PR #1088, opened 2020-04-21
Add more logging for DB connection pool status

labkey-jeckels:
#### Rationale
LabKey Server deadlocks are often related to the status of the DB connection pool. Thus, it's useful to better understand the status when doing a thread dump.

#### Changes
Add max, active, and idle counts to thread dumps
Add active and idle to Admin Console

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -984,6 +984,11 @@ public class DbScope
     {
         synchronized (_transaction)
         {
+            log.info("Data source " + toString() +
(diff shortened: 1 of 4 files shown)


labkey-jeckels:
And before you ask me to convert to use DOM builders, I tried and failed. That HTML is a mess and I don't understand what the page is trying to tell me related to the external schema. At least we're slightly more robust now.

labkey-adam:
I'm not sure I accept your code review request under such confining restrictions. Especially when couched as a thinly veiled criticism of my previous hard work.

labkey-adam (approved):

PR #1089, opened 2020-04-22
Better file cleanup and unsupported operation warning

labkey-martyp:
#### Rationale
Issue 40177: Transformation Script doesn't produce warnings as expected and deletes the input file.

#### Related Pull Requests
None

#### Changes
- On assay transform script error do better clean up of files left on server, but do not delete original file if it was already on the server
- Throw exception on assay background import with warnings as unsupported option

### DIFF
--- a/api/src/org/labkey/api/qc/TsvDataExchangeHandler.java
+++ b/api/src/org/labkey/api/qc/TsvDataExchangeHandler.java
@@ -501,7 +501,45 @@ public class TsvDataExchangeHandler implements DataExchangeHandler
         return info;
     }
 
-    public void processWarningsOutput(DefaultTransformResult result, Map<String, String> transformedProps, RunInfo info, String errorFile, List<File> files) throws ValidationException
+    private void cleanFiles(RunInfo info, DefaultTransformResult result, List<File> files, @Nullable String errorFile, @Nullable File transformedFile)
+    {
+        // If the assay does not have the "save script files for debugging" option selected, then do file cleanup
+        if (!info.isSaveScriptFiles())
+        {
+            // If file was not already on the server, erase original files from working directory
+            if (info.getOriginalFileLocation() == null)
+            {
+                for (File file : result.getUploadedFiles())
+                {
+                    FileUtils.deleteQuietly(file);
+                }
+            }
+
+            // More file cleanup on error
+            for (File file : files)
+            {
+                FileUtils.deleteQuietly(file);
+            }
+
+            if (errorFile != null)
+            {
+                FileUtils.deleteQuietly(new File(errorFile));
+            }
+
+            if (transformedFile != null)
+            {
+                FileUtils.deleteQuietly(transformedFile);
+            }
+        }
+    }
+
+    private void processWarningsOutput(
+            DefaultTransformResult result,
+            @Nullable Map<String, String> transformedProps,
+            RunInfo info,
+            @Nullable String errorFile,
+            @Nullable File transformedFile,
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/qc/TsvDataExchangeHandler.java
```diff
@@ -539,6 +577,16 @@ public void processWarningsOutput(DefaultTransformResult result, Map<String, Str
         // Display warnings case
         if(null != info.getWarningSevLevel() && info.getWarningSevLevel().equals(errLevel.WARN.name()) && null != maxSeverity && maxSeverity.equals(errLevel.WARN.name()))
         {
+            // Running in background does not support warnings
+            if(info.isBackgroundUpload())
+            {
+                cleanFiles(info, result, files, errorFile, transformedFile);
+                ValidationException exception = new ValidationException();
+                exception.addFieldError("Transform", "Background assay import does not support warnings.");
```
labkey-kevink: Should this be lower-case 'transform' and 'warning' to match the ERROR branch?

api/src/org/labkey/api/qc/TsvDataExchangeHandler.java
```diff
@@ -539,6 +577,16 @@ public void processWarningsOutput(DefaultTransformResult result, Map<String, Str
...
         if(null != info.getWarningSevLevel() && info.getWarningSevLevel().equals(errLevel.WARN.name()) && null != maxSeverity && maxSeverity.equals(errLevel.WARN.name()))
         {
+            // Running in background does not support warnings
+            if(info.isBackgroundUpload())
+            {
+                cleanFiles(info, result, files, errorFile, transformedFile);
+                ValidationException exception = new ValidationException();
+                exception.addFieldError("Transform", "Background assay import does not support warnings.");
+                exception.addFieldError("Warning", warning);
+                throw exception;
```
labkey-kevink, line 587: Throwing seems harsh for a warning, but I'm not sure.  As you said in your email, the warnings are meant to be displayed to an end user for confirmation before continuing with the import.  Since we don't have a user in the background import, maybe we can just log the warning to the pipeline log and continue with the import?

labkey-kevink (changes requested):
Looks good.

api/src/org/labkey/api/qc/TsvDataExchangeHandler.java
labkey-martyp: So my thinking on that was, the warnings are really a step where the transform script detects that there might be an issue with the data and requires a human to look at it to proceed with import.  If running in the background and that warning is not going to warn them, I thought it safer just to throw an error that the transform script is trying to raise this flag for human verification and that is not supported.  Seems like just flagging this as a warning in the logs and proceeding might allow bad data to be imported, and the warning might just get buried in the logs.  Let me know if you think there are other factors I should take into account on that

api/src/org/labkey/api/qc/TsvDataExchangeHandler.java
labkey-martyp: Fixed

labkey-kevink (approved):

api/src/org/labkey/api/qc/TsvDataExchangeHandler.java
labkey-jeckels: At least for the initial use case at SNPRC, I agree with Marty that the intent was to have a human explicitly approve bypassing the warning. I suspect that most of the warnings in the log file would never be noticed by a person.

PR #1090, opened 2020-04-22
Use DbSequence to back RowId for materials table

labkey-susanh:
#### Rationale
When doing file imports of samples, we want to be able to create detailed audit logs that use the RowId instead of the LSID as the identifier for the sample since the LSID is cumbersome, unknown in the Sample Manager application and, because it references the name of the sample, can be confusing if we get to a point where a sample can be renamed.  During file imports, the RowId is not available at the time the audit logs are created and would require a reselect.  Import of samples is already expensive enough as is, so we opt for using a Java-managed DbSequence for the RowId, the value of which is readily available during import.

#### Changes
* Upgrade script that removes the auto-increment property from RowId
* Modification to `tableInfo.xsd` to allow columns to be designated as having DbSequences and whether the sequence is "global" (i.e., associated with the root container) or not 
* Modifications to SimpleTranslator DataIterator to automatically add DbSequence columns for columns designated as having such sequences

### DIFF
--- a/api/src/org/labkey/api/data/TableInfo.java
+++ b/api/src/org/labkey/api/data/TableInfo.java
@@ -195,6 +195,8 @@ public interface TableInfo extends TableDescription, HasPermission, SchemaTreeNo
 
     Set<String> getColumnNameSet();
 
+    String getDbSequenceName(String colName);
+
     /**
      * Return a list of ColumnInfos that make up the extended set of
      * columns that could be considered a part of this table by default.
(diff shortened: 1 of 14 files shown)


api/src/org/labkey/api/data/TableInfo.java
```diff
@@ -195,6 +195,8 @@ public static IndexType getForXmlIndexType(org.labkey.data.xml.IndexType.Type.En
 
     Set<String> getColumnNameSet();
 
+    String getDbSequenceName(String colName);
+
```
labkey-adam, line 202: I'm sure you've considered this, but I'll ask anyway... would it make sense to provide a default implementation here? Like `getSchema().getName() + ":" + getName() + ":" + columnName).toLowerCase();`

labkey-adam (approved):

labkey-kevink (approved):

api/src/org/labkey/api/data/TableInfo.java
labkey-susanh: Yes, done.

PR #1058, opened 2020-04-11
Allow CreateContainerAction to enable additional modules

labkey-tchad:
This will allow us to add more functionality the tests' APIContainerHelper

PR #1094, opened 2020-04-23
Revert subtle `CreateContainerAction` behavior change

labkey-tchad:
#### Rationale
The previous change caused new subfolders to not inherit active modules from their parent in some cases. This is causing some `trialshare` tests to fail.

#### Related Pull Requests
Initial breaking change: https://github.com/LabKey/platform/pull/1058
First fix attempt: https://github.com/LabKey/platform/pull/1073

#### Changes
* Match the [original `CreateContainerAction` behavior](https://github.com/LabKey/platform/blob/9346006e632d529c05fa1d0d1db55281df585dab/core/src/org/labkey/core/CoreController.java#L745) exactly unless the `ensureModules` property is provided

### DIFF
--- a/core/src/org/labkey/core/CoreController.java
+++ b/core/src/org/labkey/core/CoreController.java
@@ -749,12 +749,8 @@ public class CoreController extends SpringActionController
                 {
                     folderType = FolderTypeManager.get().getFolderType(folderTypeName);
                 }
-                if (folderType == null)
-                {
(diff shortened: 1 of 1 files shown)


labkey-tchad:
TrialShare tests pass with this change: https://teamcity.labkey.org/viewLog.html?buildId=976384

labkey-jeckels (approved):

PR #1096, opened 2020-04-23
Item 6684: Sample Timeline - UI Dev 2 + Assay timeline events

XingY:
#### Rationale
Small refactoring to allow samplemanagement-getTimelineAction to generate meaningful labels for different event types. Also moved ExperimentAuditEvent to api so it can be used for assay timeline events

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/249

#### Changes
* added SampleTimelineEventType enum
* move ExperimentAuditEvent to api


api/src/org/labkey/api/audit/SampleTimelineAuditEvent.java
```diff
@@ -1,12 +1,66 @@
...
+        public String getActionLabel()
+        {
+            return _actionLabel;
+        }
+
+        public static SampleTimelineEventType getTypeFromEvent(@NotNull SampleTimelineAuditEvent event)
+        {
+            for (SampleTimelineEventType type : SampleTimelineEventType.values())
+            {
+                if (event.getComment().equals(type.getComment()))
```
labkey-susanh: Matching on the comment string seems fragile.  I wonder if we shouldn't just store the event type in the audit event.

labkey-susanh (changes requested):

labkey-susanh (approved):

labkey-susanh (approved):

PR #1097, opened 2020-04-23
Issue 40123: Cannot view the lineage of a source on MS-SQL when medImmune module is included

cnathe:
#### Rationale
Recent Sample Manager tests have exposed an issue on the lineage graph where the source (i.e. data class) details aren't being loaded because the lineage api response does not include the schema/query/etc. details. This only repros when the medImmune module is included in the build because that module claims the "Data" namespace which cause the ExpDataImpl.getQueryRowReference() to return the wrong response for this data class case.

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/55

#### Changes
* ExpDataImpl.getQueryRowReference() fix to check for null QueryRowReference from the DataType case
* Update @labkey/api version with fix for medImmune module fix for InvokePipeAnalysisTest failure

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpDataImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpDataImpl.java
@@ -152,6 +152,7 @@ public class ExpDataImpl extends AbstractRunItemImpl<Data> implements ExpData
         if (dc != null)
             return new QueryRowReference(getContainer(), ExpSchema.SCHEMA_EXP_DATA, dc.getName(), FieldKey.fromParts(ExpDataTable.Column.RowId), getRowId());
 
+        // Issue 40123: see MedImmuneDataHandler MEDIMMUNE_DATA_TYPE, this claims the "Data" namespace
         DataType type = getDataType();
         if (type != null)
         {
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/api/ExpDataImpl.java
```diff
@@ -148,15 +148,19 @@ public ActionURL detailsURL()
...
-        if (type != null)
-            return type.getQueryRowReference(this);
-
         ExpDataClassImpl dc = getDataClass();
         if (dc != null)
             return new QueryRowReference(getContainer(), ExpSchema.SCHEMA_EXP_DATA, dc.getName(), FieldKey.fromParts(ExpDataTable.Column.RowId), getRowId());
-        else
-            return new QueryRowReference(getContainer(), ExpSchema.SCHEMA_EXP, ExpSchema.TableType.Data.name(), FieldKey.fromParts(ExpDataTable.Column.RowId), getRowId());
+
+        DataType type = getDataType();
```
labkey-susanh, line 156: Just looking at this code without knowing the medImmune context, I still have no idea why this fixes the problem, so it might be a good idea to add a comment as to why this matters.

labkey-susanh (commented):
Deferring to others for approval on this.  Just a suggestion.

labkey-kevink (approved):

experiment/src/org/labkey/experiment/api/ExpDataImpl.java
cnathe: Added comment. Thanks for the suggestion.

labkey-nicka (approved):

PR #1102, opened 2020-04-28
Experiment: Run graph details for steps, inputs, and outputs

labkey-nicka:
#### Rationale
This PR incorporates changes to the Experiment Run Graph to allow for deeper inspection of the lineage for a particular run. Experiment Runs will display Run Step information, along with Data Inputs & Outputs, and Material Inputs & Outputs.

#### Related Pull Requests
* LabKey/labkey-ui-components#237
* LabKey/biologics#582
* LabKey/sampleManagement#254
* LabKey/commonAssays#163
* LabKey/targetedms#172

#### Changes
* Bump `@labkey/components` to incorporate UI updates.
* Expose a new property, `expType`, via the `experiment-lineage.api` which provides a more deterministic experiment type than the `type` property. (author: @labkey-kevink).
* Add ExperimentProtocolHandler to allow lineage API to provide query row reference for specific protocol types (author: @labkey-kevink).
* Use `ExperimentRunType` to provide query row reference for runs (author: @labkey-kevink).


api/src/org/labkey/api/exp/api/ExperimentJSONConverter.java
```diff
@@ -67,6 +67,8 @@
     public static final String NAME = "name";
     public static final String LSID = "lsid";
     public static final String CPAS_TYPE = "cpasType";
+    // Matches the expType parameted used in the linage api: "Data", "Material", "ExperimentRun", "Object"
```
labkey-jeckels:
```suggestion
    // Matches the expType parameter used in the linage api: "Data", "Material", "ExperimentRun", "Object"

```

api/src/org/labkey/api/exp/api/ExperimentService.java
```diff
@@ -760,9 +761,11 @@ ExpRun derive(Map<ExpMaterial, String> inputMaterials, Map<ExpData, String> inpu
 
     void registerProtocolImplementation(ProtocolImplementation impl);
 
+    void registerProtocolHandler(ExperimentProtocolHandler handler);
```
labkey-jeckels, line 764: Since ProtocolImplementation is an implementation of ExperimentProtocolHandler, what's the expectation for registering one? Register as both? Just as a ProtocolImplementation?

experiment/src/org/labkey/experiment/api/ExpProtocolApplicationImpl.java
```diff
@@ -66,6 +72,32 @@ public ActionURL detailsURL()
         return null;
     }
 
+    @Override
+    public @Nullable QueryRowReference getQueryRowReference()
+    {
+        ExpProtocolImpl protocol = getProtocol();
+        if (protocol != null)
+        {
+            ProtocolImplementation impl = protocol.getImplementation();
```
labkey-jeckels: Can this call getQueryRowReference() on the protocol? Seems like duplicate code

labkey-jeckels (changes requested):
A few small suggestions, but overall looks good. Thanks for the JavaDoc!

api/src/org/labkey/api/exp/api/ExperimentService.java
labkey-kevink: I would like to combine the ProtocolImplementation and ExperimentProtocolHandler implementations and I started to do this refactor, but it was a larger change than I wanted to make right now.  The ProtocolImplementation is geared more toward ExperimentRun protocol and requires setting a property on the protocol (which would have required an upgrade script for SampleManager protocols).  ExperimentProtocolHandler can be used for ProtocolApplication protocols (see SamplesWorkflowTaskProtocolHandler.)

labkey-jeckels (approved):

PR #1105, opened 2020-04-28
7078: Issue Designer Server Side API Updates

RosalineP:
#### Rationale
As part of the effort to replace the GWT domain designers, the Issue Designer is being converted to use a stylish new frontend React page, and its server-side API calls are being rewired to remove dependency on GWT and use a new java bean to handle issue definition properties processing.

#### Changes
* Adds IssuesDomainKindProperties bean
* Updates AbstractIssuesListDefDomainKind.createDomain() 
* Add AbstractIssuesListDefDomainKind.updateDomain()


internal/src/org/labkey/api/issues/IssuesDomainKindProperties.java
```diff
@@ -0,0 +1,82 @@
...
+
+public class IssuesDomainKindProperties
+{
+    protected String _issueDefName;
+    protected String _singularItemName;
+    protected String _pluralItemName;
+    protected String _commentSortDirection;
+
+    protected Integer _assignedToGroup;
+    protected Integer _assignedToUser;
```
labkey-klum: Shouldn't the member variables be private by default? I would think we want to force subclasses to use the accessor functions.

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
```diff
@@ -44,6 +55,105 @@
...
+            names.singularName = properties.getSingularItemName();
+            names.pluralName = properties.getPluralItemName();
+
+            String issueDefName = properties.getIssueDefName();
+
+            if (issueDefName == null || issueDefName.length() == 0)
+                throw new IllegalArgumentException("Issue definition name must not be null.");
+
+            IssueManager.saveEntryTypeNames(container, issueDefName, names);
+            this.saveIssueProperties(container, properties, issueDefName);
```
labkey-klum: using the 'this' keyword is redundant

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
```diff
@@ -44,6 +55,105 @@
...
+
+            User user = (properties.getAssignedToUser() != null) ? UserManager.getUser(properties.getAssignedToUser()) : null;
```
labkey-klum: I don't think we should use the default 'assignedToUser' to update the domain, we should probably pass in the actual user from the request.

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
```diff
@@ -44,6 +55,105 @@
...
+        }
+    }
+
+    // Note to reviewer: This was intended to hold update and create validation in common, but this seems like the only
+    // bit of validation that makes sense to do in here. Should I just factor out the check, or keep this function as is?
+    @Override
+    public void validateIssuesProperties(IssuesDomainKindProperties properties)
+    {
+        String name = properties.getIssueDefName();
+        if (name == null || name.length() == 0)
```
labkey-klum: StringUtils.isBlank is handy for these types of checks

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
```diff
@@ -44,6 +55,105 @@
...
+
+        try (DbScope.Transaction transaction = IssuesSchema.getInstance().getSchema().getScope().ensureTransaction())
+        {
+            names.singularName = properties.getSingularItemName();
+            names.pluralName = properties.getPluralItemName();
+
+            String issueDefName = properties.getIssueDefName();
+
+            if (issueDefName == null || issueDefName.length() == 0)
+                throw new IllegalArgumentException("Issue definition name must not be null.");
```
labkey-klum: A bit of confusing mix in this method between throwing exceptions and returning a ValidationException, it might be better to add this error to the returned exception to be consistent.

internal/src/org/labkey/api/issues/IssuesDomainKindProperties.java
RosalineP: True! Updated in Issues and Dataclass

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
RosalineP: Updated 👍

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
RosalineP: Oop. That's definitely a remnant from an un-factored out saveIssueProperties(), thanks for the catch

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
RosalineP: Done!

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
```diff
@@ -44,6 +56,104 @@
...
+    public void saveIssueProperties(Container container, IssuesDomainKindProperties properties, String name)
+    {
+        IssueManager.saveCommentSortDirection(container, name, Sort.SortDirection.fromString(properties.getCommentSortDirection()));
+
+        Group group = null;
+        if (properties.getAssignedToGroup() != null)
+        {
+            group = SecurityManager.getGroup(properties.getAssignedToGroup());
+            if (null == group)
+                throw new IllegalArgumentException("Group not found.");
```
labkey-bpatel: I'd include which group not found -- ' properties.getAssignedToGroup() +  " group not found." 
Plus something like refer to core.Groups for a valid list of groups.

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
```diff
@@ -44,6 +56,104 @@
...
+        }
+        IssueManager.saveAssignedToGroup(container, name, group);
+
+        User user = null;
+        if (properties.getAssignedToUser() != null)
+        {
+            user = UserManager.getUser(properties.getAssignedToUser());
+            if (null == user)
+                throw new IllegalArgumentException("User not found.");
+        }
```
labkey-bpatel, line 147: I'd include which user not found -- ' properties.getAssignedToUser() +  " user not found." 
Plus something like refer to core.Users for a valid list of Users.

labkey-bpatel (approved):
Looks great overall! Couple minor comments.

PR #1106, opened 2020-04-28
Property to disable loading resources from module source

labkey-tchad:
This serves a similar purpose to the `disableRecompileJsp` property.
Prevents generated resources from leaking between branches on TeamCity.

#### Rationale
Occasionally, on TeamCity, a feature branch build will leave generated files that get picked up by subsequent test suites that run on the same build agent. This can cause numerous spurious failures depending on the tests being run and the changes in the feature branch.

#### Related Pull Requests
* Gradle plugin 1.12.0 will set this flag when running on TeamCity (No PR)

### DIFF
--- a/api/src/org/labkey/api/module/DefaultModule.java
+++ b/api/src/org/labkey/api/module/DefaultModule.java
@@ -920,7 +920,10 @@ public abstract class DefaultModule implements Module, ApplicationContextAware
 
     public final void setSourcePath(String sourcePath)
     {
-        _sourcePath = sourcePath;
+        if (!AppProps.getInstance().isIgnoreModuleSource())
+        {
+            _sourcePath = sourcePath;
+        }
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


labkey-matthewb (commented):
Really it would be nice if setSourcePath() acted like a dumb bean and just reflected what's in the module.properties/xml.  Ideally the only place you'd have to touch is DefaultModule.computeResourceDirectory().  I would not be surprised if there are improper usages of getSourcePath() however.

labkey-tchad:
@labkey-matthewb That was what I did initially. Then I poked around and got worried about `DefaultModule.getStaticFileDirectories`, `ModuleLoader.registerResourcePrefix`,  `ModuleResourceProvider.AllModuleResourcesRoot.find`, and other odd usages of `getSourcePath` that might be around.

labkey-matthewb:
> @labkey-matthewb That was what I did initially. Then I poked around and got worried about `DefaultModule.getStaticFileDirectories`, `ModuleLoader.registerResourcePrefix`, `ModuleResourceProvider.AllModuleResourcesRoot.find`, and other odd usages of `getSourcePath` that might be around.

OK, maybe add a big comment to the effect that this code is modifying setSourcePath() because there isn't another good chokepoint and maybe cross reference computeResourceDirectory()

labkey-matthewb (approved):

PR #1107, opened 2020-04-29
Server-side 'Projects' web part

labkey-matthewb:
#### Rationale
Server-side version of "projects" web part.
Projects web part is really slow and on the "welcome" page of many sites, including labkey.org.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/331


core/src/org/labkey/core/CoreModule.java
```diff
@@ -571,7 +571,7 @@ public WebPartView getWebPartView(@NotNull ViewContext portalCtx, @NotNull WebPa
                     if (portalCtx.hasPermission(getClass().getName(), AdminPermission.class))
                     {
                         NavTree customize = new NavTree("");
-                        customize.setScript("customizeProjectWebpart(" + webPart.getRowId() + ", '" + webPart.getPageId() + "', " + webPart.getIndex() + ");");
+                        customize.setScript("(customizeProjectWebpart" + webPart.getRowId()+"||customizeProjectWebpart)(" + webPart.getRowId() + ", " + PageFlowUtil.jsString(webPart.getPageId()) + ", " + webPart.getIndex() + ");");
```
cnathe: Is the `||customizeProjectWebpart` part needed here? Looking at the changes in projects.jsp, I don't think we'd ever have a `customizeProjectWebpart()` function to call.

cnathe: Also, there is another usages of this for the subfolder case a few lines down in this file, at line 601, which would need the same change. I tested this by adding the "Subfolders" webpart locally to one of my containers and when going to the Customize option for that webpart, I get a JS error: `ReferenceError: customizeProjectWebpart is not defined`

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -1,444 +1,380 @@
...
+    if (containers.isEmpty())
+    {
+        if (getUser().isGuest())
+        {
+            %>Please log in to view the <%=h(properties.get("noun").toLowerCase())%> list.<%
+        }
+        else {
+            %>No <%=h(properties.get("noun").toLowerCase())%>s to display.<%
+        }
+        return;
```
cnathe: When I add the Projects webpart to my container and then use the Customize option to change the "Folders to Display" to a specific folder which I know has subfolders, I am getting a rendered webpart that says "No subfolders to display" and it also doesn't let me go back into the Customize dialog (resulting in a JS error). The second issue looks to be because of this return statement. The JSP never ends up adding the JavaScript code to the page because of this. I'm not sure yet what is causing the first issue.

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -1,444 +1,380 @@
...
+    }
+%>
+<%
+    JspView<Portal.WebPart> me = (JspView<Portal.WebPart>) HttpView.currentView();
+    Portal.WebPart webPart = me.getModelBean();
+    int webPartId = webPart.getRowId();
+    boolean isRootAdmin = getUser().hasRootAdminPermission();
+    boolean hasPermission;
+
+    Map<String,String> defaultProperties = Map.of("containerTypes", "project", "containerFilter", "CurrentAndSiblings", "hasCreateButton", "false", "iconSize", "large", "labelPosition", "bottom", "noun", "Project");
```
cnathe: There doesn't look to be a usage of the `hasCreateButton` prop. Was that a typo? the original code looks to have a `hideCreateButton` prop which defaults to false, so I assume you meant that here?

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -1,444 +1,380 @@
...
+    }
+
+    SimpleFilter filter = new SimpleFilter();
+    filter.addInClause(new FieldKey(null,"containerType"), Arrays.asList(StringUtils.split(properties.get("containerTypes"),",")));
+    filter.addClause(new SimpleFilter.InClause(new FieldKey(null,"entityId"), Set.of(ContainerManager.getHomeContainer().getId(), ContainerManager.getSharedContainer().getId()), false, true));
+    filter.addClause(new SimpleFilter.SQLClause(new SQLFragment("Name NOT LIKE '\\_%' ESCAPE '\\'")));
+    ContainerFilter cf = ContainerFilter.getContainerFilterByName(properties.get("containerFilter"),getUser());
+    QuerySchema core = DefaultSchema.get(getUser(),target).getSchema("core");
+    // consider: could use ContainerManager or ContainerFilter.getIds()
+    var containers = new TableSelector(core.getTable("Containers",cf), Set.of("name","entityId"),filter, (Sort)null).getArrayList(Map.class);
```
cnathe: Whenever I customize the projects webpart and select a specific folder other then the current one, I get the "No projects to display" message even when I know that my selected specific folder does have subfolders. Could there be a containerPath property usage missing in this container filter somewhere?

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -1,444 +1,380 @@
...
+                        config.containerFilter = directDescendants ? 'CurrentAndFirstChildren' : 'CurrentAndSubfolders';
+                    }
+
+                    config.hideCreateButton =  btn.up('window').down('#hideCreateButton').getValue();
+                    config.iconSize = btn.up('window').down('#style').getValue().style.iconSize;
+                    config.labelPosition = btn.up('window').down('#style').getValue().style.labelPosition;
+                    config.title = btn.up('window').down('#title').getValue();
+                    config.webPartId = webpartId;
+
+                    config['@hideCreateButton'] = '1'; // for spring binding of boolean
```
cnathe, line 378: Should this use the `config.hideCreateButton` value instead of always hard coding to '1'?

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -1,444 +1,380 @@
...
+                    },{
+                        xtype: 'checkbox',
+                        boxLabel: 'Include Workbooks',
+                        disabled: (config.containerTypes && config.containerTypes.match(/project/)) || !config.containerPath,
+                        checked: (config.containerTypes.match(/project/) || config.containerTypes.match(/workbook/)),
+                        itemId: 'includeWorkbooks'
+                    },{
+                        xtype: 'checkbox',
+                        boxLabel: 'Hide Create Button',
+                        checked: config.hideCreateButton,
```
cnathe: I had created a Projects webpart on develop and customized it for a specific folder, but left that webpart as "unchecked for the hide create button" option. However, when I switch to this fb, the customize dialog shows this "hide create button" as checked, even though the webpart correctly still shows the create button.

core/src/org/labkey/core/project/projects.jsp
cnathe: Actually this is the behavior I see for using "Subfolders" at the project level. When I use the "Subfolders" option in a non-project folder, the webpart looks to show itself (which is not what the previous implementation was showing).

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -1,444 +1,380 @@
...
+        Container c = ContainerManager.getForId((String)m.get("entityId"));
+        if (null != c)
+        {
+            HtmlString projectName = HtmlString.of(c.getProject().getName());
+            // data-project can be use in style sheet to hide projects e.g.
+            // <style>div[data-project="StudyVerifyProject"]{display:none !important;}</style>
+            if (details) {
+                %><div data-project="<%=projectName%>" class="thumb-wrap"><div style="width: 100%;" class="tool-icon thumb-wrap thumb-wrap-side"><a href="<%=h(c.getStartURL(getUser()))%>"><div class="thumb-img-side"><span class="fa fa-folder-open fa-lg"></span></div><span class="thumb-label-side"><%=h(c.getName())%></span></a></div></div><%
+            } else {
+                %><div data-project="<%=projectName%>" style="display: inline-block;" class="thumb-wrap"><div style="width: <%=width%>;" class="tool-icon thumb-wrap thumb-wrap-bottom"><a href="<%=h(c.getStartURL(getUser()))%>"><div class="thumb-img-bottom"><span class="fa fa-folder-open <%=faX%>"></span></div><span class="thumb-label-bottom"><%=h(c.getName())%></span></a></div></div><%
```
cnathe: The previous implementation was showing the DisplayName for a container in this webpart. Now it looks to have been changed to showing the container name, was that an expected change?

cnathe (changes requested):
I found several issues comparing this implementation with the implementation in develop. I noted those here but will hold off on the rest of the code review until those items / questions are addressed.

core/src/org/labkey/core/project/projects.jsp
labkey-matthewb: The @ thing is a spring trick for dealing with checkboxes.  Just from the URL you can't tell the difference between input element at all, and an unchecked checkbox.  This tells spring there is a checkbox on the page called hideCreateButton.

core/src/org/labkey/core/project/projects.jsp
labkey-matthewb: changed return to else {}

core/src/org/labkey/core/project/projects.jsp
labkey-matthewb: Nope, fixed.

core/src/org/labkey/core/project/projects.jsp
labkey-matthewb: ";" delimter not ","

core/src/org/labkey/core/project/projects.jsp
labkey-matthewb: maybe fixed with ==="true" check?

labkey-matthewb:
Thanks Cory!

core/src/org/labkey/core/project/projects.jsp
cnathe: Got it. Cool. Good to know.

core/src/org/labkey/core/project/projects.jsp
cnathe: Yup, looks like that took care of it. Thanks.

cnathe:
@labkey-matthewb I didn't find and existing tests that use the "Projects" webpart, but there was one that used the "Subfolders" wepbart (SubfolderWebPartTest.java). I ran that test on this fb and it looks like there is one failure: https://teamcity.labkey.org/viewLog.html?buildId=993039&tab=buildResultsDiv&buildTypeId=LabkeyTrunk_DailyPostgres&branch_LabkeyTrunk=projects_webpart

cnathe:
> @labkey-matthewb I didn't find and existing tests that use the "Projects" webpart, but there was one that used the "Subfolders" wepbart (SubfolderWebPartTest.java). I ran that test on this fb and it looks like there is one failure: https://teamcity.labkey.org/viewLog.html?buildId=993039&tab=buildResultsDiv&buildTypeId=LabkeyTrunk_DailyPostgres&branch_LabkeyTrunk=projects_webpart

SubfolderWebPartTest now passing on this fb on TeamCity

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -1,444 +1,385 @@
...
+
+    SimpleFilter filter = new SimpleFilter();
+    filter.addInClause(new FieldKey(null,"containerType"), Arrays.asList(StringUtils.split(properties.get("containerTypes"),";")));
+    filter.addClause(new SimpleFilter.InClause(new FieldKey(null,"entityId"), Set.of(ContainerManager.getHomeContainer().getId(), ContainerManager.getSharedContainer().getId()), false, true));
+    filter.addClause(new SimpleFilter.SQLClause(new SQLFragment("Name NOT LIKE '\\_%' ESCAPE '\\'")));
+    ContainerFilter cf = ContainerFilter.getContainerFilterByName(properties.get("containerFilter"),getUser());
+    QuerySchema core = DefaultSchema.get(getUser(),target).getSchema("core");
+    // consider: could use ContainerManager or ContainerFilter.getIds()
+    var containers = new TableSelector(core.getTable("Containers",cf), Set.of("name","entityId"),filter, (Sort)null).getArrayList(Map.class);
+    containers.sort((m1, m2) -> StringUtils.compareIgnoreCase(String.valueOf(m1.get("name")),String.valueOf(m2.get("name"))));
```
cnathe: The previous implementation took into account the sort order of projects/folders and the display name (see code here). This looks to be sorting just by name, was that an intentional change?

```
        sorters: [{
            sorterFn: function(rec1, rec2){
                // Issue 36731: use natural sort for projects webpart to match the nav menu ordering
                var sort1 = rec1.get('SortOrder') + rec1.get('DisplayName');
                var sort2 = rec2.get('SortOrder') + rec2.get('DisplayName');
                return LABKEY.internal.SortUtil.naturalSort(sort1, sort2);
            }
        }],
```

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -1,444 +1,385 @@
...
+        </div>
+        </div>
+<%
+    } // !containers.isEmpty()
+%>
+    <div><%
+        if (Boolean.TRUE != JdbcType.BOOLEAN.convert(properties.get("hideCreateButton")))
+        {
+            if ((StringUtils.equals("Project",noun) && isRootAdmin) ||
+                StringUtils.equals("Subfolder",noun) && getContainer().hasPermission(getUser(), AdminPermission.class))
```
cnathe: If the webpart is configured as a different, specific folder, shouldn't we be using that container for this permissions check instead of the `getContainer()` (i.e. current container)?

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -1,444 +1,385 @@
...
+            {
+                Container c = getContainer();
```
cnathe: Found an issue with this during manual testing. If the webpart is configured to use a different, specific folder then this looks to always use the current container for the path of the create new button instead of that other container (which the previous implementation was doing).

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -1,444 +1,385 @@
...
+    else if (!hasPermission)
+    {
+        %>You do not have permission to view this folder.<%
+        return;
+    }
+
+    SimpleFilter filter = new SimpleFilter();
+    filter.addInClause(new FieldKey(null,"containerType"), Arrays.asList(StringUtils.split(properties.get("containerTypes"),";")));
+    filter.addClause(new SimpleFilter.InClause(new FieldKey(null,"entityId"), Set.of(ContainerManager.getHomeContainer().getId(), ContainerManager.getSharedContainer().getId()), false, true));
+    filter.addClause(new SimpleFilter.SQLClause(new SQLFragment("Name NOT LIKE '\\_%' ESCAPE '\\'")));
```
cnathe, line 105: Found an issue during manual testing where if I select a specific folder which I know does not have any subfolders and I also check the box for "Include direct children only", the previous implementation would show the "No folders..." message but this shows the selected folder in the webpart. I believe the previous implementation was handling this with the following code in a client side filtering:
```
            //NOTE: path is not directly filterable, so we settle for Client-side filtering
            panel.store.on('load', function () {
                var path = this.containerPath || LABKEY.Security.currentContainer.path;
                this.filterBy(function (rec) {
                    return rec.get('Path') !== path;
                })
            }, null, {single: true});
```

cnathe (changes requested):

core/src/org/labkey/core/project/projects.jsp
labkey-matthewb: Hmm, I think the old code got this wrong too.

cnathe (approved):

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -1,444 +1,398 @@
...
+                %><%=button("Create New " + noun).href(urlProvider(AdminUrls.class).getCreateFolderURL(c, getActionURL()))%><%
+            }
+        }%>
+    </div>
+
+<script type="text/javascript">
+
+    function customizeProjectWebpart<%=webPartId%>(webpartId, pageId, index)
+    {
+        LABKEY.requiresScript(["Ext4","Ext4ClientApi"], function()
```
labkey-nicka, line 191: Why switch this to using `LABKEY.requiresScript()` instead of `dependencies.add("Ext4ClientApi")` like it was before?

labkey-nicka:
I'm just curious, was this slow due to the actual rendering performance or due to the API backing it? If it is the latter, then does this implementation result in a faster page load? Are we trading async waiting time for waiting for full page to load time?

core/src/org/labkey/core/project/projects.jsp
labkey-matthewb: The answer to this is part of the answer to you next question.  This component pulls a lot of dependencies onto this page.  In the new version, I only pull in the dependency when the customize form is rendered.  I could have also rewritten it to not use Ext4, but it works.

labkey-matthewb:
> I'm just curious, was this slow due to the actual rendering performance or due to the API backing it? If it is the latter, then does this implementation result in a faster page load? Are we trading async waiting time for waiting for full page to load time?

1) Lots of resource dependencies (Ext4 + others)
2) Async request for container info
3) Page re-layout

core/src/org/labkey/core/project/projects.jsp
labkey-nicka: Alright, I'd still recommend using `dependencies.add("Ext4")` over `LABKEY.requiresScript("Ext4")` since we're already in a JSP and support the mechanism.

PR #1113, opened 2020-04-30
Issue 40329: Sample Manager source samples: limit to 1st generation

XingY:
#### Rationale
Lineage operators returns N generations of parents and children. This PR adds operators that would return only the immediate children or parents. 

#### Related Pull Requests
* 

#### Changes
* added ImmediateChildOfCompareType, ImmediateChildOfMethod, ImmediateParentOfCompareType, ImmediateParentOfMethod


labkey-kevink (changes requested):
As we discussed over chat, let's see if it's possible to parameterize the exp:childof filer instead of adding more filter types.

PR #1115, opened 2020-04-30
Issue 40333: generate rowId from DbSequence when doing single-sample inserts

labkey-susanh:
#### Rationale
We switched to using a DbSequence to back the RowId for the exp.Materials table to better support detailed logging.  This code path for inserting single samples that does not use DataIterators was missed in that conversion

#### Related Pull Requests

#### Changes
*[ Issue 40333](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40333): Generate a rowId before using Table.insert


labkey-kevink (approved):

labkey-susanh:
Targeting wrong branch.

PR #1124, opened 2020-05-01
Fix populating exp.material for study.specimens

labkey-matthewb:
#### Rationale
With exp.material.rowid changes, we need to change implementation for populating for study specimens


labkey-susanh (changes requested):
Code changes look fine, but I think we want to target 20.5 for this fix.

PR #1126, opened 2020-05-02
Fix populating exp.material for study.specimens post material.rowid changes

labkey-matthewb:
#### Rationale
Re-enable specimen import

#### Related PRs
* https://github.com/LabKey/platform/pull/1127

PR #1127, opened 2020-05-02
Issue 40353: Remove autoincrement for rowId column of exp.materials for databases bootstrapped since the last upgrade script was implemented

labkey-susanh:
#### Rationale
Previous upgrade script mistakenly prevented the conversion of the RowId column from happening during an initial install.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1126

#### Changes
* new upgrade script that runs only if the autoIncrement property is still set and remove early exit from previous upgrade script for new installs.


labkey-adam (changes requested):
Sorry for not thinking of this while we were chatting, but could you change the previous scripts to invoke core.executeJavaInitializationCode() instead of core.executeJavaUpgradeCode()? It's a synonym, but this will make it abundantly clear which invocation needs to live forever (to remain after we consolidate these scripts into the bootstrap scripts) and which can be removed.

labkey-adam (approved):

PR #1128, opened 2020-05-02
Issue 40356: Handle null protocol when passed to AbstractAssayProvider.getPriority method

labkey-ankurj:
#### Rationale
This change avoids NPE in AbstractAssayProvider.getPriority when protocol is null.

### DIFF
--- a/assay/src/org/labkey/assay/AssayManager.java
+++ b/assay/src/org/labkey/assay/AssayManager.java
@@ -194,7 +194,7 @@ public class AssayManager implements AssayService
     @Nullable
     public AssayProvider getProvider(ExpProtocol protocol)
     {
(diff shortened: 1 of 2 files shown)


labkey-kevink (approved):

assay/src/org/labkey/assay/AssayManager.java
```diff
@@ -194,7 +194,7 @@ public AssayProvider getProvider(String providerName)
     @Nullable
     public AssayProvider getProvider(ExpProtocol protocol)
```
labkey-jeckels, line 195: Let's annotate the argument as @Nullable too

labkey-jeckels (approved):

PR #1134, opened 2020-05-04
40226: Respect "skipPrefixes" in async region

labkey-nicka:
#### Rationale
This PR fixes [40226](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40226) which describes a scenario where an asynchronous data region fails to return to the first page after applying a filter on a subsequent page.

#### Fix
Synchronous data region behavior worked in this scenario by respecting the "skip prefixes" parameter. This parameter is used to indicate any set of parameters that should be "skipped" in processing and thus no longer applied. For example, when updating a filter, the update method asks for the `offset` prefix to be skipped:

```js
var _updateFilter = function(region, filter, skipPrefixes) {
    var params = [];
    if (filter) {
        params.push([filter.getURLParameterName(region.name), filter.getURLParameterValue()]);
    }

    // Skip [OFFSET_PREFIX] plus whatever the caller wants to skip
    _setParameters(region, params, [OFFSET_PREFIX].concat(skipPrefixes));
};
```

This wasn't being respected in the async processing and so the `offset` was still being carried forward (through the next load) resulting in the user being left on the page on which they applied the filter. With this PR these skip prefixes are now processed in the async handler as well.

#### Changes
* Pass explicit "skipPrefixes" parameter to `_load()` to handle processing parameters for the next load.
* Add regression test coverage to the `DataRegionTest`.
* Replace usages of `$.each` on arrays with `Array.forEach()` and/or `Array.map()`.

### DIFF
--- a/query/webapp/QWPDemo.js
+++ b/query/webapp/QWPDemo.js
@@ -28,7 +28,8 @@
             testShowAllTotalRows: testShowAllTotalRows,
             testGetBaseFilters: testGetBaseFilters,
             testFilterOnSortColumn: testFilterOnSortColumn,
-            testButtonBarConfig: testButtonBarConfig
+            testButtonBarConfig: testButtonBarConfig,
+            testRespectExcludingPrefixes: testRespectExcludingPrefixes
         };
 
         var PAGE_OFFSET = 4;
@@ -598,7 +599,6 @@
                 },
                 listeners: {
                     render: function(qwp) {
-                        debugger;
                         if (initialLoad) {
                             initialLoad = false;
                             var filterCount = getFilterMessageBoxCount();
@@ -667,7 +667,7 @@
                                     }
                                     else if (loadCount === 3) {
                                         loadCount++;
-                                        if (assertPagingCount(dr, 3, 4)) {
+                                        if (assertPagingCount(dr, 3, TOTAL_EXPECTED_ROWS)) {
                                             LABKEY.Utils.signalWebDriverTest('testHidePagingCount');
                                         }
                                     }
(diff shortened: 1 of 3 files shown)


query/webapp/QWPDemo.js
```diff
@@ -852,6 +852,41 @@
...
+                                if (assertPagingCount(dr, 5, 6, 6)) {
+                                    loadCount++;
+
+                                    // filter to a result that includes rows on multiple page
+                                    dr.addFilter(LABKEY.Filter.create('Id', [6,5,4,3,2].join(';'), LABKEY.Filter.Types.IN));
+                                }
+                            }
+                            else if (loadCount === 1) {
+                                if (assertPagingCount(dr, 1, 2, 5)) {
+                                    LABKEY.Utils.signalWebDriverTest('testRespectExcludingPrefixes');
```
XingY, line 886: loadCount++ here?

query/webapp/QWPDemo.js
```diff
@@ -667,7 +667,7 @@
                                     }
                                     else if (loadCount === 3) {
                                         loadCount++;
-                                        if (assertPagingCount(dr, 3, 4)) {
+                                        if (assertPagingCount(dr, 3, TOTAL_EXPECTED_ROWS)) {
```
XingY, line 670: Is this change intentional? It looks odd to me that page 2 shows 2-6, but I guess that's just due to the serial of weird actions taken.

XingY (approved):

query/webapp/QWPDemo.js
labkey-nicka: Switched to updating `loadCount` before checking value.

query/webapp/QWPDemo.js
labkey-nicka: Yes, this was intentional. I'm not sure where it would show `2-6` (maybe you meant `3-6`?) and yes, it is odd. 

The test case is trying to assert that the pagination count remains hidden through different events. It uses a series of actions to test this behavior which end up showing (as a side-effect) that the original `maxRows` parameter declared on the QWP configuration is mutable by the user.

My change ends up causing the `maxRows` as specified to be "skipped" once `dr.showAllRows()` (when `loadCount === 1`) has been called which is consistent with a synchronous region's behavior.

Happy to discuss further.

PR #1137, opened 2020-05-04
Small performance fixes for getQueryDetails()

labkey-matthewb:
#### Rationale
Changes inspired by profiling getQueryDetails() as used by SM

#### Changes
cache _SqlParser objects, spending much more time creating these objects than actually parsing.  It would seem that you're supposed to hold on to them.

fixes to avoid calling new DefaultSchema() when we should be using existing object.


api/src/org/labkey/api/exp/query/SamplesSchema.java
```diff
@@ -76,11 +76,17 @@ public boolean isAvailable(DefaultSchema schema, Module module)
 
             public QuerySchema createSchema(DefaultSchema schema, Module module)
             {
-                return new SamplesSchema(schema.getUser(), schema.getContainer());
+                return new SamplesSchema(schema);
```
labkey-kevink, line 78: It looks like we call `.setDefaultSchema()` when resolving schemas in `DefaultSchema.getSchema()`.  Does that not work in this case?  Does it happen too late?

query/src/org/labkey/query/sql/SqlParser.java
```diff
@@ -1347,18 +1398,30 @@ public int LA(int i)
...
+        {
+            super(null);
             setTreeAdaptor(new LabKeyTreeAdaptor());
-            _errors = errors;
+            _errors = null;
             assert MemTracker.getInstance().put(this);
         }
 
+        public _SqlParser(String str, ArrayList<Exception> errors)
+		{
```
labkey-kevink: weird indentation

query/src/org/labkey/query/sql/SqlParser.java
```diff
@@ -116,29 +152,44 @@ public SqlParser(SqlDialect d, Container c)
         _container = c;
     }
 
+    /** see also _SqlParser.close() for AutoCloseable behavior */
+    private _SqlParser getAntlrParser()
+    {
+        synchronized (_parserPool)
+        {
+            _SqlParser ret = _parserPool.get();
```
labkey-kevink: Does this block need to lock on _parserPool if `SoftPool.get` is synchronized?

labkey-kevink (changes requested):

query/src/org/labkey/query/sql/SqlParser.java
labkey-matthewb: thanks, leftover from earlier edit.

query/src/org/labkey/query/sql/SqlParser.java
labkey-matthewb: Weird, intellij left in the tabs.

api/src/org/labkey/api/exp/query/SamplesSchema.java
labkey-matthewb: The problem isn't really here in createSchema().  In other places people call "new SampleSchema()" directly, so I wanted to make a SampleSchema(schema) constructor.  Should probably go convert those places in this PR.

PR #1140, opened 2020-05-05
Issue 38592: Rename directory to have same casing as resources directory

labkey-susanh:
#### Rationale
Using a capital-E experiment directory for some of the content copied to the web directory can cause some of the content not to be found when building on Windows.  At least that's the theory.  In any case, there's no reason to use different casing for the directories.

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/165
* https://github.com/LabKey/testAutomation/pull/328

#### Changes
* Renamed `webapp/Experiment` to `webapp/experiment`


cnathe (approved):

labkey-tchad (approved):
Will Windows folks need to do any manual steps to get this change to stick? Case-only file renames have been historically troublesome but maybe Git is better about it than SVN?

labkey-susanh:
> Will Windows folks need to do any manual steps to get this change to stick? Case-only file renames have been historically troublesome but maybe Git is better about it than SVN?

I don't really know.  @cnathe do you know?  I can send a message when the merge happens so people are at least aware (if not reading these comments as they go by).

PR #1141, opened 2020-05-05
Change domain form display options

labkey-ankurj:
#### Rationale
This PR is to pull ui-components version of domain form display options fix.

#### Related Pull Requests
https://github.com/LabKey/labkey-ui-components/pull/240

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -939,9 +939,9 @@
       "integrity": "sha1-NJIdAY//FYWYjCObiLH/TnC2adM="
     },
     "@labkey/components": {
-      "version": "0.56.3-fb-FixDomainFormDisplayOptions.1",
-      "resolved": "https://artifactory.labkey.com/artifactory/api/npm/libs-client/@labkey/components/-/@labkey/components-0.56.3-fb-FixDomainFormDisplayOptions.1.tgz",
-      "integrity": "sha1-DQkOLtSMMm/gfvKoOA6U513vtxE=",
+      "version": "0.56.3-fb-FixDomainFormDisplayOptions.2",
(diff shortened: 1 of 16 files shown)


cnathe (approved):

query/src/client/QueryMetadataEditor/QueryMetadataEditor.tsx
```diff
@@ -328,11 +328,11 @@ export class App extends PureComponent<any, Partial<IAppState>> {
...
                         useTheme={false}
                         domainFormDisplayOptions= {{
-                            showRequired: false,
+                            hideRequired: true,
                             isDragDisabled: true,
-                            showValidators: false,
+                            hideValidators: true,
                             phiLevelDisabled: true,
-                            showAddFieldsButton: false
+                            hideAddFieldsButton: true,
```
cnathe, line 335: Shouldn't this also set `hideTextOptions: true`?

cnathe (changes requested):

query/src/client/QueryMetadataEditor/QueryMetadataEditor.tsx
labkey-ankurj: it wasn't set to showTextOptions: false before, I don't recall the text options in old gwt editor. I can change if you think we need to hide them.

core/package-lock.json
```diff
@@ -11005,7 +11005,7 @@
     },
     "rgba-regex": {
       "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/rgba-regex/-/rgba-regex-1.0.0.tgz",
+      "resolved": "http://registry.npmjs.org/rgba-regex/-/rgba-regex-1.0.0.tgz",
```
cnathe: @labkey-ankurj  Need to revert this change. It usually can be fixed by clearing the node_modules for core and then rebuilding that module.

cnathe: done.

cnathe (approved):

PR #1143, opened 2020-05-05
fix module object leaks

labkey-matthewb:
#### Rationale
fix module object leaks due to FolderTypes loading WebParts from modules that may be pruned

#### Changes
* don't call registerFolderType() in Module.init()
* prune modules due to unsupported database earlier

### DIFF
--- a/api/src/org/labkey/api/module/FolderTypeManager.java
+++ b/api/src/org/labkey/api/module/FolderTypeManager.java
@@ -102,6 +102,10 @@ public class FolderTypeManager
 
     public void registerFolderType(Module sourceModule, FolderType folderType)
     {
+        if (!ModuleLoader.getInstance().isStartupInProgress() && !ModuleLoader.getInstance().isStartupComplete())
+        {
+            throw new IllegalStateException("Don't call registerFolderType() this early.  It will cause WebPart to get populated from ALL modules");
+        }
         if (_javaFolderTypes.containsKey(folderType.getName()))
         {
             String msg = "Unable to register folder type " + folderType.getName() + " from module " + sourceModule.getName() +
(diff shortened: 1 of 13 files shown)


api/src/org/labkey/api/module/FolderTypeManager.java
```diff
@@ -102,6 +102,10 @@ public void unregisterFolderType(String name)
 
     public void registerFolderType(Module sourceModule, FolderType folderType)
     {
+        if (!ModuleLoader.getInstance().isStartupInProgress() && !ModuleLoader.getInstance().isStartupComplete())
+        {
+            throw new IllegalStateException("Don't call registerFolderType() this early.  It will cause WebPart to get populated from ALL modules");
```
labkey-adam:
```suggestion
            throw new IllegalStateException("Don't call registerFolderType() this early; it can cause WebParts to get populated from modules that are about to be pruned");

```

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -634,10 +634,37 @@ private boolean isDevelopmentBuild(Module module)
...
+            if (module == core)
+                continue;
+            if (!module.getSupportedDatabasesSet().contains(coreType))
+            {
+                var e = new DatabaseNotSupportedException("This module does not support " + CoreSchema.getInstance().getSqlDialect().getProductName());
+                // In production mode, treat these exceptions as a module initialization error
+                // In dev mode, make them warnings so devs can easily switch databases
+                removeModule(iterator, module, !AppProps.getInstance().isDevMode(), e);
+            }
+        }
```
labkey-adam, line 664: Don't you want to remove this check from initialize() now?

labkey-adam (changes requested):

api/src/org/labkey/api/module/ModuleLoader.java
labkey-matthewb: Can't hurt, sure.

labkey-adam (approved):

PR #1144, opened 2020-05-06
Issue 40379: When merging data, we need to skip over columns that are backed by DbSequences.

labkey-susanh:
#### Rationale
Since we converted the RowId column for the exp.materials column to be backed by a DbSequence instead of an auto-increment column from the database, we need to be sure that we don't try to update those columns when doing a merge.  

#### Related Pull Requests
None

### DIFF
--- a/api/src/org/labkey/api/data/StatementUtils.java
+++ b/api/src/org/labkey/api/data/StatementUtils.java
@@ -726,6 +726,12 @@ public class StatementUtils
             if (done.contains(name))
                 continue;
             done.add(name);
+            if (Operation.merge == _operation)
+            {
+                ColumnInfo updatableColumn = updatable.getColumn(column.getName());
+                if (updatableColumn != null && updatableColumn.hasDbSequence())
+                    continue;
+            }
 
             SQLFragment valueSQL = new SQLFragment();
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/StatementUtils.java
```diff
@@ -726,6 +726,12 @@ public void setObjectUriPreselect(SQLFragment sqlfPreselectObject, TableInfo tab
             if (done.contains(name))
                 continue;
             done.add(name);
+            if (Operation.merge == _operation)
+            {
+                ColumnInfo updatableColumn = updatable.getColumn(column.getName());
+                if (updatableColumn != null && updatableColumn.hasDbSequence())
+                    continue;
```
labkey-matthewb: I think continue here will leave the column out of both the insert and the update part of the merge.  I think we just want _dontUpdateColumnNames.add(name) w/o the merge==_operation check.  Grain of salt alert: that's just from looking at the code w/o stepping through, however.

labkey-susanh: I wondered about that as I left my computer to think for itself overnight.  Revised.  Thanks.

labkey-matthewb (approved):

PR #1146, opened 2020-05-06
RlabkeyTest fixes related to labkey.domain.create changes with study datasets

cnathe:
#### Rationale
The RlabkeyTest has a few test cases that make use of the labkey.domain.create API helper from Rlabkey to test cases related to study dataset creation and issue list def creation. Recent server side API changes for both study dataset and issue list def look to have caused failures in this test. This PR fixes those issues (see changes mentioned below).

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/329

#### Changes
* Add DatasetDomainKind backwards compatibility support for domain create with categoryId or categoryName
* Refactor AbstractIssuesListDefDomainKind validation of the issue list def name for the create vs update case


internal/src/org/labkey/api/issues/AbstractIssuesListDefDomainKind.java
```diff
@@ -284,12 +285,9 @@ public boolean canCreateDefinition(User user, Container container)
...
                                                      @Nullable IssuesDomainKindProperties options, Container container, User user, boolean includeWarnings)
     {
-        if (options != null)
-        {
-            ValidationException exception = IssuesListDefService.get().validateIssuesProperties(options);
-            if (exception.hasErrors())
-                return exception;
-        }
+        if (options != null && StringUtils.isBlank(options.getIssueDefName()))
+            throw new IllegalArgumentException("Issue name must not be null.");
```
RosalineP: Since this function is marked as returning a ValidationException, should we keep the add-to-exception structure that existed previously rather than throwing?

RosalineP (approved):
Looks good! Only had that one comment.

internal/src/org/labkey/api/issues/AbstractIssuesListDefDomainKind.java
cnathe: Sure, I can put that back. Thanks.

PR #1149, opened 2020-05-06
Issue 38592: Rename webapp directory to have same casing as resources directory, but do it twice 'cause that's nice

labkey-susanh:
#### Rationale
Using a capital-E experiment directory for some of the content copied to the web directory can cause some of the content not to be found when building on Windows.  At least that's the theory.  In any case, there's no reason to use different casing for the directories.

And, it turns out that in order to do this renaming that just changes casing you have to do two commits, first one that renames but changes something other than casing and then one that renames to the case-changed name.  What fun!

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/165
* https://github.com/LabKey/testAutomation/pull/328

#### Changes
* Renamed `webapp/Experiment` to `webapp/experiment`


labkey-tchad (approved):
I think we need to do a non-squash merge for this to work but looks good to me otherwise.

cnathe (approved):

PR #1153, opened 2020-05-07
Item 7205: Dataset designer story4 - create dataset from a file

labkey-ankurj:
#### Rationale
This work provides the server side support for dataset creation by importing a file.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/243


core/package.json
```diff
@@ -90,8 +90,8 @@
     }
   },
   "dependencies": {
-    "@labkey/api": "0.2.8",
-    "@labkey/components": "0.56.3",
+    "@labkey/api": "0.2.4",
```
cnathe: We don't want to downgrade this @labkey/api (we need it here for the LABKEY var in LK server). Nick bump this to pick up a fix for some automated test failures. Instead, you could pull in 0.2.8 in your @labkey/components package.json to make them match.

cnathe (changes requested):
The code change in StudyController looks good, but one change to revert in the core/package.json

core/package.json
labkey-ankurj: no idea how this got changed, will change it back.

cnathe: It probably got reverted during a merge conflict resolution (i.e. merging develop into your fb).

labkey-ankurj: done

labkey-ankurj: I've it in the ui-components package.json, will merge that later.

cnathe (approved):

PR #1156, opened 2020-05-08
Bump up of some important caches that have been "limited" on production servers

labkey-adam:
#### Rationale
The limits we set on caches should be reviewed periodically as data volumes and standard RAM configurations change. We're not particularly scientific about this, but our "caches" admin page flags all caches that have been limited, providing a simple way to identify candidates whose limits might be increased. We don't blindly increase every one that's flagged; some caches are supposed to be restricted in size. This PR increases the default limit for DbCaches and doubles the limit for some widely used core caches that have been flagged on labkey.org or Atlas.

#### Changes
* Increase the DbCache default limit from 1,000 to 5,000
* Double the limit on key caches that have hit the limit production servers: Property maps, ObjectIds, Property descriptors, Domain properties, Groups, StringExpression template URLs

### DIFF
--- a/api/src/org/labkey/api/exp/OntologyManager.java
+++ b/api/src/org/labkey/api/exp/OntologyManager.java
@@ -97,9 +97,9 @@ import static org.labkey.api.search.SearchService.PROPERTY;
 public class OntologyManager
 {
     private static final Logger _log = Logger.getLogger(OntologyManager.class);
-    private static final Cache<String, Map<String, ObjectProperty>> mapCache = new DatabaseCache<>(getExpSchema().getScope(), 5000, "Property maps");
-    private static final Cache<String, Integer> objectIdCache = new DatabaseCache<>(getExpSchema().getScope(), 1000, "ObjectIds");
-    private static final Cache<Pair<String, GUID>, PropertyDescriptor> propDescCache = new BlockingCache<>(new DatabaseCache<>(getExpSchema().getScope(), 10000, CacheManager.UNLIMITED, "Property descriptors"), new CacheLoader<>()
+    private static final Cache<String, Map<String, ObjectProperty>> mapCache = new DatabaseCache<>(getExpSchema().getScope(), 10000, "Property maps");
+    private static final Cache<String, Integer> objectIdCache = new DatabaseCache<>(getExpSchema().getScope(), 2000, "ObjectIds");
+    private static final Cache<Pair<String, GUID>, PropertyDescriptor> propDescCache = new BlockingCache<>(new DatabaseCache<>(getExpSchema().getScope(), 40000, CacheManager.UNLIMITED, "Property descriptors"), new CacheLoader<>()
     {
         @Override
         public PropertyDescriptor load(@NotNull Pair<String, GUID> key, @Nullable Object argument)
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -97,9 +97,9 @@
 public class OntologyManager
 {
     private static final Logger _log = Logger.getLogger(OntologyManager.class);
-    private static final Cache<String, Map<String, ObjectProperty>> mapCache = new DatabaseCache<>(getExpSchema().getScope(), 5000, "Property maps");
-    private static final Cache<String, Integer> objectIdCache = new DatabaseCache<>(getExpSchema().getScope(), 1000, "ObjectIds");
-    private static final Cache<Pair<String, GUID>, PropertyDescriptor> propDescCache = new BlockingCache<>(new DatabaseCache<>(getExpSchema().getScope(), 10000, CacheManager.UNLIMITED, "Property descriptors"), new CacheLoader<>()
+    private static final Cache<String, Map<String, ObjectProperty>> mapCache = new DatabaseCache<>(getExpSchema().getScope(), 10000, "Property maps");
+    private static final Cache<String, Integer> objectIdCache = new DatabaseCache<>(getExpSchema().getScope(), 2000, "ObjectIds");
```
labkey-jeckels, line 101: If our default size is now 5000, should we just use that here too?

api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -174,7 +174,7 @@ private static DomainDescriptor fetchDomainDescriptorFromDB(String domainURI, Co
     }
 
     private static final BlockingCache<Integer, DomainDescriptor> domainDescByIDCache = new BlockingCache<>(new DatabaseCache<>(getExpSchema().getScope(),2000, CacheManager.UNLIMITED,"Domain descriptors by ID"), new DomainDescriptorLoader());
-    private static final BlockingCache<Pair<String, GUID>, List<Pair<String, Boolean>>> domainPropertiesCache = new BlockingCache<>(new DatabaseCache<>(getExpSchema().getScope(), 2000, CacheManager.UNLIMITED, "Domain properties"), new CacheLoader<>()
+    private static final BlockingCache<Pair<String, GUID>, List<Pair<String, Boolean>>> domainPropertiesCache = new BlockingCache<>(new DatabaseCache<>(getExpSchema().getScope(), 5000, CacheManager.UNLIMITED, "Domain properties"), new CacheLoader<>()
```
labkey-jeckels, line 177: Or perhaps make a DatabaseCache constructor that doesn't take a size and applies the default?

labkey-jeckels (approved):
A couple of optional ideas

api/src/org/labkey/api/exp/OntologyManager.java
labkey-adam: That default only applies to (legacy) DbCaches. DbCaches aren't explicitly constructed, so they must have a default size (which can be overridden in schema.xml). I prefer requiring that developers who explicitly construct caches think about the appropriate size for each cache.

api/src/org/labkey/api/exp/OntologyManager.java
labkey-adam: See above

api/src/org/labkey/api/exp/OntologyManager.java
labkey-jeckels: OK, though I imagine most sizes are basically wild guesses

labkey-adam: Yep, that's why I plan to review some production servers periodically. We might consider sending the list of "limited" caches back via mothership metrics to do a more comprehensive analysis.

PR #1160, opened 2020-05-09
Admin console defaults to "Settings"

labkey-nicka:
#### Rationale
The admin console is more often than not used to access server-wide settings. This PR changes the default to the "Settings" tab instead of "Server Information".

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/332

#### Changes
* Reorder admin console tabs and default to settings tab (`#links`).

### DIFF
--- a/api/src/org/labkey/api/view/menu/SiteAdminMenu.java
+++ b/api/src/org/labkey/api/view/menu/SiteAdminMenu.java
@@ -90,7 +90,6 @@ public class SiteAdminMenu extends NavTreeMenu
         AdminUrls adminUrls = PageFlowUtil.urlProvider(AdminUrls.class);
 
         ActionURL consoleUrl = adminUrls.getAdminConsoleURL();
-        consoleUrl.setFragment("info");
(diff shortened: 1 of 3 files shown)


labkey-jeckels (approved):
This change looks good, but I'm guessing some tests will need updates

labkey-nicka:
@labkey-jeckels I took a look at the tests and they seem to use the admin console test page pretty consistently. There may be some tests that are not using it that should be switched over. I've kicked off runs for (list to remind myself):

- ✅ Daily A Postgres - Failure in Flow tests due to an unrelated reason (NPE fixed with #1168).
- ✅ Daily B SqlServer - `MothershipReportTest` failure. Fixed with 2f3d981.
- ✅ CustomModules Postgres - failures consistent with develop
- ✅ DRT Postgres - All passing
- ✅ External Modules Postgres
- ✅ Daily C SqlServer - AssayTest, SampleSetTest failed in unrelated location -- neither reproducible locally. Other failures consistent with develop.

PR #1161, opened 2020-05-10
Issue 40403: Timeline: audit log and user permission issues

XingY:
#### Rationale
It's observed that timeline page won't load at all if user lack CanSeeAuditLogPermission. While Timeline is backed by audit log, we shouldn't require user to have CanSeeAuditLogPermission to view it in LKSM. 

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/265

#### Changes
* Allow SampleTimelineAuditEvent and ExperimentAuditEvent to be able to bypass CanSeeAuditLogPermission check


labkey-susanh (approved):

labkey-klum (commented):
I'm not a huge fan of adding a switch into the audit log service to bypass the schema security settings. Typically we would do something like add it to the contextual role in the controller actions that needed to read a specific audit table. Is there a reason that this type of an approach wouldn't have worked here or was undesirable?

XingY:
@labkey-klum That's great to know about contextual role. I'll close this PR and use it in the controller action instead.

PR #1165, opened 2020-05-12
Use $.each() instead of Array.forEach()

labkey-nicka:
#### Rationale
With #1134 I had switched several usages of `$.each()` to use `Array.forEach()` without realizing that `forEach()` does not support short-circuiting via `return false`. This PR reverts those usages that do make use of that mechanism so the logic works as expected.

This fixes the failing `LuminexEC50Test.testEC50` test case which I will confirm with a run of the Daily A test suite.

#### Related Pull Requests
* #1134

#### Changes
* Use `$.each()` instead of `Array.forEach()` in `DataRegion` for cases where short-circuit mechanism is utilized.


cnathe (approved):

labkey-nicka:
Passing `LuminexEC50Test`:

image

PR #1170, opened 2020-05-13
Epic 6800 : Issue Definition designer (Story 3)- Issues module implementation

labkey-bpatel:
#### Rationale
This PR includes setup and new development to convert Issues Definition designer to a react-based designer, available via enabling Experimental feature flag.

#### Related Pull Requests
https://github.com/LabKey/labkey-ui-components/pull/245

#### Changes
- Includes setup of Issues module for react page development
- Includes entrypoint for Issue’s Designer usage
- Includes URL param parsing, loading the existing issue model for edit and navigation on success
- Experimental feature flag
- New issues api actions to getProjectGroups and getUsersForGroup


cnathe:
@labkey-bpatel Don't forget about items `1.iv` and `4` in the "Setting up a LabKey module doc here: https://github.com/LabKey/platform/blob/develop/webpack/README.md

issues/src/org/labkey/issue/IssuesController.java
```diff
@@ -1968,21 +1977,32 @@ public ModelAndView getView(IssuesController.AdminForm adminForm, BindException
...
-                return new HtmlView(getUndefinedIssueListMessage(getViewContext(), issueDefName));
+                return ModuleHtmlView.get(ModuleLoader.getInstance().getModule("issues"), "designer");
             }
-            Domain domain = issueListDef.getDomain(getUser());
+            else
+            {
+                if (issueListDef == null)
+                {
+                    return new HtmlView(getUndefinedIssueListMessage(getViewContext(), issueDefName));
+                }
```
cnathe: This check for `issueListDef == null` and returning the HtmlView should likely stay out of this if/else block. It will still apply for the new designer ui as it is a way to show an error message if the URL param is invalid.

issues/src/org/labkey/issue/IssuesController.java
```diff
@@ -1997,6 +2017,205 @@ public NavTree appendNavTrail(NavTree root)
...
+    public class GetProjectGroupsAction extends ReadOnlyApiAction<Object>
+    {
+        @Override
+        public Object execute(Object form, BindException errors)
+        {
+            //derived from IssueServiceAction.getProjectGroups()
+            List<Principal> groups = new ArrayList<>();
+
+            SecurityManager.getGroups(getContainer().getProject(), true).stream().filter(group -> !group.isGuests() && (!group.isUsers() || getUser().hasRootAdminPermission())).forEach(group -> {
+                String displayText = (group.isProjectGroup() ? "" : "Site:") + group.getName();
```
cnathe: I think you previously had a space added after the `Site:` string. It would be nice to put that back in.

issues/src/org/labkey/issue/IssuesController.java
```diff
@@ -1997,6 +2017,205 @@ public NavTree appendNavTrail(NavTree root)
...
+
+                Principal principal = new Principal();
+                principal.setActive(group.isActive());
+                principal.setDisplayName(displayText);
+                principal.setName(group.getName());
+                principal.setType(group.getType());
+                principal.setUserId(group.getUserId());
```
cnathe: Looking at the client side code, I think we only end up using the group id and the displayName. Do you think we need to return the rest of this information?

cnathe: Alternatively, could we just build up a `List<Group>`'s here and then return that from the API? then I think we could do the displayText part on the client side since the Group object would include the `isProjectGroup` info. That way we could do away with this extra `public static class Principal` class.

issues/src/org/labkey/issue/IssuesController.java
```diff
@@ -1997,6 +2017,205 @@ public NavTree appendNavTrail(NavTree root)
...
+        }
+
+        public void setGroupId(Integer groupId)
+        {
+            this.groupId = groupId;
+        }
+
+    }
+
+    public static class Principal
```
cnathe: these class names seem backwards, isn't this one for a "Group"?

issues/src/org/labkey/issue/IssuesController.java
```diff
@@ -1997,6 +2017,205 @@ public NavTree appendNavTrail(NavTree root)
...
+        {
+            return active;
+        }
+
+        public void setActive(boolean active)
+        {
+            this.active = active;
+        }
+    }
+    public static class UserGroup
```
cnathe: these class names seem backwards, isn't this one for a "User/Principal"?

issues/src/org/labkey/issue/IssuesController.java
```diff
@@ -1997,6 +2017,205 @@ public NavTree appendNavTrail(NavTree root)
...
+                {
+                    for (User user : SecurityManager.getAllGroupMembers(group, MemberType.ACTIVE_USERS, group.isUsers()))
+                    {
+                        if (getContainer().hasPermission(user, UpdatePermission.class))
+                        {
+                            UserGroup usergrp = new UserGroup();
+                            usergrp.setGroupId(group.getUserId());
+                            usergrp.setGroupName(group.getName());
+                            usergrp.setUserId(user.getUserId());
+                            usergrp.setUserName(user.getDisplayName(getUser()));
```
cnathe: Similar question as the other response object. Does the client side code use/need all of this? I assume it only cares about the id and displayName.

issues/src/client/Designer/Designer.tsx
```diff
@@ -0,0 +1,149 @@
...
+    fetchIssuesListDefDesign
+} from "@labkey/components";
+
+import "@labkey/components/dist/components.css"
+
+type State = {
+    isLoadingModel: boolean,
+    message?: string,
+    model?: IssuesListDefModel,
+    hasDesignIssuesAdminPermission?: boolean
```
cnathe: I don't think we need this as I don't think there is a specific permissions class for issue list def admins. I believe it is just a container admin check which we already have on the IssuesController.AdminAction.

issues/src/client/Designer/Designer.tsx
```diff
@@ -0,0 +1,149 @@
...
+                    hasDesignIssuesAdminPermission: data.container.effectivePermissions.indexOf(PermissionTypes.Admin) > -1
+                }));
+            },
+            failure: (error) => {
+                this.setState(() => ({
+                    message: error.exception,
+                    hasDesignIssuesAdminPermission: false
+                }));
+            }
+        });
```
cnathe: I think this can all be removed as the action requires the permissions and the entryPoint requires admin as well.

issues/src/client/Designer/Designer.tsx
```diff
@@ -0,0 +1,149 @@
...
+            return <Alert>{message}</Alert>
+        }
+
+        if (isLoadingModel || hasDesignIssuesAdminPermission === undefined) {
+            return <LoadingSpinner/>
+        }
+
+        if (!hasDesignIssuesAdminPermission) {
+            return <Alert>You do not have sufficient permissions to create or edit a issues list definition design.</Alert>
+        }
```
cnathe: I think this can be removed as well.

cnathe (changes requested):

issues/src/org/labkey/issue/IssuesController.java
labkey-bpatel: ah youre right, now fixed

issues/src/org/labkey/issue/IssuesController.java
labkey-bpatel: yes! done

issues/src/org/labkey/issue/IssuesController.java
labkey-bpatel: Nope, not needed, I've simplified it now, let me know what you think!

issues/src/org/labkey/issue/IssuesController.java
labkey-bpatel: This class was same as client side Principle model. But its really an overkill, so now removed.

issues/src/org/labkey/issue/IssuesController.java
labkey-bpatel: This class was the same as client side's UserGroup - but now removed.

issues/src/org/labkey/issue/IssuesController.java
labkey-bpatel: nope not needed, please review my updates.

issues/src/client/Designer/Designer.tsx
labkey-bpatel: ah, good point, now removed

issues/src/client/Designer/Designer.tsx
labkey-bpatel: done

issues/src/client/Designer/Designer.tsx
labkey-bpatel: done

issues/src/org/labkey/issue/IssuesController.java
cnathe: Looks good

issues/src/org/labkey/issue/IssuesController.java
cnathe: Looks good

cnathe (changes requested):
Code changes look good. I think the only thing left are those two items missing from the "Setting up a LabKey module" doc for the README file and .gitignore changes: https://github.com/LabKey/platform/blob/develop/webpack/README.md

cnathe (approved):

PR #1174, opened 2020-05-13
Issue 40443: If changing tables in a deferred upgrade script, you als…

labkey-susanh:
#### Rationale
Previous changes updated the exp.Materials table in deferred upgrade scripts.  On SQL Server, changing tables can cause the views that reference these tables to be messed up, so we need to recreated the view after the deferred upgrade scripts have happened.

#### Related Pull Requests

#### Changes
* Issue 40443: If changing tables in a deferred upgrade script, you also need to recreate the views in SQLServer


labkey-jeckels (approved):
Could add the issue ID to the comment in the Java code, but otherwise looks great

PR #1177, opened 2020-05-14
Don't insist modules' sourceDir be non-null

labkey-tchad:
#### Rationale
When running on TeamCity with gradle plugin v1.12.0+, we `null` out the 'sourcePath' property of modules. Some places expect this to be non-null. BasicTest is failing when it attempts to visit `/labkey/admin-sql-orphanedScripts.view?`

#### Changes
* Add `null` checks


labkey-matthewb (approved):

labkey-tchad:
Fixed elsewhere

PR #1180, opened 2020-05-15
Ticket 40464: Consistency on Delete Permissions for Workbooks

bbimber:
@labkey-jeckels like we discussed, last year I made a change that made workbooks use the standard container delete pathway.  the problem is that workbooks should require DeletePermission, not AdminPermission.  These changes enforce the right permission based on what's being deleted.  Based on your suggestion, I also changed ContainerType based on your suggestion on the ticket.

Note: I dont know if team city will run tests off this.  i think it will require a branch on the labkey side of the work, but I dont think I can do this.

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -82,7 +82,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.BooleanSupplier;
-import java.util.stream.Collectors;
 
 
 /**
@@ -201,9 +200,9 @@ public class Container implements Serializable, Comparable<Container>, Securable
         return targetContainer != null && targetContainer.getContainerType().allowRowMutationFromContainer(this, targetContainer);
     }
 
-    public boolean requiresAdminToDelete()
+    public Class<? extends Permission> getPermissionNeededToDelete()
     {
-        return _containerType.requiresAdminToDelete();
+        return _containerType.getPermissionNeededToDelete();
     }
 
     public boolean isDuplicatedInContainerFilter()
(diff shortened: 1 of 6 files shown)


labkey-jeckels:
This change looks good to me.

@labkey-tchad can you assist with pointers on the workflow to get this tested on TeamCity? Ideally, if you could capture the sequence in a doc page (or share an existing link) I can follow them. I looked but couldn't find an existing set of guidelines.

labkey-tchad (changes requested):
A test in `AdminController` needs to be updated:
```
testActionPermissions(org.labkey.core.admin.AdminController$TestCase)
  Should not have permission: user (editor@actionpermission.test), action (org.labkey.core.admin.AdminController$DeleteFolderAction)
  java.lang.AssertionError: Should not have permission: user (editor@actionpermission.test), action (org.labkey.core.admin.AdminController$DeleteFolderAction)
	at org.junit.Assert.fail(Assert.java:88)
	at org.labkey.api.security.permissions.AbstractActionPermissionTest.assertNoPermission(AbstractActionPermissionTest.java:374)
	at org.labkey.api.security.permissions.AbstractActionPermissionTest.assertForAdminPermission(AbstractActionPermissionTest.java:245)
	at org.labkey.api.security.permissions.AbstractActionPermissionTest.assertForAdminPermission(AbstractActionPermissionTest.java:225)
	at org.labkey.core.admin.AdminController$TestCase.testActionPermissions(AdminController.java:10276)
```

bbimber:
@labkey-tchad thanks trey.  i just added a commit here to fix that.  do i need to explicitly merge my change from bbimber/platform:fb_workbooks into labkey/platform:fb_workbooks?

labkey-tchad:
I just updated our fb to match yours. We can't do the update via PR because the HEAD on our side needs to be on the same commit for the status checks to show up here.

labkey-tchad (approved):
Looks good now. The remaining failure has been fixed in develop.

PR #1182, opened 2020-05-16
Vendor bundles, build-dev target

labkey-nicka:
### Rationale
This PR introduces a couple of incremental improvements to our building of our client-side apps. Namely, vendor bundling, new `build-dev` target, and library (`*.lib.xml`) generation. With these changes the client-side build for a module's apps take less time as the number of apps increase and the common code is cacheable (by the browser) across apps for users.

**Vendor bundling**
Webpack allows for assets to be bundled across entry points as an optimization for application assets. These bundles can be composed in a variety of ways and a common way is to bundle all of the packages that an app relies on together; apart from the application itself. This is done via the [SplitChunksPlugin](https://webpack.js.org/plugins/split-chunks-plugin/). With this, a set of entry points can then share the bundles as they represent "common code" across them.

Here is an example of the difference in outputs:

Before - multiple applications:
image

After - multiple applications:
image

These vendor outputs are shared by all the apps in a module and you can see how small the application are stand-alone. With this, the processing for `.view.xml`s had to be updated to account for these dynamic assets. For the `dataClassDesigner` app the associated `dataClassDesigner.view.xml` now looks like this:

```xml
<view xmlns="http://labkey.org/data/xml/view" frame="none" title="Data Class Designer">
    <permissions>
        <permission name="admin"/>
    </permissions>
    <dependencies>
        <dependency path="experiment/gen/vendors~253ae210.css"/>
        <dependency path="experiment/gen/vendors~253ae210.7e75c5d563b2ab5c75d6.js"/>
        <dependency path="experiment/gen/vendors~af9f3849.158d62780c51a4c07d31.js"/>
        <dependency path="experiment/gen/dataClassDesigner~875236fa.8f4e183dee678fffe5d6.js"/>
    </dependencies>
</view>
```

Lastly, I'll note that we've been using this same pattern in Biologics for a number of years -- this update just brings it over to the general module app builds.

**Development build**
Webpack allows for different [modes](https://webpack.js.org/configuration/mode/) which define a default configuration for `development`, `production` and `none`. We use the `production` configuration for our builds since we want our local builds to reflect exactly the code we're running in production.

The downside of building the "production" code all the time is that it takes extra time to perform all the optimizations (minifying, source maps, just see this list of optimizations in modes...) which can start to significantly impact our build when we are building 10s of apps. At LabKey, for other resources, we tend to do these kind of optimizations only in our "production" build so our development experience isn't as impacted.

This change introduces a new build target, `build-dev`, which targets a new configuration specified in `platform/webpack/dev.config.js` that utilizes `mode: development`. This build takes noticeably less time (about 50-60% of the time a production build in my local runs), and, *should be* logically equivalent to the production build for the output code. I'm introducing `build-dev` with this PR so that, hopefully, we can eventually switch our client-side build to respect our Gradle setting for development v. production build.

**Library generation**
With the addition of dynamic assets being produced by Webpack (see vendor bundling) we needed a way for non-view specified applications (e.g. experiment/"RunGraph" which is served from a `.jsp`). I've replicated the logic for asset processing to produce a `.lib.xml` for an entry point when `generateLib` is set to `true`. This `generateLib` configuration flag replaces the `generateViews` flag I'd previously introduced.

**Build targets**
Brief summary of what the build targets do and how they are intended to be used:
- `build-dev`: A "development" build of the apps in a module. This build uses `mode: 'development'` and `devTool: 'eval'` source-map configuration and does not perform optimizations (e.g. vendor bundling). This can be invoked manually (`npm run build-dev`) but is not invokable via a Gradle build (for now 😛).
- `build-prod`: A "production" build of the apps in a module. This build uses `mode: 'production'` and `devTool: 'source-map'` source-map configuration and performs optimizations (e.g. vendor bundling). This is synonymous with calling `npm run build` and is invoked by our Gradle build.
- `start`: A "watch" build of the apps in a module which spins up a `webpack-dev-server` allowing for rapid app development via hot module reloading. This behavior is unchanged with this PR.

### Changes
* Introduce vendor bundling optimization across apps defined by a LK module. These vendor bundles are shared across these apps resulting in shared build processing (pays off more for each additional app declared by a module) and shared caching by client browsers.
* Adds a new `build-dev` target to all platform/<module>'s that currently have a `package.json`. This target invokes `platform/webpack/dev.config.js`.
* Renamed what was formerly `dev.config.js` to `watch.config.js` to align with our targets (`dev`, `prod`, `watch`). **Note to reviewers:** The original `dev.config.js` implementation is unchanged in `watch.config.js`.
* Factor out `processEntries` and `processPlugins` so logic can be shared between `dev` and `prod` build configurations.
* Entry point configuration flag `generateViews` replaced with `generateLib` which generates a `<app_name>.lib.xml` as an alternative to generating application views (i.e. `app.view.xml` and `appDev.view.xml`).


core/package.json
```diff
@@ -5,6 +5,8 @@
     "setup": "npm ci",
     "prebuild": "bnr clean",
     "build": "bnr build-themes:dev && bnr build-pages:prod",
+    "prebuild-dev": "bnr clean",
```
labkey-susanh: It's unfortunate that one `prebuild` script doesn't apply to all of these, but I can see that one wouldn't always want that.  In our cases, the `prebuilds` aren't all that helpful and aren't used elsewhere.  Just an observation, though. No need to change anything.

webpack/dev.config.js
```diff
@@ -4,78 +4,35 @@
...
 module.exports = {
     context: constants.context(__dirname),
 
     mode: 'development',
 
-    devServer: devServer,
+    devtool: 'eval',
 
-    entry: entries,
+    entry: constants.processEntries(entryPoints),
```
labkey-susanh, line 20: A nit: `constants` doesn't seem like the right name for a file that contains functions.  `utils` perhaps?

labkey-susanh (approved):
Nice!  I didn't test all of the packages here, but using build-dev in assay and experiment locally took at most half the time as using build-prod.  Are there any docs that need to be updated?

core/package.json
```diff
@@ -42,6 +44,13 @@
         "NODE_ENV": "production"
       }
     },
+    "build-pages:dev": {
+      "command": "webpack --config ../webpack/dev.config.js --progress --profile --colors",
+      "env": {
+        "LK_MODULE": "list",
```
cnathe: Shouldn't this be "core"?

cnathe (changes requested):
This is awesome! It will be very nice to have this build perf improvement in place.

One minor change requested. Also, there was one more platform package.json added last week for the issues module, would you mind picking that change up and adding this set of changes to issues/package.json as well?

core/package.json
labkey-nicka: Yup, fixed with 022c8c9.

labkey-nicka:
> This is awesome! It will be very nice to have this build perf improvement in place.
> 
> One minor change requested. Also, there was one more platform package.json added last week for the issues module, would you mind picking that change up and adding this set of changes to issues/package.json as well?

@cnathe Added to issues with c231b59.

cnathe (approved):

webpack/dev.config.js
labkey-nicka: I agree. The file has grown into a set of constants + utilities. For now, I'm just going to leave it as-is.

core/package.json
labkey-nicka: Removed with 6e60e707f4.

PR #1183, opened 2020-05-18
Misc performance optimizations

labkey-matthewb:
#### Rationale
getQueryDetails.api performance enhancements

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1183
https://github.com/LabKey/premium/pull/190
https://github.com/LabKey/sampleManagement/pull/269

#### Changes
Put debug-only code under asserts
Lsid.toString() is expensive,  don't call in Lsid() constructor
Pattern.match() is expensive don't use in Lsid() constructor
PropertyService.getDomains(Set<String> domainKinds) is expensive, create new method getDomains(DomainKind dk)
Avoid creating new DefaultSchema (caching constructed TableInfo objects relies on this)
Avoid any expensive SQL generation work until we need to do it (ExprColumn, QuerySelect)
AssayWellExclusion service, use provider instead of protocol to avoid expensive protocol->provider lookup
Added EOF token to SQL grammar (forum recommendation)


api/src/org/labkey/api/exp/Lsid.java
```diff
@@ -345,16 +359,16 @@ public LsidBuilder edit()
...
+            if (null == parts)
             {
                 valid = false;
                 return;
             }
-            authority = decodePart(m.group(1).toLowerCase());
-            namespace = decodePart(m.group(2));
-            objectId = decodePart(m.group(3));
-            version = decodePart(m.group(4));
+            authority = decodePart(parts[0]);
```
labkey-susanh: We did a toLowerCase before.  Not needed?

api/src/org/labkey/api/exp/property/PropertyService.java
```diff
@@ -66,9 +66,11 @@ static void setInstance(PropertyService impl)
     /** Get all the domains in the specified container and optionally project and shared. */
     List<? extends Domain> getDomains(Container container, User user, boolean includeProjectAndShared);
 
-    /** Get all the domains in the specified container and specified Domain Kinds. */
+    /** Get all the domains in the specified container and specified Domain Kinds. THIS IS SLOW */
```
labkey-susanh: Is the new method that uses a single DomainKind to be preferred here?  If so, perhaps worth a comment to that effect.

api/src/org/labkey/api/query/FilteredTable.java
```diff
@@ -61,6 +62,8 @@
  */
 public class FilteredTable<SchemaType extends UserSchema> extends AbstractContainerFilterable implements ContainerFilterable
 {
+    final private static ButtonBarConfig BUTTONBAR_NOTSET = new ButtonBarConfig(new JSONObject());
```
labkey-susanh:
```suggestion
    final private static ButtonBarConfig BUTTON_BAR_NOT_SET = new ButtonBarConfig(new JSONObject());

```

api/src/org/labkey/api/query/FilteredTable.java
```diff
@@ -714,4 +716,18 @@ public boolean hasPermission(@NotNull UserPrincipal user, @NotNull Class<? exten
         }
         return super.hasPermission(user, perm);
     }
+
+    @Override
+    public ButtonBarConfig getButtonBarConfig()
+    {
+        if (_buttonBarConfig == BUTTONBAR_NOTSET)
```
labkey-susanh: Comparison using this empty object is probably slightly less efficient than using an extra boolean `_isButtonBarSet` field but probably not enough to worry with.

assay/api-src/org/labkey/api/assay/AssayResultTable.java
```diff
@@ -223,8 +223,8 @@ else if (baseColumn.isMvIndicatorColumn())
         AssayWellExclusionService svc = AssayWellExclusionService.getProvider(_protocol);
         if (svc != null)
         {
-            addColumn(svc.createExcludedColumn(this, _protocol));
-            addColumn(svc.createExclusionCommentColumn(this, _protocol));
+            addColumn(svc.createExcludedColumn(this, getUserSchema().getProvider()));
```
labkey-susanh, line 226: Is this change also wanted on line 223 or only for getting the exclusion columns that will be in the user schema?  Is this going to be a change in behavior?  Should we have tests to cover this?

assay/src/org/labkey/assay/TSVProtocolSchema.java
```diff
@@ -122,9 +122,9 @@ private TableInfo createExclusionReportTable(ContainerFilter cf)
             AssayWellExclusionService svc = AssayWellExclusionService.getProvider(getProtocol());
             if (svc != null)
             {
-                var excludedByColumn = svc.createExcludedByColumn(this, getProtocol());
-                var excludedAtColumn = svc.createExcludedAtColumn(this, getProtocol());
-                var excludedCommentColumn = svc.createExclusionCommentColumn(this, getProtocol());
+                var excludedByColumn = svc.createExcludedByColumn(this, getProvider());
+                var excludedAtColumn = svc.createExcludedAtColumn(this, getProvider());
+                var excludedCommentColumn = svc.createExclusionCommentColumn(this, getProvider());
```
labkey-susanh, line 127: Same question here.  Are we correcting behavior here so we might need test coverage?

experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
```diff
@@ -204,6 +204,17 @@ public void registerDomainKind(DomainKind type)
         return getDomainsStream(container, user, domainKinds, includeProjectAndShared).collect(Collectors.toList());
     }
 
+    @Override
+    public List<? extends Domain> getDomains(Container container, User user, @NotNull DomainKind<?> dk, boolean includeProjectAndShared)
+    {
+        // Domain.getDomainKind() can be slow, instead just ask the passed in dk the domain matches
```
labkey-susanh:
```suggestion
        // Domain.getDomainKind() can be slow. Instead just ask the passed-in dk if the domain matches

```

query/src/org/labkey/query/sql/QField.java
```diff
@@ -170,7 +171,15 @@ public int hashCode()
...
     public BaseColumnInfo createColumnInfo(SQLTableInfo table, String alias, Query query)
     {
-        ExprColumn ret = new ExprColumn(table, alias, getRelationColumn().getValueSql(), getRelationColumn().getJdbcType());
+        final QueryRelation.RelationColumn rcol = getRelationColumn();
+        ExprColumn ret = new ExprColumn(table, alias, null, getRelationColumn().getJdbcType())
+        {
+            @Override
+            public SQLFragment getValueSql(String tableAlias)
+            {
+                return rcol.getValueSql();  // delay call in case this causes unnecessary
```
labkey-susanh: unnecessary what?  Angst?

query/src/org/labkey/query/sql/QuerySelect.java
```diff
@@ -2146,7 +2146,7 @@ void copyColumnAttributesTo(BaseColumnInfo to)
             {
                 if (_colinfo == null)
                 {
-                    _colinfo = expr.createColumnInfo(_sti, _aliasManager.decideAlias(getAlias()), _query);
+                    _colinfo = expr.createColumnInfo(_sti, "/*NOT AN OUTPUT COLUMN*/", _query);
```
labkey-susanh, line 2150: I don't pretend to know what's going on here, but this seems quite different than the previous version.  Is that intentional?  This second argument is supposed to be a name, but that seems to not be much like a decent column name.

query/src/org/labkey/query/sql/SqlParser.java
```diff
@@ -1906,13 +1991,15 @@ public void testSql()
...
-					bad(sql);
+                    bad(sql);
                 }
                 catch (Throwable t)
                 {
                     fail(sql);
                 }
             }
+            long end = System.currentTimeMillis();
+//            System.out.println("SqlParser.testSql(): " + DateUtil.formatDuration(end-start));
```
labkey-susanh: Remove commented out println?

labkey-susanh (commented):
Just a few suggestions and some comments on what seem like changes in behavior.

assay/api-src/org/labkey/api/assay/AssayResultTable.java
labkey-susanh: Ah, I see.  Signature changed.

assay/src/org/labkey/assay/TSVProtocolSchema.java
labkey-susanh: Never mind.  Signature changed.

labkey-susanh (commented):
Looks like we've got a memory leak in some SM tests.

api/src/org/labkey/api/exp/Lsid.java
labkey-matthewb: unintentional difference.  Kinda weird that toLowerCase() is inside decodePart()

api/src/org/labkey/api/exp/property/PropertyService.java
labkey-matthewb: added comment

api/src/org/labkey/api/query/FilteredTable.java
labkey-matthewb: OK

experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
labkey-matthewb: OK

assay/src/org/labkey/assay/TSVProtocolSchema.java
labkey-matthewb: next time

query/src/org/labkey/query/sql/QField.java
labkey-matthewb: check

query/src/org/labkey/query/sql/QuerySelect.java
labkey-matthewb: I'll add a comment

query/src/org/labkey/query/sql/SqlParser.java
labkey-matthewb: _log.trace()

labkey-matthewb:
> Looks like we've got a memory leak in some SM tests.

Got it already.

labkey-susanh (approved):
Changes here look good as long as the memory leak in SM goes away.

PR #1191, opened 2020-05-19
Issue 40451: Should not return html in Exception.getMessage()

labkey-jeckels:
Simplify error message when GraphViz fails to run, especially since it's on the way out


kevinkrouse (approved):

labkey-kevink (approved):
Double approved

PR #1196, opened 2020-05-20
7109: Allow QC states to be set and updated via Python API

RosalineP:
#### Rationale
To allow for QC State definitions and QC State assignments to be created, removed, and updated via Python API. 

#### Changes
* QCStateTableInfo is now writable via implementing a QueryUpdateService, rendering to table auto-accessible for API language bindings
* Gates against deletion of in-use QC state definition through API


api/src/org/labkey/api/qc/QCStateManager.java
```diff
@@ -37,6 +37,7 @@
 public class QCStateManager
 {
     private static final QCStateManager _instance = new QCStateManager();
+    private Map<String, QCStateHandler> _QCStateHandlers = new HashMap<>();
```
labkey-klum: this can probably be a static variable since QCStateManager is essentially a singleton class

core/src/org/labkey/core/CoreController.java
```diff
@@ -92,6 +92,7 @@
 import org.labkey.api.qc.AbstractManageQCStatesForm;
 import org.labkey.api.qc.DeleteQCStateForm;
 import org.labkey.api.qc.QCStateHandler;
+import org.labkey.api.qc.QCStateManager;
```
labkey-klum: unneeded import?

core/src/org/labkey/core/query/QCStateTableInfo.java
```diff
@@ -37,4 +55,51 @@ public QCStateTableInfo(CoreQuerySchema schema)
...
+                if (handler.isQCStateInUse(container, QCToDelete))
+                    return false;
+            }
+            return true;
+        }
+
+        @Override
+        protected Map<String, Object> insertRow(User user, Container container, Map<String, Object> row) throws DuplicateKeyException, ValidationException, QueryUpdateServiceException, SQLException
+        {
+            QCStateManager.getInstance().clearCache(container);
```
labkey-klum: We may want to clear the cache only after the insert successfully completes, so you'll need to save the return value, clear the cache and return the saved value. Otherwise if the insert operation were to throw then we would have prematurely cleared the cache (which isn't terrible, but unnecessary).

core/src/org/labkey/core/query/QCStateTableInfo.java
```diff
@@ -37,4 +55,51 @@ public QCStateTableInfo(CoreQuerySchema schema)
...
+            return super.insertRow(user, container, row);
+        }
+
+        @Override
+        protected Map<String, Object> deleteRow(User user, Container container, Map<String, Object> oldRowMap) throws InvalidKeyException, QueryUpdateServiceException, SQLException
+        {
+            if (!validateQCStateNotInUse(oldRowMap, container))
+                throw new QueryUpdateServiceException("QC state '" + oldRowMap.get("label") + "' cannot be deleted as it is currently in use.");
+
+            QCStateManager.getInstance().clearCache(container);
```
labkey-klum: similar to insertRow, clear the cache only after we have successfully deleted the row.

core/src/org/labkey/core/query/QCStateTableInfo.java
```diff
@@ -37,4 +55,51 @@ public QCStateTableInfo(CoreQuerySchema schema)
...
-}
\ No newline at end of file
+
+    @Override
+    public boolean hasPermission(@NotNull UserPrincipal user, @NotNull Class<? extends Permission> perm)
+    {
+        return getContainer().hasPermission(user, perm);
+    }
+
+    private class QCStateService extends DefaultQueryUpdateService
```
labkey-klum: Does this need to be an inner class? It doesn't look like if makes references to the enclosing class therefore it may be better to make this static.

[explanation of differences](https://docs.oracle.com/javase/tutorial/java/javaOO/nested.html)

labkey-klum (approved):
overall looks good but please address comments.

core/src/org/labkey/core/query/QCStateTableInfo.java
labkey-jeckels: Please use a CommitTask as well for these cache clearings. See an example in ExperimentServiceImpl.deleteDataClass().

The scenario of concern is:

1. API request comes in to update/delete/insert QC requests.
2. Transaction starts.
3. Process a row, update the DB, clear the cache.
4. Commit to DB.

Between 3 and 4, another thread can reload the cache based on the original data since the changes aren't yet committed, and cache is stale until it times out or another edit happens.

core/src/org/labkey/core/query/QCStateTableInfo.java
RosalineP: Ah, I see. Changed to a static nested class rather than a inner class

PR #1197, opened 2020-05-20
Auto retry after database deadlock

labkey-matthewb:
#### Rationale
Although best avoided deadlocks happen, and the correct thing to do is retry.  Here is a simple retry strategy to automatically retry GET requests that deadlock (probably not as common as POST, but it can happen).

### DIFF
--- a/api/src/org/labkey/api/action/SpringActionController.java
+++ b/api/src/org/labkey/api/action/SpringActionController.java
@@ -24,6 +24,7 @@ import org.json.JSONObject;
 import org.labkey.api.admin.AdminUrls;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
 import org.labkey.api.data.Container;
+import org.labkey.api.data.dialect.SqlDialect;
 import org.labkey.api.miniprofiler.MiniProfiler;
 import org.labkey.api.miniprofiler.RequestInfo;
 import org.labkey.api.module.AllowedBeforeInitialUserIsSet;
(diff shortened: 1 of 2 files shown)


labkey-matthewb:
I had this floating around, what do you guys think?

api/src/org/labkey/api/action/SpringActionController.java
```diff
@@ -525,6 +526,18 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
 
     protected void handleException(HttpServletRequest request, HttpServletResponse response, Throwable x)
     {
+        // OK, if we get here with a deadlock exception AND this is a get AND we haven't committed the response yet,^M
+        // ASK the caller to retry^M
```
labkey-kevink:
```suggestion
        // OK, if we get here with a deadlock exception AND this is a get AND we haven't committed the response yet,
        // ASK the caller to retry

```

api/src/org/labkey/api/action/SpringActionController.java
```diff
@@ -525,6 +526,18 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
 
     protected void handleException(HttpServletRequest request, HttpServletResponse response, Throwable x)
     {
+        // OK, if we get here with a deadlock exception AND this is a get AND we haven't committed the response yet,^M
+        // ASK the caller to retry^M
+        if (x instanceof Exception && SqlDialect.isTransactionException((Exception)x) && "GET".equals(request.getMethod()) && !response.isCommitted())
+        {
+            if (!request.getQueryString().contains("_retry_=1"))
```
labkey-kevink: Since we are in the SpringActionController, can't we get the ViewContext and the ActionURL instead of using the request query string?

labkey-kevink (changes requested):

labkey-kevink (approved):

PR #1199, opened 2020-05-20
Migrate NavTree appendNavTrail() to void addNavTrail()

labkey-adam:
#### Rationale
The fact that `NavTrailAction.appendNavTrail()` returns a `NavTree` has been the source of confusion and bugs in the past. Some actions have attempted to replace the passed in nav trail by returning null or a brand new NavTree, but this is futile; the primary caller, `SpringActionController.appendNavTrail()`, completely ignores the return value. Having a return value encourages chaining of the methods that append to nav trails, but chaining can have unintended consequences (e.g., `NavTree.addChild()` returns the new child, not the root... but any adds to the child will be ignored by nav trail rendering). Requiring a return value also adds unnecessary boilerplate to every action.

This PR initiates our migration from `NavTree appendNavTrail()` to `void addNavTrail()` to eliminate this return value. In Java, we can't overload based on return type, so we must change the standard method's signature. For a transition period (until after 20.7.0 ships), actions will be able to implement either `appendNavTrail()` or `addNavTrail()`. After that, `appendNavTrail()` will be removed and implementing `addNavTrail()` will be required.

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/171

#### Changes
* Add to `NavTrailAction` new method `void addNavTrail()`, whose default implementation calls `appendNavTrail()`. Deprecate `appendNavTrail()`.
* Switch SpringActionController and other callers to invoke `addNavTrail()`.
* Migrate 749 `appendNavTrail()` implementations and usages in platform and commonAssay to `addNavTrail()`. Other repos will follow shortly.


api/src/org/labkey/api/assay/actions/BaseAssayAction.java
```diff
@@ -74,13 +74,20 @@ protected DataRegion createDataRegionForInsert(TableInfo baseTable, String lsidC
...
     public NavTree appendNavTrail(NavTree root)
     {
         root.addChild("Assay List", PageFlowUtil.urlProvider(AssayUrls.class).getAssayListURL(getContainer()));
         return root;
     }
 
+    @Override
+    public void addNavTrail(NavTree root)
+    {
+        root.addChild("Assay List", PageFlowUtil.urlProvider(AssayUrls.class).getAssayListURL(getContainer()));
```
labkey-jeckels, line 88: Do we need both of these? Perhaps, with different calls to super()?

labkey-jeckels (approved):
Do you think we should add a default addNavTrail() in a base class in a subsequent refactor? Or blow up if nobody adds anything? There are quite a few implementations that no-op.

api/src/org/labkey/api/assay/actions/BaseAssayAction.java
labkey-adam: appendNavTrail() is temporary... but I learned the hard way that it needs to stay in place until every subclass is converted to addNavTrail(). The subclasses call super.appendNavTrail() which (without this implementation) throws. This will disappear once the other repos are migrated.

labkey-adam:
I'm open to that, once every action has been migrated.

PR #1201, opened 2020-05-21
Issue 40236: Resolved date loses time portion when resolved issue is updated or closed

labkey-ankurj:
#### Rationale
This change preserves the time portion of resolved date when resolved issue is updated or closed. Resolved date is specified to be JdbcType.Date and LenientSqlDateConverter.convert() returns the java.sql.Date that sets the time components to the time in the default time zone (the time zone of the Java virtual machine running the application) that corresponds to zero GMT. The change is to change resolved date to JdbcType.TIMESTAMP.

### DIFF
--- a/issues/src/org/labkey/issue/model/Issue.java
+++ b/issues/src/org/labkey/issue/model/Issue.java
@@ -264,7 +264,7 @@ public class Issue extends Entity implements Serializable, Cloneable
 
     public Date getResolved()
     {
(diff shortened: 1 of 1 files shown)


labkey-adam (approved):
Looks good to me, but would be good if @labkey-matthewb could do a quick sanity check.

labkey-matthewb (approved):

PR #1207, opened 2020-05-21
Attempt to make file watcher junit test more reliable on Linux

labkey-adam:
#### Rationale
Current FileSystemWatcherImpl junit test is flaky, especially on Linux. This may make it more reliable.

### DIFF
--- a/api/src/org/labkey/api/files/FileSystemWatcherImpl.java
+++ b/api/src/org/labkey/api/files/FileSystemWatcherImpl.java
@@ -535,6 +535,8 @@ public class FileSystemWatcherImpl implements FileSystemWatcher
             assertTrue(b.createNewFile());
             assertTrue(c.createNewFile());
 
+            waitForEvents(events, 3);
+
             try (PrintWriter pw = PrintWriters.getPrintWriter(a))
             {
                 pw.println("Hello World");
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/files/FileSystemWatcherImpl.java
```diff
@@ -545,16 +547,18 @@ public void overflow()
...
-            // On Windows, modified events are called on delete as well, but apparently not on Linux
-            int expectedEventCount = SystemUtils.IS_OS_WINDOWS ? 14 : 10;
-            Set<String> expectedModified = SystemUtils.IS_OS_WINDOWS ? Set.of("a", "b", "c") : Set.of("a", "c");
+            waitForEvents(events, SystemUtils.IS_OS_WINDOWS ? 14 : 10);
 
-            waitForEvents(events, expectedEventCount);
+            Set<String> expectedModified = SystemUtils.IS_OS_WINDOWS ? Set.of("a", "b", "c") : Set.of("a", "c");
 
             assertEquals(3, created.size());
             assertTrue(created.containsAll(Set.of("a", "b", "c")));
```
labkey-matthewb, line 562: Did you mean to use expectedModified here instead of Set.of()?

labkey-matthewb (changes requested):

api/src/org/labkey/api/files/FileSystemWatcherImpl.java
labkey-matthewb: I see it's used below the diff.  Maybe move down?

labkey-matthewb (approved):

PR #1220, opened 2020-05-25
33862: Plate Template pages needs to be more user friendly

labkey-klum:
#### Rationale
Update the manage plate template page to address this [issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=33862).

#### Changes
- As described in the issue, change the rendering on this page to replace the list of links with a dropdown and create button.

### DIFF
--- a/assay/src/org/labkey/assay/plate/view/plateTemplateList.jsp
+++ b/assay/src/org/labkey/assay/plate/view/plateTemplateList.jsp
@@ -239,7 +239,7 @@
         }
 %>
         <br/>
-        <h4>Create New Plate Templates</h4>
+        <h4>Create New Plate Template</h4>
         <labkey:form method="POST" layout="inline" id="qc_form">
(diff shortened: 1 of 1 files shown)


assay/src/org/labkey/assay/plate/view/plateTemplateList.jsp
```diff
@@ -194,36 +206,49 @@
...
-            {
-                designerURL.replaceParameter("templateType", template);
-            %>
-                <%= link("new " + sizeDesc + handler.getAssayType() + " " + template + " template", designerURL)%><br/>
-        <%  }
         }
-    }%>
+%>
+        <br/>
+        <h4>Create New Plate Templates</h4>
```
cnathe: Since you can only create one template at a time, should we drop the `s` here and go with:
```suggestion
        <h4>Create New Plate Template</h4>
```

cnathe (approved):

PR #1221, opened 2020-05-26
portal cache

labkey-matthewb:
#### Rationale
PortalPage and Portal.WebPart are cached, but not immutable.  So, of course, they are muted...  
Made immutable versions and fixed up code that broke.


api/src/org/labkey/api/module/MultiPortalFolderType.java
```diff
@@ -273,76 +272,75 @@ private void migrateLegacyPortalPage(Container container)
...
-        List<Portal.WebPart> legacyPortalParts = new ArrayList<>(Portal.getParts(container));
-        if (!legacyPortalParts.isEmpty())
-        {
-            FolderType folderType = container.getFolderType();
-            if (null == folderType.getDefaultTab() || null == StringUtils.trimToNull(folderType.getDefaultTab().getName()))
-                return;         // Nothing to do
+        List<Portal.WebPart> legacyPortalParts = Portal.getEditableParts(container);
+        if (legacyPortalParts.isEmpty())
+            return;
+
```
labkey-adam: Can we get rid of all of this migrateLegacyPortalPage() nonsense now? Dave added it 7.5 years ago, unfortunately not as an upgrade step but as a lazy upgrade. I say 7.5 years is long enough... if no one has touched a page in that period of time then no one cares about it and we can remove the migration code. Looks like @labkey-jeckels touched this code as well, so perhaps he has an opinion.

api/src/org/labkey/api/view/Portal.java
```diff
@@ -175,8 +177,10 @@ public static void clearWebPartProperties(String nameSearchText, String properti
 
     public static final String WEBPART_PROP_LegacyPageAdded = "legacyPageAdded";
 
-    /** Bean object for persisting web part configurations in the core.portalwebparts table */
-    public static class WebPart implements Serializable
+    /** Bean object for persisting web part configurations in the core.portalwebparts table
+     * NOTE: implements Factory<> so this can be used as a builder for immutable object
+     */
+    public static class WebPart implements Serializable, Factory<WebPart>
```
labkey-adam, line 183: I'm sure this would be more invasive, but do we want to make WebPart an interface?

core/src/org/labkey/core/project/projects.jsp
```diff
@@ -398,3 +399,4 @@
     }
 
 </script>
+<% } /* if webpart */ %>
```
labkey-adam: Isn't this change already pushed in a different PR?

labkey-adam (changes requested):
Mostly looks good. Just a few questions.

api/src/org/labkey/api/module/MultiPortalFolderType.java
labkey-matthewb: I love deleting code, if everyone agrees.

api/src/org/labkey/api/view/Portal.java
labkey-matthewb: Well, I had a moment where I was contemplating all the different ways we create read-only objects.  Dual interfaces, Builders/Factories, lock() methods.  And decided not to contemplate too much.  Really I could rename WebPart -> WebPartBuilder, and the anonymous class -> WebPart, and this would be pretty vanilla.

core/src/org/labkey/core/project/projects.jsp
labkey-matthewb: oops

api/src/org/labkey/api/view/Portal.java
labkey-matthewb: But more code churn that way.

core/src/org/labkey/core/project/projects.jsp
labkey-matthewb: maybe I forked off that branch, hmmm

api/src/org/labkey/api/module/MultiPortalFolderType.java
labkey-matthewb: deleted

api/src/org/labkey/api/view/Portal.java
labkey-matthewb: leaving unless you feel strongly.

labkey-adam (approved):

PR #1222, opened 2020-05-26
Fb 38234 automation

labkey-sweta:
#### Rationale
1. Adding test coverage for setting dataset default view. ( Issue 38340)
2. Adding test coverage for setting default view for dataset with name and label. (Issue 40502) 

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/345

#### Changes
New test class added in DailyA


study/test/src/org/labkey/test/tests/study/StudyDatasetDefaultView.java
```diff
@@ -0,0 +1,111 @@
...
+        datasetTableCustomizer.saveCustomView(gridName, true);
+
+        log("Setting the default view");
+        SetDefaultPageClass setDefaultPage = datasetTable.clicksetDefault();
+        setDefaultPage.selectDefaultGrid(gridName);
+
+        log("Verifying default view for the dataset.");
+        goToProjectHome();
+        clickAndWait(Locator.linkWithText(datasetName));
+        checker().verifyTrue("Not the correct default view", isElementPresent(Locator.tagWithText("span", gridName)));
```
cnathe: Maybe also add checks for the column changes made to the saved custom view as well (i.e. those columns removed and added in the lines above).

study/test/src/org/labkey/test/tests/study/StudyDatasetDefaultView.java
```diff
@@ -0,0 +1,111 @@
...
+    @Override
+    public List<String> getAssociatedModules()
+    {
+        return Arrays.asList("study");
+    }
+
+    private void doCreateSteps()
+    {
+        _containerHelper.createProject(getProjectName(), "Study");
+        importFolderFromZip(TestFileUtils.getSampleData("studies/LabkeyDemoStudy.zip"));
```
cnathe: I think there is a version of this `importFolderFromZip` helper that will not validate queries during import. This use case definitely doesn't need to do that query validation steps, so skipping that would speed things up.

cnathe (approved):
The test case coverage seems good to me. I made a couple small suggestions. Any reason to get someone from the test team to review as well?

labkey-danield (approved):
Conditional on the extra checks Cory suggested I think this looks ok.

study/test/src/org/labkey/test/tests/study/StudyDatasetDefaultView.java
```diff
@@ -0,0 +1,117 @@
...
+import org.labkey.test.components.study.ViewPreferencesPage;
+import org.labkey.test.util.DataRegionTable;
+import org.labkey.test.util.PortalHelper;
+
+import java.util.Arrays;
+import java.util.List;
+
+@Category({DailyA.class})
+@BaseWebDriverTest.ClassTimeout(minutes = 3)
+public class StudyDatasetDefaultView extends BaseWebDriverTest
```
labkey-tchad: Test class names should end with "Test"
```suggestion
public class StudyDatasetDefaultViewTest extends BaseWebDriverTest
```

labkey-tchad (changes requested):

labkey-tchad (approved):

PR #1223, opened 2020-05-26
specimen webpart cleanup

labkey-matthewb:
#### Rationale
Clean up code don't use onclick when href will do, use propert escaping helpers


study/src/org/labkey/study/view/specimen/webPart.jsp
```diff
@@ -76,18 +72,19 @@
...
         var groupingName = grouping.name;
-        if (elementId == names.group1)
+        if (elementId === names.group1)
             document.getElementById(names.heading1).innerHTML = 'Vials by ' + groupingName;
         else
             document.getElementById(names.heading2).innerHTML = 'Vials by ' + groupingName;
 
-        var innerHTML = populateGroupingContent(grouping, false);
+        var innerHTML;
+        innerHTML = populateGroupingContent(grouping);
```
labkey-kevink, line 81:
```suggestion
        var innerHTML = populateGroupingContent(grouping);

```

labkey-kevink (approved):

study/src/org/labkey/study/view/specimen/webPart.jsp
labkey-matthewb: That's just to avoid the intellij yellow warning

PR #1224, opened 2020-05-26
Item 6620 - track script processes

labkey-ankurj:
#### Rationale
This PR updates the ExpGeneratorHelper to create an experiment run with necessary steps so that provenance service can be utilized to trace the input and output of each operation.

#### Related Pull Requests
https://github.com/LabKey/provenance/pull/19


api/src/org/labkey/api/exp/api/ProvenanceRecordingParams.java
```diff
@@ -0,0 +1,169 @@
+package org.labkey.api.exp.api;
+
+import com.google.gwt.user.client.rpc.IsSerializable;
+import org.labkey.api.util.Pair;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class ProvenanceRecordingParams implements IsSerializable
```
labkey-jeckels: IsSerializable is a GWT interface. We shouldn't be introducing new references. I suspect java.io.Serializable is what you want.

api/src/org/labkey/api/exp/api/ProvenanceRecordingParams.java
```diff
@@ -0,0 +1,169 @@
...
+{
+    private String recordingId;
+    private String name;
+    private String description;
+
+    private List<String> predecessorSteps = Collections.emptyList();
+    private String inputObjectUriProperty = ProvenanceService.PROVENANCE_INPUT_PROPERTY;
+    private String outputObjectUriProperty = "lsid";
+    private Map<String, Object> properties = Collections.emptyMap();
```
labkey-jeckels, line 20: Please add comments to indicate what the Strings are intended to be here. LSIDs? GUIDs? Etc.

api/src/org/labkey/api/exp/api/ProvenanceService.java
```diff
@@ -1,18 +1,28 @@
...
+import javax.servlet.http.HttpServletRequest;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 /**
  * Service to include non-data {@link ExpObject} and non-material to a run's ProtocolApplication steps for the purposes of lineage.
- * This service can add provenance to an {@link ExpProtocolApplication} using one of teh add Methods.
+ * This service can add provenance to an {@link ExpProtocolApplication} using one of the add Methods.
+ * Also, provides support to record provenance information across LabKey HTTP API calls. - May TBD, 2020
```
labkey-jeckels: I'm generally not a big fan of logging dates in comments, since the source control history should capture that without the need to fill in TBD values. But if you want to retain, please just use "May 2020" or be sure to fill in a date

api/src/org/labkey/api/exp/api/ProvenanceService.java
```diff
@@ -36,6 +55,11 @@ static void setInstance(ProvenanceService impl)
         ServiceRegistry.get().registerService(ProvenanceService.class, impl);
     }
 
+    static Lsid getProvenanceProtocolLSID()
+    {
+        return new Lsid(PROVENANCE_PROTOCOL_LSID);
```
labkey-jeckels: Could we make the variable an Lsid instead of a String to avoid needing to create one every time this method is called?

api/src/org/labkey/api/exp/api/ProvenanceService.java
```diff
@@ -92,4 +116,34 @@ static void setInstance(ProvenanceService impl)
      */
     Map<String, Set<ExpRun>> getRunsByLsid(Set<String> lsids);
 
+    /**
+     * Start a recording session, place RecordedActionSet in http session state and
+     * @return a GUID as the recording id.
+     */
+    String startRecording(ViewContext context, JSONObject jsonObject) throws ValidationException;
```
labkey-jeckels: Consider returning a GUID object instead of a String. It's great to have the JavaDoc indicating the value is a GUID, but having it strongly typed is even better.

api/src/org/labkey/api/pipeline/RecordedAction.java
```diff
@@ -60,6 +65,19 @@
     private Date _endTime;
     private Integer _recordCount;
 
+    // Provenance map (list of from and to pairs)
```
labkey-jeckels: Pairs of what - LSIDs? Can we use the Lsid class instead of a String?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -412,6 +415,46 @@ public ExpRunImpl createExperimentRun(Container container, String name)
...
+
+                transaction.commit();
+            }
+            return ExpGeneratorHelper.insertRun(container, user, actionSet, runName, runJobId, protocol, null, null, null);
+        }
+        catch (ExperimentException | ValidationException e)
+        {
+            LOG.error(e);
+        }
+        return null;
```
labkey-jeckels: Generally, better to rethrow the exception instead of returning null. Is there an expected error case where we want to fail and return null?

experiment/src/org/labkey/experiment/pipeline/ExpGeneratorHelper.java
```diff
@@ -116,68 +108,123 @@ static private ExpData addData(Map<URI, ExpData> datas, URI originalURI, XarSour
...
-            Map<String, ExpProtocol> protocolCache = new HashMap<>();
-            List<String> protocolSequence = new ArrayList<>();
             Map<URI, String> runOutputsWithRoles = new LinkedHashMap<>();
-            Map<URI, String> runInputsWithRoles = new HashMap<>();
-
-            runInputsWithRoles.putAll(actionSet.getOtherInputs());
+            Map<URI, String> runInputsWithRoles = new HashMap<>(actionSet.getOtherInputs());
 
-            job.info("Checking files referenced by experiment run");
+            if (log != null)
```
labkey-jeckels: Consider making the Logger @NotNull, and changing callers to pass in a standard logger to route to labkey.log if we don't have a job or other more specific log file to write to.

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -3867,6 +3874,27 @@ public ApiResponse execute(ApiSaveRowsForm apiSaveRowsForm, BindException errors
             JSONObject response = executeJson(apiSaveRowsForm.getJsonObject(), CommandType.insert, true, errors);
             if (response == null || errors.hasErrors())
                 return null;
+
+            // if there is any provenance information, save it here
+            ProvenanceService svc = ProvenanceService.get();
+            if (svc != null && apiSaveRowsForm.getJsonObject().has("provenance"))
```
labkey-jeckels: Can we introduce a no-op ProvenanceService implementation that's registered in the absence of the provenance module so that all callers don't have to do null checks?

labkey-jeckels (changes requested):
Looking good. A handful of suggestions in the comments.

api/src/org/labkey/api/exp/api/ProvenanceRecordingParams.java
labkey-ankurj: Replaced

api/src/org/labkey/api/exp/api/ProvenanceRecordingParams.java
labkey-ankurj: done

api/src/org/labkey/api/exp/api/ProvenanceService.java
labkey-ankurj: removed the date, not sure why I added it in the first place.

api/src/org/labkey/api/exp/api/ProvenanceService.java
labkey-ankurj: this method was not being used, removed. Need to append recording Id to this string to create final LSID, so will be leaving this as a String.

api/src/org/labkey/api/exp/api/ProvenanceService.java
labkey-ankurj: done

api/src/org/labkey/api/pipeline/RecordedAction.java
labkey-ankurj: updated the comment

experiment/src/org/labkey/experiment/pipeline/ExpGeneratorHelper.java
labkey-ankurj: Done

query/src/org/labkey/query/controllers/QueryController.java
labkey-ankurj: Added

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-ankurj: done

api/src/org/labkey/api/exp/api/ProvenanceService.java
```diff
@@ -92,4 +120,34 @@ static void setInstance(ProvenanceService impl)
...
+    /**
+     * Start a recording session, place RecordedActionSet in http session state and
+     * @return a GUID as the recording id.
+     */
+    GUID startRecording(ViewContext context, JSONObject jsonObject) throws ValidationException;
+
+    /**
+     * Get the current recording session from http session state, and add the actionSet
+     */
+    void addRecordingStep(HttpServletRequest request, String recordingId, RecordedAction action);
```
labkey-jeckels: These recordingId values could be GUIDs too, like the return value from startRecording()

api/src/org/labkey/api/exp/api/DefaultProvenanceProvider.java
```diff
@@ -0,0 +1,135 @@
...
+import org.labkey.api.util.Pair;
+import org.labkey.api.view.ViewContext;
+
+import javax.servlet.http.HttpServletRequest;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class DefaultProvenanceProvider implements ProvenanceService
```
labkey-jeckels, line 25: Add a comment that this is a no-op implementation.

labkey-jeckels (changes requested):
Based on our call yesterday, I suspect there are more changes coming here so I'll hold off on approving

labkey-jeckels (commented):
One more suggestion

api/src/org/labkey/api/exp/api/DefaultProvenanceProvider.java
labkey-ankurj: added

api/src/org/labkey/api/exp/api/ProvenanceService.java
labkey-ankurj: done

labkey-jeckels (approved):

PR #1227, opened 2020-05-27
remove ENSURE_DOMAIN_LOCK to avoid deadlocks

labkey-matthewb:
#### Rationale
To avoid deadlocks we need to avoid doing SQL operations while holding java locks.

#### Changes
Remove ENSURE_DOMAIN_LOCK and move the ensureDomain functionality into PropertyService.


labkey-jeckels (commented):
These changes look OK to me, but it would be good to get @labkey-adam to look too in case he remembers any specifics to consider from these issues back in 2014:

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=21199

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=21128

labkey-adam:
I don't remember a lot from 2014... this looks okay to me

labkey-adam (approved):

PR #1228, opened 2020-05-27
check for null container (e.g. container deleted)

labkey-matthewb:
#### Rationale
Fix NPE caused by race condition


labkey-matthewb:
Not a very interesting fix, but does raise the question of whether we could miss deleting objects from the index because we do most index operations asynchronously.

labkey-adam:
For the record, SearchContainerListener appears to clear out everything associated with a deleted container.

labkey-adam (approved):

PR #1231, opened 2020-05-27
Remove unused parameters and consolidate duplicated methods in URL classes

labkey-jeckels:
#### Changes
Remove identical methods, and remove parameters that always have the same value

Add method to get just the relative link, of the form controller-action.view?params#fragment

### DIFF
--- a/api/src/org/labkey/api/view/ActionURL.java
+++ b/api/src/org/labkey/api/view/ActionURL.java
@@ -194,17 +194,24 @@ public class ActionURL extends URLHelper implements Cloneable
         setScheme(request.getScheme());
     }
 
+    /** @return just the controller-action.view?parameters#fragment part of the URL, omitting the server name, context path, and container path */
+    public String toContainerRelativeURL()
+    {
+        StringBuilder result = new StringBuilder(toPathStringNew(Path.emptyPath, _controller, _action, Path.emptyPath));
+        appendParamsAndFragment(result, false);
+        return result.toString();
+    }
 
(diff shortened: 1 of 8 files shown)


labkey-kevink (approved):

api/src/org/labkey/api/view/ActionURL.java
```diff
@@ -564,9 +571,9 @@ protected boolean isDirectory()
     
 
     @Override
-    public String getPath(boolean asForward)
```
labkey-matthewb, line 567: I'm surpised asForward is no longer used.  It was meant for a cases like FileServlet.service() where the contextPath needs to be dropped.

api/src/org/labkey/api/view/ActionURL.java
```diff
@@ -194,17 +194,24 @@ public ActionURL(Class<? extends Controller> actionClass, Container container)
         setScheme(request.getScheme());
     }
 
+    /** @return just the controller-action.view?parameters#fragment part of the URL, omitting the server name, context path, and container path */
+    public String toContainerRelativeURL()
```
labkey-matthewb, line 198: JspBase.buildURL() does something like this.  Generating local relative URL.  Can that be converted?

labkey-matthewb (approved):

api/src/org/labkey/api/view/ActionURL.java
labkey-jeckels: I think so - take a look at the commit I just pushed.

labkey-matthewb (approved):

PR #1233, opened 2020-05-28
migrateLegacyPortalPage() still does something that makes FolderTypes work

labkey-matthewb:
#### Rationale
Removed migrateLegacyPortalPage() in recent "cleanup" because it "looks" like it does something with only old portal definitions.  Turns out it is needed to make create folder with FolderType work.

####  Changes
revert change that removed migrateLegacyPortalPage


labkey-adam (approved):
😿

PR #1237, opened 2020-05-28
Comments for usage of getSelectName()

labkey-matthewb:
#### Rationale
Usage of getSelectName() is not clear

#### Changes
Add comments
Enforce that getSelectName() is only used for SchemaTableInfo columns


api/src/org/labkey/api/data/BaseColumnInfo.java
```diff
@@ -535,12 +535,22 @@ public void setMetaDataName(String metaDataName)
...
     {
-        return _metaDataName;      // Actual name returned by metadata; use to query meta data or to select columns enclosed in quotes
+        return _metaDataName;
     }
 
-
     @Override
     public String getSelectName()
+    {
+        assert getParentTable() instanceof SchemaTableInfo : "Use getValueSql()";
```
labkey-jeckels, line 544: Is this assert valuable? Won't it just blow up on the subsequent check with an UnsupportedOperationException anyway?

labkey-jeckels (approved):

api/src/org/labkey/api/data/BaseColumnInfo.java
labkey-matthewb: The difference is that the UnsupportedOperationException only gets thrown if _selectName is null, but there were code paths that were copying _selectName ( col.setSelectName(from.getSelectName()).  So the assert is for extra paranoia without runtime overhead.,

PR #1238, opened 2020-05-28
Issue39899: Expose Study Schema's VisitMap table for SQL Queries

RosalineP:
#### Rationale
We want to support using visit-dataset mapping data in sql queries. This requires making the VisitMap table queryable.

#### Changes
* Surfaces VisitMap for queries

### DIFF
--- a/study/src/org/labkey/study/query/VisitMapTable.java
+++ b/study/src/org/labkey/study/query/VisitMapTable.java
@@ -1,19 +1,12 @@
 package org.labkey.study.query;
 
-import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 import org.labkey.api.data.ContainerFilter;
 import org.labkey.api.data.JdbcType;
 import org.labkey.api.data.SQLFragment;
 import org.labkey.api.data.TableInfo;
-import org.labkey.api.query.DefaultQueryUpdateService;
 import org.labkey.api.query.ExprColumn;
 import org.labkey.api.query.FieldKey;
-import org.labkey.api.query.FilteredTable;
 import org.labkey.api.query.LookupForeignKey;
-import org.labkey.api.query.QueryUpdateService;
-import org.labkey.api.security.UserPrincipal;
-import org.labkey.api.security.permissions.Permission;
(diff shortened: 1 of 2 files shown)


study/src/org/labkey/study/query/VisitMapTable.java
```diff
@@ -0,0 +1,45 @@
...
+        addColumn(visitIdColumn);
+
+        SQLFragment sql = new SQLFragment(ExprColumn.STR_TABLE_ALIAS + ".DataSetId");
+        var datasetLookupCol = new ExprColumn(this, "DataSet", sql, JdbcType.VARCHAR);
+        datasetLookupCol.setFk(new LookupForeignKey(cf, "DataSetId", "Name")
+        {
+            @Override
+            public TableInfo getLookupTableInfo()
+            {
+                return new DatasetsTable(schema, getLookupContainerFilter());
```
labkey-klum: this looks like it will work but is an outdated way of hooking up a FK, Matt introduced a builder pattern which is nifty but it hard to find examples of, here's one:

var dataSetColumn = new AliasedColumn(this, "DataSet", _rootTable.getColumn("DataSet"));
dataSetColumn.setFk(studyFK().to("DataSets", "DataSetId", null));
addColumn(dataSetColumn);

study/src/org/labkey/study/query/VisitMapTable.java
```diff
@@ -0,0 +1,45 @@
...
+        addFolderColumn();
+        addWrapColumn(getRealTable().getColumn(FieldKey.fromParts("Required")));
+
+        var visitIdColumn = wrapColumn("Visit", _rootTable.getColumn("VisitRowId"));
+        LookupForeignKey visitIdFk = new LookupForeignKey()
+        {
+            @Override
+            public TableInfo getLookupTableInfo()
+            {
+                return schema.createTable("Visit");
```
labkey-klum: I think this version is deprecated due to container filter, I would use this instead:
return new VisitTable(_userSchema, getLookupContainerFilter());

labkey-klum (approved):
note comments, also might be good to edit the PR writeup since we are no longer having the table support updates.

PR #1249, opened 2020-06-01
propfind + getwebpart

labkey-matthewb:
#### Rationale
Use standard parameter validation
- don't throw unhandled exception
- bad xml formatting

### DIFF
--- a/core/src/org/labkey/core/webdav/DavController.java
+++ b/core/src/org/labkey/core/webdav/DavController.java
@@ -2001,30 +2001,6 @@ public class DavController extends SpringActionController
     }
 
 
-    private static boolean isValidXmlElementName(String name)
-    {
-            /*https://www.w3schools.com/xml/xml_elements.asp
-            Element names are case-sensitive
-            Element names must start with a letter or underscore
-            Element names cannot start with the letters xml (or XML, or Xml, etc)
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/webdav/DavController.java
```diff
@@ -2473,7 +2497,8 @@ else if (property.equals("custom"))
                         xml.writeElement(null, "prop", XMLWriter.OPENING);
                         for (String property : propertiesNotFound)
                         {
-                            xml.writeElement(null, property, XMLWriter.NO_CONTENT);
+                            if (isValidXmlElementName(property))
```
labkey-adam: Do you need another check like this later, around line 2675 in the original file?

labkey-adam (changes requested):

core/src/org/labkey/core/webdav/DavController.java
labkey-matthewb: Yes

labkey-adam (approved):

labkey-adam:
One more question: as currently written, you could end up with a \<propstat>\<prop> block with no properties. (If all propertiesNotFound are malformed.) Is that a problem? Solution would be to filter the list before checking the size, of course.

PR #1259, opened 2020-06-03
Item #7190: labbook data model refactor

labkey-kevink:
#### Rationale
The labbook schema change in labbook-20.000-20.0001.sql renames tables and drops the material FK.  However, the deferred upgrade addDbSequenceForMaterialsRowIdIfMissed invoked in exp-20.004-20.005.sql attempts to drop a constraint from a table that no longer exists.

The solution is to use core.fn_dropIfExists to gracefully handle constraints that no longer exist.

#### Related Pull Requests
* LabKey/labbook#27


labkey-susanh (approved):
Awesome indeed.

PR #1261, opened 2020-06-03
fix sharing participant group

labkey-matthewb:
#### Rationale
BrowseParticipantGroup is supposed to allow copying of non-public participant groups to allow sharing.

### DIFF
--- a/study/src/org/labkey/study/controllers/ParticipantGroupController.java
+++ b/study/src/org/labkey/study/controllers/ParticipantGroupController.java
@@ -527,7 +527,8 @@ public class ParticipantGroupController extends BaseStudyController
                         }
                         else if (form.getGroupId() != -1)
                         {
-                            // NOTE this method reads from the cache (faster), but will not find personal groups that have been shared for instance
+                            // NOTE this method reads from the cache (faster), but will not find personal groups that have been shared, for instance
+
                             ParticipantGroup group = ParticipantGroupManager.getInstance().getParticipantGroup(getContainer(), getUser(), form.getGroupId());
                             if (null == group)
                             {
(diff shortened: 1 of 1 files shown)


study/src/org/labkey/study/controllers/ParticipantGroupController.java
```diff
@@ -527,8 +527,14 @@ public ApiResponse execute(BrowseGroupsForm form, BindException errors)
...
                         {
-                            // NOTE: this can expose the participant group information to a user that can't otherwise see it via the standard UI in the study module
+                            // NOTE this method reads from the cache (faster), but will not find personal groups that have been shared for instance
                             ParticipantGroup group = ParticipantGroupManager.getInstance().getParticipantGroup(getContainer(), getUser(), form.getGroupId());
+                            if (null == group)
+                            {
+                                // NOTE: This can expose the participant group information to a user that can't otherwise see it via the standard UI in the study module.
+                                // NOTE: This is intentional as 'personal' groups are not secured and can be shared.
+                                group = ParticipantGroupManager.getInstance().getParticipantGroupFromGroupRowId(getContainer(), getUser(), form.getGroupId());
+                            }
```
labkey-susanh, line 538: Perhaps use an `else if` below this.

study/src/org/labkey/study/controllers/ParticipantGroupController.java
```diff
@@ -527,8 +527,14 @@ public ApiResponse execute(BrowseGroupsForm form, BindException errors)
                         }
                         else if (form.getGroupId() != -1)
                         {
-                            // NOTE: this can expose the participant group information to a user that can't otherwise see it via the standard UI in the study module
+                            // NOTE this method reads from the cache (faster), but will not find personal groups that have been shared for instance
```
labkey-susanh:
```suggestion
                            // NOTE this method reads from the cache (faster), but will not find personal groups that have been shared, for instance

```

labkey-susanh (approved):
Just minor suggestions

PR #1263, opened 2020-06-03
TabBufferedReader, a special BufferedReader subclass that accommodates infer fields

labkey-adam:
#### Rationale
`TabLoader` uses `BufferedReader.mark(int)` to read data for inferring column data types, then `BufferedReader.reset()` to back up and actually import the data. `mark(int)` specifies the buffer size and `BufferedReader` allocates the entire buffer upfront and never resizes it, which restricts our ability to infer a large number of rows, since every import incurs the cost of a large buffer. `TabBufferedReader` addresses this limitation and provides more flexibility. It extends a new implementation of `BufferedReader` that starts with a small read-ahead buffer and grows it as needed. The goal here is to accommodate inferring and importing very large tables while using low memory footprint for smaller tables.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1244
* https://github.com/LabKey/platform/pull/1257

### DIFF
--- a/api/src/org/labkey/api/reader/TabBufferedReader.java
+++ b/api/src/org/labkey/api/reader/TabBufferedReader.java
@@ -1,0 +1,157 @@
+package org.labkey.api.reader;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.Reader;
+import java.lang.reflect.Field;
+
+/**
+ * A subclass of {@link BufferedReader} that's specifically designed to work with {@link TabLoader}'s infer fields
+ * capability. Our previous approach used {@link BufferedReader#mark(int)} to set a fixed size read-ahead buffer prior
+ * to import. BufferedReader allocates the entire buffer upfront and never resizes it, which restricted our ability to
+ * infer a large number of rows, since every import incurred the cost of a large buffer. TabBufferedReader starts with
+ * a relatively small read-ahead buffer and grows it as needed, the goal being to accommodate inferring and importing
+ * very large tables while incurring low memory footprint for smaller tables. This class also adjusts its behavior if
+ * the exact size of the input is known, for example, when reading from a file or a CharSequence; in this case, the
+ * buffer will never grow larger than the input size.
+ *
+ * Read-ahead mode is initiated by calling {@link #setReadAhead()} and ended by calling {@link #resetReadAhead()}. The
+ * key parameters that control read-ahead buffering are immediately below and can be adjusted to refine the behavior.
+ * The read-ahead buffer is initialized to input size or INITIAL_BUFFER_SIZE, whichever is smaller. Before each line is
+ * read, the remaining buffer space is checked; if it's smaller than two times the largest row seen so far then the
+ * buffer is expanded. The new buffer size is double the previous size, though growth is always limited to
+ * MAX_BUFFER_SIZE_INCREMENT and total size is never larger than the smaller of ABSOLUTE_MAX_BUFFER_SIZE or input size.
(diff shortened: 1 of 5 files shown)


labkey-adam:
Not a lot of new code, but this deserves some real scrutiny on the code review.

api/src/org/labkey/api/reader/TabBufferedReader.java
```diff
@@ -0,0 +1,157 @@
...
+            return line;
+        }
+
+        return null;
+    }
+
+    // Try to ensure enough room in the buffer for the line that's about to be read. We don't know how long that line
+    // is, so just verify that there's enough buffer space for double the longest line seen so far; if not, attempt to
+    // expand the buffer. Return false if we can't expand the buffer any more; in that case, caller must stop reading.
+    private boolean ensureBuffer()
```
labkey-matthewb: Could this manipulation of internal field values be accomplished by using the public methods?  E.g. to grow the buffer
reset(); // get back to mark
mark(targetBufferSize); // growbuffer
skip(backToWhereWeWere); // continue on...

api/src/org/labkey/api/reader/TabBufferedReader.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public void resetReadAhead() throws IOException
+    {
+        reset();
+        _readAhead = false;
+        _markedChar = -1;
+    }
+
+    @Override
+    public String readLine() throws IOException
```
labkey-matthewb, line 84: Does this class need to override other read() methods?  Maybe they are safe, but first thought is that they should be overridden or throw Unsupported.

labkey-matthewb (changes requested):
I think you can accomplish this without the reflection with clever use of the existing public methods.  The catch would seem to be that BufferedReader doesn't have any sort of getPosition() so you need to keep track of that (or cheat and reflect for that field).  See https://commons.apache.org/proper/commons-csv/jacoco/org.apache.commons.csv/ExtendedBufferedReader.java.html

api/src/org/labkey/api/reader/TabBufferedReader.java
labkey-adam: It's not meant to be a general purpose replacement for BufferedReader, so yes, probably should. skip() as well.

api/src/org/labkey/api/reader/TabBufferedReader.java
labkey-adam: This is exactly what my first attempt did. That grows the buffer (a bit less efficiently), but the problem is I don't see a way to reliably count characters (including all line terminators) without either reflection or duplicating a lot of BufferedReader. Once I'm using reflection to get the character count I might as well set the buffer size use reflection as well.

labkey-adam:
I don't see a way to count characters consumed inside of readLine() without reflection... but if you do, I'm all for it.

api/src/org/labkey/api/reader/TabBufferedReader.java
labkey-matthewb: Yeah, that's why I mentioned ExtendedBufferedReader, it tracks position.  Could maybe copy that and add your functionality.

labkey-adam: That class tracks position if you call read(), but not readLine(), which is what TabLoader uses. We could roll our own readLine and count characters, but there's a lot of stuff in there and of course it manipulates all the same private members.

labkey-jeckels:
Since Matt's already jumped in, I am not planning to do a separate review here. Let me know if I'd be useful to break a tie or do a full independent review.

labkey-matthewb (approved):

PR #1264, opened 2020-06-04
Issue 40229: targetedms lookups target DB schema TableInfo

labkey-jeckels:
#### Rationale
We want lookups in the tables that we expose through the schema browser to be targeting other tables as exposed in the schema browser, not the lower level SchemaTableInfos

#### Changes
Improve metadata on special lookup types so that they don't report null for their schema name and give a better showing in the schema browser


labkey-jeckels:
@labkey-matthewb can you review this one?

labkey-matthewb (approved):

PR #1267, opened 2020-06-04
Item 7374: Add productId to menu models to facilitate linking between applications

labkey-susanh:
#### Rationale
The freezer management functionality being incorporated into sample manager will be in its own application and we need to be able to link from one application to the other within the product menu.  We do this by attaching productIds to each menu item in order to compare to the current product to know whether to generate an in-app URL or not.

#### Related Pull Requests
* https://github.com/LabKey/inventory/pull/35
* https://github.com/LabKey/biologics/pull/602
* https://github.com/LabKey/sampleManagement/pull/287
* https://github.com/LabKey/labkey-ui-components/pull/267

#### Changes
* Add productId to MenuItem and MenuSection models
* Update ProductRegistry to handle a set of productIds when getting menu sections
* Update ProductRegistry to allow for providing no productId to indicate all menu sections in the container should be used.


core/api-src/org/labkey/api/products/MenuItem.java
```diff
@@ -25,34 +25,26 @@
     private String _url; // the URL on the server side
     private Integer _orderNum; // ordinal for producing the primary sort order of the items
     private Boolean _requiresLogin = false; // indicates if link should be shown if not logged in.
+    private String _productId = null; // indicates the product/application this link should direct to.  Can (should?) be null if the current application is to be used.
 
-    public MenuItem(String label, String url, Integer id, String key, Integer orderNum)
+    public MenuItem(String label, String url, Integer id, String key, Integer orderNum, String productId)
```
cnathe, line 30: could have optionally left a MenuItem constructor which didn't have the new productId param so that the "null" case didn't have to change....but maybe we do want those cases to explicitly pass null so that they are intentional in setting it as "use the current productId".

core/api-src/org/labkey/api/products/ProductRegistry.java
```diff
@@ -62,19 +68,64 @@ public void registerMenuItemsProvider(ProductMenuProvider provider)
...
+    public List<MenuSection> getProductMenuSections(@NotNull ViewContext context, @NotNull Container container, @Nullable Integer itemLimit)
+    {
+        Set<String> modules = container.getActiveModules().stream().map(Module::getName).collect(Collectors.toSet());
+        List<MenuSection> sections = new ArrayList<>();
+        ProductMenuProvider userMenuProvider = null;
+        for (String module : modules)
+        {
+           if (_moduleProviderMap.containsKey(module))
+           {
+               // TODO this works for now since Inventory doesn't provide a user menu, but we'll likely need a better resolution strategy here in general
```
cnathe: I'm still a little confused from the demo of this you showed me where the user menu part was working as expected between the FM only app vs the SM and FM app combination case. Where is the user menu info coming from in the case where you only have the inventory module enabled in your container? Or maybe the issue would only show up if you had a server which only had inventory but not the SM module?

core/api-src/org/labkey/api/products/ProductRegistry.java
```diff
@@ -62,19 +68,64 @@ public void registerMenuItemsProvider(ProductMenuProvider provider)
...
+                // TODO this works for now since Inventory doesn't provide a user menu, but we'll likely need a better resolution strategy here in general
+                if (userMenuProvider == null)
+                    userMenuProvider = _productMap.get(productId);
+                sections.addAll(_productMap.get(productId).getSections(context, itemLimit));
+            }
+        }
+        // always include the user menu as the last item
+        if (userMenuProvider != null)
+            sections.add(new UserInfoMenuSection(context, userMenuProvider));
+        return sections;
```
cnathe, line 125: These two getProductMenuSections methods seem so similar, but I'm not seeing a good way to "share" any of this code between them.

cnathe (approved):

core/api-src/org/labkey/api/products/MenuItem.java
labkey-susanh: Yeah, there were just too many constructors for this object and I think we almost always want a non-null id.

core/api-src/org/labkey/api/products/ProductRegistry.java
labkey-susanh: FM now does have a user menu.  This comment is out of date.  It was working as expected only because of an accident of ordering. I've updated both these methods to accept a currentProductId in order to pick the userMenuProvider.

PR #1272, opened 2020-06-05
enable crawler

Sigmonia:
#### Rationale
Fix searchTests

#### Changes
Enable the search crawler for the duration of the Test(s) -- this is needed due to files being queued into the crawler for reindexing after a folder move


labkey-susanh (commented):
Changes seem technically fine, but the don't seem to have fixed the test.

labkey-tchad:
I think those errors were from deleting the search index immediately after enabling the crawler. I just swapped those lines (and added some logging in https://github.com/LabKey/testAutomation/pull/356)

labkey-tchad (approved):
File searches seem to be working now.
The current failures are unrelated. One is a random failure I've seen before (index not updating to reflect an updated row in a list). The other is one I haven't seen before but seems unrelated to search (failing to update the folder's security policy).

PR #1274, opened 2020-06-05
Add URL for pending requests to the thread dump output

labkey-matthewb:
#### Rationale
To add more context to thread dumps, add the URL that each request thread is processing to the output
currently only for ViewServlet


labkey-jeckels (approved):
Nice! I've thought about adding this in the past, but never got around to it. Thanks.

PR #1284, opened 2020-06-08
Issue 40594: Extend AnnouncementService to allow posting to a message thread

labkey-jeckels:
#### Rationale
We have an API to create a new thread, but not add to an existing one. We have a new use case for the latter.

#### Changes
Simple pass-through argument to specify the parent thread, and to expose the EntityId for an announcement

### DIFF
--- a/announcements/src/org/labkey/announcements/api/AnnouncementServiceImpl.java
+++ b/announcements/src/org/labkey/announcements/api/AnnouncementServiceImpl.java
@@ -15,6 +15,7 @@
  */
 package org.labkey.announcements.api;
 
+import org.jetbrains.annotations.Nullable;
 import org.labkey.announcements.AnnouncementsController;
 import org.labkey.announcements.model.AnnouncementManager;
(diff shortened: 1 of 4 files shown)


announcements/src/org/labkey/announcements/api/AnnouncementServiceImpl.java
```diff
@@ -52,11 +59,21 @@ public Announcement insertAnnouncement(Container c, User u, String title, String
...
-                              
+
         AnnouncementModel insert = new AnnouncementModel();
         insert.setTitle(title);
         insert.setBody(body);
 
+        if (parentRowId != null)
+        {
+            Announcement parentAnnouncement = getAnnouncement(c, u, parentRowId);
+            if(parentAnnouncement == null)
```
labkey-susanh:
```suggestion
            if (parentAnnouncement == null)

```

labkey-susanh (approved):

PR #1285, opened 2020-06-08
Issue 40599: Support folder aliases for redirecting action-less URLs

labkey-jeckels:
#### Rationale
We only look at current container paths for resolving this lookup. We could easily support folder aliases as well, which would enable some new use cases in Panorama Public where a container gets copied as part of a workflow, and we want to have a stable URL that points to the original container before the copy, and the new copy afterwards.

#### Changes
Use existing method that first checks for a container with the exact path, and then checks for aliased containers

### DIFF
--- a/api/src/org/labkey/api/data/ContainerManager.java
+++ b/api/src/org/labkey/api/data/ContainerManager.java
@@ -2248,6 +2248,12 @@ public class ContainerManager
     @Nullable
     private static Container resolveContainerPathAlias(String path, boolean top)
     {
+        // Strip any trailing slashes
+        while (path.endsWith("/"))
+        {
+            path = path.substring(0, path.length() - 1);
+        }
+
         // Simple case -- resolve directly (sans alias)
         Container aliased = getForPath(path);
         if (aliased != null)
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -2248,6 +2248,12 @@ public static Container resolveContainerPathAlias(String path)
     @Nullable
     private static Container resolveContainerPathAlias(String path, boolean top)
     {
+        // Strip any trailing slashes
+        while (path.endsWith("/"))
+        {
+            path = path.substring(0, path.length() - 1);
```
labkey-matthewb, line 2254: StringUtils.strip()?

labkey-matthewb (approved):

core/src/org/labkey/core/webdav/DavController.java
```diff
@@ -837,7 +837,7 @@ public final WebdavStatus doMethod() throws DavException, IOException
                 // might be a little cleaner to register a javax.servlet.Filter
                 if (x.getStatus() != WebdavStatus.SC_NOT_FOUND || !"GET".equals(method))
                     throw x;
-                Container c = ContainerManager.getForPath(getResourcePath());
+                Container c = ContainerManager.resolveContainerPathAlias(getResourcePath().toString());
```
labkey-matthewb, line 840: Do we want aliases to apply to webdav resources as well: @files, @pipeline?  We could make the resolver respect aliases, but we'd want to make sure we don't double index files.

labkey-jeckels: So far we don't have a use case for that

PR #1286, opened 2020-06-08
Create method SearchService.deleteResources() to avoid overhead of calling deleteResource() one document at a time.

labkey-matthewb:
#### Rationale
truncateSampleSet() is very slow, this reduces the cost due to updating the full-text index. 

#### Changes
* move the delete document work to the background indexing thread 
* avoid a lot of IndexManager.deleteDocument() overhead (calling maybeRefresh() for each document)

#### Related
https://github.com/LabKey/platform/pull/1286
https://github.com/LabKey/testAutomation/pull/375


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -17,6 +17,7 @@
 package org.labkey.experiment.api;
 
 import com.google.common.collect.Iterables;
+import com.google.gwt.user.client.Command;
```
labkey-adam: Remove unused import

labkey-adam (changes requested):
Just remove the unused import

labkey-adam (approved):

labkey-adam (approved):

PR #1288, opened 2020-06-08
Issue 40540: Updating a sample by re-importing does not create a timeline entry

XingY:
#### Rationale
On import with merge, sample.rowId is auto incremented as sql param. We don't re-select DbSequenceColumn columns to update those values, so the data map will always have the incremented rowId, which will be problematic for its consumer (audit log creation). 

#### Related Pull Requests
* 

#### Changes
* reselect sample rowId on merge for detailed logging

### DIFF
--- a/api/src/org/labkey/api/dataiterator/DetailedAuditLogDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/DetailedAuditLogDataIterator.java
@@ -20,15 +20,12 @@ import org.labkey.api.data.AuditConfigurable;
 import org.labkey.api.data.ColumnInfo;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.TableInfo;
-import org.labkey.api.exp.api.SampleSetService;
-import org.labkey.api.exp.query.SamplesSchema;
 import org.labkey.api.gwt.client.AuditBehaviorType;
 import org.labkey.api.query.BatchValidationException;
 import org.labkey.api.query.QueryService;
 import org.labkey.api.security.User;
 
 import java.io.IOException;
-import java.util.Collections;
 
 import static org.labkey.api.gwt.client.AuditBehaviorType.DETAILED;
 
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -135,6 +137,39 @@ public void addAuditEvent(User user, Container container, AuditBehaviorType audi
...
+            {
+                AuditBehaviorType sampleAuditType = auditBehavior;
+                if (sampleAuditType == null || getXmlAuditBehaviorType() != null)
+                    sampleAuditType = getAuditBehavior();
+
+                if (sampleAuditType == AuditBehaviorType.DETAILED || sampleAuditType == AuditBehaviorType.SUMMARY)
+                {
+                    // material.rowid is a dbsequence column that auto increments during merge, even if rowId is not updated
+                    // need to reselect rowId
+                    if (params.containsKey(LSID))
```
labkey-susanh, line 157: If there isn't an LSID or we don't get the sample back using the LSID and thus we won't have the proper ROW_ID, do we still want to add the audit event?  I guess it's better than dropping it on the floor, but only slightly.  I don't really have a better suggestion, though.

api/src/org/labkey/api/dataiterator/DetailedAuditLogDataIterator.java
```diff
@@ -88,12 +85,7 @@ public boolean next() throws BatchValidationException
                 auditType = auditConfigurable.getAuditBehavior();
 
             if (auditType == DETAILED)
-            {
-                if (_table.getPublicSchemaName().equalsIgnoreCase(SamplesSchema.SCHEMA_NAME))
-                    SampleSetService.get().addAuditEvent(_user, _container, _table, auditType, _auditAction, Collections.singletonList(((MapDataIterator) _data).getMap()));
-                else
-                    QueryService.get().addAuditEvent(_user, _container, _table, auditType, _auditAction, Collections.singletonList(((MapDataIterator) _data).getMap()));
-            }
+                _table.addAuditEvent(_user, _container, auditType, _auditAction, ((MapDataIterator) _data).getMap());
```
labkey-susanh, line 88: Definitely like this.  Thanks for doing that!

labkey-susanh (approved):
Just one question about an edge case that may never happen.

PR #1291, opened 2020-06-09
Issue 38105: Override box plot x-axis sorting for study visit labels

cnathe:
#### Rationale
Fox box plots, the default sorting function for the x-axis box labels is to sort them alphabetically. This makes sense as the default, but when the x-axis selected measure is the study visit column, this isn't ideal. In that case, the user would like for the sorting to be based on the visit's sequenceNum.

#### Changes
* when x-axis measures is ParticipantVisit/Visit explicitly sort the data by sequenceNum instead

### DIFF
--- a/visualization/resources/web/vis/chartWizard/genericChartPanel.js
+++ b/visualization/resources/web/vis/chartWizard/genericChartPanel.js
@@ -723,7 +723,7 @@ Ext4.define('LABKEY.ext4.GenericChartPanel', {
 
         // Issue 38105: For box plot of study visit labels, sort by sequenceNum
         if (this.renderType === 'box_plot' && this.measures.x && this.measures.x.fieldKey === 'ParticipantVisit/Visit') {
-            sortKey = 'ParticipantVisit/SequenceNum';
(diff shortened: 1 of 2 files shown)


visualization/resources/web/vis/chartWizard/genericChartPanel.js
```diff
@@ -719,6 +719,12 @@ Ext4.define('LABKEY.ext4.GenericChartPanel', {
     getQueryConfig : function(serialize)
     {
         var dataRegion = LABKEY.DataRegions[this.panelDataRegionName];
+        var sortKey = 'lsid'; // needed to keep expected ordering for legend data
+
+        // Issue 38105: For box plot of study visit labels, sort by sequenceNum
+        if (this.renderType === 'box_plot' && this.measures.x && this.measures.x.fieldKey === 'ParticipantVisit/Visit') {
+            sortKey = 'ParticipantVisit/SequenceNum';
```
labkey-adam: Admins can configure a custom display order... should this be something like ParticipantVisit/Visit/DisplayOrder?

cnathe: Oh yes, good point! I'll look into that.

labkey-klum (approved):

PR #1299, opened 2020-06-11
support for provenance in query apis

labkey-klum:
#### Rationale
We want to be able to pass in provenance related information to a subset of our query APIs (insert, update, delete) to allow them to participate in a provenance recording and contribute provenance mappings that will show up in the provenance run.

[master specification](https://docs.google.com/document/d/1AM7gm80JUq9iHpu529vNwPYA6KDl57oxY1zVMKlAMlU/edit#)

#### Related Pull Requests
- https://github.com/LabKey/provenance/pull/29/files
#### Changes
- Moved the code responsible for handling provenance recording params into the BaseSaveRowsAction.


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -3796,6 +3796,31 @@ protected JSONObject executeJson(JSONObject json, CommandType commandType, boole
...
+                {
+                    JSONObject provenanceJSON = json.getJSONObject("provenance");
+                    ProvenanceRecordingParams params = svc.createRecordingParams(getViewContext(), provenanceJSON, ProvenanceService.ADD_RECORDING);
+                    RecordedAction action = svc.createRecordedAction(getViewContext(), params);
+                    if (action != null && params.getRecordingId() != null)
+                    {
+                        // check for any row level provenance information
+                        if (json.has("rows"))
+                        {
+                            JSONArray jsonRows = json.getJSONArray("rows");
```
labkey-ankurj: I guess this will throw exception if 'rows' is not an array.

labkey-ankurj (changes requested):
looks good, one comment.

query/src/org/labkey/query/controllers/QueryController.java
labkey-klum: done

labkey-ankurj (approved):

PR #1301, opened 2020-06-12
Issue 40621: Lineage update is not accurately represented in timeline event details

XingY:
#### Rationale
For LKSM timeline feature, we want to track lineage (data and material inputs) updates for samples when a sample is updated. Since data and material inputs are not added to samples table by default, those information are not available in the sample's row detail information. This pr will add data/material input fields to sample full row data. 

#### Related Pull Requests
* 

#### Changes
* added getNearestParentMaterials to ExperimentService
* Include material's data and material inputs in getRow


experiment/src/org/labkey/experiment/api/SampleSetUpdateServiceDI.java
```diff
@@ -373,7 +382,55 @@ else if (lsid != null)
...
             throw new QueryUpdateServiceException("Either RowId or LSID is required to get Sample Set Material.");
 
-        return new TableSelector(getQueryTable(), filter, null).getMap();
+        Map<String, Object> sampleRow = new TableSelector(getQueryTable(), filter, null).getMap();
+
+        if (!addInputs)
+            return sampleRow;
+
+        ExperimentService experimentService = ExperimentService.get();
+        ExpMaterial seed = rowId != null ? experimentService.getExpMaterial(rowId) : experimentService.getExpMaterial(lsid);
```
labkey-susanh, line 391: Are we sure lsid is not null here?

experiment/src/org/labkey/experiment/api/SampleSetUpdateServiceDI.java
```diff
@@ -373,7 +382,55 @@ else if (lsid != null)
...
+        Set<ExpMaterial> parentSamples = experimentService.getNearestParentMaterials(container, user, seed);
+        if (parentSamples != null && !parentSamples.isEmpty())
+            addParentFields(sampleRow, parentSamples, ExpMaterial.MATERIAL_INPUT_PARENT + "/", user);
+        Set<ExpData> parentDatas = experimentService.getNearestParentDatas(container, user, seed);
+        if (parentDatas != null && !parentDatas.isEmpty())
```
labkey-susanh: It looks like `parentSamples` and `parentDatas` may always be non-null (could annotate `getNearestParent*` as such, so the `!= null` checks might be skipped.

labkey-susanh (approved):
Looks good.  Just a couple of suggestions.

experiment/src/org/labkey/experiment/api/SampleSetUpdateServiceDI.java
XingY: Exception would have been thrown if both are null

experiment/src/org/labkey/experiment/api/SampleSetUpdateServiceDI.java
XingY: Yep. Added NotNull and removed null check.

PR #1308, opened 2020-06-16
Allow App Admin to reset error mark

labkey-tchad:
#### Rationale
Our automated tests should be more flexible about the configuration of the server they wish to test. Many tests currently require database authentication against a server running locally with a full dev enlistment and a test user with site admin permission.
We can add workarounds for many of those requirements but server errors are checked frequently and tests use the `ResetErrorMarkAction` to filter out errors that are out of scope. Application Admins can already view server logs; this change would allow them to set the error mark as well.

Note: The BVT failure here are due to changes in `testAutomation` which are still under development. I'd like to get this merged independently, so that accounts-beta products will have it.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/369

#### Changes
* Allow App Admins to reset error mark


study/test/src/org/labkey/test/tests/study/StudyTest.java
```diff
@@ -137,7 +137,6 @@ protected void doCreateSteps()
     {
         if (!isQuickTest()) // StudyShortTest doesn't test alternate IDs
             SearchAdminAPIHelper.pauseCrawler(getDriver()); //necessary for the alternate ID testing
-        enableEmailRecorder();
```
labkey-jeckels: Is this meant to be included here? It's certainly not related to the title of the PR, but perhaps is relevant for enabling testing on servers without Dumbster?

labkey-tchad: It is not. I readded that line.

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -2405,7 +2405,7 @@ public void addNavTrail(NavTree root)
     }
 
 
-    @AdminConsoleAction
+    @AdminConsoleAction(ApplicationAdminPermission.class)
     @RequiresPermission(AdminPermission.class)
```
labkey-jeckels: Sorry I didn't catch this earlier, but I don't think we want both of these annotations here. Just @AdminConsoleAction(ApplicationAdminPermissions.class) should be enough, I believe. I don't see any other actions double-annotated

labkey-jeckels (changes requested):

labkey-jeckels (approved):

PR #1309, opened 2020-06-16
MethodsAllowed annotation

labkey-matthewb:
#### Rationale
Only allow GET/POST unless action explicitly allows other methods

#### Related Pull Requests
https://github.com/LabKey/cds/pull/278
https://github.com/LabKey/platform/pull/1309
https://github.com/LabKey/recipe/pull/40

### DIFF
--- a/api/src/org/labkey/api/action/BaseApiAction.java
+++ b/api/src/org/labkey/api/action/BaseApiAction.java
@@ -116,14 +116,14 @@ public abstract class BaseApiAction<FORM> extends BaseViewAction<FORM>
     @Override
     public ModelAndView handleRequest() throws Exception
     {
-        switch (getViewContext().getRequest().getMethod())
+        switch (getViewContext().getMethod())
         {
-            case "POST":
-            case "PUT":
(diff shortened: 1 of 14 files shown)


api/src/org/labkey/api/action/BaseApiAction.java
```diff
@@ -116,10 +116,17 @@ protected String getCommandClassMethodName()
     @Override
     public ModelAndView handleRequest() throws Exception
     {
-        if (isPost() || isPut() || isDelete() || isPatch())
-            return handlePost();
-        else
-            return handleGet();
+        switch (getViewContext().getRequest().getMethod())
+        {
+            case "POST":
```
labkey-jeckels: Can we introduce and use constants for these? I imagine future bugs because of inconsistent casing, etc.

labkey-jeckels (approved):
One idea about constants, and I don't love the hijacking of UnauthorizedException for non-401 responses, but I saw your note about not wanting to touch a ton of method signatures

PR #1310, opened 2020-06-17
Issue 40670: Biologics deadlocks

labkey-martyp:
#### Rationale
Issue 40670: Biologics sample import data deadlocks.  This is an issue during automated test setup. The deadlocks are on the clustered index on exp.Data table, which is getting locked by the various queries during the insert process, as well as the indexer updating lastIndexed at the same time. Adding an index was investigated but the benefit of the index did not warrant the cost.  So, instead this PR will batch exp.Data lastIndexed updates until after the import rows is complete then do the updates.

#### Changes
The batch mechanisms were added in ExperimentService where the setDataLastIndexed function was already located.  This is just applied for the exp.Data _importRowsUsingDIB use case.


labkey-matthewb:
Would it make sense to wait until the transaction commits to kick off the indexing tasks?  It's not synchronous anyway.

labkey-jeckels:
Matt's suggestion is more along the lines of what I had been thinking too.

This approach will have thread safety issues - if two different requests are trying to insert into the same table, they'll trip over each other.

I don't know that it's the right spot for this scenario, but something like switching line 438 of FileQueryUpdateService to make the call to addResource() in a commit task instead of immediately

                    getQueryTable().getSchema().getScope().addCommitTask(() -> ss.defaultTask().addResource(resource, SearchService.PRIORITY.item), DbScope.CommitTaskOption.POSTCOMMIT);

PR #1311, opened 2020-06-17
Issue 40571: Guests and Readers see "customize email template" on issue list

labkey-ankurj:
#### Rationale
Only admins should see the 'Customize email template' button in Issues list
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40571


labkey-klum (approved):
code change looks fine, I would add the issue link to the PR description (from the template) so that we can refer back to the underlying issue.

PR #1312, opened 2020-06-17
Don't propagate ContainerFilter when folder is explicitly specified in SQL

labkey-matthewb:
#### Rationale
LabKey SQL has a mechanism for specifying a table in a different folder e.g. "SELECT * FROM Project.Folder.lists.MyList".  However, it does not (yet) have a way to explicitly specify a container filter.  Often applying the query's container filter to these tables does not make sense, however, it usually works because the default container filter is often "null", and the default is used for all created tables.

#### Changes
This change explicitly sets the container filter to null (default) for tables that are specified with an explicit folder path.


labkey-klum (approved):
looks fine to me

PR #1315, opened 2020-06-17
Issue 40669: File browser error "An error occurred trying to load: Status: (401)"

labkey-ankurj:
#### Rationale
logging error instead of displaying a Alert box when a user not having Read permissions access webDav container.
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40669

### DIFF
--- a/filecontent/src/org/labkey/filecontent/FileContentController.java
+++ b/filecontent/src/org/labkey/filecontent/FileContentController.java
@@ -77,6 +77,7 @@ import org.labkey.api.query.QueryUpdateServiceException;
 import org.labkey.api.query.QueryView;
 import org.labkey.api.query.UserSchema;
 import org.labkey.api.query.ValidationError;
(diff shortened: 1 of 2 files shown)


labkey-matthewb:
There's more than this going on here.  The code is executing the API against the wrong container.  It should either a) not make the call if the result is going to be used or b) request against the correct container.  If it helps the WebDav response can be updated with more info (like owning container for each file/folder).

labkey-ankurj:
> There's more than this going on here. The code is executing the API against the wrong container. It should either a) not make the call if the result is going to be used or b) request against the correct container. If it helps the WebDav response can be updated with more info (like owning container for each file/folder).

Agreed, its not getting the correct container in webDav view but it is getting the correct container in FileContent view where ziploader patterns are used for mass spec files. Discussed this with @labkey-jeckels  and committed the change to check around the ReadPermission.

labkey-matthewb (approved):

PR #1316, opened 2020-06-18
Update default sampledata location for server-side tests

labkey-tchad:
#### Rationale
`/sampledata` has been deleted from SVN. Most of it has moved to individual modules or the `testAutomation` repository. I anticipate that there are some sample data that this won't fix but it's a start.

#### Changes
* Get sample data from 'testAutomation' if no module is specified.


labkey-susanh (approved):
Looks good to me

Sigmonia (approved):

PR #1317, opened 2020-06-18
Issue 40659: Security fixes for vulnerabilities found in security scan report

labkey-ankurj:
#### Rationale
This PR addresses the security concerns mentioned in this issue https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=40659

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/209


core/src/org/labkey/core/login/LoginController.java
```diff
@@ -673,6 +673,14 @@ public Object execute(LoginForm form, BindException errors)
                 response = new ApiSimpleResponse();
                 response.put("success", true);
 
+                Cookie emailCookie = PageFlowUtil.getCookie(request, "email");
+
```
labkey-jeckels: As discussed, let's handle this in the JS code that is the only spot that deals with the cookie right now.

server/modules/platform/core/webapp/login.js

and only set it to be secure if the page is an "https://" URL

labkey-jeckels (changes requested):

core/src/org/labkey/core/login/LoginController.java
labkey-ankurj: Done

api/webapp/clientapi/core/Utils.js
```diff
@@ -388,6 +388,10 @@ LABKEY.Utils = new function()
             if (pageonly)
                 path = location.pathname.substring(0, location.pathname.lastIndexOf('/'));
             document.cookie = name + "=" + value + expires + "; path=" + path;
+
+            if (name === "email" && window.location.href.startsWith("https")) {
```
labkey-jeckels: To be extra cautious, do a .toLowerCase() here

core/resources/views/_footer.html
```diff
@@ -1,3 +1,3 @@
 <div style="text-align: center; padding:6px;">
-    <a href="http://labkey.com" target="_blank">Powered by LabKey</a>
+    <a href="http://labkey.com" target="_blank" rel="noopener">Powered by LabKey</a>
```
labkey-jeckels: Let's make this https instead of http

labkey-jeckels: Also, looks like this should be "noopener noreferrer". https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

labkey-jeckels (changes requested):
Please also check for other places in the code where we create links with target="_blank". It should be safe to always add the new attribute, but it's only required when we target some external server (not a link to another page on the same server)  For example, I found one in server/modules/platform/core/src/org/labkey/core/admin/customizeEmail.jsp

We should also make our link rendering inject this in server/modules/platform/api/src/org/labkey/api/data/DataColumn.java, server/modules/platform/api/src/org/labkey/api/data/SimpleDisplayColumn.java, and server/modules/platform/api/src/org/labkey/api/data/DataRegion.java in their renderGridCellContents() methods.

labkey-ankurj:
> Please also check for other places in the code where we create links with target="_blank". It should be safe to always add the new attribute, but it's only required when we target some external server (not a link to another page on the same server) For example, I found one in server/modules/platform/core/src/org/labkey/core/admin/customizeEmail.jsp
> 
> We should also make our link rendering inject this in server/modules/platform/api/src/org/labkey/api/data/DataColumn.java, server/modules/platform/api/src/org/labkey/api/data/SimpleDisplayColumn.java, and server/modules/platform/api/src/org/labkey/api/data/DataRegion.java in their renderGridCellContents() methods.

Aren't the urls in renderGridCellContents of these classes targeting the same server?

labkey-jeckels:
99% of the time, yes, but they can be external URLs too, like http://foo.com/path?id=${RowId}

api/webapp/clientapi/core/Utils.js
```diff
@@ -389,7 +389,7 @@ LABKEY.Utils = new function()
                 path = location.pathname.substring(0, location.pathname.lastIndexOf('/'));
             document.cookie = name + "=" + value + expires + "; path=" + path;
 
-            if (name === "email" && window.location.href.startsWith("https")) {
+            if (name.toLowerCase() === "email" && window.location.href.startsWith("https")) {
```
labkey-jeckels: I was intending for the lower casing to be on window.location.href, not the cookie name

api/src/org/labkey/api/data/DataColumn.java
```diff
@@ -409,8 +409,9 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
                     out.write("\" style=\"");
                     out.write(css);
                 }
-                
-                out.write("\">");
+
+                out.write("\" rel=\"noopener noreferrer\"");
```
labkey-jeckels: One more tweak - let's only do this when we have a target window

api/src/org/labkey/api/data/SimpleDisplayColumn.java
```diff
@@ -144,7 +144,8 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
                 out.write(linkCls);
             }
 
-            out.write("'>");
+            out.write("' rel='noopener noreferrer'");
```
labkey-jeckels: Same here -  let's only do this when we have a target window

labkey-jeckels (approved):
Three more tiny changes, but I'll make it as approved

PR #1318, opened 2020-06-18
Issue 40415: Server JavaScript Console not displaying trigger messages fired by ETL

labkey-martyp:
#### Rationale
Update SessionAppender to store appender infos for cross thread access.  This is to allow trigger script console messages triggered from an ETL pipeline job to use the same appender info as the current view context.

#### Related Pull Requests
https://github.com/LabKey/dataintegration/pull/46

#### Changes
Add register mechanism to save appender infos to concurrent map.  Add functions to get appender info by key and to set the current thread's appender info to an already existing appender info.

### DIFF
--- a/api/src/org/labkey/api/util/SessionAppender.java
+++ b/api/src/org/labkey/api/util/SessionAppender.java
@@ -16,12 +16,15 @@
 package org.labkey.api.util;
 
 import org.apache.log4j.spi.LoggingEvent;
+import org.jetbrains.annotations.NotNull;
+import org.labkey.api.collections.ConcurrentCaseInsensitiveSortedMap;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpSession;
 import java.io.Serializable;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 
 /**
  * User: matthewb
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/util/SessionAppender.java
```diff
@@ -45,7 +48,25 @@
     }
 
     private static final ThreadLocal<AppenderInfo> localInfo = new ThreadLocal<>();
+    private static final Map<String, AppenderInfo> appenderInfos = new ConcurrentCaseInsensitiveSortedMap<>();
```
labkey-matthewb: Roaches check in, but they don't check out!

This needs to use some sort of WeakHashMap, or a SessionListener to remove AppendInfo that are associated with sessions that have gone away.

labkey-matthewb (changes requested):

api/src/org/labkey/api/util/SessionAppender.java
labkey-martyp: Dang, yes forgot the cleanup on this.  I went with a ConcurrentReferenceMap which is essentially a concurrent WeakHashMap.

api/src/org/labkey/api/util/SessionAppender.java
```diff
@@ -48,16 +48,16 @@
     }
 
     private static final ThreadLocal<AppenderInfo> localInfo = new ThreadLocal<>();
-    private static final Map<String, AppenderInfo> appenderInfos = new ConcurrentCaseInsensitiveSortedMap<>();
+    private static final Map<String, AppenderInfo> appenderInfos = new ConcurrentReferenceHashMap<>();
```
labkey-matthewb: What a great class.  You should probably specify ReferenceType.WEAK, instead of the default ReferenceType.SOFT.

labkey-matthewb (changes requested):

api/src/org/labkey/api/util/SessionAppender.java
labkey-matthewb: Also, probably worth a comment explaining how/when the cleanup should work.  e.g. References to the SessionAppender should go away when the session is destroyed allowing the ConcurrentReferenceHashMap to release the entry.

labkey-martyp: Oh yeah, I missed that was a soft reference by default. Specified to WEAK reference and added a comment about garbage collection on it.

labkey-matthewb (approved):

PR #1319, opened 2020-06-18
containerfilter annotation

labkey-matthewb:
#### Rationale
In LabKey SQL sometimes the single view-based ContainerFilter does not make sense, and the developer may want to explicitly set a ContainerFilter for a particular table.

https://docs.google.com/document/d/1A87P9Mjznv0UVQjIPHdQvCFX9zbohJeOX6IjCeM7qlI/edit#


query/src/org/labkey/query/sql/SqlBase.g
```diff
@@ -370,15 +370,34 @@ joinExpression
...
-// Usually a simple dotted identifer 'path' such as "core.users".
+tableSpecificationWithAnnotation
+    :  table=tableSpecification (tableAnnotations!)? { ((SupportsAnnotations)table.getTree()).setAnnotations(getAnnotations()); }
+    ;
+
+
+// multiple patterns for playing around with ContainerFilter annotation
+// TODO pick one
+tableAnnotations
+    // Issues[ContainerFilter='CurrentAndSubfolders'] R
```
labkey-jeckels, line 389: This one gets my vote

query/src/org/labkey/query/sql/QuerySelect.java
```diff
@@ -654,7 +655,32 @@ private QTable parseRange(QNode node)
...
+            ContainerFilter.Type cfType = null;
+            Map<String,Object> annotations = ((QUnknownNode)node).getAnnotations();
+            if (null != annotations)
+            {
+                for (var entry : annotations.entrySet())
+                {
+                    var value = entry.getValue();
+                    switch (entry.getKey())
+                    {
+                        case "containerfilter":
```
labkey-jeckels, line 668: There must be something lower-casing the keys in this map? Might be a bit cleaner as an equalsIgnoreCase()

query/src/org/labkey/query/sql/Query.java
```diff
@@ -844,20 +841,24 @@ private QueryRelation _resolveTable(
...
+        Pair<QuerySchema, TableInfo> found = _resolveCache.get(currentSchema).get(cacheKey);
         if (null != found)
         {
             TableInfo ti = found.second;
-            if (null != ti.getContainerFilter() && ti.getContainerFilter().equals(getContainerFilter()))
-                return new QueryTable(this, found.first, found.second, alias);
+            if (null != ti.getContainerFilter())
+            {
+                ContainerFilter cf = ti.getContainerFilter();
+                if (null == cfType ? cf.equals(getContainerFilter()) : cf.getType() == cfType)
```
labkey-jeckels: I'm not really following this. Maybe a comment about why it's being compared?

labkey-jeckels (approved):
Some suggestions, but nothing big. Should this go onto a backlog to help get the test coverage in place, or were you planning to do that?

labkey-jeckels (approved):

PR #1328, opened 2020-06-22
be specific about which audit event we're testing

labkey-chrisj:
#### Rationale
This updates the secondary auth test, which wrongly assumes that the first row to be returned in a request to the auth-config audit log will be the expected one.  Because the new auth config model produces more events, the test now asks more specifically for the one it needs to measure 

#### Related Pull Requests
n/a

### DIFF
--- a/devtools/test/src/org/labkey/test/tests/devtools/SecondaryAuthenticationTest.java
+++ b/devtools/test/src/org/labkey/test/tests/devtools/SecondaryAuthenticationTest.java
@@ -23,6 +23,7 @@ import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.labkey.remoteapi.CommandException;
 import org.labkey.remoteapi.Connection;
+import org.labkey.remoteapi.query.Filter;
 import org.labkey.remoteapi.query.SelectRowsCommand;
 import org.labkey.remoteapi.query.SelectRowsResponse;
 import org.labkey.remoteapi.query.Sort;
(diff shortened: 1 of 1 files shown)


devtools/test/src/org/labkey/test/tests/devtools/SecondaryAuthenticationTest.java
```diff
@@ -190,7 +202,12 @@ protected SelectRowsResponse getLatestAuditEntries()
         Connection cn = createDefaultConnection(true);
         SelectRowsCommand selectCmd = new SelectRowsCommand("auditLog", "AuthenticationProviderConfiguration");
         selectCmd.setSorts(Arrays.asList(new Sort("Created", Sort.Direction.DESCENDING)));
-        selectCmd.setMaxRows(1);
+
+        DateFormat dateFormat = new SimpleDateFormat("yyyy/MM/dd");
+        Date date = new Date();
+
+        selectCmd.setFilters(Arrays.asList(new Filter("Date", dateFormat.format(date), Filter.Operator.DATE_GTE)));
+        //selectCmd.setMaxRows(1);
```
RosalineP: Should this comment be removed?

RosalineP (approved):
Looks good, just one note on that comment, and it'd be nice to see what TC reports before merge!

devtools/test/src/org/labkey/test/tests/devtools/SecondaryAuthenticationTest.java
labkey-chrisj: It should, yes.  Thanks

PR #1330, opened 2020-06-23
Epic 5886 : ODBC for hosted clients

labkey-bpatel:
#### Rationale
Part of offering ODBC connectivity to our hosted customers is to give end-user access to the configured certificate.

#### Related Pull Requests
https://github.com/LabKey/connectors/pull/34

#### Changes
- Add External Tool View Service and Provider infrastructure
- Decouple APIKey view from SecurityController.ApiKeyAction to APIKeyProvider.
- Refactor ApiKeyAction to iterate through External Tool Providers.
- Decouple user navigation piece ('Done' button) from the individual view rendering.


core/src/org/labkey/core/security/SecurityController.java
```diff
@@ -1928,17 +1930,29 @@ public void setHideCaption(String hideCaption)
...
+            VBox view = new VBox();
+            for (ExternalToolsViewProvider externalAccessViewProvider : ExternalToolsViewService.get().getExternalAccessViewProviders())
+            {
+                for (ModelAndView providerView : externalAccessViewProvider.getViews(getUser()))
+                {
+                    view.addView(providerView);
+                }
+            }
+
+            //TODO: this is no longer shown, should this message be part of the view or silently just not display because user does not have the access?
```
labkey-jeckels: Let's replace with a "Nothing enabled" page if somebody gets here but no providers are active for them

labkey-jeckels (approved):
Let me know if you want further review, but otherwise one simple comment and don't forget the TODOs about JavaDocs

core/src/org/labkey/core/security/SecurityController.java
labkey-bpatel: Done!

core/src/org/labkey/core/security/SecurityController.java
```diff
@@ -1939,9 +1939,10 @@ public ModelAndView getView(ReturnUrlForm form, BindException errors)
                 }
             }
 
-            //TODO: this is no longer shown, should this message be part of the view or silently just not display because user does not have the access?
-//            if (!PopupUserView.allowApiKeyPage(getUser()))
-//                throw new UnauthorizedException("API keys are not configured on this site. Contact a site administrator.");
+            if (view.isEmpty())
```
labkey-jeckels: We should rename this action - it's no longer just for ApiKeys

core/src/org/labkey/core/security/nothingEnabled.jsp
```diff
@@ -0,0 +1,20 @@
...
+     * Unless required by applicable law or agreed to in writing, software
+     * distributed under the License is distributed on an "AS IS" BASIS,
+     * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     * See the License for the specific language governing permissions and
+     * limitations under the License.
+     */
+%>
+<%@ page extends="org.labkey.api.jsp.JspBase" %>
+<%@ taglib prefix="labkey" uri="http://www.labkey.org/taglib" %>
+Nothing enabled
```
labkey-jeckels: Sorry, I didn't literally mean to just show "nothing enabled". Let's rephrase to be a bit more user-friendly. How about:

This server does not have any configurations enabled that show settings here. An administrator can configure tool access via the Admin Console.

api/src/org/labkey/api/view/PopupUserView.java
```diff
@@ -84,4 +85,9 @@ public static boolean allowApiKeyPage(User user)
     {
         return AppProps.getInstance().isAllowApiKeys() || AppProps.getInstance().isAllowSessionKeys() || user.hasSiteAdminPermission();
     }
+
+    public static boolean allowExternalToolSettingsPage(User user)
+    {
+        return getContextContainer().hasPermission(user, ReadPermission.class);
```
labkey-jeckels: Is this quite right? Seems like the current user will always have read permission on the current container if the page is being rendered, and if ODBC is off and API keys aren't enabled, they'll get the page that tells them nothing is available.

Seems like we should check to see if any views would be shown to the user, and if not, omit the menu item.

labkey-jeckels (changes requested):
Another set of suggestions, hopefully the last on this side of the PR

core/src/org/labkey/core/security/nothingEnabled.jsp
labkey-bpatel: Haha yes, sorry i meant to ask you if we needed anything more user-friendly. Thanks for the text.

core/src/org/labkey/core/security/SecurityController.java
labkey-bpatel: Yes, good point! Action now renamed.

api/src/org/labkey/api/view/PopupUserView.java
labkey-bpatel: Since all the readers get odbc view at minimum, not too clear what the check should be to omit the menu item altogether, (and if odbc is disabled then its just not configured and so will get nothingEnabled page). So not clear on what the criteria is to for a user not to get external tool provider views.

labkey-jeckels: Not all deployments will have the Connectors module, and API keys aren't enabled by default.

There's no harm in showing them a blank page, of course, but it's a useless bit of menu clutter in those cases.

labkey-bpatel: Done, let me know what you think!

core/src/org/labkey/core/security/SecurityController.java
```diff
@@ -269,12 +271,17 @@ public ActionURL getFolderAccessURL(Container container)
         }
 
         @Override
-        public ActionURL getApiKeyURL(@NotNull URLHelper returnURL)
+        public ActionURL getExternalToolsViewURL(User user, @NotNull URLHelper returnURL)
```
labkey-jeckels, line 275: Let's annotate this as @Nullable, and consider this set!

labkey-jeckels (approved):
One more tiny tweak, but otherwise ready to merge so I'm approving the PR

PR #1331, opened 2020-06-23
Mechanism to register items on the Developer menu

labkey-adam:
#### Rationale
Modules need a way to register items on the Developer menu so that API doesn't need knowledge of a host of premium services.

#### Related Pull Requests
* https://github.com/LabKey/rstudio/pull/64

#### Changes
* Add PopupDeveloperView.registerMenuProvider()
* Register provider for standard menu items

### DIFF
--- a/api/src/org/labkey/api/view/NavTrees.java
+++ b/api/src/org/labkey/api/view/NavTrees.java
@@ -1,0 +1,18 @@
+package org.labkey.api.view;
+
+import java.util.Collection;
+
+public class NavTrees
+{
+    private final Collection<NavTree> _navTrees;
+
+    public NavTrees(Collection<NavTree> navTrees)
+    {
+        _navTrees = navTrees;
+    }
+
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/view/NavTrees.java
```diff
@@ -0,0 +1,18 @@
+package org.labkey.api.view;
+
+import java.util.Collection;
+
+public class NavTrees
```
labkey-matthewb, line 5: Is this really adding any value?

labkey-matthewb (approved):

api/src/org/labkey/api/view/NavTrees.java
labkey-adam: You could argue it, but wrapping keeps the providers honest... they shouldn't be getting or removing items from the collection, only adding. Same pattern we follow with ClientDependencies and other similar provider scenarios.

PR #1333, opened 2020-06-23
Recognize missing from clause entry error instead of internal error

labkey-matthewb:
#### Rationale
This SQL below is illegal because the first C.Name references a table that is not yet in scope.  We should catch this instead of throwing an internal parse exception.  Note, we still catch this too late, but still this provides a better error message (similar to postgres)

```
SELECT A.Name
FROM 
  Modules A 
     FULL JOIN 
  Modules B ON B.Name=C.Name
     FULL JOIN
  Modules C ON A.Name=C.Name
```


labkey-adam (changes requested):
Can we add a junit test that verifies we get the preferred error in this case?

labkey-adam (approved):

PR #1334, opened 2020-06-23
Issue 40659: correct jQuery references

labkey-ankurj:
#### Rationale
Removes the jQuery from timeline and refers from internal

#### Related Pull Requests
https://github.com/LabKey/commonAssays/pull/194
https://github.com/LabKey/targetedms/pull/213


labkey-ankurj:
Verified this by Running TimelineTest, it is passing locally. Let me know if more testing could be done to verify.

labkey-jeckels (approved):
Please double-check that the TimelineTest is still happy after this change, but I hope we're happy and don't have duplicates checked in now

labkey-susanh:
We probably are doing this in other places for client-side dependencies, but we should try to minimize the referencing of files from other modules.  It makes it not possible to build modules independently and we're seeing more people want to do that. Here there's really no ill effects since internal and timeline are both in the same git repo so you would have to try very hard to have the source for timeline without having the source for internal, but if we're reaching over to internal from non-platform modules we should try to find a different solution.

labkey-jeckels:
@labkey-susanh I was giving Ankur the opposite guidance. Aren't we guaranteed that API and Internal will be present on all deployments? We don't ship extra copies of ExtJS in non-platform modules - can we not take a similar approach? Or should we promote jQuery from Internal to API?

labkey-susanh:
> @labkey-susanh I was giving Ankur the opposite guidance. Aren't we guaranteed that API and Internal will be present on all deployments? We don't ship extra copies of ExtJS in non-platform modules - can we not take a similar approach? Or should we promote jQuery from Internal to API?

Yes, we are guaranteed they will be in all deployments (server won't run without them), so if this is a change for run-time use then we should be ok.  It's only if it's necessary for the build to happen that we might have a problem.  But, yes, treating it like we do ExtJS should be find.

labkey-ankurj:
> Please double-check that the TimelineTest is still happy after this change, but I hope we're happy and don't have duplicates checked in now

Timeline test is still happy

PR #1335, opened 2020-06-23
issue37736

labkey-chrisj:
#### Rationale
This adds a very quick test to address issue#37736, which provides coverage to ensure that the create project wizard highlights steps along the way

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/377


core/test/src/org/labkey/test/tests/ProjectWizardTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.experimental.categories.Category;
+import org.labkey.test.BaseWebDriverTest;
+import org.labkey.test.TestTimeoutException;
+import org.labkey.test.categories.Git;
+import org.labkey.test.components.BodyWebPart;
+
+import java.util.Arrays;
+import java.util.List;
+
+@Category({Git.class})
```
labkey-tchad: This should go in the community `Daily` suite. The `Git` suite is for premium functionality

core/test/src/org/labkey/test/tests/ProjectWizardTest.java
```diff
@@ -0,0 +1,79 @@
...
+                .clickNext()
+                .setMyUserOnly()
+                .clickNext()
+                .useDefaultLocation()
+                .clickFinish();
+
+        // confirm expected webparts in a collaboration project are present
+        new BodyWebPart(getDriver(), "Subfolders");
+        new BodyWebPart(getDriver(), "Wiki");
+        new BodyWebPart(getDriver(), "Messages");
```
labkey-tchad: You only get these webparts if the `wiki` and `announcements` modules are installed. We should avoid imposing unnecessary requirements to allow testing as wide a variety of servers as possible. Sample Manager trial doesn't have them for instance.

core/test/src/org/labkey/test/tests/ProjectWizardTest.java
```diff
@@ -0,0 +1,79 @@
...
+                .setFolderType("Collaboration")
+                .setProjectName(getProjectName())
+                .clickNext()
+                .setMyUserOnly()
+                .clickNext()
+                .useDefaultLocation()
+                .clickFinish();
+
+        // confirm expected webparts in a collaboration project are present
+        new BodyWebPart(getDriver(), "Subfolders");
```
labkey-tchad: Don't assume that the `BodyWebPart` constructor will actually find the webpart; it might get refactored to not do so at some point. Just grab the actual titles and verify that they contain what you want.
```suggestion
        assertThat("WebParts for collaboration folder type.", new PortalHelper().getWebPartTitles(), hasItem("Subfolders"))
```
_Sidenote: the thought of this is the sort of pattern is what makes me worry about helpers that return `null` or `-1` when something isn't found._

core/test/src/org/labkey/test/tests/ProjectWizardTest.java
```diff
@@ -0,0 +1,79 @@
...
+    {
+        projectMenu()
+                .navigateToCreateProjectPage()
+                .setFolderType("Collaboration")
+                .setProjectName(getProjectName())
+                .clickNext()
+                .setMyUserOnly()
+                .clickNext()
+                .useDefaultLocation()
+                .clickFinish();
```
labkey-tchad, line 77: This doesn't cover the regression described in the [related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=37736). The problem was cosmetic. This should go step by step and check that the correct step is highlighted and has the correct text.

labkey-tchad (changes requested):
This doesn't cover the regression that spawned this test.

core/test/src/org/labkey/test/tests/ProjectWizardTest.java
labkey-jeckels: @labkey-tchad  This is out of scope for this PR, but sounds like we should rename the Git suite sometime soon...

core/test/src/org/labkey/test/tests/ProjectWizardTest.java
labkey-chrisj: Huh.  Didn't realize getWebPartTitles() existed, thanks for pointing it out

core/test/src/org/labkey/test/tests/ProjectWizardTest.java
labkey-chrisj: I was relying on the waitForPage methods in SetFolderPermissionsPage and SetInitialFolderSettingsPage to check that- and the former does- but only for synchronization purposes (it's been timing out without raising errors for a long time now).
I'm updating the pages with waitForPage() methods that will verify the active list item along the way, and raise errors if they aren't there (or more to the point, they won't sit there waiting to time out without treating that like a problem)

core/test/src/org/labkey/test/tests/ProjectWizardTest.java
labkey-chrisj: Okay, I just put it in platform because that's where the fix went.

core/test/src/org/labkey/test/tests/ProjectWizardTest.java
labkey-chrisj: If you'd prefer, I can rip those checks out of the pages and put them in the test.

labkey-tchad: I vote to rip them out of the pages and put them in the test. There's no reason to have dozens of tests start failing if this regresses again.
I thought I'd seen tests pausing a bit too long on one of the wizard steps. Good to know that will be cleared up with this as well.

labkey-tchad (approved):

core/test/src/org/labkey/test/tests/ProjectWizardTest.java
labkey-chrisj: Okay, that's how this will go, then.
Pages will sync on the wizard step highlighting
if this regresses, the test will tell us, but the pages will not enforce anything.
Eventually, as most tests come to use the APIs to create projects, this test can be expanded to test the wizard properly

PR #1336, opened 2020-06-23
Issue 40670: Biologics SQL Server deadlocks on sample data import

labkey-martyp:
#### Rationale
This is an issue during automated test setup. The deadlocks are on the clustered index on exp.Data table, which is getting locked by the data upload thread and two search service threads. The search service is querying exp.Data in the SearchService:runner thread and querying and inserting into exp.Data in the SearchService:index thread, while the main thread is importing data into exp.Data.  All in rapid and unpredictable order during sample data import, causing frequent deadlocks in sql server.  The runnable being queued up for the SearchService:runner is actually not needed as it is just queuing the imported exp.Data resources for SearchService:index.  So this PR removes queuing the runnable and just calls it in a commit task.  This essentially eliminates any queries into exp.Data by the SearchService:runner thread and stops these three thread deadlocks.

#### Changes
Do not add the index task as a runnable in search service.  Call directly or use in commitTask.


labkey-jeckels (approved):
I suggest simplifying to just call DbScope.getLabKeyScope().addCommitTask(). It'll do the right thing, with or without an active transaction

PR #1343, opened 2020-06-25
Issue 40510: promote inputs to the run

labkey-ankurj:
#### Rationale
Experiment run can be created by provenance recording apis with inputs and outputs attached to the recording steps. This work promotes the inputs of a step that are not attached to the run and are not outputs of the previous step.

#### Related Pull Requests
* https://github.com/LabKey/provenance/pull/32


experiment/src/org/labkey/experiment/pipeline/ExpGeneratorHelper.java
```diff
@@ -460,11 +462,72 @@ else if (action.isEnd())
...
+            actionsList.get(i).getMaterialInputs().forEach(materialLsid -> {
+                if (!runMaterialInputs.contains(materialLsid) && !prevAction.getMaterialOutputs().contains(materialLsid))
+                {
+                    // promote material input to run
+                    addMaterialInput(run, run.getInputProtocolApplication(), materialLsid, user);
+                }
+            });
+
+            actionsList.get(i).getInputs().forEach(dataFile -> {
+                if (!runDataInputs.contains(dataFile.getURI().toString()) && !prevAction.getInputs().contains(dataFile))
```
labkey-jeckels: This doesn't feel right. Isn't it a mismatch with LSID and URI?

labkey-jeckels (changes requested):

experiment/src/org/labkey/experiment/pipeline/ExpGeneratorHelper.java
labkey-ankurj: yes, fixed it.

experiment/src/org/labkey/experiment/pipeline/ExpGeneratorHelper.java
```diff
@@ -460,11 +462,85 @@ else if (action.isEnd())
...
+                .getDataInputs()
+                .keySet()
+                .stream()
+                .map(Identifiable::getLSID)
+                .collect(Collectors.toSet());
+
+        String lsid = "lsid";
+        String datafileurl = "datafileurl";
+
+        Collection<Map<String, Object>> containerData = new TableSelector(ExperimentService.get().getTinfoData(), Set.of(lsid, datafileurl)).getMapCollection();
```
labkey-jeckels: This will pull in ALL exp.data rows, correct? On some servers that's hundreds of thousands or maybe even millions. And not all will have a DataFileURL, which means they would NPE in the forEach below. Can we filter to those in this container? Or that have a URL or LSID that matches the ones we're processing in this run?

labkey-jeckels (changes requested):

experiment/src/org/labkey/experiment/pipeline/ExpGeneratorHelper.java
labkey-ankurj: would this filter work?
```
SimpleFilter filter = SimpleFilter.createContainerFilter(container);
filter.addCondition(FieldKey.fromParts(datafileurl), null, CompareType.NONBLANK);
```

labkey-ankurj: Or that have a URL or LSID that matches the ones we're processing in this run? -- to create this filter is it okay to query multiple times?

labkey-jeckels: Your filter would definitely help, and I think still match the ones you need.

But I think it would be safe to do a series of queries for the individual rows that you need in this case. There should be a modest number of data or material objects attached to a given run, so it shouldn't be a real perf concern, and eliminates the concern about potentially pulling tons of rows into memory.

experiment/src/org/labkey/experiment/pipeline/ExpGeneratorHelper.java
```diff
@@ -520,22 +516,28 @@ static private void promoteInputs(Set<RecordedAction> actions, ExpRun run, Map<U
             });
 
             actionsList.get(i).getInputs().forEach(dataFile -> {
-                String dataLsid = containerDataLsids.get(dataFile.getURI().toString());
-                if (!runDataInputs.contains(dataLsid) && !prevAction.getOutputs().contains(dataFile))
-                {
-                    // promote data input to run
-                    ExpData data = null;
+                filter.addCondition(FieldKey.fromString(datafileurl), dataFile.getURI().toString(), CompareType.EQUAL);
```
labkey-jeckels, line 517: It looks like you're accumulating filters here. The first time through the loop, it'll match, but all subsequent times won't match because you'll have DataFileURL = X AND DataFileURL = Y, which will never match any rows. If you push the filter creation inside the loop, it should be OK. And LSID should never be null, so that filter isn't strictly necessary.

labkey-jeckels (changes requested):

experiment/src/org/labkey/experiment/pipeline/ExpGeneratorHelper.java
labkey-ankurj: I think the latest commit is fixed
Screen Shot 2020-06-29 at 9 30 19 AM

labkey-jeckels: Step 3 isn't supposed to have any outputs? Probably OK, just not sure about the test case.

labkey-jeckels (approved):
One question to double-check

labkey-ankurj:
> One question to double-check

Yes, I did not add outputs in step 3.

PR #1344, opened 2020-06-25
Allow dataset insert view form field default values to come from URL parameters

cnathe:
#### Rationale
A new client would like to be able to customize / generate a URL from the study participant page so that a user can insert a new row in a dataset and pull through the ParticipantId as the default / initial value in the dataset insert view form. We currently allow for "default values" to be set on study dataset properties, but that doesn't fit this scenario since they would like to be able to set that participant id based on the referring page.

With this PR, I have added code that will use the InsertView setInitialValues mechanism used for domain default values to also look for URL parameters as a way to set default values into the insert view form. Those URL param default values apply as follows:
* only for the insert case initial page load (i.e. not respected on error reshow)
* URL default values will be applied after the domain properties default values are set
* the URL param key must exactly match the insert view form field key

Here is an example URL that would populate the ParticipantId and Date fields of the dataset insert form based on URL params:
`/dataset-insert.view?datasetId=5010&quf_ParticipantId=999320016&quf_Date=2020-06-25`

#### Changes
* InsertUpdateAction.java update in getView() to look for URL parameters that match the insert view form field keys to populate initial values


cnathe:
@labkey-matthewb @kevinkrouse @labkey-klum I created this draft PR as a way for you guys to review this idea I was playing with as a way to allow for a client to be able to set default values in a dataset "Insert New Row" form's input fields. Let me know if you guys know of other existing mechanisms that I should explore for this or if you think there are issues with this approach. Thanks.

labkey-matthewb:
I think this is a very useful feature.  I don't particularly like the "quf_", but it is useful to have a prefix that is tied to this feature, and does not look like the filter syntax.  Westide used ".", maybe "_" by itself would work, or we could be really explicit like "default_".

cnathe:
> I think this is a very useful feature. I don't particularly like the "quf_", but it is useful to have a prefix that is tied to this feature, and does not look like the filter syntax. Westide used ".", maybe "_" by itself would work, or we could be really explicit like "default_".

@labkey-matthewb I am only using the "quf_" here because then we get the form value binding for free since that is the prefix used by these dataset insert forms. I could see using a different prefix, but that would mean a bit more code for handling this case. I can make that change if you think it is necessary.

labkey-jeckels:
I'd vote for a supporting a different prefix, since it's something we'll be telling others to use in their URLs and trying to maintain backwards compatibility with.

cnathe:
> I'd vote for a supporting a different prefix, since it's something we'll be telling others to use in their URLs and trying to maintain backwards compatibility with.

Sounds good. Two votes for a specific URL param prefix works for me. I've updated this PR to use "default." as the URL param prefix for this scenario.

cnathe:
@labkey-adam @labkey-matthewb @labkey-jeckels This PR is ready for re-review, with the change made to use a URL param prefix (i.e. "default.ParticipantId", "default.Date", etc.).

labkey-jeckels (approved):
Looks good to me.

PR #1348, opened 2020-06-27
Make participant list height dynamic

RosalineP:
#### Rationale
It was requested that the participant list right-hand display panel dynamically grow to fit its result contents, rather than a hard 470px.

#### Related Pull Requests
* N/A

#### Changes
* Remove height limitation of 470px


cnathe (commented):
I don't know the full context of this change, but just an FYI that this change affects both the narrow "Participant List" webpart and the left/main "Participant List" webpart. Also note that if a study has a large number of subjects (say > 1000) this webpart will now grow vertically to fit ALL subjects on the screen. This could result in a very very tall page with a lot of white space.

RosalineP:
Good point on the fact this also applies to the webpart! I sent a message with more context on this issue to hopefully work something out.

study/src/org/labkey/study/view/subjects.jsp
```diff
@@ -86,6 +86,9 @@
     String groupsPanelId  = "groupsPanel" + getRequestScopedUID();
 
     String viewObject = "subjectHandler" + bean.getIndex();
+    String param = getViewContext().getRequest().getParameter("pageId");
+    Boolean isParticipantsListPage = param != null && param.equals("study.PARTICIPANTS");
```
cnathe: I recommend just using the constant from StudyFolderTabs.ParticipantsPage:
```suggestion
    boolean isParticipantsListPage = StudyFolderTabs.ParticipantsPage.PAGE_ID.equalsIgnoreCase(param);
```

study/src/org/labkey/study/view/subjects.jsp
```diff
@@ -744,7 +744,7 @@ Ext4.onReady(<%=viewObject%>.render, <%=viewObject%>);
                     <%= !bean.getWide() ? "<br/>" : "" %>
                     <span id="<%=divId%>.status" name="status" style="margin-bottom: 10px;">Loading...</span>
                 </div>
-                <div style="overflow-y:auto; height: 470px;" id="<%= listDivId %>"></div>
+                <div style="<%=conditionalStyling%>" id="<%= listDivId %>"></div>
```
cnathe: I'd imagine that IntelliJ is complaining about this being "unsafe" (i.e. red squiggles):
```suggestion
                <div style="<%=h(conditionalStyling)%>" id="<%= listDivId %>"></div>
```

cnathe (changes requested):

study/src/org/labkey/study/view/subjects.jsp
RosalineP: Oddly not... 
image
Thank you for the reminder, though!

cnathe (approved):

labkey-adam (approved):

PR #1349, opened 2020-06-29
Item 7465: Add sample timeline audit events when inventory items change

labkey-susanh:
#### Rationale
When inventory items change, we want to track that in the sample timeline events.  We also want to be able to distinguish these events from other types of audit events so we can display them specially within applications

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/310
* https://github.com/LabKey/inventory/pull/46

#### Changes
* Update SampleTimelineAuditEvent to add an inventoryUpdateType field
* Add action for creating an auditEvent with a InventoryUpdateType


cnathe (commented):
@labkey-susanh the SampleInventoryUpdateType.java class seems so "inventory" specific. Have you thought about moving that over to the inventory module and then just having the new SampleTypeService.addAuditEvent() method take in that type as a string (and the message as the comment)?

labkey-susanh:
> @labkey-susanh the SampleInventoryUpdateType.java class seems so "inventory" specific. Have you thought about moving that over to the inventory module and then just having the new SampleTypeService.addAuditEvent() method take in that type as a string (and the message as the comment)?

Yes, that's where I started, but then I moved it here because it also seemed sample-specific (since nothing else uses this enum currently).   I could see it either way, but we may want this type also for more specific audit logs for the InventoryAuditEvents, so, yeah, I guess I'll move it back.

api/src/org/labkey/api/exp/api/SampleInventoryUpdateType.java
```diff
@@ -0,0 +1,40 @@
+package org.labkey.api.exp.api;
+
+import java.util.Map;
+
+public enum SampleInventoryUpdateType
```
cnathe: discussed moving this back to the inventory module so that it can use this type and just pass over the type string to the addAuditEvent

cnathe (changes requested):

cnathe (approved):

PR #1355, opened 2020-06-30
Fb validator reference

labkey-matthewb:
#### Rationale
Remove unnecessary table exp.validatorreference.  This m-m junction table has caused deadlocks and is unnecessary.  Also be consistent about always providing WHERE clauses to use the PK (container,propertyid,rowid)


experiment/resources/schemas/dbscripts/sqlserver/exp-20.006-20.007.sql
```diff
@@ -0,0 +1,26 @@
+EXEC sp_rename 'exp.propertyvalidator', 'pv_old'
+GO
+
+CREATE TABLE exp.PropertyValidator
+(
+    RowId        int identity(1,1) not null,
+    Name         varchar(50) not null,
+    Description  varchar(200),
+    TypeURI      varchar(200) not null,
+    Expression   text,
```
labkey-jeckels: I think we should be using NVARCHAR, and NVARCHAR(MAX) instead of TEXT

experiment/resources/schemas/dbscripts/sqlserver/exp-20.006-20.007.sql
```diff
@@ -0,0 +1,26 @@
...
+    ErrorMessage text,
+    Properties   text,
+    Container    entityid not null constraint fk_pv_container references core.containers (entityid),
+    PropertyId   int not null constraint fk_pv_descriptor references exp.propertydescriptor,
```
labkey-jeckels: Indices on Container and PropertyId?

labkey-jeckels (changes requested):
Couple of schema nits

experiment/resources/schemas/dbscripts/sqlserver/exp-20.006-20.007.sql
labkey-matthewb: This is as it was previously. I was wondering if we really need the propertyid constraint, or should add an index.  container should be fine because of the PK.

labkey-jeckels: My heuristic is that if it needs a FK, it probably needs an index too, but these will be small tables so no big deal. I think the NVARCHAR comment is relevant regardless (maybe you're already swapping that one)

labkey-jeckels (approved):

PR #1361, opened 2020-07-02
Fix #36517, #40790, and #40799

labkey-adam:
#### Rationale
The Upload Participant Mapping page is not clear that an error during import (e.g., unrecognized participant ID) means nothing is imported. In the process of fixing that issue, I discovered that exporting the mapping can result in a MUTATING SQL exception and that Container is far too strict when it attempts to set a container type that it doesn't recognize.

#### Changes
* Fix [Issue 36517: Alternate participant ID import fails at job level](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=36517)
* Fix [Issue 40790: java.lang.IllegalStateException: MUTATING SQL executed as part of handling action: GET org.labkey.study.controllers.StudyController$ExportParticipantTransformsAction](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40790)
* Fix [Issue 40799: java.lang.IllegalArgumentException: Unknown container type: labbook](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40799)

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -94,6 +94,8 @@ import java.util.function.BooleanSupplier;
  */
 public class Container implements Serializable, Comparable<Container>, SecurableResource, ContainerContext, HasPermission, Parameter.JdbcParameterValue
 {
+    private static final Logger LOG = Logger.getLogger(Container.class);
+
     private GUID _id;
     private Path _path;
     private Date _created;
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/data/Container.java
```diff
@@ -1322,9 +1322,11 @@ public String getType()
 
     public void setType(String typeString)
     {
-        _containerType = ContainerTypeRegistry.get().getType(typeString);
-        if (_containerType == null)
-            throw new IllegalArgumentException("Unknown container type: " + typeString);
+        ContainerType type = ContainerTypeRegistry.get().getType(typeString);
+        if (type == null)
+            Logger.getLogger(Container.class).warn("Unknown container type: " + typeString);
```
labkey-jeckels: I know we don't have any other usages yet, but might as well make this a final static variable

labkey-jeckels (approved):

api/src/org/labkey/api/data/Container.java
labkey-adam: Done

labkey-adam:
Wrong branch name

PR #1372, opened 2020-07-07
Issue 40830 getkeycolumns

labkey-matthewb:
#### Rationale
As a performance fix we no longer generate database SQL when we generate a TableInfo from a query.  This means that the "_resolved" flag is also not set.  This test case found an assert that I believe is unnecessary as we will resolve fields on demand (SelectColumn.getResolvedField()).


labkey-matthewb:
Sorry, wrong base branch.  Will need to cherry pick the two commits.

PR #1375, opened 2020-07-08
fix display path (use container.path instead of docid)

labkey-matthewb:
#### Rational
Display text in search result was not remapped from guid -> path

#### Related
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40785


labkey-tchad (approved):

labkey-adam (approved):
Please add a reference to the 20.7 issue, which I think is https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40785

PR #1378, opened 2020-07-08
Item 7490: Upgrade to Log4J 2 - Part 1

labkey-ankurj:
#### Rationale
This PR is the starting point towards the effort to upgrade our logging to log4j2. After the code review of this branch, we will not merge this branch in develop. For the next part of the work, a new branch will be created from this branch and pull request from the new branch to this branch will be created and code review for that story and merge it back. This will keep on happening until we are done and then this original branch will be merged to develop. Let me know if someone has suggestion for an easier approach.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/395
* https://github.com/LabKey/commonAssays/pull/205
* https://github.com/LabKey/ehrModules/pull/57

#### Changes
Shorter version:
* ```org.apache.log4j.Logger.getLogger()``` --> ```org.apache.logging.log4j.LogManager.getLogger()```
* ```org.apache.log4j``` --> ```org.apache.logging.log4j```


labkey-ankurj:
The builds and teamcity tests will fail as the build.gradle which is in SVN will not be checked in until the work is complete.

labkey-adam:
In the past, I've added conditionals to the (svn-based) build.gradle file to keep it working with both develop and my branch. And then I remove the conditionals after I finally merge to develop. In my case, I was conditionalizing on the presence of the assay module, but you could conditionalize on something else: https://svn.mgt.labkey.host/stedi/trunk/server/build.gradle/?p=64176

labkey-ankurj:
> In the past, I've added conditionals to the (svn-based) build.gradle file to keep it working with both develop and my branch. And then I remove the conditionals after I finally merge to develop. In my case, I was conditionalizing on the presence of the assay module, but you could conditionalize on something else: https://svn.mgt.labkey.host/stedi/trunk/server/build.gradle/?p=64176

There is more to that, the gradle plugin needs to read the changed log4j.xml (which is also in SVN).

labkey-susanh:
> > In the past, I've added conditionals to the (svn-based) build.gradle file to keep it working with both develop and my branch. And then I remove the conditionals after I finally merge to develop. In my case, I was conditionalizing on the presence of the assay module, but you could conditionalize on something else: https://svn.mgt.labkey.host/stedi/trunk/server/build.gradle/?p=64176
> 
> There is more to that, the gradle plugin needs to read the changed log4j.xml (which is also in SVN).

We can make a short-term change to the plugin to pick up a different log4j.xml based on, say, the presence of some property if that makes things easier, but since you're having to change all the import paths for log4j, I'm not seeing a clear way to have both log4j options available based on some property or other without a lot of extra work.

core/src/org/labkey/core/CoreController.java
```diff
@@ -2636,4 +2638,89 @@ public ApiResponse execute(TransformWikiForm form, BindException errors)
...
+
+        }
+
+
+        @Override
+        protected ModelAndView handleGet()
+        {
+            Runnable testThread = this::randomSentence;
+
+            for (int i=0; i < 4000; i++)
```
labkey-jeckels, line 2715: Do we get value from having this many threads? 4000 is quite a stress test. Just calling randomSentence() directly in the for loop seems like it would be fine

api/src/org/labkey/api/util/MXBeanAppender.java
```diff
@@ -31,20 +37,28 @@
  * Date: 2012-02-28
  * Time: 4:10 PM
  */
-public class MXBeanAppender extends org.apache.log4j.AppenderSkeleton implements ErrorsMXBean
+@Plugin(name = "SessionAppender", category = "Core", elementType = "appender", printObject = true)
```
labkey-jeckels: Looks like name was copy/pasted, should probably be MXBeanAppender?

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -7627,17 +7628,18 @@ public ApiResponse execute(Object o, BindException errors)
...
             }
             catch (NumberFormatException x) {}
-            LoggingEvent[] events = SessionAppender.getLoggingEvents(getViewContext().getRequest());
+            LogEvent[] events = SessionAppender.getLoggingEvents(getViewContext().getRequest());
             ArrayList<Map<String, Object>> list = new ArrayList<>(events.length);
-            for (LoggingEvent e : events)
+            for (LogEvent e : events)
             {
-                if (eventId==0 || eventId<Integer.parseInt(e.getProperty("eventId")))
+                // TODO - find replacement - maybe a ThreadLocal using a WeakHashMap of Event->EventId?
```
labkey-jeckels: Is this on the list for the next story?

labkey-jeckels (commented):
I guess I'll hold the final approval until it's ready to merge?

labkey-martyp (changes requested):
Tested out both ETL and data entry methods for SessionAppender and they look good.

Not sure if you saw my comments here.  That eventIdMap in the SessionAppender is accessible by muliple threads so should be thread safe.
https://github.com/LabKey/platform/pull/1423

labkey-martyp:
Also tutorialModules repo has some build issues with these changes, not sure if that is coming in another round of updates

labkey-ankurj:
> Tested out both ETL and data entry methods for SessionAppender and they look good.
> 
> Not sure if you saw my comments here. That eventIdMap in the SessionAppender is accessible by muliple threads so should be thread safe.
> #1423

Here you go: https://github.com/LabKey/platform/pull/1378/commits/45f909b18c7dbf4a9e9dc01d3c1119dcebf07fd6. I pushed the changes in closed branch.

labkey-ankurj:
> Also tutorialModules repo has some build issues with these changes, not sure if that is coming in another round of updates

thanks for letting me know. I have not upgraded log4j in tutorialModules, will check it out as well.

labkey-martyp (approved):

api/src/org/labkey/api/pipeline/PipelineJob.java
```diff
@@ -1498,28 +1498,19 @@ public void setErrorStatus(Object message)
 
         public void write(String message, @Nullable Throwable t, String level)
         {
-            String formattedDate = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX").format(new Date());
+            String formattedDate = new SimpleDateFormat("dd MMM yyyy HH:mm:ss,SSS").format(new Date());
```
labkey-jeckels: Don't create a new date format for every invocation - use DateUtil.formatDateTime() which caches and uses a thread-safe implementation

labkey-jeckels (changes requested):

labkey-jeckels (approved):

PR #1379, opened 2020-07-09
Item 7169: Biologics sample create - Update GetQueryDetailsAction return correctly cased query name

labkey-martyp:
#### Rationale
GetQueryDetailsAction is returning the passed in query name parameter as the query name of successfully retrieved tableinfos. This does not always have the correct casing. If table info is found and it's not a wrapped table (has the same name but possibly different casing) set the query name in the response to be that of the table info.  Let server be source of truth on what the table name is.

#### Related Pull Requests
https://github.com/LabKey/biologics/pull/628

#### Changes
Set correct query name after table info found


query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
```diff
@@ -171,6 +171,11 @@ public ApiResponse execute(Form form, BindException errors)
         if (auditHistoryUrl != null)
             resp.put("auditHistoryUrl", auditHistoryUrl);
 
+        // If name is the same name just not the right casing, update response to what is stored on the server. Don't
+        // change name if the query is a wrapper around the tinfo
+        if (form.getQueryName().toLowerCase().equals(tinfo.getName().toLowerCase()))
+            resp.put("name", tinfo.getName());
```
labkey-jeckels: I don't understand the comment here - if the query is a wrapper around the table, it wouldn't have the same name (regardless of casing) as the table, would it?

It seems like always replacing the value with the canonical name from the TableInfo would be best here. There are cases where we support backwards compatible names when a table is renamed, and in those cases I think we should return the new, preferred name

labkey-jeckels (changes requested):

query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
labkey-martyp: As we discussed I changed this to the tinfo public name and tested across test suites.  Please see related test update PRs.
https://github.com/LabKey/targetedms/pull/233
https://github.com/LabKey/customModules/pull/41

PR #1381, opened 2020-07-10
Issue #40862

labkey-adam:
#### Rationale
[Issue 40862: Migrate wiki ad hoc \<select> elements to SelectBuilders](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=40862)

#### Changes
* Convert some ad hoc \<select> elements to SelectBuilders
* Override default `cellTpl` for `Ext4.tree.Column`

### DIFF
--- a/api/webapp/ext-4.2.1/ext-patches.js
+++ b/api/webapp/ext-4.2.1/ext-patches.js
@@ -1031,31 +1031,3 @@ Ext4.override(Ext4.window.MessageBox, {
     }
 });
 
-/**
- * @Override
- * Ext4.tree.Column does not HTML encode display values. This overrides the default cell template
- * to encode the value using {value:htmlEncode}.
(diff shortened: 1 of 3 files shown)


labkey-adam:
Build and BVTs are green. Now running the Wiki suite (disguised as BVT Postgres).

labkey-jeckels (approved):

labkey-nicka:
In addition to the ExtJS fix I've pushed a fix for the `WikiLongTest` failure.

labkey-jeckels:
Will this potentially double-encode in places like the schema browser, where I think/hope we've already done encoding?

labkey-nicka:
> Will this potentially double-encode in places like the schema browser, where I think/hope we've already done encoding?

Likely. It looks like we ask the server to do the encoding in that case.

```js
LABKEY.ActionURL.buildURL('query', 'getSchemaQueryTree.api', null, {showHidden : this.showHidden, withHtmlEncoding:true})
```

PR #1382, opened 2020-07-10
Issue 40848: Adding Percent Neutralization Max and Percent Neutralization Initial Dilution to the NAb Assay results and copied-to-study dataset

cnathe:
#### Rationale
See info from issue [40848](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40848) and related support ticket [40640](https://www.labkey.org/Atlas/support%20tickets/issues-details.view?issueId=40640). A lab using the LabKey NAb assay tool requested that two new values be calculated and added to the NAb assay results grid. These values were simple calculations off of the DilutionData values we already stored for the NAb assay runs.

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/206

#### Changes
* add ExprColumn for two calculated values to the NAbSpecimenTable
* expose the new calculated columns in the NabRunDataTable, but don't add them to the default view
* override isValidDataProperty() in NAbDataHandler so that it considers the two new column names valid

### DIFF
--- a/assay/api-src/org/labkey/api/assay/nab/query/NAbSpecimenTable.java
+++ b/assay/api-src/org/labkey/api/assay/nab/query/NAbSpecimenTable.java
@@ -113,8 +113,7 @@ public class NAbSpecimenTable extends FilteredTable<AssayProtocolSchema>
     {
         return new SQLFragment("(SELECT MAX(PercentNeutralization) FROM ")
             .append(DilutionManager.getTableInfoDilutionData(), "dd")
-            .append(" WHERE dd.RunDataId = ").append(ExprColumn.STR_TABLE_ALIAS + ".RowId")
-            .append(" GROUP BY dd.RunDataId)");
+            .append(" WHERE dd.RunDataId = ").append(ExprColumn.STR_TABLE_ALIAS + ".RowId)");
     }
 
     private SQLFragment getPercentNeutralizationInitialDilution()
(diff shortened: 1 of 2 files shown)


assay/api-src/org/labkey/api/assay/nab/query/NAbSpecimenTable.java
```diff
@@ -91,11 +109,37 @@ private SQLFragment getSelectedCurveFitAUC(boolean positive)
         return sql;
     }
 
+    private SQLFragment getPercentNeutralizationMax()
+    {
+        return new SQLFragment("(SELECT MAX(PercentNeutralization) FROM ")
+            .append(DilutionManager.getTableInfoDilutionData(), "dd")
+            .append(" WHERE dd.RunDataId = ").append(ExprColumn.STR_TABLE_ALIAS + ".RowId")
+            .append(" GROUP BY dd.RunDataId)");
```
labkey-jeckels: This GROUP BY seems harmless, but unnecessary

labkey-jeckels (approved):

assay/api-src/org/labkey/api/assay/nab/query/NAbSpecimenTable.java
cnathe: Good catch. I had it in there when I was testing the SQL in the schema browser and including some more columns in the SELECT list. Removed.

labkey-klum (approved):

PR #1384, opened 2020-07-11
Issue 40882: Migrate HTML encoding from server actions to client

labkey-nicka:
#### Rationale
This PR addressed issues [40882](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40882) which outlines the desire to migrate HTML encoding away from our server actions and into the client. Additionally, this addresses [40883](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=40883) by patching Ext3 and Ext4 to HTML encode their respective tree components.

Note: There are a number of client-side usages of these endpoints so there may be additional test failures that need to be addressed before merging this PR.

#### Changes
* Update actions outlined in [40082](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40882) to no longer HTML encode properties in their payload.
* Added a patch for Ext3 to encode the display text for the `Ext.tree.TreeNodeUI` component.
* Added a patch for Ext4 to encode the display text for the `Ext4.tree.Column` component.
* Additionally revise `query-getSchemaQueryTree.api` to no longer specify a `withHtmlEncoding` parameter nor pass a `qtip` property in the payload -- let the client be explicit about it.

### DIFF
--- a/query/webapp/query/browser/Tree.js
+++ b/query/webapp/query/browser/Tree.js
@@ -21,7 +21,7 @@ Ext4.define('LABKEY.query.browser.Tree', {
                 extend: 'Ext.data.Model',
                 proxy: {
                     type: 'ajax',
-                    url: LABKEY.ActionURL.buildURL('query', 'getSchemaQueryTree.api', null, {showHidden : this.showHidden, withHtmlEncoding:true}),
+                    url: LABKEY.ActionURL.buildURL('query', 'getSchemaQueryTree.api', null, { showHidden : this.showHidden }),
                     // don't use the "_dc" defeat cache parameter
                     noCache: false,
                     listeners: {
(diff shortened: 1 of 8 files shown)


query/webapp/query/browser/Tree.js
```diff
@@ -36,7 +36,7 @@ Ext4.define('LABKEY.query.browser.Tree', {
                     {name: 'description'},
                     {name: 'hidden', type: 'boolean', defaultValue: false},
                     {name: 'name'},
-                    {name: 'qtip'},
+                    {name: 'qtip', convert: function(_, rec) { return Ext4.htmlEncode(rec.raw.description); } },
```
labkey-nicka, line 39: I investigated default encoding for Ext4 QuickTip as well, however, I wasn't as confident about it's usages so I encoded the value into the model here.

labkey-jeckels (approved):
Looks good to me.

labkey-nicka:
Additionally ran DailyA, B, C, and DRTs with success (or as much success as `develop`).

PR #1386, opened 2020-07-12
Eliminate usages and remove deprecated methods & classes

labkey-adam:
#### Rationale
Make Matt happy by eliminating `DelegatingContainerFilter` and `ParameterMap`. Reduce maintenance by removing other deprecated and unused (or barely used) methods and classes. Switch callers of `InputTag.getValue()` to use our best practice for tags: pass in a String and let the builder encode (except in one case where we actually need to pass an HTML value).

#### Changes
* Eliminate `DelegatingContainerFilter` and `ParameterMap`
* Remove other deprecated and unused (or barely used) methods and classes
* Switch all but one caller of `InputTag.getValue()` to pass in string


labkey-matthewb:
Yeah

labkey-matthewb (approved):

PR #1387, opened 2020-07-13
Fast-forward for 20.3.10

labkey-tchad:


labkey-susanh (approved):
Test failure seems to be not related to these changes (missing tandem, etc.). Perhaps the TC project configuration needs to be updated?

PR #1391, opened 2020-07-13
40893: enforce display column width using overflow-wrap

labkey-nicka:
#### Rationale
This PR addresses [40893](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40893) by updating `DisplayColumn` to use `overflow-wrap: anywhere` in instead of `word-break: break-all`. Here is the description of `anywhere` from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap):
> To prevent overflow, an otherwise unbreakable string of characters — like a long word or URL — may be broken at any point if there are no otherwise-acceptable break points in the line. No hyphenation character is inserted at the break point. Soft wrap opportunities introduced by the word break are considered when calculating min-content intrinsic sizes.

This allows for the cells to wrap on common break points (e.g. spaces, hyphens, etc) but will enforce the content breaks if a common break point cannot be found. Here is an example of two long running values in the "description" column:

image

#### Changes
* Enforce explicit display column width by using `overflow-wrap: anywhere`.
* Updated issue comment to most recent.

### DIFF
--- a/api/src/org/labkey/api/data/DisplayColumn.java
+++ b/api/src/org/labkey/api/data/DisplayColumn.java
@@ -1011,7 +1011,7 @@ public abstract class DisplayColumn extends RenderColumn
         if (_nowrap)
             style += "white-space:nowrap;";
 
-        // 34871: Support for column display width
+        // 40893: Support for column display width
         if (!StringUtils.isBlank(getWidth()))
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
Looks great, thanks for the continued iteration.

PR #1392, opened 2020-07-14
Fb notebook api 7479

labkey-klum:
#### Rationale
This adds support for the initial set of server Notebook APIs so that the ELN client can be eventually hooked up to store notebook state into the database. Notebooks will use wiki's as the backend to store the contents of a notebook entry, changes were made to the wikiService to support this usage.

#### Related Pull Requests
- https://github.com/LabKey/labbook/pull/32

#### Changes
Many of the high level wiki model objects (Wiki and WikiVersion) are not part of the API, a wiki can be uniquely identified by a name/container combination. This change adds support to the WikiService to :
- getContent
- updateContent
- deleteWiki


wiki/src/org/labkey/wiki/WikiManager.java
```diff
@@ -898,6 +898,52 @@ private void fireWikiDeleted(User user, Container c, String name)
...
+                        return updateWiki(user, wiki, version, false);
+                    }
+                }
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public void deleteWiki(Container c, User user, String wikiName, boolean deleteSubtree) throws SQLException
```
labkey-nicka, line 938: nit: I realize that the `WikiManager.deleteWiki()` method is `void` but could consider having this service/interface method return `boolean`. I'm mainly thinking of the case where the wiki is `null`.

labkey-nicka (approved):

labkey-nicka:
@labkey-klum Should we prevent the search indexer from indexing the content of these Notebook wiki's?

labkey-kevink (approved):

PR #1393, opened 2020-07-14
40910: mute mothership client-side errors when page unloads

labkey-nicka:
#### Rationale
This PR disables client-side mothership reporting of errors once the page fires the `beforeunload` event. This is being done to avoid reporting flaky errors that occur when outbound async requests are aborted/cancelled by the browser upon navigation.

I could have hooked the `unload` page event instead, however, in my local testing that event fires well after request processing has failed so I elected to bind `beforeunload` instead. A drawback of this approach is that a user could invoke the `beforeunload` event but subsequently elect to stay on the page and mothership would stay disabled. I've elected to forego attempting to handle that scenario as there isn't really a clear solution to recognize this case.

#### Changes
* Disable client-side mothership reporting upon `beforeunload` event.

### DIFF
--- a/internal/webapp/mothership.js
+++ b/internal/webapp/mothership.js
@@ -773,7 +773,7 @@ LABKEY.Mothership = (function () {
         window.addEventListener('error', report);
 
         // 40910: mute reporting errors when page unloads
-        window.addEventListener('beforeunload', function() { enable(true); });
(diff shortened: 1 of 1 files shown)


labkey-tchad (approved):

cnathe (approved):
Clever and clean! Looks good to me.

PR #1407, opened 2020-07-15
Issue 40921: ResultSetImpl.getSize updates to ensure resultset iterated before calling getSize

labkey-martyp:
#### Rationale
ResultSetImpl.getSize behavior changed to require the result set to be fully iterated first before getSize can be called.  Added tracking to enforce that and an unsupported exception if trying to call before fully iterated.  Added method to iterate and calculate size if usage does not iterate already.

#### Changes
- ResultSetImpl and descendants call hasNext when returning from Next() which will set flag if resultset is fully iterated
- getSize will throw error if the fully iterated flag is not true
- countAll will iterated (if not already done) and return the size

### DIFF
--- a/api/src/org/labkey/api/data/DataRegion.java
+++ b/api/src/org/labkey/api/data/DataRegion.java
@@ -1099,7 +1099,7 @@ public class DataRegion extends DisplayElement
         if (useTableWrap)
             out.write("</tbody></table>");
 
-        if (usesResultSet() && rs instanceof TableResultSet && ((TableResultSet) rs).iterateAll() && ((TableResultSet) rs).getSize() != -1)
+        if (usesResultSet() && rs instanceof TableResultSet && ((TableResultSet) rs).countAll() != -1)
         {
             _rowCount = ((TableResultSet) rs).getSize();
             if (_complete && _totalRows == null)
--- a/api/src/org/labkey/api/data/DataRegion.java
+++ b/api/src/org/labkey/api/data/DataRegion.java
@@ -1099,7 +1099,7 @@ public class DataRegion extends DisplayElement
         if (useTableWrap)
             out.write("</tbody></table>");
 
-        if (usesResultSet() && rs instanceof TableResultSet && ((TableResultSet) rs).getSize() != -1)
+        if (usesResultSet() && rs instanceof TableResultSet && ((TableResultSet) rs).iterateAll() && ((TableResultSet) rs).getSize() != -1)
         {
             _rowCount = ((TableResultSet) rs).getSize();
             if (_complete && _totalRows == null)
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/data/TableResultSet.java
```diff
@@ -40,4 +40,9 @@
 
     /** @return the number of rows in the result set. -1 if unknown */
     int getSize();
+
+    default int countAll() throws SQLException
+    {
+        return -1;
```
labkey-jeckels, line 46: This means that many implementations will just return -1. Could the default call getSize() instead? Or add a comment about -1 being unknown, like we have on getSize() already?

api/src/org/labkey/api/data/DataRegion.java
```diff
@@ -1099,7 +1099,7 @@ private void _renderTableNew(RenderContext ctx, Writer out, ResultSet rs) throws
         if (useTableWrap)
             out.write("</tbody></table>");
 
-        if (usesResultSet() && rs instanceof TableResultSet && ((TableResultSet) rs).getSize() != -1)
+        if (usesResultSet() && rs instanceof TableResultSet && ((TableResultSet) rs).countAll() != -1)
```
labkey-jeckels, line 1102: I'd probably leave this one as getSize() - it shouldn't have any more iteration to do at this point, and if it does, it's probably some other kind of bug

api/src/org/labkey/api/data/ResultSetImpl.java
```diff
@@ -94,28 +94,55 @@ void setComplete(boolean isComplete)
         _isComplete = isComplete;
     }
 
+    protected boolean hasNext(boolean hasNext)
```
labkey-jeckels, line 97: Please add a comment on this one.

labkey-jeckels (changes requested):
I like this change, but I think we can defer it to develop if we make this simple tweak to AssayProtocolSchema:
```
Index: api/src/org/labkey/api/assay/AssayProtocolSchema.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- api/src/org/labkey/api/assay/AssayProtocolSchema.java	(revision d6ace0670242e226f78d036ce654597f890e8391)
+++ api/src/org/labkey/api/assay/AssayProtocolSchema.java	(date 1593188694757)
@@ -700,7 +700,9 @@
                             QueryView allResultsQueryView = createAllResultsQueryView(viewContext, qs);
 
                             DataView dataView = allResultsQueryView.createDataView();
-                            try (Results r = dataView.getDataRegion().getResults(dataView.getRenderContext()))
+                            RenderContext renderContext = dataView.getRenderContext();
+                            renderContext.setCache(true);
+                            try (Results r = dataView.getDataRegion().getResults(renderContext))
                             {
                                 final int rowCount = r.getSize();
 
```

PR #1408, opened 2020-07-16
Issue 40918: Sample Import has become slower with change to prevent deadlocks

labkey-jeckels:
#### Rationale
We like speed

#### Changes
* To speed up the queuing for the indexing, do a single big query instead of a bunch of little queries (one per row). This mirrors the approach already taken in ExpDataClassDataTableImpl.persistRows()


labkey-martyp (approved):
Nice, looks like we already did this for exp.datas too

labkey-susanh (approved):
Confirmed with a local build that import times are now back to 20.6 levels.

labkey-martyp (approved):
I ran this a handful of times on TC and did not get any deadlocks. Looks good for 20.7

PR #1409, opened 2020-07-16
Issue 40921: QC message about hidden rows not being displayed

labkey-martyp:
#### Rationale
Due to updates in size calculations of resultsets and data regions, the QC row count comparison used to generate QC hidden rows message is not working.  This updates AssayProtocolSchema's calculations to generate that message

#### Changes
- Update calculation of all rows, not filtered for QC to use cached result sets. This will have calculated size.
- Update data region to iterate through its rows on an aggregate operation to get total row count.


labkey-jeckels (approved):
Looks good, but please add a note in the comment to point at issue 40921, and to the line that calls setCache(false) too.

PR #1414, opened 2020-07-16
Allow HEAD request on whoami action

labkey-martyp:
#### Rationale
[Issue 40927](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=40927): Allow HEAD request on whoami action

#### Changes
- Allow HEAD and GET methods on WhoAmIAction
- Cleanup imports in this file

### DIFF
--- a/core/src/org/labkey/core/login/LoginController.java
+++ b/core/src/org/labkey/core/login/LoginController.java
@@ -2625,7 +2625,7 @@ public class LoginController extends SpringActionController
     @SuppressWarnings("unused")
     @RequiresNoPermission
     @AllowedOutsideImpersonationProject
-    @MethodsAllowed(HttpUtil.Method.HEAD)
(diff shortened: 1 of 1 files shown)


labkey-matthewb (commented):
If we're going to accept HEAD we should probably actually do what this method means

9.4 HEAD
The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response. The metainformation contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request. This method can be used for obtaining metainformation about the entity implied by the request without transferring the entity-body itself. This method is often used for testing hypertext links for validity, accessibility, and recent modification.

The response to a HEAD request MAY be cacheable in the sense that the information contained in the response MAY be used to update a previously cached entity from that resource. If the new field values indicate that the cached entity differs from the current entity (as would be indicated by a change in Content-Length, Content-MD5, ETag or Last-Modified), then the cache MUST treat the cache entry as stale.

PR #1417, opened 2020-07-19
Issue 40921: ResultSetImpl.getSize() must iterate through result set first

labkey-martyp:
#### Rationale
ResultSetImpl was updated to calculate size on iteration, therefore calls to getSize should throw an exception if the result set is not completely iterated.  This PR includes tracking for when the result set is iterated and throws an exception if getSize is called before completely iterated.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1409

#### Changes
- ResultSetImple.getSize() throws exception if result set not iterated
- ResultSetImpl.hasNext() added to track when iterated. Called from ResultSetImpl.next() and classes overriding this function.
- Update to ResultSetImpl.next() to ensure counting is done for all rows case
- Add countAll() function to iterate and count a result set

### DIFF
--- a/api/src/org/labkey/api/assay/AssayProtocolSchema.java
+++ b/api/src/org/labkey/api/assay/AssayProtocolSchema.java
@@ -670,7 +670,7 @@ public abstract class AssayProtocolSchema extends AssaySchema
                             DataView dataView = allResultsQueryView.createDataView();
                             try (Results r = dataView.getDataRegion().getResults(dataView.getRenderContext()))
                             {
-                                final int rowCount = r.getSize();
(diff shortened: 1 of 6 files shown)


labkey-jeckels (changes requested):
These changes look good, but don't yet have the assay header aspect folded in.

labkey-martyp:
@labkey-jeckels Ok I updated the AssayProtocolSchema to use the countAll method and not the cached result set.  Verified it passed the QC tests on biologics and core.

labkey-jeckels (approved):

PR #1421, opened 2020-07-21
Use readily available test data for study export test

labkey-tchad:
#### Rationale
`StudyExportTest` started failing because it depends on a list archive that was intended for Rlabkey testing and isn't necessarily present. The test doesn't actually seem to care what list archive you import.

#### Changes
* Use a list archive from `testAutomation/data`


labkey-danield (approved):
I guess this looks ok. It looks like the tests doesn't care about the content of the file, it just needs some file to import and export.

PR #1423, opened 2020-07-22
Item 7554: part 2 of log4jupgrade

labkey-ankurj:
#### Rationale
This is part two of log4j upgrade work, includes changes to custom appenders, event id handling for server javascript console and admin logging UI changes

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1378


api/src/org/labkey/api/util/SafeFileAppender.java
```diff
@@ -83,4 +114,71 @@ public void append(LogEvent loggingEvent)
         }
     }
 
+    private void logJobMessage(LogEvent logEvent, @Nullable Throwable t)
+    {
+        if (logEvent.getLevel().compareTo(Level.DEBUG) == 0)
```
labkey-jeckels: Can these use equals() instead of compareTo()?

And while we seldom use it, this is missing the TRACE level

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -7628,17 +7628,17 @@ public ApiResponse execute(Object o, BindException errors)
                     eventId = Integer.parseInt(s);
             }
             catch (NumberFormatException x) {}
-            LogEvent[] events = SessionAppender.getLoggingEvents(getViewContext().getRequest());
-            ArrayList<Map<String, Object>> list = new ArrayList<>(events.length);
-            for (LogEvent e : events)
+            Map<LogEvent, String> events = SessionAppender.getLoggingEvents(getViewContext().getRequest());
```
labkey-jeckels, line 7631: @labkey-martyp if you haven't already looked at this, can you make sure that it doesn't break the ETL logging scenario that you recently worked on?

labkey-jeckels (changes requested):

api/src/org/labkey/api/util/SafeFileAppender.java
labkey-ankurj: Should I add the trace level (It wasn't there earlier.)? Are there pipeline jobs that have the trace logging ?

labkey-ankurj: Here is old code in Pipeline.java
Screen Shot 2020-07-22 at 3 00 18 PM

labkey-jeckels: OK, but let's make the new code better than the old. :)

labkey-jeckels: I doubt anyone's using trace now (since it doesn't work, apparently), but let's support it. I added trace level logging for some of the recent ODBC work, so I think there are legit use cases within our codebase

labkey-ankurj: sounds good, added it.

labkey-jeckels (approved):
Looks good, but let's wait for Marty confirm on the JS console logging for the ETL scenario if you haven't already connected

api/src/org/labkey/api/util/LoggerWriter.java
```diff
@@ -0,0 +1,36 @@
...
+ * This is the helper class to provide a custom logger writer which can be
+ * used to route log messages to a different location than the logger. ex - to a file.
+ *
+ * User : ankurj
+ * Date : Jul 23, 2020
+ * */
+
+public interface LoggerWriter
+{
+    void write(String message, @Nullable Throwable t);
```
labkey-jeckels, line 15: dumpThreads() is only calling a single method, so I think this interface could much simpler:

void debug(String message)

or you call call it write() if that feels better. I don't expect we'll need to add other log levels in the future since it's such a targeted use case

api/src/org/labkey/api/util/FileLoggerWriter.java
```diff
@@ -0,0 +1,126 @@
...
+{
+    private final String LINE_SEP = System.getProperty("line.separator");
+    private static Logger LOG = LogManager.getLogger(FileLoggerWriter.class);
+
+    private File _file;
+
+    public void writeToFile(File file)
+    {
+        // Make sure that we try to mount the drive (if needed) before using the file
+        NetworkDrive.exists(file);
```
labkey-jeckels: Since we only have a single use case, I recommend having the constructor open the BufferedWriter, adding a close() method, and calling it after all of the write() calls are done. That will make it so that you don't have to keep opening and closing the file for every line that gets written

labkey-jeckels (changes requested):

api/src/org/labkey/api/util/LoggerWriter.java
labkey-ankurj: there is a call to DbScope.logCurrentConnectionState in DebugInfoDumper.dumpThreads (L282) which uses info, call to ConnectionWrapper.dumpOpenConnections at (L290) which uses debug and call to ConnectionWrapper.dumpLeaksForThread (L272) which uses error.

labkey-jeckels: Indeed. I'm not sure that the debug/error split is all that compelling, but given the differentiation let's stick with your approach.

labkey-jeckels (approved):

api/src/org/labkey/api/util/SessionAppender.java
```diff
@@ -15,30 +15,45 @@
...
 /**
  * User: matthewb
  * Date: Aug 18, 2009
  * Time: 2:18:21 PM
  */
-@Plugin(name = "SessionAppender", category = "Core", elementType = "appender", printObject = true)
+@Plugin(name = "SessionAppender", category = Core.CATEGORY_NAME, elementType = Appender.ELEMENT_TYPE, printObject = true)
 public class SessionAppender extends AbstractAppender
 {
+    private static ThreadLocal<WeakHashMap<LogEvent, String>> _eventIds = new ThreadLocal<>();
```
labkey-martyp, line 47: Is this being used?

api/src/org/labkey/api/util/SessionAppender.java
```diff
@@ -54,7 +69,7 @@ protected SessionAppender(String name, Filter filter, Layout<? extends Serializa
 
         final String key;
         boolean on;
-        final List<LogEvent> list = new LinkedList<>();
+        final Map<LogEvent, String> eventIdMap = new WeakHashMap<>();
```
labkey-martyp, line 72: I think this should probably be a ConcurrentReferenceHashMap.  These can be accessed from multiple threads, so that will at least make the map operations thread safe.  See appenderInfos implementation of that map type

api/src/org/labkey/api/util/SessionAppender.java
```diff
@@ -90,25 +105,23 @@ public void append(LogEvent event)
...
+        synchronized (info.eventIdMap)
         {
-            // TODO - find replacement - maybe a ThreadLocal using a WeakHashMap of Event->EventId?
-            //            event.setProperty("eventId", String.valueOf(++info.eventId));
-            info.list.add(event);
-            if (info.list.size() > 1000)
-                info.list.remove(0);
+            info.eventIdMap.put(event, String.valueOf(++info.eventId));
+            if (info.eventIdMap.size() > 1000)
+                info.eventIdMap.remove(0);
```
labkey-martyp, line 112: This is something I should have fixed, but would probably be safer to override the put() function of that map with this check.

api/src/org/labkey/api/util/SessionAppender.java
labkey-martyp: Possibly something like this?  Still not great to just randomly remove index(0) but not sure how to get the eldest entry without using a LinkedHashMap, which does not have a weak reference option.

```suggestion
        final Map<LogEvent, String> eventIdMap = new ConcurrentReferenceHashMap<>(16, ConcurrentReferenceHashMap.ReferenceType.WEAK)
        {
            @Override
            public String put(LogEvent key, String value)
            {
                // Safeguard against runaway size. ConcurrentReferenceHashMap does not have removeEldestEntry so have  
                // to randomly remove an entry
                if (size() > 1000)
                    remove(0);

                return super.put(key, value);
            }
        };
```

api/src/org/labkey/api/util/SessionAppender.java
labkey-martyp: See suggestion above

labkey-martyp (changes requested):
Sorry, my delay getting to this probably got your branches out of sync with develop.  I'm having difficulty getting this to build and looks like its not building on TC. Let me know if I can be any help fixing the build, but assume you can do it a lot quicker for this branch.

PR #1431, opened 2020-07-28
add pipeline status details page test helper

labkey-kevink:
#### Rationale
- fix S3FileRootMigrationTest: support s3 cloud status files in StatusDetailsBean
- fix RedCapTest: find expected errors in log text

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/413
* https://github.com/LabKey/redcap/pull/29
* https://github.com/LabKey/cloud/pull/74
* https://github.com/LabKey/dataintegration/pull/58

### DIFF
--- a/pipeline/src/org/labkey/pipeline/status/FileDisplayColumn.java
+++ b/pipeline/src/org/labkey/pipeline/status/FileDisplayColumn.java
@@ -120,7 +120,7 @@ public class FileDisplayColumn extends SimpleDisplayColumn
     {
         Path parent = p.getParent();
 
-        if (Files.exists(parent))
+        if (NetworkDrive.exists(parent))
         {
             // calculate base name of the .status file
             String statusName = p.getFileName().toString();
--- a/pipeline/src/org/labkey/pipeline/status/FileDisplayColumn.java
+++ b/pipeline/src/org/labkey/pipeline/status/FileDisplayColumn.java
@@ -21,6 +21,7 @@ import org.labkey.api.data.RenderContext;
 import org.labkey.api.data.SimpleDisplayColumn;
 import org.labkey.api.pipeline.PipelineProvider;
 import org.labkey.api.pipeline.PipelineService;
+import org.labkey.api.util.FileUtil;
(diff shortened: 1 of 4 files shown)


pipeline/src/org/labkey/pipeline/status/FileDisplayColumn.java
```diff
@@ -91,49 +100,51 @@ public void renderDetailsCellContents(RenderContext ctx, Writer out) throws IOEx
...
+    }
+
+    public static List<Path> listFiles(Path p, Container c, PipelineProvider provider)
             throws IOException
     {
-        File dir = f.getParentFile();
+        Path parent = p.getParent();
 
-        if (NetworkDrive.exists(dir))
+        if (Files.exists(parent))
```
labkey-jeckels: We need a variant of NetworkDrive.exists() that takes a Path, so that we do the mounting of the drive if required on Windows.

labkey-jeckels (changes requested):

pipeline/src/org/labkey/pipeline/status/FileDisplayColumn.java
labkey-kevink: Done

pipeline/src/org/labkey/pipeline/status/details.jsp
```diff
@@ -232,7 +232,7 @@
         <% } else { %>
         <div id="log-data">
             <% for (var record : status.log.records) { %>
-            <pre class="labkey-log-text <%=logTextClass(record)%>"
+            <pre class="labkey-log-text <%=h(logTextClass(record))%>"
```
labkey-adam, line 235: I have a slight preference for making `logTextClass()` return HtmlString and then directly outputting that return value (pushing the encoding assertion to where the content is generated), but this is fine since we're all in the same JSP file.

labkey-adam (approved):
Josh should probably re-review the Path/ensure() changes, but looks good to me.

labkey-jeckels (approved):

PR #1432, opened 2020-07-28
7540: Module Editor Story 2

RosalineP:
#### Rationale
Permissions information must be passed with each file object to support module editor changes. 
Module editor file updates sometimes cause a path of "/" within audit logs, which should be replaced by the actual path.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/308
* https://github.com/LabKey/moduleEditor/pull/8

#### Changes
* Updates file's passed permissions
* Changes audit log output to account for 'root' being passed


api/src/org/labkey/api/webdav/AbstractWebdavResource.java
```diff
@@ -638,7 +638,8 @@ else if ("deleted".equalsIgnoreCase(message))
         }
         else if ("replaced".equalsIgnoreCase(message))
         {
-            message = "File replaced in " + c.getContainerNoun() + ": " + c.getPath();
+            String path = (c.getPath().equals("/")) ? c.getPath() : this.getPath().toString();
```
cnathe: Minor note, usually when using string .equals(), I'll put the hard coded string on the left hand side. For example, if for some reason your `c.getPath()` is `null` the current implementation will throw an NPE but if you flip it it won't.

cnathe (approved):

api/src/org/labkey/api/webdav/AbstractWebdavResource.java
RosalineP: Good point! Done

PR #1443, opened 2020-07-31
Issue 40918: Sample Import perf

labkey-jeckels:
#### Rationale
Improves performance for bulk-loading samples, especially those that have parent materials or datas

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1408

#### Changes
* Bulk load sample name->RowId map, up to a million records

* Add caching for:
ProtocolActions within a Protocol
SampleType by name for resolving parent samples
Poke more ExpData and ExpMaterials into import-scoped cache for possible reuse


api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -279,10 +298,19 @@ private Object fetch(Triple<ColumnInfo, ColumnInfo, MultiValuedMap<?,?>> triple,
...
         private TableSelector createSelector(ColumnInfo pkCol, ColumnInfo altKeyCol, Object value)
         {
-            SimpleFilter filter = new SimpleFilter(altKeyCol.getFieldKey(), value);
-            return new TableSelector(_targetTable, Arrays.asList(altKeyCol, pkCol), filter, null);
+            return createSelector(pkCol, altKeyCol, new SimpleFilter(altKeyCol.getFieldKey(), value));
+        }
+
+        private TableSelector createSelector(ColumnInfo pkCol, ColumnInfo altKeyCol, SimpleFilter filter)
+        {
+            return new TableSelector(_targetTable, Arrays.asList(altKeyCol, pkCol), filter, null).setMaxRows(1_000_000);
```
labkey-susanh, line 313: If the number of rows is more than MaxRows, will they simply be missed?

labkey-susanh (approved):
Seems fine to me.  Just one question about the million-row limit.

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
labkey-jeckels: We'll bulk pre-load at most 1 million rows. We will still query individually if a given value isn't in the cache, which would happen if either there are more than a million rows or someone requests a value that isn't in the table.

PR #1456, opened 2020-08-04
Make Junit Go action a standard MutatingApiAction

labkey-adam:
#### Rationale
JunitTest fails periodically claiming that the Go action returned "null". I don't know what's causing that, but we might as well use our standard API action instead of constructing the response "manually."

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/423

#### Changes
* Convert Go action into a MutatingApiAction

### DIFF
--- a/api/src/org/labkey/api/util/ExceptionUtil.java
+++ b/api/src/org/labkey/api/util/ExceptionUtil.java
@@ -1066,6 +1066,12 @@ public class ExceptionUtil
                 {
                     return user;
                 }
+
+                @Override
+                public String getMethod()
+                {
+                    return "GET";
+                }
             };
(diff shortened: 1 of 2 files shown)


labkey-tchad (changes requested):
This seems to be causing trouble for the redirect check in `org.labkey.api.util.ExceptionUtil$TestCase.testUnauthorized`: https://teamcity.labkey.org/viewLog.html?buildId=1079268&buildTypeId=LabkeyTrunk_BvtBPostgres
Running premium JUnit tests for extra validation:
https://teamcity.labkey.org/viewQueued.html?itemId=1079524
https://teamcity.labkey.org/viewQueued.html?itemId=1079526

labkey-tchad (approved):
Nothing new in the Git JUnit suite. Assuming BVTs look good (my `ExceptionUtil` fix worked locally), this is good to merge.

PR #1457, opened 2020-08-04
Issue 41058: GetQueryDetailsAction does not return public name of query

labkey-martyp:
#### Rationale
[41058](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41058): GetQueryDetailsAction returns the passed in query name as the actual query name in the response from this action, instead of the actual public name from the server.  This fixes this action to return the public name.  This affects the details in the schema browser, so needed to update the name on a handful of queries across a few repos.

#### Related Pull Requests
https://github.com/LabKey/targetedms/pull/241
https://github.com/LabKey/testAutomation/pull/424
https://github.com/LabKey/adjudication/pull/42

#### Changes, 
- Set name in GetQueryDetailsAction response to public name
- Pass through and set the correct name on AbstractMatrixSchema
- Update NAbSpecimenTable to use the same public name as the list of names in the schema

### DIFF
--- a/query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
+++ b/query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
@@ -171,6 +171,9 @@ public class GetQueryDetailsAction extends ReadOnlyApiAction<GetQueryDetailsActi
         if (auditHistoryUrl != null)
             resp.put("auditHistoryUrl", auditHistoryUrl);
 
+        // Return public name for query
+        resp.put("name", tinfo.getPublicName());
+
         resp.put("title", tinfo.getTitle());
         resp.put("titleColumn", tinfo.getTitleColumn());
 
--- a/assay/api-src/org/labkey/api/assay/matrix/AbstractMatrixProtocolSchema.java
+++ b/assay/api-src/org/labkey/api/assay/matrix/AbstractMatrixProtocolSchema.java
@@ -70,13 +70,13 @@ public abstract class AbstractMatrixProtocolSchema extends AssayProtocolSchema
     {
         if (name.equals(getDataBySampleTableName()))
         {
-            return getDataBySampleTable(cf, rowAxisId, colAxisId, valueMeasureName, title); //TODO: change
+            return getDataBySampleTable(cf, rowAxisId, colAxisId, valueMeasureName, name, title); //TODO: change
         }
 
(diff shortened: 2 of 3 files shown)


query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
```diff
@@ -171,6 +171,9 @@ public ApiResponse execute(Form form, BindException errors)
         if (auditHistoryUrl != null)
             resp.put("auditHistoryUrl", auditHistoryUrl);
 
+        // Return public name for query
+        resp.put("name", tinfo.getPublicName());
```
labkey-nicka, line 175: Rather than overriding what "name" represents I would recommend we use a new property instead. There are many usages of the `getQueryDetail.api` and I think we should be wary of backwards breaking changes where usages would have expected "name" to reflect what as in the `form.getQueryName()`. I would recommend "publicName".

With regards to the "In biologics this can cause caching of query details in multiple casings" as mentioned in the ticket -- the cache for `QueryInfo` found in `@labkey/components` is case-insensitive. If you are talking about another cache somewhere else I'd like to discuss that implementation.

labkey-nicka (changes requested):

query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
labkey-martyp: Yeah I was concerned about the same, but I ran a ton of our automated tests and I didn't find anywhere we're relying on the name in the response from the server to match the name passed in.  Do you know of anywhere where that's the case? Or you concerned more about external implementations?  The only place I found any issues was in the schema browser because GetSchemaQueryTreeAction uses getVisibleTableNames, which was different than public name in a few cases.

labkey-jeckels: "name" as currently implemented isn't all that helpful. In cases where it doesn't match what would be returned from getPublicName(), it tends to be the raw table name from the underlying database, which is of no interest to the client.

labkey-nicka: It's not the raw table name but rather reflecting back to the client the value from `form.getQueryName()`.

labkey-nicka (approved):

PR #1462, opened 2020-08-05
Change SimpleTableDomainKind to use getValueSql

labkey-martyp:
#### Rationale
Looks like SimpleTableDomainKind needs to use getValueSql instead of getSelectName after the related PR.
https://teamcity.labkey.org/repository/download/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_SnprcEhrSqlserver/1080510:id/tomcatLogs.tar.gz!/labkey-errors.log

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1458

#### Changes
Update SimpleTableDomainKind.sqlObjectIdsInDomain to use getValueSql


labkey-matthewb (changes requested):
It's not obvious, but It is a little suspect to mix and match the SchemaTableInfo (variable table) and a columns from the SimpleTable (variable table).  Makes more sense to stick with only objects from table or only objects from schemaTable.

Also avoid using  String + SQLFragment.  That may not do what you want.

sql.append("SELECT o.ObjectId FROM ").append(table.getFromSql("me")).append(...).append(objectUriColumn.getValueSql("me").append(...)

labkey-matthewb (approved):

PR #1463, opened 2020-08-05
Package updates

labkey-nicka:
#### Rationale
NPM package updates across platform modules.

#### Changes
- Package updates for `assay`, `core`, `experiment`, `issues`, `list`, and `study` modules.
- Ran `npm audit fix` across all modules.
- Improve error handling in database authentication configuration app. Replace usages of `alert` and display errors to the user.
- Fix jest test no longer allowing mock of `window.location.assign`.


core/src/client/components/DatabaseConfigurationModal.tsx
```diff
@@ -87,6 +101,8 @@ export default class DatabaseConfigurationModal extends PureComponent<Props, Sta
                 </Modal.Header>
 
                 <Modal.Body>
+                    {hasError && <Alert bsStyle="danger">{error || initError}</Alert>}
```
cnathe, line 104: would there ever be a case where both errors would exist? If so, I could see having something like this be sufficient to catch the either-or case:
```suggestion
                    {error && <Alert bsStyle="danger">{error}</Alert>}
                    {initError && <Alert bsStyle="danger">{initError}</Alert>}
```

cnathe (approved):

cnathe:
Adding @RosalineP as a heads up to these auth config component changes.

core/src/client/components/DatabaseConfigurationModal.tsx
labkey-nicka: I don't think so. Init errors will only occur on `componentDidMount` and i prevented update in the case that there are init errors since we shouldn't allow mutation if we don't have current server props.

core/src/client/components/DatabaseConfigurationModal.tsx
```diff
@@ -59,24 +65,32 @@ export default class DatabaseConfigurationModal extends PureComponent<Props, Sta
     };
 
     saveChanges = (): void => {
+        if (this.state.error) {
+            this.setState({ error: undefined });
+        }
+
         Ajax.request({
-            url: ActionURL.buildURL('login', 'SaveDbLoginProperties'),
+            url: ActionURL.buildURL('login', 'saveDbLoginProperties.api'),
```
RosalineP, line 73: Is the effect of adding .api to these places mainly to make the code more readable?

RosalineP (approved):

PR #1465, opened 2020-08-06
Aggregate JSP print(String) statistics by file

labkey-adam:
#### Rationale
Collecting and logging JSP print(String) calls by JSP file will help prioritize the suites and JSP files to address first.

#### Changes
* Collect JSP print(String) statistics by file and log in order of occurrences
* Move all logging to JspStringPrints.log

### DIFF
--- a/api/src/org/labkey/api/jsp/LabKeyJspWriter.java
+++ b/api/src/org/labkey/api/jsp/LabKeyJspWriter.java
@@ -15,12 +15,11 @@
  */
 package org.labkey.api.jsp;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/jsp/LabKeyJspWriter.java
```diff
@@ -70,8 +69,9 @@ public void print(char[] s) throws IOException
     public void print(String s) throws IOException
     {
         StackTraceElement[] stackTrace = Thread.currentThread().getStackTrace();
```
labkey-jeckels, line 71: Probably not an issue because we're actively eliminating callers, but keep in mind that this is an expensive call

labkey-jeckels (approved):

PR #1467, opened 2020-08-07
On folder delete, disable the submit button after click

bbimber:
Several times recently we've had situations where a user deletes a folder or workbook.  After hitting 'delete' there is a delay while the delete is happening.  The user gets impatient and re-hits the delete button.  The server gives a 'there is a delete is progress' message, which is good; however, it seems like we should just disable the delete button after click.  This patch does that

### DIFF
--- a/core/src/org/labkey/core/admin/deleteFolder.jsp
+++ b/core/src/org/labkey/core/admin/deleteFolder.jsp
@@ -192,7 +192,7 @@
                 <%
             }
             %>
-            <%= button("Delete").submit(true) %>
+            <%= button("Delete").disableOnClick(true).submit(true) %>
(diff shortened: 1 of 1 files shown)


bbimber:
@labkey-jeckels I dont seem to be able to request reviewers here.  this is non-urgent, but this delete problem has happened enough times that I thought it was worth a patch.

labkey-jeckels:
Looks good. Created a copy on the LabKey repo to get TeamCity to run builds.

https://github.com/LabKey/platform/pull/1473

bbimber:
@labkey-jeckels we can close this PR now, right?

labkey-jeckels:
Yes!

PR #1470, opened 2020-08-07
Eliminate JSP print(String) warnings from community module test suites

labkey-adam:
#### Rationale
We want to eliminate rendering of raw Strings from JSPs. This checkpoint addresses most of the community module test suites.

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/224

#### Changes
* Remove unused `SelectTag`
* Return safe wrappers from more helper methods
* Add new capabilities to builders
* Clear all warnings from the platform & commonAssays module test suites (Assays, FileContent, Flow, Issues, Luminex, Mothership, MS2, Nab, SampleType, Search, Visualization, and Wiki)


api/src/org/labkey/api/util/element/Input.java
```diff
@@ -797,15 +830,21 @@ public T type(String type)
...
+        public T value(@Nullable HasHtmlString hhs)
+        {
+            _value = null != hhs ? hhs.getHtmlString() : null;
+            return (T)this;
+        }
+
         public T value(String value)
         {
-            _value = HtmlString.of(value);
+            _value = HtmlString.of(value); // TODO: Should null result in _value == null?
```
labkey-susanh: Why would you expect null to become non-null?

api/src/org/labkey/api/util/element/Input.java
```diff
@@ -905,6 +944,24 @@ public T stepValue(Integer stepValue)
...
+            return (T)this;
+        }
+
+        public T style(String style)
+        {
+            _styles.add(style);
+            return (T)this;
+        }
+
+        public T style(List<String> styles)
```
labkey-susanh: The naming of these methods seems odd to me.  I would expect something like `addStyle` and `addStyles`, though I see this has just been moved from the Select class to here.

issues/src/org/labkey/issue/model/Issue.java
```diff
@@ -719,16 +720,26 @@ public String getCreatedByName(User currentUser)
...
 
-        public void setComment(String comment)
+        public void setHtmlComment(HtmlString comment)
         {
             this.comment = comment;
         }
 
+        public String getComment()
+        {
+            throw new IllegalStateException("Callers should use getHtmlComment()");
```
labkey-susanh: `IllegalStateException` seems a little odd.  I would think `UnsupportedOperationException` perhaps?

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
```diff
@@ -277,7 +278,7 @@ public int createIssue(Container container, User user, @NotNull String issueDefN
         issue.setPriority("3");
         issue.setType("Todo");
         if (body != null)
-            issue.addComment(user, body);
+            issue.addComment(user, HtmlString.unsafe(body));
```
labkey-susanh: The body of the issue seems like it could indeed be unsafe, but perhaps we deal with that filtering elsewhere?

labkey-susanh (approved):

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
labkey-adam: In most cases, we have to simply trust that issue body Strings are safe, since we render and persist them as HTML in the database (unfortunate, of course). This specific method, however, has exactly one caller, in EvaluationContent. Makes sense to force callers (who construct the body) to assert correctness by passing in an HtmlString. To avoid adding another repo to this FB, I'll add the new method for now and then later update the caller & remove the old method.

api/src/org/labkey/api/util/element/Input.java
labkey-adam: Agreed. Changed.

api/src/org/labkey/api/util/element/Input.java
labkey-adam: Because `null` can be annoying? I added the comment when I noticed the inconsistency between not setting and explicitly setting value to null, but it didn't seem important to address immediately since blank and null are both handled by Input rendering. I'll clean it up.

issues/src/org/labkey/issue/model/Issue.java
labkey-adam: `UnsupportedOperationException` always strikes me as "we haven't gotten around to implementing this yet"... or that some other implementing class does support the method. Not "never ever call this method, you bad developer, it's illegal!" But the good news is I just deleted the method and nothing broke (yet). That's probably an even better indication that the method shouldn't be called...

labkey-adam: I sometimes forget that `BeanObjectFactory` (used by our ORM layer) and Spring binding have different requirements for setters and getters.

PR #1472, opened 2020-08-08
Issues 37376 and 35065: Pivot tables not respecting column displayWidths

labkey-martyp:
#### Rationale
In pivot tables, neither the pivot columns nor the non-pivot columns are respecting display widths set on the column metadata, such as in .query.xml files.  This PR fixes both cases.  Since the table-layout is default to auto, the tables are essentially resizing to contents and the container size.  This PR moves the metadata defined styles from the th tags to a div inside the th tag around the content.  The tables will then resize to the content.  Also removed the width: 100% from pivot queries only, as this was stretching out those column widths in some cases and ignoring the set widths.  Tested on pivot and non-pivot tables, with and without data in Chrome, FF and Edge.

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/427

#### Changes
- Move style applied to table column headers to a div wrapping the contents in the th tag instead of on the th itself. For consistency, the div wrapper is always on the th content regardless if style is applied.
- For pivot queries only, set width: auto on upper two rows of th tags


api/src/org/labkey/api/data/DisplayColumn.java
```diff
@@ -734,6 +719,21 @@ public void renderGridHeaderCell(RenderContext ctx, Writer out, String headerCla
...
+
+        // 34871: Support for column display width
+        if (!StringUtils.isBlank(getWidth()))
+            style += ";width:" + getWidth() + "px;";
+
+        if (!"".equals(style))
+        {
+            out.write("<div style=\"");
+            out.write(style);
+            out.write("\">");
```
labkey-jeckels: Is this right? Seems like we wouldn't want to close it here, and you're also closing it at line 759

labkey-jeckels (changes requested):

api/src/org/labkey/api/data/DisplayColumn.java
labkey-martyp: That was just closing the opening tag.  But I've updated this a bit to always include the div tag around th content, whether there is style applied or not.  This is for a consistent DOM across all column headers, for CSS and testing.

labkey-jeckels (approved):
I was misreading the code on the <div> generation so my previous comment was wrong. Having the consistent structure like you do now makes sense too.

PR #1473, opened 2020-08-09
On folder delete, disable the submit button after click

labkey-jeckels:
#### Rationale
Avoid letting users double-click and get an error. Creating a PR from the LabKey side to get TeamCity to run the builds.

#### Changes
* Disable the button after the click!


bbimber:
@labkey-jeckels It seems like those sqlserver errors are a config problem?

labkey-adam:
@bbimber It's related to the logging changes... that's just the test query to determine if group_concat is installed or not. The exception stack trace is supposed to be suppressed, but the previous log level check started failing with the log4j 2 upgrade. It should be fixed on develop.

labkey-jeckels:
I just merged the latest from develop, so I expect the next run should be clean

labkey-jeckels:
TeamCity is happy on SQLServer now. I just realized that Cory's offline today. Adam or Ankur, if you can approve, that would be great.

labkey-ankurj (approved):

PR #1475, opened 2020-08-10
Item 7600: Validate query params and filter the suspicious requests

labkey-ankurj:
#### Rationale
This PR add the validations for query params in QueryView form and marks the incorrectly formatted requests as suspicious.


api/src/org/labkey/api/data/WorkbookContainerType.java
```diff
@@ -149,7 +149,7 @@ public Container getContainerFor(DataType dataType, Container currentContainer)
     {
         switch (dataType)
         {
-            //The intent is that outside of these blacklisted actions, return the current container (parent otherwise)
+            //The intent is that outside of these blockListed actions, return the current container (parent otherwise)
```
labkey-jeckels: This seems a little awkward now. What about "outside of these specially supported actions" instead?

api/src/org/labkey/api/query/QuerySettings.java
```diff
@@ -172,7 +178,7 @@ public void setSortFilter(PropertyValues pvs)
             }
             catch (IllegalArgumentException ex)
             {
-                _showRows = ShowRows.PAGINATED;
+                throw new BadRequestException(SC_BAD_REQUEST, "Illegal value for parameter '" + QueryParam.showRows.name() + "'", ex);
```
labkey-jeckels: Let's include the bogus value in the message too.

api/src/org/labkey/api/query/QuerySettings.java
```diff
@@ -226,31 +232,40 @@ public void init(PropertyValues pvs)
...
                 try
                 {
                     long offset = Long.parseLong(offsetParam);
                     if (offset > 0)
                         _offset = offset;
                 }
-                catch (NumberFormatException ignored) { }
+                catch (NumberFormatException nfe)
+                {
+                    throw new BadRequestException(SC_BAD_REQUEST, "Could not parse parameter 'offset'", nfe);
```
labkey-jeckels: Same thing here and a few other spots - include the bogus value in the error message

api/src/org/labkey/api/query/QuerySettings.java
```diff
@@ -226,31 +232,40 @@ public void init(PropertyValues pvs)
...
+            if (isNotBlank(ignoreFilter))
+                _ignoreUserFilter = (Boolean) ConvertUtils.convert(ignoreFilter, Boolean.class);
+
+            String reportId = _getParameter(param(QueryParam.reportId));
+            if (isNotBlank(reportId))
             {
-                _ignoreUserFilter = true;
+                var identifier = ReportService.get().getReportIdentifier(reportId, null, null);
+                if (null == identifier)
+                    throw new NotFoundException();
```
labkey-jeckels: Would be helpful to give a message like "Could not find report for reportId: " + reportId

api/src/org/labkey/api/view/BadRequestException.java
```diff
@@ -29,6 +29,12 @@
     final int status;
     boolean _useBasicAuthentication = false;
 
+    public BadRequestException(int status, String message)
```
labkey-jeckels: We've got a mix of argument orders. Let's standardize on:

message
status
x

since they're strictly additive in that sequence

labkey-jeckels (changes requested):

labkey-jeckels (approved):
This looks good, but the build is unhappy. I didn't look, but I assume another repo or two may need an update for the constructor parameter ordering?

labkey-ankurj:
> This looks good, but the build is unhappy. I didn't look, but I assume another repo or two may need an update for the constructor parameter ordering?

required an update from develop. Passing now.

PR #1484, opened 2020-08-13
Auth Config App: fix delete request success handler scope

labkey-nicka:
#### Rationale
Upon successfully deleting an authentication configuration the callback was incorrectly scoped to call `this.setState` on the component. As a result, the SecondaryAuthenticationTest was failing upon clean-up due to the erroneous dereference.

#### Changes
* Use `() => {}` rather than `function() {}` to ensure component scope.


labkey-tchad (approved):
I'll take your word for it

labkey-nicka:
> I'll take your word for it

Hah! Thanks. I'm running the test in TC to be sure.

**EDIT** Test passed!

PR #1486, opened 2020-08-13
Catch and add BadRequestException message in jsp views

labkey-ankurj:
#### Rationale
This branch catches the BadRequestException in WebPartView returned via JSPs when query or data grids are called with bad query parameters.

### DIFF
--- a/api/src/org/labkey/api/view/WebPartView.java
+++ b/api/src/org/labkey/api/view/WebPartView.java
@@ -387,6 +387,10 @@ public abstract class WebPartView<ModelBean> extends HttpView<ModelBean>
                 {
                     errorHtml = HtmlString.of("Not Found : " + x.getMessage());
                 }
+                catch (BadRequestException x)
+                {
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-matthewb this seems consistent with other error handling, but do you think we should be checking if the response is committed already. If not, we could try to route through ExceptionUtil to send the appropriate error HTTP response code?

labkey-matthewb:
> @labkey-matthewb this seems consistent with other error handling, but do you think we should be checking if the response is committed already. If not, we could try to route through ExceptionUtil to send the appropriate error HTTP response code?

This is a interesting because QuerySettings is used both as an internal configuration object as well as an action form (indirectly via QueryForm).  I think this solution is consistent with our strategy of trying to make sure a broken view can't kill a portal page.  So when used as a "view", we don't want to kill the whole request.  When used directly in executeQuery it returns 400.

labkey-jeckels (approved):

PR #1491, opened 2020-08-14
Issue 41132: XML formatted responses from query API's shouldn't run async

labkey-jeckels:
… and write leading spaces to the stream

#### Rationale
Async queries are good, allowing us to free up database resources if the client disconnects before the query has finished running. But they're no good if we return malformed XML to the request

#### Changes
* Run API queries as async if we're writing back JSON, but not XML

### DIFF
--- a/api/src/org/labkey/api/action/ApiQueryResponse.java
+++ b/api/src/org/labkey/api/action/ApiQueryResponse.java
@@ -85,6 +85,8 @@ public class ApiQueryResponse implements ApiResponse
     private List<FieldKey> _columnFilter;
     private boolean _includeMetaData;
 
+    private ApiResponseWriter.Format _format = ApiResponseWriter.Format.JSON;
+
     // TODO: This is silly... switch to builder pattern, or at least a constructor that takes reasonable strategies
     public ApiQueryResponse(QueryView view, boolean schemaEditable, boolean includeLookupInfo,
                             String schemaName, String queryName, long offset, List<FieldKey> fieldKeys, boolean metaDataOnly,
@@ -114,7 +116,6 @@ public class ApiQueryResponse implements ApiResponse
 
     public ApiQueryResponse()
     {
-        _includeLookupInfo = true;
         _metaDataOnly = true;
     }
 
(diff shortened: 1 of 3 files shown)


labkey-kevink:
It's been a while, but I think it is legal to have whitespace between the <?xml> decl and the document xml element.

labkey-jeckels:
Yes, I imagine we could get fancier on this, but the XML response format is very rarely used, and I think that your change for https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=37051 effectively eliminates it for JSON API responses too.

api/src/org/labkey/api/action/ApiQueryResponse.java
```diff
@@ -114,7 +116,6 @@ public ApiQueryResponse(QueryView view, boolean schemaEditable, boolean includeL
 
     public ApiQueryResponse()
     {
-        _includeLookupInfo = true;
```
labkey-adam: Okay as is, but I have a slight preference for leaving this, making `_includeLookupInfo` final, and eliminating the initialization at declaration time. That would force decisions to be made if additional constructors or builder are implemented. (As the code comment suggests.)

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -3104,6 +3105,7 @@ else if (getRequestedApiVersion() >= 9.1)
                         form.isIncludeDisplayValues());
             }
             response.includeStyle(form.isIncludeStyle());
+            response.setFormat(getResponseFormat());
```
labkey-adam, line 3107: Doesn't matter functionally, of course, but switch order for consistency?

labkey-adam (approved):

api/src/org/labkey/api/action/ApiQueryResponse.java
labkey-jeckels: Swapped back

query/src/org/labkey/query/controllers/QueryController.java
labkey-jeckels: Switched

PR #1496, opened 2020-08-18
41131: Change permission required for CreateNewUserAction & AddUsersAction

labkey-klum:
#### Rationale
We need to support scenarios where a client is able to create and register a custom security role that grants all permissions associated with a project admin but doesn't allow them to create new users.

#### Changes
- Create new permission : AddUserPermission
- Add new permission to the ProjectAdminRole as well as AppAdminRole
- Check for this permission in the CreateNewUserAction and AddUsersAction

### DIFF
--- a/api/src/org/labkey/api/security/roles/ProjectAdminRole.java
+++ b/api/src/org/labkey/api/security/roles/ProjectAdminRole.java
@@ -20,6 +20,7 @@ import org.labkey.api.security.Group;
 import org.labkey.api.security.SecurableResource;
 import org.labkey.api.security.SecurityManager;
 import org.labkey.api.security.SecurityPolicy;
+import org.labkey.api.security.permissions.AddNewUserPermission;
 import org.labkey.api.security.permissions.UserManagementPermission;
 
 import java.util.Collections;
@@ -36,7 +37,7 @@ public class ProjectAdminRole extends AbstractRole
         super("Project Administrator",
             "Project Administrators have full control over the project, but not the entire system.",
             FolderAdminRole.PERMISSIONS,
-            Collections.singletonList(UserManagementPermission.class)
+            Collections.singletonList(AddNewUserPermission.class)
         );
 
         excludeGuests();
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/security/roles/ProjectAdminRole.java
```diff
@@ -20,6 +20,10 @@
 import org.labkey.api.security.SecurableResource;
 import org.labkey.api.security.SecurityManager;
 import org.labkey.api.security.SecurityPolicy;
+import org.labkey.api.security.permissions.AddNewUserPermission;
+import org.labkey.api.security.permissions.UserManagementPermission;
```
cnathe: `UserManagementPermission` isn't actually used here is it? Looks like the new perm added for a ProjectAdminRole is only `AddNewUserPermission`

cnathe (approved):
Looks good to me. In the PR "Changes" list it mentions `Add UserManagementPermission to the ProjectAdminRole` but shouldn't that say `Add AddNewUserPermission to the ProjectAdminRole`?

labkey-adam:
Seems odd to me that `UserManagementPermission` grants the ability to delete, deactivate, and update users, but add requires a separate permission. I know that's specifically what the client asked for, but I can't think of any other scenario where we grant delete but not insert. I would think we'd either separate these into two permissions (AddDeleteUser and UpdateUser) or three permissions (AddUser, DeleteUser, UpdateUser).

labkey-adam:
Also, if this is being requested for merge into 20.7 then the issue milestone and triage fields need to be updated.

api/src/org/labkey/api/security/roles/ProjectAdminRole.java
labkey-klum: done `9c30b8f`

labkey-klum:
@labkey-adam Agreed that having only an add user permission seems strange. I propose that for 20.7 we have the minimal change and just introduce the AddUserPermission. In 20.11 we can introduce DeleteUserPermission and UpdateUserPermission, chase down all usages of UserManagementPermission and switch to the new permissions where appropriate. We can also address the permission checking in CreateNewUserAction to simplify the usage of hasRootPermission

labkey-adam:
Sounds fine. Let's make sure the issue gets reopened for 20.11 to finish this off. Also worth clarifying with the client that those with their special role will be able to delete users, but not create them.

labkey-adam:
Also for 20.11, we could add the listener framework to ensure that their special role gets all the permissions desired, regardless of registration sequence.

labkey-adam (approved):
Okay as is for 20.7.x

PR #1500, opened 2020-08-20
Issue 40668: Convert npm builds to use build-dev by default instead of build-prod for efficiency

labkey-susanh:
#### Rationale
The "build-dev" target that was introduced for all platform modules that have a `package.json` is  a "development" build of the apps in a module. This build uses webpack mode: `development` and devTool: `eval` source-map configuration and does not perform optimizations (e.g. vendor bundling).

To take advantage of this we, we'll switch the default `npm run build` command to use `builid-dev` instead of `build-prod`.   This should improve build times on local development machines.  TeamCity uses `-PdeployMode=prod`, which means it already uses `npm run build-prod` so no changes are required there.  

Version 1.13.0 of the Gradle Plugins introduced the `useNpmProd` property, which can be used if a local build needs to opt into the prod build (though you can also just use `-PdeployMode=prod`).

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1182
* https://github.com/LabKey/gradlePlugin/pull/72

#### Changes
* Switch the npm `build` script to use `build-dev`

### DIFF
--- a/assay/package.json
+++ b/assay/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "setup": "npm ci",
-    "build": "npm run build-prod",
+    "build": "npm run build-dev",
     "start": "bnr build:watch",
     "build-dev": "npm run clean && bnr build:dev",
     "build-prod": "npm run clean && bnr build:prod",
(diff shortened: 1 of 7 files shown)


labkey-nicka (commented):
Let's make the same change in `core/package.json`. Now that we are going to target development build we can switch `build` target to be just like the other packages:

```json
"build": "npm run build-dev",
```

labkey-nicka (approved):

PR #1501, opened 2020-08-20
fix bug in FilteredTable.addCondition()

labkey-matthewb:
#### Rationale
Address automated test failure.
fix bug in FilteredTable.addCondition() where code assumes wrapped table is a SchemaTableInfo

### DIFF
--- a/api/src/org/labkey/api/query/FilteredTable.java
+++ b/api/src/org/labkey/api/query/FilteredTable.java
@@ -16,6 +16,7 @@
 
 package org.labkey.api.query;
 
+import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
Is it possible to add a unit test for this?

Spacing seems messed up - tabs vs spaces maybe?

labkey-matthewb:
> Is it possible to add a unit test for this?
> 
> Spacing seems messed up - tabs vs spaces maybe?

Hmm, not sure why intellij decided to use tabs...  I really don't think this needs a test, getValueSql() is the method used _everywhere_ else, and this code path get hits a lot (just not the one overload of addCondition used with a querytableinfo)

labkey-jeckels (approved):

PR #1502, opened 2020-08-21
Issue 38105: Chart x-axis visit label sorting should be based on sequenceNum not label text

cnathe:
#### Rationale
A client notified us of an issue with the box plot charts x-axis sorting behavior which was sorting based on the visit label instead of the visit sequence number. There was an initial fix for this applied to 20.7 but that fix did not account for custom study subject information being set and was only applied to box plots. This PR fixes the hard coded subject table name issue and also applies the sortFn fix to the bar and scatter plot types.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1291

#### Changes
* Use the study context subject table name instead of hardcoding to "Participant"
* Apply x-axis sortFn changes to bar and scatter plot types in addition to box plots

### DIFF
--- a/visualization/resources/web/vis/chartWizard/genericChartPanel.js
+++ b/visualization/resources/web/vis/chartWizard/genericChartPanel.js
@@ -720,9 +720,10 @@ Ext4.define('LABKEY.ext4.GenericChartPanel', {
         var dataRegion = LABKEY.DataRegions[this.panelDataRegionName];
         var sortKey = 'lsid'; // needed to keep expected ordering for legend data
 
-        // Issue 38105: For box plot of study visit labels, sort by visit display order and then sequenceNum
-        if (this.renderType === 'box_plot' && this.measures.x && this.measures.x.fieldKey === 'ParticipantVisit/Visit') {
-            sortKey = 'ParticipantVisit/Visit/DisplayOrder, ParticipantVisit/SequenceNum';
+        // Issue 38105: For plots with of study visit labels on x-axis, sort by visit display order and then sequenceNum
+        var tableName = LABKEY.vis.GenericChartHelper.getStudySubjectInfo().tableName + 'Visit';
+        if (this.measures.x && this.measures.x.fieldKey === tableName + '/Visit') {
+            sortKey = tableName + '/Visit/DisplayOrder, ' + tableName + '/SequenceNum';
         }
 
         var config = {
--- a/visualization/resources/web/vis/chartWizard/genericChartPanel.js
+++ b/visualization/resources/web/vis/chartWizard/genericChartPanel.js
@@ -720,10 +720,12 @@ Ext4.define('LABKEY.ext4.GenericChartPanel', {
         var dataRegion = LABKEY.DataRegions[this.panelDataRegionName];
         var sortKey = 'lsid'; // needed to keep expected ordering for legend data
 
-        // Issue 38105: For plots with of study visit labels on x-axis, sort by visit display order and then sequenceNum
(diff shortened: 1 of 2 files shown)


visualization/resources/web/vis/chartWizard/genericChartPanel.js
```diff
@@ -720,9 +720,10 @@ Ext4.define('LABKEY.ext4.GenericChartPanel', {
         var dataRegion = LABKEY.DataRegions[this.panelDataRegionName];
         var sortKey = 'lsid'; // needed to keep expected ordering for legend data
 
-        // Issue 38105: For box plot of study visit labels, sort by visit display order and then sequenceNum
-        if (this.renderType === 'box_plot' && this.measures.x && this.measures.x.fieldKey === 'ParticipantVisit/Visit') {
-            sortKey = 'ParticipantVisit/Visit/DisplayOrder, ParticipantVisit/SequenceNum';
+        // Issue 38105: For plots with of study visit labels on x-axis, sort by visit display order and then sequenceNum
```
labkey-susanh:
```suggestion
        // Issue 38105: For plots with study visit labels on x-axis, sort by visit display order and then sequenceNum

```

visualization/resources/web/vis/chartWizard/genericChartPanel.js
```diff
@@ -720,9 +720,10 @@ Ext4.define('LABKEY.ext4.GenericChartPanel', {
         var dataRegion = LABKEY.DataRegions[this.panelDataRegionName];
         var sortKey = 'lsid'; // needed to keep expected ordering for legend data
 
-        // Issue 38105: For box plot of study visit labels, sort by visit display order and then sequenceNum
-        if (this.renderType === 'box_plot' && this.measures.x && this.measures.x.fieldKey === 'ParticipantVisit/Visit') {
-            sortKey = 'ParticipantVisit/Visit/DisplayOrder, ParticipantVisit/SequenceNum';
+        // Issue 38105: For plots with of study visit labels on x-axis, sort by visit display order and then sequenceNum
+        var tableName = LABKEY.vis.GenericChartHelper.getStudySubjectInfo().tableName + 'Visit';
```
labkey-susanh: It seems unfortunate that this very specific logic for a Visit table will sit inside this generic function for getting the queryConfig.  I suppose that was true before this change as well.  Is there nowhere closer to Study that this special logic could be applied?  In any case, I might change the variable to something like `visitTableName`.

And is this intended to no longer cover the `ParticipantVisit/Visit` case or will `LABKEY.vis.GenericChartHelper.getStudySubjectInfo().tableName` return `Participant` in that case?

labkey-susanh (approved):
Though I don't particularly like the special study logic in the otherwise generic function, we had special logic before, so this change doesn't introduce it.

visualization/resources/web/vis/chartWizard/genericChartPanel.js
cnathe: done

visualization/resources/web/vis/chartWizard/genericChartPanel.js
cnathe: Yes, `LABKEY.vis.GenericChartHelper.getStudySubjectInfo().tableName` will return `Participant`. 

I agree with your other comment about this study specific logic being in a generic `getQueryConfig` function. I'm looking into it now to see if this could be passed in as a param or something along those lines.

cnathe: After review, I don't see a great way to separate the "study specificity" part of this check / code. We have a few other study specific parts of this generic chart file, though I'm not surprised as this charting did originate in study land.

labkey-susanh (approved):

PR #1504, opened 2020-08-21
Fix Crawler failures due to BRE

labkey-ankurj:
#### Rationale
This catches BadRequestExceptions thrown in project views by the templates.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1493


core/src/org/labkey/core/view/template/bootstrap/PageTemplate.java
```diff
@@ -291,6 +292,11 @@ public String getProjectTitle()
                     }
                     catch (Exception x)
                     {
+                        if (x instanceof BadRequestException)
+                        {
+                            // re-throw BadRequestException
+                            throw new BadRequestException(x.getMessage(), x);
```
labkey-ankurj: @labkey-jeckels let me know if there is a better approach.

labkey-jeckels: See the changes I just pushed.

core/src/org/labkey/core/portal/ProjectController.java
```diff
@@ -343,9 +345,18 @@ public ModelAndView getView(PageForm form, BindException errors)
...
 
             Template t = isPrint() ? Template.Print : Template.Home;
-            HttpView template = t.getTemplate(getViewContext(), new VBox(), page);
+            try
+            {
+                template = t.getTemplate(getViewContext(), new VBox(), page);
+            }
+            catch (BadRequestException e)
+            {
+                ExceptionUtil.handleException(getViewContext().getRequest(), getViewContext().getResponse(), e, e.getMessage(), false);
```
labkey-jeckels: Why is this required? I'd expect it to get routed to ExceptionUtil via the standard exception handling, like in other actions, without needing this code

labkey-jeckels (changes requested):

core/src/org/labkey/core/portal/ProjectController.java
labkey-ankurj: It wasn't reaching there before when I kept the debug point there or may be I am wrong. I can re-try without having this code.

labkey-ankurj: or I guess before the changes to PageTemplate, we were catching the exception in PageTemplate and logging the error but now we are re-throwing so it would reach there.

labkey-jeckels (approved):

PR #1508, opened 2020-08-22
Issue 41194: Improve test email admin page

labkey-jeckels:
#### Rationale
The test email page is useful, but has some layout and usability problems

#### Changes
* Improve spacing and form element sizing
* Add cancel button

### DIFF
--- a/core/src/org/labkey/core/admin/emailProps.jsp
+++ b/core/src/org/labkey/core/admin/emailProps.jsp
@@ -17,16 +17,22 @@
 %>
 <%@ page import="org.labkey.api.util.MailHelper" %>
 <%@ page import="java.util.Properties" %>
+<%@ page import="java.util.Set" %>
+<%@ page import="org.labkey.api.collections.CaseInsensitiveHashSet" %>
 <%@ page extends="org.labkey.api.jsp.JspBase" %>
 <%
     Properties emailProps = MailHelper.getSession().getProperties();
+    Set<String> obscuredProps = new CaseInsensitiveHashSet(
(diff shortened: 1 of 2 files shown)


core/src/org/labkey/core/admin/emailProps.jsp
```diff
@@ -17,16 +17,22 @@
...
 <%@ page import="org.labkey.api.util.MailHelper" %>
 <%@ page import="java.util.Properties" %>
+<%@ page import="java.util.Set" %>
+<%@ page import="org.labkey.api.collections.CaseInsensitiveHashSet" %>
 <%@ page extends="org.labkey.api.jsp.JspBase" %>
 <%
     Properties emailProps = MailHelper.getSession().getProperties();
+    Set<String> obscuredProps = new CaseInsensitiveHashSet(
+            "mail.smtp.user",
+            "mail.smtp.password"
```
labkey-susanh, line 27: Did you also want to include `mail.smtp.from` here?

labkey-jeckels: I intentionally omitted it here. Since it will be part of the outgoing email itself it doesn't seem confidential. But I wasn't sure exactly what to include/exclude, so very I'm open to suggestions

labkey-susanh: I would have thought the password was sufficient but hiding the user name seems fine too.  Quick glance at other smtp properties doesn't suggest any others that you might want to secure.

labkey-susanh (approved):

PR #1510, opened 2020-08-25
40933: Query dependency report sort order

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40933

#### Changes
- sort by type (query, report, table)
- sort by schema display name
- sort by query or report name

### DIFF
--- a/query/webapp/query/browser/view/QueryDetails.js
+++ b/query/webapp/query/browser/view/QueryDetails.js
@@ -363,7 +363,27 @@ Ext4.define('LABKEY.query.browser.view.QueryDetails', {
     formatDependencies : function () {
 
         const dependencies = this.queriesCache.getDependencies(LABKEY.container.id, this.schemaName, this.queryName);
+
+        // issue : 40993 sort dependencies by type, schemaName and name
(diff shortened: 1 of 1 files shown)


query/webapp/query/browser/view/QueryDetails.js
```diff
@@ -363,7 +363,27 @@ Ext4.define('LABKEY.query.browser.view.QueryDetails', {
...
+            if (schemaName !== 0)
+                return schemaName;
+
+            // query name
+            return a.name.toLowerCase().localeCompare(b.name.toLowerCase());
+        };
+
         if (dependencies){
+            dependencies.dependees.sort(sortFn);
+            dependencies.dependents.sort(sortFn);
```
cnathe, line 385: Unrelated to this set of code changes, but I was looking at this file and the related `webapp/query/browser/Caches.js` file to understand this a bit better and in that Caches.js file, I saw that there are several usages of both spellings: `dependent` and 'dependant`. I assume it isn't affecting functionality, but might be good to get cleaned up.

cnathe (approved):

PR #1512, opened 2020-08-25
Item 7514: Allow users to attach comments for the audit log when inserting, updating, or deleting rows

labkey-susanh:
#### Rationale
When checking items in and out of the freezer, we want to allow users to optionally provide comments on the action.  These comments are naturally associated with the audit event, not with the sample or the inventory item.  They are to be surfaced with the sample timeline and inventory location history views within the applications.

Though I added a constant for the field name `UserComment` to `AbstractAuditTypeProvider`, I did not add this as a standard default field since the field is not useful unless there is a way for a user to provide the comment.  Only the InventoryAuditEvent uses this field at present.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/351
* https://github.com/LabKey/labkey-api-js/pull/73
* https://github.com/LabKey/labkey-ui-components/pull/324
* https://github.com/LabKey/inventory/pull/75

#### Changes
* Add an optional userComment parameter to the AuditHandler method signatures for creating log entries
* Add an AuditUserComment field that can be supplied with an update, insert, or delete rows action.
* Update `QueryController` to look for the field `auditUserComment` in the JSON payload and pass that through the context along with the AuditBehaviorType that may have been provided in the request.

### DIFF
--- a/api/src/org/labkey/api/audit/SampleTimelineAuditEvent.java
+++ b/api/src/org/labkey/api/audit/SampleTimelineAuditEvent.java
@@ -20,8 +20,8 @@ public class SampleTimelineAuditEvent extends DetailedAuditTypeEvent
         MERGE("Sample was registered or updated.", "Registered"),
         UPDATE("Sample was updated.", "Updated");
 
-        private String _comment;
-        private String _actionLabel;
+        private final String _comment;
+        private final String _actionLabel;
 
         SampleTimelineEventType(String comment, String actionLabel)
         {
@@ -75,6 +75,7 @@ public class SampleTimelineAuditEvent extends DetailedAuditTypeEvent
     private boolean _isLineageUpdate;
     private String _metadata;
     private String _inventoryUpdateType;
+    private String _userComment;
 
     public SampleTimelineAuditEvent()
     {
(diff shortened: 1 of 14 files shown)


api/webapp/clientapi/core/Query.js
```diff
@@ -703,8 +704,9 @@ LABKEY.Query = new function()
         * @param {Integer} [config.timeout] The maximum number of milliseconds to allow for this operation before
         *       generating a timeout error (defaults to 30000).
         * @param {boolean} [config.transacted] Whether all of the updates should be done in a single transaction, so they all succeed or all fail. Defaults to true
-         * @param {string} [config.auditBehavior] Audit behavior for this particular action, used to override behavio
+         * @param {string} [config.auditBehavior] Audit behavior for this particular action, used to override behavior
          * set on the underlying table.
+         * @param {string} [config.auditUserComment] Can be used to provide a comment from the user that will be attached to certain detailed audit logs.
```
cnathe:
```suggestion
         * @param {string} [config.auditUserComment] Can be used to provide a comment from the user that will be attached to detailed audit log records.
```

api/src/org/labkey/api/audit/SampleTimelineAuditEvent.java
```diff
@@ -12,6 +12,7 @@
     public static final String EVENT_TYPE = "SampleTimelineEvent";
 
     public static final String SAMPLE_TIMELINE_EVENT_TYPE = "SampleTimelineEventType";
+    public static final String AUDIT_USER_COMMENT = "userComment";
```
cnathe: Could any usages of this `AUDIT_USER_COMMENT` variable just use `COLUMN_NAME_USER_COMMENT` instead? or is that casing difference the issue?

cnathe (approved):

PR #1513, opened 2020-08-26
Fix DataClassSearchTest

labkey-tchad:
#### Rationale
Slight changes to test preamble/postamble have caused the post-test check for this test to fail because it is on the wrong page.

#### Related Pull Requests
* LabKey/testAutomation#431

#### Changes
* Navigate to search page before verifying search results


labkey-tchad:
Verifying fix: https://teamcity.labkey.org/viewQueued.html?itemId=1097944

labkey-chrisj (approved):

PR #1514, opened 2020-08-26
Fix StudyExportTest

labkey-tchad:
#### Rationale
The participant list data region takes a while to load sometimes.

#### Changes
* Increase timeout for 'demoDataRegion' to load.


labkey-tchad:
Verifying fix: https://teamcity.labkey.org/viewQueued.html?itemId=1097942

labkey-chrisj (approved):

PR #1518, opened 2020-08-26
Item 7629: Redesign of Error page

labkey-ankurj:
#### Rationale
This work is to update the current error page for LabKey server. We will address modern UI/UX patterns with error page content. This will improve the user experience in the events of errors by smoother transitions from the error page to a guided LabKey server page.


api/src/org/labkey/api/view/template/PageConfig.java
```diff
@@ -65,7 +65,8 @@
         Wizard,
         Body,
         Custom,  // must be handled by module
-        App;
+        App,
+        Error;
```
labkey-ankurj: note to self - remove Error

core/src/client/entryPoints.js
```diff
@@ -4,15 +4,23 @@
...
+    //     name: 'authenticationConfiguration',
+    //     title: 'Authentication',
+    //     permission: 'admin',
+    //     path: './src/client/AuthenticationConfiguration'
+    // },{
+    //     name: 'components',
+    //     title: '@labkey/components',
+    //     permission: 'admin',
+    //     path: './src/client/LabKeyUIComponentsPage'
+    // },
```
labkey-ankurj: note to self - put these back

core/src/client/ErrorHandler/ErrorType.tsx
```diff
@@ -0,0 +1,101 @@
+import React, { ReactNode } from 'react';
+import { imageURL } from '@labkey/components';
+
+const ERROR_HEADING = 'Oops! An error has occurred.';
+
+const GENERAL_SUBHEADING = <>It seems like something went wrong. The requested page cannot be found.</>;
+const GENERAL_INSTRUCTION = (
+    <>
```
labkey-ankurj, line 8: Cory, I know you told me to just have string constants here but I thought of keeping them like this because of different pieces of text and links in some. Let me know if I should make a component for each of them and pass props or sth else.

cnathe: Yeah, if they have links and other non-text items in them, then leaving them as-is makes sense. It looks like the SUBHEADING constants are just text, so I would likely change those.

api/src/org/labkey/api/util/ExceptionUtil.java
```diff
@@ -844,8 +851,75 @@ else if (isJSON || isXML)
                 log.error("Global.handleException", x);
             }
         }
+        else if (context != null && AppProps.getInstance().isExperimentalFeatureEnabled(AppProps.EXPERIMENTAL_ERROR_PAGE))
+        {
+            // 7629: Error page redesign -- development in-progress. This path shows the new error view.
+            ErrorRenderer renderer = getErrorRenderer(responseStatus, message, ex, request, false, startupFailure);
+            renderer.setErrorType(ErrorRenderer.ErrorType.execution);
```
labkey-jeckels: Similar to how we default to a particular responseStatus in handleException() and then update it for certain kinds of exceptions (NotFoundException sets it to a 404, UnauthorizedException sets it to SC_UNAUTHORIZED or SC_FORBIDDEN), choose the appropriate errorType as we go and then use it down here.

labkey-jeckels (changes requested):

api/src/org/labkey/api/util/ErrorRenderer.java
```diff
@@ -41,6 +41,62 @@
     private final ErrorRendererProperties _errorRendererProps;
     private final String _title;
 
+    // TODO: ErrorPage use these in React app
```
cnathe, line 44: Is this a TODO or just a comment saying that the React component uses them? Just curious what is the TODO here.

core/src/client/ErrorHandler/ErrorHandler.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+interface ErrorHandlerProps {
+    context: AppContext;
+}
+
+interface ErrorHandlerState {
+    showDetails: boolean;
+}
+
+export class ErrorHandler extends React.Component<ErrorHandlerProps, ErrorHandlerState> {
```
cnathe: use React.PureComponent

core/src/client/ErrorHandler/ErrorHandler.tsx
```diff
@@ -0,0 +1,57 @@
...
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            showDetails: false,
+        };
+    }
+
+    loadBack = (): void => {
```
cnathe: this function name confused me, what is the `load` part about? could it be renamed to something like `goBack` or `onBackClick`?

core/src/client/ErrorHandler/ErrorHandler.tsx
```diff
@@ -0,0 +1,57 @@
...
+        window.history.back();
+    };
+
+    loadViewDetails = (): void => {
```
cnathe: same question here about the use of `load` in the function name, should this be named `onViewDetailsClick` or something along those lines?

core/src/client/ErrorHandler/ErrorType.tsx
```diff
@@ -0,0 +1,101 @@
...
+        <div className="labkey-error-instruction">
+            Please contact your admin or reference the <a href="#">LabKey support forum.</a>
+        </div>
+        <div className="labkey-error-instruction">
+            If you would like to file a <a href="#"> LabKey support ticket</a>, your unique reference code is:
```
cnathe, line 15: I assume these hrefs will be filled in with the next story? If so, I'd add a TODO somewhere here about this.

core/src/client/ErrorHandler/ErrorType.tsx
```diff
@@ -0,0 +1,101 @@
...
+const EXECUTION_SUB_HEADING = <>It seems like there is an issue with this installation of LabKey server.</>;
+const EXECUTION_INSTRUCTION = (
+    <>
+        <div className="labkey-error-instruction">
+            Please report this bug to <a href="#"> LabKey Support </a> by copying and pasting both unique reference code
+            and code under 'view details'.
+        </div>
+        <div className="labkey-error-instruction">Your unique reference code is:</div>
+    </>
+);
```
cnathe, line 35: After thinking about what we discussed the other day for these consts and looking at the enum below. What do you think about putting the headings and instructions into a single object? That way they could be accessed with a single line instead of a switch / ifelse. Something like this:
```
const ERROR_TYPE_INFO = {
   general: {heading: ..., instruction: ...},
   permission:  {heading: ..., instruction: ...},
   ...
};

ERROR_TYPE_INFO[errorType].heading
```

core/src/client/ErrorHandler/ErrorType.tsx
```diff
@@ -0,0 +1,101 @@
...
+    permission = 'permission',
+    configuration = 'configuration',
+    execution = 'execution',
+}
+
+export const getErrorHeading = (): ReactNode => {
+    return <div className="labkey-error-heading"> {ERROR_HEADING} </div>;
+};
+
+export const getImage = (errorType: ErrorType): ReactNode => {
```
cnathe, line 71: If we went the route of putting the error type heading and instructions into a single object, I'd put the image string in there as well. Then all of these functions would be much simpler (i.e. `getImage`, `getSubHeading`, `getInstruction `).

```
export const getImage = (errorType: ErrorType): ReactNode => {
   const path = ERROR_TYPE_INFO[errorType].heading; // probably want to add a check to make sure errorType exists in the ERROR_TYPE_INFO
   return <img alt="LabKey Error" src={imageURL('_images', path)} />;
};
```

core/src/client/ErrorHandler/dev.tsx
```diff
@@ -0,0 +1,23 @@
...
+    );
+};
+
+App.registerApp<AppContext>('errorHandler', render, true);
+
+declare const module: any;
+
+if (module.hot) {
+    module.hot.accept();
+}
```
cnathe, line 23: @labkey-nicka Rosaline mentioned in her PR for the module editor case (which is also using the App.registerApp approach), that these lines weren't needed for the HMR to work. I haven't tried it, just an FYI.

core/src/client/components/ErrorTopSection.tsx
```diff
@@ -0,0 +1,39 @@
...
+    loadBack: () => void;
+    loadViewDetails: () => void;
+}
+
+export class ErrorTopSection extends React.PureComponent<ErrorTopSectionProps> {
+    render(): ReactNode {
+        const { errorType, loadBack, loadViewDetails } = this.props;
+        return (
+            <>
+                <Row className="panel-body">
```
cnathe: seems a bit odd to put the `panel-body` class on a `Row`. Should this be wrapped in a `Panel` or div and that class be applied there?

core/src/client/components/ErrorTopSection.tsx
```diff
@@ -0,0 +1,39 @@
...
+                    <Col md={1} />
```
cnathe: I'm curious why the `<Col md={1} />` on the left and right sides? Is that just to make the display area narrower?

core/src/client/entryPoints.js
```diff
@@ -14,5 +14,11 @@ module.exports = {
         title: '@labkey/components',
         permission: 'admin',
         path: './src/client/LabKeyUIComponentsPage'
+    },{
+        name: 'errorHandler',
+        title: 'Error Handler',
+        permission: 'admin',
+        path: './src/client/ErrorHandler',
+        generateLib: true
```
cnathe: could add a quick comment here saying where this lib is used:
```suggestion
        generateLib: true // used in errorView.jsp
```

core/src/client/entryPoints.js
```diff
@@ -14,5 +14,11 @@ module.exports = {
         title: '@labkey/components',
         permission: 'admin',
         path: './src/client/LabKeyUIComponentsPage'
+    },{
+        name: 'errorHandler',
+        title: 'Error Handler',
+        permission: 'admin',
```
cnathe: I assume this permission isn't actually used in the `generateLib` case and can be removed. That permission is used in the views/*.xml files but this error page will have permissions controlled elsewhere.

cnathe (changes requested):
I reviewed the client side pieces of this PR. I'll defer on the server side stuff to Josh/Nick.

api/src/org/labkey/api/util/ErrorRenderer.java
labkey-ankurj: That's a TODO, I am keeping the TODOs for this story consistent with format // TODO: ErrorPage so that they'd be searchable in next story. The todo here is to use/ remove these booleans from the old error view into new error page.

core/src/client/ErrorHandler/ErrorHandler.tsx
labkey-ankurj: sure

core/src/client/ErrorHandler/ErrorHandler.tsx
labkey-ankurj: sure

core/src/client/ErrorHandler/ErrorType.tsx
labkey-ankurj: Yes, I think this looks cleaner.

core/src/client/ErrorHandler/ErrorType.tsx
labkey-ankurj: done

core/src/client/ErrorHandler/ErrorType.tsx
labkey-ankurj: added

core/src/client/components/ErrorTopSection.tsx
labkey-ankurj: yes, done.

core/src/client/components/ErrorTopSection.tsx
labkey-ankurj: I guess I added because invision has a wider view, but after removing this it looks better. Thanks.

core/src/client/entryPoints.js
labkey-ankurj: removed

core/src/client/entryPoints.js
labkey-ankurj: added

cnathe (approved):
Approving React / client side related code changes.

api/src/org/labkey/api/util/ExceptionUtil.java
```diff
@@ -844,8 +860,81 @@ else if (isJSON || isXML)
...
+                {
+                    // TODO : ErrorPage, this app template doesn't work
+                    errorView = PageConfig.Template.App.getTemplate(context, new ErrorTemplate(renderer), pageConfig);
+                    pageConfig.addClientDependencies(errorView.getClientDependencies());
+                    errorView.getView().render(errorView.getModel(), request, response);
+
+                }
+                catch (Exception exc)
+                {
+                    log.error("Global.handleException", e);
```
labkey-jeckels: Seems like we should log exc too

api/src/org/labkey/api/util/ExceptionUtil.java
```diff
@@ -844,8 +860,81 @@ else if (isJSON || isXML)
                 log.error("Global.handleException", x);
             }
         }
+        else if (AppProps.getInstance().isExperimentalFeatureEnabled(AppProps.EXPERIMENTAL_ERROR_PAGE))
```
labkey-jeckels, line 863: This method is already ridiculously long. Can you pull this out into a separate method that's called from here?

api/src/org/labkey/api/util/ExceptionUtil.java
```diff
@@ -595,14 +597,19 @@ public static boolean isClientAbortException(Throwable ex)
...
+
+    // This is called by SpringActionController (to display unhandled exceptions) and called directly by AuthFilter.doFilter() (to display startup errors and bypass normal request handling)
+    public static ActionURL handleException(@NotNull HttpServletRequest request, @NotNull HttpServletResponse response, Throwable ex, @Nullable String message, boolean startupFailure, @Nullable ViewContext context, @Nullable PageConfig pageConfig)
+    {
+        return handleException(request, response, ex, message, startupFailure, SearchService.get(), LOG, context, pageConfig);
     }
 
     static ActionURL handleException(@NotNull HttpServletRequest request, @NotNull HttpServletResponse response, Throwable ex, @Nullable String message, boolean startupFailure,
-        SearchService ss, Logger log)
+        SearchService ss, Logger log, @Nullable ViewContext context, @Nullable PageConfig pageConfig)
```
labkey-jeckels: It looks like pageConfig is only used to get the template, correct? If so, consider switching to take that as an argument since I'm not sure we want to carry all of the other config into the error page.

Also, did you see if you could call HttpView.hasCurrentView() and then HttpView.currentContext() instead of passing ViewContext here?

labkey-jeckels (changes requested):

api/src/org/labkey/api/util/ExceptionUtil.java
labkey-ankurj: PageConfig is used to get template and also to create the template. I think it is required here. 

On the ViewContext part, yes I am able to use HttpView.currentContext().

api/src/org/labkey/api/util/ExceptionUtil.java
labkey-ankurj: sure

api/src/org/labkey/api/util/ExceptionUtil.java
labkey-ankurj: OK

labkey-jeckels (approved):

PR #1525, opened 2020-08-28
Issue 40302: Unable to use samples or data class with integer like names as material or data input

labkey-kevink:
#### Rationale
The magic lineage columns used in DataClass and SampleType import may contain rowId values or string values for looking up a data or material by alternate key (typically by name.)  There were two issues with resolving:

* When using the bulk tsv import method, you would receive a "Expected comma separated list or a JSONArray of parent names" error.  Since the magic columns are not known columns on the table, TabLoader would infer types of the magic columns by attempting to parse the value.   If all values are numeric, the TabLoader would parse the values as Integer while ExpDataIterator.DerivationDataIterator expects a String value or JSONArray.

* If the String can be parsed as a number, the ExperimentService.findExpMaterial() method would attempt to parse the value as an Integer to find the material by row id and return null if not found.  If the value doesn't parse as an Integer, we attempt to find by alternate key.  If the material has a numeric name, we wouldn't resolve by name.

#### Changes
* resolve lineage parent by numeric name if resolving by rowId fails
* force TabLoader to load lineage columns as strings

### DIFF
--- a/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
+++ b/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
@@ -990,7 +990,7 @@ public abstract class AbstractAssayTsvDataHandler extends AbstractExperimentData
                 {
                     String ssName = byNameSS != null ? byNameSS.getName() : null;
                     Container lookupContainer = byNameSS != null ? byNameSS.getContainer() : container;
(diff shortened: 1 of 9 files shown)


labkey-jeckels (approved):
I worry about ambiguous numeric valued that could resolve to either a RowId or a name, but impossible to support both and always do the right thing.

labkey-nicka (approved):

PR #1526, opened 2020-08-28
Don't double-log exceptions in BaseApiAction and ExceptionUtil

labkey-jeckels:
#### Rationale
Keep refining our logging and HTTP response codes for various kinds of bogus requests

#### Changes
* Use common base class for exceptions that should relay a specific HTTP status code to the client

* Avoid useless null check


api/src/org/labkey/api/view/NotFoundException.java
```diff
@@ -23,20 +24,20 @@
...
-public class NotFoundException extends RuntimeException implements SkipMothershipLogging
+public class NotFoundException extends HttpStatusException
 {
     public NotFoundException()
     {
-        super("" + HttpServletResponse.SC_NOT_FOUND + ": page not found");
+        this(null);
     }
 
     public NotFoundException(String string)
```
labkey-ankurj: seems like String argument can be marked as @Nullable

api/src/org/labkey/api/view/NotFoundException.java
```diff
@@ -23,20 +24,20 @@
...
     {
-        super(string);
+        this(string, null);
     }
     
-    public NotFoundException(String string, Throwable cause)
+    public NotFoundException(String string, @Nullable Throwable cause)
```
labkey-ankurj: here as well.

query/src/org/labkey/query/reports/ReportsController.java
```diff
@@ -877,37 +877,31 @@ public ApiResponse execute(BEAN bean, BindException errors) throws Exception
...
-                LinkedHashSet<ClientDependency> dependencies = resultsView.getClientDependencies();
-                LinkedHashSet<String> cssScripts = new LinkedHashSet<>();
-                addScriptDependencies(bean, dependencies, cssScripts);
+            Map<String, Object> resultProperties = new HashMap<>();
 
-                LinkedHashSet<String> includes = new LinkedHashSet<>();
-                LinkedHashSet<String> implicitIncludes = new LinkedHashSet<>();
-                PageFlowUtil.getJavaScriptFiles(getContainer(), dependencies, includes, implicitIncludes);
+            LinkedHashSet<ClientDependency> dependencies = resultsView.getClientDependencies();
+            LinkedHashSet<String> cssScripts = new LinkedHashSet<>();
```
labkey-ankurj, line 883: use interface type (Set) for the reference?

query/src/org/labkey/query/reports/ReportsController.java
```diff
@@ -877,37 +877,31 @@ public ApiResponse execute(BEAN bean, BindException errors) throws Exception
...
+            addScriptDependencies(bean, dependencies, cssScripts);
 
-                MockHttpServletResponse mr = new MockHttpServletResponse();
-                mr.setCharacterEncoding(StringUtilsLabKey.DEFAULT_CHARSET.displayName());
-                resultsView.render(getViewContext().getRequest(), mr);
+            LinkedHashSet<String> includes = new LinkedHashSet<>();
+            LinkedHashSet<String> implicitIncludes = new LinkedHashSet<>();
```
labkey-ankurj, line 887: use interface type (Set) for the reference?

labkey-ankurj (changes requested):

api/src/org/labkey/api/view/NotFoundException.java
labkey-jeckels: True, and this also got mangled in a merge so it lost the default message. Now restored and annotated

query/src/org/labkey/query/reports/ReportsController.java
labkey-jeckels: Similar - getJavaScriptFiles() needs a LinkedHashSet for its arguments

query/src/org/labkey/query/reports/ReportsController.java
labkey-jeckels: addScriptDependencies() takes a LinkedHashSet since it cares about ordering, so this needs to be declared to match

labkey-ankurj (approved):

PR #1529, opened 2020-08-30
Issue 40420: remove first delete folder page

labkey-ankurj:
#### Rationale
When you delete a folder (via admin > folder > management), the first page reads "You are about to delete the following folder: " but doesn't say which one. Nothing is listed, just the message that there are subfolders. The next page shows all the details. This PR is cutting out the middle part and directly going to the final confirmation page.

#### Changes
* Remove first delete confirmation view.
* Change Button wording from 'Delete' -> 'Yes, delete all'


labkey-jeckels (approved):
Code looks OK but this could certainly benefit from testing of the various permutations. And will there be any test harness updates needed now that we don't show the intermediate confirmation page?

labkey-ankurj:
I will check with the test team.

PR #1531, opened 2020-08-31
7627: Module Editor Story 3

RosalineP:
#### Rationale
We surface links to the new Module Editor page in the UI.

#### Related Pull Requests
* https://github.com/LabKey/moduleEditor/pull/9

#### Changes
* Adds module editor link in Developer Links dropdown
* Adds module editor link from Known Modules page
* Resolves ModuleEditor permissions-related TODO


cnathe:
@RosalineP should the PR name be referring to "Story 3" instead of 2?

api/src/org/labkey/api/reports/report/ModuleReportDescriptor.java
```diff
@@ -51,8 +51,7 @@ default boolean canEdit(User user, List<ValidationError> errors)
         if (null == ModuleEditorService.get().getFileForModuleResource(getModule(), getSourceFile().getPath()))
             errors.add(new SimpleValidationError("The source for this module report is not editable."));
         else if (!user.hasRootPermission(EditModuleResourcesPermission .class))
-            // TODO add role name here instead of permission
-            errors.add(new SimpleValidationError("You must have EditModuleResourcesPermission to edit module resources."));
+            errors.add(new SimpleValidationError("You must possess ModuleEditorRole to edit module resources."));
```
cnathe: I believe we usually leave the perm/role names out of these types of error messages and say something like "You do not have permissions to edit module resources on this server."

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -8065,6 +8065,12 @@ private ModulesView(Collection<ModuleContext> contexts, String type, HtmlString
                 setTitle(type + " Modules");
             }
 
+            private Link.LinkBuilder getModuleEditorLink(ModuleContext moduleContext)
+            {
+                String linkText = AppProps.getInstance().isDevMode() ? "Edit module" : "Update module";
```
cnathe: It looks like the only place that is calling this `getModuleEditorLink` method is already doing the `isDevMode()` check so I think we can just set the linkText to "Edit module" here.

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -8168,6 +8176,12 @@ private ActionURL getUpdateURL(String name)
             return url;
         }
 
+        private ActionURL getModuleEditorURL(String name)
+        {
+            ActionURL url = ModuleEditorService.get().getModuleEditorURL(name);
+            return url;
```
cnathe:
```suggestion
            return ModuleEditorService.get().getModuleEditorURL(name);
```

cnathe (changes requested):

core/src/org/labkey/core/admin/AdminController.java
RosalineP: Oop. Thank you for catching—the text is supposed to say 'Edit module' if LK is in dev mode, and 'Update module' otherwise, as per the spec. I've updated the code accordingly in the last push.

cnathe: So is the link target going to go to a different page / action than in the non-dev mode case? Or is it just the text of the link that changes? I assume the former, but from the code in this PR it looks like both will still just go to the module editor URL. Should the code that was removed below for the "Update Module" case be called to get the `getUpdateURL()`?

cnathe (changes requested):
Still one question about the `Edit module` vs `Update module case`.

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -8092,7 +8092,8 @@ protected void renderView(Object model, PrintWriter out)
                                 TD(cl("labkey-column-header"),"Class"),
                                 TD(cl("labkey-column-header"),"Location"),
                                 TD(cl("labkey-column-header"),"Schemas"),
-                                null == externalModulesDir ? null : TD(cl("labkey-column-header"),""),    // update actions
+                                !AppProps.getInstance().isDevMode() ? null : TD(cl("labkey-column-header"),""),    // edit actions
```
RosalineP, line 8095: This was reordered because it looks peculiar when there are empty columns on either side of a generally populated column

cnathe (approved):

PR #1532, opened 2020-08-31
41280: Expose new permissions : DeleteUserPermission and UpdateUserPermission

labkey-klum:
#### Rationale
As a follow on to this [change](https://github.com/LabKey/platform/pull/1496), for completeness sake we build on the AddUserPermission and introduce
- DeleteUserPermission
- UpdateUserPermission

In addition, to help with external development of admin security roles, we've introduced a listener to allow the automatic addition of registered permissions to admin roles.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1496

#### Changes
- add new permissions
- update instances of UserManagementPermission to check for either DeleteUserPermission or UpdateUserPermission where appropriate
- assign new permissions to AppAdminRole
- add AdminRoleListener and RoleManager.addAdminRoleListener


api/src/org/labkey/api/security/AuthenticationManager.java
```diff
@@ -495,7 +496,7 @@ public static boolean isAccountExpirationEnabled()
      */
     public static boolean canSetUserExpirationDate(User userManager, Container container)
     {
-        boolean isUserManager = userManager.hasRootPermission(UserManagementPermission.class);
+        boolean isUserManager = userManager.hasRootPermission(UpdateUserPermission.class);
```
cnathe: Very minor, but since we aren't using `UserManagementPermission` here anymore, might be worth changing the variable names: `userMangaer` -> `user`, `isUserManager` -> `canUpdateUser`

core/src/org/labkey/core/security/SecurityApiActions.java
```diff
@@ -2131,7 +2131,7 @@ else if (c.isRoot() && !getUser().hasRootPermission(AddUserPermission.class))
     /**
      * Invalidate existing password and send new password link
      */
-    @RequiresPermission(UserManagementPermission.class)
+    @RequiresPermission(UpdateUserPermission.class)
```
cnathe, line 2134: Looking at the `SecurityApiActions.TestCase` it looks like there are some asserts that will need to be updated for the `@RequiresPermission(UserManagementPermission.class)` changes.

core/src/org/labkey/core/security/SecurityController.java
```diff
@@ -1587,7 +1588,7 @@ protected SecurityMessage createMessage(EmailForm form)
     /**
      * Invalidate existing password and send new password link
      */
-    @RequiresPermission(UserManagementPermission.class)
+    @RequiresPermission(UpdateUserPermission.class)
```
cnathe, line 1591: Just above this action, there is another related action `ShowResetEmailAction`, should the requires permission check there be updated to use `UpdateUserPermission` as well?

core/src/org/labkey/core/user/UserController.java
```diff
@@ -1944,7 +1964,7 @@ else if(form.getIsPasswordPrompt() || _isPasswordPrompt)
         @Override
         public ActionURL getSuccessURL(UserForm form)
         {
-            boolean isUserManager = getUser().hasRootPermission(UserManagementPermission.class);
+            boolean isUserManager = getUser().hasRootPermission(UpdateUserPermission.class);
```
cnathe: Another minor case where it might be worth changing the variable name to `canUpdateUser`

core/src/org/labkey/core/user/UserController.java
```diff
@@ -591,7 +611,7 @@ public void setDelete(boolean delete)
         }
     }
 
-    @RequiresPermission(UserManagementPermission.class)
+    @RequiresPermission(DeleteUserPermission.class)
     public class DeleteUsersAction extends FormViewAction<UserIdForm>
```
cnathe, line 616: There is another action in this controller which I think should be updated: `UpdateUsersStateApiAction`. It does both de/reactive changes and deletions, so should likely be changed to something like `@RequiresAllOf({UpdateUserPermission.class, DeleteUserPermission.class})`

core/src/org/labkey/core/user/changeEmail.jsp
```diff
@@ -35,7 +36,7 @@
     JspView<UserController.UserForm> me = (JspView<UserController.UserForm>) HttpView.currentView();
     UserController.UserForm form = me.getModelBean();
     ActionURL cancelURL = new ActionURL(UserController.DetailsAction.class, getContainer()).addParameter("userId", form.getUserId());
-    boolean isUserManager = getUser().hasRootPermission(UserManagementPermission.class);
+    boolean isUserManager = getUser().hasRootPermission(UpdateUserPermission.class);
```
cnathe: change `isUserManager` variable name?

core/src/org/labkey/core/user/changeEmail.jsp
```diff
@@ -23,6 +23,7 @@
 <%@ page import="org.labkey.core.login.LoginController" %>
 <%@ page import="org.labkey.core.user.UserController" %>
 <%@ page import="org.labkey.core.user.UserController.ChangeEmailAction" %>
+<%@ page import="org.labkey.api.security.permissions.UpdateUserPermission" %>
```
cnathe, line 26: can the `UserManagementPermission` import above be removed?

cnathe (changes requested):
In addition to comments added to the changed files, I found a user in UsersTable.java which we might want to update to use the specific add / delete users perm checks:

```
        if (schema.getUser().hasRootPermission(UserManagementPermission.class))
        {
            setDeleteURL(new DetailsURL(new ActionURL(UserController.DeleteUsersAction.class, schema.getContainer())));
            setInsertURL(new DetailsURL(new ActionURL(SecurityController.AddUsersAction.class, schema.getContainer())));
        }
```

Also, the `PipelineController.java` and `ShowGroupMambersAction.java` classes both look to have a `UserManagementPermissions` check that looks to be related to an add user case.

labkey-adam:
It's possible (though not likely, honestly) that the old permission (UserManagementPermission) is serialized for webpart permissions in the database and/or folder archives. If you want, you can address this by implementing getSerializationAliases(), e.g., on UpdateUserPermission. See DesignSampleTypePermission as a random example of this.

core/src/org/labkey/core/security/SecurityApiActions.java
labkey-klum: I agree that the method name is no longer accurate and could be changed to something like : assertForUserPermissions but the test still should pass since application admins will have all of the new permissions. Were you noticing something different?

cnathe: Ah yes, that's right. I forgot how those assert checks worked. Looking at it closer, that makes sense.

cnathe (approved):

PR #1537, opened 2020-09-01
Rethrow HttpStatusExceptions while building menus in a template to get handled in ExceptionUtil.handleException

labkey-ankurj:
#### Rationale
This re-throws HttpStatusExceptions thrown in project views by the page template.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1493


core/src/org/labkey/core/view/template/bootstrap/PageTemplate.java
```diff
@@ -287,9 +288,9 @@ public String getProjectTitle()
                             }
                         }
                     }
-                    catch (BadRequestException x)
+                    catch (BadRequestException | NotFoundException x)
```
labkey-jeckels: I suggest using the new HttpStatusException that I introduced yesterday instead of listing multiple here

labkey-jeckels (changes requested):

labkey-jeckels (approved):

PR #1538, opened 2020-09-01
Fix 41171

labkey-alan:
#### Rationale
The GridPanelPage was re-using QueryModel ids for every model, which was causing test failures due to us using the model id as the selectionKey for selections.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/454
* https://github.com/LabKey/labkey-ui-components/pull/332

#### Changes
* Add a unique id for grids in the GridPanelPage


core/src/client/LabKeyUIComponentsPage/GridPanelPage.tsx
```diff
@@ -34,27 +32,19 @@ class GridPanelExampleImpl extends PureComponent<{} & InjectedQueryModels, State
...
     };
 
     render() {
-        const { queryModels, actions } = this.props;
-        const { model } = queryModels;
-        const { schemaName, queryName } = this.state;
+        const { id, queryName, schemaName, schemaQuery } = this.state;
         let body = (
             <div>
                 Enter a Schema, Query, View
```
labkey-nicka: nit: It doesn't look like "View" is supported. Probably shouldn't mention setting it here.

core/src/client/LabKeyUIComponentsPage/GridPanelPage.tsx
```diff
@@ -34,27 +32,19 @@ class GridPanelExampleImpl extends PureComponent<{} & InjectedQueryModels, State
...
         }
 
-        const schemaQuery = SchemaQuery.create(schemaName, queryName);
-
-        if (model !== undefined) {
-            actions.setSchemaQuery(model.id, schemaQuery);
-        } else {
-            actions.addModel({ schemaQuery, id: 'model' }, true);
-        }
+        this.setState({ id: `gpe-${schemaName}-${queryName}`, schemaQuery: SchemaQuery.create(schemaName, queryName) });
```
labkey-nicka: nit: Could consider storing the `queryConfig` on state rather than the separate properties. That way the component won't necessarily re-render every time.

labkey-nicka (approved):
Approved with only nits.

PR #1544, opened 2020-09-02
fb_parse_warnings

labkey-matthewb:
#### Rationale
Allow parse warnings to be generated by the parser (SqlBase.g)


query/src/org/labkey/query/sql/SqlBase.g
```diff
@@ -103,6 +104,31 @@ tokens
         _annotations = null;
         return ret;
     }
+
+    //protected LinkedHashSet<QueryParseWarning> _warnings = new LinkedHashSet<>();
```
labkey-adam: Delete this?

labkey-adam (changes requested):

query/src/org/labkey/query/sql/SqlBase.g
labkey-matthewb: thanks

labkey-adam (approved):

PR #1547, opened 2020-09-04
CompareType: support Long format dates

labkey-nicka:
#### Rationale
Adds support for filtering based on the `Long` value of date when specified for a date-based filter (e.g. `DATE_LESS_THAN`). Since there is already expected filtering behavior for strings and numbers as strings this introduces a supported suffix ("L" or "l") to avoid backwards compatibility breaks.

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/667

#### Changes
* Support `Long` formatted date values.
* Add unit tests for `CompareType.asDate()`.

### DIFF
--- a/api/src/org/labkey/api/data/CompareType.java
+++ b/api/src/org/labkey/api/data/CompareType.java
@@ -1465,13 +1465,26 @@ public abstract class CompareType
                 return ((Calendar) v).getTime();
             String s = v.toString();
 
+            // Support Long notation for dates
+            if (s.endsWith("L") || s.endsWith("l"))
+            {
+                String couldBeLong = s.substring(0, s.length() - 1);
+
+                // Only if rest of the value is numeric is it considered a valid Long
+                if (NumberUtils.isDigits(couldBeLong))
+                    return new Date(Long.parseLong(couldBeLong));
+            }
+
             if (!s.startsWith("-") && !s.startsWith("+"))
                 return new Date(DateUtil.parseDateTime(s));
 
             boolean add = s.startsWith("+");
             s = s.substring(1);
+
+            // Default to number representing number of "days"
             if (NumberUtils.isDigits(s))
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/CompareType.java
```diff
@@ -2180,6 +2192,55 @@ public static SQLFragment getMemberOfSQL(SqlDialect dialect, SQLFragment userIdS
...
+            try
+            {
+                asDate(invalidNumber);
+            }
+            catch (ConversionException e)
+            {
+                threwExpectedException = true;
+            }
+
+            assertTrue("Expected numeric value to throw ConversionException", threwExpectedException);
```
labkey-jeckels, line 2241: Slightly easier to just fall fail() inside of the try block after asDate() and treat the exception as ignored

api/src/org/labkey/api/data/CompareType.java
```diff
@@ -2180,6 +2192,55 @@ public static SQLFragment getMemberOfSQL(SqlDialect dialect, SQLFragment userIdS
...
+
+            // Formatting causes rounding of date (e.g. loss of context like minutes, seconds, etc)
+            // Rehydrate a time from the Date parsed from the format.
+            long dateFromFormat = DateUtil.parseDateTime(ContainerManager.getRoot(), formattedDateStr);
+
+            assertEquals(dateFromFormat, asDate(formattedDateStr).getTime());
+
+            // Long value
+            String longNow = dateNow.getTime() + "L";
+            assertEquals(dateNow.getTime(), asDate(longNow).getTime());
```
labkey-jeckels, line 2227: Consider adding an expected failure if the provided number exceeds Long's maximum value

labkey-jeckels (approved):
A couple of ideas for the unit test, but change looks good.

PR #1549, opened 2020-09-06
Finalize JSP print() work

labkey-adam:
#### Rationale
Miscellaneous cleanup now that `print(String)` and unsafe `print(Object)` invocations have been removed from all JSP files.

#### Related Pull Requests
* Many

#### Changes
* Seems confusing that methods producing JavaScript content would return `HtmlString`, so introduce `SafeToRenderBuilder` and use it to render content that intermingles HTML and JavaScript (e.g., `PageFlowUtil.getStandardIncludes()`).
* Add a copy of `JspWriter` source from Tomcat to assist with future static analysis of JSP `print(Object)` invocations
* Add `JspBase.q(URLHelper)` and use it
* Stop calling `h()` on primitives like int, long, et al
* Remove pointless ternary operators (`x != null ? x : null`) from JSPs
* Use `urlFor()` instead of `buildURL()`, hard-coded strings, and `new ActionURL()`
* Remove some dodgy String-based URL manipulation


experiment/src/org/labkey/experiment/ProtocolSteps.jsp
```diff
@@ -218,7 +215,7 @@
                     </td>
                     <td width="200px"><%=h(mpo.getLSID())%></td>
                     <td width="80px"><%=h(criteria != null ? criteria.getTypeName() : null)%></td>
-                    <td width="20px"><%=h(mpo.getMinOccurs())%></td>
+                    <td width="20px"><%=mpo.getMinOccurs()%></td>
                     <td width="20px"><%=h(mpo.getMaxOccurs())%></td>
```
labkey-klum, line 219: don't encode this int?

labkey-adam: `getMinOccurs()` is an int, but `getMaxOccurs()` is an Integer. `h()` treats nulls differently than `JspWriter.out(Object)`, so I went conservative here. I only "unencode" the primitives.

api/src/org/labkey/api/jsp/JspBase.java
```diff
@@ -706,18 +700,20 @@ public String getMethod()
...
-    // Provides a unique integer within the context of this request.  Handy for generating element ids, etc. See UniqueID for caveats and warnings.
+    // Provides a unique integer within the context of this request. Handy for generating element ids, etc. See UniqueID for caveats and warnings.
     protected int getRequestScopedUID()
     {
         return UniqueID.getRequestScopedUID(getViewContext().getRequest());
     }
 
     /** simple link to different action in same container w/no parameters */
+    @Deprecated // Eliminate usages and delete
     protected String buildURL(Class<? extends Controller> actionClass)
```
labkey-klum, line 717: Wasn't this one of the points of these PRs? Are there still some lingering usages of buildURL or are we just being extra careful.

labkey-klum (approved):

api/src/org/labkey/api/jsp/JspBase.java
labkey-adam: I tried removing the method, but concept.jsp still uses it. Matt is moving this JSP from search to ontology... but right now there are copies in both places. I didn't want to bother fixing both copies, so this keeps us building until Matt kills the search version. I'll clean this up in a follow-on PR.

PR #1557, opened 2020-09-11
use consistent "now" value for repeatability

labkey-matthewb:
#### Rationale
Make CompareType test case more reliable


labkey-matthewb:
To CR my one PR.  This still doesn't work because  asDate(daysInFuture) and asDate(daysInPast) both still do an implicit currentTimeMillis().

labkey-nicka:
@labkey-matthewb Happy to discuss this further. I understand wanting a consistent value of for "now" across the test cases, however, I'm not sure how it makes it more repeatable.

> To CR my one PR. This still doesn't work because asDate(daysInFuture) and asDate(daysInPast) both still do an implicit currentTimeMillis().

Yes, the test for "days in future/past" is really just to ensure that the number is successfully translated into days. I guess we could create a mock for `System.currentTimeMillis()` and then have a repeatable value.

labkey-matthewb:
The problem is that the value returned from System.currentTimeMillis() can tick-over while this test is running causing some of the equals() tests to fail.  A better attempt than my first try might be to do something like abs(a-b)<2 or something like that.

labkey-nicka (approved):

PR #1561, opened 2020-09-14
Issue 41150: ETL - Columns with spaces in name are being nulled out when not in src query 

labkey-martyp:
#### Rationale
Columns with spaces in name are getting nulled out if not in source query of an ETL.  Currently comparing sql selector with column names when determining what columns not to update.  Need to compare names to names.

#### Changes
Update StatementUtils.createStatement, Base Table Insert section.  Store col and value sql fragments in map with column name as key so comparison of column names can be done when skipping updates to columns in _dontUpdateColumnNames.


api/src/org/labkey/api/data/StatementUtils.java
```diff
@@ -659,47 +660,47 @@ public ParameterMapStatement createStatement(Connection conn, @Nullable Containe
         // BASE TABLE INSERT()
         //
 
-        List<SQLFragment> cols = new ArrayList<>();
-        List<SQLFragment> values = new ArrayList<>();
+        Map<String, SQLFragment> cols = new HashMap<>();
+        Map<String, SQLFragment> values = new HashMap<>();
```
labkey-matthewb: It's not a good assumption that cols.values() and values.values() will return items in the same order.  I would consider changing "List<SQLFragment> cols"  to "List<ColumnInfo> cols" instead.  You could use LinkedHashMap<> but I don't really see the point in using the Map<>.

labkey-matthewb (changes requested):

api/src/org/labkey/api/data/StatementUtils.java
labkey-martyp: @labkey-matthewb 
I wasn't thinking order would matter here but I have seen pretty small changes in order make big changes in performance on queries, so that makes sense.

What is the suggestion here? "I would consider changing "List cols" to "List cols" instead."

Really the whole point of the map is just because we need the column name (not SQLFragment) to see if it exists in _dontUpdateColumnNames (line 840).  I could make a third list to hold column names but thought converting cols and values to maps would be a little more concrete than relying on indexes aligning across three lists.

labkey-matthewb: Sorry, apparently angle brakets means something in markup (see comment source).  That was List of ColumnInfo instead of List of SQLFragment.  It's not a performance thing, but the order of column in the INSERT clause must match the order of columns in the SELECT clause.  I don't think you're really using the Map functionality you're just tracking both the "name" and the "selectnam", you could just carry around the ColumnInfo which also tracks the "name" and "selectname".

labkey-martyp: Got it. Good call. Updated to use List<ColumnInfo>

api/src/org/labkey/api/data/StatementUtils.java
```diff
@@ -835,12 +836,13 @@ else if (column.getName().equalsIgnoreCase(updatable.getObjectURIColumnName()) &
             int updateCount = 0;
             for (int i = 0; i < cols.size(); i++)
             {
-                FieldKey fk = new FieldKey(null, cols.get(i).getSQL());
-                if (keys.containsKey(fk) || null != _dontUpdateColumnNames && _dontUpdateColumnNames.contains(fk.getName()))
+                SQLFragment colSql = new SQLFragment(cols.get(i).getSelectName());
+                FieldKey fk = new FieldKey(null, colSql.getSQL());
```
labkey-matthewb: I think this can now be `fk = col.getFieldKey()` which is much better than the original code!

labkey-matthewb (changes requested):

labkey-matthewb (approved):
Nice cleanup, thanks Marty.

PR #1563, opened 2020-09-14
Item 7840: @labkey/components package update and switch build to using scss files instead of components.css

cnathe:
#### Rationale
Issue [40839](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40839): Styling on audit log grids in SM and FM apps is different in production build vs. hot-reloading build

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/342
* https://github.com/LabKey/biologics/pull/681
* https://github.com/LabKey/moduleEditor/pull/11
* https://github.com/LabKey/platform/pull/1563
* https://github.com/LabKey/provenance/pull/37
* https://github.com/LabKey/sampleManagement/pull/364
* https://github.com/LabKey/inventory/pull/91

#### Changes
* Switch to using @labkey/components scss files instead of components.css generated files


labkey-susanh (approved):
As we discussed, when looking to make the webpack stuff accessible across all modules, we should also look into making a common file that includes the import of the ui-components scss so we might get rid of many of these one-line scss files.

PR #1572, opened 2020-09-17
Fix crawler error in mothership data

labkey-jeckels:
#### Rationale
If you go to /home/mothership-begin.view in the /Home project and change the folder filter to All Folders, you can see data from /_mothership. Clicking on a link to an exception will give you a 404 because it's targeting the current container instead of row's owning container

#### Changes
* Switch to fancier URL generation approach already in use for the details URL


labkey-jeckels:
FWIW, not related to any of the other recent mothership changes. And not a real-world usage scenario. But worth fixing to keep the crawler happy!

labkey-ankurj (approved):

labkey-danield (approved):

PR #1574, opened 2020-09-17
Declare api and implementation dependencies to better align with code usages

labkey-susanh:
#### Rationale
It is generally best practice to explicitly declare dependencies that are used in your code rather than relying on transitive dependencies to supply them for you if for no other reason than it documents what is in use.  Though in reality the risk of these transitive dependencies disappearing is low and would usually be easily noticed, the declaration of proper `api` dependencies will assure that our published `pom` files are accurate and others will not have to separately declare some dependencies that we rely on when using our jar files.

The changes made here were done with the help of this [Gradle dependency-analysis plugin](https://plugins.gradle.org/plugin/com.autonomousapps.dependency-analysis), which is still not at v1.0 and seems to have a few quirks, but my spot checks of the suggestions given seem like they provide solid advice.

Some of the dependencies we have been declaring in the `api` module are no longer (perhaps never were) used within that module, so a few other modules have been updated to include these dependencies.  This means, of course, that some jar files will be included twice for certain distributions, but there are not many and this is part of the price of keeping things modular.  However, if the repetition of things like `commons-math3` seems cumbersome or silly we can go back to including these in `api` (at least for now), but it is really best practice to own up to your dependencies.

#### Related Pull Requests
* Coming soon

#### Changes
* Update dependency declarations


api/build.gradle
```diff
@@ -54,7 +45,7 @@ List jackson = [
 ]
 
 List charting = [
-        "org.jfree:jfreechart:${jfreechartVersion}"
+        "org.jfree:jfreechart:${jfreechartVersion}",
```
labkey-adam: Trailing comma?

labkey-jeckels: Intentional trailing comma?

labkey-jeckels (approved):

labkey-adam (approved):

labkey-tchad (approved):

PR #1575, opened 2020-09-18
Issue 41435: support long property values in getProperties()

labkey-kevink:


labkey-jeckels (approved):

labkey-adam:
Yes, this will work, but I hate it when we create side `switch` statements on our own enums. Adding methods directly to the enum is far preferable. If we move `getValue()` to the enum there's never a chance that we'll forget to implement it (as we did with BIGINT and are still doing with BINARY... and will continue to with any new constant) and there's no need for the IllegalStateException fall through. You'll have to pass in the ObjectProperty as a param, but it seems a lot cleaner to me. I'd say the same with the `init()` switch as well.

PR #1576, opened 2020-09-18
7857: Apply moduleEditor feedback

RosalineP:
#### Rationale
On Known Modules page, 'Edit Module' link should only be shown if moduleEditor is not included in the build, or if the relevant module is not editable.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/345
* https://github.com/LabKey/moduleEditor/pull/12

#### Changes
* Add helper function to DefaultModule
* Add checks for moduleEditor's existence and module validity before displaying 'Edit Module' link

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -8111,9 +8111,7 @@ public class AdminController extends SpringActionController
                                         if (!rel.startsWith(".."))
                                             shortPathToModule = rel.toString();
                                     }
-                                    boolean modIsValid = (module instanceof DefaultModule) && ((DefaultModule) module).isValidSourceDirectory();
                                     ActionURL moduleEditorUrl = getModuleEditorURL(moduleContext.getName());
-                                    boolean showModuleEditor = ModuleLoader.getInstance().hasModule("ModuleEditor") && (null != moduleEditorUrl) && modIsValid;
 
                                     return TR(cl(rowCount.getAndIncrement()%2==0 ? "labkey-alternate-row" : "labkey-row").at(style,"vertical-align:top;"),
                                         TD(moduleContext.getName()),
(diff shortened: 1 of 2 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -8111,6 +8111,8 @@ protected void renderView(Object model, PrintWriter out)
                                         if (!rel.startsWith(".."))
                                             shortPathToModule = rel.toString();
                                     }
+                                    boolean modIsValid = (module instanceof DefaultModule) && ((DefaultModule) module).isValidSourceDirectory();
```
RosalineP: @labkey-matthewb 
This looked like the correct check to invoke based off of your comments on the ModuleEditor spec regarding when to allow editing of module resources. Happy to make any fixes if I've misapprehended anything, or if there are additional gates to include when determining if a module is editable or not!

labkey-matthewb: This logic shouldn't be scattered around the product, it should be centralized in ModuleEditorService.  Presumably null != ModuleService.getModuleEditorURL() should be accurate.

labkey-matthewb (changes requested):

core/src/org/labkey/core/admin/AdminController.java
RosalineP: Hopefully addressed as per offline gChat convo.

labkey-matthewb: Doesn't address the original comment however.  Presumably getModuleEditorURL() is correctly determining if this module can be edited and what the URL should be.  Why second guess that code???  Just check != null.

labkey-matthewb (changes requested):

core/src/org/labkey/core/admin/AdminController.java
RosalineP: Ah, the intention of the `isValidSourceDirectory()` check goes back to a comment you made on the ModuleEditor spec (under the 'Open questions' section) regarding when to allow editing of module resources—you mention that we should allow editing if we have a valid source directory, and to take a look at logic implemented in `computeResourceDirectory()` for that. I thought that this was what you were referring to when you mention in the Issue opened that if a module isn't editable, the link to the ModuleEditor shouldn't appear. (My bad—should have made this more clear than my vague comment above about the ModuleEditor spec!)

It sounds like I was probably misinterpreting that, though, and the nullcheck with getModuleEditorURL() is sufficient here for determining whether to display the link to the ModuleEditor on the known modules page! Changes pushed.

labkey-matthewb (approved):

PR #1582, opened 2020-09-21
Migrate theming to @labkey/themes

labkey-nicka:
#### Rationale
See https://github.com/LabKey/labkey-ui-components/pull/347 for background and rationale.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/347

#### Changes
* Migrate theme publishing source and assets to `@labkey/themes`.
* Remove theme building targets in core module and replace with targets to copy the distribution assets for `@labkey` packages (i.e. `@labkey/api` and `@labkey/themes`).
* Remove unnecessary dependency on `react-toggle-button`.


labkey-susanh (approved):
My only suggestion would be to possibly rename `build.js` to something like `copy-distributions.js` since `build` implies (to me) more general heavy lifting than is going on.  But only a suggestion `build.js` will serve well if the heavy lifting comes later.

cnathe (approved):

PR #1587, opened 2020-09-22
Add test for LabKey SQL container filter annotations

labkey-tchad:
#### Rationale
Automated test for new LabKey SQL feature.


query/test/src/org/labkey/test/tests/query/ContainerFilterQueryTest.java
```diff
@@ -0,0 +1,195 @@
...
+
+        DataRegionTable table = createQuery(getProjectName(), queryName, "core", sql);
+        List<String> containerNames = table.getColumnDataAsText("Name");
+        assertEquals("Wrong containers for 'Current' container filter.",
+            List.of(getProjectName()), containerNames);
+
+        table.setContainerFilter(DataRegionTable.ContainerFilterType.ALL_FOLDERS);
+        containerNames = table.getColumnDataAsText("Name");
+        assertEquals("ContainerFilter annotation should ignore data region container filter.",
+            List.of(getProjectName()), containerNames);
```
labkey-danield, line 99: You sure you want to do an assertEquals here. If it fails is there no value in doing the next check? (Same question applies to some of the following tests as well.)

labkey-danield (commented):
It looks ok to me, but Matt should review for test coverage.

labkey-chrisj (approved):

labkey-matthewb (approved):

PR #1589, opened 2020-09-22
Item 7767: View details section of ErrorPage Redesign - Story 2

labkey-ankurj:
#### Rationale
This is the continued work of error page redesign.

#### Related Pull Requests
* https://github.com/LabKey/LabDevKitModules/pull/54

#### Changes
* View details section on Error Page


core/src/client/ErrorHandler/ErrorHandler.tsx
```diff
@@ -18,38 +18,51 @@ interface ErrorHandlerState {
...
+            window.location.href = ActionURL.getBaseURL(false);
+        }
     };
 
     onViewDetailsClick = (): void => {
-        this.setState(() => ({
-            showDetails: true,
-        }));
+        const { showDetails } = this.state;
+        this.setState({ showDetails: !showDetails });
```
cnathe: You'll want to use the state at the time of the callback.
```suggestion
        this.setState((state) => ({ showDetails: !state.showDetails }));
```

core/src/client/ErrorHandler/ErrorHandler.tsx
```diff
@@ -18,38 +18,51 @@ interface ErrorHandlerState {
...
-                {/* TODO : ErrorPage, following section in next story*/}
-                {showDetails && <h3 className="labkey-error">{message}</h3>}
+                <div className="error-details-body">
+                    <div className="row">
+                        <div className="col-md-8">
+                            <div className="labkey-error-top">
+                                {getErrorHeading()}
+                                {getSubHeading(errorDetails)}
+                                {getInstruction(errorDetails)}
+                                <button className="btn btn-info error-backButton" onClick={this.onBackClick}>
```
cnathe: I know the mockups had this as the bootstrap blue "info" button, but we should stick to our standard button classes and let our core scss (which will hopefully remove some of the bootstrap overrides in the future).
```suggestion
                                <button className="btn btn-primary error-backButton" onClick={this.onBackClick}>
```

core/src/client/ErrorHandler/ErrorType.tsx
```diff
@@ -1,90 +1,297 @@
...
+        <div className="labkey-error-details labkey-error-details-question">What else can I do?</div>
+        <div className="labkey-error-details">
+            If you are subscribed to LabKey, then reach out to your account manager for immediate help. If you are using
+            the community edition, your support ticket will be resolved within the next 2 to 3 weeks.
+        </div>
+        <br />
+        <div className="labkey-error-details">
+            {helpLinkNode('default', 'LabKey support documentation')} is another immediate resource that is available to
+            all users. A search through the documentation or previous forum questions may also help troubleshoot your
+            issue.
```
cnathe: Does anyone know if this text has been reviewed by the support team? Looking at [this page](https://www.labkey.org/home/Support/project-begin.view?), it looks like the response time of community forum posts is "as time allows". Ankur, if you haven't chatted with the support team and reviewed these pages with them yet, I'm happy to do so.

core/src/client/ErrorHandler/ErrorType.tsx
```diff
@@ -1,90 +1,297 @@
...
+            Please contact your admin or reference the{' '}
+            <a href="https://www.labkey.org/project/home/Support/begin.view?" rel="noopener noreferrer" target="_blank">
+                LabKey support forum.
+            </a>
         </div>
+        {errorCode !== undefined && (
+            <div className="labkey-error-instruction">
+                If you would like to file a{' '}
+                <a
+                    href="https://www.labkey.org/Support%20Tickets/wiki-edit.view?"
```
cnathe: wiki-edit.view? that can't be the right place to send them, is it? I would expect any of the support related links would just go to this URL: https://www.labkey.org/home/Support/project-begin.view?

cnathe (changes requested):
Just some quick comments. I'll do a more thorough code review after manual testing.

core/src/client/ErrorHandler/ErrorType.tsx
labkey-ankurj: I asked the services team around the wording of config exception in the specs but not all the pages. Yes, it'd be great if you don't mind chatting with the team. Thanks.

core/src/client/ErrorHandler/ErrorType.tsx
labkey-ankurj: I found that link in the specs to linked to filing a ticket. I can update if that is not correct.

labkey-jeckels: It's definitely not the right spot. I just send an invite for a meeting next week to get this and other text finalized.

core/src/client/ErrorHandler/ErrorHandler.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+describe('ErrorHandler', () => {
+    test('Not found exception', () => {
+        const errorDetails: ErrorDetails = {
+            errorType: ErrorType.notFound,
+            errorCode: '123XYZ',
+            message: 'This is a not found exception',
+        };
+        const wrapper = mount(<ErrorHandler context={{ errorDetails }} />);
+        expect(wrapper.find('.error-details-container')).toHaveLength(0);
```
cnathe, line 18: is there some header text / element  that we can check based on the `errorType`? Just checking that `error-details-container` isn't visible until after the state change is more of a check of the view details button and not the error type differences.

core/src/client/ErrorHandler/ErrorHandler.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+        wrapper.setState({ showDetails: true });
+        const question = wrapper.find('.error-details-container');
+        expect(question.text().startsWith('What went wrong?')).toBeTruthy();
```
cnathe, line 22: any way we can also check for the `message` from the ErrorDetails context somewhere in this test case too?

core/src/client/ErrorHandler/ErrorHandler.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+        wrapper.unmount();
+    });
+
+    test('Configuration exception', () => {
+        const errorDetails: ErrorDetails = {
+            errorType: ErrorType.configuration,
```
cnathe, line 30: similar comment as above, this test case is the same as the above one in what things it is checking for even though the `errorType` is different. It would be good to check more about the error type changes in these cases if possible.

core/src/client/ErrorHandler/ErrorHandler.tsx
```diff
@@ -16,40 +16,60 @@ interface ErrorHandlerProps {
...
+            window.location.href = ActionURL.getBaseURL(false);
+        }
     };
 
     onViewDetailsClick = (): void => {
-        this.setState(() => ({
-            showDetails: true,
-        }));
+        const { showDetails } = this.state;
+        const viewDetailsBtnText = !showDetails ? this._hideDetails : this._viewDetails;
```
cnathe: I would probably just move this if/else logic to the button text display instead of stashing it in state. We already have the `showDetails` state variable we can use for that button text display.

core/src/org/labkey/core/webdav/DavController.java
```diff
@@ -72,27 +72,10 @@
...
-import org.labkey.api.util.MemTracker;
-import org.labkey.api.util.PageFlowUtil;
-import org.labkey.api.util.Pair;
-import org.labkey.api.util.Path;
-import org.labkey.api.util.ResponseHelper;
-import org.labkey.api.util.ShutdownListener;
-import org.labkey.api.util.StringUtilsLabKey;
-import org.labkey.api.util.URLHelper;
-import org.labkey.api.util.UnexpectedException;
+import org.labkey.api.util.*;
```
cnathe, line 75: was this change to the `*` import intentional or just something that IntelliJ did for you?

core/src/client/ErrorHandler/ErrorType.tsx
```diff
@@ -1,90 +1,297 @@
...
+                    {getServerContext().user.isSignedIn
+                        ? 'You are currently logged in as: ' +
+                          (getServerContext().impersonatingUser !== undefined
+                              ? getServerContext().impersonatingUser.displayName
+                              : getServerContext().user.displayName)
+                        : 'You are not logged in.'}
+                </li>
+            </ul>
+        </div>
+        {getServerContext().impersonatingUser !== undefined && (
```
cnathe, line 133: It would be good to get some jest tests for the various different displays here for the permissions error with and without the impersonate user.

cnathe (changes requested):

labkey-nicka:
@labkey-ankurj I've closed the related `@labkey/components` PR so this branch can just update to the latest version of that package (which might result in no diff from develop).

core/src/org/labkey/core/webdav/DavController.java
```diff
@@ -427,6 +410,24 @@ WebdavStatus sendError(WebdavStatus status, Exception x)
             String message = x.getMessage() != null ? x.getMessage() : status.message;
             if (x instanceof ConfigurationException)
                 message += "\nThis may be a server configuration problem.  Contact the site administrator.";
+
+            ErrorRenderer renderer = ExceptionUtil.getErrorRenderer(HttpServletResponse.SC_BAD_REQUEST, message, x, getViewContext().getRequest(), false, false);
```
labkey-nicka: I'm curious why we switch to error view here? The `sendMessage` function is intended to handle how the error response is configured. We don't want to return an HTML view when the response is requested as JSON.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -2235,7 +2238,9 @@ public void export(ConvertArraysToExcelForm form, HttpServletResponse response,
             catch (JSONException | ClassCastException e)
             {
                 // We can get a ClassCastException if we expect an array and get a simple String, for example
-                HttpView errorView = ExceptionUtil.getErrorView(HttpServletResponse.SC_BAD_REQUEST, "Failed to convert to Excel - invalid input", e, getViewContext().getRequest(), false);
+                ErrorRenderer renderer = ExceptionUtil.getErrorRenderer(HttpServletResponse.SC_BAD_REQUEST, "Failed to convert to Excel - invalid input", e, getViewContext().getRequest(), false, false);
```
labkey-nicka: Expecting JSON?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -2305,8 +2310,14 @@ public void export(ConvertArraysToExcelForm form, HttpServletResponse response,
             }
             catch (JSONException e)
             {
-                HttpView errorView = ExceptionUtil.getErrorView(HttpServletResponse.SC_BAD_REQUEST, "Failed to convert to table - invalid input", e, getViewContext().getRequest(), false);
-                errorView.render(getViewContext().getRequest(), getViewContext().getResponse());
+                ErrorRenderer renderer = ExceptionUtil.getErrorRenderer(HttpServletResponse.SC_BAD_REQUEST, "Failed to convert to table - invalid input", e, getViewContext().getRequest(), false, false);
```
labkey-nicka: Expecting JSON?

labkey-nicka (changes requested):
I'd like to discuss the different usages of `ErrorView` and see if we can consolidate them so that the implementation that is specific to the view isn't the callers responsibility.

api/src/org/labkey/api/util/ExceptionUtil.java
```diff
@@ -988,6 +918,19 @@ private static void renderErrorPage(Throwable ex, int responseStatus, String mes
     }
 
 
+    public static void renderErrorView(ViewContext context, PageConfig pageConfig, ErrorRenderer.ErrorType errorType, int responseStatus, String message, @Nullable Throwable ex, boolean isPart, boolean isStartupFailure) throws Exception
```
labkey-ankurj, line 923: @labkey-nicka added this method

core/src/org/labkey/core/webdav/DavController.java
```diff
@@ -469,6 +453,19 @@ WebdavStatus sendError(WebdavStatus status, String message)
                             super.setStatus(HttpServletResponse.SC_OK);
                         }
                     }
+                    else if (CONTENT_TYPE_HTML.equals(getRequest().getHeader("Content-Type)")) ||
```
labkey-ankurj, line 473: @labkey-nicka Let me know if this looks right. I tried to test through the URL link that you showed yesterday and it renders the error page in browser since the content type is text/html but returns json when apt content type is passed in the request. 
Screen Shot 2020-09-29 at 10 43 03 AM

api/src/org/labkey/api/util/ExceptionUtil.java
labkey-nicka: Very nice!

core/src/org/labkey/core/webdav/DavController.java
labkey-nicka: Yes, this is now appropriately scoped and looks to still respond with JSON.

It is now assuming that the response code is 404 (not found) but then returning a status of 400 (bad request). What happens if a different type of error occurs? If is a 404 shouldn't we respond as such?

core/src/org/labkey/core/webdav/DavController.java
```diff
@@ -470,6 +453,19 @@ WebdavStatus sendError(WebdavStatus status, String message)
...
+                            accept.contains(CONTENT_TYPE_HTML))
+                    {
+                        try
+                        {
+                            ExceptionUtil.renderErrorView(getViewContext(), new PageConfig(), ErrorRenderer.ErrorType.notFound, HttpServletResponse.SC_BAD_REQUEST, message, null, false, false );
+                            return WebdavStatus.SC_BAD_REQUEST;
+                        }
+                        catch (Exception e)
+                        {
+                            return sendError(status, message);
```
labkey-nicka: This could cause an infinite loop as this is calling the same method. I believe this should call `super.sendError(status.code, message);` and not return.

labkey-nicka (changes requested):

core/src/org/labkey/core/webdav/DavController.java
labkey-ankurj: updated the status code here, but error type here I think should always be not found as this is being utilized only for not found pages.

core/src/org/labkey/core/webdav/DavController.java
labkey-ankurj: updated. thanks.

core/src/org/labkey/core/webdav/DavController.java
labkey-ankurj: this was intellij.. ran the optimize imports.

core/src/client/ErrorHandler/ErrorHandler.spec.tsx
labkey-ankurj: Added a text check for the sub heading in the top section and text check in details section.

core/src/client/ErrorHandler/ErrorHandler.spec.tsx
labkey-ankurj: done

core/src/client/ErrorHandler/ErrorHandler.spec.tsx
labkey-ankurj: Added a text check for the sub heading in the top section and text check in details section.

core/src/client/ErrorHandler/ErrorHandler.tsx
labkey-ankurj: done

core/src/client/ErrorHandler/ErrorType.tsx
labkey-ankurj: Not sure how to impersonate in jest tests, do you have any pointers to existing example?

api/src/org/labkey/api/view/template/errorView.jsp
```diff
@@ -1,32 +1,58 @@
...
-    LABKEY.App.loadApp('errorHandler', <%=q(appId)%>, {
-        message: "<%=unsafe(model.getHeading())%>",
-        errorType: "<%=unsafe(model.getErrorType())%>"
+    // pulling in theme based css for the not found pages to still respect the site's theme
+    LABKEY.requiresCss(<%=q("/core/css/" + PageFlowUtil.resolveThemeName(getContainer()) + ".css")%>);
+
+    LABKEY.requiresScript('core/gen/errorHandler.js', function() {
+    // LABKEY.requiresScript('http://localhost:3001/errorHandler.js', function() {
+
+        LABKEY.App.__app__.isDOMContentLoaded = true;
```
labkey-nicka, line 51: @labkey-ankurj Can you open a ticket to me about this. It's fine to leave this "as-is" in this PR but the app loader should correctly identify when the DOM is loaded so you don't have to touch it's "private" state.

api/src/org/labkey/api/view/template/errorView.jsp
```diff
@@ -1,32 +1,58 @@
...
+        }
+    }
+%>
+
 <script type="application/javascript">
-    LABKEY.App.loadApp('errorHandler', <%=q(appId)%>, {
-        message: "<%=unsafe(model.getHeading())%>",
-        errorType: "<%=unsafe(model.getErrorType())%>"
+    // pulling in theme based css for the not found pages to still respect the site's theme
+    LABKEY.requiresCss(<%=q("/core/css/" + PageFlowUtil.resolveThemeName(getContainer()) + ".css")%>);
```
labkey-nicka, line 46: nit: I might further update this comment to state why the CSS is being pulled in here rather than in the client dependencies for the JSP. Something like:

```
This error page may be invoked without the themes having been loaded for this container. We load the theme artifact for this container here to ensure the correct theme is loaded as this cannot be resolved during "addClientDependencies()" for this view.

**Review by labkey-nicka — COMMENTED** [review]
Looks good to me. Only a couple nit items. I'll let @cnathe approve the PR once he's reviewed again.

**Review by labkey-nicka — APPROVED** [review]

File: core/src/client/ErrorHandler/ErrorType.tsx (hunk 10 above)
Comment (cnathe) [reply core/src/client/ErrorHandler/ErrorType.tsx:133]: If it is just a matter of setting some specific context on the LABKEY object for a certain test case (i.e. the LABKEY.impersonatingUser in this case), then we have an example of that in labkey-ui-components which might work here:
labkey-ui-components/packages/components/src/internal/app/utils.spec.ts

**Review by cnathe — APPROVED** [review]

File: api/src/org/labkey/api/util/ExceptionUtil.java [hunk 19]
```diff
@@ -961,19 +897,24 @@ private static void renderErrorPage(Throwable ex, int responseStatus, String mes
...
+                errorView = PageConfig.Template.Body.getTemplate(new ViewContext(request, response, new ActionURL(ActionURL.getBaseServerURL())), new ErrorView(renderer), pageConfig);
             }
 
             pageConfig.addClientDependencies(errorView.getClientDependencies());
             errorView.getView().render(errorView.getModel(), request, response);
         }
         catch (Exception e)
         {
+            // config exceptions that occur before jsps have been initialized
+            if (ex instanceof ConfigurationException && e instanceof NullPointerException)
```
labkey-jeckels: Can we do a more direct check if JSPs are ready to render? What's null that's causing this exception?

PR #1590, opened 2020-09-23
Issue 39386: Importing sample type design XAR makes first field required

labkey-kevink:
#### Rationale
SampleType.getIdCols() is strange in that it includes the first domain property if
no ID columns have been explicitly set.  We were not properly round-tripping the ID columns
when using 'Name' as the ID column.

#### Changes
* On export of a SampleType to XAR, include the 'Name' ID column when using it as the ID col.
* On import, don't validate the 'Name' property exists in the domain when setting it as the ID col.
* Only mark the ID columns as required when not using a name generator expression
* Read the SampleType label color on import.
* Add clarifying javadoc


labkey-jeckels (approved):

labkey-danield (commented):
Thanks for the heads up on this change/bug fix. It looks ok to me but I'm going to defer to a Dev to approve.

PR #1593, opened 2020-09-23
Specimen aliases #39672

labkey-adam:
#### Rationale
Column import aliases on specimens don't work. Let's make them work. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=39672

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/510

#### Changes
* Push import aliases into `ImportableColumn`
* Teach various import code paths to use import aliases


study/src/org/labkey/study/importer/SpecimenImporter.java
```diff
@@ -3117,19 +3183,14 @@ public DataIterator getDataIterator(final DataIteratorContext context)
...
-                        public Object call() throws Exception
-                        {
-                            Object computedValue = cc.getValue(_rowMap);
-                            if (computedValue instanceof Parameter.TypedValue)
-                                return ((Parameter.TypedValue) computedValue).getJdbcParameterValue();
-                            else
-                                return computedValue;
-                        }
+                    Callable<Object> call = () -> {
+                        Object computedValue = cc.getValue(_rowMap);
```
labkey-matthewb, line 3230: Pretty sure that the implementations of ComputedColumn.getValue() won't benefit from the importAlias mapping.  The column names in _rowMap are the still the TSV names, and the ComputedColumn impls probably use the "expected" name when they call _rowMap.get(String).

study/src/org/labkey/study/importer/SpecimenImporter.java
```diff
@@ -3138,36 +3199,34 @@ public Object call() throws Exception
                 if (seen.add(ic.getLegalDbColumnName(d)))
                 {
                     String boundInputColumnName = null;
-                    if (tsvColumnNames.contains(ic.getTsvColumnName()))
-                        boundInputColumnName = ic.getTsvColumnName();
+                    Optional<String> firstImportName = ic.getImportNames().stream()
```
labkey-matthewb: The SDIB is a little stricter.  a) In case of ambiguity it prefers column name match over import alias match (because of how it creates the map, insert labels first, then names)  b) it enforces that no column can be matched two source columns (e.g instead of using .first() enforce that there is exactly one match).

labkey-matthewb (commented):
To solve the computed column problem, it's probably easiest to have SpecimenImporterBuilder do something like

   MapDataIterator in = dib.getDataIterator(context);
   remap = new SpecimenRemapIterator(this, in, context);
   new SpecimenImportIterator(this, remap, context);

study/src/org/labkey/study/importer/SpecimenImporter.java
labkey-matthewb: I scanned some of the ComputedColumn impls.  I think these are the columns that are used, and therefore might not work with import aliases?

  global_unique_specimen_id
  specimen_number
  Ptid
  VisitValue
  draw_timestamp

study/src/org/labkey/study/importer/SpecimenImporter.java
labkey-matthewb: new method createImportMap() helps keep this mapping consistent.

labkey-matthewb (approved):

PR #1596, opened 2020-09-24
40664: Add startup property that creates API keys

labkey-adam:
#### Rationale
Providing a mechanism to create API keys and associate them with specific users at bootstrap time would be useful for testing remote LabKey instances plus other scenarios.

#### Changes
* Add startup property that creates API keys. Example: `ApiKey.info@labkey.com = apikey|9d6036f5e4d910a4365b4cae03b169b2`


labkey-jeckels (approved):
Do API keys need to start with "apikey|" to be recognized for authentication? If so, we should validate the desired value matches the pattern.

labkey-adam:
Yes they do. I can add that validation. They should also be long random strings, but I wasn't planning any other validation (e.g., length or entropy or whatnot).

labkey-jeckels (approved):
Looks good. I agree, up to the caller to choose a secure API key.

curiositycasualty:
posthumous 👍

PR #1600, opened 2020-09-25
ClusterStartup exit code, file root overrides for workbooks

labkey-jeckels:
This PR has two changes:

ClusterStartup will give a simple non-zero exit when the taskstatus is ERRROR, but otherwise exit with 0.

This will allow workbooks to have a file root override, and adds an integration test to that effect. This is a minimal change, and the folder management UI is still restricted to the parent folder. However, this should let custom code in modules alter workbook file roots.

### DIFF
--- a/api/src/org/labkey/api/data/WorkbookContainerType.java
+++ b/api/src/org/labkey/api/data/WorkbookContainerType.java
@@ -15,6 +15,7 @@
  */
 package org.labkey.api.data;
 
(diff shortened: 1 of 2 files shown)


labkey-jeckels:
@labkey-ankurj can you approve this? I can't because I created the PR from the LabKey repo, but I'm comfortable with the changes Ben has proposed.

labkey-ankurj (approved):

PR #1602, opened 2020-09-26
Fix HTML link generation to avoid extra quote

labkey-jeckels:
#### Rationale
Switch to using HtmlString when rendering a grid or details value to ensure proper encoding. Fix up many problematic subclasses.

When we have a link with a target tab/window, we're injecting an extra quote into the HTML, because we're adding a closing quote when we append the next attribute

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1317

#### Changes
* Remove deprecated getFormattedValue(), consolidate on getFormattedHtml()
* Use HtmlString to signal this is specifically for HTML
* Remove the extra quote

### DIFF
--- a/api/src/org/labkey/api/util/HtmlString.java
+++ b/api/src/org/labkey/api/util/HtmlString.java
@@ -25,9 +25,10 @@ import java.util.Objects;
 public final class HtmlString implements SafeToRender, DOM.Renderable, Comparable<HtmlString>
 {
     // Helpful constants for convenience (and efficiency)
-    public static HtmlString EMPTY_STRING = HtmlString.of("");
-    public static HtmlString NBSP = HtmlString.unsafe("&nbsp;");
-    public static HtmlString NDASH = HtmlString.unsafe("&ndash;");
+    public static final HtmlString EMPTY_STRING = HtmlString.of("");
+    public static final HtmlString NBSP = HtmlString.unsafe("&nbsp;");
+    public static final HtmlString NDASH = HtmlString.unsafe("&ndash;");
+    public static final HtmlString BR = HtmlString.unsafe("<br/>");
 
     private final @NotNull String _s;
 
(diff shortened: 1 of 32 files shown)


labkey-adam:
Ugh, why don't we replace all of this code with `LinkBuilder`?

labkey-adam:
Thanks, Josh! Much nicer, IMO. Up to you, but for brevity, you should be able to simply set renderURLTitle(ctx), getLinkCls(), getOnClick(), and getCssStyle(ctx) on the LinkBuilder, without any if checks. getLinkTarget() might still need the if since we want to add the rel().

api/src/org/labkey/api/util/HtmlString.java
```diff
@@ -36,19 +37,24 @@
      * @param s A String. A null value results in an empty HtmlString (equivalent of HtmlString.of("")).
```
labkey-ankurj:
```suggestion
     * @param s A char sequence. A null value results in an empty HtmlString (equivalent of HtmlString.of("")).
```

api/src/org/labkey/api/util/HtmlString.java
```diff
@@ -36,19 +37,24 @@
...
 
+    public static @NotNull HtmlString of(@Nullable Object o)
+    {
+        return of(o == null ? null : o.toString());
+    }
+
     /**
      * Returns an HtmlString that wraps an HTML encoded version of the passed in String, with the option to preserve
      * whitespace.
      * @param s A String. A null value results in an empty HtmlString (equivalent of HtmlString.of("")).
```
labkey-ankurj:
```suggestion
     * @param s A char sequence. A null value results in an empty HtmlString (equivalent of HtmlString.of("")).
```

labkey-ankurj (approved):

PR #1605, opened 2020-09-28
41472: Copied to study assay results column names can change

labkey-klum:
#### Rationale
This change fixes issue [41472](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41472). While this change does alter the generated copied to study column names, it only does so for the previously broken case where a study name consisted entirely of numbers. This scenario resulted in unstable column names when multiple copied to study names were added that only consisted of numbers in the name.

### DIFF
--- a/api/src/org/labkey/api/assay/AssayProtocolSchema.java
+++ b/api/src/org/labkey/api/assay/AssayProtocolSchema.java
@@ -843,7 +843,9 @@ public abstract class AssayProtocolSchema extends AssaySchema
                 String studyName = assayDataset.getStudy().getLabel();
                 if (studyName == null)
                     continue; // No study in that folder
-                String studyColumnName = "copied_to_" + sanitizeName(studyName);
+
+                // issue 41472 include the prefix as part of the sanitization process
+                String studyColumnName = sanitizeName("copied_to_" + studyName);
 
                 // column names must be unique. Prevent collisions
                 while (usedColumnNames.contains(studyColumnName))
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/assay/AssayProtocolSchema.java
```diff
@@ -843,7 +843,15 @@ else if (!first)
                 String studyName = assayDataset.getStudy().getLabel();
                 if (studyName == null)
                     continue; // No study in that folder
-                String studyColumnName = "copied_to_" + sanitizeName(studyName);
+
+                String studyColumnName;
+                if (sanitizeName(assayDataset.getStudy().getContainer().getName()).isEmpty())
```
labkey-jeckels: Why do we look at the container name in the if check but then use the study label when generating the column name?

labkey-jeckels (changes requested):

api/src/org/labkey/api/assay/AssayProtocolSchema.java
labkey-klum: I was attempting to constrain the column name changes to only the case where the entire study folder name consisted of numbers but a study folder name of 123FooBar would still create a column name of : copied_to_FooBar_Study which would be the same as prior to this change.

labkey-jeckels: Yes, that makes sense, but the container name doesn't necessarily match the study's label. You could have a container name that doesn't start with a number, but a study label that does. Wouldn't you want to check the same string in the if() statement as you're sanitizing?

labkey-klum: good point, line 848 could be replaced with : `if (sanitizeName(studyName).isEmpty())`

labkey-jeckels: I assume this update is in your queue, @labkey-klum? I can approve when it's pushed.

labkey-jeckels (approved):

PR #1608, opened 2020-09-29
Issue 41470: ETL'ing a decimal value into a column of type decimal is rounding off decimal

labkey-martyp:
#### Rationale
When performing an ETL merge, values are stored in variables of the same type as the target column.  These were not including precision and scale which in SQL Server defaults to 0 scale, thus rounding off the decimal value.  Postgres does not coerce values if no precision and scale defined so it was not affected by this issue.  This does not affect LK built in datatypes (dataset, list, etc).

#### Changes
- Add Decimal Digits (scale) metadata to JDBC metadata reader
- Add Decimal Digits parameter to SQL dialects
- Add Decimal Digits to BaseMicrosoftSqlServerDialect metadata query
- Add Precision to column info (via ColumnRenderPropertiesImpl)
- Set precision and scale for decimal values in BaseColumnInfo.createFromDatabaseMetaData.
- Add Precision and Scale to Parameter
- Add precision and scale to Decimal variables in StatementUtils.variableDeclaration
- Update JdbcType._toBigDecimal to not use BigDecimal(double) constructor as this provides unreliable results. BigDecimal(string) is reliable according to docs


api/src/org/labkey/api/data/ColumnRenderProperties.java
```diff
@@ -171,5 +171,7 @@ else if (Date.class.isAssignableFrom(javaClass))
 
     int getScale();
```
labkey-jeckels, line 176: Suggest adding some comments here about what scale and decimal digits mean (and how they related to the "normal" use of those terms)

bigiron/src/org/labkey/bigiron/mysql/MySqlDialect.java
```diff
@@ -165,6 +165,7 @@ private MySqlColumnMetaDataReader(ResultSet rsCols)
             _sqlTypeKey = "DATA_TYPE";
             _sqlTypeNameKey = "TYPE_NAME";
             _scaleKey = "COLUMN_SIZE";
+            _decimalDigitsKey = "DECIMAL_DIGITS";
```
labkey-jeckels, line 168: Have you tested on these other DBs or at least verified that they match?

labkey-jeckels (commented):
Changes look good to me, but would be great to get Matt's quick review too.

labkey-matthewb:
Why the use of "DecimalDigits" instead of scale which is the common term for database documentation and java BigDecimal?  I think it adds a lot of value to not require translating from database-ese to labkey-ese.

https://docs.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql?view=sql-server-ver15
https://www.postgresql.org/docs/9.1/datatype-numeric.html

labkey-matthewb:
Is this right?

Parameter.length == columninfo.scale == SQL precision
Parameter.decimaldigits == columninfo.decimaldigits == SQL scale

We can't use different names like this, and the columninfo.scale thing just needs to be fixed.

labkey-martyp:
@labkey-matthewb Yes those are correct.  I filed an issue for triage on our use of "scale".  We're also using it in PropertyDescriptors, XSDs and our api response, so I was trying to find a work around to not have multiple definitions of scale in LK and still be backwards compatible with our current version.  Not sure exactly how to proceed on the conflict of LK scale vs SQL scale.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41497

> Is this right?
> 
> Parameter.length == columninfo.scale == SQL precision
> Parameter.decimaldigits == columninfo.decimaldigits == SQL scale
> 
> We can't use different names like this, and the columninfo.scale thing just needs to be fixed.

labkey-matthewb:
I really don't think you will break anything if you give it the correct meaning for scale when the type is DECIMAL.  We don't use decimal very much, and we can just make a release not that we've cleaned up that property in our api.

> @labkey-matthewb Yes those are correct. I filed an issue for triage on our use of "scale". We're also using it in PropertyDescriptors, XSDs and our api response, so I was trying to find a work around to not have multiple definitions of scale in LK and still be backwards compatible with our current version. Not sure exactly how to proceed on the conflict of LK scale vs SQL scale.
> 
> https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41497
> 
> > Is this right?
> > Parameter.length == columninfo.scale == SQL precision
> > Parameter.decimaldigits == columninfo.decimaldigits == SQL scale
> > We can't use different names like this, and the columninfo.scale thing just needs to be fixed.

labkey-martyp:
Ok, then add a precision field to ColumnInfo as well?  This would only be used for DECIMAL values?  Right now precision is being stored in scale.

> I really don't think you will break anything if you give it the correct meaning for scale when the type is DECIMAL.  We don't use decimal very much, and we can just make a release not that we've cleaned up that property in our api.
> 
> > @labkey-matthewb Yes those are correct. I filed an issue for triage on our use of "scale". We're also using it in PropertyDescriptors, XSDs and our api response, so I was trying to find a work around to not have multiple definitions of scale in LK and still be backwards compatible with our current version. Not sure exactly how to proceed on the conflict of LK scale vs SQL scale.
> > https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41497
> > > Is this right?
> > > Parameter.length == columninfo.scale == SQL precision
> > > Parameter.decimaldigits == columninfo.decimaldigits == SQL scale
> > > We can't use different names like this, and the columninfo.scale thing just needs to be fixed.

bigiron/src/org/labkey/bigiron/mysql/MySqlDialect.java
labkey-martyp: I did not test these but I got them from the docs for each db type

api/src/org/labkey/api/data/ColumnRenderProperties.java
labkey-martyp: Comments added

labkey-martyp:
Ok I removed decimalDigits and added precision to ColumnRenderProperties and I'm correctly setting scale and precision for Decimals in BaseColumnInfo.createFromDatabaseMetaData.  Also updated Parameter to have scale and precision.  @labkey-matthewb let me if that looks good.

labkey-matthewb (approved):

PR #1610, opened 2020-09-30
Issue 41472: Include study name in copied to study assay column name when sanitized name equals "study"

cnathe:
#### Rationale
Issue [41472](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41472): Copied to study assay results column names can change when multiple studies are used for the same assay

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1605

#### Changes
* Add check for if the sanitized study name equals "study", ignoring case, when prepending "copied_to_study"

### DIFF
--- a/api/src/org/labkey/api/assay/AssayProtocolSchema.java
+++ b/api/src/org/labkey/api/assay/AssayProtocolSchema.java
@@ -845,7 +845,8 @@ public abstract class AssayProtocolSchema extends AssaySchema
                     continue; // No study in that folder
 
                 String studyColumnName;
-                if (sanitizeName(studyName).isEmpty())
+                String sanitizedStudyName = sanitizeName(studyName);
+                if (sanitizedStudyName.isEmpty() || "study".equalsIgnoreCase(sanitizedStudyName))
                 {
                     // issue 41472 include the prefix as part of the sanitization process
                     studyColumnName = sanitizeName("copied_to_" + studyName);


labkey-klum (changes requested):
Sorry for not looking at this sooner, I don't think your change would work in all cases because the user may have altered the default study name (100-my study, 200-my study) would still expose the existing problem.

We could just say that the study column name is always
`studyColumnName = sanitizeName("copied_to_" + studyName)`

But it would mean the change potentially affects more users.

cnathe:
> Sorry for not looking at this sooner, I don't think your change would work in all cases because the user may have altered the default study name (100-my study, 200-my study) would still expose the existing problem.
> 
> We could just say that the study column name is always
> `studyColumnName = sanitizeName("copied_to_" + studyName)`
> 
> But it would mean the change potentially affects more users.

@labkey-klum I agree that this won't fix the case where a user can alter the default study name and that the change to something like `studyColumnName = sanitizeName("copied_to_" + studyName)` would be a more general fix (but also less backwards compatible). This change to check specifically for "study" was to address just the default study name scenario.

cnathe:
@labkey-klum do you think we should ditch this change/PR until we decide what we want to do more generally, or do you think this change to handle the default study name case is worth taking for 20.11?

labkey-klum (approved):
I'm okay with this change for 20.11

PR #1620, opened 2020-10-07
Allow enabling troubleshooting stacktraces on TeamCity

labkey-tchad:
#### Rationale
We encounter occasional memory leaks on TeamCity. Being able to quickly enable stacktraces would be useful for debugging such failures.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/493

#### Changes
* New action to enable troubleshooting stacktraces via API

### DIFF
--- a/core/src/org/labkey/core/admin/miniprofiler/MiniProfilerController.java
+++ b/core/src/org/labkey/core/admin/miniprofiler/MiniProfilerController.java
@@ -169,6 +169,18 @@ public class MiniProfilerController extends SpringActionController
         }
     }
 
+    // Invoked by test framework
+    @RequiresSiteAdmin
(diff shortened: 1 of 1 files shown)


labkey-danield (approved):
It looks ok to me, but I don't have a lot of experience with this class/code.
I'm assuming the second parameter is there because it is an override. I wonder if that will cause any problems, or confusion later.

labkey-jeckels:
This seems OK, but is a separate approach from what's proposed in https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41534

I guess this may be slightly better because the stack traces aren't really an "experimental" feature.

labkey-adam:
This approach is fine with me

labkey-adam (approved):

PR #1621, opened 2020-10-08
Item 7882: Add support for discovery of data involved in transactions through detailed audit logging

labkey-susanh:
#### Rationale
Within LKSM, we want to enable users to work with the samples and sources that have just been imported, either via a file import or import within a grid.  Currently, this is possible when importing from a grid because the records created from using the `insertRows` operation are returned in the api response.  When importing from a file, only a count of the rows is returned (which is reasonable).  Here we add in an ability to find the data items that were inserted or updated during an upload operation by way of a new audit event type, which is created when the transaction begins and contains a unique transactionId that can then be stored in other audit logs as a reference.  

* [Issue 39594](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=39594)

#### Related Pull Requests
* 

#### Changes
* Add TransactionAuditProvider, Domain and EventType
* Add transactionID field to QueryUpdateEvents, SampleTimelineEvents, and SampleTypeEvents
* Add method to QueryService for retrieving audit records given a transactionId
* Add action in AuditController for retrieving data ids based on a transaction id (for sources and samples only at this point)
* Convert type of rowId in AuditTypeEvent from int to long to support use of DbSequence as a rowId


api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -143,6 +149,17 @@ protected boolean hasPermission(User user, Class<? extends Permission> acl)
...
 
+    public static TransactionAuditProvider.TransactionAuditEvent createTransactionAuditEvent(Container container, QueryService.AuditAction auditAction)
+    {
+        long auditId = DbSequenceManager.get(ContainerManager.getRoot(), DB_SEQUENCE_NAME).next();
+        return new TransactionAuditProvider.TransactionAuditEvent(container.getId(), auditAction, auditId);
+    }
+
+    public static void addTransactionAuditEvent(DbScope.Transaction transaction,  User user, TransactionAuditProvider.TransactionAuditEvent auditEvent)
+    {
+        transaction.addCommitTask(() -> AuditLogService.get().addEvent(user, auditEvent), DbScope.CommitTaskOption.POSTCOMMIT);
```
labkey-jeckels: I'd make this PRECOMMIT instead so that it gets committed in the same transaction as everything else.

audit/src/org/labkey/audit/AuditController.java
```diff
@@ -317,4 +336,125 @@ public void setAuditEventType(String auditEventType)
...
+                    else if (newRecord.containsKey("LSID"))
+                        lsids.add(newRecord.get("LSID"));
+
+                }
+            });
+            if (!lsids.isEmpty())
+            {
+                SimpleFilter filter = SimpleFilter.createContainerFilter(container);
+                filter.addCondition(FieldKey.fromParts("LSID"), lsids, CompareType.IN);
+                TableSelector selector = new TableSelector(ExperimentService.get().getTinfoData(), Collections.singleton("RowId"), filter, null);
```
labkey-jeckels: Think we should sort the IDs?

labkey-jeckels (changes requested):
One request and one question.

audit/src/org/labkey/audit/AuditController.java
labkey-susanh: Could do.  It's not at all necessary for my current use case, though.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
labkey-susanh: Done.

labkey-jeckels (approved):

api/src/org/labkey/api/query/AbstractQueryImportAction.java
```diff
@@ -409,12 +413,22 @@ else if (!dataFileDir.exists())
 
             configureLoader(loader);
 
-            int rowCount = importData(loader, file, originalName, ve, getAuditBehaviorType());
+            TransactionAuditProvider.TransactionAuditEvent auditEvent = null;
+            // Check first if the audit behavior has been defined for the table either in code or through XML.
```
XingY, line 417: This logic seems problematic to me. audit behavior is set code, then override by action, and then override by xml. If the code set audit to Summary, but XML set audit to NONE, we should respects XML's NONE and ignore action's audit setting.

audit/src/org/labkey/audit/AuditController.java
```diff
@@ -317,4 +336,125 @@ public void setAuditEventType(String auditEventType)
             this.auditEventType = auditEventType;
         }
     }
+
+    @RequiresPermission(ReadPermission.class)
+    public static class GetTransactionRowIdsAction extends ReadOnlyApiAction<AuditTransactionForm>
```
XingY, line 341: Should this action be moved to samplemanagement instead? The function seems very much tied to samplemanager and its client util selectIdsFromTransactionId is in samplemanagement only (maybe I missed some other usages).

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -5473,6 +5499,32 @@ public void setChecked(boolean checked)
         }
     }
 
+    @RequiresPermission(ReadPermission.class)
+    public static class ReplaceSelectedAction extends MutatingApiAction<SetCheckForm>
```
XingY, line 5489: Is the shared component ComponentWithQueryModels replaceSelections wired up to this new action?

XingY (changes requested):

api/src/org/labkey/api/query/AbstractQueryImportAction.java
labkey-susanh: I think that is the effect here since the logic for getting it from the table will defer to the XML setting unless I'm misreading.

audit/src/org/labkey/audit/AuditController.java
labkey-susanh: I've put it here anticipating that Biologics will also want to use this when they turn on detailed audit logging for samples and because the event types in question are both defined in platform modules.  We will likely move the `selectIdsFromTransactionId` from sampleManagement into ui-components when the time comes.

query/src/org/labkey/query/controllers/QueryController.java
labkey-susanh: Yes.

XingY: hmm, I'm still seeing clearSelections followed by setSelections for replaceSelections in withQueryModels.tsx

api/src/org/labkey/api/query/AbstractQueryImportAction.java
XingY: The problematic check is behaviorType == AuditBehaviorType.NONE . if XML set to NONE, then we should treat it as NONE, instead of setting it to action's auditBehavior

query/src/org/labkey/query/controllers/QueryController.java
labkey-susanh: Ahhh!  Sorry.  I did not change QueryModels.  I shall do so.

api/src/org/labkey/api/data/TableInfo.java
```diff
@@ -576,6 +576,40 @@ default AuditBehaviorType getAuditBehavior()
         return AuditBehaviorType.NONE;
     }
 
+    /**
+     * Retrieves the audit behavior for this table taking into account, in order of precedence:
+     *  - the setting from the XML file (always returned if there is a value set)
+     *  - the value supplied by this table's implementation (returned if not NONE or overrideValue is null)
```
XingY: I'm not understanding the order of precedence here. My understanding is that XML > param override > table implementation. If override has the lowest precedence, then it doesn't sound like it can override anything. If xml is null, table is summary, param is detailed, then the winner should be detailed. But based on this logic, summary would win.

XingY (changes requested):

api/src/org/labkey/api/data/TableInfo.java
labkey-susanh: Ah yes.  'tis fixed.

XingY (approved):

PR #1624, opened 2020-10-09
fix for experimental feature "No Question Marks in URLs"

labkey-matthewb:
#### Rationale
bug fix

### DIFF
--- a/study/src/org/labkey/study/view/subjects.jsp
+++ b/study/src/org/labkey/study/view/subjects.jsp
@@ -77,6 +77,9 @@
     ActionURL subjectUrl = new ActionURL(ParticipantAction.class, container);
     if ("participant2".equals(getViewContext().getActionURL().getAction()))
         subjectUrl = new ActionURL(Participant2Action.class, container);
+    String subjectUrlString = subjectUrl.toString();
+    if (!subjectUrlString.endsWith("?"))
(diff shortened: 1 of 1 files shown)


study/src/org/labkey/study/view/subjects.jsp
```diff
@@ -77,6 +77,9 @@
     ActionURL subjectUrl = new ActionURL(ParticipantAction.class, container);
     if ("participant2".equals(getViewContext().getActionURL().getAction()))
         subjectUrl = new ActionURL(Participant2Action.class, container);
+    String subjectUrlString = subjectUrl.toString();
```
labkey-kevink: Instead of using string concatenation on the client-side, we can use URL objects.  I'll send you a patch.

labkey-kevink (changes requested):

study/src/org/labkey/study/view/subjects.jsp
labkey-adam: 👍

labkey-matthewb: I'll be out for a week, feel free to commit to branch!

labkey-kevink (approved):

labkey-adam (approved):

PR #1628, opened 2020-10-12
20.7 fb elisa vis 7808

labkey-klum:
#### Rationale
This is a follow on story to the initial [ELISA assay improvements](https://docs.google.com/document/d/131Mh8D1HMs89YltSr263w931cbsqBi96B--xmTVu4OU/edit) which added support for multi-plate, multi-well data formats. Some work was done on the StatsService to allow more general handling of curve fit options between assay providers as well as methods to serialize/deserialize fit parameters to JSON.

This PR also introduces a new four parameter fit which is better adapted to handle the types of data we expect to see from the new ELISA changes. The algorithm uses a hybrid simplex-based approach that will evaluate fit errors between two optimization algorithms:
- Nelder-Mead
- Powell
And select the results which use the best fit.

#### Related Pull Requests
https://github.com/LabKey/commonAssays/pull/245

#### Changes
- Four parameter simplex curve fit
- JSON serialization support for curve fits
- AssayPlateProvider support for declaring which types of curve fits are available

### DIFF
--- a/core/src/org/labkey/core/statistics/FourParameterSimplex.java
+++ b/core/src/org/labkey/core/statistics/FourParameterSimplex.java
@@ -100,7 +100,6 @@ public class FourParameterSimplex extends ParameterCurveFit implements Multivari
             _minError = optimum.getValue();
             _bestFit = optimum.getPoint().clone();
             _lastOptimizer = optimizer;
-            //_log.info("optimizer: " + optimizer.toString() + " fit error : " + _minError + " evals: " + optimizer.getEvaluations());
         }
     }
 
--- a/core/src/org/labkey/core/statistics/FourParameterSimplex.java
+++ b/core/src/org/labkey/core/statistics/FourParameterSimplex.java
@@ -1,0 +1,195 @@
+package org.labkey.core.statistics;
+
+import org.apache.commons.math3.analysis.MultivariateFunction;
+import org.apache.commons.math3.exception.TooManyEvaluationsException;
+import org.apache.commons.math3.optim.InitialGuess;
+import org.apache.commons.math3.optim.MaxEval;
+import org.apache.commons.math3.optim.PointValuePair;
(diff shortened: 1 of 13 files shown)


core/src/org/labkey/core/statistics/FourParameterSimplex.java
```diff
@@ -0,0 +1,195 @@
...
+                optimize(powellOptimizer, start);
+            }
+            catch (TooManyEvaluationsException e)
+            {
+                _log.info(e.getMessage());
+            }
+        }
+
+        _log.info("optimizer : " + _lastOptimizer + " fit error : " + calculateFitError(createParams(_bestFit)));
+        _log.info("params " + createParams(_bestFit).toMap());
```
cnathe, line 80: Do you want to leave in this log info?

core/src/org/labkey/core/statistics/FourParameterSimplex.java
```diff
@@ -0,0 +1,195 @@
...
+                simplex,
+                GoalType.MINIMIZE,
+                new InitialGuess(start));
+
+        if (optimum.getValue() < _minError)
+        {
+            _minError = optimum.getValue();
+            _bestFit = optimum.getPoint().clone();
+            _lastOptimizer = optimizer;
+            //_log.info("optimizer: " + optimizer.toString() + " fit error : " + _minError + " evals: " + optimizer.getEvaluations());
```
cnathe: Leave in this log info or remove?

cnathe (approved):

core/src/org/labkey/core/statistics/FourParameterSimplex.java
labkey-klum: yes, keep for now

core/src/org/labkey/core/statistics/FourParameterSimplex.java
labkey-klum: I'll remove this one.

PR #1630, opened 2020-10-14
Item 7959: NAb assay updates to for more data access from NAbSpecimen table (and copied-to-study dataset)

cnathe:
#### Rationale
Ticket [41023](https://www.labkey.org/Atlas/support%20tickets/issues-details.view?issueId=41023): Adding full curve NAb data to export
Feature Request [41520](https://www.labkey.org/Dataspace/Feature%20Requests/issues-details.view?issueId=41520): NAb - dilution data improvements

The HVTN lab would like to be able to export all of the fields associated with the dilutions curves from NAb data from the study folders in addition to their normal “ldo_export” view, explained in ticket 41023, so that the dilution data can be submitted to the SCHARP stats team. They would like to incorporate it as a standard dataset available after the NAb data is copied to study.

Additionally, the DataSpace team has been pulling the dilution data, as well as the fit parameter data, for the CAVD DataSpace portal. Adding dilution data, cell/virus control aggregates, and fit parameter data to be accessible from the copied-to-study dataset would also benefit them as they could simplify their current process that uses the selectRows API against the study dataset and then uses the getNAbRuns API against the assay folder in a multi step process.

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/247

#### Changes
* Save fit parameters as JSON with NAbSpecimen row on new run upload and update via QC exclusion workflow
* Show "FitParameters" column (not in default view column list) for NAbSpecimen table
* DilutionManager helper to get NAbSpecimen row from a specimenLsid value

### DIFF
--- a/assay/api-src/org/labkey/api/assay/dilution/DilutionDataHandler.java
+++ b/assay/api-src/org/labkey/api/assay/dilution/DilutionDataHandler.java
@@ -198,6 +198,7 @@ public abstract class DilutionDataHandler extends AbstractExperimentDataHandler
                 // TODO: factor this out in the nab data handlers
                 props.put("VirusWellGroupName", group.getProperty("VirusWellGroupName"));
 
+                // TODO once 20.7 changes merged in, use fitParams.toJSON helper
                 CurveFit.Parameters fitParams = dilution.getCurveParameters(assayResults.getRenderedCurveFitType());
                 props.put(FIT_PARAMETERS_PROPERTY_NAME, fitParams != null ? new JSONObject(fitParams.toMap()) : null);
             }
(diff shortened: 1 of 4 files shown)


assay/api-src/org/labkey/api/assay/dilution/DilutionDataHandler.java
```diff
@@ -194,6 +197,9 @@ public DilutionDataFileParser(ExpData data, File dataFile, ViewBackgroundInfo in
 
                 // TODO: factor this out in the nab data handlers
                 props.put("VirusWellGroupName", group.getProperty("VirusWellGroupName"));
+
+                CurveFit.Parameters fitParams = dilution.getCurveParameters(assayResults.getRenderedCurveFitType());
+                props.put(FIT_PARAMETERS_PROPERTY_NAME, fitParams != null ? new JSONObject(fitParams.toMap()) : null);
```
labkey-klum, line 203: this is roughly the same code but once we merge the ELISA changes into develop we probably want to use the built in method for Parameters.toJSON. Maybe add a TODO here so we don't forget.

labkey-klum (approved):

assay/api-src/org/labkey/api/assay/dilution/DilutionDataHandler.java
cnathe: TODO added.

PR #1631, opened 2020-10-14
7711: Surface 'add user' button on project users page

RosalineP:
#### Rationale
We want a Project Administrator to be able to add new users to their folder/project. After adding the new user, the Project Admin will then use existing functionality to configure the user's access to the folder/project.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/501

#### Changes
* Update what permissions unlock the 'Add User' button on user grid view
* Make getAddUsersURL() respect container

Note:
The two areas in which the changes within the `populateUserGridButtonBar()` surface are within ShowUsersAction, and DetailsAction. ShowUsersAction is the desired area of change, and DetailsAction specifies its button bar is read-only, and so doesn't surface the 'Add User' button (as wanted).


cnathe (approved):
Looks good. I pulled down the changes and verified the changes locally.

PR #1633, opened 2020-10-15
Bugfix to container path encoding

bbimber:
@labkey-jeckels, see: https://www.labkey.org/ONPRC/Support%20Tickets/issues-details.view?issueId=41591


labkey-jeckels (approved):
We've been working to make our HTML encoding more robust, and I'm guessing that the place that's rendering this URL is now properly encoding it when it hadn't before.

Normally I'd create a LabKey copy of the FB to get TeamCity to build it but this seems very straightforward and safe so I'll merge directly.

labkey-jeckels:
Nevermind, I'll have to create the LabKey repo copy after all, since I can't merge until TeamCity has reported a successful build.

labkey-jeckels:
Looks good, but need the LK side to get TeamCity to trigger a validation build.

https://github.com/LabKey/platform/pull/1634

PR #1634, opened 2020-10-16
Bugfix to container path encoding

labkey-jeckels:
#### Rationale
We want to encode content for HTML exactly once. The ActionURL shouldn't be encoded, it should be encoded when it's rendered as HTML 

#### Related PRs
* https://github.com/LabKey/platform/pull/1633

#### Changes
* Don't pre-render URL as HTML


bbimber:
@labkey-jeckels i keep forgetting to move to the LK side for testing: is there an efficient and/or scriptable way to do that?  or are you doing this through the browser?

labkey-jeckels:
I'm not sure that you'll have permission to create the branch in the LabKey copy of the repo, but here's what I've been doing:

git fetch origin pull/<ID>/head:<fb_branchname>
git checkout <fb_branchname>
git push --set-upstream origin <fb_branchname>

where <ID> is the number of the PR. I think that would be enough to trigger the TeamCity builds and report the status into your original PR, so maybe I don't need to create the PR from the LabKey copy of the FB.

labkey-adam (approved):

PR #1635, opened 2020-10-19
Issue 40239: DataClass: grid not rendering link for column URL property

labkey-kevink:
#### Rationale
DataClass table was manually constructing `ExprColumn` for each of the domain columns instead of using `addWrapColumn` since it can only be used for columns on the same table.  Sadly, the the URL from the wrapped column wasn't propagated to the `ExprColumn`.  The fix is to use the new `wrapColumnFromJoinedTable` helper method which create the wrapped column correctly.


labkey-matthewb:
This looks good.  Would it be possible to add comments to wrapColumn() and wrapColumnFromJoinedTable() to explain the difference the next time some comes across them?

labkey-matthewb (approved):

PR #1638, opened 2020-10-19
40909: Lookup field doesn't resolve when bulk importing into a sampleset

labkey-klum:
#### Rationale
We allow FK imports on int columns by alternate key column but not for string columns, for example dataset ptid to lsid translations. This change enables that if the raw FK value does not resolve to the target table.

tracking [issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40909)

#### Changes
- For datasets, mark the subject ID column as unique so it can be used as an alternate key
- RemapPostConvertColumn was modified to accept a flag `includePkLookup` which will cause it to attempt a match by primary key before trying by alternate key. We will enable this flag for String to String lookups but not for int lookups. The implementation wasn't as clean as I would like since the fetch signature for PK looks different than AK.


api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -745,7 +746,23 @@ protected Object convert(Object o)
         {
             try
             {
-                return _convertCol.convert(o);
+                Object value =  _convertCol.convert(o);
+                ForeignKey fk = _toCol.getFk();
+                // issue 40909 : allow String columns to resolve lookups by alternate key if the raw lookup fails to resolve
+                if (fk != null && Objects.equals(o, value) && String.class.equals(_toCol.getJavaObjectClass()))
```
labkey-matthewb: getJdbcType().isText()

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -745,7 +746,23 @@ protected Object convert(Object o)
...
+                {
+                    if (fk.getLookupTableInfo() != null)
+                    {
```
labkey-matthewb, line 823: I think _remapper is doing a lot of work to cache lookups.  This is not and TableSelector.getObject() is really expensive.  Wouldn't it make more sense to have a flag on RemapPostConvert to also lookup by PK first, then by AK?  Just like it has a flag called _includeTitleColumn?

labkey-matthewb (changes requested):

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -299,6 +332,39 @@ private Object fetch(Triple<ColumnInfo, ColumnInfo, MultiValuedMap<?,?>> triple,
...
+                Object v = map.get(k);
+                if (v == MISS)
+                    return null;
+                return v;
+            }
+            else
+            {
+                TableSelector ts;
+                if (_allowBulkLoads)
+                    ts = createSelector(pkCol, pkCol);
```
labkey-matthewb: Looks good.  Won't this do a full table select for every miss when _allowBulkLoads == true?  Should we remember if we've done a full select so as to only select once here?

labkey-matthewb (changes requested):

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
labkey-klum: Good catch, fixed [here](https://github.com/LabKey/platform/pull/1638/commits/6e67cdec440ad044d4581484ed59cad34ff3046c).

labkey-matthewb (approved):

PR #1648, opened 2020-10-21
Issue 41454: Problem with adding > 100 vials to a request

labkey-jeckels:
#### Rationale
We've switched our UX to act on all of the rows in a selection, even if they span multiple pages of a grid. This JS-based code wasn't following the same approach, so attempting to add > 100 vials to a request would drop those that aren't visible

#### Changes
* Use API call to get the full selection state to find the selected IDs that aren't on the current grid page


RosalineP (approved):
Having to do js-in-java is truly painful, but looks good, and works. 👍

PR #1649, opened 2020-10-21
Use standard JspBase.getWebappURL() and JspBase.getScriptTag() helper methods

labkey-adam:
#### Rationale
We have helper methods that safely and succinctly create `<script>` and `` tags.

#### Related Pull Requests
* See below

#### Changes
* Use standard JspBase.getWebappURL() and JspBase.getScriptTag() helper methods instead of hand assembling script include tags and image URLs
* Best not to HTML encode SafeToRender objects... might result in double encoding


api/src/org/labkey/api/view/GWTView.jsp
```diff
@@ -34,12 +34,11 @@
...
 <div id="<%= h(bean.getModuleName()) %>-Root" class="<%=h(bean.getLoadingStyleName())%>"></div>
 <%
-String contextPath = request.getContextPath();
 String jsPath = bean.getModuleName() + "/" + bean.getModuleName() + ".nocache.js";
-String hashedPath = contextPath + "/" + jsPath + "?" + PageFlowUtil.getServerSessionHash();
+String hashedPath = "/" + jsPath + "?" + PageFlowUtil.getServerSessionHash();
 %>
 <script id="__gwt_marker_<%=h(bean.getModuleName())%>"></script>
-<script type="text/javascript" src="<%=h(hashedPath)%>"></script>
+<%=getScriptTag(hashedPath)%>
```
labkey-matthewb: Shouldn't this be getScriptTag(jsPath)?  getScriptTag() adds the hash.

labkey-adam: `getScriptTag()` doesn't add the hash if the path contains "nocache" (see `staticResourceUrl()`). So, passing just `jsPath` would remove the hash in this case. I suspect it doesn't matter either way... not like the GWT stuff is changing often.

labkey-adam: Oops... my bad, it skips the hash if the path contains ".cache."

labkey-adam: Fixed

labkey-matthewb (approved):

PR #1650, opened 2020-10-22
40297: fail to generate sample names from expression in legacy wizard UI

labkey-kevink:
#### Rationale
When generating sample names outside of the normal query insertRows/updateRows DataIterator pipeline, we need to increment the table counters.  The "genId" value needs to be incremented for each row and put in the NameGenerator context.  I opted to increment the counters in the NameGenerator itself since there were a few usages that generate sample names outside of "normal" query code path.  The counters are incremented using a mini DataIterator pipeline.

ExperimentController.DeriveSamplesAction switched to use query insertRows to create the derived samples.  However a few code paths still create ExpMaterial manually and need to generate a sample name using the name expression while updating the table counters.

#### TODO
The generated counter values are not being persisted on the table which may mean the counter incrementing should happen outside of NameGenerator.  Other table counters are not implemented yet.

#### Changes
- increment "genId" counter when generating sample names outside query insert/update DataIterator


experiment/src/org/labkey/experiment/api/ExpSampleTypeTestCase.java
```diff
@@ -423,6 +433,90 @@ public void testNameExpression() throws Exception
...
+        }
+
+        // verify via query
+        UserSchema schema = QueryService.get().getUserSchema(user, c, SchemaKey.fromParts("Samples"));
+        TableInfo table = schema.getTable("Samples");
+        Collection<Map<String, Object>> inserted = new TableSelector(table, TableSelector.ALL_COLUMNS, null, new Sort("rowId")).getMapCollection();
+
+        Iterator<Map<String, Object>> iter = inserted.iterator();
+        Map<String, Object> row0 = iter.next();
+//        assertEquals(1, ((Number)row0.get("genId")).intValue());
```
labkey-jeckels: Should this check be removed or enabled?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -543,3 +612,96 @@ public String getName()
...
+    @Override
+    public Map<String, Object> getMap()
+    {
+        return _currentMap;
+    }
+
+    @Override
+    public int getColumnCount()
+    {
+        return _columns.size()-1;
```
labkey-jeckels, line 679: Why subtract one?

labkey-jeckels (approved):
A couple of comments. Looks OK, but would be good if @labkey-matthewb has a chance to review too

labkey-nicka:
@labkey-jeckels Given that [this issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40297) was closed over a year ago and that the underlying files have changed significantly I propose we close this PR and delete this feature branch.

PR #1655, opened 2020-10-23
Add LabBook notification options to EmailOption

labkey-alan:
#### Rationale
We need to add some new options for LabBook email settings

#### Related Pull Requests
* https://github.com/LabKey/labbook/pull/55

#### Changes
* Add LABOOK_NONE and LABOOK_ALL settings to EmailOption

### DIFF
--- a/core/module.properties
+++ b/core/module.properties
@@ -1,6 +1,6 @@
 Name: Core
 ModuleClass: org.labkey.core.CoreModule
-SchemaVersion: 20.002
+SchemaVersion: 20.003
 Label: Administration and Essential Services
 Description: The Core module provides central services such as login, \
     security, administration, folder management, user management, \
--- a/api/src/org/labkey/api/announcements/EmailOption.java
+++ b/api/src/org/labkey/api/announcements/EmailOption.java
@@ -39,7 +39,9 @@ public enum EmailOption
     FILES_INDIVIDUAL(513),
     FILES_DAILY_DIGEST(514),
     SAMPLEMANAGER_NONE(701),
-    SAMPLEMANAGER_ALL(702);
+    SAMPLEMANAGER_ALL(702),
+    LABBOOK_NONE(801),
+    LABBOOK_ALL(802);
 
     private final int value;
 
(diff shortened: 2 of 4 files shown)


labkey-alan:
I think I did this correctly, but this is also my first set of SQL upgrade scripts since 2012 or 2013.

core/module.properties
```diff
@@ -1,6 +1,6 @@
 Name: Core
 ModuleClass: org.labkey.core.CoreModule
-SchemaVersion: 20.002
+SchemaVersion: 20.003
```
labkey-nicka, line 3: I presume you're holding off on bumping the module version in the core module until you're ready to merge?

labkey-nicka: Ah, just saw your PR comment. Yea, I think you want to bump the version in the core module not in the module.properties. @labkey-klum correct me if I'm wrong.

labkey-alan: No, I didn't realize I needed to do that. I haven't done one of these in almost a decade 😀 .

labkey-alan: Where does one bump the module version? I am seeing no other references to `20.002` in the core module.

labkey-klum: Actually, you can do either. Most java modules override `getSchemaVersion` but the core module uses the module.properties file. I'm not exactly sure why this is the case or what the advantage is, but it has always been something that was mildly confusing to me.

labkey-klum (approved):

core/module.properties
labkey-adam: Correct, for core, just bump the SchemaVersion property in module.properties (as I think you've done). That's the only option for non-Java (file-based) modules and we've discussed switching Java modules to manage SchemaVersion in module.properties as well, for consistency. We made the change for core a while back to try it out... but haven't ever migrated the other Java modules.

labkey-nicka: TIL

labkey-alan: alright

labkey-nicka (approved):

PR #1656, opened 2020-10-23
40894: Tooltips in schema browser's tree are too narrow on Safari, truncated in Firefox

labkey-klum:
#### Rationale
Ext4 quick tips were rendering incorrectly on versions of Safari and Firefox. This was an Ext bug with dynamic sizing of tooltips where the box size wasn't calculated correctly for those browsers. Sencha fixed the problem in a subsequent release of Ext but that fix was never backported to 4.2.1

[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40894)

This affect all quick tips and is not limited to the schema browser, anywhere in the application that we used quick tips would show the problem:
- file browser
- views and scripting
- others

#### Changes
- backport the fix via ext-patches and constrain to Safari and Gecko user agents.

### DIFF
--- a/api/webapp/ext-4.2.1/ext-patches.js
+++ b/api/webapp/ext-4.2.1/ext-patches.js
@@ -1061,11 +1061,11 @@ Ext4.override(Ext4.tree.Column, {
 });
 
 // issue : 40894 quick tips width being set too narrow
-// This is a backport of the fix implemented in Ext 5.0.1 :
-//
+// The problem was fixed in Ext 5.0.1 but not backported to earlier versions :
 // EXTJS-12511 Narrow tooltips in Safari Mac
 // EXTJS-13334 Tooltip is horizontally squished in Safari
 //
+// This override was found in a stackoverflow posting : https://stackoverflow.com/questions/15834689/extjs-4-2-tooltips-not-wide-enough-to-see-contents
 if (Ext4.isSafari || Ext4.isGecko) {
     Ext4.override(Ext4.tip.QuickTip, {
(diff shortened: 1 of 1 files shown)


api/webapp/ext-4.2.1/ext-patches.js
```diff
@@ -1059,3 +1059,44 @@ Ext4.override(Ext4.tree.Column, {
         '</tpl>'
     ]
 });
+
+// issue : 40894 quick tips width being set too narrow
```
labkey-nicka, line 1063: nit: Do you have a link to a sencha issue or somewhere you found this override? This is something we normally try to document for these types of override fixes.

api/webapp/ext-4.2.1/ext-patches.js
```diff
@@ -1059,3 +1059,44 @@ Ext4.override(Ext4.tree.Column, {
...
+        {
+            var me = this;
+
+            me.target = me.target || Ext4.getDoc();
+            me.targets = me.targets || {};
+            me.callParent();
+
+            me.on('move', function ()
+            {
+                var offset = me.hasCls('x-tip-form-invalid') ? 35 : 12,
```
labkey-nicka, line 1082: I'm rusty on my ExtJS, and I'm sure you tested this out that it works, but the Ext4 CSS namespace prefix is `x4-` so I'm curious if this should be updated as well? I'd presume it is always setting the offset to `12` (which may be suitable).

labkey-nicka (approved):
Looks good. Couple questions.

PR #1658, opened 2020-10-26
41618: provenance: delete performance while deleting large assay run

labkey-kevink:
#### Rationale
When deleting a exp.data input or a set of assay result rows in an assay run, we clean up any associated exp.edge and provenance map rows.  Instead of iterating over each ExpData individually, we can use an IN clause to delete all rows at once.  In addition, issuing separate DELETE requests is much faster than using a WHERE clause with OR.

#### Related Pull Requests
* https://github.com/LabKey/provenance/pull/41

#### Changes
- use IN clause to delete ExpData grouped by source run in AbstractAssayTsvDataHandler.beforeDeleteData
- delete run provenance data before deleting ExpData
- avoid using OR when deleting edges and provenance

### DIFF
--- a/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
+++ b/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
@@ -89,6 +89,8 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static java.util.stream.Collectors.toList;
+
 /**
  * User: jeckels
  * Date: Jan 3, 2008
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
```diff
@@ -281,114 +283,130 @@ private void adjustFirstRowOrder(List<Map<String, Object>> dataRows, DataLoader
...
+                {
+                    domain = AbstractAssayProvider.getDomainByPrefix(protocol, ExpProtocol.ASSAY_DOMAIN_DATA);
+                }
+                catch (IllegalStateException ignored)
+                {
+                    domain = null;
+                    // Be tolerant of not finding a domain anymore, if the provider has gone away
+                }
+
+                // TODO: create assayResultLsidSql when provider no longer exists
```
labkey-ankurj, line 351: is this to be done in this PR?

labkey-ankurj (approved):
Looks good, one comment regd TODO

api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
labkey-kevink: No, this is a comment from prior to this change.

PR #1662, opened 2020-10-27
Add an alternate path for the view files generated by webpack so Gradle can use this for caching purposes

labkey-susanh:
#### Rationale
In order to take advantage of gradle caching and better up-to-date checks, we need to separate the directory where files generated by webpack exist from the directory where source code may live.  The changes here introduce an alternate path where view HTML files may be found, namely,  `resources/views/gen`.  Once this is in place, we can modify the webpack build to actually put files in this directory, then modify the gradlePlugin to designate this directory as an output directory.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/98

#### Changes
* Add resources/web/gen as an alternate path for finding HTML views and introduce the corresponding resource cache
* Export a method `getGeneratedViewPath` for getting the path to this new directory


cnathe (approved):
Awesome!

cnathe (approved):

PR #1663, opened 2020-10-27
fix inconsistency between DataLoaderIterator and DataLoader._DataIterator

labkey-matthewb:
#### Rationale
If DataLoader has colums with column.active==false, then the returned _DataIterator does not correctly return values via .get(int)

#### Changes
introduce getActiveColumns() method to make consistency easier
assert that columns line up


labkey-adam (changes requested):
Looks okay to me, but sql.Query$QueryTestCase appears to be failing now

labkey-adam:
FreezerPro tests are green on this branch: https://teamcity.labkey.org/viewLog.html?buildId=1153939&buildTypeId=LabKey_Trunk_LabkeyPremiumTrunk_GitModules_FreezerProPostgres95

labkey-adam (approved):

PR #1666, opened 2020-10-27
Additional pre-cleaning for SearchTest

labkey-tchad:
#### Rationale
Other tests share sample data used by the search test. If they don't clean up after themselves (usually due to failing), the search tests fail when they search expecting no results.

#### Changes
* Delete projects that contain search results that would interfere with `SearchTest`


labkey-danield (approved):
I don't see any technical issues with this. However I am curious how this will prevent future tests from using the same test data and not getting added to this cleaned up code. 
You could change the test data for the search test and then partition it off, or isolate it in some fashion. But then you would have to change all of the search tests. (not fun).
Would it be useful to add a ReadMe file stored with the test data that warns about using the same data as the search tests?

PR #1667, opened 2020-10-27
watchkey check

labkey-matthewb:
#### Rationale
Remove watchkey check that fails with upload module feature.


labkey-adam (changes requested):
Undo the `ModuleLoader` change? I'm assuming that's wasn't meant to be pushed.

labkey-matthewb:
> Undo the `ModuleLoader` change? I'm assuming that's wasn't meant to be pushed.

thanks

labkey-adam (approved):

PR #1673, opened 2020-11-01
Issue 41314: Wikis containing forward slashes get duplicated upon export/import

labkey-adam:
#### Rationale
Wikis with special characters in their names get exported correctly, but WikiImporterFactory gets confused at import time. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41314

### DIFF
--- a/wiki/src/org/labkey/wiki/export/WikiImporterFactory.java
+++ b/wiki/src/org/labkey/wiki/export/WikiImporterFactory.java
@@ -40,6 +40,7 @@ import org.labkey.wiki.WikiSelectManager;
 import org.labkey.wiki.model.Wiki;
 import org.labkey.wiki.model.WikiVersion;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


wiki/src/org/labkey/wiki/export/WikiImporterFactory.java
```diff
@@ -213,18 +218,20 @@ private Wiki importWiki(String name, String title, boolean shouldIndex, boolean
             return wiki;
         }
 
-        private Pair<String, InputStream> findContentStream(VirtualFile dir, String wikiName) throws ImportException, IOException
+        // Currently, content filename is the same as the folder name, plus an extension that determines the type
+        // TODO: We should store the type and/or the content filename
```
labkey-matthewb, line 222: I actually like the current strategy since it allows 1) webdav 2) folder archive 3) module wikis (if we supported such a thing to all support the exact same layout w/o extra files.

labkey-matthewb (approved):

wiki/src/org/labkey/wiki/export/WikiImporterFactory.java
labkey-adam: I'm not proposing any changes to the layout. Just suggesting that we could export a couple additional hints in the wiki.xml file... in addition to name, title, parent, showAttachments, and shouldIndex (which are all required to properly roundtrip) we'd add (optional) folder and contentFilename attributes.

PR #1675, opened 2020-11-02
Check first 100 rows for columns in AbstractQUS._toDataIterator

labkey-martyp:
#### Rationale
Ran into this while testing SND APIs, which don't currently fill in missing columns for every row passed to the API.  Found we're only looking for columns in the first row.  There may be a different fix but thought looking at the first 100 rows would at least increase the probability of getting all the columns while not iterating through very large row sets.

#### Changes
* AbstractQueryUpdateService._toDataIterator iterate first 100 rows looking for column names

### DIFF
--- a/api/src/org/labkey/api/query/AbstractQueryUpdateService.java
+++ b/api/src/org/labkey/api/query/AbstractQueryUpdateService.java
@@ -95,9 +95,12 @@ import java.io.StringReader;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import static java.util.Objects.requireNonNull;
 import static org.labkey.api.audit.TransactionAuditProvider.DB_SEQUENCE_NAME;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -383,7 +386,9 @@ protected DataIterator _toDataIterator(String debugName, List<Map<String, Object
 
         if (rows.size() > 0 && rows.get(0) instanceof ArrayListMap)
         {
-            colNames = ((ArrayListMap)rows.get(0)).getFindMap().keySet();
+            // Check first 100 rows for column names
```
labkey-matthewb, line 389: This doesn't seem quite right.  You can see the "else" case already scans all the rows.  Right now we overload "is arraylist" to mean we can optimize.  I think the right fix is to find the data loader that is returning ArrayListMaps doesn't know all its columns, OR be explicit and have a way for the dataloader  to indicate it does/doesn't know all its columns.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -383,7 +386,9 @@ protected DataIterator _toDataIterator(String debugName, List<Map<String, Object
 
         if (rows.size() > 0 && rows.get(0) instanceof ArrayListMap)
         {
-            colNames = ((ArrayListMap)rows.get(0)).getFindMap().keySet();
+            // Check first 100 rows for column names
+            colNames = rows.stream().limit(100).map(Map::keySet).flatMap(Collection::stream)
```
labkey-jeckels, line 390: If we already have them all in memory as a List, is there any reason to stop at 100? Performance? @labkey-matthewb what do you think?

labkey-matthewb: Hmm, did my previous comment not submit?  It should be there now.

labkey-nicka:
Related, this is why we have the `ensureAllFieldsInAllRows` backstop method on the client.

https://github.com/LabKey/labkey-ui-components/blob/40d1eba99fa23409b4ca0524db61bbe494b459ca/packages/components/src/internal/query/api.ts#L714

PR #1677, opened 2020-11-03
Alphabetize by name in thread dumps

labkey-jeckels:
#### Rationale
In the Admin Console's running threads page, we alphabetize threads to groups like with like. We don't do that for the thread dumps we write to log files.

#### Changes
* Let a TreeMap do the hard work of sorting by name

### DIFF
--- a/api/src/org/labkey/api/util/DebugInfoDumper.java
+++ b/api/src/org/labkey/api/util/DebugInfoDumper.java
@@ -36,10 +36,12 @@ import java.nio.file.Path;
 import java.nio.file.StandardWatchEventKinds;
 import java.text.DecimalFormat;
 import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.Date;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeMap;
 
 /**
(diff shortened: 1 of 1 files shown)


labkey-susanh (approved):

api/src/org/labkey/api/util/DebugInfoDumper.java
```diff
@@ -242,7 +243,7 @@ public static synchronized void dumpThreads(LoggerWriter logWriter)
                 FileUtils.byteCountToDisplaySize(used) + " from a max of " +
                 FileUtils.byteCountToDisplaySize(max) + " (" + DecimalFormat.getInstance().format(used) + " / " + DecimalFormat.getInstance().format(max) + " bytes)");
         logWriter.debug("*********************************************");
-        Map<Thread,StackTraceElement[]> threads = Thread.getAllStackTraces();
+        Map<Thread,StackTraceElement[]> threads = new TreeMap<>(Thread.getAllStackTraces());
```
labkey-adam: Do you need to supply a `Comparator`? Is the natural order of `Thread`s deterministic?

labkey-jeckels: It's not. I should have done more validation than just spot checking. Now consistent with what we do in the Admin Console.

PR #1687, opened 2020-11-04
Issue 41364: Having a date field in an assay run result creates a bogus entry in the materials table

labkey-jeckels:
#### Rationale
The original theory behind creating this sample records was to let you find assay data that was probably derived from the same sample, even if you don't have a sample ID. I doubt it's ever helped anyone, and ends up creating a bunch of bogus records.

#### Changes
* Match against an existing material in the exceedingly unlikely case where it already exists, but otherwise don't bother creating it, and make sure callers handle nulls.


labkey-kevink (approved):
Excellent.  Any test changes required?

labkey-jeckels:
@labkey-kevink  I'll be checking the test results via TeamCity. I hope not, but it's possible an assay test or two will be checking for these as inputs

PR #1696, opened 2020-11-06
Add optional container to BulkEmailer.addMessage

labkey-alan:
#### Rationale
The bulk emailer doesn't associate a container with an email message, even though the MailHelper.send method supports passing a container in for auditing purposes. This PR adds an optional container to addMessage which we thread through to MailerHelper.send. 

#### Related Pull Requests
* https://github.com/LabKey/labbook/pull/66

#### Changes
* Add  optional container arg to BulkEmailer.addMessage

### DIFF
--- a/api/src/org/labkey/api/util/MailHelper.java
+++ b/api/src/org/labkey/api/util/MailHelper.java
@@ -505,7 +505,8 @@ public class MailHelper
      */
     public static class BulkEmailer extends Thread
     {
-        private final Map<Collection<String>, MimeMessage> _map = new HashMap<>(10);
+        private final Map<Collection<String>, MimeMessage> _messageMap = new HashMap<>(10);
+        private final Map<Collection<String>, Container> _containerMap = new HashMap<>(10);
         private final User _user;
 
         // User is for audit purposes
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/util/MailHelper.java
```diff
@@ -505,7 +505,8 @@ public void setTemplate(EmailTemplate template, Container c) throws MessagingExc
      */
     public static class BulkEmailer extends Thread
     {
-        private final Map<Collection<String>, MimeMessage> _map = new HashMap<>(10);
+        private final Map<Collection<String>, MimeMessage> _messageMap = new HashMap<>(10);
+        private final Map<Collection<String>, Container> _containerMap = new HashMap<>(10);
```
labkey-nicka: Rather than caching whole `Container` objects it might be better to cache the container's Id and then lookup the container when running using `ContainerManager.getForId(_containerMap.get(emails));`

labkey-nicka (approved):

api/src/org/labkey/api/util/MailHelper.java
labkey-alan: Done

PR #1698, opened 2020-11-06
20.11 fb study datetime

labkey-matthewb:
#### Rationale
can't use <datatype>timestamp</datatype>, this gets interpreted by SqlDialect and is therefore not right on SQL Server.  see BaseColumnInfo.getJdbcType()

                int type = d.sqlTypeIntFromSqlTypeName(_sqlTypeName);
 
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41766

### DIFF
--- a/study/resources/schemas/study.xml
+++ b/study/resources/schemas/study.xml
@@ -920,7 +920,7 @@
       </column>
       <column columnName="DrawTimestamp">
           <description>The timestamp of specimen collection.</description>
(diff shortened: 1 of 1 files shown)


labkey-tchad (approved):
Could we add some kind of check to `DbSchemaTest` to complain about ‘`timestamp`’ columns for schemas from modules that support both PG and mssql to try to prevent this sort of inconsistency in the future?

PR #1699, opened 2020-11-07
Issue 40482: Prefer using apiKey replacement parameter

labkey-kevink:
#### Rationale
External R scripts include a ${httpSessionId} token replacement that can be either a JSESSIONID or a temporary apiKey depending on the context the script is executed in.  The ${sessionCookieName} can either be "JSESSIONID" or "LabKeyTransformSessionId" to indicate which to use.  To reduce confusion, we are adding a new ${apiKey} token that will always include the temporary apiKey.  The ${httpSessionId} token replacement will continue to work but will print a deprecation warning when it is used.

#### TODO
* Thread warnings back to the client when an assay design is saved that uses transform script with a deprecated token replacement.
* For RReports, should we use: labkey.setDefaults(apiKey=\"${apiKey}\")
* Should we print deprecation warning when using the "LabKeyTransformSessionId" cookie for authentication

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/256
* https://github.com/LabKey/snprcEHRModules/pull/323
* https://github.com/LabKey/testAutomation/pull/525
* https://github.com/LabKey/premium/pull/211

#### Changes
- add new "apiKey" replacement when executing RReport or Assay QC/transform scripts
- warn when saving an assay design with a transform script that uses a deprecated replacement
- warn when executing a transform script with a deprecated replacement
- add a help link to error messages

### DIFF
--- a/api/src/org/labkey/api/action/ApiResponseWriter.java
+++ b/api/src/org/labkey/api/action/ApiResponseWriter.java
@@ -26,9 +26,9 @@ import org.labkey.api.query.ValidationError;
 import org.labkey.api.query.ValidationException;
 import org.labkey.api.util.ExceptionUtil;
 import org.labkey.api.util.ExpectedException;
+import org.labkey.api.util.HelpTopic;
 import org.labkey.api.util.Pair;
 import org.labkey.api.view.HttpStatusException;
-import org.labkey.api.view.NotFoundException;
 import org.springframework.validation.Errors;
 import org.springframework.validation.FieldError;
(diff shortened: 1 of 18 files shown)


labkey-jeckels (approved):

labkey-jeckels:
I'm not too optimistic that the deprecation warnings will spur much action, but I still think it's worth logging during authentication

labkey-adam (approved):

labkey-jeckels (approved):

PR #1700, opened 2020-11-10
ErrorPage Fixes: fix error during partial startup and display exception class

labkey-ankurj:
#### Rationale
This PR fixes two issues: 
* 41680 - Display the exception class in View Details in Execution exception.
* 41673 - Handle startup errors.

### DIFF
--- a/api/src/org/labkey/api/util/ExceptionUtil.java
+++ b/api/src/org/labkey/api/util/ExceptionUtil.java
@@ -898,12 +898,16 @@ public class ExceptionUtil
 
             addDependenciesAndRender(responseStatus, pageConfig, errorView, ex, request, response);
         }
+        catch (ConfigurationException ce)
+        {
+            throw new ConfigurationException(ce.getMessage());
+        }
         catch (Exception e)
         {
-            // config exceptions that occur before jsps have been initialized
-            if (ex instanceof ConfigurationException || null != ModuleLoader.getInstance().getStartupFailure())
+            // non config exceptions like SqlScriptException that occur during startup
+            if (null != ModuleLoader.getInstance().getStartupFailure())
             {
-                throw new ConfigurationException(ex.getMessage());
+                throw new ConfigurationException(ex.getMessage(), ex);
             }
 
             // try to render just the react app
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/util/ExceptionUtil.java
```diff
@@ -899,7 +901,7 @@ private static void renderErrorPage(Throwable ex, int responseStatus, String mes
         catch (Exception e)
         {
             // config exceptions that occur before jsps have been initialized
-            if (ex instanceof ConfigurationException && null != ModuleLoader.getInstance().getStartupFailure())
+            if (ex instanceof ConfigurationException || null != ModuleLoader.getInstance().getStartupFailure())
             {
                 throw new ConfigurationException(ex.getMessage());
```
labkey-jeckels: This will lose the original stack trace. I recommend:

1. Adding a separate catch block for ConfigurationException and simply rethrowing it.
2. Remove the instanceof check here and pass ex as a second argument to the constructor.

api/src/org/labkey/api/util/PageFlowUtil.java
```diff
@@ -2146,10 +2146,13 @@ public static JSONObject jsInitObject(ContainerUser context, @Nullable PageConfi
...
-        json.put("user", User.getUserProps(user, container));
-        if (user.isImpersonated())
-            json.put("impersonatingUser", User.getUserProps(user.getImpersonatingUser(), container));
-        SecurityLogger.outdent();
+        if (null != user)
+        {
+            json.put("user", User.getUserProps(user, container));
+            if (user.isImpersonated())
+                json.put("impersonatingUser", User.getUserProps(user.getImpersonatingUser(), container));
+            SecurityLogger.outdent();
```
labkey-jeckels, line 2154: Either move the indent() inside the if(), or move the outdent() out.

labkey-jeckels (changes requested):

api/src/org/labkey/api/util/ExceptionUtil.java
```diff
@@ -896,12 +898,16 @@ private static void renderErrorPage(Throwable ex, int responseStatus, String mes
 
             addDependenciesAndRender(responseStatus, pageConfig, errorView, ex, request, response);
         }
+        catch (ConfigurationException ce)
+        {
+            throw new ConfigurationException(ce.getMessage());
```
labkey-jeckels: Just have this throw the ce directly - no need to create a new exception (and this current code will lose the original stack trace).

```suggestion
            throw ce;
```

labkey-jeckels (approved):
One more tweak, but no need for further review unless you have questions or hit problems with the suggestion

PR #1701, opened 2020-11-10
41693: survey datetime field not rendering properly

labkey-klum:
#### Rationale
The custom survey component which should render both a date and time portion is throwing a javascript error after you save a value and attempt to reshow the survey. 

#### Related Issue
- [41693](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41693)

#### Changes
It's unclear when this was broken but we were relying on a variable : DATEFORMATS which is no longer defined. The solution is to replace with the standard alternate date formats returned by LABKEY.Utils.getDateAltFormats. I will also add a test team backlog item to get some automation coverage for this component.

### DIFF
--- a/survey/webapp/survey/DateTimeField.js
+++ b/survey/webapp/survey/DateTimeField.js
@@ -290,7 +290,7 @@ Ext4.define('Ext.ux.form.field.DateTime', {
         var formats = [];
         if (format)
             formats.push(format);
-        formats = formats.concat(DATEFORMATS);
+        formats = formats.concat(LABKEY.Utils.getDateAltFormats().split('|'));
 
         var val;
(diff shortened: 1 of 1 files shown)


cnathe (approved):
@labkey-klum I found a reference to the `DATEFORMATS` variable in the LDK module here: https://github.com/LabKey/LabDevKitModules/blob/develop/LDK/resources/web/LDK/ConvertUtils.js

Thinking back, I think this DateTimeField.js, file was based off of a related file from the LDK module (which got the component / file from the Sencha extensions). 

My guess is that this survey usage was always throwing this JS error based on the missing `DATEFORMATS` variable, but we never caught it. Your change here looks to match what's in that LDK utils file.

PR #1702, opened 2020-11-10
Issue 40081: Fix Mothership test failing on sql server because of date coming as string "null" 

labkey-ankurj:
#### Rationale
Check for date as "null" string on configure mothership page.

### DIFF
--- a/mothership/src/org/labkey/mothership/MothershipManager.java
+++ b/mothership/src/org/labkey/mothership/MothershipManager.java
@@ -480,7 +480,7 @@ public class MothershipManager
     {
         Map<String, String> props = getProperties(c);
         String buildDate = props.get(CURRENT_BUILD_DATE_PROP);
-        return  null == buildDate || "null".equals(buildDate) ? null : new Date(DateUtil.parseDateTime(c, buildDate));
+        return  null == buildDate ? null : new Date(DateUtil.parseISODateTime(buildDate));
     }
 
     private String getStringProperty(Container c, String name)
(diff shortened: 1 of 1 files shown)


mothership/src/org/labkey/mothership/MothershipManager.java
```diff
@@ -480,7 +480,7 @@ public Date getCurrentBuildDate(Container c)
     {
         Map<String, String> props = getProperties(c);
         String buildDate = props.get(CURRENT_BUILD_DATE_PROP);
-        return  null == buildDate ? null : new Date(DateUtil.parseDateTime(c, buildDate));
+        return  null == buildDate || "null".equals(buildDate) ? null : new Date(DateUtil.parseDateTime(c, buildDate));
```
labkey-jeckels: Let's avoid storing the string "null" instead. Looks like it's being introduced by the String.valueOf() in setCurrentBuildDate(). I recommend checking that value for null, and using DateUtil.formatDateTime() instead.

labkey-jeckels (changes requested):

mothership/src/org/labkey/mothership/MothershipManager.java
labkey-adam: Since we're storing a date as a string and expecting to parse it later, please use `formatDateTimeISO8601(Date)` in `setCurrentBuildDate()` and `parseISODateTime(String)` in `getCurrentBuildDate()` (or `parseDateTime(String)` if we need to parse non-ISO formats for backward compatibility purposes). The container-based variants should only be used for displaying results in human readable form, not for serializing dates that need to be parsed. Folder display formats can be lossy and there's no guarantee that we'll be able to parse the results.

labkey-ankurj: @labkey-jeckels the initial (when landing on configure mothership page for the first time) "null" string value of property CURRENT_BUILD_DATE_PROP  is not coming from String.valueOf() in setCurrentBuildDate(). Also, looks like it only happens on sql server.

labkey-jeckels (approved):

labkey-adam (approved):

PR #1703, opened 2020-11-11
Fix BulkEmailer NPE

labkey-alan:
#### Rationale
There is an NPE happening when a container is not passed to BulkEmailer, this is causing some test failures. This PR fixes the NPE.

#### Related Pull Requests
* n/a

#### Changes
* Fix NPE in BulkEmailer

### DIFF
--- a/api/src/org/labkey/api/util/MailHelper.java
+++ b/api/src/org/labkey/api/util/MailHelper.java
@@ -519,7 +519,9 @@ public class MailHelper
         public void addMessage(Collection<String> emails, MimeMessage m, @Nullable Container c)
         {
             _messageMap.put(emails, m);
-            _containerMap.put(emails, c.getId());
+
+            if (c != null)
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/util/MailHelper.java
```diff
@@ -547,7 +549,11 @@ public void run()
             {
                 Collection<String> emails = entry.getKey();
                 MimeMessage m = entry.getValue();
-                Container c = ContainerManager.getForId(_containerMap.get(emails));
+                String containerId = _containerMap.get(emails);
+                Container c = null;
+
+                if (containerId != null)
+                    c = ContainerManager.getForId(_containerMap.get(emails));
```
labkey-nicka: nit: `c = ContainerManager.getForId(containerId);`

labkey-nicka (approved):

PR #1707, opened 2020-11-12
Introduce TransformSession auto-closeable

labkey-adam:
#### Rationale
Introduce `TransformSession` auto-closeable to simplify transform session handling for callers

#### Related Pull Requests
* See below


labkey-kevink (approved):
much nicer!

PR #1710, opened 2020-11-12
Issue 41655: Record only the fields that have changed in a detailed dataset audit log

labkey-susanh:
#### Rationale
We do not need to show all of the fields in a dataset when only some of them have changed.  Instead we capture just the fields that have changed during an update.

#### Changes
* extract method for finding the fields that have been modified into a static method in AuditHandler and use that for getting the old and new records for dataset audit events in which rows have changed.

### DIFF
--- a/api/src/org/labkey/api/audit/AuditHandler.java
+++ b/api/src/org/labkey/api/audit/AuditHandler.java
@@ -1,5 +1,6 @@
 package org.labkey.api.audit;
 
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.data.AuditConfigurable;
 import org.labkey.api.data.Container;
@@ -7,13 +8,16 @@ import org.labkey.api.data.TableInfo;
 import org.labkey.api.gwt.client.AuditBehaviorType;
 import org.labkey.api.query.QueryService;
 import org.labkey.api.security.User;
+import org.labkey.api.util.Pair;
 
+import java.text.SimpleDateFormat;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.Date;
 
 public abstract class AuditHandler
 {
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/audit/AuditHandler.java
```diff
@@ -150,4 +133,46 @@ else if (isExtraAuditField)
...
+        Map<String, Object> modifiedRow = new HashMap<>();
+
+        for (Map.Entry<String, Object> entry : row.entrySet())
+        {
+            boolean isExtraAuditField = extraFieldsToInclude.contains(entry.getKey());
+            if (updatedRow.containsKey(entry.getKey()))
+            {
+                Object newValue = updatedRow.get(entry.getKey());
+                // compare dates using string values to allow for both Date and Timestamp types
+                if (newValue instanceof Date && entry.getValue() != null) {
```
labkey-klum: javascript brace style versus java style?

api/src/org/labkey/api/audit/AuditHandler.java
```diff
@@ -150,4 +133,46 @@ else if (isExtraAuditField)
...
+                    Object oldValue = entry.getValue();
+                    String oldString = oldValue instanceof Date ? formatter.format((Date) oldValue) : oldValue.toString();
+                    if (!newString.equals(oldString) || isExtraAuditField)
+                    {
+                        originalRow.put(entry.getKey(), oldValue);
+                        modifiedRow.put(entry.getKey(), newValue);
+                    }
+                }
+                else
+                    if (!Objects.equals(entry.getValue(), newValue) || isExtraAuditField)
```
labkey-klum: strange newline here

api/src/org/labkey/api/audit/AuditHandler.java
```diff
@@ -150,4 +133,46 @@ else if (isExtraAuditField)
...
+
+        for (Map.Entry<String, Object> entry : row.entrySet())
+        {
+            boolean isExtraAuditField = extraFieldsToInclude.contains(entry.getKey());
+            if (updatedRow.containsKey(entry.getKey()))
+            {
+                Object newValue = updatedRow.get(entry.getKey());
+                // compare dates using string values to allow for both Date and Timestamp types
+                if (newValue instanceof Date && entry.getValue() != null) {
+                    SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
```
labkey-klum, line 151: Do we need this specific datetime format ore would : `DateUtil.formatDateTimeISO8601` suffice?

labkey-klum (approved):
Looks fine, although does this mean that existing audit entries will still render unchanged fields and are we okay with that?

api/src/org/labkey/api/audit/AuditHandler.java
labkey-susanh: Fixed.  (So wonderful that these conventions are different for no good reason.)

api/src/org/labkey/api/audit/AuditHandler.java
labkey-susanh: Fixed.

api/src/org/labkey/api/audit/AuditHandler.java
labkey-susanh: I believe this is the format a Timestamp.toString will produce.  I borrowed this from `encodeForDataMap` where it says this: 
```
            if (value instanceof Date)
            {
                // issue: 35002 - normalize Date values to avoid Timestamp/Date toString differences
                // issue: 36472 - use iso format to show date-time values
                SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
                stringMap.put(entry.getKey(), formatter.format((Date) value));
            }
```
Using DateUtil.formatdateTimeISO8601 does not record seconds, and I think we want to use the same value as is encoded in the string.

labkey-susanh:
Updating existing audit logs seems like a bad idea and there does not seem to be a desire to update `dataset-datasetAuditHistory.view` at this point.

PR #1711, opened 2020-11-13
use WeakReference to hold the module of the loaded resources

labkey-matthewb:
#### Rationale
Fix memory leak caused by not freeing empty cache entries in ModuleResourceCache


labkey-matthewb:
Note: The ClientDependency leak still needs to be merged forward from 20.11

api/src/org/labkey/api/module/ModuleResourceCache.java
```diff
@@ -70,25 +71,68 @@
...
+            super(m, q);
+            this.moduleName = m.getName();
+            this.second = second;
+        }
+
+        Module getModule()
+        {
+            return get();
+        }
+        synchronized V getValue()
```
labkey-adam: Why `synchronized`?

labkey-matthewb: First version cleared the value in processQueue().  Changed to remove entry from cache instead.

labkey-adam (approved):

PR #1712, opened 2020-11-13
Issue 41768: Add alias column to exp.data

labkey-kevink:
#### Rationale
Expose the alias column on exp.Data query table which is useful when querying across DataClass types.


experiment/src/org/labkey/experiment/api/ExpDataClassDataTestCase.java
```diff
@@ -374,24 +388,77 @@ private void testInsertAliases(ExpDataClassImpl dataClass, TableInfo table) thro
...
+        assertEquals(new HashSet<>(actualAliases), expectedAliases);
+
+        verifyAliasesViaSelectRows("exp", "Data", expDataRowId, expectedAliases);
+        verifyAliasesViaSelectRows("exp.data", table.getPublicName(), expDataRowId, expectedAliases);
+    }
+
+    private void verifyAliasesViaSelectRows(String schemaName, String queryName, int expDataRowId, Set<String> expectedAliases)
+            throws Exception
+    {
+        String apiKey = SecurityManager.beginTransformSessionApiKey(TestContext.get().getUser());
```
labkey-jeckels: Side note - we should make beginTransformSessionApiKey() return an AutoClosable that can give you the APIKey so we can use try-with-resources

labkey-jeckels (approved):

experiment/src/org/labkey/experiment/api/ExpDataClassDataTestCase.java
labkey-adam: Great idea, @labkey-jeckels ... I did this yesterday 😄 https://github.com/LabKey/platform/pull/1707  Will merge momentarily...

PR #1714, opened 2020-11-13
Fix provenance tests now that we don't auto-create dummy samples for assays

labkey-jeckels:
#### Rationale
The fix for 41364 broke provenance tests, which assumed that uploading assays created dummy samples

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1687
* https://github.com/LabKey/provenance/pull/45

#### Changes
* Create sample types and reference samples in the assay data


labkey-jeckels:
@labkey-susanh and @labkey-ankurj this is file in platform that's needed to fix the provenance tests, which are now passing on TeamCity.

labkey-susanh (approved):
Seems fine to me.

PR #1716, opened 2020-11-14
insertRow() default implementation throws

labkey-adam:
#### Rationale
We'd like all `QueryUpdateService` subclasses to use `DataIterator` for load, import, and insert operations. However, many QueryUpdateService subclasses override `insertRow()` to provide custom handling. We need to migrate these overrides to follow the DataIterator pattern instead.

With this change, the default implementation of `insertRows()` throws and good subclasses that never use insertRows() no longer need to override it. This allows us to easily identify the subclasses that still need migration, a small step toward our goal.

#### Related Pull Requests
* https://github.com/LabKey/harvest/pull/131


labkey-matthewb (commented):
Why the change from Map<Enum,Object> to Map<String,Object>?

labkey-adam:
Not changing any types... this **removes** the `configParameters` parameter completely (unused in the method), but the diff looks a bit odd

labkey-matthewb (approved):

PR #1718, opened 2020-11-16
Deprecate PostgreSQL 9.5

labkey-adam:
#### Rationale
PostgreSQL 9.5 will reach end-of-life on [February 11, 2021](https://www.postgresql.org/support/versioning/), before the LabKey Server 21.3 release. Since 9.5 will no longer be supported, we will deprecate it in 21.3 (allowing its use for this version, but warning administrators to upgrade ASAP).

#### Changes
* Deprecate PostgreSQL 9.5
* Eliminate most references and reduce visibility of unsupported versions
* Correct inaccurate "java version requirement" message by removing it

### DIFF
--- a/api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
+++ b/api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
@@ -72,7 +72,7 @@ import java.util.stream.Collectors;
 // track changes we've implemented for each version over time.
 public abstract class PostgreSql91Dialect extends SqlDialect
 {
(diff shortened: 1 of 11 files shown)


labkey-jeckels (approved):
Does this impact our ability to connect to older versions of Postgres for external schemas?

labkey-adam:
Not really... this merely deprecates PostgreSQL 9.5. Site administrators will get a warning if they're using PG 9.5 as primary or external data source, but that's it. The current plan is to remove support for PG 9.5 in 21.7. We don't have a "primaryDataSource" check for PostgreSQL (we do for SQL Server, to continue allowing 2008 and 2008R2). Are you aware of clients still using 9.5?

PR #1720, opened 2020-11-17
Perf improvement for AtomicDatabaseInteger

labkey-matthewb:
#### Rationale
Perf  improvement for AtomicDatabaseInteger.  Table.update() is expensive when you don't need the generality.


labkey-adam (approved):

labkey-adam:
Actually, just noticed that you can delete _user, since it's no longer used

labkey-matthewb:
> Actually, just noticed that you can delete _user, since it's no longer used

OK, had to fix call in OConnorLabModules

labkey-adam:
Is that the only code that uses these things?

labkey-matthewb:
> Is that the only code that uses these things?

Looks like it.  Usages might have migrated to DbSequenceManager?

labkey-adam:
Ah, right, very different things

PR #1727, opened 2020-11-18
collection of page load perf related fixes.

labkey-matthewb:
#### Rationale
Fixes for page load/render speed
inspired mostly by suggestions from lighthouse

### DIFF
--- a/api/src/org/labkey/api/view/template/PageConfig.java
+++ b/api/src/org/labkey/api/view/template/PageConfig.java
@@ -25,6 +25,8 @@ import org.json.JSONObject;
 import org.labkey.api.data.DataRegion;
 import org.labkey.api.module.Module;
 import org.labkey.api.util.HelpTopic;
+import org.labkey.api.util.HtmlString;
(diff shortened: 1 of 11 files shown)


labkey-kevink (approved):

labkey-adam (approved):

api/src/org/labkey/api/view/template/PageConfig.java
```diff
@@ -340,31 +342,44 @@ private String getCanonicalLink(URLHelper current)
         return null == u ? null : u.getURIString();
     }
 
+    public HtmlString getPreloadTags()
+    {
+        final List<String> fonts = List.of(
```
labkey-nicka, line 347: I'm now seeing this from the Biologics app:

image

I'd like to investigate how to resolve this.

labkey-nicka: > I thought biologics had a completely different page template?

It does use the `app` page template, however, that still ends up calling `getPreloadTags()` in `pageTemplate.jsp` (since it is also used by `AppTemplate`).

PR #1729, opened 2020-11-20
Remove Ext3  dependent date format

labkey-martyp:
#### Rationale
This Date.format does not work on pages with no Ext3. Use Ext4 version

#### Changes
* Use Ext4.Date


labkey-jeckels (changes requested):
Thanks for the fix. Would you mind retargeting for 20.11 though, as this will probably hit other centers too.

labkey-jeckels (approved):

labkey-martyp:
Merging to 20.11 in another PR

PR #1730, opened 2020-11-20
Show diff in audit logs when doing a merge on import if possible

labkey-susanh:
#### Rationale
When importing data, but in particular samples, and indicating that we want to merge existing records, the audit log that is created currently looks like an audit log for an insert instead of an audit log created when using an updateRows call.  We would like the audit logs and the behavior to be similar for these two types of actions when importing from a file and doing detailed audit logging, as we are for the Sample Manager application.  Since we were already querying to get an ExpMaterial in order to use the proper rowId of existing samples in our audit log, this change shouldn't affect the performance.

There was also a desire to change the audit log comment from the "Register or updated" to be "registered" or "updated" as applicable.

#### Changes
* update AuditHandler to check for the existence of oldRow and updatedRow to determine if an update-type audit log can be created
* Add 'created' and 'createdBy' to the dontUpdate columns passed to the TableInsertDataIterator from ExpDataIterators
* update comment from SampleTimelineAuditEvent.

### DIFF
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -51,6 +51,7 @@ import org.labkey.api.exp.api.ExpSampleType;
 import org.labkey.api.exp.api.ExperimentService;
 import org.labkey.api.exp.property.Domain;
 import org.labkey.api.exp.property.DomainProperty;
+import org.labkey.api.exp.query.ExpDataTable;
 import org.labkey.api.query.AbstractQueryUpdateService;
 import org.labkey.api.query.BatchValidationException;
 import org.labkey.api.query.QueryUpdateService;
@@ -87,9 +88,7 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import static org.apache.commons.lang3.StringUtils.equalsIgnoreCase;
-import static org.labkey.api.exp.api.ExperimentJSONConverter.CREATED;
-import static org.labkey.api.exp.api.ExperimentJSONConverter.CREATED_BY;
-import static org.labkey.api.exp.api.ExperimentJSONConverter.LSID;
+
 
 public class ExpDataIterators
 {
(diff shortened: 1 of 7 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -87,6 +87,9 @@
 import java.util.stream.Collectors;
 
 import static org.apache.commons.lang3.StringUtils.equalsIgnoreCase;
+import static org.labkey.api.exp.api.ExperimentJSONConverter.CREATED;
+import static org.labkey.api.exp.api.ExperimentJSONConverter.CREATED_BY;
+import static org.labkey.api.exp.api.ExperimentJSONConverter.LSID;
```
labkey-kevink: These constants are for API serialization.  The usages below are for database persistence so should probably use ExpDataTable.Column constants instead to keep them separate.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -95,8 +92,11 @@
 import java.util.Set;
 
 import static java.util.Objects.requireNonNull;
+import static org.labkey.api.exp.api.ExperimentJSONConverter.CREATED;
```
labkey-kevink: Same here

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -156,15 +158,39 @@ public void addAuditEvent(User user, Container container, AuditBehaviorType audi
...
-                    if (params.containsKey(LSID))
+                    if (newRow.containsKey(LSID))
                     {
-                        ExpMaterial sample = ExperimentService.get().getExpMaterial((String) params.get(LSID));
-                        if (sample != null)
-                            params.put(ROW_ID, sample.getRowId());
+                        try
+                        {
+                            SampleTypeUpdateServiceDI qus = (SampleTypeUpdateServiceDI) getUpdateService();
+                            if (qus != null) {
```
labkey-kevink: brace on newline

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -156,15 +158,39 @@ public void addAuditEvent(User user, Container container, AuditBehaviorType audi
...
+                                if (existingRows != null && !existingRows.isEmpty()) {
+                                    NOT_FOR_UPDATE.forEach(name -> existingRows.get(0).remove(name));
+                                    existingRow = existingRows.get(0);
+                                    rowId = existingRow.get(ROW_ID);
+                                }
+                            }
+                        }
+                        catch (Exception e)
+                        {
+                            ExpMaterial sample = ExperimentService.get().getExpMaterial((String) newRow.get(LSID));
```
labkey-kevink: Catching exception seems overly broad.  When does the exception get thrown and does it make sense to try and recover from it?

labkey-kevink (approved):
Some minor comments

PR #1739, opened 2020-11-26
Audit log changes to container aliases

labkey-jeckels:
#### Rationale
We track other changes to containers, including deletion and creation. Why not capture other useful configuration changes too?

#### Changes
* Create audit events for changes to container aliases
* Stop using arrays in container alias API

### DIFF
--- a/core/src/org/labkey/core/project/projects.jsp
+++ b/core/src/org/labkey/core/project/projects.jsp
@@ -71,7 +71,7 @@
     properties.putAll(me.getModelBean().getPropertyMap());
 
     String containerTypes = properties.get("containerTypes");
-    String noun = !isBlank(containerTypes) && !"project".equalsIgnoreCase(containerTypes) ? "Subfolder" : "Project";
+    String noun = !isBlank(containerTypes) && !"project".equals(containerTypes) ? "Subfolder" : "Project";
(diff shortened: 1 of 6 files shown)


core/src/org/labkey/core/project/projects.jsp
```diff
@@ -71,15 +71,15 @@
     properties.putAll(me.getModelBean().getPropertyMap());
 
     String containerTypes = properties.get("containerTypes");
-    String noun = !isBlank(containerTypes) && !"project".equals(containerTypes) ? "Subfolder" : "Project";
+    String noun = !isBlank(containerTypes) && !"project".equalsIgnoreCase(containerTypes) ? "Subfolder" : "Project";
```
labkey-adam: Looks like this file was meant for your other PR, but no harm done.

labkey-adam (approved):

core/src/org/labkey/core/project/projects.jsp
labkey-adam: Although if this change is important for 20.11 then you'll need to move it

labkey-jeckels: Not all that important either way, but I've moved it to the other branch since it's more closely related there.

PR #1740, opened 2020-11-27
Remove Ext3 reference from search panel

labkey-martyp:
#### Rationale
Date.format does not work on pages with no Ext3. Use Ext4 version

#### Changes
* Use Ext4.Date


labkey-jeckels (approved):
Opened a bug since this is a core platform change in a release branch. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41936

PR #1741, opened 2020-11-28
Issue 41675: deleting exp.object 

labkey-kevink:
#### Rationale
As a part of fixing Issue 37518, we added `objectId` column to the `exp.data`, `exp.material`, and `exp.experimentrun` tables and created an `exp.object` when a new row was inserted.  When deleting an experiment run, the `exp.data` is detached from the run and ExperimentDataHandler.deleteData() is called to clean up any data handler specific stuff.  The underlying issue is that the `exp.object` for the `exp.data` row was deleted.  When the `exp.data` file was used to import into another run, we created a new `exp.object` for the same `exp.data` and ended up getting out of sync with the `objectId`.

The issue only repros if the first assay run is imported from a file in the pipeline root that was not uploaded via WebDAV.

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/263
* https://github.com/LabKey/platform/pull/249

#### Related Issues
* [Issue 37518](https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=37518): deadlock when concurrently inserting samples with lineage

#### Changes
- associate assay import background job with the resulting run
- add FK from exp.data, exp.material, exp.experimentrun to exp.object by LSID and objectId
- don't delete exp.object in assay ExperimentDataHandler subclasses; it will be cleaned up when the exp.data is deleted
- added AssayIntegrationTestCase.java regression test

### DIFF
--- a/assay/src/org/labkey/assay/AssayIntegrationTestCase.java
+++ b/assay/src/org/labkey/assay/AssayIntegrationTestCase.java
@@ -1,0 +1,213 @@
+package org.labkey.assay;
+
+import com.google.common.base.Charsets;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.hamcrest.CoreMatchers;
+import org.junit.Before;
+import org.junit.Test;
+import org.labkey.api.assay.AssayDataCollector;
+import org.labkey.api.assay.AssayDomainService;
+import org.labkey.api.assay.AssayProvider;
+import org.labkey.api.assay.AssayRunCreator;
+import org.labkey.api.assay.AssayRunUploadContext;
+import org.labkey.api.assay.AssayService;
+import org.labkey.api.data.Container;
+import org.labkey.api.data.JdbcType;
+import org.labkey.api.data.PropertyStorageSpec;
+import org.labkey.api.exp.OntologyManager;
+import org.labkey.api.exp.OntologyObject;
+import org.labkey.api.exp.PropertyDescriptor;
+import org.labkey.api.exp.api.ExpData;
+import org.labkey.api.exp.api.ExpExperiment;
+import org.labkey.api.exp.api.ExpMaterial;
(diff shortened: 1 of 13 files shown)


assay/src/org/labkey/assay/AssayIntegrationTestCase.java
```diff
@@ -0,0 +1,213 @@
...
+        AssayDomainService assayDomainService = new AssayDomainServiceImpl(context);
+        GWTProtocol assayTemplate = assayDomainService.getAssayTemplate("General");
+        assayTemplate.setName(ASSAY_NAME);
+        List<GWTDomain<GWTPropertyDescriptor>> domains = assayTemplate.getDomains();
+
+        // clear the batch domain fields
+        GWTDomain<GWTPropertyDescriptor> batchDomain = domains.stream().filter(d -> "Batch Fields".equals(d.getName())).findFirst().orElseThrow();
+        batchDomain.getFields().clear();
+
+        // clear the batch domain fields
```
labkey-jeckels:
```suggestion
        // clear the run domain fields
```

experiment/resources/schemas/dbscripts/sqlserver/exp-20.012-20.013.sql
```diff
@@ -0,0 +1,31 @@
+-- create exp.object for exp.data without one
+INSERT INTO exp.object (objecturi, container)
```
labkey-jeckels, line 2: Are we confident that exp.material and exp.experimentrun won't have orphaned objectid references?

labkey-jeckels (approved):

experiment/resources/schemas/dbscripts/sqlserver/exp-20.012-20.013.sql
labkey-kevink: I reviewed usages and I didn't find any places where we delete exp.object out from under a material or run anywhere.

PR #1742, opened 2020-11-29
Remove unused actions and flag non-existing crawler actions

labkey-adam:
#### Rationale
These two delete actions are unused and will throw if invoked (they are mutating actions that allow GET). Add a tools action that flags non-existent actions referenced by the crawler.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/537

### DIFF
--- a/devtools/src/org/labkey/devtools/ToolsController.java
+++ b/devtools/src/org/labkey/devtools/ToolsController.java
@@ -35,7 +35,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedList;
@@ -410,8 +409,8 @@ public class ToolsController extends SpringActionController
                 addActionIds(actionIds, file);
             }
 
-            List<ControllerActionId> missingModules = new ArrayList<>();
-            List<ControllerActionId> missingActions = new ArrayList<>();
+            Set<ControllerActionId> missingModules = new TreeSet<>();
+            Set<ControllerActionId> missingActions = new TreeSet<>();
 
             for (ControllerActionId actionId : actionIds)
             {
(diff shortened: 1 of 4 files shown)


devtools/src/org/labkey/devtools/ToolsController.java
```diff
@@ -365,4 +378,168 @@ public void addNavTrail(NavTree root)
...
+            public String getAction()
+            {
+                return _action;
+            }
+
+            @Override
+            public String toString()
+            {
+                return "/" + _controller + "/" + _action;
+            }
```
labkey-tchad: Should this use modern `controller-action` style?
```suggestion
            {
                return "/" + _controller + "-" + _action;
            }
```

devtools/src/org/labkey/devtools/ToolsController.java
```diff
@@ -365,4 +378,168 @@ public void addNavTrail(NavTree root)
...
+                if (!file.exists())
+                {
+                    errors.reject(ERROR_MSG, FileUtil.getAbsoluteCaseSensitiveFile(file).getAbsolutePath() + ": path not found!");
+                    return new SimpleErrorView(errors);
+                }
+                addActionIds(actionIds, file);
+            }
+
+            List<ControllerActionId> missingModules = new ArrayList<>();
+            List<ControllerActionId> missingActions = new ArrayList<>();
```
labkey-tchad: Make sets to avoid duplicate errors?
```suggestion
            Set<ControllerActionId> missingModules = new HashSet<>();
            Set<ControllerActionId> missingActions = new HashSet<>();
```

labkey-tchad (approved):

devtools/src/org/labkey/devtools/ToolsController.java
labkey-adam: Yes it should!

devtools/src/org/labkey/devtools/ToolsController.java
labkey-adam: Switched to TreeSet<>

PR #1743, opened 2020-11-29
Issue 41908: ETL columns with special characters (including Benchling WH columns)

labkey-martyp:
#### Rationale
ETLing columns with non-alphanumeric characters like $ in the column name, does not correctly bind the parameter values.  The StatementUtils creation uses the select column name and the ResultSetDataIterator uses only the alias, causing parameter values on a SQL statement to not be filled in correctly.  This PR uses columnInfos from Results metadata for ResultSetDataIterator columnInfos, if Results type of ResultSet passed in.  This is used in StatementDataIterator.init() to match result set values with sql statement parameters.

Verified on PG and MSSQL and ran nearly all major automated tests to ensure no issues.

#### Changes
- In ResultSetDataIterator constructor use Results columnInfos for metadata if Results type passed in for ResultSet.

### DIFF
--- a/api/src/org/labkey/api/dataiterator/QueryDataIteratorBuilder.java
+++ b/api/src/org/labkey/api/dataiterator/QueryDataIteratorBuilder.java
@@ -161,7 +161,7 @@ public class QueryDataIteratorBuilder implements DataIteratorBuilder
         try
         {
             ResultSet rs = qs.select(t, selectCols, _filter, null, _parameters, false);
-            return new ResultSetDataIterator(rs, context, selectCols);
+            return new ResultSetDataIterator(rs, context);
         }
         catch (QueryParseException x)
         {
--- a/api/src/org/labkey/api/dataiterator/QueryDataIteratorBuilder.java
+++ b/api/src/org/labkey/api/dataiterator/QueryDataIteratorBuilder.java
@@ -161,7 +161,7 @@ public class QueryDataIteratorBuilder implements DataIteratorBuilder
         try
         {
             ResultSet rs = qs.select(t, selectCols, _filter, null, _parameters, false);
-            return new ResultSetDataIterator(rs, context);
+            return new ResultSetDataIterator(rs, context, selectCols);
         }
         catch (QueryParseException x)
         {
--- a/api/src/org/labkey/api/dataiterator/ResultSetDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/ResultSetDataIterator.java
@@ -67,12 +67,10 @@ public class ResultSetDataIterator extends AbstractDataIterator implements Scrol
                 ColumnInfo ci = null;
                 if (rs instanceof Results)
                 {
-                    for (ColumnInfo col : ((Results) rs).getFieldMap().values())
+                    // Use field map to get size of columnInfos which is field map size + 1
+                    if (i <= ((Results) rs).getFieldMap().size())
                     {
-                        if (col.getName().equals(rsmd.getColumnName(i)) || col.getAlias().equals(rsmd.getColumnName(i)))
-                        {
-                            ci = col;
-                        }
+                        ci = ((Results) rs).getColumn(i);
                     }
                 }
                 
--- a/api/src/org/labkey/api/dataiterator/ResultSetDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/ResultSetDataIterator.java
@@ -16,6 +16,7 @@
 
 package org.labkey.api.dataiterator;
 
+import org.jetbrains.annotations.Nullable;
 import org.junit.Assert;
 import org.junit.Test;
 import org.labkey.api.data.BaseColumnInfo;
@@ -29,6 +30,7 @@ import org.labkey.api.util.ResultSetUtil;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
+import java.util.Collection;
 
 /**
  * {@link DataIterator} implementation backed by a JDBC {@link ResultSet}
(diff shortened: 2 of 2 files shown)


labkey-jeckels (commented):
This approach looks OK, but a couple of questions:

1. The ResultSet that's passed here will actually be a Results object which already has info about the ColumnInfos. We could consider using that instead of adding another argument.
2. Why create a new BaseColumnInfo instead of just using the selectCol that you've matched?

It would be great to get @labkey-matthewb 's review too.

api/src/org/labkey/api/dataiterator/QueryDataIteratorBuilder.java
```diff
@@ -161,7 +161,7 @@ public DataIterator getDataIterator(DataIteratorContext context)
         try
         {
             ResultSet rs = qs.select(t, selectCols, _filter, null, _parameters, false);
-            return new ResultSetDataIterator(rs, context);
+            return new ResultSetDataIterator(rs, context, selectCols);
```
labkey-matthewb: qs.select() returns a Results (not a ResultSet).  Results already have the correct ColumnInfo associated with each column in the result.

labkey-martyp:
> This approach looks OK, but a couple of questions:
> 
> 1. The ResultSet that's passed here will actually be a Results object which already has info about the ColumnInfos. We could consider using that instead of adding another argument.
> 2. Why create a new BaseColumnInfo instead of just using the selectCol that you've matched?
> 
> It would be great to get @labkey-matthewb 's review too.

Thanks for taking a look @labkey-jeckels 
1. The ResultSet does have columnInfos with it, however when you get the list of columnInfos using ResultSet.getMetaData it only returns columnInfos with the alias as the name.  The underlying actual column name is not accessible.  Those columnInfos are limited in other ways in their metadata, also missing things like parentTable which might have helped here.  PgResultSetMetaData has getBaseColumnName() which works to get the actual column name, but I can't find an equivalent function for sql server.
2. We can use the existing select columnInfos if you think that's appropriate.  They are a child type of BaseColumnInfo (RelationColumnInfo).  I was just trying to be consistent with the BaseColumnInfo type.  Also trying to create the new ColumnInfos free and clear of changes to the passed in SelectCols columnInfos, should they change elsewhere.

labkey-jeckels:
@labkey-martyp for the approach that I'm proposing with #1 you'd have to check if it's an instanceof Results and then you could call getFieldMap().values() to get the Collection<ColumnInfo>.

And I think that all of ColumnInfos are hopefully immutable based on the work that Matt's done previously, so you shouldn't need to be defensive and make your own copy.

labkey-martyp:
> @labkey-martyp for the approach that I'm proposing with #1 you'd have to check if it's an instanceof Results and then you could call getFieldMap().values() to get the Collection.
> 
> And I think that all of ColumnInfos are hopefully immutable based on the work that Matt's done previously, so you shouldn't need to be defensive and make your own copy.

@labkey-jeckels I see what you're saying.  I'll check that out.

labkey-matthewb:
Note that Results has a getColumn(index) method, which might be easier to use.  You wouldn't have to match the ColumnInfo to the right column in the ResultSet.

api/src/org/labkey/api/dataiterator/ResultSetDataIterator.java
```diff
@@ -62,7 +63,21 @@ protected ResultSetDataIterator(ResultSet rs,  DataIteratorContext context)
             _columns = new ColumnInfo[rsmd.getColumnCount()+1];
             _columns[0] = new BaseColumnInfo("_row", JdbcType.INTEGER);
             for (int i=1 ; i<=rsmd.getColumnCount() ; i++)
-                _columns[i] = new BaseColumnInfo(rsmd, i);
+            {
+                ColumnInfo ci = null;
+                if (rs instanceof Results)
+                {
+                    for (ColumnInfo col : ((Results) rs).getFieldMap().values())
```
labkey-matthewb: Doesn't ((Results)rs).getColumn(i) work?

labkey-matthewb (changes requested):

api/src/org/labkey/api/dataiterator/QueryDataIteratorBuilder.java
labkey-martyp: Yes using the Results columnInfos looks like the way to go.  This should be updated now to not pass in the selectCols.

api/src/org/labkey/api/dataiterator/ResultSetDataIterator.java
labkey-martyp: The indexes don't match up very well. Example below.  Results has an extra _rowNumber columnInfo in its columnInfoList, but that's not in its fieldMap or in the ResultSet metadata.  The ResultSet metadata has the container columnInfo which Results does not have.  However sometimes the container is in the Results object.  I can probably piece something together something using indexes and offsets, but it seemed the more reliable way was just to match on name/alias.  Or I can look into making Results columnInfo indexes more closely match the ResultSet metadata columns.

image

labkey-matthewb: Remember that the ResultSet api's are 1-based.  I see (xKey,field1,field2) in the same order on both sides.  It would be really, really broken if those didn't match.

It seems that we don't have the columnInfo's for the "suggested columns" on the query side (I should probably look into that).  Interestingly , that means that this extra column (container) was not explicitly specified by the user in the query, and should probably not participate in the ETL!  QueryService.get().selectResults() has a parameter "strictColumnList" that does not pull in extra columns.

labkey-martyp: @labkey-matthewb Whose idea was it to make ResultSets 1-based?  Ok I updated to do direct access of the columnInfo with Results.getColumn().  All appears to be working now. Running tests on TC

I didn't do anything with strictColumnList but can file a follow up issue if container columnInfo shouldn't be there.  I was just running ETLListTest.testMergeEtl to test.

labkey-matthewb (approved):

PR #1744, opened 2020-11-30
Update path for inputs to jsp compilation tasks to align with Gradle tasks

labkey-susanh:
#### Rationale
To be able to support build caching, the output of each Gradle tasks need to be independent of the input directories.  In release 1.21.0 of the gradle plugins, the JSP compilation and Jar tasks were marked as cacheable and the intermediate directories needed for compiling the JSPs were changed to avoid the overlap with the output directory.  These changes make the same adjustments for the RecompilingJspClassLoader.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/98

#### Changes
* Use jspWebappDir instead of jspTempDir as input to compilation task


labkey-adam (approved):
Reviewed and verified with a few local recompiles

labkey-jeckels (approved):
I verified that it fixed the error for me on develop.

PR #1746, opened 2020-12-01
Epic 8080 : WNPRC Purchasing Module - Initial Setup

labkey-bpatel:
#### Rationale
Initial implementation of WNPRC Purchasing workflow

#### Related Pull Requests
https://github.com/LabKey/wnprc-modules/pull/46
https://github.com/LabKey/ehrModules/pull/95
https://github.com/LabKey/distributions/pull/116

#### Changes
Add EHRPurchasingTemplateType

### DIFF
--- a/api/schemas/domainTemplate.xsd
+++ b/api/schemas/domainTemplate.xsd
@@ -184,4 +184,11 @@
             </xs:extension>
         </xs:complexContent>
     </xs:complexType>
+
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
We really need a new way to distinguish these template types. We shouldn't have to enumerate them all in the main XSD, but we can save that for another day.

PR #1749, opened 2020-12-02
Pipeline log display not showing some logs on ERROR

labkey-martyp:
#### Rationale
The pipeline log display is checking the log file right after executing the job.  There is a race condition between displaying the error and the logger writing the logs to file (caught with intermittent failures on TC).  This PR adds a time check to start the log rechecking process if the job has completed in the last minute.

#### Changes
- Add a check if the pipeline job modified time is within one minute of the current time.  If so, go into the pipeline log updating code.

### DIFF
--- a/pipeline/src/org/labkey/pipeline/status/details.jsp
+++ b/pipeline/src/org/labkey/pipeline/status/details.jsp
@@ -26,6 +26,7 @@
 <%@ page import="java.time.LocalDateTime" %>
 <%@ page import="org.labkey.api.settings.FolderSettingsCache" %>
 <%@ page import="java.time.format.DateTimeFormatter" %>
+<%@ page import="java.util.Date" %>
 <%@ page extends="org.labkey.api.jsp.JspBase" %>
 <%@ taglib prefix="labkey" uri="http://www.labkey.org/taglib" %>
 <%
@@ -39,6 +40,7 @@
     ActionURL showListURL = bean.showListUrl;
     ActionURL showFolderURL = bean.showFolderUrl;
     ActionURL showDataURL = bean.dataUrl;
+    Date modified = bean.modified;
 %>
 <%!
     // keep in sync with JavaScript logTextClass function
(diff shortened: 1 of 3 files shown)


pipeline/src/org/labkey/pipeline/status/details.jsp
```diff
@@ -342,8 +346,13 @@
     scrollLog(false);
 </script>
 
-<%-- fetch updates if the job is active or there was an error reading the log file on first render. --%>
-<% if (status.active || status.log == null || !status.log.success) { %>
+<%-- fetch updates if the job is active, there was an error reading the log file on first render, or the job has
+ended very recently. --%>
+<%
+    LocalDateTime modifiedTime = LocalDateTime.parse(status.modified,
```
labkey-jeckels: Can we change the bean to store the Date object, and format it when we display it, instead of converting to String and then back to a date?

labkey-jeckels (changes requested):

pipeline/src/org/labkey/pipeline/status/details.jsp
labkey-martyp: Yeah I wasn't sure how the Date would serialize or if it was much savings to do that since the Date will have to be formatted in both Java and JS, depending on the view/api called.

labkey-kevink: Yes, using Date objects in the `StatusDetailsBean` is a good option, but you'd need to format it the client as you pointed out.  Alternatively, since this check of the modified field is only used on the initial render, you could add the `modified` field or a method to the `StatusController.DetailsBean`

labkey-kevink (changes requested):

pipeline/src/org/labkey/pipeline/status/details.jsp
labkey-martyp: Good call.  Added modified to DetailsBean to just be passed to the view for initial render.

labkey-jeckels (approved):

labkey-kevink (approved):
Looks good

PR #1755, opened 2020-12-04
Issue 41981: Inject filterGroups from query xmls defined in fk definition of a column in query-importData.view

labkey-ankurj:
#### Rationale
FK definition in query xml has child element filters that can be utilized to filter column values during insert and update and it was not getting applied as filterArray in query-importData.view because it is named filterGroups. This PR converts them to filterArray in Utils.js.

#### Related Pull Requests
* https://github.com/LabKey/tnprc_ehr/pull/397

### DIFF
--- a/api/webapp/clientapi/ext4/Util.js
+++ b/api/webapp/clientapi/ext4/Util.js
@@ -810,13 +810,12 @@
             }
 
             if (l.filterGroups) {
-                const scope = this;
-                let filterArray = [];
+                var filterArray = [];
                 Ext4.each(l.filterGroups, function (filterGroup) {
                     Ext4.each(filterGroup.filters, function (filter) {
-                        let columnName = filter.column;
-                        let value = filter.value;
-                        let operand = scope.getFilterType(filter.operator)
+                        var columnName = filter.column;
+                        var value = filter.value;
+                        var operand = LABKEY.Filter.getFilterTypeForURLSuffix(filter.operator)
 
                         filterArray.push(LABKEY.Filter.create(columnName, value, operand));
                     });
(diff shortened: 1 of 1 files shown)


api/webapp/clientapi/ext4/Util.js
```diff
@@ -1028,6 +1043,108 @@
...
+                case 'memberof':
+                    filter = LABKEY.Filter.Types.MEMBER_OF;
+                    break;
+                case 'exp:childof':
+                    filter = LABKEY.Filter.Types.EXP_CHILD_OF;
+                    break;
+                case 'exp:parentof':
+                    filter = LABKEY.Filter.Types.EXP_PARENT_OF;
+                    break;
+                case 'q':
```
labkey-ankurj: @labkey-jeckels could not find corresponding filtertypes for following operators in queryCustomView.xsd
- inornull
- notinornull
- exp:lineageof
- where

labkey-jeckels:
@labkey-nicka Can you comment on this? For example, can we use const and let in these JS files? How do we synchronize these changes (if at all) with the versioned JS API? Anything else?

api/webapp/clientapi/ext4/Util.js
```diff
@@ -809,6 +809,21 @@
                 config.filterArray = l.filterArray;
             }
 
+            if (l.filterGroups) {
+                const scope = this;
```
labkey-nicka: For consistency I would recommend against using `const` and `let` in this file. Use `var` declarations instead.

api/webapp/clientapi/ext4/Util.js
```diff
@@ -1028,6 +1043,108 @@
         resizeToViewport: function(extContainer, width, height, paddingX, paddingY, offsetX, offsetY)
         {
             LABKEY.ext4.Util.resizeToContainer.apply(this, arguments);
+        },
+
+        getFilterType: function (filterType) {
```
labkey-nicka: There is already a method available to do this type of mapping: `LABKEY.Filter.getFilterTypeForURLSuffix(operator)`. This will return the associated filter type.

api/webapp/clientapi/ext4/Util.js
labkey-ankurj: 👍

api/webapp/clientapi/ext4/Util.js
```diff
@@ -809,6 +809,20 @@
                 config.filterArray = l.filterArray;
             }
 
+            if (l.filterGroups) {
+                var filterArray = [];
```
labkey-jeckels, line 813: Probably best to append this to any existing filterArray on the config object already.

labkey-jeckels (approved):
One more suggestion, but otherwise looks good. I was talking with Nick about semi-related JS API stuff so I know he's on board with the approach

labkey-nicka (approved):

api/webapp/clientapi/ext4/Util.js
labkey-ankurj: yes, I changed this but curious to know when can that happen(both filterGroups and filterArray are present), my understanding is that metadata gets applied either from query.xml or from model/source js file.

labkey-jeckels: Yes, those are the only likely scenarios in the real world. But since this is something that other code could call with whatever crazy combination it wants, when it's no harder to support both, why not?

PR #1761, opened 2020-12-06
Issue tracker pull request links

labkey-kevink:
#### Rationale
The links in the issue tracker pull request text box aren't clickable.  While we're at it, let's make them nice short github pull request references.  Even better, let's attempt to fetch the PR title from github.  The current method of fetching the PR title only will work for public pull requests.

#### Changes
- Render the pull request column as a set of short PR reference links (e.g., platform#1234)
- Attempt to fetch the pull request title


labkey-tchad (commented):
I don't think its worth grabbing the PR title. It often just matches the issue title and there are often multiple PRs with the same title which would just be confusing.

labkey-klum:
I'm all for making the links clickable but it makes me nervous to be issuing additional github requests to pull in information that may not be that useful. At the very least, we should have a way to disable this functionality or just get rid of the piece that tries to pull the title from the PR

labkey-kevink:
Removed the github title fetching.

issues/src/org/labkey/issue/query/IssuesTable.java
```diff
@@ -856,3 +866,91 @@ public String parseUserDisplayName(String part)
...
+            String organization = m.group("org");
+            String project = m.group("project");
+            try
+            {
+                int pullRequestId = Integer.parseInt(m.group("pullId"));
+                return Tuple3.of(organization, project, pullRequestId);
+            }
+            catch (NumberFormatException e)
+            {
+                IssuesTable.LOG.warn("Failed to parse pull request ID: " + url);
```
labkey-tchad: Could we include the issueId in the log message to make these warnings a bit easier to remedy?

labkey-tchad (approved):

PR #1763, opened 2020-12-08
41982: Copy-to-Study Should Support Dates Provided via Lookups

labkey-klum:
#### Rationale
To address this [issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41982), if we don't find the expected assay date column we will accept an alternative date column provided there is only a single column of that type.


labkey-jeckels (approved):
This change looks OK, but can we more strongly link finding an alternative date column to the alternative PTID around line 975? Even if it's just a comment.

Note also that they behave slightly differently in that the first PTID match is used, while we only use the date column if we find exactly one of them.

Do we want to to filter out "created" or "modified" system columns from consideration?

PR #1764, opened 2020-12-09
fix matching of trailing URL in plain text

labkey-matthewb:
#### Rationale
If text ends with a URL, last char of the link is not included.

### DIFF
--- a/api/src/org/labkey/api/util/PageFlowUtil.java
+++ b/api/src/org/labkey/api/util/PageFlowUtil.java
@@ -262,7 +262,7 @@ public class PageFlowUtil
                 case 'm':
                     if (encodeLinks)
                     {
-                        CharSequence sub = s.subSequence(i, s.length() - 1);
+                        CharSequence sub = s.subSequence(i, s.length());
                         if (StringUtilsLabKey.startsWithURL(sub))
                         {
                             Matcher m = urlPatternStart.matcher(sub);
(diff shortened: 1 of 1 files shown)


labkey-matthewb:
Maybe this should be for 20.11??  It breaks the trailing link in some emails from trial servers.

api/src/org/labkey/api/util/PageFlowUtil.java
```diff
@@ -2611,6 +2611,24 @@ private boolean isLegalId(String id)
...
             return !id.isEmpty() && Character.isLetter(id.charAt(0)) && id.replaceAll("[0-9A-Za-z\\-_:.]", "").isEmpty();
         }
+
+        @Test
+        public void testFilterURL()
+        {
+            String html;
+
+            html = filter("click here http://this/is/a/test.view.", true, true);
+            assertTrue(html.contains("href=\"http:"));
```
labkey-jeckels, line 2621: Any reason to not do an assertEquals() on the full HTML result?

labkey-jeckels (approved):
Is this a regression, or a long-standing bug?

It seems well-suited to a 20.11 hotfix, but not knowing which emails are problematic it's hard for me to make the call

api/src/org/labkey/api/util/PageFlowUtil.java
labkey-matthewb: Not really, but this would be less sensitive to any changes like using LinkBuilder or something that changes or adds attributes or styles or something.

labkey-matthewb: > Is this a regression, or a long-standing bug?
> 
> It seems well-suited to a 20.11 hotfix, but not knowing which emails are problematic it's hard for me to make the call

I assume it was this - https://github.com/LabKey/platform/commit/bb9eb5ab573423d894074d2637e2e3fd0f9a36f6#diff-acca962e3bad5eca2b9cda00a7226af0ac44f39ba5f0a789a21dbc05545e5086

labkey-jony:
@labkey-jeckels Do we want to revisit the hotfix idea for 20.11? I just worked with ONPRC and they reported this issue. Thankfully, we have workarounds for it and I was able to confirm with them that eventually upgrading to 21.X will take care of the problem. But I don't know how long they'll be on 20.11 for.

labkey-jeckels:
@labkey-jony they are scheduled to upgrade within a few weeks, so if the workarounds are viable for a bit longer, I'd say we can hold off. But I don't know their scenario or how painful the workarounds are.

labkey-jony:
@labkey-jeckels Workaround is to add a new line and put in a single character, like a dot. I don't think that would be too painful and they seem okay about it.

PR #1765, opened 2020-12-09
convert NAb run details display column to be a standard DetailsColumn

labkey-klum:
#### Rationale
To address this [issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41748), we needed to migrate from the custom display column to the standard DetailsColumn so we get consistent behavior from both export as well as look and feel.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/551
- https://github.com/LabKey/commonAssays/pull/268


cnathe (approved):
The DetailsURL approach looks much cleaner!

PR #1770, opened 2020-12-10
Item 6775: Asynchronous import for samples and sources

XingY:
#### Rationale
A file size limit of up to 5mb has been applied to LKSM currently since importing large volume of samples can take a long time. To allow importing of large files, we will enable asynchronous import in LKSM. This PR adds support for importing sources and samples. Assay data already has back end support for async import, and will be wired up in a separate story.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1770
* https://github.com/LabKey/labkey-api-js/pull/81
* https://github.com/LabKey/labkey-ui-components/pull/410
* https://github.com/LabKey/sampleManagement/pull/427

#### Changes
* added 'useAsync' import param to AbstractQueryImportAction, which allow import in background job
* new QueryImportPipelineJob, which are wired up to experiment-importSamples and importData for samples and sources background import


api/src/org/labkey/api/pipeline/PipelineJob.java
```diff
@@ -454,6 +455,25 @@ public Path getLogFilePath()
         return FileUtil.hasCloudScheme(_logFilePathName) ? FileUtil.stringToPath(getContainer(), _logFilePathName) : new File(FileUtil.createUri(_logFilePathName)).toPath();
     }
 
+    /** Finds a file name that hasn't been used yet, appending ".2", ".3", etc as needed */
+    public static File findUniqueLogFile(File primaryFile, String baseName)
+    {
+        // need to look in both the assay data and archived dirs for any unused log file names (issue 20987)
```
labkey-susanh: The comment could be generalized slightly.

api/src/org/labkey/api/query/AbstractQueryImportAction.java
```diff
@@ -393,10 +406,35 @@ else if (!dataFileDir.exists())
...
+
+                        UserSchema schema = _target.getUserSchema();
+                        if (schema != null)
+                        {
+                            String schemaName = schema.getSchemaName();
+                            String queryName = _target.getName();
+
+                            QueryImportPipelineJob job = new QueryImportPipelineJob(getQueryImportProviderName(), info, root, dataFile, _hasColumnHeaders, multipartfile.getContentType(),
+                                schemaName, queryName, getRenamedColumns(),
+                                _insertOption, behaviorType, _importLookupByAlternateKey, _importIdentity, hasLineageColumns(), getQueryImportDescription());
```
labkey-susanh: wow.  That's quite a constructor.  It may win for the most arguments.

api/src/org/labkey/api/query/AbstractQueryImportAction.java
```diff
@@ -376,7 +389,7 @@ else if (getViewContext().getRequest() instanceof MultipartHttpServletRequest)
                     // can't read the multipart file twice so create temp file (12800)
                     dataFile = File.createTempFile("~upload", multipartfile.getOriginalFilename());
                     PipeRoot root = PipelineService.get().findPipelineRoot(getContainer());
-                    if (null != root && Boolean.parseBoolean(saveToPipeline))
+                    if (null != root && (Boolean.parseBoolean(saveToPipeline) || _useAsync))
```
labkey-susanh, line 393: It seems like saveToPipeline and useAsync are kind of serving the same purpose. Is there a need to have both?  if so, it's probably a good idea to add comments in this method about when to use one or the other.

api/src/org/labkey/api/query/AbstractQueryImportAction.java
```diff
@@ -536,28 +618,33 @@ protected ActionURL getSuccessURL(FORM form)
     /* TODO change prototype to take DataIteratorBuilder, and DataIteratorContext */
     protected int importData(DataLoader dl, FileStream file, String originalName, BatchValidationException errors, @Nullable AuditBehaviorType auditBehaviorType, TransactionAuditProvider.@Nullable TransactionAuditEvent auditEvent) throws IOException
     {
-        if (_target != null)
+        return importDataImpl(dl, _target, _updateService, _insertOption, _importLookupByAlternateKey, _importIdentity, errors, auditBehaviorType, auditEvent, getUser(), getContainer());
+    }
+
+    public static int importDataImpl(DataLoader dl, TableInfo target, QueryUpdateService updateService, QueryUpdateService.InsertOption insertOption, boolean importLookupByAlternateKey, boolean importIdentity, BatchValidationException errors, @Nullable AuditBehaviorType auditBehaviorType, TransactionAuditProvider.@Nullable TransactionAuditEvent auditEvent, User user, Container container) throws IOException
```
labkey-susanh: Same question here about the `Impl` suffix.   The parameters are quite different from the other `importData`, so I would think both could be called `importData`.

api/src/org/labkey/api/query/AbstractQueryImportAction.java
```diff
@@ -438,18 +475,23 @@ else if (!dataFileDir.exists())
...
         }
+
     }
 
     protected void configureLoader(DataLoader loader) throws IOException
+    {
+        configureLoaderImpl(loader, _target, getRenamedColumns(), hasLineageColumns());
+    }
+
+    public static void configureLoaderImpl(DataLoader loader, @Nullable TableInfo target, @Nullable Map<String, String> renamedColumns, boolean includeLineageColumns) throws IOException
```
labkey-susanh: If this is meant to be a public method, can it just be called `configureLoader`?  The set of arguments is quite different than the other `configureLoader`.

api/src/org/labkey/api/query/QueryImportPipelineJob.java
```diff
@@ -0,0 +1,126 @@
...
+        try
+        {
+            setStatus(TaskStatus.running);
+            getLogger().info("Starting importing " + getDescription());
+
+            UserSchema schema = QueryService.get().getUserSchema(getUser(), getContainer(), _schemaName);
+            TableInfo target = schema.getTable(_queryNama, true);
+
+            if (null == target)
+                throw new ServletException("TableInfo not found.");
```
labkey-susanh: I don't generally like the pattern of a method throwing and then immediately catching its own exception (mainly because throwing is relatively expensive), but I guess many of the other methods that are called here can also throw exceptions so this provides an easy, common way to handle all such errors in a common way.  In any case "ServletException" doesn't really seem appropriate here.  A simple Exception seems sufficient since we do nothing special with it.

api/src/org/labkey/api/query/AbstractQueryImportAction.java
XingY: agree it was excessive... refactored so a builder is used

api/src/org/labkey/api/query/AbstractQueryImportAction.java
XingY: saveToPipeline only save file to pipeline, but doesn't use background import. comment added.

api/src/org/labkey/api/query/AbstractQueryImportAction.java
```diff
@@ -393,10 +407,48 @@ else if (!dataFileDir.exists())
...
+                                .setAuditBehaviorType(behaviorType)
+                                .setImportLookupByAlternateKey(_importLookupByAlternateKey)
+                                .setImportIdentity(_importIdentity)
+                                .setHasLineageColumns(hasLineageColumns())
+                                .setJobDescription(getQueryImportDescription());
+
+                            QueryImportPipelineJob job = new QueryImportPipelineJob(getQueryImportProviderName(), info, root, importContextBuilder);
+                            PipelineService.get().queueJob(job);
+
+                            JSONObject response = new JSONObject();
```
labkey-kevink, line 445: It might be nice to return the queued job id to the client so we can check on progress.

`PipelineService.get().getJobId(getUser(), getContainer(), job.getJobGUID());`

labkey-susanh (approved):

api/src/org/labkey/api/query/AbstractQueryImportAction.java
XingY: good idea. done.

PR #1777, opened 2020-12-11
Issue 42023

labkey-adam:
#### Rationale
Issue 42023: Experiment module has implicit dependency on assay module

#### Changes
* null check the AssayService

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -3928,7 +3928,7 @@ public class ExperimentServiceImpl implements ExperimentService
             transaction.commit();
         }
 
-        if (null != assayService)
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -3927,7 +3928,8 @@ public void deleteProtocolByRowIds(Container c, User user, int... selectedProtoc
             transaction.commit();
         }
 
-        AssayService.get().unindexAssays(Collections.unmodifiableCollection(expProtocols));
+        if (null != assayService)
```
Sigmonia:
```suggestion
        if (assayService != null)
```
for consistency with other calls.

Sigmonia (approved):

labkey-klum (approved):

PR #1778, opened 2020-12-11
Item 7342: Add query parameters for GetUserNotificationsAction so a container, type labels and maxRows can be provided

labkey-susanh:
#### Rationale
For displaying notifications within the Sample Management application, we want to be able to show only certain types of notifications for the current container and don't need to display all the notifications. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/412

#### Changes
* Update GetUserNotificationsAction to accept container, typeLabel and maxRows parameters
* Add method for getting notifications by type label
* Add method for marking all of a user's notifications as read without having to provide the rowIds

### DIFF
--- a/api/src/org/labkey/api/admin/notification/NotificationService.java
+++ b/api/src/org/labkey/api/admin/notification/NotificationService.java
@@ -74,6 +74,12 @@ public interface NotificationService
     List<Notification> getNotificationsByType(Container container, @NotNull String type, int notifyUserId, boolean unreadOnly);
 
     /*
+     * Returns a list of notifications for a specific user based on the specified type label (e.g, 'Pipeline').
+     */
+    List<Notification> getNotificationsByTypeLabel(Container container, @NotNull String typeLabel, int notifyUserId, boolean unreadOnly);
+
+
+    /*
      * Returns a notification based on the notification's RowId.
      */
     Notification getNotification(int rowId);
--- a/api/src/org/labkey/api/admin/notification/NotificationService.java
+++ b/api/src/org/labkey/api/admin/notification/NotificationService.java
@@ -74,9 +74,9 @@ public interface NotificationService
     List<Notification> getNotificationsByType(Container container, @NotNull String type, int notifyUserId, boolean unreadOnly);
 
     /*
-     * Returns a list of notifications for a specific user based on the specified type label (e.g, 'Pipeline').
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/admin/notification/NotificationService.java
```diff
@@ -73,6 +73,11 @@ Notification sendMessageForRecipient(Container c, User createdByUser, User recip
      */
     List<Notification> getNotificationsByType(Container container, @NotNull String type, int notifyUserId, boolean unreadOnly);
 
+    /*
+     * Returns a list of notifications for a specific user based on the specified type label (e.g, 'Pipeline').
+     */
+    List<Notification> getNotificationsByTypeLabel(Container container, @NotNull List<String> typeLabels, int notifyUserId, boolean unreadOnly);
```
cnathe: since the param takes a list of type labels, should we make the method name plural?
`getNotificationsByTypeLabels`

api/src/org/labkey/api/admin/notification/NotificationService.java
```diff
@@ -73,6 +73,11 @@ Notification sendMessageForRecipient(Container c, User createdByUser, User recip
      */
     List<Notification> getNotificationsByType(Container container, @NotNull String type, int notifyUserId, boolean unreadOnly);
 
+    /*
+     * Returns a list of notifications for a specific user based on the specified type label (e.g, 'Pipeline').
```
cnathe:
```suggestion
     * Returns a list of notifications for a specific user based on the specified type labels (e.g, 'Pipeline').

```

cnathe (approved):

core/src/org/labkey/core/notification/NotificationController.java
```diff
@@ -234,30 +266,84 @@ public void addNavTrail(NavTree root)
...
+                }
+                i++;
+                if (notification.getReadOn() == null)
+                    unreadCount++;
             }
 
             ApiSimpleResponse response = new ApiSimpleResponse();
             response.put("notifications", notificationList);
+            response.put("totalRows", notifications.size());
+            response.put("unreadCount", unreadCount);
```
XingY, line 310: maybe rename to totalUnreadCount?

labkey-susanh: I think I'll leave this as is since the same field is used in GetUserNotificationsForPanelAction.  Not a hugely compelling argument, I realize.

PR #1780, opened 2020-12-11
8012: Field Editor Row Selection

RosalineP:
#### Rationale
Update component version to support field row selection. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/413
* https://github.com/LabKey/sampleManagement/pull/454

#### Changes
* Update package versions


core/package.json
```diff
@@ -48,8 +48,8 @@
     }
   },
   "dependencies": {
-    "@labkey/api": "1.1.3",
-    "@labkey/components": "1.7.1",
+    "@labkey/api": "1.1.1",
```
cnathe: I don't think we want to revert the @labkey/api version here back down to 1.1.1. I assume this was just a bad merge, right?

cnathe (changes requested):

core/package.json
RosalineP: Oh, yup! Fixed

cnathe (approved):

PR #1783, opened 2020-12-12
Issue 41448: Migrate PropertyType switch statements to enum methods

labkey-adam:
#### Rationale
[Issue 41448: Migrate PropertyType switch statements to enum methods](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41448)

### DIFF
--- a/api/src/org/labkey/api/exp/PropertyType.java
+++ b/api/src/org/labkey/api/exp/PropertyType.java
@@ -882,9 +882,9 @@ public enum PropertyType
         return getFromURI(concept, datatype, RESOURCE);
     }
 
-    abstract public void init(PropertyRow row, Object value);
-    abstract public void setValue(ObjectProperty property, Object value);
-    abstract public Object getValue(ObjectProperty property);
+    abstract protected void init(PropertyRow row, Object value);
+    abstract protected void setValue(ObjectProperty property, Object value);
+    abstract protected Object getValue(ObjectProperty property);
 
     static
     {
--- a/api/src/org/labkey/api/exp/PropertyType.java
+++ b/api/src/org/labkey/api/exp/PropertyType.java
@@ -882,6 +882,7 @@ public enum PropertyType
         return getFromURI(concept, datatype, RESOURCE);
     }
 
+    @Deprecated // Eliminate this along with PropertyRow? Or at least combine with setValue() below.
     abstract protected void init(PropertyRow row, Object value);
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/exp/PropertyType.java
```diff
@@ -525,6 +882,9 @@ public static PropertyType getFromURI(String concept, String datatype)
         return getFromURI(concept, datatype, RESOURCE);
     }
 
+    abstract public void init(PropertyRow row, Object value);
```
labkey-kevink: Can we make these methods protected since they aren't intended for general use?  I'm not sure I totally get the difference between `init` and `setValue`.

api/src/org/labkey/api/exp/PropertyType.java
```diff
@@ -48,410 +50,765 @@
...
+        }
+
+        @Override
+        public Object getValue(ObjectProperty property)
+        {
+            return property.getStringValue();
+        }
+    };
+
+    private final String typeURI;
```
labkey-kevink, line 803: :+1:  for final

api/src/org/labkey/api/exp/PropertyType.java
```diff
@@ -48,410 +50,765 @@
 public enum PropertyType
 {
     BOOLEAN("http://www.w3.org/2001/XMLSchema#boolean", "Boolean", 'f', JdbcType.BOOLEAN, 10, null, CellType.BOOLEAN, Boolean.class, Boolean.TYPE)
-            {
-                @Override
-                protected Object convertExcelValue(Cell cell) throws ConversionException
-                {
-                    return cell.getBooleanCellValue();
-                }
+    {
```
labkey-kevink, line 53: :+1:  for fixing indentation

labkey-kevink (approved):
Looks good.  If we can get rid of insert tab delimited, I think we can do away with PropertyRow since it seems redundant with ObjectProperty.

api/src/org/labkey/api/exp/PropertyType.java
labkey-adam: Now `protected`... good idea!

My first step was this fairly mechanical refactor. But I like the idea of consolidating `init()` and `setValue()`... or (better) eliminating `init()` along with `PropertyRow`. I'll add some comments in the code.

PR #1786, opened 2020-12-14
Issue 22302: Merge user-defined and built-in queries in schema browser

labkey-jeckels:
#### Rationale
Users often don't pick up on the fact that we split queries and tables in the schema browser. And even if they understand that, they don't necessarily know which type a given entry is. Thus, in schemas with many dozens of tables and queries, it's often frustrating to find the one you want.

#### Changes
* Flatten the queries and tables so that they're direct children of their parent schema
* Show child schema list in the schema details panel
* Use icons to distinguish between tables, module-defined queries, and user-defined queries


query/src/org/labkey/query/controllers/GetSchemaQueryTreeAction.java
```diff
@@ -226,7 +207,7 @@ protected JSONObject getSchemaProps(SchemaKey schemaName, QuerySchema schema)
         return schemaProps;
     }
 
-    protected void addQueryToList(SchemaKey schemaName, String qname, String label, String description, boolean hidden, JSONArray list)
+    protected void addQueryToList(SchemaKey schemaName, String qname, String label, String description, boolean hidden, Map<String, JSONObject> list, boolean table, boolean moduleSupplied)
```
labkey-nicka: nit: These new parameters `table` and `moduleSupplied` are unused.

query/src/org/labkey/query/controllers/GetSchemaQueryTreeAction.java
```diff
@@ -38,9 +39,11 @@
 
 import javax.servlet.http.HttpServletResponse;
 import java.util.ArrayList;
+import java.util.Collections;
```
labkey-nicka: nit: Unused import.

query/webapp/query/browser/Browser.js
```diff
@@ -261,6 +258,9 @@ Ext4.define('LABKEY.query.browser.Browser', {
                     callback.call((scope || this), true, schemaNode);
                 }
             });
+            if (selectInTree) {
```
labkey-nicka: Consider doing this in the `expand()` callback so it is guaranteed to be after the node has been expanded.

```js
schemaNode.expand(false, function(node) {
    if (Ext4.isFunction(callback)) {
        callback.call((scope || this), true, node);
    }

    if (selectInTree) {
        tree.getSelectionModel().select(node, false, true);
    }
});
```

query/webapp/query/browser/Tree.js
```diff
@@ -41,6 +41,9 @@ Ext4.define('LABKEY.query.browser.Tree', {
                     {name: 'queryName'},
                     {name: 'queryLabel'},
                     {name: 'text'},
+                    {name: 'icon', convert: function(_, rec) {
+                        if (rec.raw.table === undefined) { return undefined; }
+                        return LABKEY.ActionURL.getContextPath() + '/reports/' + (rec.raw.table ? 'grid.gif' : (rec.raw.moduleSupplied ? 'grid-sql.png' : 'grid-sql-editable.png' )); }},
```
labkey-nicka: I'm not seeing these icons as I'd expect. I think this is due to those unused properties above not being handed down to the client.

labkey-nicka (changes requested):

query/webapp/query/browser/view/QueryDetails.js
```diff
@@ -560,11 +560,18 @@ Ext4.define('LABKEY.query.browser.view.QueryDetails', {
             }]
         }];
 
-        if (queryDetails.isUserDefined && queryDetails.moduleName) {
+        if (queryDetails.isUserDefined) {
+            children.push({
+                tag: 'span',
+                style: 'cursor: default;',
+                html: 'LabKey SQL query' + (queryDetails.moduleName ? ' defined in ' + Ext4.htmlEncode(queryDetails.moduleName) + ' module' : '')
```
labkey-nicka, line 567: nit: Should this use the new `moduleSupplied` property instead of if the `moduleName` has a value or vice versa? Seems like overlapping logic.

query/webapp/query/browser/Browser.js
labkey-jeckels: Done

query/src/org/labkey/query/controllers/GetSchemaQueryTreeAction.java
labkey-jeckels: Simplified to one new parameter - table - and re-introduced the code that sends it to the client

query/webapp/query/browser/view/QueryDetails.js
labkey-jeckels: I removed the moduleSupplied property, and can keep relying on the preexisting moduleName property

query/webapp/query/browser/view/QueryDetails.js
```diff
@@ -560,11 +560,18 @@ Ext4.define('LABKEY.query.browser.view.QueryDetails', {
             }]
         }];
 
-        if (queryDetails.isUserDefined && queryDetails.moduleName) {
+        if (queryDetails.isUserDefined) {
```
labkey-kevink, line 563: Can you check the behavior with linked schemas that expose a table and query in the source container and also has a custom sql query defined in the target (linked schema) container?

labkey-nicka (approved):

PR #1787, opened 2020-12-15
Issue 39785: Don't return error page HTML from APIs that normally return JSON

labkey-jeckels:
#### Rationale
API callers don't like when we send back HTML responses, because they're expecting the content to be JSON, XML, or other structured content. Thus, even when we hit error conditions, we should respond in kind.

#### Changes
* Use an HttpServletRequest attribute to remember which type of response we should send
* Centralize storage of what response format to use
* Capture preferred response as part of checking permissions to ensure we capture even 401 responses

### DIFF
--- a/api/src/org/labkey/api/action/BaseApiAction.java
+++ b/api/src/org/labkey/api/action/BaseApiAction.java
@@ -61,6 +61,7 @@ import java.util.Map;
  */
 public abstract class BaseApiAction<FORM> extends BaseViewAction<FORM>
 {
+    public static final String RESPONSE_FORMAT_PARAMETER_NAME = "respFormat";
     private final Marshaller _marshaller;
 
     private ApiResponseWriter.Format _reqFormat = null;
@@ -116,7 +117,7 @@ public abstract class BaseApiAction<FORM> extends BaseViewAction<FORM>
 
     private void setResponseFormat()
     {
-        ApiResponseWriter.setResponseFormat(getViewContext().getRequest(), ApiResponseWriter.Format.getFormatByName(getViewContext().getRequest().getParameter("respFormat"), ApiResponseWriter.Format.JSON));
+        ApiResponseWriter.setResponseFormat(getViewContext().getRequest(), ApiResponseWriter.Format.getFormatByName(getViewContext().getRequest().getParameter(RESPONSE_FORMAT_PARAMETER_NAME), ApiResponseWriter.Format.JSON));
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/action/BaseApiAction.java
```diff
@@ -106,6 +105,20 @@ private Marshaller findMarshaller()
...
+    {
+        // Set the preferred response format here so we can use it for 401s
+        setResponseFormat();
+
+        super.checkPermissions();
+    }
+
+    private void setResponseFormat()
+    {
+        ApiResponseWriter.setResponseFormat(getViewContext().getRequest(), ApiResponseWriter.Format.getFormatByName(getViewContext().getRequest().getParameter("respFormat"), ApiResponseWriter.Format.JSON));
```
labkey-kevink: Create constant for "respFormat"?

api/src/org/labkey/api/action/BaseApiAction.java
```diff
@@ -106,6 +105,20 @@ private Marshaller findMarshaller()
         return null;
     }
 
+    @Override
+    public void checkPermissions() throws UnauthorizedException
+    {
+        // Set the preferred response format here so we can use it for 401s
```
labkey-kevink, line 112: This feels like a strange place to set the response format.  I get that it has to be done early, but maybe getting the "respFormat" attribute and stashing it in the attribute should be done by SpringActionController.handleRequest?  It does lots of other similar things.

labkey-kevink (approved):
Looks good in general, just a few comments.

api/src/org/labkey/api/action/BaseApiAction.java
labkey-susanh: Is 'responseFormat' really too verbose here?  I read this initially as standing for "respectFormat".

labkey-jeckels: I agree, this is a poorly named parameter. But the name predates this change, and changing it would break backwards compatibility with APIs

labkey-jeckels: I'll introduce a constant. This is the only usage of it though.

api/src/org/labkey/api/action/BaseApiAction.java
labkey-jeckels: BaseApiAction.createResponseWriter() is the only place that has been using this (and now ExceptionUtil). I thought about putting it somewhere else, but non-API actions will ignore it and return HTML, so it didn't seem worth a bigger refactor.

PR #1790, opened 2020-12-16
Issue 42078: TSV export not resolving Lookup Columns by display column name

labkey-martyp:
#### Rationale
Due to this [PR](https://github.com/LabKey/platform/pull/1743), we are now passing through full column infos to ResultSetDataIterator results. This includes the DisplayColumnFactories. TSV export uses the DisplayColumnFactory to populate data for the export but cannot correctly load values with FK by the Lookup names (ex. JobRunId/JobRunId). Need to use NOLOOKUP display factory for TSV export.

#### Changes
- Check for FK on the columns in TSVGridWriter.init and if it exists, use NOLOOKUP DisplayColumnFactory

### DIFF
--- a/api/src/org/labkey/api/data/TSVGridWriter.java
+++ b/api/src/org/labkey/api/data/TSVGridWriter.java
@@ -94,15 +94,7 @@ public class TSVGridWriter extends TSVColumnWriter implements ExportWriter
         List<DisplayColumn> dataColumns = new LinkedList<>();
 
         for (ColumnInfo col : cols)
-        {
-            DisplayColumnFactory df = col.getDisplayColumnFactory();
-            if (col.getFk() != null)
-            {
-                df = ColumnInfo.NOLOOKUP_FACTORY;
-            }
-
-            dataColumns.add(df.createRenderer(col));
-        }
+            dataColumns.add(col.getDisplayColumnFactory().createRenderer(col));
 
         return init(dataColumns);
     }
--- a/api/src/org/labkey/api/data/TSVGridWriter.java
+++ b/api/src/org/labkey/api/data/TSVGridWriter.java
@@ -94,7 +94,15 @@ public class TSVGridWriter extends TSVColumnWriter implements ExportWriter
         List<DisplayColumn> dataColumns = new LinkedList<>();
 
         for (ColumnInfo col : cols)
-            dataColumns.add(col.getDisplayColumnFactory().createRenderer(col));
+        {
+            DisplayColumnFactory df = col.getDisplayColumnFactory();
+            if (col.getFk() != null)
+            {
+                df = ColumnInfo.NOLOOKUP_FACTORY;
+            }
+
+            dataColumns.add(df.createRenderer(col));
+        }
 
         return init(dataColumns);
     }


labkey-jeckels:
I haven't tested this, but it looks like it would change all TSV exports so that they render the raw FK value instead of the display value from the target table.

If you do a TSV export from an issues grid (for example) do you get user names, or user IDs for columns like Assigned To and Created By?

labkey-matthewb:
> I haven't tested this, but it looks like it would change all TSV exports so that they render the raw FK value instead of the display value from the target table.
> 
> If you do a TSV export from an issues grid (for example) do you get user names, or user IDs for columns like Assigned To and Created By?

Marty's change only affects the case where TSVGridWriter is handed a bare Results object (w/o explicit list of DisplayColumn objects).  You would expect the latter to be the common case e.g.

``` Results results = rgn.getResults(view.getRenderContext());```
``` try (TSVGridWriter tsv = new TSVGridWriter(results, getExportColumns(rgn.getDisplayColumns())))```

Still, I'm surprised how many usages there are of ```new TSVGridWriter(Results)```

labkey-matthewb:
The bug report isn't very specific.  It this showing up as a general problem, or a specific code path that could possibly be fixed by calling ```tsvWriter.setApplyFormats(false);```

labkey-martyp:
> The bug report isn't very specific. It this showing up as a general problem, or a specific code path that could possibly be fixed by calling `tsvWriter.setApplyFormats(false);`

The bug was found by this test [failing test](https://teamcity.labkey.org/viewLog.html?buildId=1188150&tab=buildResultsDiv&buildTypeId=LabKey_2011Release_Premium_ModulesSuites_NciPostgres).  I can check out tsvWriter.setApplayFormats.  The issue is that the column name in RenderContext is not the same as the display value for these lookup columns, so when trying to retrieve the actual value for the TSV file, the names are not matching and the value is not found.  I can try out the setApplyFormats path but I believe it still ultimately calls DisplayColumn.getDisplayValue(RenderContext)

labkey-martyp:
> > The bug report isn't very specific. It this showing up as a general problem, or a specific code path that could possibly be fixed by calling `tsvWriter.setApplyFormats(false);`
> 
> The bug was found by this test [failing test](https://teamcity.labkey.org/viewLog.html?buildId=1188150&tab=buildResultsDiv&buildTypeId=LabKey_2011Release_Premium_ModulesSuites_NciPostgres). I can check out tsvWriter.setApplayFormats. The issue is that the column name in RenderContext is not the same as the display value for these lookup columns, so when trying to retrieve the actual value for the TSV file, the names are not matching and the value is not found. I can try out the setApplyFormats path but I believe it still ultimately calls DisplayColumn.getDisplayValue(RenderContext)

I did test this fix out on all the BVT, Dailys, CustomModules, NCI, Nab and DataIntegration tests.

labkey-matthewb:
I see so it looks like an ETL is running to generate a TSV that later code expects to be able to import.  The default mode of TSVGridWriter is _not_ to generate round-trippable files.  It is to generate data that looks like the source 'dataregion' (assuming there is one).  Since this probably isn't the intention of most TSV generated by ETL, the fix is probably to default to TSVGridWriter.setApplyFormats(false) in TransformTask.appendToTargetFile().  This will give higher fidelity output (e.g. floating point and date precision).  

It might make sense to be able to seperately configure ```use lookups``` and ```use formats``` in TSVGridWriter, but they are tied together now.

labkey-matthewb:
PS: I wasn't involved in the initial spec for TSV support in ETL.  I don't know if there is a need for formatted TSV output (doesn't seem too likely)

labkey-martyp:
> I see so it looks like an ETL is running to generate a TSV that later code expects to be able to import. The default mode of TSVGridWriter is _not_ to generate round-trippable files. It is to generate data that looks like the source 'dataregion' (assuming there is one). Since this probably isn't the intention of most TSV generated by ETL, the fix is probably to default to TSVGridWriter.setApplyFormats(false) in TransformTask.appendToTargetFile(). This will give higher fidelity output (e.g. floating point and date precision).
> 
> It might make sense to be able to seperately configure `use lookups` and `use formats` in TSVGridWriter, but they are tied together now.

Indeed it looks like the setApplyFormats solution resolves the issue for this case.  I'll test it out on other suites.

labkey-martyp:
Closed in lieu of this PR

https://github.com/LabKey/dataintegration/pull/69

PR #1796, opened 2020-12-17
42017: Encode Data Region Messages upon writing to DOM

labkey-nicka:
#### Rationale
[Issue 42017](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42017) describes how messages in our `@labkey/components` query grid are being double encoded. This is stemming from those grid's being backed by the selectRows API with the "metadata" flag turned on. When this flag is on, then we supply messages from the (server-side) Data Region in the API response.

Until now, these messages have been HTML encoded. This is not a desirable format as the targeted layout may not want the HTML generated by the server and instead may do layout based on the other properties of the message (e.g. `area` and `type`) as well.

This changes the messages generated by `DataRegion` to not be HTML encoded upon creation and leaves it up to the user of the message to encode these (if they end up displaying them in HTML). As for the Data Region, the rendering has been updated to encode the message content upon writing the Data Region DOM.

Additionally, I'll note that this still works as a fix for [Issue 35895](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=35895) which had previously been fixed by attempting to encode the messages when they're generated (what is being reversed here).

#### Changes
* Remove HTML encoding and HTML layout from Data Region messages.
* For default Data Region rendering encode the message contents as they are written to the DOM.

### DIFF
--- a/api/src/org/labkey/api/data/DataRegion.java
+++ b/api/src/org/labkey/api/data/DataRegion.java
@@ -181,6 +181,11 @@ public class DataRegion extends DisplayElement
     }
     private List<GroupTable> _groupTables = new ArrayList<>();
 
+    /**
+     * Messages that are displayed to the user and included in Query API responses.
+     * These messages should NOT be HTML encoded as the responsibility for encoding is left to the caller.
+     * See https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42017
+     */
     public static class Message
     {
         private String _area;
--- a/api/src/org/labkey/api/data/DataRegion.java
+++ b/api/src/org/labkey/api/data/DataRegion.java
@@ -181,7 +181,6 @@ public class DataRegion extends DisplayElement
     }
     private List<GroupTable> _groupTables = new ArrayList<>();
 
-    /** HTML message to show the user attached to the region. Caller is responsible for HTML encoding as needed */
     public static class Message
     {
         private String _area;
@@ -974,7 +973,7 @@ public class DataRegion extends DisplayElement
                     _errorCreatingResults = true;
                     _showPagination = false;
                     _allowHeaderLock = false;
-                    addMessage(new Message("<span class=\"labkey-error\">" + PageFlowUtil.filter(x.getMessage()) + "</span><br>", MessageType.ERROR, MessagePart.header));
+                    addMessage(new Message(x.getMessage(), MessageType.ERROR, MessagePart.header));
                 }
             }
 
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/data/DataRegion.java
```diff
@@ -181,7 +181,6 @@ public void setGroupHeadings(List<String> groupHeadings)
     }
     private List<GroupTable> _groupTables = new ArrayList<>();
 
-    /** HTML message to show the user attached to the region. Caller is responsible for HTML encoding as needed */
     public static class Message
```
labkey-jeckels, line 190: Please add JavaDoc about expectations that this is isn't supposed to be HTML encoded, and a note on getContent() that the caller is responsible for encoding

labkey-jeckels (approved):

api/src/org/labkey/api/data/DataRegion.java
labkey-nicka: Check that. I'll add docs for `getContent()`.

labkey-nicka: Done.

labkey-matthewb (commented):
These changes don't address that all the current uses currently do filter the messages.  Search for usages of DataRegion.addMessage() and subclasses of DataRegion.MessageSupplier.

labkey-nicka:
> These changes don't address that all the current uses currently do filter the messages. Search for usages of DataRegion.addMessage() and subclasses of DataRegion.MessageSupplier.

@labkey-matthewb thanks for taking a look. I specifically looked at all the usages of `addMessage` and that is why many of the changes are directly inline in a `addMessage` call. Additionally, I did trace the implementors of `MessageSupplier` and that is how I found the change necessary in `QueryView`. That said, I'm coming up with two implementors of `MessageSupplier` and I'd like to know if you are seeing additional ones:

image

labkey-matthewb:
> > These changes don't address that all the current uses currently do filter the messages. Search for usages of DataRegion.addMessage() and subclasses of DataRegion.MessageSupplier.
> 
> @labkey-matthewb thanks for taking a look. I specifically looked at all the usages of `addMessage` and that is why many of the changes are directly inline in a `addMessage` call. Additionally, I did trace the implementors of `MessageSupplier` and that is how I found the change necessary in `QueryView`. That said, I'm coming up with two implementors of `MessageSupplier` and I'd like to know if you are seeing additional ones:
> 
> image

I was expecting more files to be touch, but it looks like a lot of the scattered usages were just text.  I do see a usage in ReportDataRegion that doesn't seem to have been addressed.

labkey-nicka:
@labkey-matthewb I've switched `ReportDataRegion` to serialize the report metadata to JSON rather than writing to a message. A message is subsequently created on the client if report metadata is available.

PR #1797, opened 2020-12-17
Issue 42081: allow wildcard * in sql when joining between two tables containing the same fields

labkey-adam:
#### Rationale
We added support for duplicate column names in LabKey SQL queries when a special flag was passed to `createTable()` or `getTable()`. Client requested that this be made the default behavior. See [Issue 42081: allow wildcard * in sql when joining between two tables containing the same fields](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42081)

#### Changes
* Set the flag to true by default
* Remove parameter from `createTable()` and `getTable()`
* Add new junit cases for that test duplicate column names. Move previously failing tests that are now passing.


labkey-matthewb (approved):
Looks fine, I do think a parse warning would be nice.

PR #1798, opened 2020-12-17
Use JSONB data type for metrics in mothership reporting

labkey-jeckels:
#### Rationale
Servers report a variety of metrics via JSON. We can make much better use of them by leveraging Postgres's JSON data type

#### Changes
* Allow pass-through for Postgres JSON operators and functions
* Drop SQL Server support for the mothership module, as we have no production deployments there
* Add queries that expose the JSON data in a relational-DB-like model for ease of reporting


labkey-jeckels:
@labkey-matthewb I'd appreciate your review of the JSON pass-through approach. Possible areas of concern:

Is there a better, relatively low effort way to pass through to the Postgres JSON operators?

Should we immediately do this passthrough more centrally in Query so that it's available in all Postgres instances instead of just the few ones that have the Mothership module?

Do want to capture the data type more directly in JdbcType or other enum-like metadata?

Do we want to invest in SQL Server JSON support now? There's overlap in functionality but it's far from a 1:1 simple mapping.

mothership/src/org/labkey/mothership/MothershipModule.java
```diff
@@ -68,14 +77,75 @@ public String getName()
...
     }
 
     @Override
     protected void init()
     {
         addController("mothership", MothershipController.class);
         MothershipSchema.register(this);
+
+
+        if (MothershipManager.get().getSchema().getSqlDialect().isPostgreSQL())
```
labkey-matthewb: You can omit this check if you pass in the PostgresDialect as the last argument to registerPassthroughMethod().  Then it will work for external postgres schemas regardless of the primary database type.

labkey-matthewb: Is there a reason to limit this to servers that have the mothership module?  Obviously these are global functions once they are registered.

mothership/src/org/labkey/mothership/MothershipModule.java
```diff
@@ -68,14 +77,75 @@ public String getName()
...
+                @Override
+                public SQLFragment getSQL(SqlDialect dialect, SQLFragment[] arguments)
+                {
+                    SQLFragment rawOperator = arguments[1];
+                    String operatorRawString = rawOperator.getSQL();
+                    if (!rawOperator.getParams().isEmpty() || !operatorRawString.startsWith("'") || !operatorRawString.endsWith("'"))
+                    {
+                        throw new IllegalArgumentException("Unsupported JSON operator: " + rawOperator);
+                    }
+                    SQLFragment transformedOperator = new SQLFragment(operatorRawString.substring(1, operatorRawString.length() - 1));
```
labkey-matthewb: SQL injection

labkey-matthewb: Verify that this is indeed an expected operator

mothership/src/org/labkey/mothership/MothershipModule.java
```diff
@@ -68,14 +77,75 @@ public String getName()
...
+
+
+        if (MothershipManager.get().getSchema().getSqlDialect().isPostgreSQL())
+        {
+            // Wire up JSON and JSONB data type support for Postgres, as described here:
+            // https://www.postgresql.org/docs/9.5/functions-json.html
+            // Ultimately we need to promote these out of the mothership module but this will enable it for JSON metric
+            // reporting for now.
+
+            // Pretend that the JSON operators are a function instead so that we don't need them to be fully supported
```
labkey-matthewb: Probably wouldn't be that hard to update SqlBase.g to accept these operators and just translate it on the spot.  E.g. 
The parser could translate```a #>> b``` into ```json_op(a, '#>>', b)```

labkey-matthewb (changes requested):

mothership/src/org/labkey/mothership/MothershipModule.java
```diff
@@ -68,14 +77,75 @@ public String getName()
...
+                public SQLFragment getSQL(SqlDialect dialect, SQLFragment[] arguments)
+                {
+                    SQLFragment rawOperator = arguments[1];
+                    String operatorRawString = rawOperator.getSQL();
+                    if (!rawOperator.getParams().isEmpty() || !operatorRawString.startsWith("'") || !operatorRawString.endsWith("'"))
+                    {
+                        throw new IllegalArgumentException("Unsupported JSON operator: " + rawOperator);
+                    }
+                    SQLFragment transformedOperator = new SQLFragment(operatorRawString.substring(1, operatorRawString.length() - 1));
+                    return new SQLFragment(arguments[0]).append(transformedOperator).append(arguments[2]);
```
labkey-matthewb: need to rap arguments[0] and arguments[2] in parens "(" and ")".  Precedence of "," in function argument list is different than precedence of the transformedOperator.

mothership/src/org/labkey/mothership/MothershipModule.java
```diff
@@ -68,14 +77,75 @@ public String getName()
...
+            // Pretend that the JSON operators are a function instead so that we don't need them to be fully supported
+            // for query parsing
+            QueryService.get().registerMethod("json_op", new AbstractMethodInfo(JdbcType.VARCHAR)
+            {
+                @Override
+                public SQLFragment getSQL(SqlDialect dialect, SQLFragment[] arguments)
+                {
+                    SQLFragment rawOperator = arguments[1];
+                    String operatorRawString = rawOperator.getSQL();
+                    if (!rawOperator.getParams().isEmpty() || !operatorRawString.startsWith("'") || !operatorRawString.endsWith("'"))
```
labkey-matthewb, line 109: see Method.validate() so that this is a handled/displayed error.

mothership/src/org/labkey/mothership/MothershipModule.java
labkey-jeckels: This limits our exposure while we experiment as very few distributions include it

mothership/src/org/labkey/mothership/MothershipModule.java
labkey-jeckels: Added

mothership/src/org/labkey/mothership/MothershipModule.java
labkey-jeckels: Done

mothership/src/org/labkey/mothership/MothershipModule.java
labkey-jeckels: Method (and QNode, etc) isn't part of the query module's API so it's not available here. I changed to throw a QueryParseException instead, which is a little better but still gets reported poorly. When we promote this to more general usage we can register in Query and improve this.

labkey-matthewb (approved):

labkey-jeckels:
@labkey-tchad I just pushed a change to mark the Mothership tests as Postgres only. Can you adjust the GitHub and/or TeamCity settings to stop triggering them on SQL Server too?

PR #1805, opened 2020-12-18
add DATETIME to _sqlTypeNameMap

labkey-matthewb:
#### Rationale
Can't use TIMESTAMP in schema.xml because it is ambiguous.  So always support DATETIME as a valid name for java.sql.Types.TIMESTAMP.


labkey-adam:
Remove this same mapping from `BaseMicrosoftSqlServerDialect`?

labkey-matthewb:
> Remove this same mapping from `BaseMicrosoftSqlServerDialect`?

I thought about that for 2 seconds, but since that's basically a list of "weird" dialect specific names, it didn't seem out of place (even if redundant).

labkey-adam (approved):

PR #1807, opened 2020-12-19
Issue 41050: ContainerDisplayColumn shows wrong container name in queries

labkey-jeckels:
#### Rationale
ContainerDisplayColumn does its best to resolve the full container path and name. It wasn't working well in some custom queries

#### Changes
* Use ContainerContext which is smarter about finding the right EntityId

### DIFF
--- a/api/src/org/labkey/api/data/ContainerDisplayColumn.java
+++ b/api/src/org/labkey/api/data/ContainerDisplayColumn.java
@@ -27,6 +27,7 @@ import org.labkey.api.action.ApiJsonWriter;
 import org.labkey.api.action.ApiResponseWriter;
 import org.labkey.api.action.ExtendedApiQueryResponse;
 import org.labkey.api.action.NullSafeBindException;
+import org.labkey.api.query.DetailsURL;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.QueryService;
 import org.labkey.api.query.QuerySettings;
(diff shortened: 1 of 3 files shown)


labkey-matthewb:
It's entirely possible I'm missing something, but this is my first reaction...

Other than the fact that the lookup is done in memory instead of SQL.  This seems to act pretty much like a regular display column when _boundColHasEntityId==true.  It seems it would be good to try to minimize or remove the other cases.  I only see three places where _boundColHasEntityId is false.  

1) MsController.ShowAllRunsAction.  I think that's a bug since it is in fact bound to an entityid column.
2) ContainerTable.init("Path").  Since this is just a wrapped column, it would make more sense to wrap "entityid" instead of "name", then it would be like all the other cases.
3) ContainerTable.init("Name").  I don't entirely understand this.  Since the ContainerTable already has a column "DisplayName" which seems to have the same behavior (e.g. it's like container.getTitle()), what's the need for this odd duck?

labkey-jeckels:
I'm sure I'll regret touching this, but I just removed _boundColHasEntityId.

api/src/org/labkey/api/data/ContainerDisplayColumn.java
```diff
@@ -105,69 +94,33 @@ public Object getDisplayValue(RenderContext ctx)
...
-        return keys;
+        return (String)ctx.get(getBoundColumn().getFieldKey());
     }
 
-    private FieldKey getEntityIdFieldKey(RenderContext ctx)
+    private Container getContainer(RenderContext ctx)
     {
-        for(FieldKey fk : getEntityIdFieldKeys())
+        // Issue 41050 - use URL's container context when possible to resolve the referenced container
+        if (getURLExpression() instanceof DetailsURL)
```
labkey-matthewb: Do we expect this to be different than getEntityIdValue()?  If so maybe a comment how/when that can happen, of if not an assert?

labkey-matthewb:
Hopefully, no regrets.  I think we have fewer mysteries when we avoid lookup/displaycolumn/etc that assume much about the state of the world except for the data they directly have control over.

labkey-matthewb (approved):

api/src/org/labkey/api/data/ContainerDisplayColumn.java
labkey-jeckels: Good point - now that we're always wrapping an EntityId column this can be removed. Just pushed a followup

PR #1810, opened 2020-12-21
41982: Copy-to-Study Should Support PTIDs and Dates Provided via Lookups

labkey-klum:
#### Rationale
The fix for this issue was merged into develop earlier, at the request of the client this PR backports the change to 20.11.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/1763

### DIFF
--- a/api/src/org/labkey/api/study/query/PublishResultsQueryView.java
+++ b/api/src/org/labkey/api/study/query/PublishResultsQueryView.java
@@ -33,6 +33,7 @@ import org.labkey.api.data.DataColumn;
 import org.labkey.api.data.DataRegion;
 import org.labkey.api.data.DetailsColumn;
 import org.labkey.api.data.DisplayColumn;
+import org.labkey.api.data.JdbcType;
 import org.labkey.api.data.RenderContext;
 import org.labkey.api.data.ShowRows;
 import org.labkey.api.data.SimpleDisplayColumn;
@@ -86,6 +87,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * User: brittp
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/study/query/PublishResultsQueryView.java
```diff
@@ -997,9 +974,28 @@ protected Object calculateValue(RenderContext ctx)
...
+
         ColumnInfo runIdCol = colInfos.get(runIdFieldKey);
         ColumnInfo assayVisitIDCol = colInfos.get(assayVisitIDFieldKey);
         ColumnInfo assayDateCol = colInfos.get(assayDateFieldKey);
+        if (assayDateCol == null)
+        {
+            // issue 41982 : look for an alternate date column if the standard assay date field does not exist
+            List<ColumnInfo> dateCols = selectColumns.stream()
+                    .filter(c -> JdbcType.TIMESTAMP.equals(c.getJdbcType()) &&
+                            (!c.getName().equalsIgnoreCase("CreatedBy") && !c.getName().equalsIgnoreCase("ModifiedBy")))
```
labkey-jeckels: I didn't notice this earlier, but shouldn't these column names be "Created" and "Modified"? CreatedBy and ModifiedBy shouldn't be TIMESTAMPs

labkey-jeckels (changes requested):

api/src/org/labkey/api/study/query/PublishResultsQueryView.java
labkey-klum: shoot, good catch! The misnamed columns should get fixed on develop when we merge this forward.

labkey-jeckels (approved):

labkey-jeckels (approved):

PR #1811, opened 2020-12-21
42074: Query report charting options

labkey-klum:
#### Rationale
The client has requested the ability to create charts from query reports by exposing the charting button on the data region. This is a simple fix, and AFAIK there wasn't a reason to exclude the button originally and has always been that way.

#### Related Issue
- https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42074


cnathe (approved):
Looks good, I too don't know of a reason why this would have been disabled/removed.

PR #1825, opened 2020-12-23
Issue 41459: Allow user to specify schema/query for EditableGrid in core-components.view

cnathe:
#### Rationale
We have a shared components testing page in the core module. A few of the components (QueryGridPanel, GridPanel, and EditableGridPanel) would benefit from allowing the user/tester to provide a schema/query name in a consistent way so that our automated tests can setup various scenarios using tables or lists with certain field properties.

#### Changes
* factor out SchemaQueryInputProvider.tsx for use in component pages for GridPanel and QueryGridPanel

### DIFF
--- a/core/src/client/LabKeyUIComponentsPage/EditableGridPage.tsx
+++ b/core/src/client/LabKeyUIComponentsPage/EditableGridPage.tsx
@@ -2,7 +2,7 @@
  * Copyright (c) 2019 LabKey Corporation. All rights reserved. No portion of this work may be reproduced in
  * any form or by any electronic or mechanical means without written permission from LabKey Corporation.
  */
-import React from 'react'
+import React, {PureComponent} from 'react'
 import {List, Map} from "immutable";
 import {
     SchemaQuery,
@@ -12,8 +12,11 @@ import {
     EditableGridPanel,
     getQueryGridModel,
     getStateQueryGridModel,
-    gridInit
+    gridInit,
+    QueryGridModel
 } from "@labkey/components";
+
+import {SchemaQueryInputContext, SchemaQueryInputProvider} from "./SchemaQueryInputProvider";
 
 const columnMetadata = Map<string, EditableColumnMetadata>({
     "Name": {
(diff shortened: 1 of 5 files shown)


core/src/client/LabKeyUIComponentsPage/EditableGridPage.tsx
```diff
@@ -71,33 +93,59 @@ export class EditableGridPage extends React.Component<any, any> {
...
+                        {...defaultProps}
+                        model={this.getEditorQueryGridModel()}
+                    />
+                    : <EditableGridPanel
+                        {...defaultProps}
+                        model={editorModelWithData}
+                        columnMetadata={columnMetadata}
+                    />
+                }
+            </>
```
labkey-susanh, line 142: This seems a bit odd/hacky to me that we'll use either the editorModelWithData or getEditroQueryGridModel and yet we display the loading spinner based on one and do the timeout based on the other.  Of course, this isn't actual application code so it's less of a concern, but it still seems odd.

core/src/client/LabKeyUIComponentsPage/SchemaQueryInputProvider.tsx
```diff
@@ -0,0 +1,75 @@
...
+            schemaName: undefined,
+            error: undefined,
+            model: undefined,
+            queryConfig: undefined,
+        };
+
+        onFormChange = (e: ChangeEvent<HTMLInputElement>) => {
+            const { name, value } = e.target;
+            this.setState(() => ({
+                ...this.state,
```
labkey-susanh, line 33: Is this line necessary?

labkey-susanh (approved):

core/src/client/LabKeyUIComponentsPage/SchemaQueryInputProvider.tsx
cnathe: Yes, without it there is a compilation error because of Typescript, related to the line which sets the state var based on the `name` (i.e. `[name]: value`).

core/src/client/LabKeyUIComponentsPage/EditableGridPage.tsx
cnathe: Yes, this is odd indeed. I did remove the loading spinner since the default editable grid model (when no schema/query is provided, which is used to show the features of the editable grid component) doesn't actually take any time to load since it is all inline. I'm going to leave the rest as is for now and we can adjust accordingly after the test team starts making use of this for automated tests.

PR #1826, opened 2020-12-24
Issue 42134: Selecting view metadata on schema/IssueListDef table gives an error

labkey-ankurj:
#### Rationale
EditMetadata link on issue list definitions table has the queryName as the label of the query which is causing the page to be not found.

#### Changes
* return the correct name for IssueListDefTable

### DIFF
--- a/issues/src/org/labkey/issue/query/IssuesListDefTable.java
+++ b/issues/src/org/labkey/issue/query/IssuesListDefTable.java
@@ -110,7 +110,7 @@ public class IssuesListDefTable extends FilteredTable<IssuesQuerySchema>
     private void addAllColumns()
     {
         setDescription("Contains one row for each issue list");
-        setName("Issue List Definitions");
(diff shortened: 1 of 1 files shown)


issues/src/org/labkey/issue/query/IssuesListDefTable.java
```diff
@@ -241,6 +241,12 @@ public ActionURL getImportDataURL(Container container)
         return LINK_DISABLER_ACTION_URL;
     }
 
+    @Override
```
labkey-jeckels: I suggest changing the call to setName() on line 113 to setTitle(). It should then default to using IssueListDef as both the name and publicName. @labkey-klum speak up if you disagree - looks like you added the call to setName("Issue ListDefinitions") back in 2016.

labkey-jeckels (changes requested):

labkey-jeckels (approved):

PR #1827, opened 2020-12-24
Fail startup if Redshift JDBC driver is found in <tomcat>/lib

labkey-adam:
#### Rationale
We are now distributing the Redshift JDBC driver in the Redshift module and need to warn admins if a Redshift driver remains in \<tomcat>/lib, as we already do with jtds.jar, postresql.jar, and mysql.jar. [Issue 42040: Ship Redshift JDBC driver with Redshift module](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42040)

#### Related Pull Requests
* https://github.com/LabKey/redshift/pull/11

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -26,6 +26,7 @@ import org.jetbrains.annotations.Nullable;
 import org.junit.Assert;
 import org.junit.Test;
 import org.labkey.api.cache.Cache;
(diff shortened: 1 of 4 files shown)


labkey-jeckels (approved):
Looks fine, but we will probably want to let modules register the JDBC drivers they include in the future so we don't need to have a hard-coded list in API.

labkey-adam:
Good idea. Might as well do it now.

labkey-jeckels (approved):
Very nice

PR #1829, opened 2020-12-28
Issue 42105: Preserve order and concurrency of logevents

labkey-ankurj:
#### Rationale
Change in session appender during Log4J upgrade changed the server side JS LogEvents to be stored in ConcurrentReferenceHashMap from LinkedList which destroyed the logging order. This PR fixes the problem of preserving log events while having synchronization.

### DIFF
--- a/api/src/org/labkey/api/util/SessionAppender.java
+++ b/api/src/org/labkey/api/util/SessionAppender.java
@@ -69,14 +69,11 @@ public class SessionAppender extends AbstractAppender
         boolean on;
         final Map<LogEvent, String> eventIdMap = Collections.synchronizedMap(new LinkedHashMap<>()
         {
+            // Safeguard against runaway size.
             @Override
-            public String put(LogEvent key, String value)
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/util/SessionAppender.java
```diff
@@ -67,18 +67,18 @@ protected SessionAppender(String name, Filter filter, Layout<? extends Serializa
...
+        final Map<LogEvent, String> eventIdMap = Collections.synchronizedMap(new LinkedHashMap<>()
         {
             @Override
             public String put(LogEvent key, String value)
             {
                 // Safeguard against runaway size. ConcurrentReferenceHashMap does not have removeEldestEntry so have
                 // to randomly remove an entry
                 if (size() > 1000)
-                    remove(0);
+                    remove(eventIdMap.keySet().toArray()[0]);
```
labkey-kevink: Instead of creating the keySet on every put, you can override `removeEldestEntry` to do the same logic.

labkey-kevink (approved):
Looks good.

PR #1834, opened 2020-12-29
Enable merge option for datasets (phase 1)

Sigmonia:
#### Rationale
Enable the Merge option for bulk updating Datasets

TODO: (Phase 2) Improve Audit Logging

#### Changes
* Enabled UI option for the Dataset bulk insert page and form
* Passed importOption and auditLogBehavior parameters through to the DataIterator context
* Added a DetailedAuditLogDataIterator to the DI queue

### DIFF
--- a/study/src/org/labkey/study/model/DatasetDefinition.java
+++ b/study/src/org/labkey/study/model/DatasetDefinition.java
@@ -1609,15 +1609,13 @@ public class DatasetDefinition extends AbstractStudyEntity<DatasetDefinition> im
         @Override
         public void addAuditEvent(User user, Container container, AuditBehaviorType auditBehavior, @Nullable String userComment, QueryService.AuditAction auditAction, Map<String, Object> parameters)
         {
-            if (auditAction != QueryService.AuditAction.MERGE)
+            Map<String,Object> oldRow = null;
+            if (auditAction == QueryService.AuditAction.MERGE)
             {
-                super.addAuditEvent(user, container, auditBehavior, userComment, auditAction, parameters);
-                return;
+                //REVIEW: I'm guessing this is a terrible idea for multiple reasons, but seems to work...
+                String lsid = String.valueOf(parameters.get("lsid"));
+                oldRow = DatasetDefinition.this.getDatasetRow(user, lsid);
             }
-
-            //REVIEW: I'm guessing this is a terrible idea for multiple reasons, but seems to work...
-            String lsid = String.valueOf(parameters.get("lsid"));
-            Map<String,Object> oldRow = DatasetDefinition.this.getDatasetRow(user, lsid);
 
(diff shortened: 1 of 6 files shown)


Sigmonia:
One thing I just noticed is the Created date field's value is getting changed on the update instead of the Modified field. I'll need to look and see how to correct it.

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1586,6 +1590,35 @@ public PHI getMaxContainedPhi()
...
+            }
+        }
+
+        @Override
+        public void addAuditEvent(User user, Container container, AuditBehaviorType auditBehavior, @Nullable String userComment, QueryService.AuditAction auditAction, Map<String, Object> parameters)
+        {
+            Map<String,Object> oldRow = null;
+            if (auditAction == QueryService.AuditAction.MERGE)
+            {
+                //REVIEW: I'm guessing this is a terrible idea for multiple reasons, but seems to work...
```
labkey-susanh, line 1615: This is certainly where we take the performance hit, but it is akin to how we handle the detailed audit logging for samples as well and is, I believe, the best we can do without more significant changes.  What reasons were you seeing other than performance for this to be a bad idea?

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -2096,20 +2129,26 @@ public DataIteratorBuilder getInsertDataIterator(User user, DataIteratorBuilder
         TableInfo table = getTableInfo(user, false);
         boolean needToHandleQCState = table.getColumn(DatasetTableImpl.QCSTATE_ID_COLNAME) != null;
 
+        DataIteratorBuilder dib = in;
```
labkey-susanh, line 2132: The extra variable here doesn't seem necessary.

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1586,6 +1590,35 @@ public PHI getMaxContainedPhi()
...
+
+                StudyServiceImpl.addDatasetAuditEvent(user, DatasetDefinition.this, row, updatedRow);
+            }
+        }
+
+        @Override
+        public void addAuditEvent(User user, Container container, AuditBehaviorType auditBehavior, @Nullable String userComment, QueryService.AuditAction auditAction, Map<String, Object> parameters)
+        {
+            Map<String,Object> oldRow = null;
+            if (auditAction == QueryService.AuditAction.MERGE)
```
labkey-susanh, line 1613: We want to do the detailed logging only when the audit behavior is DETAILED.

labkey-susanh (changes requested):

Sigmonia:
Split into two PRs:
1) for enabling the merge option https://github.com/LabKey/platform/pull/1839
2) Adding the audit logging https://github.com/LabKey/platform/pull/1840

PR #1837, opened 2020-12-30
Don't register generic action class (SimpleAction.class) in static _classToDescriptor map

labkey-matthewb:
#### Rationale
Fix memory leak by not registering SimpleAction.class in the static variable SpringActionController._classToDescriptor

### DIFF
--- a/api/src/org/labkey/api/action/SpringActionController.java
+++ b/api/src/org/labkey/api/action/SpringActionController.java
@@ -139,25 +139,25 @@ public abstract class SpringActionController implements Controller, HasViewConte
         assert null == prev || prev == ad;
     }
 
-    @NotNull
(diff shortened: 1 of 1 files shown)


labkey-adam (approved):

api/src/org/labkey/api/action/SpringActionController.java
```diff
@@ -139,25 +139,25 @@ protected static void registerAction(ActionDescriptor ad)
         assert null == prev || prev == ad;
     }
 
-    @NotNull
     static ActionDescriptor getActionDescriptor(Class<? extends Controller> actionClass)
```
labkey-adam, line 143: Annotate as `@Nullable`?

labkey-adam (approved):

PR #1839, opened 2020-12-31
Enable merge option for datasets (phase 1)

Sigmonia:
#### Rationale
Enable the Merge option for bulk updating Datasets

TODO: (Phase 2) Improve Audit Logging

#### Changes
* Enabled UI option for the Dataset bulk insert page and form
* Passed importOption and auditLogBehavior parameters through to the DataIterator context

### DIFF
--- a/study/src/org/labkey/study/controllers/DatasetController.java
+++ b/study/src/org/labkey/study/controllers/DatasetController.java
@@ -137,32 +137,6 @@ public class DatasetController extends BaseStudyController
             if (newRecord != null)
             {
                 newData = DatasetAuditProvider.decodeFromDataMap(newRecord);
-                String lsid = newData.get("lsid");
-                if (lsid != null)
-                {
-                    // If we have a current record, display it
-                    Dataset ds = StudyManager.getInstance().getDatasetDefinition(getStudyRedirectIfNull(), datasetId);
-                    if (null != ds)
-                    {
-                        TableInfo datasetTable = ds.getTableInfo(getUser());
-
-                        TableViewForm objForm = new TableViewForm(datasetTable);
-                        objForm.set("lsid", lsid);
-                        objForm.set(DatasetDefinition.DATASETKEY, datasetId);
-
-                        DetailsView objView = new DetailsView(objForm);
-
-                        view.addView(objView);
-                    }
-                }
-                else
-                {
-                    view.addView(new NoRecordView());
-                }
-            }
-            else
-            {
-                view.addView(new NoRecordView());
             }
 
             if (oldRecord != null)
@@ -188,6 +162,10 @@ public class DatasetController extends BaseStudyController
                 }
                 view.addView(new AuditChangesView(comment, oldData, newData));
             }
+            else
+            {
+                view.addView(new NoRecordView());
+            }
 
             return view;
         }
@@ -197,7 +175,7 @@ public class DatasetController extends BaseStudyController
             @Override
             protected void renderInternal(Object model, PrintWriter out)
             {
-                out.write("<p>No current record found</p>");
+                out.write("<p>No additional details recorded</p>");
             }
         }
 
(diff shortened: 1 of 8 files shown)


study/src/org/labkey/study/controllers/DatasetController.java
```diff
@@ -197,7 +175,7 @@ public ModelAndView getView(DatasetAuditHistoryForm form, BindException errors)
             @Override
             protected void renderInternal(Object model, PrintWriter out)
             {
-                out.write("<p>No current record found</p>");
+                out.write("<p>No additional details recorded</p>");
```
labkey-susanh, line 178: Any tests that need updating as a result of this text change?

study/src/org/labkey/study/controllers/DatasetController.java
```diff
@@ -137,32 +137,6 @@ public ModelAndView getView(DatasetAuditHistoryForm form, BindException errors)
             if (newRecord != null)
             {
                 newData = DatasetAuditProvider.decodeFromDataMap(newRecord);
-                String lsid = newData.get("lsid");
```
labkey-susanh, line 140: Does the removal of this check for the lsid mean we'll never see the details?  Or is the case when we have the details to show handled elsewhere?

study/src/org/labkey/study/controllers/DatasetController.java
Sigmonia: Not that i could find.

study/src/org/labkey/study/controllers/DatasetController.java
Sigmonia: This removes the duplicated block seen on the new inserts. The block starting at Line 172 (old): 146 (change) should continue to show the new row or diff as appropriate.

labkey-susanh: Got it.

labkey-susanh (approved):

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1576,15 +1581,43 @@ public TableInfo getSchemaTableInfo()
             return m;
         }
 
+        public PHI getMaxContainedPhi()
+        {
+            return _maxContainedPhi;
+        }
+
         @Override
-        public CaseInsensitiveHashSet skipProperties()
```
Sigmonia, line 1580: Removed as it matches the underlying method

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1576,15 +1581,43 @@ public TableInfo getSchemaTableInfo()
             return m;
         }
 
+        public PHI getMaxContainedPhi()
```
Sigmonia, line 1584: Not changed, just a bad diff.

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1576,15 +1581,43 @@ public TableInfo getSchemaTableInfo()
...
+        public void addAuditEvent(User user, Container container, AuditBehaviorType auditBehavior, @Nullable String userComment, QueryService.AuditAction auditAction, Map<String, Object> parameters)
         {
-            return _maxContainedPhi;
+            Map<String,Object> oldRow = null;
+            //Leaving this block as reference for getting update diff
+//            if (auditAction == QueryService.AuditAction.MERGE)
+//            {
+//                String lsid = String.valueOf(parameters.get("lsid"));
+//                oldRow = DatasetDefinition.this.getDatasetRow(user, lsid);
+//            }
```
labkey-susanh: This code is in the other branch (https://github.com/LabKey/platform/tree/20.11_fb_dataset_merge_auditing),  so I don't think this commented out block is necessary.

labkey-susanh (approved):
As we discussed, let's add the check on auditBehavior in the two addAuditLog methods just in case the methods are called not from the DataIterator.  

Otherwise, looks good.

PR #1840, opened 2020-12-31
Adding detailed audit logging for the Merge option for importing datasets

Sigmonia:
#### Rationale
Add detailed audit logging for merging bulk imports for datasets

#### Changes
* Added a DetailedAuditLogDataIterator to the DI queue
* Added DatasetSchemaTable.addAuditEvent override methods

### DIFF
--- a/study/src/org/labkey/study/model/DatasetDefinition.java
+++ b/study/src/org/labkey/study/model/DatasetDefinition.java
@@ -41,6 +41,7 @@ import org.labkey.api.dataiterator.DataIterator;
 import org.labkey.api.dataiterator.DataIteratorBuilder;
 import org.labkey.api.dataiterator.DataIteratorContext;
 import org.labkey.api.dataiterator.DataIteratorUtil;
+import org.labkey.api.dataiterator.DetailedAuditLogDataIterator;
 import org.labkey.api.dataiterator.Pump;
 import org.labkey.api.dataiterator.StandardDataIteratorBuilder;
 import org.labkey.api.exp.ChangePropertyDescriptorException;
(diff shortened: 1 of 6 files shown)


study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1586,6 +1590,34 @@ public PHI getMaxContainedPhi()
...
+
+                StudyServiceImpl.addDatasetAuditEvent(user, DatasetDefinition.this, row, updatedRow);
+            }
+        }
+
+        @Override
+        public void addAuditEvent(User user, Container container, AuditBehaviorType auditBehavior, @Nullable String userComment, QueryService.AuditAction auditAction, Map<String, Object> parameters)
+        {
+            Map<String,Object> oldRow = null;
+            if (auditAction == QueryService.AuditAction.MERGE)
```
labkey-susanh, line 1613: Don't we still need to check if the auditBehavior is DETAILED here?

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1586,6 +1590,34 @@ public PHI getMaxContainedPhi()
...
+            {
+                //REVIEW: I'm guessing this is a terrible idea for multiple reasons, but seems to work...
```
labkey-susanh, line 1615: Same question as before.  Is there a reason other than performance that this seems like a terrible idea?  I don't know how else we would get the existing record.

Sigmonia:
Changes were pulled into https://github.com/LabKey/platform/pull/1839

PR #1841, opened 2021-01-01
New Assay Type Select Page

labkey-martyp:
#### Rationale
Create a new React page to replace the current "Choose Assay Type" page.  Uses @labkey/components AssayPicker which will be used in LKB as well.  New NPM entry point for this page. React page handles XAR upload here in Platform as this will not be used in LKB. Update actions to load the UI and handle XAR submission. 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/425
- https://github.com/LabKey/testAutomation/pull/575

#### Changes
- Add action in AssayController to get list of available assay providers, list of possible containers and the default container. Reuses code from prior implementation just no longer a FormViewAction.
- Add action in ExperimentController to upload the XAR. This reuses XAR upload from previous FormViewAction implementation.
- Add NPM entry point
- Add AssayTypeSelect component to wrap AssayPicker and navigate or upload XAR on submit
- Remove unused actions and jsps


assay/src/client/entryPoints.js
```diff
@@ -14,5 +14,10 @@ module.exports = {
         title: 'Assay Data Import',
         permission: 'insert',
         path: './src/client/AssayDataImport'
+    },{
+        name: 'assayTypeSelect',
+        title: 'New Assay Design',
+        permission: 'insert',
```
cnathe, line 20: The action where this new entryPoint is used, i.e. ChooseAssayTypeAction, requires DesignAssayPermission. Since that is the main usage of this entryPoint, I'd suggest that if we leave this generated view in place we raise this to "admin" permissions here. As it stands now, a user in the Author group can get to the assay-assayTypeSelect.view URL action/view successfully but it isn't really functional.

assay/src/client/AssayTypeSelect/AssayTypeSelect.tsx
```diff
@@ -0,0 +1,110 @@
...
+        setProvider(provider);
+    }, [])
+
+    const onContainerSelect = useCallback((cont) => {
+        setContainer(cont);
+    }, [])
+
+    const onSubmit = useCallback(() => {
+        const cont = container ?? getServerContext().container.path;
+        if (isFileUpload && xar) {
```
cnathe: if `isFileUpload` is true but we don't have a `xar`, this would go to the else statement and redirect. Is that desired? Not sure the user would actually be able to get to this `onSubmit` function in that case.

assay/src/client/AssayTypeSelect/AssayTypeSelect.scss
```diff
@@ -0,0 +1,18 @@
...
+.assay-type-select-hdr {
+  font-size: 16px;
+  font-weight: bold;
+  height: 45px;
+  padding: 10px 15px;
+}
+
+.assay-type-select-panel {
+  min-width: 1000px;
+  margin-top: 25px;
```
cnathe: was this extra top spacing in the mockups? it looks a little excessive to me

assay/src/client/AssayTypeSelect/AssayTypeSelect.tsx
```diff
@@ -0,0 +1,110 @@
...
+                    </Panel.Heading>
+                </div>
+                <Panel.Body>
+                    <AssayPicker
+                        showImport={true}
+                        onProviderSelect={onSelect}
+                        onContainerSelect={onContainerSelect}
+                        onFileChange={setXar}
+                        setIsFileUpload={setIsFileUpload}
+                        selectedTab={tab}
```
cnathe: Since this component will be used in LKS and LKB, I guess I was expecting that the `AssayPicker` itself would be the one maintaining the state of the selected inputs within the component. Was the reason for needing to put this state in this wrapped component because of the XAR import? I was imagining that all of that state and import action would be within ui-components and then this wrapper would just be how the LKS one passes in the props about allowing XAR import and handling navigation after success or on cancel.

As a comparison, the AssayDesigner.tsx wrapper in this module uses the AssayDesignerPanels in this way (i.e. all of the state and logic about the assay designer is in the shared component and then this wrapper is just for what to do onCancel and onComplete).

assay/src/client/AssayTypeSelect/AssayTypeSelect.tsx
```diff
@@ -0,0 +1,110 @@
+import React, { FC, memo, useCallback, useEffect, useState, useMemo } from 'react'
+import { AssayPicker } from '@labkey/components';
+import { Button, Panel } from "react-bootstrap";
+
+import "./AssayTypeSelect.scss"
+import { ActionURL, Ajax, Utils, getServerContext } from "@labkey/api";
+
+export function uploadXarFile(
```
cnathe: Does this function need to be exported? It looks to only be used here in this component.

assay/src/client/entryPoints.js
cnathe: Also, another option (though likely overkill here) would be to do what the AssayDesigner.tsx does. It checks the user permissions via an API call and then shows an insufficient permissions Alert (line 207) where applicable.

assay/src/org/labkey/assay/AssayController.java
```diff
@@ -622,42 +624,112 @@ public ActionURL getReturnURL()
...
-            _createIn = form.getCreateContainer();
 
-            return true;
+            beans.sort(Comparator.comparing(AssayProviderBean::getName));
+            return beans;
         }
 
         @Override
-        public ActionURL getSuccessURL(CreateAssayForm form)
+        public Object execute(Object o, BindException errors) throws Exception
```
cnathe: should the return type here be something  like `ApiResponse` or `ApiSimpleResponse`?

assay/src/org/labkey/assay/AssayController.java
```diff
@@ -622,42 +624,112 @@ public ActionURL getReturnURL()
...
-            return PageFlowUtil.urlProvider(AssayUrls.class).getDesignerURL(_createIn, form.getProviderName(), returnURL);
+            ApiSimpleResponse response = new ApiSimpleResponse();
+
+            List<AssayProviderBean> providers = getProviders();
+            JSONArray jsonProvider = new JSONArray();
+            for (AssayProviderBean provider : providers)
+            {
+                jsonProvider.put(new JSONObject(provider));
+            }
+            response.put("providers", jsonProvider);
```
cnathe: Do we need to iterate through the providers here to convert them to `JSONObject`? I would have expected that just doing the following would be sufficient: `response.put(getProviders)`.

cnathe (changes requested):

assay/src/org/labkey/assay/AssayController.java
labkey-martyp: Was having some issues with Jackson marshalling, but got it working.  Removed this manual creation of JSONObjects

assay/src/org/labkey/assay/AssayController.java
labkey-martyp: Fixed

assay/src/client/AssayTypeSelect/AssayTypeSelect.tsx
labkey-martyp: No, fixed

assay/src/client/AssayTypeSelect/AssayTypeSelect.tsx
labkey-martyp: As we discussed, the AssayPicker is maintaining state (so it can be reused in LKB) and passing up selection data to this parent component.  Refactored per your suggestion to use a single onChange handler and passing up a model of selections.

assay/src/client/AssayTypeSelect/AssayTypeSelect.scss
labkey-martyp: The goal was to make the top margin on this page match the top margin of the Assay Designer for continuity when this page navigates to that page

assay/src/client/AssayTypeSelect/AssayTypeSelect.tsx
labkey-martyp: This was refactored

assay/src/client/entryPoints.js
labkey-martyp: As we discussed, updated @labkey/build to allow permissionClasses in entry points.  Added DesignAssayPermission class as a permissionClass.

assay/src/client/AssayTypeSelect/AssayTypeSelect.tsx
```diff
@@ -0,0 +1,107 @@
...
+    }, [])
+
+    const tab = useMemo(() => ActionURL.getParameter('tab'), [])
+
+    const onCancel = useCallback(() => {
+        window.location.href = returnUrl || ActionURL.buildURL('project', 'begin', getServerContext().container.path);
+    }, [returnUrl])
+
+    const onChange = useCallback((model: AssayPickerSelectionModel) => {
+        setAssayPickerSelection(model);
```
cnathe, line 55: do we want to add the dirty state check/handling in this component wrapper?

cnathe (approved):

PR #1842, opened 2021-01-02
Fix url handling and simplify ButtonTag

labkey-adam:
#### Rationale
With the "No Question Marks in URLs" experimental feature turned on, the "Specimen Requests" page buttons were all broken due to their use of String concatenation. We're fixing that specific page and making some changes to move away from String-based URL manipulation.

image

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/281

#### Changes
* Simplify and modernize the generation of specimen requests buttons. Use `ActionURLs` and `usePost()`.
* Eliminate `ButtonTag` methods `href(String)` and `action()`.


labkey-kevink (approved):
Much nicer

PR #1844, opened 2021-01-04
Issue41815: Switch "Single sign on" to "Single sign-on" in UI 

RosalineP:
#### Rationale
Docs and product are switching to use "single sign-on". 
I'm using "Single Sign-on" since title-case is used in the relevant text, and prepositions in hyphenated words typically aren't capitalized in title-case. 

#### Related Pull Requests
* n/a

#### Changes
* Update phrasing
* Update jest snapshots

### DIFF
--- a/core/src/client/components/AuthConfigMasterPanel.tsx
+++ b/core/src/client/components/AuthConfigMasterPanel.tsx
@@ -74,7 +74,7 @@ class PrimaryTab extends PureComponent<PrimaryTabProps> {
 
                 <div>
                     <div className="configurations__config-section-title">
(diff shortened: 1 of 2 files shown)


labkey-jeckels (changes requested):
The convention appears to be that if "Sign" is capitalized, "On" should be too.

https://www.google.com/search?q=single+sign+on

RosalineP:
> The convention appears to be that if "Sign" is capitalized, "On" should be too.
> 
> https://www.google.com/search?q=single+sign+on

Oops, not regular title casing then! Updated.

labkey-jeckels (approved):
Yes. English is weird

PR #1847, opened 2021-01-04
Improve vocabulary domain support

labkey-nicka:
#### Rationale
In support of product features this PR contains a variety of changes to make Vocabulary domains more easily usable by non-platform modules.

#### Related Pull Requests
* https://github.com/LabKey/labbook/pull/80
* https://github.com/LabKey/biologics/pull/762
* https://github.com/LabKey/snd/pull/79

#### Changes
* Introduces `PropertyService.get().findVocabularyProperties()` to house common implementation for this commonly needed logic.
* Introduce `insertNullValues` option for both `insertProperties()` and `updateObjectProperties()`. This will allow for a null value to be persisted to the `exp.objectproperty` explicitly rather than not adding a row at all. This allows for recognition of properties being tied to an object that don't necessarily have a value.
* Update `OntologyManager.insertProperties` to take a `User`. The former implementation is now deprecated but still resolves the user from `currentView`. Additionally, consolidates implementation by removing the private method variant.
* Expose `_VirtualTable` from `StorageProvisioner` to allow for other modules to use the same class to override schema metadata.
* Removes `_WorkbookDataIteratorBuilder` as this class is no longer utilized.


experiment/src/org/labkey/experiment/api/VocabularyDomainKind.java
```diff
@@ -128,6 +128,15 @@ public Priority getPriority(String domainURI)
         return getKindName().equals(lsid.getNamespacePrefix()) ? Priority.MEDIUM : null;
     }
 
+    @Override
+    public String generateDomainURI(String schemaName, String queryName, Container container, User user)
+    {
+        // This makes "schemaName" synonymous with "vocabularyName" which it isn't, however, this is the
+        // interface exposed variant of generateDomainURI() allowing for resolution of a VocabularyDomainKind URI
+        // without needing access to the class.
```
labkey-kevink, line 136: Using queryName might make more sense?

I'm having trouble finding where `generateDomainURI` is used to resolve to the domain kind.  Where do we do that?

labkey-kevink (approved):
Adding null support for ontology is excellent!

api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -2059,21 +2004,85 @@ public static void deletePropertyDescriptor(PropertyDescriptor pd)
...
     }
 
+    /***
+     * @deprecated Use {@link #insertProperties(Container, User, String, ObjectProperty...)} so that a user can be
+     * supplied.
+     */
     public static void insertProperties(Container container, String ownerObjectLsid, ObjectProperty... properties) throws ValidationException
     {
-        insertProperties(container, ownerObjectLsid, false, properties);
+        User user = HttpView.hasCurrentView() ? HttpView.currentContext().getUser() : null;
```
labkey-klum, line 2013: ooh, I hate using thread local to yank out the context... I'm guessing there was no way to thread the User object all the way through.

labkey-klum (approved):

experiment/src/org/labkey/experiment/api/VocabularyDomainKind.java
labkey-nicka: I've updated this to use `queryName` instead.

api/src/org/labkey/api/exp/OntologyManager.java
labkey-nicka: There are a lot of usages of this method that are not straight-forward to convert away from implicitly getting the associated user.

PR #1850, opened 2021-01-05
Issue 38722: Migrate ExcelWriter, TSVGridWriter, DataRegion, et al to use ResultsFactory or similar

labkey-adam:
#### Rationale
`ExcelWriter` and `TSVGridWriter` are `AutoCloseable`, but they take `Results` that are produced by their callers. In some cases, both the writer and `Results` are passed around through separate code paths. This pattern makes it difficult to close these resources consistently (including error cases).

This change moves `ExcelWriter` and `TSVGridWriter` to a factory pattern. Instead of passing in `Results`, callers pass in a `ResultsFactory`. The writers generate the `Results` before rendering and close it immediately after, via try-with-resources. A `StashingResultsFactory` is provided for code paths that must inspect the `Results` before initializing the writer.

[Issue 38722: Migrate ExcelWriter, TSVGridWriter, DataRegion, et al to use ResultsFactory or similar](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=38722)

### DIFF
--- a/api/src/org/labkey/api/data/ResultsFactory.java
+++ b/api/src/org/labkey/api/data/ResultsFactory.java
@@ -1,0 +1,6 @@
+package org.labkey.api.data;
+
+public interface ResultsFactory
+{
+    Results get() throws Exception;
+}
--- a/api/src/org/labkey/api/data/ResultsFactory.java
+++ b/api/src/org/labkey/api/data/ResultsFactory.java
@@ -1,6 +1,9 @@
 package org.labkey.api.data;
 
+import java.io.IOException;
+import java.sql.SQLException;
+
 public interface ResultsFactory
 {
-    Results get() throws Exception;
+    Results get() throws IOException, SQLException;
 }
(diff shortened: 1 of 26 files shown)


api/src/org/labkey/api/data/ResultsFactory.java
```diff
@@ -0,0 +1,6 @@
+package org.labkey.api.data;
+
+public interface ResultsFactory
+{
+    Results get() throws Exception;
```
labkey-jeckels: Can we make this throw SQLException, IOException, or be responsible for translating the RuntimeExceptions? Throwing Exception tends to leave to crummy handling code, though it doesn't seem like it's had much impact in the PRs for these changes.

labkey-jeckels (approved):
One question about exception handling

api/src/org/labkey/api/data/ResultsFactory.java
labkey-adam: It was SQLException, IOException, until I converted `ScriptEngineReport` (`generateResults()` throws Exception because `createQueryView()` throws it). But it's probably best for ScriptEngineReport's factory to catch Exception and throw RuntimeException, since it's clearly the outlier. I'll adjust...

PR #1852, opened 2021-01-05
better handling of race condition in ensureDomainDescriptor()

labkey-matthewb:
#### Rationale
Use INSERT WHERE to avoid race condition with current SELECT -> INSERT code.
This will hopefully address sporadic failure in AbstractQueryService$TestCase

### DIFF
--- a/api/src/org/labkey/api/exp/OntologyManager.java
+++ b/api/src/org/labkey/api/exp/OntologyManager.java
@@ -1764,10 +1764,31 @@ public class OntologyManager
         {
             try
             {
-                dd = Table.insert(null, getTinfoDomainDescriptor(), ddIn);
-                // We may have a cached miss that we need to clear
-                uncache(dd);
-                return getDomainDescriptor(ddIn.getDomainURI(), ddIn.getContainer());
+                // ensureDomainDescriptor() shouldn't fail if there is a race condition, however Table.insert() will throw if row exists, so can't use that
+                // also a constraint violation will kill any current transaction
+                // CONSIDER to generalize add an option to check for existing row to Table.insert(ColumnInfo[] keyCols, Object[] keyValues)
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -1764,10 +1764,31 @@ public static DomainDescriptor ensureDomainDescriptor(DomainDescriptor ddIn)
...
-                return getDomainDescriptor(ddIn.getDomainURI(), ddIn.getContainer());
+                // ensureDomainDescriptor() shouldn't fail if there is a race condition, however Table.insert() will throw if row exists, so can't use that
+                // also a constraint violation will kill any current transaction
+                // CONSIDER to generalize add an option to check for existing row to Table.insert(ColumnInfo[] keyCols, Object[] keyValues)
+                String timestamp = getExpSchema().getSqlDialect().getSqlTypeName(JdbcType.TIMESTAMP);
+                String templateJson = null==ddIn.getTemplateInfo() ? null : ddIn.getTemplateInfo().toJSON();
+                SQLFragment insert = new SQLFragment(
+                        "INSERT INTO " + getTinfoDomainDescriptor().getSelectName() +
+                        " (Name, DomainURI, Description, Container, Project, StorageTableName, StorageSchemaName, ModifiedBy, Modified, TemplateInfo)\n" +
+                        "SELECT ?,?,?,?,?,?,?,CAST(NULL AS INT),CAST(NULL AS " + timestamp + "),?\n",
```
labkey-jeckels, line 1775: In most codepaths we set the modified and modifiedBy to be the same as the created and createdBy fields. Seems best to be consistent.

labkey-jeckels (approved):
One comment, but don't feel strongly about a change.

api/src/org/labkey/api/exp/OntologyManager.java
labkey-matthewb: Oddly, this table doesn't have a created and createdby!

PR #1862, opened 2021-01-06
Specimen module

labkey-adam:
#### Rationale
We want to separate the specimen repository functionality out of study into its own module to simplify the product (esp. in cases where specimens are not used), provide flexibility, and reduce maintenance costs. This WIP PR creates a study-api module and moves 100+ specimen-related classes from study-main into study-api, which ensures that these classes have no direct dependencies on study-main.

Future PRs will move the remaining specimen classes into study-api and then, as study-main dependencies on these classes are severed, they will be moved into a new specimen module.

#### Related Pull Requests
* https://github.com/LabKey/wnprc-modules/pull/57


api/src/org/labkey/api/study/StudyService.java
```diff
@@ -150,6 +157,9 @@ static StudyService get()
 
     DbSchema getDatasetSchema();
 
+    @Deprecated // Use SpecimenSchema instead
+    DbSchema getStudySchema();
```
labkey-klum, line 161: I'm a little confused by this new method, I'm assuming it will be temporary scaffolding to allow us to pull out the specimen code.

labkey-adam: Yes, hopefully temporary (and most uses are gone). For now, key specimen tables remain in study (e.g., all the sample request stuff and locations) and I'm not sure if it's worth moving them. And specimen references some of the core study tables, like participants, visits, and cohorts. We may be able to abstract all of these usages through service interfaces, but again, have to trade-off that effort.

labkey-klum (approved):

PR #1868, opened 2021-01-07
Issue 41864: DataRegion column analytics updates to lazy load Ext4 and vis dependencies

cnathe:
#### Rationale
Issue [41864](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41864): Remove dataregion dependency on ExtJS (SummaryStatisticsAnalyticsProvider)

#### Changes
* moves LABKEY.ColumnSummaryStatistics.showDialogFromDataRegion from ColumnSummaryStatistics.js to DataRegion.js
* remove ColumnSummaryStatistics dependency and vis dependency from AnalyticsProvider classes
* fix DataRegion update to column header menu item disabled state on add/remove column visualization
* wrap ColumnVisualizationAnalytics.js functions in calls to LABKEY.requiresVisualization()

### DIFF
--- a/api/src/org/labkey/api/view/PopupMenuView.java
+++ b/api/src/org/labkey/api/view/PopupMenuView.java
@@ -157,6 +157,8 @@ public class PopupMenuView extends HttpView<PopupMenu>
     protected static void renderTreeItem(NavTree item, String cls, Writer out) throws IOException
     {
         out.write("<li");
+        if (item.getId() != null)
+            out.write(" id=\"" + item.getId() + "\"");
         if (item.isDisabled())
             cls = cls != null ? cls + " disabled" : "disabled";
         if (cls != null)
--- a/api/src/org/labkey/api/view/PopupMenuView.java
+++ b/api/src/org/labkey/api/view/PopupMenuView.java
@@ -157,8 +157,6 @@ public class PopupMenuView extends HttpView<PopupMenu>
     protected static void renderTreeItem(NavTree item, String cls, Writer out) throws IOException
     {
         out.write("<li");
-        if (item.getId() != null)
-            out.write(" id=\"" + item.getId() + "\"");
         if (item.isDisabled())
             cls = cls != null ? cls + " disabled" : "disabled";
         if (cls != null)
--- a/api/webapp/clientapi/dom/DataRegion.js
+++ b/api/webapp/clientapi/dom/DataRegion.js
@@ -2746,9 +2746,7 @@ if (!LABKEY.DataRegions) {
             var view = _getViewFromQueryDetails(queryDetails, viewName);
             if (view && _viewContainsColumn(view, colFieldKey)) {
                 _addAnalyticsProviderToView.call(this, view, colFieldKey, providerName, false);
-
-                var elementId = this.name + ':' + colFieldKey + ':analytics-' + providerName;
-                _updateAnalyticsProviderMenuItem(elementId, true);
+                _updateAnalyticsProviderMenuItem(this.name + ':' + colFieldKey, providerName, true);
             }
         }, null, this);
     };
@@ -2765,9 +2763,7 @@ if (!LABKEY.DataRegions) {
             var view = _getViewFromQueryDetails(queryDetails, viewName);
             if (view && _viewContainsColumn(view, colFieldKey)) {
                 _removeAnalyticsProviderFromView.call(this, view, colFieldKey, providerName, false);
-
-                var elementId = this.name + ':' + colFieldKey + ':analytics-' + providerName;
-                _updateAnalyticsProviderMenuItem(elementId, false);
+                _updateAnalyticsProviderMenuItem(this.name + ':' + colFieldKey, providerName, false);
             }
         }, null, this);
     };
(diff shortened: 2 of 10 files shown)


api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -2831,6 +2827,16 @@ if (!LABKEY.DataRegions) {
         }
     };
 
+    var _updateAnalyticsProviderMenuItem = function(elementId, disable) {
+        var el = $('#' + elementId.replace(/:/g, '\\:'));
```
labkey-nicka: nit: This line could use a comment. Not exactly clear why this replacement is being done.

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -4742,4 +4748,44 @@ LABKEY.AggregateTypes = {
...
+            var region = LABKEY.DataRegions[dataRegionName];
+            if (region) {
+                var regionViewName = region.viewName || "",
+                        column = region.getColumn(colFieldKey);
+
+                if (column) {
+                    region.getColumnAnalyticsProviders(regionViewName, colFieldKey, function(colSummaryStats) {
+                        Ext4.create('LABKEY.ext4.ColumnSummaryStatisticsDialog', {
+                            queryConfig: region.getQueryConfig(),
+                            filterArray: LABKEY.Filter.getFiltersFromUrl(region.selectAllURL, 'query'), //Issue 26594
```
labkey-nicka: Shouldn't this be using the data region's `dataRegionName`?

```js
filterArray: LABKEY.Filter.getFiltersFromUrl(region.selectAllURL, region.dataRegionName),
```

I realize this code is just copied across so take it or leave it.

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -4742,4 +4748,44 @@ LABKEY.AggregateTypes = {
     AVG: 'mean'
 
     // TODO how to allow premium module additions to aggregate types?
+};
+
+LABKEY.ColumnSummaryStatistics = new function () {
```
labkey-nicka: Since this is now defined in `DataRegion.js` it might be better to just incorporate this into `DataRegon` much like customize view. Then the callers can call `LABKEY.DataRegion[regionName].showSummaryStatisticsDialog(colFieldKey)` or some such.

This would remove the need for this extra `LABKEY.ColumnSummaryStatistics` namespace and feel more "native" to a data region.

api/src/org/labkey/api/view/PopupMenuView.java
```diff
@@ -157,6 +157,8 @@ else if ("-".equals(child.getText()))
     protected static void renderTreeItem(NavTree item, String cls, Writer out) throws IOException
     {
         out.write("<li");
+        if (item.getId() != null)
```
labkey-nicka: Hmm, I don't think this is a good idea. This id is not guaranteed to be unique or compliant with the DOM id accepted syntax. Additionally, it ends up being generated for a lot more elements than just the one we're looking to target with this change.

That said, I presume you got this code to invoke `_updateAnalyticsProviderMenuItem` (I couldn't quite find the scenario). Could we instead load Ext4 at the time this is called as well and then still do the component query?

cnathe: Yeah, I see what you are saying about the DOM id. We could revert back to the Ext4 component query, but from what I could tell that was actually broken as well since it was trying to do an elementId query and from what I could tell that wasn't applicable any more (i.e. these aren't Ext4 menu anymore). I think we could instead do a jQuery selector without having to add this id. I'll take another look.

api/webapp/clientapi/dom/DataRegion.js
cnathe: Good call, I'll make that change.

api/webapp/clientapi/dom/DataRegion.js
cnathe: Yeah, I just copied this code over and wasn't planning to change it at all, but I can take a look at this suggestion.

api/webapp/clientapi/dom/DataRegion.js
cnathe: done

api/webapp/clientapi/dom/DataRegion.js
cnathe: Turns out that the `region.selectAllURL` which is being used in this call to `LABKEY.Filter.getFiltersFromUrl()` does actually use the constant `query` as the url param prefix instead of the `dataRegionName`.

api/webapp/clientapi/dom/DataRegion.js
cnathe: Done

api/src/org/labkey/api/view/PopupMenuView.java
cnathe: Done

api/webapp/clientapi/dom/DataRegion.js
labkey-nicka: Eesh, that sounds like a bug...

labkey-nicka (approved):
Awesome. I think this is a good change.

PR #1870, opened 2021-01-09
Render and log JSP compilation errors

labkey-adam:
#### Rationale
At some point, we stopped reporting runtime JSP compilation errors. This PR adds them to the exception message so they're logged and rendered in exception pages. It also logs a link to the JSP file for one-click access to the file from the IntelliJ log. [Issue 42206: Display and log JSP compilation errors, not server stacktrace](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42206)


api/src/org/labkey/api/jsp/RecompilingJspClassLoader.java
```diff
@@ -205,31 +205,54 @@ private String getCompleteSourcePath(String finderSource, String jspSource)
...
+            }
+        }
+
+        @Override
+        public StackTraceElement[] getStackTrace()
         {
-            super("Error compiling " + jspFilename);
-            _errors = errors;
+            // Don't care about the stack trace
+            return new StackTraceElement[]{};
```
labkey-jeckels, line 249: Sneaky!

labkey-jeckels (approved):
Thanks. I had noticed this but not gotten to logging an issue or investigating a fix.

PR #1878, opened 2021-01-11
Issue 42210: Sample update details are not showing when using an import alias

XingY:
#### Rationale
When importing using alias, the alias field, as well as the "dataInputs/XX" field is present in updatedRow. The check for input fields should be case insensitive.

#### Related Pull Requests
* 

#### Changes
* case insensitive check for input field names


labkey-susanh (approved):
Ah, case-sensitivity.  Still a thing in 2021, I guess.

PR #1879, opened 2021-01-12
test fixes

labkey-matthewb:
#### Rationale
fixes for reimport PR, 
StudyReloadColumnInferenceTest

### DIFF
--- a/api/src/org/labkey/api/dataiterator/HashDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/HashDataIterator.java
@@ -108,7 +108,8 @@ public class HashDataIterator extends SimpleTranslator
             for (int i = start; i < length; i++)
             {
                 Object v = arr[i];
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/dataiterator/HashDataIterator.java
```diff
@@ -108,7 +108,8 @@ public static String calculateRowHash(Object[] arr, int start, int length)
             for (int i = start; i < length; i++)
             {
                 Object v = arr[i];
-                assert null == v || v instanceof String || v instanceof Number;
+                // Unfortunately, dataset import still requires inferTypes, so we have to handle parsed types here.
```
labkey-kevink, line 111: Should we include a setting on HashDataIterator to only allow parsed/unparsed data from the data loader?

labkey-kevink (approved):

PR #1881, opened 2021-01-13
Test fixes

labkey-adam:
#### Rationale
Crawler does not enjoy the NPEs it gets when it sends bad parameters to `UpdateDatasetVisitMappingAction`.

`SecurityManager$TestCase.testCreateUser()` continues to fail authenticating despite a strong password. I suspect bad luck... random password occasionally includes three random characters from the email address. Log the test password so we can verify this theory.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1872

#### Changes
* Improve UpdateDatasetVisitMappingAction validation to appease crawler
* Log test password when authentication fails

### DIFF
--- a/api/src/org/labkey/api/security/SecurityManager.java
+++ b/api/src/org/labkey/api/security/SecurityManager.java
@@ -2523,7 +2523,7 @@ public class SecurityManager
                 SecurityManager.setPassword(email, password);
 
                 User user2 = AuthenticationManager.authenticate(ViewServlet.mockRequest("GET", new ActionURL(), null, null, null), rawEmail, password);
-                assertNotNull(rawEmail + " failed to authenticate; check labkey.log around timestamp " + DateUtil.formatDateTime(new Date(), "HH:mm:ss,SSS") + " for the reason", user2);
+                assertNotNull("\"" + rawEmail + "\" failed to authenticate with password \"" + password + "\"; check labkey.log around timestamp " + DateUtil.formatDateTime(new Date(), "HH:mm:ss,SSS") + " for the reason", user2);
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -2523,7 +2523,7 @@ public void testCreateUser() throws Exception
                 SecurityManager.setPassword(email, password);
 
                 User user2 = AuthenticationManager.authenticate(ViewServlet.mockRequest("GET", new ActionURL(), null, null, null), rawEmail, password);
-                assertNotNull(rawEmail + " failed to authenticate; check labkey.log around timestamp " + DateUtil.formatDateTime(new Date(), "HH:mm:ss,SSS") + " for the reason", user2);
+                assertNotNull("\"" + rawEmail + "\" failed to authenticate with password \"" + password + "\"; check labkey.log around timestamp " + DateUtil.formatDateTime(new Date(), "HH:mm:ss,SSS") + " for the reason", user2);
```
labkey-jeckels, line 2526: Does this mean that it successfully set the password to something that we're then rejecting as insufficiently secure?

labkey-jeckels (approved):

api/src/org/labkey/api/security/SecurityManager.java
labkey-adam: Yes, the setter that the test uses bypasses the password validation that happens in the controller. SecurityManager and therefore this test lives in API but the controller and password rules are in Core. A refactor is probably in order here, but I'd like to verify my theory before jumping in.

PR #1883, opened 2021-01-13
Issue 41058: Return public name from GetQueriesAction

labkey-martyp:
#### Rationale
GetQueryDetailsAction has been updated to return public name.  This PR updates GetQueriesAction to also return public name and updates a handful of tables that need public name updated.  Extensive tests run across our test suites.

#### Changes
- Update QueryController.getQueryProps(), used in GetQueriesAction, to return public name for query
- Add getPublicName() overrides to tables missing public names


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -5875,6 +5875,7 @@ public ApiResponse execute(GetQueriesForm form, BindException errors)
                         if (columns.size() > 0)
                             qinfo.put("columns", columns);
                     }
+                    name = table.getPublicName();
                     if (table instanceof DatasetTable)
```
labkey-jeckels: I think we should eliminate this special case and always send the title back. I remember making a similar change to stop special-casing DatasetTable in another API a while ago but I can't find the commit.

labkey-jeckels (approved):

query/src/org/labkey/query/controllers/QueryController.java
labkey-martyp: Done

PR #1887, opened 2021-01-14
Issue 39785: JSON responses for APIs targeting invalid container

labkey-jeckels:
#### Rationale
The previous change used checkPermissions() to signal an action's preferred response format. That doesn't get invoked when the URL is targeting a bogus container, so API-oriented actions replied with HTML. 

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1787

#### Changes
* Let actions indicate preferred response format via separate method


labkey-kevink (approved):

labkey-nicka (approved):
Pulled down locally and verified against python API integration tests and against issue repro scenario. Both succeeded and returned JSON as expected.

PR #1900, opened 2021-01-18
42141: importing file into assay after deleting the run creates duplicate exp.data

labkey-kevink:
#### Rationale
When importing a file from the filebrowser into an assay, the existing exp.data for the file isn't used.  This can happen when a file is imported into the assay a second time after the original assay run has been deleted.

During import the existing file is moved into the assay upload temp directory, a new exp.data for the new file path is created, and the file is moved the file back to the original location.  The result is two exp.data for the same file path.  This PR skips moving the file into the assay upload temp directory if the file already exists in the pipeline root so that the existing exp.data can be found during import.

#### Related Issues
* [Issue 26726](https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=26726): Transform script in assay not handling files when imported via pipeline

#### Changes
- don't copy uploaded file into assay upload temp dir if it already exists under the pipe root
- add regression test to verify the original exp.data is reused if the file is imported a second time
- remove unused `AssayDataCollector.getRoot` method

### DIFF
--- a/api/src/org/labkey/api/assay/AbstractTempDirDataCollector.java
+++ b/api/src/org/labkey/api/assay/AbstractTempDirDataCollector.java
@@ -28,7 +28,6 @@ import org.labkey.api.util.NetworkDrive;
 
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


labkey-jeckels (approved):
This seems OK to me.

@labkey-martyp you've worked on some of assay file handling edge cases with transform scripts. Any concerns?

labkey-martyp (approved):
Looks good to me.  If I recall correctly, the only reason we're using a temp directory is to keep input and script generated files together for easier cleanup and debugging.  We did some work with the provenance work to better handle file cleanup and making sure not to delete the original pipeline inputs, but looks like we didn't handle exp.data.  The only slight inconvenience for the user I could see here is if they select to Save Script Data for Debugging in the assay design, they would have to look in one location for the pipeline inputs and another for all the script data.  Pretty minor inconvenience for a rarely used feature I would guess.

PR #1903, opened 2021-01-18
Issue40467: Submit buttons for list insert and update are not tab-traversable

RosalineP:
#### Rationale
For accessibility and consistency with other buttons, button builder buttons should be focusable and triggerable using the keyboard.

#### Related Pull Requests
* n/a

#### Changes
* Make button builder buttons focusable
* Make button builder buttons trigger on enter


RosalineP:
(Not expecting two CRs, just whoever gets to review these changes first is sufficient)

Currently the buttons are only enter-triggerable, and not space-triggerable. In commit 3aaa3eb, 'Space keystroke on focus triggers submit buttons', I got space-trigger working for 'submit' builder buttons using the 'onkeypress' attribute that Josh suggested, but this was not workable for non-submit buttons (the root reason being essentially because the buttons are <a/> elements and not <input/> elements. I can go more into explaining this if desired. Workarounds that I could see were unreliable and also gross). 

Since in our email chain we decided space-triggerability would only be a nice-to-have, I decided to hold off on it. If there's a place to include this issue as a note for future accessibility improvements, let me know!

labkey-jeckels:
Looks promising to me, but others may have a better sense of the quirks of having the "#" href. I know I've been surprised by browser behavior around stuff like this.

api/src/org/labkey/api/util/Button.java
```diff
@@ -232,10 +232,11 @@ public HtmlString getHtmlString()
         boolean iconOnly = getIconCls() != null;
         String submitId = GUID.makeGUID();
         final HtmlString tip = (null != tooltip ? HtmlString.of(tooltip) : (!iconOnly ? null : html));
+        String hrefValue = (null == getHref()) ? "#" : getHref();
 
         var attrs = at(attributes)
             .id(getId())
-            .at(Attribute.href, getHref(), title, tip, onclick, generateOnClick(submitId), Attribute.rel, getRel(), Attribute.name, getName(), Attribute.style, getStyle(), Attribute.target, getTarget())
+            .at(Attribute.href, hrefValue, title, tip, onclick, generateOnClick(submitId), Attribute.rel, getRel(), Attribute.name, getName(), Attribute.style, getStyle(), Attribute.target, getTarget(), tabindex, 0)
```
cnathe: I don't think we need the `tabIndex=0` part here. In local testing using both list and wiki save buttons, the change to make sure each button `<a>` tag had an href was sufficient to make it tab traversable and handle click on enter key.

cnathe (changes requested):

api/src/org/labkey/api/util/Button.java
RosalineP: Oop, looks like you're right! Removed, thanks

labkey-nicka (approved):
To be strict, if we don't have an `href` we shouldn't be using an `<a>` tag since they are intended to be for navigation and `href` is a required attribute for valid markup. The value `"#"` for the `href` is a valid target, however, if this anchor tag is clicked, and does not have a corresponding `onclick` handler then it will push on the user's browser history while effectively not navigating them anywhere (just appending the `"#"` to the URL).

That said, it is OK if the `onclick` handles and prevents invocation of the `href` but you can see how the interplay between these attributes on an `<a>` can cause subtle undesirable behavior if not generated properly.

cnathe (approved):

PR #1913, opened 2021-01-20
Issue 42269: SM: "select them in the grid" link can cause invalid grid selection state

XingY:
#### Rationale
Query-SetCheckedAction does not check against valid ids. In the case of selecting samples/sources using transaction audit id, the sets of ids associated with a transaction might have become invalid overtime. This PR allows SetCheckedAction to optionally validate the set of ids it receives.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1913
* https://github.com/LabKey/labkey-ui-components/pull/432
* https://github.com/LabKey/sampleManagement/pull/463

#### Changes
* allow validating ids before setting checked


api/src/org/labkey/api/data/DataRegionSelection.java
```diff
@@ -297,7 +298,7 @@ public static void clearAll(ViewContext context)
         return getSelected(view, form.getQuerySettings().getSelectionKey());
     }
 
-    public static List<String> getSelected(QueryView view, String key) throws IOException
+    public static Pair<DataRegion, RenderContext> getDataRegionContext(QueryView view)
```
cnathe: Should/can `getDataRegionContext` be private?

cnathe (approved):

PR #1914, opened 2021-01-21
Issue 42207: List items get indexed twice with different document IDs

Sigmonia:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-update.view?issueId=42207
* We generate our entity ids as lowercase strings and set them to index. MS SQL returns them as uppercase strings, so the original ids aren't found and are reindexed. 

#### Changes
* Refresh list definition at indexing time.

### DIFF
--- a/list/src/org/labkey/list/model/ListManager.java
+++ b/list/src/org/labkey/list/model/ListManager.java
@@ -386,7 +386,8 @@ public class ListManager implements SearchService.DocumentProvider
 
             Runnable r = () ->
             {
-                ListDefinition list = ListDefinitionImpl.of(def);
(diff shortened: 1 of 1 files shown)


labkey-matthewb (approved):

labkey-tchad (commented):
Verifying fix on TeamCity:
https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_ModuleSuites_SearchPostgres/1235935?buildTab=overview
https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_ModuleSuites_SearchSqlserver/1235933?buildTab=overview

PR #1873, opened 2021-01-09
Specimen module - Part 2

labkey-adam:
#### Rationale
Continue work migrating specimen functionality out of the study module

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1862

#### Changes
* Move more location methods to LocationManager and shore up SpecimenEventManager. Remove many methods from SpecimenManager.
* Move ExtendedSpecimenRequestView, SpecimenRequest, and request requirement classes.
* Rename nearly all remaining "sample" references to "specimen" for consistency: actions, classes, methods, jsps, variables, comments, etc.
* Migrate AutoCompleteAction, ActorNotificationRecipientSet, SpecimenCommentAuditDomainKind
* Create specimen module and set up dependencies
* Move/register specimen security roles and register request attachment type in specimen module
* Migrate comment audit classes. Register audit provider, domain kinds, and import strategy in SpecimenModule
* Migrate webparts, query helper, study cache, etc.
* Extract and move SpecimenToolsWebPartFactory
* Extract and move SpecimensPage, making the "Specimens" tab optional. Provide mechanism for registering study tabs, such as "Specimens," from other modules.
* Move ParticipantGroup to study-api
* Enable Specimen module in tests that require it
* Create SpecimenRequestManager and migrate most SpecimenManager request methods to it
* Create SimpleStudyExportContext and BaseStudyWriter. Move AbstractSpecimenWriter to study-api.
* Factor SpecimenColumns and SpecimenTableManager (code shared by writers and importers) out of SpecimenImporter
* Disentangle and move SpecimenSettingsWriter, SpecimenWriter, and SpecimenArchiveWriter to study-api. Move specimens.xsd to API (for now).

PR #1916, opened 2021-01-21
Simplify default repository settings

labkey-adam:
#### Rationale
`AdvancedImportOptionsTest.testImportToMultipleFolders` started failing because the default specimen repository settings weren't matching its expectations. The goal of this PR is to restore the previous defaults without breaking a bunch of other tests.

In more detail, study's handling of specimen repository default settings has been complicated and somewhat unpredictable. `StudyImpl` had a `getRepositorySettings()` method that ended up getting invoked by reflection by the table layer before each study was created. This method used the ensure pattern to proactively save default repository settings. The refactor to extract specimen functionality from study changed this, first removing the getter and therefore the early call, and then removing the ensure pattern; we simply return the default values until they're explicitly saved. Problem now is that the default settings actually change based on whether a study is present or not. I can't see a good reason for this, so this PR shifts to always return the same default settings.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1873
* https://github.com/LabKey/platform/pull/1894

#### Changes
* Return the same default repository values (standard repository, requests disabled) regardless of whether a study exists or not

### DIFF
--- a/study/api-src/org/labkey/api/specimen/settings/RepositorySettings.java
+++ b/study/api-src/org/labkey/api/specimen/settings/RepositorySettings.java
@@ -59,11 +59,11 @@ public class RepositorySettings
     {
         this(container);
         String simple = map.get(KEY_SIMPLE);
-        _simple = null != simple && Boolean.parseBoolean(simple);
+        _simple = Boolean.parseBoolean(simple);
         String enableRequests = map.get(KEY_ENABLE_REQUESTS);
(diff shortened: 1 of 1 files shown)


labkey-adam:
AdvancedImportOptionsTest and the entire Specimen suite passed (one failure above is unrelated intermittent) on this branch

labkey-danield (approved):

PR #1917, opened 2021-01-21
Item 8357: LKFM Inventory data folder archive export/import

cnathe:
#### Rationale
As a precursor to allowing for LKSM/FM evaluation data to be imported via a folder archive, this story supports the inventory data part of the LKFM app. This PR adds the folder archive data type for "Inventory data" to the folder.xsd and list of FolderArchiveDataTypes.

#### Related Pull Requests
* https://github.com/LabKey/dataintegration/pull/86
* https://github.com/LabKey/platform/pull/1917
* https://github.com/LabKey/inventory/pull/174

#### Changes
* Add "inventory" export dir type to folder.xsd for folder.xml
* Add INVENTORY type to FolderArchiveDataTypes
* Minor updates to SampleTypeAndDataClassFolderImporter for consistency with other importers


experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderImporter.java
```diff
@@ -239,6 +242,7 @@ public FolderImporter create()
         @Override
         public int getPriority()
         {
+            // TODO add a note about why this value is set as such? what other importer does it need to be after?
```
cnathe: @labkey-klum I was looking at this folder importer while doing the dev for the LKFM inventory data importer. I noticed that this importer priority was set, but I wasn't sure why. Can you help me fill in this TODO about which importer we are making sure happens before the sample type and data class import?

labkey-klum: I set this so that it would run after the experiments & runs importer, both importers use the XAR format to capture the lineage for sample types and data classes and I can't remember the exact problem but I think it was related to having an assay run with references to sample type information, wouldn't get hooked up correctly if the order was inverted.

cnathe: That makes sense. I'll update that comment accordingly. Thanks.

labkey-klum (approved):

PR #1918, opened 2021-01-21
Fb issue 40274

labkey-klum:
#### Rationale
The dataset domain kind allows a file link field type but does not appear to support it in either the data iterators or table info. This PR addresses these issues:

- https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40274

#### Changes
- Persist the uploaded files into the datasetdata subfolder
- Render the uploaded file using `FileLinkDisplayColumn`

### DIFF
--- a/study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
+++ b/study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
@@ -17,7 +17,6 @@ package org.labkey.study.model;
 
 import org.apache.commons.beanutils.ConvertUtils;
 import org.apache.commons.beanutils.Converter;
-import org.labkey.api.attachments.AttachmentFile;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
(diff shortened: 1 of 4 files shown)


study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
```diff
@@ -804,5 +820,43 @@ else if (_defaultQCState != null)
                 return null;
             }
         }
+
+        class FileColumn implements Supplier<Object>
```
labkey-adam: Would `FileColumn` be useful elsewhere? If so, would be easy to generalize and move to api.

labkey-adam (approved):

study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
labkey-klum: Possibly, I could move it there for other data iterators to use.

PR #1919, opened 2021-01-22
DataRegion fix for Ext4 dependency loading related to prompt to save session view

cnathe:
#### Rationale
Found another case where an action in the DataRegion.js code need to explicitly request Ext4 lib before showing prompt to save the session view.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1868

#### Changes
* Move call to call to LABKEY.DataRegion.loadViewDesigner() to beginning of _saveSessionShowPrompt()

### DIFF
--- a/api/webapp/clientapi/dom/DataRegion.js
+++ b/api/webapp/clientapi/dom/DataRegion.js
@@ -3327,7 +3327,9 @@ if (!LABKEY.DataRegions) {
                     return;
                 }
 
-                _saveSessionShowPrompt(this, json);
+                LABKEY.requiresExt4Sandbox(function() {
+                    _saveSessionShowPrompt(this, json);
+                }, this);
             },
             scope: region
         });
--- a/api/webapp/clientapi/dom/DataRegion.js
+++ b/api/webapp/clientapi/dom/DataRegion.js
@@ -3327,62 +3327,60 @@ if (!LABKEY.DataRegions) {
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -3327,7 +3327,9 @@ if (!LABKEY.DataRegions) {
                     return;
                 }
 
-                _saveSessionShowPrompt(this, json);
+                LABKEY.requiresExt4Sandbox(function() {
```
labkey-nicka: nit: I think I'd prefer to keep this load request call in one place. Instead, I'd change how `_saveSessionShowPrompt` loads the config and then requests to load the designer:

```js
var _saveSessionShowPrompt = function(region, queryDetails) {
    LABKEY.DataRegion.loadViewDesigner(function() {
        var config = Ext4.applyIf(/* ... */);
        
        LABKEY.internal.ViewDesigner.Designer.saveCustomizeViewPrompt(config);
    });
};
```

labkey-nicka (approved):

api/webapp/clientapi/dom/DataRegion.js
cnathe: Ah yes, I agree. I didn't see that `loadViewDesigner` inside of `_saveSessionShowPrompt`. Just moving the config generation inside of that callback is a much better change.

PR #1920, opened 2021-01-22
Issue 42295: ClassCastException in MetricJSONDisplayColumn

labkey-jeckels:
#### Rationale
Our new approach of using a JSONB column for these metrics means we don't need to use a tricky DisplayColumn to pull out specific metrics. Thus, we can remove this now unused code.

I considered moving it out of Mothership into Core or API because it's potentially useful in other scenarios. But without a concrete use case, I figured it was best to shed the third-party dependency and reduce our code footprint.

#### Changes
* Remove MetricJSONDisplayColumn, its factory, and the small library it used for pull the JSON data


labkey-adam:
Looks like some jayway references in MothershipController

labkey-ankurj (approved):

labkey-adam (changes requested):
Probably should build

labkey-jeckels:
Picky, picky

labkey-adam (approved):
Fabulous

PR #1925, opened 2021-01-22
Switch usages to TableInsertDataIteratorBuilder

labkey-nicka:
#### Rationale
Migrates all remaining usages of `TableInsertDataIterator.create()` to using `new TableInsertDataIteratorBuilder()`. Additionally, makes `ScriptTriggerFactory` defensive against a `null` value for the `container` provided to the builder

#### Related Pull Requests
* https://github.com/LabKey/labbook/pull/87
* https://github.com/LabKey/harvest/pull/142
* https://github.com/LabKey/commonAssays/pull/294

#### Changes
* Switch usages to use `TableInsertDataIteratorBuilder`.
* Remove previously deprecated variants of `TableInsertDataIterator.create()`.
* Removes redundant override of `persistRows` in `DatasetDefinition`.
* Annotates `TriggerFactory` to indicate `container` argument is `@Nullable`.
* Makes `ScriptTriggerFactory` defensive against a `null` container value and immediately returns an empty collection. All other variants of `TriggerFactory` were not using the container value at all.

### DIFF
--- a/api/src/org/labkey/api/data/AbstractTableInfo.java
+++ b/api/src/org/labkey/api/data/AbstractTableInfo.java
@@ -1573,7 +1573,7 @@ abstract public class AbstractTableInfo implements TableInfo, AuditConfigurable,
     }
 
     @Override
-    public boolean hasTriggers(Container c)
+    public boolean hasTriggers(@Nullable Container c)
     {
         return !getTriggers(c).isEmpty();
     }
@@ -1594,7 +1594,7 @@ abstract public class AbstractTableInfo implements TableInfo, AuditConfigurable,
     private Collection<Trigger> _triggers = null;
 
     @NotNull
-    protected Collection<Trigger> getTriggers(Container c)
+    protected Collection<Trigger> getTriggers(@Nullable Container c)
     {
         if (_triggers == null)
         {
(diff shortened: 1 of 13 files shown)


api/src/org/labkey/api/dataiterator/TableInsertDataIterator.java
```diff
@@ -61,30 +60,14 @@
     private Connection _conn = null;
     private Set<DomainProperty> _adhocPropColumns = new LinkedHashSet<>();
 
-    @Deprecated // use TableInsertDataIteratorBuilder
```
labkey-matthewb, line 64: Yeah!

api/src/org/labkey/api/data/AbstractTableInfo.java
```diff
@@ -1573,7 +1573,7 @@ public void addTriggerFactory(TriggerFactory factory)
     }
 
     @Override
-    public boolean hasTriggers(Container c)
+    public boolean hasTriggers(@Nullable Container c)
```
labkey-matthewb, line 1576: Comment on behavior when c==null

labkey-matthewb (approved):

api/src/org/labkey/api/data/AbstractTableInfo.java
labkey-nicka: Added a comment in `ScriptTriggerFactory`.

PR #1926, opened 2021-01-25
Regression coverage for Files in Datasets disappear after editing the…

labkey-sweta:
#### Rationale
Automation Test
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=42309

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/607

#### Changes
study/test/src/org/labkey/test/tests/study/StudyDatasetFileFieldTest.java


study/test/src/org/labkey/test/tests/study/StudyDatasetFileFieldTest.java
```diff
@@ -0,0 +1,105 @@
...
+                "intField", "25"));
+
+        log("Edit the dataset");
+        DataRegionTable table = new DataRegionTable("Dataset", getDriver());
+        table.clickEditRow(0);
+        setFormElement(Locator.name("quf_textField"), "Welcome..!");
+        clickButton("Submit");
+
+        log("Verify file field is not deleted");
+        checker().verifyEquals("Incorrect file name ", " datasetdata\\xml_sample.xml", table.getDataAsText(0, "fileField"));
```
labkey-tchad: This looks like a Windows file path and I suspect that it will fail on Linux or MacOS.

study/test/src/org/labkey/test/tests/study/StudyDatasetFileFieldTest.java
```diff
@@ -0,0 +1,105 @@
...
+                "SequenceNum", "2",
+                "date", "2020-08-04",
+                "fileField", filePath.toString(),
+                "textField", "Hello World..!",
+                "intField", "25"));
+
+        log("Edit the dataset");
+        DataRegionTable table = new DataRegionTable("Dataset", getDriver());
+        table.clickEditRow(0);
+        setFormElement(Locator.name("quf_textField"), "Welcome..!");
```
labkey-tchad, line 88: The issue mentions that there should be a thumbnail or link to the existing file on the update row page. We should check for that here.

labkey-tchad (changes requested):

study/test/src/org/labkey/test/tests/study/StudyDatasetFileFieldTest.java
labkey-klum: Yes, that would be a good check, I would consider a different file from xml, since it will just render as a download link and not as an inline image with a popup.

labkey-sweta: Changed the Verification

labkey-tchad: I was suggesting that this should verify the download link and thumbnail on the update page before submitting. I think the file would need to be an image of some kind to actually have a thumbnail.

labkey-tchad (changes requested):

labkey-tchad (approved):

PR #1928, opened 2021-01-25
Issue 33004: Template.Framed appears to not be a thing

labkey-adam:
#### Rationale
Remove unused Template.Framed option, but mostly an excuse to introduce a bit of Java 14 syntax (switch expressions)


labkey-nicka (approved):
> but mostly an excuse to introduce a bit of Java 14 syntax

I was wondering why you were so quick to pick this up. Now I see.

labkey-matthewb (approved):

PR #1936, opened 2021-01-27
remove one usage of Ext4

labkey-matthewb:
#### Rationale
Remove large dependency for a very small usage. (e.g. remove Ext4 from customer portal pages with message web part)


labkey-klum (approved):
this looks great, thanks for doing this! Eventually we may want to remove our dependency on jquery for a pure javascript solution.

PR #1937, opened 2021-01-27
40384: Fix Chart Filtering

labkey-nicka:
#### Rationale
This PR addresses issues [37153](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=37153) and [40384](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40384) by clearly delineating when and how filters are applied to charts from the different sources.

#### Changes
* Refactor how the set of filters is constructed that fulfills the query configuration filters backing charts.
* Switch to applying both immutable and mutable (aka "removable") filters to the associated QueryWebPart (QWP).
* Outline (as a comment) the supported filtering sources.

### DIFF
--- a/visualization/resources/web/vis/chartWizard/genericChartPanel.js
+++ b/visualization/resources/web/vis/chartWizard/genericChartPanel.js
@@ -681,9 +681,10 @@ Ext4.define('LABKEY.ext4.GenericChartPanel', {
             removableFilters = removableFilters.concat(this.getDataRegionFilters());
         }
 
+        var allFilters = this.getUniqueFilters(filters.concat(removableFilters));
         var userSort = LABKEY.Filter.getSortFromUrl(this.getFilterURL(), this.dataRegionName);
 
-        this.currentFilterStr = this.createFilterString(removableFilters);
+        this.currentFilterStr = this.createFilterString(allFilters);
         this.currentParameterStr = Ext4.JSON.encode(this.parameters);
 
         var wpConfig = {
@@ -714,7 +715,7 @@ Ext4.define('LABKEY.ext4.GenericChartPanel', {
         // save the dataregion
         this.panelDataRegionName = wp.dataRegionName;
 
-        // issue 21418: support for parameterized queries
+        // Issue 21418: support for parameterized queries
         wp.on('render', function(){
             if (wp.parameters)
                 Ext4.apply(this.parameters, wp.parameters);
@@ -723,6 +724,10 @@ Ext4.define('LABKEY.ext4.GenericChartPanel', {
         wp.render(renderTo);
     },
 
+    /**
+     * Returns the filters applied to the associated Data Region. This region's presence is optional.
+     * If this region is not available, then this method will return any filters found explicitly on the URL.
+     */
     getDataRegionFilters : function()
     {
         var dataRegion = LABKEY.DataRegions[this.dataRegionName];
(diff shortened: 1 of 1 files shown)


visualization/resources/web/vis/chartWizard/genericChartPanel.js
```diff
@@ -663,29 +664,38 @@ Ext4.define('LABKEY.ext4.GenericChartPanel', {
...
+            // If the region does not exist, then apply it's filters from the URL
+            // and allow them to be mutated on the QWP.
+            removableFilters = removableFilters.concat(this.getDataRegionFilters());
+        }
 
-        var userSort = LABKEY.Filter.getSortFromUrl(filterUrl, this.dataRegionName);
+        var userSort = LABKEY.Filter.getSortFromUrl(this.getFilterURL(), this.dataRegionName);
 
-        this.currentFilterStr = this.createFilterString(userFilters);
+        this.currentFilterStr = this.createFilterString(removableFilters);
```
cnathe: since `this.currentFilterStr` is used to determine if the chart needs to be re-rendered when the user switches between the "View Data" and "View Chart" button options, I think this needs to use the `filters` instead of `removableFilters`:
```suggestion
        this.currentFilterStr = this.createFilterString(filters);
```

visualization/resources/web/vis/chartWizard/genericChartPanel.js
```diff
@@ -778,20 +814,13 @@ Ext4.define('LABKEY.ext4.GenericChartPanel', {
             config.scope = this;
         }
 
-        var filters;
+        // Filter scenarios (Issue 37153, Issue 40384)
+        // 1. Filters defined explicitly on chart configuration (this.userFilters)
+        // 2. Filters defined on associated QWP (panelDataRegionName)
+        // 3. Filters defined on associated Data Region (dataRegionName)
+        // 4. Filters defined on URL for associated Data Region (overlap with #3)
```
cnathe, line 838: Thanks for writing this up and including it in a comment. After this issue is resolved and PR merged, we should probably pass this over to the test team as a TODO for adding a regression test for the various combinations of these filter scenarios.

cnathe (approved):
Overall looks good. Manual testing done locally with charts that have saved filters as part of their config and charts without. Then I tested applying this run URL filter on top of both and from there adding/removing "View Data" QWP filters. All looks good.

The one comment about the "should the chart re-render" check was all I found, but approving as no need to re-review.

visualization/resources/web/vis/chartWizard/genericChartPanel.js
labkey-nicka: I've pointed Chris to these and we've gone over these scenarios. He is going to take on updating the Visualization tests that need to be updated due to this and we'll work through ensuring coverage for these scenarios.

visualization/resources/web/vis/chartWizard/genericChartPanel.js
labkey-nicka: It ended up being even more subtle than this. I've updated this to include all filters since the user could modify any of these filters (e.g. either on the Data Region directly or on the QWP directly). See 1ba0ae4.

visualization/resources/web/vis/chartWizard/genericChartPanel.js
labkey-nicka: I ended up triaging the tests and they had found edge cases with the `filterUrl` parameter. I've since patched with 3d4d6ce and the tests are looking good.

PR #1938, opened 2021-01-27
Item 8294: Job Template Import and Export

XingY:
#### Rationale
Sample Manager uses exp.protocol to define job templates. Folder export option "Experiments and Runs" was previously only available when there are exp.runs in the folder. To support export/import of sample manager job templates, "Experiments and Runs" has been changed to "Experiments, Protocols, and Runs", to allow protocol only export. 

#### Related Pull Requests
* 

#### Changes
* change label to "Experiments, Protocols, and Runs"
* show "Experiments, Protocols, and Runs" option when there are protocols defined
* add protocol to exported xar
* check option based on LKSM module presence


cnathe:
@XingY FYI, that I had to update the TrialShare related tests for the new "Inventory locations and items" folder archive option. I assume this change to the experiments folder archive option will need a small tweak to these files as well: https://github.com/LabKey/trialShare/pull/83

XingY:
> @XingY FYI, that I had to update the TrialShare related tests for the new "Inventory locations and items" folder archive option. I assume this change to the experiments folder archive option will need a small tweak to these files as well: [LabKey/trialShare#83](https://github.com/LabKey/trialShare/pull/83)

That's interesting... For this PR though, no new option is added, but renamed. The old "Experiments and runs" is not among the existing list of options that it currently checks. Maybe the inventory one gets checked because it's shown at root folder? For Experiments and runs", the shown/hide is controlled by data, so probably it's not gonna be an issue.. I kicked off an TC run on git suite to see how it goes. Thanks for this info!

XingY:
> @XingY FYI, that I had to update the TrialShare related tests for the new "Inventory locations and items" folder archive option. I assume this change to the experiments folder archive option will need a small tweak to these files as well: [LabKey/trialShare#83](https://github.com/LabKey/trialShare/pull/83)

@cnathe So, the trialshare test did fail for me. But after looking into it, it seems to be caused by me unintentionally removing the selectedByDefault(AbstractFolderContext.ExportType type) override in FolderXarWriterFactory. I checked in fix and will kick off another TC run. Thanks!

labkey-susanh:
> > @XingY FYI, that I had to update the TrialShare related tests for the new "Inventory locations and items" folder archive option. I assume this change to the experiments folder archive option will need a small tweak to these files as well: [LabKey/trialShare#83](https://github.com/LabKey/trialShare/pull/83)
> 
> @cnathe So, the trialshare test did fail for me. But after looking into it, it seems to be caused by me unintentionally removing the selectedByDefault(AbstractFolderContext.ExportType type) override in FolderXarWriterFactory. I checked in fix and will kick off another TC run. Thanks!

@XingY We want to change that selectedByDefault to be true now for the "Experiments, Protocols, and Runs" option, per Hannah's comments in the spec.  Sorry for not capturing that in the tasks.

labkey-susanh (approved):

PR #1941, opened 2021-01-29
Issue 42244: Add hacky check for 1904-based dates in excel files

labkey-susanh:
#### Rationale
[Issue 42244](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42244)

#### Changes
* Add method that can be used to determine the date system in use by an Excel file being imported.

### DIFF
--- a/api/src/org/labkey/api/reader/ExcelLoader.java
+++ b/api/src/org/labkey/api/reader/ExcelLoader.java
@@ -57,8 +57,10 @@ import java.io.InputStream;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Calendar;
 import java.util.Collection;
 import java.util.Date;
+import java.util.GregorianCalendar;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -74,6 +76,7 @@ public class ExcelLoader extends DataLoader
 {
     public static FileType FILE_TYPE = new FileType(Arrays.asList(".xlsx", ".xls"), ".xlsx",
             Arrays.asList("application/" + ExcelFactory.SUB_TYPE_BIFF8, "application/" + ExcelFactory.SUB_TYPE_XSSF, "application/" + ExcelFactory.SUB_TYPE_BIFF5));
+    private Boolean _isStartDate1904 = null;
 
     static {
         FILE_TYPE.setExtensionsMutuallyExclusive(false);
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/reader/ExcelLoader.java
```diff
@@ -155,6 +158,45 @@ public ExcelLoader(Workbook workbook, boolean hasColumnHeaders, Container mvIndi
...
+            Row row = sheet.createRow(sheet.getLastRowNum() + 1);
+            Cell cell = row.createCell(0);
+            cell.setCellValue(0.0);
+            double value = cell.getNumericCellValue();
+            Date date = cell.getDateCellValue();
+            Calendar cal = new GregorianCalendar();
+            cal.setTime(date);
+            long year1900 = cal.get(Calendar.YEAR)-1900;
+            long yearEst1900 = Math.round(value/(365.25));
+            _isStartDate1904 =  year1900 > yearEst1900;
```
labkey-matthewb: isn't this just a constant  e.g. year1900 > 0?

api/src/org/labkey/api/reader/ExcelLoader.java
```diff
@@ -155,6 +158,45 @@ public ExcelLoader(Workbook workbook, boolean hasColumnHeaders, Container mvIndi
...
+    // "Excel supports two date systems. Each date system uses a unique starting date from which all other workbook
+    // dates are calculated. Excel 2008 for Mac and earlier Excel for Mac versions calculate dates based on the 1904
+    // date system. Excel for Mac 2011 uses the 1900 date system, which guarantees date compatibility with Excel for
+    // Windows. All versions of Excel for Windows calculate dates based on the 1900 date system."
+    //
+    // When using the SAXParser, we need to know whether this current workbook/sheet is 1904-base or 1900-based.
+    // We are not using getDateCellValue in that case, which knows how to determine which date system is in use, so
+    // we hack up our own way to detect this.  This is derived from this posting:
+    // https://alandix.com/code2/apache-poi-detect-1904-date-option/
+    void setIsStartDate1904()
```
labkey-matthewb: minor thing.  As a habit I've started avoiding JavaBean like method names for things that aren't getter/setter.  Like using computeIsStarteDate1904() or something like that.  Not that it probably matters here, but I've had to change code due to odd errors in reflection/binding code.

labkey-matthewb (approved):

api/src/org/labkey/api/reader/ExcelLoader.java
labkey-susanh: Good point. Simplified.

api/src/org/labkey/api/reader/ExcelLoader.java
labkey-susanh: Good to note.  Updated to `computeIsStartDate1904`

PR #1942, opened 2021-01-29
Update components view with OntologyBrowserPanel

Sigmonia:
#### Rationale
Add the new OntologyBrowser component to the UI components view.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1942
* https://github.com/LabKey/labkey-ui-components/pull/439
* https://github.com/LabKey/ontology/pull/21

#### Changes
* Update core-components.view with the new OntologyBrowserPanel
* Bumped core/package.json to latest w/ alpha


core/src/client/LabKeyUIComponentsPage/LabKeyUIComponentsPage.tsx
```diff
@@ -364,6 +366,11 @@ export class App extends React.Component<any, State> {
                         <LoadingSpinner msg={'Loading message goes here...'}/>
                     )
                 }
+                {selected === 'OntologyBrowser' &&
+                    this.renderPanel('OntologyBrowser',
+                        <OntologyBrowserPanel ontologyId={ 7 } />
```
cnathe: if we end up leaving this on the components page, we'll want to have a way for the user to set the ontologyId for this param, but we can figure that part out later (especially since we might end up moving this panel to the ontology view/page).

cnathe (approved):

PR #1945, opened 2021-01-29
Issue 42355: Perf problem during copy-to-study with many rows

labkey-jeckels:
#### Rationale
Users want to be able to copy large assay runs to study datasets. It's painfully slow because we are iterating on each row and doing lots of DB queries

#### Changes
* Switch AssayProvider to encourage implementations to query rows in bulk
* Cache ExpData, ExpRun, and ExpExperiment references during process as they're likely to be shared by result rows
* Give user more feedback on what's happening during confirmation page
* Optimize date math for visit calculation

### DIFF
--- a/api/src/org/labkey/api/assay/AbstractAssayProvider.java
+++ b/api/src/org/labkey/api/assay/AbstractAssayProvider.java
@@ -801,13 +801,6 @@ public abstract class AbstractAssayProvider implements AssayProvider
         return result;
     }
 
-//    @Override
-//    public ExpData getDataForDataRow(Object dataRowId, ExpProtocol protocol)
-//    {
-//        Set<ExpData> datas = getDatasForDataRows(Collections.singleton(dataRowId), protocol, new ResolverCache());
-//        return datas.isEmpty() ? null : datas.iterator().next();
-//    }
-
     @Nullable
     public final ExpData getDataForDataRow(int resultRowId, ExpProtocol protocol)
     {
@@ -833,6 +826,7 @@ public abstract class AbstractAssayProvider implements AssayProvider
             {
                 return null;
             }
+            // Don't use computeIfAbsent() because it treats null values as if they weren't in the map at all
             if (cache.containsKey(key))
             {
                 return cache.get(key);
--- a/api/src/org/labkey/api/assay/AbstractAssayProvider.java
+++ b/api/src/org/labkey/api/assay/AbstractAssayProvider.java
@@ -133,6 +133,7 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Supplier;
 
 /**
  * User: jeckels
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/assay/AbstractAssayProvider.java
```diff
@@ -757,52 +758,106 @@ public ExpProtocol createAssayDefinition(User user, Container container, String
...
         {
+            Container container = null;
+            if (data.getRunId() != null)
+            {
+                ExpRun run = cache.getRun(data.getRunId());
+                if (run != null)
+                {
+                    // Ignore Results domain TargetStudy for now.
+                    // The participant resolver will find the TargetStudy on the row.
+                    ExpObject source = switch (pair.first)
```
labkey-kevink, line 783: Ooh, switch expression is fun!

api/src/org/labkey/api/assay/AbstractAssayProvider.java
```diff
@@ -757,52 +758,106 @@ public ExpProtocol createAssayDefinition(User user, Container container, String
...
+                            container = ContainerManager.getForId(targetStudyId);
+                    }
+                }
+            }
+            result.add(container);
         }
+        return result;
+    }
+
+//    @Override
```
labkey-kevink: remove?

api/src/org/labkey/api/assay/AbstractAssayProvider.java
```diff
@@ -757,52 +758,106 @@ public ExpProtocol createAssayDefinition(User user, Container container, String
...
+
+        private <K, T extends ExpObject> T get(K key, Map<K, T> cache, Supplier<T> supplier)
         {
-            Map<String, Object> properties = OntologyManager.getProperties(source.getContainer(), source.getLSID());
-            String targetStudyId = (String) properties.get(targetStudyColumn.getPropertyURI());
+            if (key == null)
+            {
+                return null;
+            }
+            if (cache.containsKey(key))
```
labkey-kevink, line 830: Can this be replaced with `cache.computeIfAbsent(key, k -> supplier.get())`?

study/src/org/labkey/study/model/StudyImpl.java
```diff
@@ -1233,7 +1234,19 @@ public static int daysBetween(Calendar startDate, Calendar endDate)
...
+        int daysBetween = (int)(diffInMillis / TimeUnit.DAYS.toMillis(1));
+
+        if (daysBetween > 0)
+        {
+            // Round down to ensure consistent results with original implementation, which added one day at a time and
+            // was a performance bottleneck when the supplied dates were many years apart
+            daysBetween--;
+            date.add(Calendar.DAY_OF_MONTH, daysBetween);
+        }
+
```
labkey-kevink, line 1249: Shouldn't the while loop after this change be removed now that are calculating the date diff more efficiently?

labkey-kevink (approved):

api/src/org/labkey/api/assay/AbstractAssayProvider.java
labkey-jeckels: Yes

api/src/org/labkey/api/assay/AbstractAssayProvider.java
labkey-jeckels: computeIfAbsent() computes if the value is null, which means that it wouldn't cache misses. I'll add a comment.

study/src/org/labkey/study/model/StudyImpl.java
labkey-jeckels: Just pushed a test case to ensure consistency with the original (and questionable) implementation, plus comments.

labkey-klum (approved):

PR #1946, opened 2021-01-30
Fb sample type export fixes

labkey-klum:
#### Rationale
In order to address some recent issues with sample type and data class export/import we have been reconsidering the roles that the XAR and query update service play. In the current design, rows that represent derivation or lineage are captured in the XAR format and all other rows are captured as TSV data. In this PR we pivot so that all rows are exported/imported as TSV, this will happen first and allows any special handling that the QUS may provide on all rows. The XAR handlers will then just be responsible for wiring up and creating derivation runs but will not be tasked with creating materials or datas.

Related issues:
- 42359: SampleTypeFolderExportImportTest.testExportImportDerivedSamples on sqlserver
- 42356: Folder import fails when there is an existing sample type of the same name
- 42161: Unable to import Data Classes as inputs/outputs for Sample Types

#### Changes
- Allow for creation of 2 XAR files, one for the sample type and data class definitions and another to represent any derivation protocol runs
- Export/import all rows via TSV
- Make `XarReader` a bit smarter about resolving existing sample types or data classes 
- Add XarImportContext to share XAR related settings between folder importers

### DIFF
--- a/experiment/src/org/labkey/experiment/LSIDRelativizer.java
+++ b/experiment/src/org/labkey/experiment/LSIDRelativizer.java
@@ -57,8 +57,6 @@ public enum LSIDRelativizer implements SafeToRenderEnum
             {
                 ExpData data = (ExpData)o;
                 // Most DataClass data don't have a dataFileUrl, but some do -- like NucSequence imported from a genbank file
-                // UNDONE: Don't have user and fetching DataClass is expensive-ish. Maybe add data.hasDataClass() or something
-                // UNDONE: Maybe there is a better way to detect when we should use ${AutoFileLSID}?
                 if (data.getDataFileUrl() == null || data.getDataClass() != null)
                 {
                     // If we don't have a URL for this data object, we can't use AutoFileLSID. Instead,
--- a/experiment/src/org/labkey/experiment/LSIDRelativizer.java
+++ b/experiment/src/org/labkey/experiment/LSIDRelativizer.java
@@ -56,7 +56,10 @@ public enum LSIDRelativizer implements SafeToRenderEnum
             if (o instanceof ExpData)
             {
                 ExpData data = (ExpData)o;
-                if (data.getDataFileUrl() == null)
+                // Most DataClass data don't have a dataFileUrl, but some do -- like NucSequence imported from a genbank file
+                // UNDONE: Don't have user and fetching DataClass is expensive-ish. Maybe add data.hasDataClass() or something
+                // UNDONE: Maybe there is a better way to detect when we should use ${AutoFileLSID}?
+                if (data.getDataFileUrl() == null || data.getDataClass() != null)
                 {
                     // If we don't have a URL for this data object, we can't use AutoFileLSID. Instead,
                     // try the next best option
@@ -86,6 +89,9 @@ public enum LSIDRelativizer implements SafeToRenderEnum
             }
             else if ("Data".equals(prefix))
             {
+                // UNDONE: Now that "Data" prefix is used for DataClass, the AutoFileLSID is not a good default.
+                // UNDONE: Can we be more restrictive about which LSIDs this is applied to?  Maybe only if the objectId part of the LSID includes a "/" (%2F) or something?
+                // UNDONE: Maybe there is a better way to detect when we should use ${AutoFileLSID}?
                 return AutoFileLSIDReplacer.AUTO_FILE_LSID_SUBSTITUTION;
             }
             else if (suffix != null && SUFFIX_PATTERN.matcher(suffix).matches())
--- a/experiment/src/org/labkey/experiment/XarExporter.java
+++ b/experiment/src/org/labkey/experiment/XarExporter.java
@@ -146,9 +146,10 @@ public class XarExporter
     AssayProvider.XarCallbacks assayCallbacks = null;
 
 
-    public XarExporter(LSIDRelativizer lsidRelativizer, URLRewriter urlRewriter, User user)
+    public XarExporter(LSIDRelativizer.RelativizedLSIDs relativizedLSIDs, URLRewriter urlRewriter, User user)
     {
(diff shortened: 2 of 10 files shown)


experiment/src/org/labkey/experiment/LSIDRelativizer.java
```diff
@@ -86,6 +89,9 @@ else if ("Sample".equals(prefix) || "Material".equals(prefix))
             }
             else if ("Data".equals(prefix))
             {
+                // UNDONE: Now that "Data" prefix is used for DataClass, the AutoFileLSID is not a good default.
+                // UNDONE: Can we be more restrictive about which LSIDs this is applied to?  Maybe only if the objectId part of the LSID includes a "/" (%2F) or something?
+                // UNDONE: Maybe there is a better way to detect when we should use ${AutoFileLSID}?
```
labkey-kevink, line 92: @labkey-jeckels Any thoughts about these comments?  We don't want to use ${AutoFileLSID} for DataClass data and I don't think we have enough context from the LSID alone to know if it is an exp.data for a file or a DataClass.

experiment/src/org/labkey/experiment/XarExporter.java
```diff
@@ -146,19 +146,30 @@
     AssayProvider.XarCallbacks assayCallbacks = null;
 
 
-    public XarExporter(LSIDRelativizer lsidRelativizer, URLRewriter urlRewriter, User user)
+    public XarExporter(LSIDRelativizer.RelativizedLSIDs relativizedLSIDs, URLRewriter urlRewriter, User user)
     {
-        _relativizedLSIDs = new LSIDRelativizer.RelativizedLSIDs(lsidRelativizer);
+        // UNDONE: Is it ok to share the relativizedLSIDs across XarExporters and tsv writers?
```
labkey-kevink, line 151: It seems ok

experiment/src/org/labkey/experiment/XarExporter.java
```diff
@@ -517,7 +528,7 @@ private void populateXmlBeanValue(SimpleValueType xValue, AbstractParameter para
     }
 
     private String relativizeLSIDPropertyValue(String value, SimpleTypeNames.Enum type)
-    {
+    {// NOTE: this is interesting - we fixup LSID values in the xar xml
```
labkey-kevink, line 531: comment was a note to self. remove comment

experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
```diff
@@ -147,29 +155,41 @@ public void write(Container object, ImportContext<FolderDocument.Folder> ctx, Vi
...
-            selection.addRuns(exportedRuns);
-            exportXar = true;
+            runsSelection.addRuns(exportedRuns);
+            exportRuns = true;
         }
         VirtualFile xarDir = vf.getDir(DEFAULT_DIRECTORY);
 
-        if (exportXar)
+        // UNDONE: The other exporters use FOLDER_RELATIVE, but it wants to use ${AutoFileLSID} replacements for DataClass LSIDs when exporting the TSV data.. see comment in ExportLsidDataColumn
+        LSIDRelativizer.RelativizedLSIDs relativizedLSIDs = new LSIDRelativizer.RelativizedLSIDs(LSIDRelativizer.FOLDER_RELATIVE);
```
labkey-kevink, line 164: I think the `relativizedLSIDs` instance should be shared across all of the exporters by stashing it in the `ImportContext<FolderDocument.Folder>`

experiment/src/org/labkey/experiment/XarReader.java
```diff
@@ -1353,8 +1353,23 @@ private Data loadData(DataBaseType xbData,
         checkDataCpasType(declaredType);
 
         String dataLSID = LsidUtils.resolveLsidFromTemplate(xbData.getAbout(), context, declaredType, new AutoFileLSIDReplacer(xbData.getDataFileUrl(), getContainer(), _xarSource));
-
         ExpDataImpl expData = ExperimentServiceImpl.get().getExpData(dataLSID);
+        ExpDataClass expDataClass = ExperimentService.get().getDataClass(declaredType);
+        if (expData == null && expDataClass != null)
+        {
+            // Try resolving it by name within the data class in case we have it under a different LSID
+            ExpData data = expDataClass.getData(getContainer(), xbData.getName());
```
labkey-jeckels: Should be able to avoid this instanceof by using ExperimentServiceImpl to fetch a ExpDataClassImpl. Its getData() returns a ExpDataImpl

experiment/src/org/labkey/experiment/LSIDRelativizer.java
```diff
@@ -56,7 +56,10 @@ protected String relativize(ExpObject o, RelativizedLSIDs lsids)
             if (o instanceof ExpData)
             {
                 ExpData data = (ExpData)o;
-                if (data.getDataFileUrl() == null)
+                // Most DataClass data don't have a dataFileUrl, but some do -- like NucSequence imported from a genbank file
+                // UNDONE: Don't have user and fetching DataClass is expensive-ish. Maybe add data.hasDataClass() or something
```
labkey-jeckels: Are these UNDONEs intended to stay here?

labkey-jeckels (approved):

labkey-kevink (commented):
This looks good but I think we need another pass to clear up some of the UNDONE comments.

experiment/src/org/labkey/experiment/LSIDRelativizer.java
labkey-jeckels: I don't know what the DataClass LSIDs look like, but maybe we should build out RelativizedLSIDs.relativize(ExpObject) so that we can use more of the ExpData to help sniff the right approach (checking for a DataFileURL, for example).

PR #1949, opened 2021-02-01
Conditionalize Admin Console link in nav trail based on container

labkey-jeckels:
#### Rationale
The pipeline status action (grid view) is used both for individual containers and site-wide, as linked from the Admin Console

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1940

#### Changes
* Don't show Admin Console link when the list is viewed for a single container

### DIFF
--- a/api/src/org/labkey/api/Constants.java
+++ b/api/src/org/labkey/api/Constants.java
@@ -49,7 +49,7 @@ public class Constants
      */
     public static String getDocumentationVersion()
     {
-        return "20.11";
+        return "21.3";
     }
 
     /**
--- a/api/src/org/labkey/api/admin/AdminUrls.java
+++ b/api/src/org/labkey/api/admin/AdminUrls.java
@@ -24,6 +24,7 @@ import org.labkey.api.util.URLHelper;
 import org.labkey.api.util.emailTemplate.EmailTemplate;
 import org.labkey.api.view.ActionURL;
 import org.labkey.api.view.NavTree;
+import org.springframework.web.servlet.mvc.Controller;
 
 /**
  * User: jgarms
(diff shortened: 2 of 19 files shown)


labkey-adam (commented):
I suspect this is not the only admin action with this need and indeed the `addAdminNavTrail()` method purports to conditionalize based on root vs. not, but the container is not getting passed through for non-AdminController actions. Can we adjust or create a varitant of `addAdminNavTrail` to take the current container or pull it from the URL if non-null?

labkey-jeckels:
Change pushed. Feel free to run with this. I'd love to have this be more general purpose, but figured I'd create a PR instead of a bug opened to you...

labkey-jeckels:
We could inspect the ActionURL to resolve a Container, but it only has an getExtraPath() method. We could call ContainerManager.getForPath() instead of having an overloaded method.

labkey-adam:
I'll take a look later today. One option: one method that doesn't take URL or container, for dedicated Admin Console actions, and another that always takes container and current action class, same as the AdminController methods. We'd always have a proper container to check and would avoid the need for boilerplate URL construction in every action. Would be more invasive, but we could deprecate the current method and migrate repos over time.

labkey-adam:
I opted for a single method going forward... always pass in action class and container, and we should do the right thing, whether root or not. This addresses only platform actions, of course. Once we review and merge this PR, we can circle back to the other repos that use the deprecated method.

labkey-jeckels (commented):
This approach looks fine to me. I can't approve because I opened the PR.

labkey-adam:
Yep, was waiting for green tests before approving. I just did the doc version commit, so it'll be a few minutes...

labkey-adam (approved):

PR #1951, opened 2021-02-01
Compliance: OwnerEntityId, DownloadSnapshot

labkey-nicka:
#### Rationale
Exposes `downloadSnapshot` and `getSignedSnapshots` on the `ComplianceService`. See related [compliance PR](https://github.com/LabKey/compliance/pull/103) for more information.

Co-author: @labkey-klum 

#### Related Pull Requests
* https://github.com/LabKey/compliance/pull/103
* https://github.com/LabKey/biologics/pull/782
* https://github.com/LabKey/labbook/pull/90

#### Changes
* Update `ComplianceService` interface.
* Update `SignedSnapshot` model to include `OwnerEntityId` property.


api/src/org/labkey/api/compliance/ComplianceService.java
```diff
@@ -120,10 +127,22 @@ public SignedSnapshot getSignedSnapshot(Container container, int snapshotId)
...
+        }
+
         @Override
         public Map<String, Object> deleteSignedSnapshot(Container container, User user, int snapshotId)
         {
             return null;
         }
+
+        @Override
+        public Pair<AttachmentParent, String> downloadSnapshot(Container container, User user, SignedSnapshot snapshot)
```
labkey-adam, line 143: `@Nullable`?

api/src/org/labkey/api/compliance/ComplianceService.java
```diff
@@ -120,10 +127,22 @@ public SignedSnapshot getSignedSnapshot(Container container, int snapshotId)
             return null;
```
labkey-adam, line 127: `@Nullable`? Obviously pre-existing code, but it's concerning that this isn't annotated and existing callers (e.g., NotebookManager) don't check null. Is ELN expected to blow up if compliance isn't present?

api/src/org/labkey/api/compliance/ComplianceService.java
labkey-adam: Should be annotating interface as well, of course

labkey-adam (approved):

api/src/org/labkey/api/compliance/ComplianceService.java
labkey-nicka: With regards to both of your comments -- the story is a bit deeper. 

`ComplianceService` interface is defined in `platform/api` and the "real" implementation is `ComplianceServiceImpl` in the `compliance` module. However, if `compliance` module is not available, then the `DefaultComplianceService` will be returned which essentially is just an interface to `null` (most of the methods return null).

`NotebookManager` does not check for null because the `compliance` module is required for Notebook's to work as it relies on signed snapshots for persisting artifacts. Compliance was made an explicit dependency at build time with https://github.com/LabKey/labbook/pull/31. I guess we could attempt to re-check at runtime but it seems a bit overwrought. If the service implementation ever returns `DefaultComplianceService` notebooks will not be having a good time.

Open to other ideas and/or patterns where the implementation lives in a premium module.

api/src/org/labkey/api/compliance/ComplianceService.java
labkey-nicka: See https://github.com/LabKey/platform/pull/1951#discussion_r568338411

PR #1962, opened 2021-02-03
ExistingRecordDataIterator

labkey-matthewb:
#### Rationale
For consistency and performance create a standard DataIterator pattern for detailed audit logging.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1962
https://github.com/LabKey/inventory/pull/185

#### Changes
* Select existing record using ExistingRecordDataIterator
* Consolidate logging in AuditHandler classes
* Support bulk insert of audit records in AuditLogService and LogManager


api/src/org/labkey/api/audit/AuditHandler.java
```diff
@@ -6,6 +6,8 @@
 import org.labkey.api.data.AuditConfigurable;
```
labkey-susanh, line 6: Are you planning to leave the abstract class inside the AuditHandler interface?

api/src/org/labkey/api/audit/DefaultAuditProvider.java
```diff
@@ -51,6 +51,11 @@ public boolean isViewable()
         return null;
     }
 
+    @Override
+    public <K extends AuditTypeEvent> void addEvents(User user, List<K> events)
```
labkey-susanh, line 55: Java's syntax for this is weird.

api/src/org/labkey/api/dataiterator/AbstractDataIterator.java
```diff
@@ -140,4 +141,23 @@ protected String getDataSource()
...
+
+    protected Integer indexOfExistingRecord;
+
+    @Override
+    public @Nullable Map<String, Object> getExistingRecord()
+    {
+        if (null == indexOfExistingRecord)
+        {
+            indexOfExistingRecord = -1;
+            for (int i=1 ; i<=getColumnCount() ; i++)
```
labkey-susanh, line 155: Why does this start at 1? (probably a good reason, but maybe a comment?)

api/src/org/labkey/api/dataiterator/AbstractDataIterator.java
```diff
@@ -140,4 +141,23 @@ protected String getDataSource()
...
+            {
+                if (ExistingRecordDataIterator.EXISTING_RECORD_COLUMN_NAME.equals(getColumnInfo(i).getName()))
+                    indexOfExistingRecord = i;
+            }
+        }
+        assert( supportsGetExistingRecord() ? indexOfExistingRecord>1 : indexOfExistingRecord<0);
```
labkey-susanh: Seems like it can be 1, shouldn't that be `>=1`?

api/src/org/labkey/api/dataiterator/DataIterator.java
```diff
@@ -35,6 +36,8 @@
  */
 public interface DataIterator extends DataIteratorBuilder, Closeable
 {
+    String ROWNUMBER_COLUMNNAME = "_rowNumber";        // TODO change to something like DataIterator.class().getName() + "#_rowNumber"
```
labkey-susanh, line 39: Is this a TODO for now or later?

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -0,0 +1,220 @@
...
+{
+    public static final String EXISTING_RECORD_COLUMN_NAME = "_" + ExistingRecordDataIterator.class.getName() + "#EXISTING_RECORD_COLUMN_NAME";
+
+    Map<String,Object> currentExistingRecord = null;
+
+    final TableInfo target;
+    final ArrayList<ColumnInfo> pkColumns = new ArrayList<>();
+    final ArrayList<Supplier<Object>> pkSuppliers = new ArrayList<>();
+
+    final int existingColIndex;
```
labkey-susanh, line 40: So as not to suffer from the confusing name affliction again, does this mean the index for the column that holds the existing record?  Would it be better just called `colIndex`?

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -0,0 +1,220 @@
...
+    {
+        return existingColIndex;
+    }
+
+    @Override
+    public ColumnInfo getColumnInfo(int i)
+    {
+        if (i<existingColIndex)
+            return _delegate.getColumnInfo(i);
+        return new BaseColumnInfo(EXISTING_RECORD_COLUMN_NAME, JdbcType.OTHER);
```
labkey-susanh, line 76: Will `i` never be larger than `existingColIndex`?

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -0,0 +1,220 @@
...
+        if (i<existingColIndex)
+            return _delegate.getConstantValue(i);
+        return null;
+    }
+
+    @Override
+    public boolean next() throws BatchValidationException
+    {
+        if (pkColumns.size() > 1)
+            throw new UnsupportedOperationException("NYI");
```
labkey-susanh, line 115: Perhaps a more informative message here: "Support for multiple key values not yet implemented"?

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -0,0 +1,220 @@
...
+        currentExistingRecord = null;
+        boolean ret = super.next();
+        if (ret && !pkColumns.isEmpty())
+        {
+            SimpleFilter filter = new SimpleFilter();
+            for (int i=0 ; i<pkColumns.size() ; i++)
+                filter.addClause(new CompareType.EqualsCompareClause(pkColumns.get(i).getFieldKey(), CompareType.EQUAL, pkSuppliers.get(i).get()));
```
labkey-susanh, line 122: Looping isn't strictly necessary here since the size will always be 1 at this point, but no harm.

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -0,0 +1,220 @@
...
+            currentExistingRecord = null==record ? Map.of() : record;
+        }
+        return ret;
+    }
+
+
+    @Override
+    public boolean supportsGetExistingRecord()
+    {
+        return !pkColumns.isEmpty();
```
labkey-susanh, line 133: Should this be `pkColumns.size() == 1`?

api/src/org/labkey/api/dataiterator/WrapperDataIterator.java
```diff
@@ -108,6 +109,24 @@ public Object getConstantValue(int i)
...
+    {
+        return _delegate.supportsGetExistingRecord();
+    }
+
+
+    /* if !supportsGetExistingRecord() then this method should return null
+     * if supportsGetExistingRecord() this method should return a map, an empty map means no existing record found
+     */
+    @Override
+    public Map<String, Object> getExistingRecord()
```
labkey-susanh, line 123:
```suggestion
    @Nullable public Map<String, Object> getExistingRecord()
```

audit/src/org/labkey/audit/AuditLogImpl.java
```diff
@@ -161,15 +166,19 @@ public boolean isViewable()
                 {
                     if (_logToDatabase.get())
                     {
-                        LogManager.get()._insertEvent(user, event);
+                        for (var event : events)
+                            LogManager.get().insertEvent(user, event);
```
labkey-susanh, line 170: Could use a `forEach` here and below if you were so inclined.

audit/src/org/labkey/audit/model/LogManager.java
```diff
@@ -69,39 +73,82 @@ public DbSchema getSchema()
...
+        if (optimize)
         {
-            Container c = ContainerManager.getForId(type.getContainer());
+            for (var event : events)
+            {
+                if (!Objects.equals(type.getEventType(), event.getEventType()))
+                    optimize = false;
+                if (!Objects.equals(type.getContainer(), event.getContainer()))
+                    optimize = false;
+            }
```
labkey-susanh: You could optimize the optimize loop by bailing as soon as `optimize` is false, perhaps using `stream()` and `findFirst()`.

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -842,8 +844,8 @@ private void addAttachments(User user, Container c, Map<String, Object> row, Str
         public DataIteratorBuilder createImportDIB(User user, Container container, DataIteratorBuilder data, DataIteratorContext context)
         {
             StandardDataIteratorBuilder etl = StandardDataIteratorBuilder.forInsert(getQueryTable(), data, container, user, context);
-
-            DataIteratorBuilder dib = ((UpdateableTableInfo)getQueryTable()).persistRows(etl, context);
+            DataIteratorBuilder existing = ExistingRecordDataIterator.createBuilder(etl, getQueryTable(), Set.of(ExpDataTable.Column.LSID.toString()));
```
labkey-susanh: Do you want to rename `etl` to `standard` as you did elsewhere (better name, I think).

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1599,48 +1605,115 @@ public PHI getMaxContainedPhi()
...
+                return;
+            Objects.requireNonNull(rows);
+            AuditLogService auditLog = AuditLogService.get();
+
+            if (null == existingRows && action == QueryService.AuditAction.MERGE)
             {
-                @Override
-                public void addSummaryAuditEvent(User user, Container c, TableInfo table, QueryService.AuditAction action, Integer dataRowCount)
+                assert false : "Please provide list of existing rows!";
+                for (int i=0; i < rows.size(); i++)
```
labkey-susanh: Uh, why is this loop here if we `assert false` above?

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1599,48 +1605,115 @@ public PHI getMaxContainedPhi()
...
+            else
+            {
+                newRecordString = DatasetAuditProvider.encodeForDataMap(c, record);
+            }
+            event.setLsid(lsid == null ? null : lsid.toString());
+
+            if (oldRecordString != null) event.setOldRecordMap(oldRecordString);
+            if (newRecordString != null) event.setNewRecordMap(newRecordString);
+
+            return event;
```
labkey-susanh, line 1701: Seems like we're duplicating some logic from AuditHandler here.  Is it necessary to do that?

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -2171,15 +2244,20 @@ public DataIteratorBuilder getInsertDataIterator(User user, DataIteratorBuilder
         b.setKeyList(lsids);
 
         Container target = getDataSharingEnum() == DataSharing.NONE ? getContainer() : getDefinitionContainer();
-        DataIteratorBuilder dib = StandardDataIteratorBuilder.forInsert(table, b, target, user, context);
+        DataIteratorBuilder standard = StandardDataIteratorBuilder.forInsert(table, b, target, user, context);
 
         if (context.getConfigParameter(DetailedAuditLogDataIterator.AuditConfigs.AuditBehavior) == AuditBehaviorType.DETAILED)
-            dib = DetailedAuditLogDataIterator.getDataIteratorBuilder(this.getTableInfo(user, false), dib, context.getInsertOption() == QueryUpdateService.InsertOption.MERGE ? QueryService.AuditAction.MERGE : QueryService.AuditAction.INSERT, user, target);
+        {
+        }
```
labkey-susanh: Is this `if` block really empty? Can it be removed now?

labkey-susanh (approved):
Just a few suggestions here, but nothing that needs changing, I think.  The SampleManagement and Inventory timeline and location history tests are a good indication of whether the ordering of old record and new record is as it should be.  I'm seeing diffs that seem like they should already be in develop now that 21.2 has been merged there.  Perhaps you need to merge from develop still?

Anyway, very excited to see this go in.  It'll be a great performance gain!

api/src/org/labkey/api/audit/AuditHandler.java
labkey-matthewb: Right.  Let's fix that...

api/src/org/labkey/api/dataiterator/AbstractDataIterator.java
labkey-matthewb: added comment.
PS 0 is always the row number

api/src/org/labkey/api/dataiterator/AbstractDataIterator.java
labkey-matthewb: thank you  (would _strange_ to be 1, but yes that assert is wrong)

api/src/org/labkey/api/dataiterator/DataIterator.java
labkey-matthewb: Later, just seems better if special columns have really special names.

study/src/org/labkey/study/model/DatasetDefinition.java
labkey-matthewb: oops

study/src/org/labkey/study/model/DatasetDefinition.java
labkey-matthewb: Nice catch, kinda looks like it.  I think I'll tackle in the next PR, just so I can get this in...

PR #1963, opened 2021-02-04
Fix auto-complete NPEs

labkey-adam:
#### Rationale
The previous PR's migration of auto-complete URL handling from String toward ActionURL introduced NPE opportunities since the final rendering was still String-based, causing tests to fail. Those NPEs are fixed by pushing ActionURLs all the way to render time, which completes the migration. "Auto-completion completion", if you will. (And even if you won't.)

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1953

#### Changesg
* Push ActionURL through to DataColumn and publish study rendering of auto-complete elements
* Mark `completionBase` parameters as `@Nullable` throughout
* While we're at it, don't grab the current request from thread local when you have a RenderContext in hand

### DIFF
--- a/api/src/org/labkey/api/data/DataColumn.java
+++ b/api/src/org/labkey/api/data/DataColumn.java
@@ -46,6 +46,7 @@ import org.labkey.api.util.element.Option;
 import org.labkey.api.util.element.Option.OptionBuilder;
 import org.labkey.api.util.element.Select;
 import org.labkey.api.util.element.TextArea;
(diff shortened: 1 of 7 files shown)


labkey-adam:
Elispot + Elisa and NAb tests are all green on this branch, except for NAb SQL Server which appears to have unrelated failures. FYI, @labkey-danield

labkey-susanh (approved):
Seems fairly mechanical.

PR #1968, opened 2021-02-07
Product MenuSection Updates

labkey-nicka:
#### Rationale
Adds support for `sectionKey` to `MenuSection` to allow for differentiating the section from the `key`.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/443
* https://github.com/LabKey/biologics/pull/787
* https://github.com/LabKey/sampleManagement/pull/479
* https://github.com/LabKey/inventory/pull/186

#### Changes
* Additional `MenuItem` constructor to support passing a key.
* Add `_sectionKey` prop to `MenuSection`.

### DIFF
--- a/core/api-src/org/labkey/api/products/MenuSection.java
+++ b/core/api-src/org/labkey/api/products/MenuSection.java
@@ -38,9 +38,14 @@ public abstract class MenuSection
     protected ViewContext _context;
     protected String _label;
     protected Integer _itemLimit;
+    /**
+     * A unique key identifying this MenuSection.
+     * This is used by the client to match client configuration to the configuration supplied from LKS.
+     */
     protected String _key;
     private List<MenuItem> _allItems;
     private String _productId;
+    /** Used by the client to determine the first part of an app-relative URL. Defaults to the same as "key". */
     private String _sectionKey;
 
     public MenuSection(@NotNull ViewContext context, @NotNull String label, @NotNull String key, @Nullable Integer itemLimit, @Nullable String productId)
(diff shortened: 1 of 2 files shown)


core/api-src/org/labkey/api/products/MenuSection.java
```diff
@@ -41,12 +41,14 @@
...
     private List<MenuItem> _allItems;
     private String _productId;
+    private String _sectionKey;
 
     public MenuSection(@NotNull ViewContext context, @NotNull String label, @NotNull String key, @Nullable Integer itemLimit, @Nullable String productId)
     {
         _context = context;
         _label = label;
         _key = key;
+        _sectionKey = key;
```
labkey-susanh, line 56: Will 'key' and 'sectionKey' always be the same?  Do we need both?

labkey-nicka: `BiologicsWorkflowMenuSection` sets them differently. The reason is that the client formerly used the `key` as the first part of an `AppURL`, however, I wanted to have `biologicsWorkflow` and `workflow` not share the same key but share the same `AppURL` part. Thus, I made the client always look at the `sectionKey` which defaults to the same value as `key`.

I purposely didn't update the constructor to take as a separate argument since it is a bit of an edge case and don't want to trip up future usages.

labkey-susanh: Let's add a comment that explains the difference.

labkey-susanh (approved):

core/api-src/org/labkey/api/products/MenuSection.java
labkey-nicka: eba38a2.

PR #1971, opened 2021-02-08
Issue 41458: Custom date-time fields show diffs with no changes on update

labkey-ankurj:
#### Rationale
This fixes the issue in change summary of issue tracker that displays the custom date-time field diff when there are no changes.

### DIFF
--- a/issues/src/org/labkey/issue/actions/ChangeSummary.java
+++ b/issues/src/org/labkey/issue/actions/ChangeSummary.java
@@ -283,8 +283,8 @@ public class ChangeSummary
 
         if (!Objects.equals(from, to))
         {
-            String fromStr = from instanceof Date ? DateUtil.formatDate(container, (Date) from) : String.valueOf(from);
-            String toStr = to instanceof Date ? DateUtil.formatDate(container, (Date) to) : String.valueOf(to);
+            String fromStr = from instanceof Date ? DateUtil.formatDate(container, (Date) from) : null == from ? "" : String.valueOf(from);
+            String toStr = to instanceof Date ? DateUtil.formatDate(container, (Date) to) : null == to ? "" : String.valueOf(to);
             sbText.append(encField);
             if (newIssue)
             {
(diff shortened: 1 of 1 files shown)


labkey-adam (changes requested):
This doesn't look right... oldValue could be anything, an Integer or String or whatever. This code seems to be converting random objects to Strings, parsing the result as a date, and formatting back as a date.

I believe the correct approach is to NOT convert oldValue and newValue to Strings. `_appendCustomColumnChange()`, `_appendColumnChange()`, and `_appendChange()` should all take these as `Object` parameters. Then do a null-safe equals check on these objects (e.g., `Objects.equals()`) in `_appendChange()` and only then convert them to strings, including formatting date objects.

labkey-ankurj:
want to make sure that oldValues and newValues were getting converted to string before this change and we want to change the approach.

labkey-adam:
This problem is caused by converting these values to String too early. My suggestion is to compare Object values and convert to Strings only if the object values differ; the String values should be used only for rendering purposes. Let me know if you want to discuss.

issues/src/org/labkey/issue/actions/ChangeSummary.java
```diff
@@ -247,33 +247,44 @@ static public ChangeSummary createChangeSummary(ViewContext context, IssueListDe
...
         // Use custom caption if one is configured
         String encField = PageFlowUtil.filter(caption);
-        from = from == null ? "" : from;
-        to = to == null ? "" : to;
 
-        if (!from.equals(to))
+        if (!Objects.equals(from, to))
         {
+            String fromStr = from instanceof Date ? DateUtil.formatDate(container, (Date) from) : String.valueOf(from);
+            String toStr = to instanceof Date ? DateUtil.formatDate(container, (Date) to) : String.valueOf(to);
```
labkey-adam: `from` and `to` can be `null` now, right? `String.valueOf(null)` returns the String "null", which I'm sure we don't want. I think you want to set `fromStr` and `toStr` to either `null` or `""` in the null case.

labkey-adam (commented):
Looks much better. Just the one comment about null handling in `_appendChange()`. Or let me know if I've misunderstood something there.

labkey-adam (approved):

PR #1979, opened 2021-02-09
Issue 42420: Support source query filter via ETL XML

labkey-jeckels:
#### Rationale
To make it easier to create ETLs that run against a remote server where the ETL developer may not have permission to create custom queries, it's more convenient to allow the ETL definition to add filters to the source table/query.

#### Changes
* Add filters to ETL XSD
* Parse and propagate them
* Apply them when fetching the source data

### DIFF
--- a/api/src/org/labkey/api/dataiterator/CopyConfig.java
+++ b/api/src/org/labkey/api/dataiterator/CopyConfig.java
@@ -18,6 +18,7 @@ package org.labkey.api.dataiterator;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import org.labkey.api.query.SchemaKey;
+import org.labkey.remoteapi.query.Filter;
 
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


bbimber:
@labkey-jeckels So does this count as clean test results? if you have other thoughts on the Filter serialization problem I'm happy to try to implement something other than what I proposed here: https://github.com/LabKey/labkey-api-java/pull/17/files

labkey-jeckels:
Yes, the tests are looking happy but as Marty mentioned here:

https://github.com/LabKey/dataintegration/pull/90

the same filter should be applied to the check for modified data query. I'm mid-refactor to do that.

bbimber:
@labkey-jeckels OK, that is a good point. If sound like you're already doing that change, but feel free to push any of this back to me if it would help get this closed out.

labkey-martyp (approved):

PR #1980, opened 2021-02-10
Fix switched new/old rows for dataset audit logging

Sigmonia:
#### Rationale
Found an issue with the audit logging for datasets where the audit event was swapping the new/old values

#### Changes
* Changed order to match other overload and existing methods
* Fixed issue where the audit event's new row data would be set incorrectly


labkey-susanh (approved):
Thanks for catching that.

PR #1982, opened 2021-02-10
Return ExpRuns ordered by their RowId

vagisha:
#### Rationale
Runs copied as part of folder export/import should be imported in the same order in the destination folder as they were in the source folder.

#### Changes
ExperimentService.getExpRuns(Container container, @Nullable ExpProtocol parentProtocol, @Nullable ExpProtocol childProtocol) returns ExpRuns in the order of their RowId


labkey-jeckels (approved):

labkey-jeckels:
I've created a matching FB in the LabKey copy of the repo to fire off TeamCity validation.

https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_VerifyJavaBuild/1257540

PR #1991, opened 2021-02-12
Issue 42468: Respect email subscription settings for message board moderators

labkey-jeckels:
#### Rationale
We introduced innovative, industry-leading moderator review workflow for message boards to reduce spam posts. It currently emails everyone who has admin access to the folder, including all site admins.

That's working OK for us on labkey.org (our primary use case) but it's causing more mail than desired on other installs

#### Changes
* Use the message board's email settings first, and then filter to exclude non-admins for receiving the moderator review.

### DIFF
--- a/announcements/src/org/labkey/announcements/model/AnnouncementManager.java
+++ b/announcements/src/org/labkey/announcements/model/AnnouncementManager.java
@@ -341,7 +341,7 @@ public class AnnouncementManager
 
         if (toList.isEmpty())
         {
-            LogManager.getLogger(AnnouncementManager.class).warn("New " + name.toLowerCase() + " requires moderator review, but no moderators are subscribed in this folder: " + c.getPath());
+            LogManager.getLogger(AnnouncementManager.class).warn("New " + name.toLowerCase() + " requires moderator review, but no moderators are subscribed to receive 'all' messages in this folder: " + c.getPath());
         }
         else
         {
--- a/announcements/src/org/labkey/announcements/model/AnnouncementManager.java
+++ b/announcements/src/org/labkey/announcements/model/AnnouncementManager.java
@@ -326,13 +326,22 @@ public class AnnouncementManager
     {
         String name = AnnouncementManager.getMessageBoardSettings(c).getConversationName();
         BulkEmailer emailer = new BulkEmailer(user);
-        List<String> toList = SecurityManager.getUsersWithPermissions(c, Collections.singleton(AdminPermission.class)).stream()
+
+        // Only admins should be part of the moderator review process, but we will still respect their email subscription
+        // preferences for this container
+        IndividualEmailPrefsSelector sel = new IndividualEmailPrefsSelector(c);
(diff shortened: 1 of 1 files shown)


announcements/src/org/labkey/announcements/model/AnnouncementManager.java
```diff
@@ -326,13 +326,22 @@ private static void notifyModerators(Container c, User user, AnnouncementModel a
...
+        admins.retainAll(subscribers);
+
+        List<String> toList = admins.stream()
             .map(User::getEmail)
             .collect(Collectors.toList());
 
         if (toList.isEmpty())
         {
-            LogManager.getLogger(AnnouncementManager.class).warn("New " + name.toLowerCase() + " requires moderator review, but no moderators are authorized in this folder: " + c.getPath());
+            LogManager.getLogger(AnnouncementManager.class).warn("New " + name.toLowerCase() + " requires moderator review, but no moderators are subscribed in this folder: " + c.getPath());
```
labkey-adam: I think this means all admins subscribed to individual emails (regardless of whether they've selected "all" or "mine") will be notified. That seems fine (I guess you could quibble that these posts wouldn't qualify as "mine"), but does it make sense to update the logging here to be a bit more specific?

labkey-jeckels: Just pushed a change - let me know if it conveys what you're suggesting.

announcements/src/org/labkey/announcements/model/AnnouncementManager.java
```diff
@@ -326,13 +326,22 @@ private static void notifyModerators(Container c, User user, AnnouncementModel a
...
+        admins.retainAll(subscribers);
+
+        List<String> toList = admins.stream()
             .map(User::getEmail)
             .collect(Collectors.toList());
 
         if (toList.isEmpty())
         {
-            LogManager.getLogger(AnnouncementManager.class).warn("New " + name.toLowerCase() + " requires moderator review, but no moderators are authorized in this folder: " + c.getPath());
+            LogManager.getLogger(AnnouncementManager.class).warn("New " + name.toLowerCase() + " requires moderator review, but no moderators are subscribed to receive 'all' messages in this folder: " + c.getPath());
```
labkey-adam:
```suggestion
            LogManager.getLogger(AnnouncementManager.class).warn("New " + name.toLowerCase() + " requires moderator review, but no moderators are subscribed to receive 'Individual' notifications in this folder: " + c.getPath());

```

labkey-adam: I believe the filter will include all admins that subscribe to "Individual" notifications (whether 'all' or 'mine'), hence the suggestion

labkey-jeckels: OK, works for me.

labkey-adam (approved):
I approve, though with some reluctance due to the characterization of this feature as "simple"

labkey-jeckels:
@labkey-adam fixed the typo.

PR #1993, opened 2021-02-12
Fix assert to handle duplicate columns in DataLoader

labkey-matthewb:
#### Rationale
DataLoader consistency check did not handle duplicate column names
Issue 41816


labkey-susanh (approved):
So I guess the original issue would not (generally) have come up in a production instance since this is changing only assert behavior here.

PR #1996, opened 2021-02-12
Issue 42476: Support JSON and JSONB operators and functions on Postgres

labkey-jeckels:
#### Rationale
Late last year I added pass-through operator and function support for Postgres's native JSON and JSONB data type. Initially, I only exposed it via the Mothership module so I could use it for metric reporting.

However, we're hearing growing demand to support it more generally

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1798

#### Changes
* Move registration of operators and functions into Core module, out of Mothership


labkey-jeckels:
I'm not sure this is the best place to register the functions, but it seemed reasonable. Open to alternatives

labkey-matthewb (approved):
Since this is a core module, you don't really need to use the registerMethod() service.  This could all just be added to org.labkey.query.sql.Method to live with the other pass through functions.

labkey-jeckels:
@labkey-matthewb thanks, I knew there was an easier approach

PR #2001, opened 2021-02-16
Item 8335: LabKey product navigation header icon and menu

cnathe:
#### Rationale
As more premium clients start using our applications alongside their LKS usages, we need a way to navigate between the various products deployed to a given LK server (i.e. LKS <-> LKB, LKSM). This PR adds in some ReadOnlyApiActions to get data for the registered products on the given LK server and the current containers folder tab info. It also puts the ProductNavigation shared component onto the LKS header when applicable (i.e. for non-guest users on premium servers).

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/88
* https://github.com/LabKey/labkey-ui-components/pull/447
* https://github.com/LabKey/biologics/pull/792
* https://github.com/LabKey/sampleManagement/pull/488
* https://github.com/LabKey/inventory/pull/190
* https://github.com/LabKey/platform/pull/2001
* https://github.com/LabKey/testAutomation/pull/638

#### Changes
* Issue 41508: Update @labkey/api version to pick up fix for LABKEY.App isDOMContentLoaded property
* Update @labkey/components package version in platform modules
* Add ProductMenuProvider.getProductName and ProductRegistery.getRegisteredProducts
* Add AdminController.GetFolderTabsAction API and ProductController.GetRegisteredProductsAction API
* Add core module entryPoint for ProductNavigation.tsx to be used via loadApp in LKS header via header.jsp


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -9007,6 +9007,28 @@ public void setNewFolderType(String newFolderType)
...
+                    .collect(Collectors.toList());
+        }
+
+        private Map<String, Object> getProperties(NavTree navTree)
+        {
+            Map<String, Object> props = new HashMap<>();
+            props.put("id", navTree.getId());
+            props.put("text", navTree.getText());
+            props.put("href", navTree.getHref());
+            props.put("disabled", navTree.isDisabled());
```
labkey-susanh, line 9027: I wonder when or if we will want to get the disabled tabs.  No big deal, of course since it won't be lots of data either way but a parameter to the action that allows you to include/exclude disabled tabs might be wanted at some point.  Saves filtering them out on the client side.

core/src/org/labkey/core/view/template/bootstrap/header.jsp
```diff
@@ -171,6 +175,56 @@
 <%
     }
 
+    // only show the product navigation menu item if we are on a premium LK server or LKSM hosted only
+    if (isRealUser && c.hasPermission(user, ReadPermission.class) && (hasPremiumModule || isSMHostedOnly))
```
labkey-susanh, line 179: Is the check for read permissions on the current container what we want here?  If I land on a container that I don't have read permissions to (somehow) and then want to get out of that container, the product menu could be a way to do that.

core/src/org/labkey/core/view/template/bootstrap/header.jsp
```diff
@@ -171,6 +175,56 @@
...
+                </ul>
+            </li>
+            <script type="text/javascript">
+                +function($){
+                    // wait to load the product navigation dependencies until the hover over the header icon
+                    var productNavLoaded = false;
+                    $(document).on('mouseenter', '#headerProductDropdown .dropdown-toggle', function(e) {
+                        if (!productNavLoaded) {
+                            LABKEY.requiresScript('core/gen/productNavigation', loadProductNav);
+                            // LABKEY.requiresScript('http://localhost:3001/productNavigation.js', loadProductNav);
```
labkey-susanh, line 199: Remove?  Or leave in for hot reload testing, but add a comment?

labkey-susanh (approved):
Just a couple of comments/questions, but no need for re-review.

core/src/org/labkey/core/admin/AdminController.java
cnathe: Good point. I'll add a TODO for now.

core/src/org/labkey/core/view/template/bootstrap/header.jsp
cnathe: Correct, this line is for HMR. I'll add a comment.

core/src/org/labkey/core/view/template/bootstrap/header.jsp
cnathe: I plan to defer this for now and have added a note to my TODO list to look into fixing up this case. The current issue is that the product nav menu makes API calls to the current container, and those return "user does not have perm" error messages which we currently just show in the menu instead of gracefully handling this case.

PR #2002, opened 2021-02-16
Item 8417: Support for changing lineColor in trendline plots

labkey-ankurj:
#### Rationale
Support to add a new property in trend line graphs to provide the line color.

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/337

#### Changes
* new lineColor prop in LABKEY.vis.TrendingLinePlotType


internal/webapp/vis/src/plot.js
```diff
@@ -1630,6 +1630,7 @@ boxPlot.render();
  * @param {Object} [config.properties.mouseOverFnScope] (Optional) The scope to use for the call to mouseOverFn.
  * @param {Function} [config.properties.pointClickFn] (Optional) The function to call on data point click. The parameters to
  *                  that function will be the click event and the row of data for the selected point.
+ * @param {String} [config.properties.lineColor] (Optional) The data property name for the color to be used for the trend line connecting data points.
```
cnathe: The description here mentions that this is a "data property name for the color", is that right? I think it should just say "The color to be used...".

internal/webapp/vis/src/plot.js
```diff
@@ -2202,7 +2203,12 @@ boxPlot.render();
                     }
                     pathLayerConfig.data = seriesDataArr;
 
-                    pathLayerConfig.aes.pathColor = config.properties.groupBy;
+                    if (config.properties.lineColor) {
+                        pathLayerConfig.aes.color = config.properties.lineColor;
```
cnathe: I tested this on the Demo LabKey Plots page by making changes to the `leveyJenningsPlot` case in the /vis/demo/demo_scripts.js file, and it looks like the color I define for the `lineColor` isn't showing in the plot. For exmaple, if I change the properties to the following:
```
        showTrendLine: true,
        groupBy: 'colorlabel',
        lineColor: 'red',
        color: 'colorlabel',
        colorRange: ['red', 'blue', 'green', 'brown'],
```
I do see that the line color doesn't match the different group colors of the points, but the line is not red. It seems that the main thing here that is preventing the different line colors is just that we aren't going into the `else` block below.

cnathe (changes requested):

internal/webapp/vis/src/plot.js
labkey-ankurj: yeah I guess the color black that I was trying was coming from default color in geom.js. fixed now.

internal/webapp/vis/src/plot.js
labkey-ankurj: done

cnathe (approved):

PR #2004, opened 2021-02-17
batch select existing records

labkey-matthewb:
#### Rationale
For performance we want to be able to batch selecting existing records for detailed audit logging.
However, sampletypes depends on the behavior of SampleTypeUpdateServiceDI.getRows().  So provide a use QUS.getRows() option (not fast).


api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -154,13 +174,129 @@ public static DataIteratorBuilder createBuilder(DataIteratorBuilder dib, TableIn
...
+        @Override
+        protected void prefetchExisting() throws BatchValidationException
+        {
+            try
+            {
+                existingRecords.clear();
+                Map<String,Object> keys = CaseInsensitiveHashMap.of();
+                for (int p=0 ; p<pkColumns.size() ; p++)
+                    keys.put(pkColumns.get(p).getColumnName(), pkSuppliers.get(p).get());
+                List<Map<String, Object>> list = qus.getRows(user, c, List.of(keys));
```
labkey-susanh: Hmm.  Not sure if this is going to do the trick because this method sets addInputs to false.  We need to have that true.

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -154,13 +174,129 @@ public static DataIteratorBuilder createBuilder(DataIteratorBuilder dib, TableIn
...
+                Map<String,Object> existing = list.isEmpty() ? Map.of() : list.get(0);
+                existingRecords.put((Integer)_delegate.get(0), existing);
```
labkey-susanh, line 289: Why is it that this version can't fetch multiple rows?  Does the SQL to get the key values not work here?

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -154,13 +174,129 @@ public static DataIteratorBuilder createBuilder(DataIteratorBuilder dib, TableIn
...
+                existingRecords.put(r,Map.of());
+            list.forEach(map -> {
+                int r = (Integer)map.get("_row_number_");
+                map.remove("_row_number_");
+                map.remove("_row"); // I think CachedResultSet adds "_row"
+                existingRecords.put(r,(Map<String,Object>)map);
+            });
+            // backup to where we started so caller can iterate through them one at a time
+            ((CachingDataIterator)_delegate).reset();
+            _delegate.next();
```
labkey-susanh, line 255: That's a bit crafty :-).

labkey-susanh (changes requested):

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
labkey-matthewb: I should have added a comment.  Even though getRows() looks like a bulk API, the implementations all just loop row-by-row.

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
labkey-matthewb: Thanks

labkey-matthewb: Not part of the QUS interface of course...

labkey-susanh (commented):
This seems good to me, but there's still a failing test for SM Lineage that will need some further investigation.

labkey-susanh (approved):

PR #2013, opened 2021-02-20
Add controller alias to match @labkey/api expectations

labkey-adam:
#### Rationale
Many tests failed last night due to the new version of `@labkey/api` expecting `specimens-api` controller name instead of `specimen-api`. Add an alias to stop the carnage.

Boneheaded mistake by yours truly. Credit @labkey-matthewb for pointing out the mismatch.

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/87

### DIFF
--- a/specimen/src/org/labkey/specimen/SpecimenModule.java
+++ b/specimen/src/org/labkey/specimen/SpecimenModule.java
@@ -101,7 +101,7 @@ public class SpecimenModule extends SpringModule
 
         AttachmentService.get().registerAttachmentType(SpecimenRequestEventType.get());
 
-        addController("specimen-api", SpecimenApiController.class, "study-samples-api");
+        addController("specimen-api", SpecimenApiController.class, "study-samples-api", "specimens-api");
 
         // Register early -- some modules don't declare a runtime dependency on specimen module, but will use the
         // service if it's available


labkey-tchad (approved):

specimen/src/org/labkey/specimen/SpecimenModule.java
```diff
@@ -101,7 +101,7 @@ protected void init()
 
         AttachmentService.get().registerAttachmentType(SpecimenRequestEventType.get());
 
-        addController("specimen-api", SpecimenApiController.class, "study-samples-api");
+        addController("specimen-api", SpecimenApiController.class, "study-samples-api", "specimens-api");
```
cnathe, line 104: Should "specimen-api" still be in the list (i.e. was that a typo or do we need/want both)?

cnathe (approved):

specimen/src/org/labkey/specimen/SpecimenModule.java
labkey-adam: I meant it to be "specimen-api" everywhere, but screwed up in labkey-api-js. I'll circle back and fix it there, but this is the quickest fix. Doesn't hurt to have multiple aliases.

PR #2015, opened 2021-02-22
Issue 41202: Allow submitter to insert rows into a dataset

labkey-susanh:
#### Rationale
Currently to insert rows in a data set, a user must have full Edit permissions, but the Submitter role, which grants a user only Insert permission, is designed to allow for users that provide data but are not allowed to edit it or remove it.  Here we separate out checks for insert permission from overall edit permission to make the Submitter role effective for datasets.

#### Changes
* Check for insert permissions separately from update permissions as appropriate for dataset functionality

### DIFF
--- a/study/src/org/labkey/study/model/DatasetDefinition.java
+++ b/study/src/org/labkey/study/model/DatasetDefinition.java
@@ -921,6 +921,10 @@ public class DatasetDefinition extends AbstractStudyEntity<DatasetDefinition> im
                 {
                     result.addAll(RoleManager.getRole(SiteAdminRole.class).getPermissions());
                 }
+                else if (getStudy().getContainer().hasPermission(user, InsertPermission.class))
+                {
+                    result.add(InsertPermission.class);
+                }
                 else if (getStudy().getContainer().hasPermission(user, UpdatePermission.class))
                 {
                     // Basic write access grants insert/update/delete for datasets to everyone who has update permission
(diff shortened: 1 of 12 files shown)


labkey-adam:
While we're at it, I recommend that we similarly refactor `hasEdit()` into `hasUpdate()` and `hasDelete()`. LabKey's standard roles always take these together, but we have at least one client with a custom role that provides insert/update without delete. IMO, this would also make the code easier to understand.

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -921,6 +921,10 @@ public Container getDefinitionContainer()
                 {
                     result.addAll(RoleManager.getRole(SiteAdminRole.class).getPermissions());
                 }
+                else if (getStudy().getContainer().hasPermission(user, InsertPermission.class))
```
labkey-adam: Doesn't this mean Editors and Authors will never receive Update/Delete permissions in this case? (Because they have insert as well.) I think this logic needs to be reworked and perhaps simplified, esp. if delete is split out.

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -986,6 +994,18 @@ public boolean canWrite(UserPrincipal user)
         return getPermissions(user).contains(UpdatePermission.class);
     }
 
+    @Override
```
labkey-adam, line 1015: Factor out the common checks here... especially if we add `canDelete()`.

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -986,6 +994,18 @@ public boolean canWrite(UserPrincipal user)
...
     }
 
+    @Override
+    public boolean canInsert(UserPrincipal user)
+    {
+        if (getStudy().isDataspaceStudy())
+            return false;
+        if (user instanceof User && !canAccessPhi((User)user))
+            return false;
+        if (getContainer().hasPermission(user, AdminPermission.class))
```
labkey-adam, line 1020: I know this is pre-existing, but I'm puzzled about the separate admin check. I guess they're always exempt from dataset security rules? Comment would be nice...

PR #2018, opened 2021-02-22
add more logging to bootstrap property loading

curiositycasualty:
#### Rationale
This pull request adds `debug` and `trace` level logging to a few parts of the bootstrap properties loading process to indicate negatives cases that previously failed silently.


labkey-jeckels (approved):
Looks good!

PR #2023, opened 2021-02-23
Adjust comment message applied to dataset detailed audit log entries

Sigmonia:
#### Rationale
Update audit log message for merged dataset rows that contain no changes

#### Changes
* Adjust messaging for merges applied to dataset detailed audit log entries

### DIFF
--- a/study/src/org/labkey/study/model/DatasetDefinition.java
+++ b/study/src/org/labkey/study/model/DatasetDefinition.java
@@ -1662,7 +1662,8 @@ public class DatasetDefinition extends AbstractStudyEntity<DatasetDefinition> im
         @Override
         protected DatasetAuditProvider.DatasetAuditEvent createDetailedAuditRecord(User user, Container c, AuditConfigurable tInfo, QueryService.AuditAction action, @Nullable String userComment, @Nullable Map<String, Object> record, Map<String, Object> existingRecord)
         {
-            String auditComment = null!=userComment ? userComment :
+            boolean userCommentIsBlank = StringUtils.trimToNull(userComment) == null;
+            String auditComment = !userCommentIsBlank ? userComment :
                     switch (action)
                     {
                         case INSERT -> "A new dataset record was inserted";
@@ -1670,13 +1671,6 @@ public class DatasetDefinition extends AbstractStudyEntity<DatasetDefinition> im
                         case UPDATE, MERGE ->  null!=existingRecord && !existingRecord.isEmpty() ? "A dataset record was modified" : "A new dataset record was inserted";
                         default -> "A dataset record was modified";
                     };
-
-            DatasetAuditProvider.DatasetAuditEvent event = new DatasetAuditProvider.DatasetAuditEvent(c.getId(), auditComment);
-
-            if (c.getProject() != null)
-                event.setProjectId(c.getProject().getId());
-            event.setDatasetId(getDatasetId());
-            event.setHasDetails(true);
 
             String oldRecordString = null;
             String newRecordString = null;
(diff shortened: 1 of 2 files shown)


study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1662,7 +1662,8 @@ protected AuditTypeEvent createSummaryAuditRecord(User user, Container c, AuditC
         @Override
         protected DatasetAuditProvider.DatasetAuditEvent createDetailedAuditRecord(User user, Container c, AuditConfigurable tInfo, QueryService.AuditAction action, @Nullable String userComment, @Nullable Map<String, Object> record, Map<String, Object> existingRecord)
         {
-            String auditComment = null!=userComment ? userComment :
+            boolean userCommentIsBlank = StringUtils.trimToNull(userComment) == null;
+            String auditComment = !userCommentIsBlank ? userComment :
```
labkey-susanh: I'm not sure this is really what's wanted here.  "userComment" was introduced for sampleManagement timeline and is meant to be text captured from a user when a change is made that likely documents why the change was made. Is there UI available for adding a userComment for datasets?  Or is the thought that this will possibly come in through an API call?

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1686,16 +1680,31 @@ protected AuditTypeEvent createSummaryAuditRecord(User user, Container c, AuditC
...
                 Pair<Map<String, Object>, Map<String, Object>> rowPair = AuditHandler.getOldAndNewRecordForMerge(record, existingRecord, Collections.emptySet());
                 oldRecordString = DatasetAuditProvider.encodeForDataMap(c, rowPair.first);
                 newRecordString = DatasetAuditProvider.encodeForDataMap(c, rowPair.second);
+
+                // Check if no fields changed, if so adjust messaging
+                if (rowPair.second.size() == 0 && userCommentIsBlank)
+                {
+                    auditComment = "Dataset row was processed, but no changes detected";
+                    newRecordString = DatasetAuditProvider.encodeForDataMap(c, record);
+                }
```
labkey-susanh, line 1695: You could make an `if-else` out of this and line 1687, but I still question the test for `userCommentIsBlank`.

study/src/org/labkey/study/model/DatasetDefinition.java
Sigmonia: userComment is coming from the AbstractAuditHandler.createDetailedAuditRecord API, while I agree in this case it is always null and probably not exposed. If/when it is implemented, then I think it would take precedence.

Maybe it would be better to create an AuditRecord builder class? Then it wouldn't need to be passed in here?

study/src/org/labkey/study/model/DatasetDefinition.java
Sigmonia: fixed.

study/src/org/labkey/study/model/DatasetDefinition.java
labkey-susanh: We actually use a separate field for the UserComment because the "comment" field is meant to still record the more generic "A row was inserted" or "A row was updated" and the user comment is allow the user to say why it was done..  See InventoryDomainKind .  This userComment field was not added in every domain (so, not in AbstractAuditDomainKind), but I think it should be used similarly in other places when there is call for it.

A builder class could make sense, but I wouldn't do it in this PR.

labkey-susanh (changes requested):

labkey-susanh (approved):
Thanks for adding the javadocs.

PR #2027, opened 2021-02-24
Remove exception when AppPropsImpl encounters unknown config property

labkey-jeckels:
#### Rationale
My previous commit assumed that AppPropsImpl was the only thing reading from the SiteSettings config property scope. Other code also looks for properties there, so it can't be super-strict.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2024

#### Changes
* Remove check that rejects a SiteSettings-scoped config property that's not part of AppPropsImpl


cnathe (approved):
We can discuss later if I should move those CoreModule properties to another category/scope.

PR #2028, opened 2021-02-24
Handle InterruptedException in StatementDataIterator

labkey-matthewb:
#### Rationale
Fix error caused by unexpected InterruptedException
cf AbstractQueryUpdateService$TestCase.MERGE


labkey-jeckels (approved):
Looks promising

PR #2030, opened 2021-02-25
Issue 42222: Hide Admin Console's System Properties and Environment V…

labkey-jeckels:
…ariables from non-Site Admins

#### Rationale
These values may contain secrets that are being propagated through from the host OS or container. Thus, they should have the highest level of permissions

#### Changes
* Only allow Site Admins to see System Properties and Environment Variables


labkey-adam (approved):
Your changes look good. Pre-existing code, but I don't love the "no permission class" cases here. Ideally, we'd simply pass in the class and addLink() would infer the permissions from the action annotation. But that's for another PR...

PR #2031, opened 2021-02-25
Issue 42519: fix insert api for Submitter role

labkey-ankurj:
#### Rationale
Submitter role claims that the users may submit new information, but may not read or change anything. When a submitter (not reader) tries to insert information through Query.insertRows, the user is not successful because of ReadPermission check. This change allows the user to insert into a folder when the user is only submitter and not reader.

#### Related Pull Requests
* https://github.com/LabKey/premium/pull/230
* https://github.com/LabKey/gel/pull/4
* https://github.com/LabKey/cloud/pull/83
* https://github.com/LabKey/fileTransfer/pull/30

### DIFF
--- a/api/src/org/labkey/api/security/permissions/AbstractActionPermissionTest.java
+++ b/api/src/org/labkey/api/security/permissions/AbstractActionPermissionTest.java
@@ -162,7 +162,7 @@ public abstract class AbstractActionPermissionTest extends Assert
         return users;
     }
 
-    public void assertForReadPermission(User user, boolean skipNoPermissionCheck, PermissionCheckableAction... actions)
+    public void assertForReadPermission(User user, boolean submitterAllowed, PermissionCheckableAction... actions)
     {
         for (PermissionCheckableAction action : actions)
         {
@@ -178,7 +178,7 @@ public abstract class AbstractActionPermissionTest extends Assert
                     _users.get(APPLICATION_ADMIN_EMAIL), _users.get(SITE_ADMIN_EMAIL)
             );
 
-            if (!skipNoPermissionCheck)
+            if (!submitterAllowed)
             {
                 assertNoPermission(_c, action,
                         _users.get(SUBMITTER_EMAIL)
--- a/api/src/org/labkey/api/security/permissions/AbstractActionPermissionTest.java
+++ b/api/src/org/labkey/api/security/permissions/AbstractActionPermissionTest.java
@@ -162,6 +162,11 @@ public abstract class AbstractActionPermissionTest extends Assert
         return users;
     }
 
+    public void assertForReadPermission(User user, PermissionCheckableAction... actions)
+    {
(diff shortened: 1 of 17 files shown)


labkey-jeckels (changes requested):
I don't think we want to modify getRows() like this. Can we check in the code that's calling getRows() instead? It looks like we might need to pull hasPermission() into the QueryUpdateService interface.

Have you already tested this with a number of roles and no permissions to see that it produces the expected results?

labkey-jeckels (approved):
Looks good. With the use of @RequiresAnyOf this feels like a much safer change.

labkey-jeckels:
Please address the test failure before merging though.

labkey-ankurj:
seems like I have to modify the failing test for checking permissions

api/src/org/labkey/api/security/permissions/AbstractActionPermissionTest.java
```diff
@@ -178,7 +178,7 @@ public void assertForReadPermission(User user, boolean skipNoPermissionCheck, Pe
                     _users.get(APPLICATION_ADMIN_EMAIL), _users.get(SITE_ADMIN_EMAIL)
             );
 
-            if (!skipNoPermissionCheck)
+            if (!submitterAllowed)
```
labkey-jeckels, line 186: Can you make this verify that submitter has access, instead of just skipping the check that it has no access?

labkey-ankurj: I think https://github.com/LabKey/platform/pull/2031/files#diff-98586cebe484870a914a05d55618a046b47349818c6cf16811f81dab2cf1365cR7031 is doing that. let me know if I am misunderstanding

labkey-jeckels: Which file are you looking at? That just links to the whole PR for me.

I'm looking right here, where it does a check if submitterAllowed is false, but doesn't actually verify that the submitter has access when it submitterAllowed is true

labkey-ankurj: new addition in QueryController.java L7031

labkey-ankurj: Are you saying that I check for insertPermission for submitter?

labkey-jeckels: I think it's as simple as adding an else that calls assertPermission(_c, action, _users.get(SUBMITTER_EMAIL)), so that the call from QueryController.java L7031 actually validates the submitter permission instead of just skipping over it

labkey-jeckels (changes requested):

labkey-jeckels (approved):
Lovely

labkey-jeckels (approved):

PR #2032, opened 2021-02-25
Move specimen import to its own pipeline provider

labkey-tchad:
#### Rationale
We are unable to import anything via the pipeline if the study module is installed without the specimen module. We shouldn't add the "Import Specimen Data" pipeline action if the specimen module isn't present.

#### Changes
* Move specimen import to its own pipeline provider
* Update study tests to be less reliant on specimen module


labkey-adam (changes requested):
Minor request: update obsolete comment at StudyPipeline:59

labkey-adam (approved):

PR #2035, opened 2021-02-25
Issue 41827: Exp schema very slow (and sometimes non-responsive) to lookup to or open the schema browser

XingY:
#### Rationale
ensureFileData crawls all files under the current directory (every 5 minutes or when accessed) to make sure all files have a corresponding exp.data record. This could be a very expensive action if the number of files are large. In the case of browsing schema tree for exp, there is no need to sync files since only basic metadata of the table is needed. 

#### Related Pull Requests
* 

#### Changes
* skip ensureFileData for exp.files on expanding schema tree

### DIFF
--- a/api/src/org/labkey/api/data/AbstractTableInfo.java
+++ b/api/src/org/labkey/api/data/AbstractTableInfo.java
@@ -240,6 +240,10 @@ abstract public class AbstractTableInfo implements TableInfo, AuditConfigurable,
         }
     }
 
+    public void afterConstruct(boolean forSchemaTree)
+    {
+        afterConstruct();
+    }
 
     protected Map<String, ColumnInfo> constructColumnMap()
     {
--- a/api/src/org/labkey/api/query/AbstractSchema.java
+++ b/api/src/org/labkey/api/query/AbstractSchema.java
@@ -159,12 +159,12 @@ abstract public class AbstractSchema implements QuerySchema
         return _container;
     }
 
-    protected void afterConstruct(TableInfo info)
+    protected void afterConstruct(TableInfo info, boolean forSchemaTree)
     {
         if (info instanceof AbstractTableInfo)
         {
             AbstractTableInfo t = ((AbstractTableInfo)info);
-            t.afterConstruct();
+            t.afterConstruct(forSchemaTree);
         }
     }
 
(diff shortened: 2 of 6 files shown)


labkey-kevink (changes requested):
I like the idea of being able to construct a table for the purposes of "metadata only" in general.  I think @labkey-matthewb had some thoughts that in the past so maybe he can chime in.

However, I question the need to crawl the file system root at all when creating the table.  It seems orthogonal to ensuring there are exp.data rows which could be done in a background service.  Crawling the fileroot every 5 minutes seems excessive to me as well, but I don't know the original scenario.

XingY:
> I like the idea of being able to construct a table for the purposes of "metadata only" in general. I think @labkey-matthewb had some thoughts that in the past so maybe he can chime in.
> 
> However, I question the need to crawl the file system root at all when creating the table. It seems orthogonal to ensuring there are exp.data rows which could be done in a background service. Crawling the fileroot every 5 minutes seems excessive to me as well, but I don't know the original scenario.

For background, see original spec at https://docs.google.com/document/d/1DvpuDaAtDMCz00Tk1cAVBNmLRhHsz4ShMatQ_9yscrI/edit

labkey-jeckels:
Two ideas to reduce the scope of the change but perhaps still meet the original requirements:

1. Make the file system scan run in a background thread. Since we're not guaranteeing it's fully consistent (due to the 5 minute timer) this might be a trivial change.
2. Keep the scan synchronous, but move it from the constructor to when we generate SQL, which would prevent it from running when we're creating the TableInfo for metadata purposes. This could be done by overriding getFromSQL(), kicking off the scan, and then calling the superclass method.

labkey-matthewb:
To answer Kevin's question.  We do have a lightweight object called TableDescription, it would be great to use this in more places that just need to know that a table exists and its name, but not ColumnInfo.  It could be very useful for UserSchema to implement getTableDescriptions() or something like that.  But that's too much for 21.3.

To Josh's suggestion.  Yes!  I've done a ton of work to push work out of TableInfo constructors into
TableInfo.getFromSQL() or ColumnInfo.getValueSql().  There is a lot of work that can be delayed until the TableInfo is actually used get fetch data.  I always support this!

_getTableOrQuery is already so complicated that (especially now with caching), I'm wary of having yet another kind of TableInfo construction option.

XingY:
new PR https://github.com/LabKey/platform/pull/2043

PR #2036, opened 2021-02-25
Issue40946: Updates to secure message board options

RosalineP:
#### Rationale
Allow for a third secure message board option that allows for secure message boards to send contents of emails.
Migrate previous `secure = false` option to 'secureOff', previous `secure = true` option to 'secure without emails', and retain default message board setting as 'secureOff'. 
A message board that is 'secure without emails' will notify that a message board update has occurred without including body content, while 'secure with emails' will notify and include body content.

#### Related Pull Requests
* n/a

#### Changes
* Toggle boolean 'secure' property to corresponding default string using upgrade scripts
* Bump announcement version number
* Update wording and options on announcements-customize page


RosalineP:
Notes from sync with Josh: 
Due to the simplicity of the new string secure setting, I opted to not make it a enum, and just clarify using the string values.

Existing integration tests check whether or not various roles can various actions such as view, update, delete. Since there's no test that currently toggles and verifies the Security setting—and no tests checking email content—I think auto tests on this should be deferred.

Either PR would be appreciated depending on who gets the bandwidth, thank you! 👍

RosalineP:
Looking at test outages that popped up with the upgrade script push....

internal/src/org/labkey/api/announcements/DiscussionService.java
```diff
@@ -214,16 +214,32 @@ public void setConversationName(String itemName)
...
         {
             _secure = secure;
         }
 
+        public boolean isSecureOff() {
+            return ("secureOff").equals(_secure);
+        }
+
+        public boolean isSecureOn() {
+            return ("secureWithoutEmail").equals(_secure) || ("secureWithEmail").equals(_secure);
```
labkey-jeckels: Since these strings are reused can they be constants and also referenced from the JSP?

announcements/resources/schemas/dbscripts/sqlserver/comm-21.000-21.001.sql
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+UPDATE prop.properties
+SET value = 'secureOff'
+WHERE name = 'secure' AND value = 'false' AND
+    set IN (SELECT set FROM prop.propertysets WHERE category = 'messageBoardSettings');
```
labkey-jeckels: Haven't looked at the test failures, but given your comment I'm guessing you need to double-quote "SET" as it's a reserved word and I remember one of the DBs being picky

```suggestion
    "set" IN (SELECT "set" FROM prop.propertysets WHERE category = 'messageBoardSettings');
```

labkey-jeckels (changes requested):

announcements/resources/schemas/dbscripts/sqlserver/comm-21.000-21.001.sql
RosalineP: Though the outages are only on sqlserver, tested this out and added quotes to both upgrade scripts for consistency 👍

internal/src/org/labkey/api/announcements/DiscussionService.java
RosalineP: Done! 👍

labkey-jeckels (changes requested):
As discussed separately, it's legal syntax in the full UPDATE statement but it's not working the way you want
SELECT 'set' FROM prop.propertysets WHERE category = 'messageBoardSettings'
will return the value 'set' as a string, not the value of the "set" column

labkey-jeckels (approved):

PR #2037, opened 2021-02-26
Item 8583: Ontology concept picker integration for field editor concept annotation feature

cnathe:
#### Rationale
With the Field Editor changes in the related ui-components PR, we are ready to use the OntologyBrowserPanel to allow a user to configure a field's Concept Annotation. This PR pulls in the @labkey/components package updates to put those Field Editor changes in place for the various designers in the platform modules. In addition, we then use that field concept annotation to show additional info in the Data Region column header on hover.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/452
* https://github.com/LabKey/platform/pull/2037
* https://github.com/LabKey/ontology/pull/22
* https://github.com/LabKey/sampleManagement/pull/496

#### Changes
* Update platform module @labkey/components package version to pick up ontology related Field Editor changes
* Remove usages of OntologyBrowserPanel from core-components page now that it is integration with ontology module pages
* Data Region update to show concept annotation label/code on column header hover


api/src/org/labkey/api/data/DisplayColumn.java
```diff
@@ -681,17 +683,33 @@ public void renderGridHeaderCell(RenderContext ctx, Writer out, String headerCla
         {
             tooltip.append(getDescription());
         }
-        if (null != getColumnInfo() && !getColumnInfo().getFieldKey().toString().equals(getColumnInfo().getLabel()))
+        if (null != getColumnInfo())
```
Sigmonia, line 687:
```suggestion
        
        if (null != getColumnInfo())
```

api/src/org/labkey/api/data/DisplayColumn.java
```diff
@@ -681,17 +683,33 @@ public void renderGridHeaderCell(RenderContext ctx, Writer out, String headerCla
...
             }
-            tooltip.append(getColumnInfo().getFieldKey().toString());
-            if (suffix)
+
+            if (null != getColumnInfo().getPrincipalConceptCode())
             {
-                tooltip.append(")");
+                String conceptDisplay = getColumnInfo().getPrincipalConceptCode();
+                var ontologyService = OntologyService.get();
+                if (null != ontologyService)
```
Sigmonia, line 707: Is there a way for the getPrincipalConceptCode to be set, but the OntologyService is not registered? If not should this check be moved up?

Sigmonia (approved):

api/src/org/labkey/api/data/DisplayColumn.java
cnathe: done

api/src/org/labkey/api/data/DisplayColumn.java
cnathe: Since `principalConceptCode` is a prop on the ColumnType of  tableInfo.xsd, I think it is possible that this property could be set in XML without the OntologyService being there. Very unlikely I would guess.

PR #2042, opened 2021-02-26
Item 8516: Create aliquot using API

XingY:
#### Rationale
Aliquots are distinguished from the existing "derivatives". An aliquot or subaliquot is generated from a single parent sample or aliquot of the same sample type, and can have its own additional properties, in addition to the inherited sample's properties. This PR adds some preliminary work to support aliquots for LKSM. 

#### Related Pull Requests
* 

#### Changes
* new "Sample Aliquot" experimental flag
* add new PropertyDescriptor "DerivationDataScope" to support tagging Aliquot specific” fields
* add RootMaterialLSID and AliquotedFromLSID to exp.material
  * RootMaterialLSID is used to fetch sample metadata for aliquots and subaliquots
  * AliquotedFromLSID stores the aliquots direct parent for convenience
* fetch aliquot field values using a self join on provisioned material table in ExpMaterialImpl.getFromSql 
* update DerivationDataIterator to also handle aliquot creation, using a new "Sample Aliquot Protocol"
** customize samples/aliquot insert and update with new UpdateMaterialQueryRowAction and InsertMaterialQueryRowAction


experiment/resources/schemas/dbscripts/sqlserver/exp-20.012-20.013.sql
```diff
@@ -29,3 +29,4 @@ ALTER TABLE exp.material ADD CONSTRAINT FK_Material_ObjectId
     FOREIGN KEY (objectid) REFERENCES exp.object (objectid);
 ALTER TABLE exp.experimentrun ADD CONSTRAINT FK_ExperimentRun_ObjectId
     FOREIGN KEY (objectid) REFERENCES exp.object (objectid);
+GO
```
labkey-susanh, line 32: Is this necessary if it's the last line in the file?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -422,28 +441,41 @@ public DataIterator getDataIterator(DataIteratorContext context)
         final Container _container;
         final User _user;
         final boolean _isSample;
+        final boolean _skipAliquot; // skip aliquot validation, used for update/updates cases
 
-        protected DerivationDataIterator(DataIterator di, DataIteratorContext context, Container container, User user, boolean isSample)
+        final List<String> _aliquotLsids; // check if lsid is aliquot with absent "AliquotedFrom", for merge
```
labkey-susanh: This comment is confusing to me.  I think it's partially a naming issue (see below), but maybe the comment could be updated to something like "used to check if ...". or "LSIDs present without AliquotFrom field.  Used for validation."

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -422,28 +441,41 @@ public DataIterator getDataIterator(DataIteratorContext context)
...
                 {
                     _parentCols.put(entry.getValue(), entry.getKey());
                 }
+                else if (_isSample && "AliquotedFrom".equalsIgnoreCase(name))
+                {
+                    aliquotParentCol = entry.getValue();
+                }
             }
+
+            _aliquotParentCol = aliquotParentCol;
```
labkey-susanh, line 479: Looks like the local `aliquotParentCol` variable is not necessary.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -461,9 +493,37 @@ public boolean next() throws BatchValidationException
...
+                            getErrors().addRowError(new ValidationException("Expected string value for aliquot parent name: " + o, "AliquotedFrom"));
+                        }
+
+                        if (aliquotParentName != null)
+                            _aliquotParents.put(lsid, aliquotParentName);
+                    }
+                }
+                else if (!_skipAliquot)
+                {
+                    _aliquotLsids.add(lsid);
```
labkey-susanh: Am I reading this correctly that if there isn't an `AliquotedFrom` column then the lsid is added to the `aliquotLsids`?  That seems the opposite of what I would expect, but I think it's just a naming thing.  `aliquotLsids` are really samlpeLsids that will be checked to determine if they are aliquots or not, right?  Maybe just call this `sampleLsids`?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -404,15 +414,24 @@ public DataIterator getDataIterator(DataIteratorContext context)
             {
                 return pre;
             }
-            return LoggingDataIterator.wrap(new DerivationDataIterator(pre, context, _container, _user, _isSample));
+            return LoggingDataIterator.wrap(new DerivationDataIterator(pre, context, _container, _user, _isSample, _skipAliquot));
         }
     }
 
+    static List<ExpMaterialImpl> getAliquots(List<String> lsids)
```
labkey-susanh: Looks like this is only ever used to check for absence or presence or aliquots.  Can you use the `exists()` method on `TableSelector` instead?  It should be more efficient.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -524,13 +584,31 @@ else if (o instanceof Number)
...
+                    if (_isSample && _context.getInsertOption().mergeRows && _aliquotParentCol == -1)
+                    {
+                        if (!_aliquotLsids.isEmpty())
+                        {
+                            if (!getAliquots(_aliquotLsids).isEmpty())
+                            {
+                                // AliquotedFrom is used to determine if aliquot/meta field value should be retained or discarded
+                                // In the case of merge, one can argue AliquotedFrom can be queried for existing data, instead of making it a required field.
+                                // But that would be too expensive. For performance reasons, merge will error out if any aliquots are present but 'AliquotedFrom' column is missing.
+                                throw new ValidationException("Aliquots are present but 'AliquotedFrom' column is missing.");
```
labkey-susanh: This behavior is a little different than for other fields where the absence of a column means we don't want to change the current value in the database.  Does it need it be?

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -643,11 +665,37 @@ public SQLFragment getFromSQL(String alias)
...
-            sql.append(comma).append("ss.*");
+        {
+            for (ColumnInfo propertyColumn : provisioned.getColumns())
+            {
+                // don't select lsid twice
+                if ("lsid".equalsIgnoreCase(propertyColumn.getColumnName()))
+                    continue;
+
+                sql.append(comma);
+                if ("ChildOnly".equalsIgnoreCase(propertyColumn.getDerivationDataScope()))
```
labkey-susanh: Maybe add constants for "ChildOnly" and its partner?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -239,20 +269,53 @@ public int mergeRows(User user, Container container, DataIteratorBuilder rows, B
...
+            {
+                validRowCopy.put(updateField, updateValue);
+            }
+            else if (isAliquot && isSampleMetaField)
+            {
+                LOG.warn("Sample metadata update has been skipped for an aliquot");
+            }
+            else if (!isAliquot && isAliquotField)
+            {
+                LOG.warn("Aliquot specific field update has been skipped for a sample.");
```
labkey-susanh:
```suggestion
                LOG.warn("Aliquot-specific field update has been skipped for a sample.");
```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -5903,72 +5903,94 @@ public ExpRun deriveSamples(Map<ExpMaterial, String> inputMaterials, Map<ExpMate
...
+    {
+        helper.addRunParams(run._object, user.getUserId());
+
+        // protocol applications
+        ExpProtocolImpl parentProtocol = run.getProtocol();
+
+        List<ExpProtocolActionImpl> actions = parentProtocol.getSteps();
+        if (actions.size() != 3)
+        {
+            throw new IllegalArgumentException("Protocol has the wrong number of steps for a simple protocol, it should have three");
```
```suggestion
            throw new IllegalArgumentException("Protocol has the wrong number of steps for a simple protocol; it should have three.");
```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -5903,72 +5903,94 @@ public ExpRun deriveSamples(Map<ExpMaterial, String> inputMaterials, Map<ExpMate
...
-                helper.addProtocolApp(protApp3, date, action3, outputProtocol, run, runRecord);
-
-                if ((count % MAX_RUNS_IN_BATCH) == 0)
+                if (runRecord.getAliquotInput() != null)
                 {
-                    helper.saveBatch();
+                    countA++;
+                    ExpRunImpl run = createAliquotRun(runRecord.getAliquotInput(), runRecord.getAliquotOutputs(), info);
+                    _prepareRun(aliquotHelper, run, runRecord, user, date);
+                    if ((countA % MAX_RUNS_IN_BATCH) == 0)
```
labkey-susanh, line 5974: I'm unclear on the needs for batches since we're sort of moving away from using them.  It seems right that we would have separate batches for aliquots than for derivatives.  Thoughts on this @labkey-kevink?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -6157,6 +6221,54 @@ else if (rec._action.getActionSequence() == SIMPLE_PROTOCOL_OUTPUT_STEP_SEQUENCE
...
+        {
+            // get the protocol application rows id's
+            Map<String, Integer> protAppRowMap = new HashMap<>();
+
+            for (ProtocolAppRecord rec : protAppRecords)
+                protAppRowMap.put(rec._protApp.getLSID(), null);
+
+            TableInfo pa = getTinfoProtocolApplication();
+            SQLFragment sqlfilter = new SimpleFilter(FieldKey.fromParts("LSID"), protAppRowMap.keySet(), IN).getSQLFragment(pa, "pa");
+            new SqlSelector(pa.getSchema(), new SQLFragment("SELECT Lsid, RowId FROM " + pa /* + (pa.getSqlDialect().isSqlServer() ? " WITH (UPDLOCK, HOLDLOCK)" : "") */ + " ")
```
labkey-susanh: Is the `WITH (UPDLOCK ...` meant to be commented out still?

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -85,6 +86,9 @@
     private static final Logger _log = LogManager.getLogger(UploadSamplesHelper.class);
     private static final String MATERIAL_LSID_SUFFIX = "ToBeReplaced";
 
+    private static final String INVALID_ALIQUOT_PROPERTY = "An aliquot specific property [%1$s] value has been ignored for a non-aliquot sample.";
```
labkey-susanh:
```suggestion
    private static final String INVALID_ALIQUOT_PROPERTY = "An aliquot-specific property [%1$s] value has been ignored for a non-aliquot sample.";

```

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -316,6 +359,24 @@ public UploadSampleRunRecord(Map<ExpMaterial, String> inputMaterial, Map<ExpMate
...
+        {
+            ExpSampleType sampleType = sampleTypes.computeIfAbsent(dataType, (name) -> SampleTypeService.get().getSampleType(c, user, name));
+            if (sampleType == null)
+                throw new ValidationException("Invalid sample type: " + dataType);
+
+            aliquotParent = findMaterial(c, user, sampleType, dataType, aliquotedFrom, cache, materialMap);
+
+            if (aliquotParent == null)
+            {
+                String message = "Aliquot parent '" + aliquotedFrom + "' not found";
```
labkey-susanh:
```suggestion
                String message = "Aliquot parent '" + aliquotedFrom + "' not found.";

```

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -329,6 +390,12 @@ public UploadSampleRunRecord(Map<ExpMaterial, String> inputMaterial, Map<ExpMate
                 {
                     if (!isEmptyParent)
                     {
+                        if (isAliquot)
+                        {
+                            String message = "Sample derivation parent input not allowed for aliquot";
```
labkey-susanh:
```suggestion
                            String message = "Sample derivation parent input is not allowed for aliquots.";

```

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -351,11 +418,17 @@ public UploadSampleRunRecord(Map<ExpMaterial, String> inputMaterial, Map<ExpMate
...
                     {
-                        if (isMerge)
+                        if (isMerge && !isAliquot)
                             parentSampleTypesToRemove.add(namePart);
                     }
                     else
                     {
+                        if (isAliquot)
+                        {
+                            String message = "Sample derivation parent input not allowed for aliquot";
```
labkey-susanh:
```suggestion
                            String message = "Sample derivation parent input is not allowed for aliquots.";

```

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -374,7 +447,15 @@ else if (parts[0].equalsIgnoreCase(ExpMaterial.MATERIAL_OUTPUT_CHILD))
                     {
                         ExpMaterial sample = findMaterial(c, user, sampleType, namePart, parentValue, cache, materialMap);
                         if (sample != null)
-                            childMaterials.put(sample, sampleRole(sample));
+                        {
+                            if (StringUtils.isEmpty(sample.getAliquotedFromLSID()))
+                                childMaterials.put(sample, sampleRole(sample));
+                            else
+                            {
+                                String message = "Sample derivation output not allowed for aliquot";
```
labkey-susanh:
```suggestion
                                String message = "Sample derivation output is not allowed for aliquots.";

```

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -441,25 +522,56 @@ else if (parts[0].equalsIgnoreCase(ExpData.DATA_OUTPUT_CHILD))
...
+                    isExistingAliquot = !StringUtils.isEmpty(currentMaterial.getAliquotedFromLSID());
+
+                    if (isExistingAliquot && !isAliquot)
+                        throw new ValidationException("AliquotedFrom is absent for aliquot: " + currentMaterial.getName());
+                    else if (!isExistingAliquot && isAliquot)
+                        throw new ValidationException("Unable to change sample to aliquot: " + currentMaterial.getName());
+                    else if (isExistingAliquot)
+                    {
+                        if (!currentMaterial.getAliquotedFromLSID().equals(aliquotParent.getLSID()))
+                            throw new ValidationException("Aliquot parent cannot be updated");
```
labkey-susanh:
```suggestion
                            throw new ValidationException(Strring.format("Aliquot parents cannot be updated.  Change parent value from %s to %s.",currentMaterial.getAliquotedFromLSID(), aliquotParent.getLSID()));
```
This should help a user find the offending row in the input.

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -441,25 +522,56 @@ else if (parts[0].equalsIgnoreCase(ExpData.DATA_OUTPUT_CHILD))
...
-                    ExpSampleType sampleType = materialParent.getSampleType();
-                    String role = sampleRole(materialParent);
-                    if (sampleType != null && !parentMaterials.containsValue(role) && !parentSampleTypesToRemove.contains(role))
-                        existingParentMaterials.put(materialParent, role);
-                });
-                parentMaterials.putAll(existingParentMaterials);
+                if (isExistingAliquot && currentParents.second.size() != 1)
+                {
+                    if (currentParents.second.size() == 0)
+                        throw new ValidationException("Unable to find aliquot parent for " + runItem.getName());
```
labkey-susanh:
```suggestion
                        throw new ValidationException("Unable to find aliquot parent for " + runItem.getName() + ".");

```

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -441,25 +522,56 @@ else if (parts[0].equalsIgnoreCase(ExpData.DATA_OUTPUT_CHILD))
...
+                    else
+                        throw new ValidationException("Invalid parents for aliquot: " + runItem.getName());
```
labkey-susanh:
```suggestion
                        throw new ValidationException("Invalid parents for aliquot " + runItem.getName() + ".");

```

labkey-susanh (changes requested):

experiment/resources/schemas/dbscripts/sqlserver/exp-20.012-20.013.sql
XingY: I assume it's helpful in the roll up script. But I added it here because @mbellew added a 'GO' to exp-20.010-20.011

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: The problem is _aliquotParentCol is a final fields that cannot be updated. So a temp var is used to process intermediate value.

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: It doesn't needs to be. But for reasons listed in comment, it might not be so horrible to make it required. Also it would require a big change to StatementUtil to prepare a customized update/insert statement for samples. Maybe something to consider for future when we got compliants?

experiment/resources/schemas/dbscripts/sqlserver/exp-20.012-20.013.sql
labkey-adam: Fine to include it, thanks @XingY . When we eventually roll up scripts, we do test and add GO statements if required. So not critical either way.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Got it.  Thanks.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Isn't the hasAliquots method already querying for the LSID?  Granted, it's now doing an exists query instead of retrieving the materials so will be more efficient.  

Sorry if I'm being dense here, but why would the customized insert/update statement be needed?  Do you mean in the case that we don't have the AliquotedFrom value so we would somehow need to know which fields to update and which to discard?  Yes, that would be ugly.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Still need this commented out part of the SQLFragment?

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: Yes, the check is for differentiating meta vs aliquots fields. Insert/update into the material & properties table are already done by this point. And for merge without 'aliquotedFrom' column, the update might have wrongly updated meta/aliquot fields. This check is meant as a workaround after the update to fail the transaction. It does check for hasAliquots, but like I said, this is way past the actual update/insert. 

To properly update aliquot/meta fields, a custom StatementUtil will be needed at an earlier step. But if it's purely sql based, log warning for bad records wouldn't be possible.

Alternatively, we can add a FixDerivationDataFieldDataIterator, which identifies (find all merged sample with any aliquot-specific field populated, and aliquots with meta field populated) and fix(re-update) aliquot vs meta fields for merge without 'aliquotedFrom' case.

labkey-susanh: Got it.  Thanks for the clarification.  Definitely a different kind of merge we're doing here.  We may want to revisit this at some point if this special treatment from the user becomes problematic, but I think we're good for now.

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: This was copied from Matt's code, removed.

PR #2045, opened 2021-03-01
Issue 42496: Lingering General/GPAT terminology in UI after rename

labkey-jeckels:
#### Rationale
We've renamed General assay to be Standard assay, and there are lingering references to the old name in the UI.

Since we want this for 21.3 I'll retarget before merging but this will get some test coverage underway

#### Changes
* Introduce label for assay providers
* Use the label in a handful of UI text generating places
* Update some comments and metadata


labkey-martyp (approved):
Looks good. Thanks for fixing these

PR #2047, opened 2021-03-01
Refactor publish start and confirm actions

labkey-klum:
#### Rationale
In order to integrate Sample Types into Studies we are planning to support the copy to study workflow for Sample Types. This ability currently exists for Assays but the workflow needs to be refactored to allow future usage with Sample Types.

The goal of this feature branch is to tease apart the assay specific and non-assay specific code into a new framework and to get this code working with assays. We plan to implement the sample type specific code in a future feature branch.

#### Changes
Two main actions handle the copy to study workflow:
- PublishStartAction :  takes the initial row selections and displays the dropdown of the valid publish targets
- PublishConfirmAction : performs validation on the row selections and displays the confirmation and validation grid

The approach taken is to move the common code into abstract base actions and extend the actions with the (in this case) assay specific actions. The code which handles the actual publishing of the data is unchanged but is planned to be refactored in a later feature branch.


api/src/org/labkey/api/study/query/PublishResultsQueryView.java
```diff
@@ -273,7 +248,7 @@ protected DataRegion createDataRegion()
     protected Set<String> getHiddenColumnCaptions()
     {
         HashSet<String> hidden = new HashSet<>(Collections.singleton("Assay Match"));
-        if (_targetStudyDomainProperty != null && _targetStudyDomainProperty.first != ExpProtocol.AssayDomainTypes.Result)
+        //if (_targetStudyDomainProperty != null && _targetStudyDomainProperty.first != ExpProtocol.AssayDomainTypes.Result)
```
RosalineP, line 251: Note: May be redundant logic, may be touched in later story in this epic. I don't think we had an actionable on this other than me keeping it in mind for downstream dev, or perhaps getting an optional explanatory comment getting added here.

study/src/org/labkey/study/assay/AssayPublishStartAction.java
```diff
@@ -0,0 +1,248 @@
...
+
+@RequiresPermission(InsertPermission.class)
+public class AssayPublishStartAction extends AbstractPublishStartAction<AssayPublishStartAction.AssayPublishForm>
+{
+    private ExpProtocol _protocol;
+    private AssayProvider _provider;
+    private List<Integer> _ids = new ArrayList<>();
+    private List<Integer> _runIds = new ArrayList<>();
+
+    public static class AssayPublishForm extends ProtocolIdForm implements PublishStartForm
```
RosalineP: Rename 'AssayPublishForm' => 'AssayPublishStartForm'?

The intention would be to align naming so that `AssayPublishStartForm [...] implements PublishStartForm` is parallel with `AssayPublishConfirmForm extends PublishConfirmForm`

(Just as `AssayPublishStartAction extends AbstractPublishStartAction` is already parallel with `AssayPublishConfirmAction extends AbstractPublishConfirmAction`)

api/src/org/labkey/api/study/query/PublishResultsQueryView.java
```diff
@@ -229,7 +201,10 @@ public DataView createDataView()
     protected DataRegion createDataRegion()
     {
         DataRegion dr = new DataRegion();
-        initializeDataRegion(dr);
+
+        //initializeDataRegion(dr);
```
RosalineP: Flagged for deletion. `configureDataRegion` below does same thing, in non-assay-tied way.

study/src/org/labkey/study/publish/AbstractPublishConfirmAction.java
```diff
@@ -1,592 +1,427 @@
...
+                mismatched,
+                form.isIncludeTimestamp(),
+                getAdditionalColumns(form),
+                getHiddenFormFields(form));
+
+        List<ActionButton> buttons = new ArrayList<>();
+        URLHelper returnURL = form.getReturnURLHelper();
+        if (null == returnURL)
+        {
+//            returnURL = PageFlowUtil.urlProvider(AssayUrls.class).getAssayRunsURL(getContainer(), _protocol).addParameter("clearDataRegionSelectionKey", publishConfirmForm.getDataRegionSelectionKey());
```
RosalineP, line 220: Flagged--I believe for this one, because it's assay-specific these lines need to be pulled out, but it remains to be seen whether this will be a useful bit of logic that will have to be touched in the SampleType implementation story?
Explanatory comment to be added by Karl.

RosalineP (approved):

RosalineP:
Also, I believe PR titles should be of form '[Scrumwise Item Number]: [Title]' 👍

RosalineP:
And thanks again for insight! :)

PR #2049, opened 2021-03-01
Issue 42242: Skip suggested columns in ETL

labkey-martyp:
#### Rationale
I didn't get a complete fix on this the first time.  Add getTable option to UserSchema that skips suggested columns.  Called from data integration.

#### Related Pull Requests
* https://github.com/LabKey/dataintegration/pull/92

#### Changes
* Add getTable option to UserSchema which skips suggested columns


labkey-matthewb:
I don't really love the collection of flags we're collecting on getTableI(), but this seems pragmatic for now.  I'd rather see us add use   the ColumnInfos (e.g. use isReadOnly(), or add a new property if necessary isQuerySuggested()).

labkey-matthewb (approved):

PR #2053, opened 2021-03-03
8540: Support Tagging Participant and Time Point Columns in Sample Types

RosalineP:
#### Rationale
Bump @labkey/components version, which will allow for ‘VisitDate’ and ‘VisitID’ Data Types on domain fields for Sample Types. This will support the larger integration of Sample Types with Study.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/460
* https://github.com/LabKey/biologics/pull/811
* https://github.com/LabKey/sampleManagement/pull/502

#### Changes
* Bump @labkey/components version

### DIFF
--- a/api/gwtsrc/org/labkey/api/gwt/client/model/GWTDomain.java
+++ b/api/gwtsrc/org/labkey/api/gwt/client/model/GWTDomain.java
@@ -43,6 +43,7 @@ public class GWTDomain<FieldType extends GWTPropertyDescriptor> implements IsSer
     private boolean allowFileLinkProperties;
     private boolean allowAttachmentProperties;
     private boolean allowFlagProperties;
+    private boolean allowTimepointProperties;
     private boolean showDefaultValueSettings;
     private DefaultValueType defaultDefaultValueType = null;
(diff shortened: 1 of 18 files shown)


RosalineP:
Note that the version bump in List is purely so it's possible to verify that the two new dropdown options don't exist in a non-sampleType domain, with the new package. 

@cnathe Is it good style to do a version bump in all places the Designers surface (the set of package.jsons touched in the previous Field Editor PRs)?

cnathe (approved):

cnathe:
> Note that the version bump in List is purely so it's possible to verify that the two new dropdown options don't exist in a non-sampleType domain, with the new package.
> 
> @cnathe Is it good style to do a version bump in all places the Designers surface (the set of package.jsons touched in the previous Field Editor PRs)?

Yes, in all most all cases for develop when we bump the components package version in platform, we will want to bump it for all 7 of the modules within platform that use it. The main case where we often will bump only specific platform modules is for a patch fix to a release branch where the fix is only needed (or only applies) in a subset of modules.

labkey-klum (approved):

PR #2054, opened 2021-03-03
Issue 42519: Allow submitter to insert into custom issue fields

labkey-ankurj:
#### Rationale
Previous work fixed the general insert apis to allow submitter. This fixes the actual issue of allowing submitter not able to insert into custom issue definition fields for which the prev work started.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2031

#### Changes
* update check on IssuePage


labkey-jeckels (approved):
Please add a comment that this approach is designed to support the submitter role.

PR #2057, opened 2021-03-03
Issue 41202: Allow submitter to insert rows into a dataset

labkey-adam:
#### Rationale
In editable dataset modes, dataset editing currently requires Admin or Editor role. This means Authors and Submitters, who can insert into other tables, are not allowed to insert into datasets. Here we replace a single `canWrite()` method, that checks for update, with separate `canInsert()`, `canUpdate()`, and `canDelete()` methods. We also fix a variety of associated inconsistencies (dataset button bars not respecting dataset security, incorrect action permissions) and unnecessarily complex permission checks. All code paths should now delegate to `DatasetDefinition.hasPermission()`.

This also updates the per-dataset drop-downs to remove unnecessary roles ("No Permissions", "Restricted Reader"), add "Author", switch to standard "Reader" and "Editor" terminology, and add a sensible sort.

#### Dataset Security Rules
For the record, these are the rules we now follow:

Regardless of security type:

- Admins in the current folder can always import, insert, update, and delete rows in every dataset
- All dataset access requires folder read permission, at a minimum

Additional rules for each security type:

- Basic security with read-only datasets
  - Users with folder read permissions can read all datasets. Editors, Authors, etc. have no ability to edit datasets (edit is restricted to Admins)
- Basic security with editable datasets
  - Dataset permissions respect folder permissions. Editors can insert, update, delete, and read; Authors can insert and read; etc. (previously, only update was checked and implied insert/delete)
- Custom security with read-only datasets
  - Users assigned to groups designated as "READ ALL" can read all datasets
  - Users assigned to groups granted "PER DATASET" access can read only the datasets marked as "Read" for that group
- Custom security with editable datasets
  - Read permissions are handled as above (Custom security with read-only datasets)
  - Users assigned to groups designated as "EDIT ALL" are granted full Editor rights in all datasets. Under this type, users' permissions may be "escalated" beyond their folder permissions; for example, a user with only Reader role will become an Editor in all datasets.
  - Users assigned to groups granted "PER DATASET" access are granted full Editor rights in the datasets marked as "Edit" for that group. Again, users' permissions may be escalated beyond their folder permissions.

### DIFF
--- a/study/src/org/labkey/study/model/DatasetDefinition.java
+++ b/study/src/org/labkey/study/model/DatasetDefinition.java
@@ -86,6 +86,7 @@ import org.labkey.api.query.SimpleValidationError;
 import org.labkey.api.query.ValidationException;
 import org.labkey.api.reports.model.ViewCategory;
 import org.labkey.api.reports.model.ViewCategoryManager;
+import org.labkey.api.security.HasPermission;
 import org.labkey.api.security.MutableSecurityPolicy;
 import org.labkey.api.security.SecurityPolicy;
 import org.labkey.api.security.SecurityPolicyManager;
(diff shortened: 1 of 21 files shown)


labkey-susanh (approved):
Nice resolution of some of the complications.

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -917,40 +918,19 @@ public Container getDefinitionContainer()
...
-                    }
-                    // A user can be part of multiple groups, which are set to both Edit All and Per Dataset permissions
-                    // so check for a custom security policy even if they have UpdatePermission on the study's policy
-                    if (studyPolicy.hasPermission(user, ReadSomePermission.class))
-                    {
-                        // Advanced write grants dataset permissions based on the policy stored directly on the dataset
-                        result.addAll(SecurityPolicyManager.getPolicy(this).getPermissions(user));
-                    }
+                    // Advanced write grants dataset permissions based on the policy stored directly on the dataset
+                    result.addAll(SecurityPolicyManager.getPolicy(this).getPermissions(user));
```
labkey-matthewb, line 938: why not use copyEditPerms() here?

labkey-matthewb (approved):

study/src/org/labkey/study/model/DatasetDefinition.java
labkey-adam: No reason... just missed it.

PR #2061, opened 2021-03-03
Issue 40895: Renaming a query breaks editing metadata via UI

labkey-ankurj:
#### Rationale
Updating query metadata after renaming the query was causing two table elements to be written in metadata xml - one for the old name and one for the new name, which was breaking the query metadata editor UI because of the improper xml.

#### Changes
* Update the query xml metadata on renaming

### DIFF
--- a/query/src/org/labkey/query/controllers/QueryController.java
+++ b/query/src/org/labkey/query/controllers/QueryController.java
@@ -2302,6 +2302,8 @@ public class QueryController extends SpringActionController
                     return false;
                 }
 
+                // Issue 40895: update queryName in xml metadata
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
Looks OK, but would be nice to avoid duplication of XML serialization with MetadataTableJSON. Maybe a new method on QueryDefinition that takes a TablesDocument and does the XmlOption stuff?

labkey-ankurj:
Can I keep this as a follow up item for future?

labkey-jeckels:
Yes, that's OK. It's not much code.

PR #2064, opened 2021-03-04
Issue 42386: Don't show reserved fields as inferred fields

labkey-susanh:
#### Rationale
When inferring fields from a file for a domain, we separate the response fields into two sets: the reserved fields and the others.  This allows the client to handle these sets differently if desired.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/463

#### Changes
* Add optional `domainKindName` property to form for `property-inferFields`.
* Update @labkey/components in various packages that use the `DomainForm`


experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -844,6 +849,8 @@ private ApiSimpleResponse getInferDomainResponse(DataLoader loader, Integer numL
             }
 
             response.put("fields", fields);
+            if (!reservedFields.isEmpty())
+                response.put("reservedFields", reservedFields);
```
cnathe: probably could just remove the if statement and always send back the reservedFields, even if empty. That way the client code can always expect it.

list/src/org/labkey/list/model/ListDomainKind.java
```diff
@@ -231,7 +232,7 @@ public PropertyStorageSpec getPropertySpec(PropertyDescriptor pd, Domain domain)
     @Override
     public Set<String> getReservedPropertyNames(Domain domain)
     {
-        Set<String> properties = new LinkedHashSet<>();
+        Set<String> properties = new CaseInsensitiveHashSet();
```
cnathe, line 235: Did the other domain kinds already use CaseInsensitiveHashSet?

cnathe (approved):

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
labkey-susanh: OK.  Wasn't sure which I preferred.

list/src/org/labkey/list/model/ListDomainKind.java
labkey-susanh: Yes.

PR #2065, opened 2021-03-04
Issue 39822: Visit calculations are not as expected in the published child Date study

labkey-martyp:
#### Rationale
Need to recalculate participant visits after updating top level study info as they are dependent on study start date.

#### Changes
* Add updateParticipantVisits call to TopLevelStudyPropertiesImporter.process


labkey-jeckels (changes requested):
Your change looks OK, Marty, but it looks like this PR includes commits from develop, post 21.3-branch. I think you'll need to recreate the FB from the 21.3 branch.

PR #2070, opened 2021-03-05
Improvements to File Root Settings

bbimber:
This PR adds two changes:

- Create standalone action to set file roots, not linked to import wizard. The usage is to allow containers like workbooks to more easily set their file root (which is allowed). The standard Folder Management doesnt permit this. Adding this new action seemed like the least redundancy in code, since there is a lot of complexity around setting the file root.

- Note related to above: I noticed that when one changes the file root to a custom root, LK give you options to kick off a pipeline job to update this. However, if the folder has a custom root and you change to a different path, these options dont show up. I think we want them to, right? This adds a onchange listener to do that. Note: there is some complexity around cloud roots I dont quite get, and based on the pre-existing comment "TODO: must we prevent moving when switching *TO* unmanaged cloud root" it seems like maybe the right behavior isnt clear?

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -5121,7 +5121,8 @@ public class AdminController extends SpringActionController
 
     /**
      * This standalone file root management action can be used on folder types that do not support
-     * the normal 'Manage Folder' UI.
+     * the normal 'Manage Folder' UI. Not currently linked in the UI, but available for direct URL
+     * navigation when a workbook needs it.
      */
     @RequiresPermission(AdminPermission.class)
     public class ManageFileRootAction extends FormViewAction<FileRootsForm>
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -5119,6 +5119,52 @@ public class AdminController extends SpringActionController
         }
     }
 
+    /**
+     * This standalone file root management action can be used on folder types that do not support
+     * the normal 'Manage Folder' UI.
+     */
+    @RequiresPermission(AdminPermission.class)
+    public class ManageFileRootAction extends FormViewAction<FileRootsForm>
+    {
+        @Override
+        public ModelAndView getView(FileRootsForm form, boolean reShow, BindException errors)
+        {
+            JspView view = getFileRootsView(form, errors, getReshow());
+            getPageConfig().setTitle("Manage File Root");
+            return view;
+        }
+
+        @Override
+        public void validateCommand(FileRootsForm form, Errors errors)
+        {
+            validateCloudFileRoot(form, getContainer(), errors);
+        }
+
+        @Override
+        public boolean handlePost(FileRootsForm form, BindException errors) throws Exception
+        {
+            return handleFileRootsPost(form, errors);
+        }
+
+        @Override
+        public ActionURL getSuccessURL(FileRootsForm form)
+        {
+            ActionURL url = getContainer().getStartURL(getUser());
+
+            if (getViewContext().getActionURL().getReturnURL() != null)
+            {
+                url.addReturnURL(getViewContext().getActionURL().getReturnURL());
+            }
+
+            return url;
+        }
+
+        @Override
+        public void addNavTrail(NavTree root)
+        {
+        }
+    }
+
     @RequiresPermission(AdminPermission.class)
     public class FileRootsAction extends FolderManagementViewPostAction<FileRootsForm>
     {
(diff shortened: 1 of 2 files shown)


bbimber:
@labkey-jeckels this is related to a discussion a while back about enabling workbooks to set their file roots. While this is technically possible today, getting UI to do this provided tricky, so this proposal adds a standalone action to do that.

josheckels:
@bbimber How does one get to this new action?

bbimber:
@josheckels Well currently it would be if someone provided a link. This is an admin action. Since LabKey has the existing folder management UI with that same UI, there did not seem to be a point in putting it into any existing LabKey UI I am aware of. If you think there is some UI where it'd be useful I would be happy to do that.

josheckels:
@bbimber are you planning to add a link somewhere on your side then? I'm trying to understand the planned usage

bbimber:
@josheckels My planned usage is to provide a reasonable way to update the file root of workbooks. I originally thought that perhaps I'd write more like an API-level action; however, when I looked closer into the code, I concluded that the UI and POST handling seemed complicated enough that it made more sense to go through that codepath. 

My immediate use case is to just navigate there a few times to update workbooks I need. This is not all that common an event. However, yes, maybe I would make a more formal link to this action somewhere in our UI. 

I'm happy to look at other ways to enable this. The rationale for this additional action was that it seemed to involve the least duplication of code. 

A more sophisticated implementation might be to poke this into the admin menu, specifically of a workbook, so that menu might have "Manage Folder", which is the existing link that goes to the parent, and also "Manage Workbook Root", which goes to this page. I would also be willing to try to adapt the core manage folder UI to also work on a workbook. My impression on these ideas is that a) workbooks are not really that heavily used across much of LK, and b) we all remember all the issues implementing ContainerType. Doing the above properly would need to hook into that. I havent actually tried to code anything, but my initial impression is that it would get convoluted pretty fast. Am happy to take some additional steps on this feature if you want though.

josheckels:
@bbimber OK, that makes sense. Thanks. I'm not pushing back on the approach, just wanting to understand the use case.

I'll make a couple of small suggestions and create the LK-side copy of the branch afterwards to get TeamCity to build it.

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -5119,6 +5119,52 @@ public void addNavTrail(NavTree root)
         }
     }
 
+    /**
+     * This standalone file root management action can be used on folder types that do not support
+     * the normal 'Manage Folder' UI.
```
josheckels:
```suggestion
     * the normal 'Manage Folder' UI. Not currently linked in the UI, but available for direct URL navigation when a workbook needs it

```

core/src/org/labkey/core/admin/view/filesProjectSettings.jsp
```diff
@@ -310,6 +310,42 @@
         }
     }
 
+    function onRootChange()
+    {
+
+        var value = document.getElementById('folderRootPath').value;
+        var isChangeFromExisting = value != '<%=h(bean.getFolderRootPath())%>';
```
josheckels, line 337:
```suggestion
        var isChangeFromExisting = value != <%=q(bean.getFolderRootPath())%>;

```

bbimber:
Yeah, very reasonable. I would like to integrate this more meaningfully into the UI if there was a way that made sense to do it. As I wrote that, poking something into the admin menu probably make the most sense, but I dont know how cleanly we could do that without making the core code "workbook-aware"

core/src/org/labkey/core/admin/view/filesProjectSettings.jsp
```diff
@@ -310,6 +310,42 @@
         }
     }
 
+    function onRootChange()
```
josheckels, line 333: Can we share this with the existing code instead of copy/pasting it? Might need to introduce an argument or two, but looks like it's otherwise a duplicate, correct?

josheckels (changes requested):

bbimber:
@josheckels I have those changes over on my fork, but I dont think I have permission to push them to the LK side:

https://github.com/bbimber/platform/tree/fb_fileroot

I was able to clean up a couple other pre-existing cases where I think the migrate options field wasnt getting appropriately cleared, such as toggling to one setting and them toggling back to the original (i.e. there is no change).

josheckels:
I mirrored the change into the LK repo to get TeamCity testing it. I'll do a followup code review later today.

https://github.com/LabKey/platform/tree/fb_fileroot

labkey-jeckels (approved):

labkey-jeckels:
TeamCity looks happy and this is working in my testing. Merging in a moment.

bbimber:
@labkey-jeckels Hi Josh, is there a chance you would consider accepting ManageFileRootAction into release21.3? I am considering how to push this out to our servers. While I could try to deploy using a different core branch, I think it would be logistically easiest for me to continue deploying using the official release branch. I dont know what approval/logistical hurdles are required to add changes to a release these days.

PR #2074, opened 2021-03-08
Remove support for PostgreSQL 9.5

labkey-adam:
#### Rationale
PostgreSQL is now considered end-of-life (EOL) and is no longer supported; see https://www.postgresql.org/support/versioning. We deprecated PostgreSQL 9.5 in 21.3.0 and are removing it completely for 21.7.0.

Also:

- Move admin message handling to the base PostgreSQL dialect since we no longer use method overrides to convey deprecation warnings
- Speed up the `DialectRetrievalTestCases` by not testing an absurd number of invalid cases
- Correct some comments


labkey-tchad:
I've updated TeamCity to use 9.6 for compatibility testing on develop.

josheckels (approved):

PR #2076, opened 2021-03-08
Improvements for MxN Panorama reproducibility reports - plotting functions and error reporting

labkey-jeckels:
#### Rationale
Panorama is leveraging our visualizations and needs more click and rendering options

#### Changes
* Support a new 'sequential' point arrangement that orders the scatter plot deterministically (unlike jittering) and based on the input data ordering
* Support clickFn on box plots
* Show error when an AJAX details tooltip fails to load

### DIFF
--- a/internal/webapp/vis/src/internal/D3Renderer.js
+++ b/internal/webapp/vis/src/internal/D3Renderer.js
@@ -655,10 +655,14 @@ LABKEY.vis.internal.D3Renderer = function(plot) {
     // position log gutter data relative to x or y axis
     var logGutterPointsOffset = 15;
 
-    // Constants for position settings
+    // Constants for position and scaleType settings
     var position = {
         jitter: 'jitter',
         sequential: 'sequential'
+    };
+    var scaleType = {
+        discrete: 'discrete',
+        continuous: 'continuous'
     };
 
     var initLabelElements = function() {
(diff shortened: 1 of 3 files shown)


internal/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -1936,15 +1936,37 @@ LABKEY.vis.internal.D3Renderer = function(plot) {
                 colorAcc, sizeAcc, shapeAcc, hoverTextAcc, jitterIndex = {};
         layer = getLayer.call(this, geom);
 
-        if (geom.xScale.scaleType == 'discrete' && geom.position == 'jitter') {
+        // For sequential jitters, keep track of the current count for a given x value
+        var jitters = {};
+
+        if (geom.xScale.scaleType === 'discrete' && (geom.position === 'jitter' || geom.position === 'sequential')) {
```
labkey-ankurj: seems like strings 'discrete', 'jitter' and 'sequential' can me made as constants.

labkey-ankurj (changes requested):

internal/webapp/vis/src/internal/D3Renderer.js
labkey-jeckels: Indeed they can, as could many other strings that aren't constants in this file. I made constants for the position and scaleType values but don't want to revamp everything as part of this change.

cnathe (approved):
Looks good. I did some quick local testing using our index.html and demo_data code in /platform/internal/webapp/vis/demo/ folder.

labkey-ankurj (approved):

PR #2077, opened 2021-03-08
Location.locationPath and Location.locationPathDisplay

labkey-matthewb:
#### Rationale
Precompute location paths for performance and query simplicity


labkey-susanh (approved):
These changes seem straightforward enough. Did you want the inventory changes also reviewed?

PR #2079, opened 2021-03-09
Issue 42617: search index deleted and last indexed values reset on every module version change

labkey-adam:
#### Rationale
Clearing the "last indexed" values can be very time consuming for large tables. The original intent was to delete the index and clear last indexed on every search module upgrade, but it's actually happening on every module upgrade, which can be annoying to developers. Change to search module upgrades only.


labkey-kevink (approved):
Reviewed and verified fixed

PR #2080, opened 2021-03-09
Improve study functionality without specimen module

labkey-tchad:
#### Rationale
The Specimen module is no longer included in our community distribution. There are numerous hidden dependencies within the study module that break functionality when the specimen module isn't present. This fixes some of those and adds regression test suites to help as we continue to pull specimen functionality out of the study module.

#### Changes
* Fix some study imports when specimen module is missing [product]
* Update some tests to work when specimen module is missing
* Create `studySansSpecimen` and `assaySansSpecimen` test suites


labkey-adam (approved):

study/test/src/org/labkey/test/tests/study/StudyPublishTest.java
```diff
@@ -496,97 +514,100 @@ else if (visits.length > 0)
...
+
+                    if ((removePhiColumns && (response.getRowCount().intValue() != 0))
+                            || (!removePhiColumns && (response.getRowCount().intValue() != 10)))
+                    {
+                        // Fail with a useful screenshot
+                        t1.ensureColumnPresent(field);
+                        t1.setFilter(field, filterOperator.getDisplayValue(), null);
+                        assertEquals("SelectRows didn't return the same number of rows as contained in data region (both should have been zero)", response.getRowCount().intValue(), t1.getDataRowCount());
+                        fail("PHI column was not exported as expected: " + (removePhiColumns ? "PHI not removed" : "PHI removed") + " for " + field);
+                    }
```
labkey-danield, line 561: I had to look at this code for a while before I understood what it was doing (make sure column is visible, and filter result to useful data, then fail because the count is wrong or because the PHI column was not exported. Right?). Maybe you could elaborate in the comment a bit more?

study/test/src/org/labkey/test/tests/study/StudyPublishTest.java
```diff
@@ -496,97 +514,100 @@ else if (visits.length > 0)
...
-                    t1.ensureColumnPresent(field);
-                    t1.setFilter(field, filterOperator.getDisplayValue(), null);
-                    assertEquals("SelectRows didn't return the same number of rows as contained in data region (both should have been zero)", response.getRowCount().intValue(), t1.getDataRowCount());
-                    fail("PHI column was not exported as expected: " + (removePhiColumns ? "PHI not removed" : "PHI removed") + " for " + field);
+                    t2.setFilter("MouseId", "Does Not Equal Any Of (example usage: a;b;c)", createOneOfFilterString(ptids));
+                    assertTextPresent("No data to show.");
+                    t2.clearFilter("MouseId");
                 }
+                t2.setFilter("SequenceNum", "Does Not Equal Any Of (example usage: a;b;c)", createOneOfFilterString(visits));
+                assertTextPresent("No data to show.");
```
labkey-danield, line 577: Do you really want to assert here and two lines above and stop any future testing? Would checker() be more appropriate?

study/test/src/org/labkey/test/tests/study/StudyPublishTest.java
```diff
@@ -496,97 +514,100 @@ else if (visits.length > 0)
...
             {
-                t2.setFilter("MouseId", "Does Not Equal Any Of (example usage: a;b;c)", createOneOfFilterString(ptids));
-                assertTextPresent("No data to show.");
-                t2.clearFilter("MouseId");
+                waitForText(WAIT_FOR_JAVASCRIPT, "No specimens found.");
+            }
+            // verify that the specimen request options are hidden from the manage study page
+            goToManageStudy();
+            assertTextNotPresent("Specimen Repository Settings", "Repository Type", "Display and Behavior", "Specimen Request Settings");
+            assertTextPresent("Note: specimen repository and request settings are not available for published studies.");
```
labkey-danield, line 595: Same question regarding assert vs checker(). Assert for a text value will stop all future testing if any part of the text string changes.

labkey-danield (approved):
Just a couple of questions/comments in the test code.

labkey-tchad:
@labkey-danield Those lines you commented on are unchanged from the existing test. I just hid some sections of the tests behind checks for the specimen module. The actual scope of the changes is clearer if you [hide whitespace changes](https://github.com/LabKey/platform/pull/2080/files?diff=split&w=1).

PR #2081, opened 2021-03-09
@labkey/build updates to webpack alias from /src instead of /dist

cnathe:
#### Rationale
This PR pulls in the @labkey/build updates and related package updates for being able to use HMR in the LK module to build from /src for all @LabKey packages for the various platform modules.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/471
* https://github.com/LabKey/biologics/pull/822
* https://github.com/LabKey/inventory/pull/204
* https://github.com/LabKey/sampleManagement/pull/512
* https://github.com/LabKey/platform/pull/2081

#### Changes
* package.json updates and related jest updates
* move import of .scss files to the app.tsx and dev.tsx (core module)
* use @labkey/components-scss alias in the entryPoints scss files
* add in main.d.ts file for LABKEY typings check when using start-link
* remove tsconfig.json file


assay/package.json
```diff
@@ -12,9 +12,11 @@
     "clean": "rimraf resources/web/assay/gen && rimraf resources/views/gen && rimraf resources/views/*.*"
   },
   "dependencies": {
-    "@labkey/components": "2.11.1"
+    "@labkey/components": "2.11.0-fb-webpackAliasFromSrc.0"
   },
   "devDependencies": {
-    "@labkey/build": "2.0.1"
+    "@labkey/build": "2.0.1-fb-webpackAliasFromSrc.3",
+    "@types/jest": "26.0.20",
```
labkey-nicka: Did you add these dependencies on `jest` for the assay module (I see you added it to others as well like `issues`) to be consistent or because it was now necessary? I guess I'm wondering if previously it was just getting picked up and/or if there were any jest tests in the module.

core/src/client/LabKeyUIComponentsPage/LabKeyUIComponentsPage.tsx
```diff
@@ -389,7 +387,14 @@ export class App extends React.Component<any, State> {
                     )
                 }
                 {selected === 'PermissionAssignments' &&
-                    <PermissionAssignmentsPage/>
+                    <PermissionAssignmentsPage
```
labkey-nicka: Hmm, I wonder why this wasn't error'ing previously. Best I can tell all of these props have been marked as required on `PermissionsProviderProps` since at least April of 2020. That said, having them all be `undefined` smells of actually meaning they should all be optional. Don't have to make a change, just interesting.

cnathe: Thank you for the reminder. I threw in those undefineds to get the start-link build working but then forgot to go back and investigate. I agree that I am no sure why this wasn't failing before this change. I'll take a look in the morning.

assay/package.json
cnathe: I added them because in the start-link case when the @labkey/components src is included in what is build, there are some test helpers that use jest so these modules that didn't have the jest package dependencies were failing. Maybe there was a different way to handle this by doing something in the components package. Thoughts?

core/src/client/LabKeyUIComponentsPage/LabKeyUIComponentsPage.tsx
labkey-nicka: Looks good.

assay/package.json
labkey-nicka: With https://github.com/LabKey/labkey-ui-components/pull/471/commits/7804468c I updated `@labkey/components` to no longer make it necessary for packages to need to require jest for `npm run start-link` to work.

labkey-nicka (approved):

PR #2085, opened 2021-03-11
fix for targetStudyTest

labkey-klum:
#### Rationale
The targetStudyTest was failing because the target study column did not appear in the publish confirm grid. This was a bit of code that we had questions about during the most recent refactor and had commented out the logic which hid the column only if it did not appear in the results domain.

Factored out that bit of code into the assay publish confirm action.

### DIFF
--- a/study/src/org/labkey/study/publish/AbstractPublishConfirmAction.java
+++ b/study/src/org/labkey/study/publish/AbstractPublishConfirmAction.java
@@ -191,6 +191,15 @@ public abstract class AbstractPublishConfirmAction<FORM extends PublishConfirmFo
      */
     protected abstract ActionURL copyToStudy(FORM form, Container targetStudy, Map<Integer, PublishKey> publishData, List<String> publishErrors);
 
+    /**
+     * Specifies the columns in the publish results query view that should not be visible (but still be in the data view)
+     * @return
+     */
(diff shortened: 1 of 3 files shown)


study/src/org/labkey/study/publish/AbstractPublishConfirmAction.java
```diff
@@ -191,6 +191,15 @@ public URLHelper getSuccessURL(FORM form)
      */
     protected abstract ActionURL copyToStudy(FORM form, Container targetStudy, Map<Integer, PublishKey> publishData, List<String> publishErrors);
 
+    /**
+     * Specifies the columns in the publish results query view that should not be visible (but still be in the data view)
```
RosalineP, line 195: Ahhh 👍

study/src/org/labkey/study/publish/AbstractPublishConfirmAction.java
```diff
@@ -210,7 +219,8 @@ public ModelAndView getView(FORM form, boolean reshow, BindException errors) thr
                 mismatched,
                 form.isIncludeTimestamp(),
                 getAdditionalColumns(form),
-                getHiddenFormFields(form));
+                getHiddenFormFields(form),
```
RosalineP, line 222: 👍

RosalineP (approved):
Neat! Good to know what was going on. Changes look good

PR #2086, opened 2021-03-12
Issue 42168: Show queue position for jobs in the Data Pipeline webpart

josheckels:
#### Rationale
On shared servers, users' jobs may sit in the pipeline queue for some time. We can help them understand their place in line, at least when the server is configured to use JMS.

#### Changes
* Show job queue position in pipeline details page, hooking into page refresh
* Add column to pipeline job grid too

### DIFF
--- a/pipeline/src/org/labkey/pipeline/api/PipelineQuerySchema.java
+++ b/pipeline/src/org/labkey/pipeline/api/PipelineQuerySchema.java
@@ -41,7 +41,6 @@ import org.labkey.api.query.UserSchema;
 import org.labkey.api.security.User;
 import org.labkey.api.security.permissions.AdminPermission;
 import org.labkey.api.util.HtmlString;
-import org.labkey.pipeline.mule.EPipelineQueueImpl;
 import org.labkey.pipeline.query.TriggerConfigurationsTable;
 
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


pipeline/src/org/labkey/pipeline/api/PipelineQuerySchema.java
```diff
@@ -108,6 +114,9 @@ public FieldKey getContainerFieldKey()
                 table.setContainerFilter(new ContainerFilter.AllFolders(getUser()));
             }
 
+            MutableColumnInfo positionCol = table.addWrapColumn("QueuePosition", table.getRealTable().getColumn("RowId"));
+            positionCol.setDisplayColumnFactory(QueuePositionDisplayColumn::new);
```
labkey-kevink, line 117: Probably need to disable sort and filter on this column.

pipeline/src/org/labkey/pipeline/api/PipelineQueueImpl.java
```diff
@@ -257,6 +260,13 @@ public synchronized PipelineJobData getJobDataInMemory(Container c)
         return ret;
     }
 
+    @Override
+    @NotNull
+    public Map<Integer, Integer> getQueuePositions()
+    {
+        return Collections.emptyMap();
```
labkey-kevink: I don't know enough about the basic pipeline queue implementation.  Is it not possible to know the queue positions?

labkey-kevink (approved):

pipeline/src/org/labkey/pipeline/api/PipelineQuerySchema.java
josheckels: Yes, this is already handled via the display column

pipeline/src/org/labkey/pipeline/api/PipelineQueueImpl.java
josheckels: Thanks for the nudge. I thought I'd have to dig into the bowels of Mule's internal state but PipelineQueueImpl had lists for the running and queued jobs so it was easy. Now supported on both.

PR #2088, opened 2021-03-12
can you give me more HtmlString?

labkey-matthewb:
#### Rationale
Find and fix more places that return String that should return HtmlString

### DIFF
--- a/api/src/org/labkey/api/util/HtmlStringBuilder.java
+++ b/api/src/org/labkey/api/util/HtmlStringBuilder.java
@@ -54,6 +54,12 @@ public class HtmlStringBuilder implements HasHtmlString, SafeToRender
         return this;
     }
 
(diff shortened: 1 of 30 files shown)


api/src/org/labkey/api/util/HtmlStringBuilder.java
```diff
@@ -72,6 +76,26 @@ public HtmlStringBuilder append(HasHtmlString hhs)
...
+        return this;
+    }
+    public HtmlStringBuilder endTag(@NotNull String tag)
+    {
+        assert(tagPattern.matcher(tag).matches());
+        _sb.append("</").append(tag).append(">");
+        return this;
+    }
+
+
```
labkey-adam, line 104: 👍

labkey-adam (approved):
Need to fix build break in adjudication

PR #2097, opened 2021-03-16
Address "There is already an object named 'xxx' in the database" study design exceptions

labkey-adam:
#### Rationale
TeamCity is seeing frequent errors such as "There is already an object named 'c43d424_productantigen' in the database." This PR proactively creates the study design provisioned tables at study creation time, instead of using an "ensureDomain()" anti-pattern. It also makes the "AssaySpecimen" and "AssaySpecimenVisit" tables visible when the specimen module is missing or inactive; crazy me, I had thought these specimen tables were related to specimens.

Partial fix for [Issue 42641: Intermittent table name collision when importing study archive](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42641)


experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
```diff
@@ -166,7 +176,21 @@ private String _create(DbScope scope, DomainKind<?> kind, Domain domain)
 
             change.setForeignKeys(domain.getPropertyForeignKeys());
 
-            change.execute();
+            try
+            {
+                log.info("Attempting to create " + tableName);
+                change.execute();
+                RECENTLY_CREATED_TABLES.put(tableName, Thread.currentThread().getStackTrace());
```
labkey-jeckels, line 183: Should we log if the table was somehow already in the cache?

labkey-jeckels (approved):

experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
labkey-adam: I didn't bother because 1) the only way you'd be in that situation is if you created the same table twice and 2) Cache.put() is void, unlike a map.

PR #2098, opened 2021-03-16
8591: Link to Study UI Elements

RosalineP:
#### Rationale
For Sample Type designer on LKS, add the form 'Auto-Link Data to Study' dropdown element to 'Properties' panel, and support full roundtrip of this new attribute.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/518
* https://github.com/LabKey/labkey-ui-components/pull/477
* https://github.com/LabKey/biologics/pull/829

#### Changes
* Bump ui-components version to pull in Sample Type designer UI changes
* Add 'AutoLinkTargetContainerId' column to 'exp.MaterialSource' table to hold new 'Auto-Link Data to Study' attribute data 
* Roundtrip new 'Auto-Link Data to Study' attribute data on creation and update of Sample Types
* Surface 'Link To Study' button in
    - Sample Types webpart
    - showSampleType view (experiment-showSampleType.view?rowId=`N`)
* Surface 'Link To Study History' button in Sample Type properties panel


experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -320,6 +320,17 @@ public void setMetricUnit(String metricUnit)
         return _object.getMetricUnit();
     }
 
+    public void setAutoLinkTargetContainerId(String autoLinkTargetContainerId)
```
RosalineP: Thought we might want to leave this intentionally not public in the `ExpSampleType` interface, similar to `setMetricUnit()` above at line 312?

experiment/src/org/labkey/experiment/api/ExpSampleTypeTableImpl.java
```diff
@@ -97,6 +98,7 @@ protected void populateColumns()
         addColumn(ExpSampleTypeTable.Column.NameExpression).setHidden(true);
         addColumn(ExpSampleTypeTable.Column.LabelColor).setHidden(true);
         addColumn(ExpSampleTypeTable.Column.MetricUnit).setHidden(true);
+        addColumn(ExpSampleTypeTable.Column.AutoLinkTargetContainerId).setHidden(true);
```
RosalineP: Assuming this should also be a hidden column

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -599,13 +598,13 @@ public ExpSampleTypeImpl createSampleType(Container c, User u, String name, Stri
...
         return createSampleType(c, u, name, description, properties, indices, idCol1, idCol2, idCol3,
-                parentCol, nameExpression, templateInfo, null, null, null);
+                parentCol, nameExpression, templateInfo, null, null, null, null);
     }
 
     @NotNull
     @Override
     public ExpSampleTypeImpl createSampleType(Container c, User u, String name, String description, List<GWTPropertyDescriptor> properties, List<GWTIndex> indices, int idCol1, int idCol2, int idCol3, int parentCol,
-                                              String nameExpression, @Nullable TemplateInfo templateInfo, @Nullable Map<String, String> importAliases, @Nullable String labelColor, @Nullable String metricUnit)
+                                              String nameExpression, @Nullable TemplateInfo templateInfo, @Nullable Map<String, String> importAliases, @Nullable String labelColor, @Nullable String metricUnit, @Nullable String autoLinkTargetContainerId)
```
RosalineP: Would it be a good idea to check whether `autoLinkTargetContainerId` is valid for the current container around line 655, here? (Following the pattern of the other checks-and-throws within the function)

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
labkey-klum: That sounds reasonable, as the interface looks immutable (no setters)

experiment/src/org/labkey/experiment/api/ExpSampleTypeTableImpl.java
labkey-klum: yep

experiment/resources/schemas/exp.xml
```diff
@@ -355,6 +355,7 @@
       <column columnName="NameExpression"/>
       <column columnName="LabelColor"/>
       <column columnName="MetricUnit"/>
+      <column columnName="AutoLinkTargetContainerId"/>
```
RosalineP: Noticed that `DatabaseDiagnosticsTest.databaseCheckTest` was running red due to `ERROR: Inconsistency in Schema exp
ERROR: Table "materialsource", column "autolinktargetcontainerid" missing from XML`. Checked out some examples and updated expTypes.xsd and exp.xml accordingly

experiment/resources/schemas/dbscripts/postgresql/exp-21.002-21.003.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE exp.MaterialSource ADD COLUMN AutoLinkTargetContainerId VARCHAR(50) NULL;
```
labkey-klum: If this field in the DB is truly a container ID we should model it as type: ENTITYID which is defined in core-0.00-18.30.sql as VARCHAR(36). Just so we are consistent across our data models on container fields.

experiment/resources/schemas/exp.xml
labkey-klum: expTypes.xsd is used in the XAR import/export for sample types. I assume you didn't add any code yet to support the new property but maybe add a note in the spec for one of the future stories.

labkey-klum (changes requested):
We should discuss how we represent the autoLinkTarget in our data models and interfaces. I think I'm more in favor of using Container objects instead of IDs on the database serialization/deserialization boundary and other than in the JSON serialization for the DomainKind.createDomain, it would be preferable to return container objects.

experiment/resources/schemas/dbscripts/postgresql/exp-21.002-21.003.sql
RosalineP: Updated type accordingly!

experiment/resources/schemas/exp.xml
RosalineP: Note added to spec!

experiment/src/org/labkey/experiment/api/MaterialSource.java
```diff
@@ -149,12 +148,12 @@ public void setMetricUnit(String metricUnit)
...
     }
 
-    public String getAutoLinkTargetContainerId()
+    public Container getAutoLinkTargetContainerId()
     {
         return _autoLinkTargetContainerId;
     }
 
-    public void setAutoLinkTargetContainerId(String autoLinkTargetContainerId)
+    public void setAutoLinkTargetContainerId(Container autoLinkTargetContainerId)
```
RosalineP: Would have been nice to rename these `getAutoLinkTargetContainer` and `setAutoLinkTargetContainer` since we're using Containers instead of Strings now, but renaming seems to break alignment with the SampleTypeDomainKindProperties bean

labkey-klum: I don't think I'm following this completely. Can you give an example or point me to the code where there is an implicit need to keep these methods aligned?

RosalineP: (Updated in patch as of 0ddf377)

labkey-klum (approved):

PR #2099, opened 2021-03-16
Fb sample type dataset

labkey-klum:
#### Rationale
In preparation for supporting sample type datasets in study, this feature branch provides preliminary work to abstract away much of the assay specific code in studies as well as the code which handles the creation and population of assay backed datasets. With these changes publishing and creation of these types of datasets should be more general purpose.

#### Changes
- Change AssayPublishService to StudyPublishService
- QC state updates to generalize the defaultAssayQCState property
- Create a builder pattern for dataset definitions to help with general dataset creation
- Update the dataset table to rename protocolId to publishSourceId and add a publishSourceType enum
- Generalize DatasetQueryView
- Dataset interface changes to remove: isAssayData and getAssayProtocol
- Introduce Dataset.PublishSource enum to help support both assays and samples


study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -2831,6 +2857,173 @@ public static void cleanupOrphanedDatasetDomains()
         }
     }
 
+    public static class Builder implements org.labkey.api.data.Builder<DatasetDefinition>
```
labkey-adam, line 2862: 👍

labkey-adam (approved):

api/src/org/labkey/api/study/Dataset.java
```diff
@@ -91,12 +200,19 @@
     Date getModified();
 
     /**
-     * @return true if this dataset is backed by assay data within LabKey Server. Note that if a dataset happens
-     * to contain assay data but isn't linked to an assay provider in the server (ie., when importing a study archive), this method will return false.
+     * @return true if this dataset is backed by published data (assay, sample type etc). Note that if a dataset happens
+     * to contain publised data but isn't linked to the publish source in the server (ie., when importing a study archive), this method will return false.
```
RosalineP:
```suggestion
     * to contain published data but isn't linked to the publish source in the server (ie., when importing a study archive), this method will return false.

```

study/src/org/labkey/study/writer/StudyExportContext.java
```diff
@@ -78,13 +78,15 @@ private static StudyDocument getStudyDocument()
 
     private void initializeDatasets(StudyImpl study)
     {
+        // TODO, add support for sample datasets
```
RosalineP, line 81: Is note to do work to support sample dataset export support, when a study is being exported? (Related to the Sample Type import/export work we've noted for the 'Miscellaneous Wrap Up' story?)

RosalineP (approved):
👍 👍

study/src/org/labkey/study/writer/StudyExportContext.java
labkey-klum: yes, we will eventually want to be able to independently select sample datasets from the study export UI

PR #2101, opened 2021-03-17
Prevent Container field by checking in DomainKind validation options.

Sigmonia:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=42636

#### Changes
* Update the base DomainKind to prevent creation of a field named `Container`

### DIFF
--- a/api/src/org/labkey/api/exp/property/DomainKind.java
+++ b/api/src/org/labkey/api/exp/property/DomainKind.java
@@ -331,5 +331,10 @@ abstract public class DomainKind<T>  implements Handler<String>
 
         if (isUpdate && !canEditDefinition(user, domain))
             throw new UnauthorizedException("You don't have permission to edit this domain");
+
+        //Case insensitive match
+        if (updatedDomainDesign.getFieldByName("Container") != null)
+            throw new IllegalArgumentException("Property name 'Container' is a reserved name.");
+
     }
 }
--- a/api/src/org/labkey/api/exp/property/DomainKind.java
+++ b/api/src/org/labkey/api/exp/property/DomainKind.java
@@ -331,10 +331,5 @@ abstract public class DomainKind<T>  implements Handler<String>
 
         if (isUpdate && !canEditDefinition(user, domain))
             throw new UnauthorizedException("You don't have permission to edit this domain");
-
-        //Case insensitive match
-        if (updatedDomainDesign.getFieldByName("Container") != null)
-            throw new IllegalArgumentException("Property name 'Container' is a reserved name.");
-
     }
 }
--- a/experiment/src/org/labkey/experiment/api/DataClassDomainKind.java
+++ b/experiment/src/org/labkey/experiment/api/DataClassDomainKind.java
@@ -19,6 +19,7 @@ import com.google.common.collect.Sets;
 import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.labkey.api.collections.CaseInsensitiveHashSet;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.DbSchema;
 import org.labkey.api.data.DbSchemaType;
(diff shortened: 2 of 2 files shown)


labkey-adam (approved):

api/src/org/labkey/api/exp/property/DomainKind.java
```diff
@@ -331,5 +331,10 @@ public void validateOptions(Container container, User user, T options, String na
 
         if (isUpdate && !canEditDefinition(user, domain))
             throw new UnauthorizedException("You don't have permission to edit this domain");
+
+        //Case insensitive match
+        if (updatedDomainDesign.getFieldByName("Container") != null)
+            throw new IllegalArgumentException("Property name 'Container' is a reserved name.");
```
labkey-susanh: One wonders if this is the only universally unavailable field name, but this will at least prevent some problems.

labkey-susanh (commented):
I believe our domain designers should currently prevent the creation of a field that matches a reserved field name.  If not, I wonder if some domains need to have their reserved field names updated as that may lead to a more graceful failure.  This change will be a general safeguard (if creating a domain via an API call perhaps?).

api/src/org/labkey/api/exp/property/DomainKind.java
labkey-jeckels: Yes, see DomainKind.getReservedPropertyNames(). Some implementations include "Container" and others don't. Can we tap into that mechanism instead of adding another place where we check for disallowed names? I think it might provide a better user experience in the designer too.

Sigmonia: This (or alternatively the calling method) is a nice choke point for the DomainForm create/update methods which cover the client apis and domain designers. ​I think adding a full validation loop for the reserved fields here would be appropriate, and would allow centralizing the existing checks out of the individual domains.

Given that this is going out as a hotfix, I was attempting to be a bit conservative and just check for the one field that is truly problematic. I think adding the typical fields `container, created, createdby, modified, modifiedby` to the base domain reserved properties would be the most direct solution, but again was trying to be a bit conservative for a patch.

It is also worth noting that Susan's recent changes for 42386 also address this on the client-side in a more elegant fashion in the `develop` branches.

labkey-susanh: This method is overridable, so it may not be a choke point.  It seems that it is only overridden in one place currently, which first calls this method, but in theory this method might be bypassed.  

I certainly like the idea of adding these field to the base domain's reserved properties but wonder if there are domains that didn't have these reserved and fields got defined and now they become reserved.  Perhaps not likely and perhaps not a problem, but, well, it's data and can appear in so many forms.  

If this is going into 21.3, I think we should pull forward the changes made in #2064 to use case-insensitive sets and that add at least "Container" as a reserved field where it was missing.

labkey-jeckels: Ignoring the call to super, DomainKind.validateOptions() is only called in one place, in PropertyController. There are many other places that save domains, so this is pretty limited protection. See callers to DomainKind.updateDomain() and DomainKind.createDomain(), for example.

We can do a limited fix for 21.3, but in develop I think we need to:

Make getReservedPropertyNames() aware of whatever names are actually reserved for each domain.
Be sure that the list is respected in all of the various codepaths.
If it's the responsibility of getReservedPropertyNames() to return a case insensitive set, make it explicit in the method signature. At the moment, the comment in DomainKind says it's the responsibility of the caller to do stuff in a case insensitive way if needed. Otherwise we'll keep adding new problematic implementations.

Sigmonia: Will address the broader discussion via email.

But pulled Susan's changes from develop for DataClassDomainKind and reverted my change.

labkey-susanh (approved):

labkey-jeckels (approved):
Looks good to me too

PR #2102, opened 2021-03-17
Issue 42712: Assay fields with OOR enabled are exported incorrectly

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42712

#### Changes
* Remove OutOfRangeDisplayColumn.getTsvFormattedValue override since it ends up calling back though OutOfRangeDisplayColumn.getDisplayValue which handles the concatenation of the OOR indicator and numberic value


labkey-jeckels (approved):
This change is an improvement over the previous behavior.

Note that a configured format string on the primary column doesn't get applied in the TSV export.

PR #2105, opened 2021-03-17
Consolidate domain designer entryPoints into core module

cnathe:
#### Rationale
Our platform repo has 7 different modules which have versions of the domain designer / field editor. This means that when we make an @labkey/components package update we usually update all 7 of those package.json files to pick up the changes. This also means that 7 modules have to install node_modules and run the npm build. Since are have vendor bundling setup for our prod build, this also means that each of those modules will have a version / copy of those vendor bundles. 

In an effort to streamline this, this PR moves all of the domain designer related React platform module entryPoints into the core module. This allowed me to remove the React build related pieces from 4 of those platform modules (list, query, issues, and study). This also allowed me to share some SCSS across all of these entryPoints. Finally, this resulted in a 3 minute reduction in build time on TeamCity for the community build (went from ~15 min on develop to ~12 min in this fb).

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/674

#### Changes
* move entryPoint info and React code for all domain designer usages into the core/src/client dir
* remove React build related items (package.json, package-lock.json, .npmrc, main.d.ts, etc.) from issues, list, query, and study modules
* consolidate each entryPoint to use the same DomainDesigner.scss
* update the related controller actions to get ModuleHtmlView resource from "core" module


labkey-susanh (approved):
Looks good!

labkey-nicka (approved):

PR #2113, opened 2021-03-19
Participant purge pipeline job

labkey-adam:
#### Rationale
TeamCity is seeing frequent failures like `ERROR: relation "c132d1770_sia_minus_1" already exists` caused by the Participant Purge timer thread running in parallel with study import. A simple workaround for now is to run the Participant Purge handling as a pipeline job, ensuring that it will never run in parallel with another pipeline job. [Issue 42641: Intermittent table name collision when importing study archive](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42641)

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2097

#### Changes
* Key the POTENTIALLY_DELETED_PARTICIPANTS map on String (containerId) instead of Container. Safer and happier serialization.
* If the map is not empty, the timer thread now queues a participant purge pipeline job instead of directly invoking this code

### DIFF
--- a/study/src/org/labkey/study/visitmanager/PurgeParticipantsJob.java
+++ b/study/src/org/labkey/study/visitmanager/PurgeParticipantsJob.java
@@ -1,7 +1,5 @@
 package org.labkey.study.visitmanager;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.dialect.SqlDialect;
@@ -18,22 +16,19 @@ import java.io.File;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 
 public class PurgeParticipantsJob extends PipelineJob
 {
-    private final Map<String, Set<String>> _potentiallyDeletedParticipants;
-
-    @SuppressWarnings("unused")
-    @JsonCreator
-    public PurgeParticipantsJob(@JsonProperty("_potentiallyDeletedParticipants") Map<String, Set<String>> potentiallyDeletedParticipants)
+    @SuppressWarnings("unused") // For deserialization
+    public PurgeParticipantsJob()
     {
-        _potentiallyDeletedParticipants = potentiallyDeletedParticipants;
     }
 
-    PurgeParticipantsJob(ViewBackgroundInfo info, PipeRoot pipeRoot, Map<String, Set<String>> potentiallyDeletedParticipants)
+    PurgeParticipantsJob(ViewBackgroundInfo info, PipeRoot pipeRoot)
     {
         super(null, info, pipeRoot);
-        _potentiallyDeletedParticipants = potentiallyDeletedParticipants;
         setLogFile(new File(pipeRoot.getLogDirectory(), FileUtil.makeFileNameWithTimestamp("purge_participants", "log")));
     }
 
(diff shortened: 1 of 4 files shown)


study/src/org/labkey/study/visitmanager/PurgeParticipantsTask.java
```diff
@@ -36,99 +36,41 @@
...
-                        // throw them back on the queue
-                        VisitManager vm = StudyManager.getInstance().getVisitManager(study);
-                        vm.scheduleParticipantPurge(potentiallyDeletedParticipants);
-                    }
-                }
-            }
+        try
+        {
+            PipelineJob job = new PurgeParticipantsJob(vbi, root, _potentiallyDeletedParticipants);
+            LOG.info("Queuing PurgeParticipantsJob [thread " + Thread.currentThread().getName() + " to " + PipelineService.get().toString() + "]");
```
labkey-jeckels: Do we want to tone this down to a debug level message?

study/src/org/labkey/study/visitmanager/PurgeParticipantsTask.java
```diff
@@ -36,99 +36,41 @@
...
-                    {
-                        // throw them back on the queue
-                        VisitManager vm = StudyManager.getInstance().getVisitManager(study);
-                        vm.scheduleParticipantPurge(potentiallyDeletedParticipants);
-                    }
-                }
-            }
+        try
+        {
+            PipelineJob job = new PurgeParticipantsJob(vbi, root, _potentiallyDeletedParticipants);
```
labkey-jeckels: This may queue up multiple jobs if there are multiple deletes from datasets while a purge task is in the queue. That's not horrible, but it might clutter the list.

The alternative would be to not queue another job if there's already one in the list (see TransformManager.transformIsPending() for how ETLs avoid this) and to not pass the collection to the job constructor but instead have it reach back to use PurgeParticipantsTask. _potentiallyDeletedParticipants directly.

study/src/org/labkey/study/visitmanager/PurgeParticipantsJob.java
```diff
@@ -0,0 +1,152 @@
...
+import org.labkey.study.model.StudyManager;
+
+import java.io.File;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+public class PurgeParticipantsJob extends PipelineJob
+{
+    private final Map<String, Set<String>> _potentiallyDeletedParticipants;
```
labkey-jeckels: Because jobs are serialized/deserialized as part of queuing (to test that they round-trip cleanly), this won't end up being the same object as in PurgeParticipantTask. Thus, the fancy synchronization below is effectively useless here.

labkey-jeckels (approved):
This is OK as-is, but see comments for ideas on avoiding double-queuing purge jobs and how we might share references to the collection of purgeable participants

study/src/org/labkey/study/visitmanager/PurgeParticipantsTask.java
labkey-adam: Done

study/src/org/labkey/study/visitmanager/PurgeParticipantsTask.java
labkey-adam: Should be addressed. I refactored so that DatasetUpdateService's purge invocation bypasses the pipeline job and runs synchronously, as it used to. This allows the job to work directly on POTENTIALLY_DELETED_PARTICIPANTS. I used a simple volatile boolean to avoid queueing multiple jobs, instead of the transformIsPending() approach.

study/src/org/labkey/study/visitmanager/PurgeParticipantsJob.java
labkey-adam: Fixed

study/src/org/labkey/study/visitmanager/PurgeParticipantsJob.java
```diff
@@ -0,0 +1,171 @@
...
+                    int deleted = VisitManager.performParticipantPurge(study, _potentiallyDeletedParticipants);
+                    if (deleted > 0)
+                    {
+                        StudyManager.getInstance().getVisitManager(study).updateParticipantVisitTable(null, null);
+                    }
+                    _info.accept("Finished purging participants in " + _container.getPath());
+                }
+                catch (TableNotFoundException tnfe)
+                {
+                    // Just move on if container went away
```
labkey-jeckels: I find this comment misleading - it's effectively commenting a non-existent else block?

labkey-jeckels (approved):
One gripe about a preexisting comment. New approach looks good.

study/src/org/labkey/study/visitmanager/PurgeParticipantsJob.java
labkey-adam: Yes, I think it's commenting on the fall through case. I updated it for clarity.

PR #2114, opened 2021-03-19
Issue 40737: Remove legacy lineage calculations

labkey-jeckels:
#### Rationale
We've iterated on the lineage calculations and the legacy version has been only available as an opt-in experimental feature for a while. We've confirmed the one deployment that still had it turned on no longer needs it, so it's time to consolidate.

#### Changes
* Remove the legacy approach, the experimental feature, and the code that diff'd the new and legacy approaches
* Clean up the Dataset API slightly


experiment/src/org/labkey/experiment/api/LineagePerfTest.java
```diff
@@ -333,7 +320,7 @@ public void lineagePerformance() throws Exception
 //        lineageQueries("OLD", oldLineageQuery, oldLineageGraph, oldInsertMoreTimer, ss, firstData);
 
         LOG.info("TEST querying with exp.edge lineage: ");
-        ExperimentalFeatureService.get().setFeatureEnabled(ExperimentServiceImpl.EXPERIMENTAL_LEGACY_LINEAGE, false, _user);
+
```
labkey-kevink, line 315: Can you also remove the commented out call to `lineageQueries` and the associated timers since we won't need to compare them any longer?

labkey-kevink (approved):
Looks good.  Just need to remove the old lineage parts from the lineage perf test.

experiment/src/org/labkey/experiment/api/LineagePerfTest.java
labkey-jeckels: Done

PR #2116, opened 2021-03-20
Issue 42452: Incessant per-request logging spam from WsRemoteEndpointImplServer.doClose()

labkey-jeckels:
#### Rationale
Some dev machines (like mine) see lots of logging from Tomcat complaining that WebSockets aren't being closed appropriately. This is from a combination of java.util.logging and Tomcat's JULI configuration that is getting reset by some third-party library during initialization. I've failed to isolate the exact cause, but it seems like moving the logging level reset to later in the codepath has the intended result.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/62

#### Changes
* Set the log level when we're starting to wire up WebSocket code


labkey-kevink (approved):
I haven't experienced this issue, but the change seems fine.

PR #2120, opened 2021-03-23
Issue 42620: AliasedColumn should never be considered key field for the TableInfo

cnathe:
#### Rationale
Related to NAb runs grid performance issue: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42620

The fix made in the related PR for 21.3 was just to explicitly set the AliasedColumns for the given NAb case as `setKeyField(false)`. However, an AliasedColumn should never be considered a key field, so this PR sets that in the AliasedColumn constructor after it does the `copyAttributesFrom(column)` call.

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/307

#### Changes
* Call setKeyField(false) in AliasedColumn constructor after copyAttributesFrom

### DIFF
--- a/api/src/org/labkey/api/query/AliasedColumn.java
+++ b/api/src/org/labkey/api/query/AliasedColumn.java
@@ -38,6 +38,9 @@ public class AliasedColumn extends BaseColumnInfo
         super(key, parent);
         copyAttributesFrom(column);
 
+        // Issue 42620: alias columns should never be considered key fields
+        setKeyField(false);
+
         Map<FieldKey, FieldKey> remap = new HashMap<>();
         remap.put(column.getFieldKey(), key);
         if (parent != null && parent != column.getParentTable())


labkey-matthewb (approved):

cnathe:
@labkey-matthewb There were some TeamCity test failures which pointed out an issue with this change. For example there are usages of the `AliasedColumn` which are used to wrap all columns of a table. For example, the `StorageProvisionerImpl.wrappAllColumns()` case. In that case, we would want to not override the `setKeyField` on these columns. Thoughts on how I can handle this case?

labkey-matthewb:
some ideas in no particular order 1) In this sort of scenario, I've been tending toward hiding the public constructors of the class, and creating static constructor methods that indicate the usage purpose (copyUp(), renameColumn(), duplicateColumn()).  That would also get us closer to being able to ditch having both AliasedColumn and WrappedColumnInfo.  2) adding a new flag to the constructor would force the caller to make a choice, but doesn't seem to get to the heart of the problem 3) Maybe just focus on the TableInfo helpers.  wrapColumn() could propagate isKey by default if column name is unchanged (or use a list of key names to set isKey).

cnathe:
I looked into this more and I don't know these code paths well enough to feel confident in knowing where the right touch points are.

PR #2130, opened 2021-03-26
Item 8638: Aliquot grid and lineage update

XingY:
#### Rationale
A "isAliquot" column is now added to the samples grid to provide indication if a sample is an aliquot. Additionally, modifications to lineage query and graph are done to better support aliquot vs derivative differentiation. 

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2130
* https://github.com/LabKey/labkey-api-js/pull/92
* https://github.com/LabKey/labkey-ui-components/pull/485
* https://github.com/LabKey/inventory/pull/212
* https://github.com/LabKey/sampleManagement/pull/529

#### Changes
* add "IsAliquot" query column to samples grid
* add "materialLineageType" to lineage node response
* support lineage filter option by MaterialRunType
* fix broken lineage query when a seed has no child


api/src/org/labkey/api/exp/api/ExpLineage.java
```diff
@@ -382,6 +383,16 @@ private JSONObject nodeToJSON(Identifiable node, User user, JSONArray parents, J
         {
             json = ExperimentJSONConverter.serialize(node, user, settings);
 
+            if (node instanceof ExpMaterial)
```
labkey-kevink: Can you add this to ExperimentJSONConverter instead?  I'd like to keep all the serialization there if possible.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -2478,6 +2480,15 @@ public SQLFragment generateExperimentTreeSQL(SQLFragment lsidsFrag, ExpLineageOp
                     parents.add(options.getCpasType());
                 }
 
+                if (excludeAliquotation || excludeDerivation)
```
labkey-kevink: I'd prefer not making this limited to these two types of runs.  We've had requests to filter the lineage by assay run protocol.

experiment/src/org/labkey/experiment/api/ExperimentRunGraph2.jsp
```diff
@@ -36,6 +36,10 @@
     var CONCAT = HtmlString.unsafe(dialect.isPostgreSQL() ? "||" : "+");
 
     assert "ALL".equals(expType) || "Data".equals(expType) || "Material".equals(expType) || "ExperimentRun".equals(expType) || "Object".equals(expType);
+
+  String materialRunType = StringUtils.defaultString(bean.getMaterialRunType(), "ALL");
```
labkey-kevink: See other comment about including/excluding other protocols by LSID.

labkey-kevink (changes requested):
Let's support filtering by other protocol LSIDs as well.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -149,6 +151,17 @@ public MutableColumnInfo createColumn(String alias, Column column)
                 return wrapColumn(alias, _rootTable.getColumn("RootMaterialLSID"));
             case AliquotedFromLSID:
                 return wrapColumn(alias, _rootTable.getColumn("AliquotedFromLSID"));
+            case IsAliquot:
+            {
+                ExprColumn columnInfo = new ExprColumn(this, FieldKey.fromParts("IsAliquot"), new SQLFragment(
+                        "(CASE WHEN RootMaterialLSID IS NULL THEN ? ELSE ? END)").add(false).add(true), JdbcType.BOOLEAN);
```
labkey-susanh, line 157: Is the parameterization here necessary since the values being used are constants?

api/src/org/labkey/api/exp/api/ExpLineage.java
```diff
@@ -20,6 +20,7 @@
 import org.labkey.api.exp.Identifiable;
 import org.labkey.api.security.User;
 import org.labkey.api.util.Pair;
+import org.springframework.util.StringUtils;
```
labkey-susanh: Is this import used?

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -469,11 +469,17 @@ else if (parts[0].equalsIgnoreCase(ExpData.DATA_INPUT_PARENT))
...
                     {
-                        if (isMerge)
+                        if (isMerge && !isAliquot)
                             parentDataTypesToRemove.add(namePart);
                     }
                     else
                     {
+                        if (isAliquot)
+                        {
+                            String message = "Data input is not allowed for aliquots";
```
labkey-susanh: Is this the message users see in the UI?  I wonder if we can be more informative so they will know which field in their data is problematic by including the name of the column. Something like "${parentColName} is not allowed for aliquots."  ("Data input" just sounds like we don't allow any kind of data.)

labkey-susanh (approved):
Just one comment on error message wording.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
XingY: turns out it's best to use parameterization due to sql dialect handling. Sql uses bit but postgres uses boolean type.

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -469,11 +469,17 @@ else if (parts[0].equalsIgnoreCase(ExpData.DATA_INPUT_PARENT))
...
                     {
-                        if (isMerge)
+                        if (isMerge && !isAliquot)
                             parentDataTypesToRemove.add(namePart);
                     }
                     else
                     {
+                        if (isAliquot)
+                        {
+                            String message = parentColName + " is not allowed for aliquots";
```
Sigmonia, line 479: if the parentColName is aliased in the import via a parent alias this might not match up to the user's import name. 
```suggestion
                            String message = "Sample derivation inputs are not allowed for aliquots.";
```

Or maybe a string constant to match with line 396.

XingY:
> Let's support filtering by other protocol LSIDs as well.

@kevinkrouse Kevin, can you please review this PR again so merge can be unblocked? Thanks.

labkey-kevink (approved):

PR #2131, opened 2021-03-26
Add support for generated Unique ID fields in Sample Types

labkey-susanh:
#### Rationale
For Sample Types, we want to give users the ability to add one or more LabKey-controlled UniqueID fields. The values in these fields are meant to be globally (within a given database) distinct across a server.  They are not user-editable.  

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/486

#### Changes
* Add a new concept URI for the UniqueID field type
* Add method to SimpleTranslator that adds columns for the UniqueID fields during dataIterator insertion
* Update DomainImpl to backfill UniqueID fields with values after added to an existing domain.
* Add metrics for counting the number of UniqueID fields on a server and the number of sample types with these fields


api/src/org/labkey/api/dataiterator/GenerateUniqueDataIterator.java
```diff
@@ -0,0 +1,234 @@
+package org.labkey.api.dataiterator;
```
labkey-susanh, line 1: @labkey-matthewb I didn't end up using this here.  Should we keep it around?

labkey-matthewb (approved):

api/src/org/labkey/api/data/BaseColumnInfo.java
```diff
@@ -597,6 +600,12 @@ public void setConceptURI(String conceptURI)
     {
         checkLocked();
         _conceptURI = conceptURI;
+        if (STORAGE_UNIQUE_ID_CONCEPT_URI.equals(_conceptURI))
```
cnathe: `ColumnInfo` below has a `isUniqueIdField`, can that be used here or put somewhere that it can be shared?

experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
```diff
@@ -663,8 +665,15 @@ public SQLFragment getValueSql(String tableAlias)
                     {
                         return super.getValueSql(tableAlias);
                     }
+
                 };
                 to.setHidden(from.isHidden());
+                if (STORAGE_UNIQUE_ID_CONCEPT_URI.equals(from.getConceptURI()))
```
cnathe: similar comment here, is there an `is unique concept` helper somewhere that can be reached here? Maybe all cases could use a single static method of `BaseColumnInfo` which takes in a conceptURI param to check?

PR #2134, opened 2021-03-29
Item 8703: Refactor usages of EntityInsertPanel to pass in fileImportParameters instead of having to define the import file handler

cnathe:
#### Rationale
The test team has been adding test cases that use the core-components page for general component testing. Chris pointed out that the file import tab of the sample create/import (i.e. EntityInsertpanel) was not working as expected. This related PR adds a callback handler for the file import case to actually importData into the shared components code to be used by the EntityInsertpanel.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/490
* https://github.com/LabKey/platform/pull/2134
* https://github.com/LabKey/sampleManagement/pull/533
* https://github.com/LabKey/biologics/pull/843

#### Changes
* Update @labkey/components version to pick up EntityInsertpanel changes for handling file import


core/src/client/LabKeyUIComponentsPage/SampleInsertPage.tsx
```diff
@@ -16,6 +25,35 @@ export const SampleInsertPage: FC = memo(() => {
...
+                        reject({msg: response.errors._form})
+                    }
+                }).catch((error) => {
+                    console.error(error);
+                    reject({msg: error.exception})
+                });
+            });
+        }, []
+    );
+
```
labkey-susanh: As discussed, let's move this functionality into the EntityInsertPanel since it's very much the same in all usages currently.

labkey-susanh (changes requested):

labkey-susanh (approved):
Yay!  Thanks for doing this refactor.

PR #2140, opened 2021-03-30
Remove default sample type upgrade code

labkey-adam:
#### Rationale
Earliest upgrade is now 19.1, which means the `deleteDefaultSampleType()` code that was added in 19.1 no longer needs to run. At this point, no installation (whether new or upgraded) will have a default sample type.

### DIFF
--- a/experiment/src/org/labkey/experiment/ExperimentModule.java
+++ b/experiment/src/org/labkey/experiment/ExperimentModule.java
@@ -279,9 +279,6 @@ public class ExperimentModule extends SpringModule implements SearchService.Docu
     @Override
     protected void startupAfterSpringConfig(ModuleContext moduleContext)
     {
-        // delete the default "Unspecified" SampleType TODO: move to an upgrade script in 19.2
-        SampleTypeServiceImpl.get().deleteDefaultSampleType();
-
         // TODO move to an upgrade script
         ExperimentUpgradeCode.upgradeMaterialSource(null);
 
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -279,9 +279,6 @@ public WebPartView getWebPartView(@NotNull ViewContext portalCtx, @NotNull Porta
     @Override
     protected void startupAfterSpringConfig(ModuleContext moduleContext)
     {
-        // delete the default "Unspecified" SampleType TODO: move to an upgrade script in 19.2
-        SampleTypeServiceImpl.get().deleteDefaultSampleType();
-
         // TODO move to an upgrade script
         ExperimentUpgradeCode.upgradeMaterialSource(null);
```
labkey-kevink, line 283: I know it's not a part of this change, but can we move `upgradeMaterialSource` into an upgrade script?  Or at least add a note about when we can remove it?  I think for 22.3?

labkey-kevink (changes requested):
There are a few usages of `SampleTypeService.getDefaultSampleTypeLsid` that should be removed -- especially the `ExperimentServiceImpl.createExpMaterials` one that could save new samples into the default sample type.

experiment/src/org/labkey/experiment/ExperimentModule.java
labkey-adam: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40714

labkey-adam:
Oops... I didn't see your other comment, @labkey-kevink  -- I'll need to look at that

labkey-kevink (approved):
As we discussed, we'll remove the other usages of the getDefaultSampleTypeMaterialLsidPrefix in a separate PR.

PR #2141, opened 2021-03-30
Sample Type Properties Panel Auto-Link Fix

RosalineP:
#### Rationale
Within the 'Sample Type Properties Panel,' we want the value for 'Auto-Link Target Container' to be a folder path, consistent with the Sample Type Designer 'Auto-Link Data to Study' dropdown value, rather than a Container ID.

#### Related Pull Requests
- https://github.com/LabKey/sampleManagement/pull/527
- https://github.com/LabKey/biologics/pull/840

#### Changes
* Set `autoLinkTargetContainer` column to invisible
* Add `SimpleDisplayColumn` of relevant values for Sample Type Properties Panel

### DIFF
--- a/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
+++ b/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
@@ -609,7 +609,7 @@ public class ExperimentController extends SpringActionController
                 autoLinkTargetColumn.setVisible(false);
 
                 SimpleDisplayColumn displayAutoLinkTargetColumn = new SimpleDisplayColumn();
-                displayAutoLinkTargetColumn.setCaption("Auto Link Target Container:");
+                displayAutoLinkTargetColumn.setCaption("Auto-Link Target Container:");
                 String path = autoLinkContainer.getPath();
                 displayAutoLinkTargetColumn.setDisplayHtml(path.equals("/") ? "" : path);
                 detailsView.getDataRegion().addDisplayColumn(displayAutoLinkTargetColumn);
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -602,6 +602,19 @@ public ModelAndView getView(ExpObjectForm form, BindException errors)
...
+            if (null != autoLinkContainer)
+            {
+                DisplayColumn autoLinkTargetColumn = detailsView.getDataRegion().getDisplayColumn("autoLinkTargetContainer");
+                autoLinkTargetColumn.setVisible(false);
+
+                SimpleDisplayColumn displayAutoLinkTargetColumn = new SimpleDisplayColumn();
+                displayAutoLinkTargetColumn.setCaption("Auto-Link Target Container:");
+                String path = autoLinkContainer.getPath();
+                displayAutoLinkTargetColumn.setDisplayHtml(path.equals("/") ? "" : path);
+                detailsView.getDataRegion().addDisplayColumn(displayAutoLinkTargetColumn);
```
RosalineP, line 615: Wasn't totally sure if this (hiding the DisplayColumn and creating a SimpleDisplayColumn) was the correct approach, but I didn't see a `setDisplayHtml` on the DisplayColumn.

labkey-klum: I think this approach works, although more work than I would have expected. Another approach would be to use a `ContainerDisplayColumn` to handle the rendering but I think you would still have the problem with the empty '/' path.

I think optimally the proper display column for that field would be on the source table but it looks like we are using the `MaterialSource` hard table instead of the virtual table.

labkey-klum (approved):

PR #2146, opened 2021-03-31
Issue 42822: Merging, alternative keyed, detailed logging ETLs fail on resolving PK

labkey-jeckels:
#### Rationale
Improvements to how we do detailed audit logging broke an ETL scenario where we're merging based on a column that's not the target table's PK

#### Related Pull Requests
* https://github.com/LabKey/dataintegration/pull/99

#### Changes
* Propagate alternateKeys early enough that the DataIteratorContext can be configured properly for the loadRows() call
* Use alternateKeys for finding rows in ExistingRecordDataIterator
* Leverage QueryUpdateService setup for DataIteratorContext in more places
* Use loadRows() for merge ETLs instead of mergeRows()

### DIFF
--- a/api/src/org/labkey/api/dataiterator/CachingDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/CachingDataIterator.java
@@ -158,6 +158,17 @@ public class CachingDataIterator extends AbstractDataIterator implements Scrolla
     @Override
     public Object get(int i)
     {
+        if (_currentRowArray == null)
+        {
+            // No row available
+            if (i == 0)
+            {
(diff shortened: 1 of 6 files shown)


labkey-martyp:
Awesome, so looks like we should be good to go with auditing merges with alternate keys again?  Without custom QUS. 

Looks good to me but I'll let Matt review since it's his recent work.

labkey-jeckels:
@labkey-martyp yes. I only tested the TNPRC scenario that broke, but I'm optimistic it will fix others as well.

labkey-matthewb (approved):

PR #2158, opened 2021-04-02
Remove buttons that navigate to root folder file browsing

labkey-jeckels:
#### Rationale
Get rid of dead code, avoid showing buttons that take users to bad places, and prevent lint warnings

#### Changes
* Shuffle pipeline button bar to group related buttons more closely
* Better generics for many PipelineService and other API classes
* Delete non-JMS job queue status page now that we show the queue position in the grid view
* Delete legacy pipeline job details page


labkey-kevink (approved):
Excellent

PR #2159, opened 2021-04-03
21.3 fb issue 42758

labkey-klum:
#### Rationale
Sample type and data class export can be slow even if the number of referenced samples or data is small. This is because we take all of the sample derivation runs and check to see if any of the material or data inputs/outputs match those that are selected. If the folder has a large number of derivation runs, this can be expensive since the runs need to be fully populated to do these checks.

#### Changes
Reverse the logic and use queries to determine the list of runs which are using the specified samples or datas.

### DIFF
--- a/experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
+++ b/experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
@@ -437,7 +437,7 @@ public class SampleTypeAndDataClassFolderWriter extends BaseFolderWriter
     }
 
     /**
-     * Generate a query to get the runIds where the supplied set of ExpData's were used as inputs
+     * Generate a query to get the runIds where the supplied set of ExpData's were used as inputs or outputs
      * CONSIDER : moving this to ExperimentService on trunk
      */
     private List<? extends ExpRun> getRunsForExpData(List<ExpData> datas)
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
```diff
@@ -450,6 +436,45 @@ public SampleTypeAliasColumnFactory(ColumnInfo aliasColumn)
         }
     }
 
+    /**
+     * Generate a query to get the runIds where the supplied set of ExpData's were used as inputs
+     * CONSIDER : moving this to ExperimentService on trunk
```
labkey-kevink: How is this method different from the `getRunsUsingDatas` method?

experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
```diff
@@ -450,6 +436,45 @@ public SampleTypeAliasColumnFactory(ColumnInfo aliasColumn)
         }
     }
 
+    /**
+     * Generate a query to get the runIds where the supplied set of ExpData's were used as inputs
```
labkey-kevink:
```suggestion
     * Generate a query to get the runIds where the supplied set of ExpData's were used as inputs or outputs
```

labkey-kevink (approved):

experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
labkey-klum: oh geez, I didn't notice that method! I thought there was only support for materials.

PR #2160, opened 2021-04-03
Defer initialization of data sources until first use

labkey-adam:
#### Rationale
Connecting to every data source proactively at init time slows down server startup and prevents some testing scenarios (e.g., LabKey connecting to itself via JDBC). Adding a level of indirection also prepares us for graceful handling of data sources coming and going.

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -305,7 +305,7 @@ public class DbScope
     {
         _dbScopeLoader = loader;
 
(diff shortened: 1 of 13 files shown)


labkey-jeckels (approved):
Is the behavior prior to this change that we attempt to connect at startup and if it fails, we don't try again? If so, looks good as-is. If not, we should add a release note.

labkey-adam:
Yes, connect once at startup and then never again is the current behavior

PR #2165, opened 2021-04-06
Issue 42857: name expression containing lookup fails to convert lookup string value

labkey-kevink:
#### Rationale
SampleTypes with a name expression containing a lookup such as ${CellLine/Name} failed to remap an incoming lookup string display value to the lookup primary key.  The fix is to pass the target column's ForeignKey into the call to addConvertColumn to make use of the alternate key remapping during import.

#### Changes
- In _SamplesCoerceDataIterator, pass the query column's foreign key into addConvertColumn to make use of remap
- consolidate three overloaded addConvertColumn cases into a single method
- add remap test case and javadoc


api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -779,48 +786,48 @@ public Object convert(Object value)
...
 
-        PropertyConvertColumn(String fieldName, int fromIndex, int mvIndex, PropertyDescriptor pd, PropertyType pt)
+        PropertyConvertColumn(String fieldName, int fromIndex, int mvIndex, boolean supportsMisisngValue, @Nullable PropertyType pt, @Nullable JdbcType type)
         {
-            super(fieldName, fromIndex, mvIndex, pt.getJdbcType());
-            this.pd = pd;
+            super(fieldName, fromIndex, mvIndex, type != null ? type : pt != null ? pt.getJdbcType() : null);
             this.pt = pt;
-            this.supportsMissingValue = pd.isMvEnabled();
+            this.supportsMissingValue = supportsMisisngValue;
```
labkey-matthewb:
```suggestion
            this.supportsMissingValue = supportsMissingValue;
```

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -779,48 +786,48 @@ public Object convert(Object value)
...
 
 
     private class PropertyConvertColumn extends MissingValueConvertColumn
     {
-        PropertyDescriptor pd;
-        PropertyType pt;
+        @Nullable PropertyType pt;
 
-        PropertyConvertColumn(String fieldName, int fromIndex, int mvIndex, PropertyDescriptor pd, PropertyType pt)
+        PropertyConvertColumn(String fieldName, int fromIndex, int mvIndex, boolean supportsMisisngValue, @Nullable PropertyType pt, @Nullable JdbcType type)
```
labkey-matthewb:
```suggestion
        PropertyConvertColumn(String fieldName, int fromIndex, int mvIndex, boolean supportsMissingValue, @Nullable PropertyType pt, @Nullable JdbcType type)
```

api/src/org/labkey/api/dataiterator/CoerceDataIterator.java
```diff
@@ -70,7 +69,7 @@ void init(TableInfo target, boolean useImportAliases, boolean outputAllColumns)
                 else if (to.getFk() instanceof MultiValuedForeignKey)
                     addColumn(to.getName(), i); // pass-through multi-value columns -- converting will stringify a collection
                 else
-                    addConvertColumn(to.getName(), i, to.getJdbcType(), false);
+                    addConvertColumn(to.getName(), i, to.getJdbcType(), to.getFk(), true);
```
labkey-matthewb, line 72: What does this change (mv=true) do?

labkey-matthewb: Shouldn't this depend on whether the target columnn supports MV?

labkey-kevink: The last parameter is now `useOriginalValueOnRemapFailure` instead of `mv`.  When the lookup to the alternate key value fails, we don't want to throw an error just yet to allow triggers the chance to fix the error.  The later StandardDataIterator will add convert columns with the `useOriginalValueOnRemapFailure` set to true.

labkey-matthewb (approved):

PR #2168, opened 2021-04-06
Issue 42885: Provide for parameter substitution in authentication configuration fields 

RosalineP:
#### Rationale
The motivating change desired is to allow the ACS URL setting within SAML auth config modals to dynamically update the hardcoded url '/labkey/saml-validate.view?configuration=configurationId'. We want to replace 'configurationId' with the id value saved on the SAML auth config in question.

We opted for a more general and reusable approach that allows for customization of <FixedHtml/>s through parameter substitution. Upon detecting the parameter substitution pattern ('${someVariable}') we attempt to replace it with the passed dynamic variable value 'someVariable' existing on 'authConfig'. If we cannot, we replace the pattern with the text 'someVariable' itself.

#### Related Pull Requests
* https://github.com/LabKey/saml/pull/39

#### Changes
* Lint
* Supply 'authConfig' prop to <DynamicFields/>
* Support <FixedHtml/> parameter substitution
* Update snapshot


labkey-adam:
Probably best for Alan to review... seems like some changes aren't directly related to parameter substitution, but probably fine changes to take.

core/src/client/components/DynamicFields.tsx
```diff
@@ -121,20 +131,41 @@ interface FixedHtmlProps {
...
+        let stringTemplatedHtml = html;
+        const match = /\${(.*?)}/;
+        const matchResults = match.exec(html);
+        if (matchResults) {
+            const matchedTemplate = matchResults[0];
+            const matchedString = matchResults[1];
+            const replacementVar = authConfig[matchedString];
+            stringTemplatedHtml = html.replace(
+                matchedTemplate,
+                replacementVar ? replacementVar.toString() : matchedString
```
labkey-adam: HTML encode replacement value?

RosalineP: Addressed! 
Note: JS built-in `encodeURI ()` encodes all special characters except `, / ? : @ & = + $ #`

core/src/client/components/DynamicFields.tsx
```diff
@@ -121,20 +131,32 @@ interface FixedHtmlProps {
...
 export class FixedHtml extends PureComponent<FixedHtmlProps> {
     render() {
-        const { description, caption, html } = this.props;
+        const { description, caption, html, authConfig } = this.props;
+
+        // Issue42885: Upon detecting the parameter substitution pattern ('${someVariable}') we attempt to replace it
+        // with the passed dynamic variable value existing on 'authConfig'. If we cannot, we replace the pattern with
+        // the text 'someVariable' itself.
+        const stringTemplatedHtml = html.replace(/\${(.*?)}/g, (match, value) => {
+            return encodeURI(authConfig[value] ? authConfig[value].toString() : value);
```
labkey-adam: encodeURI is not HTML encoding. One extreme test: give your configuration a description like `SAML Configuration <script>alert("hello");</script>` and include ${description} in a FixedHtml field. I didn't see an alert, but the <script> tag and everything after it are being output into the page without any HTML encoding. Instead, you should see all the text, just like you do on the main auth config page.

core/src/client/components/DynamicFields.tsx
```diff
@@ -121,20 +131,32 @@ interface FixedHtmlProps {
...
 
 export class FixedHtml extends PureComponent<FixedHtmlProps> {
     render() {
-        const { description, caption, html } = this.props;
+        const { description, caption, html, authConfig } = this.props;
+
+        // Issue42885: Upon detecting the parameter substitution pattern ('${someVariable}') we attempt to replace it
+        // with the passed dynamic variable value existing on 'authConfig'. If we cannot, we replace the pattern with
+        // the text 'someVariable' itself.
+        const stringTemplatedHtml = html.replace(/\${(.*?)}/g, (match, value) => {
```
labkey-adam, line 145: Did you test this with a bunch of different replacements? For me, it's still only working on the first one. Perhaps you need `replaceAll()`?

labkey-adam (changes requested):

core/src/client/components/DynamicFields.tsx
RosalineP: (Adam reported as resolved)

core/src/client/components/DynamicFields.tsx
RosalineP: Found and used  LK JS API fn `encodeHtml()` 👍

core/src/client/components/DynamicFields.tsx
```diff
@@ -121,20 +132,32 @@ interface FixedHtmlProps {
...
+            return Utils.encodeHtml(authConfig[value] ? authConfig[value].toString() : value);
```
cnathe: Is it possible for `authConfig[value]` to be zero here? If so, that would go to the else block. Do you instead want the check to be `authConfig[value] !== undefined && authConfig[value] !== null ? authConfig[value].toString() : value`?

cnathe (approved):

core/src/client/components/DynamicFields.tsx
RosalineP: Done!

labkey-adam (approved):

labkey-adam:
I've performed manual testing locally... good to go!

PR #2170, opened 2021-04-08
Link to Study Actions for Sample Type

RosalineP:
#### Rationale
For Sample Type designer on LKS, allow users to link Sample Type rows to studies. 

#### Related Pull Requests
* n/a

#### Changes
* Rosaline
    * Move files related to publishing from `platform/study/.../publish` to `platform/api/.../publish`
    * Support hidden form field parameter 'returnUrl' (within `SampleTypeContentsView.createDataView()`), allowing users to hit 'cancel' to return from link-to-study pages to previous page
    * Implement `SampleTypePublishStartAction.java`, which 
        * displays page prompting user to choose a target study
        * posts information such as rows selected from sample type, selected target study, and original sample type to `SampleTypePublishConfirmAction`
    * Implement SampleTypePublishConfirmAction.java, which
        * Displays page displaying copied sample set row information, with with auto-populated or blank form fields for aligning Participant and Visit/Timepoint data. Validation errors are displayed on this page, as well as an option to select a subset of rows to copy
        * On button click, executes Link to Study action, which redirects to created or existing Sample Type Dataset
    * Set link within Sample Set webpart view's 'Link to Study' button to land on `experiment-sampleTypePublishStart.view` page (`SampleTypeWebPart.java`)


experiment/src/org/labkey/experiment/controllers/exp/SampleTypeContentsView.java
```diff
@@ -52,9 +55,37 @@ public static ActionButton getDeriveSamplesButton(@NotNull Container container,
...
+            else
+                returnURL = getViewContext().getActionURL().toString();
+        }
+
+        view.getDataRegion().addHiddenFormField(ActionURL.Param.returnUrl, returnURL);
+
+        return view;
+    }
+
+    public static ActionButton getLinkToStudyButton(DataView view, int rowId, Container container)
```
labkey-klum: this doesn't need to be static since it's only used in this class, I would also make it private.

experiment/src/org/labkey/experiment/controllers/exp/SampleTypeContentsView.java
```diff
@@ -52,9 +55,37 @@ public static ActionButton getDeriveSamplesButton(@NotNull Container container,
...
     {
-        ActionURL linkToStudyURL = new ActionURL(); // Rosaline: TODO in LinkToStudyAction story
+        ActionURL linkToStudyURL = PageFlowUtil.urlProvider(ExperimentUrls.class).getLinkToStudyURL(container);
+        view.getDataRegion().addHiddenFormField("rowId", String.valueOf(rowId));
```
labkey-klum: I would add the hidden form field in `createDataView` just to consolidate actions on the data view into one place. Since at the time it's called we already know the source sample type.

experiment/src/org/labkey/experiment/controllers/exp/SampleTypeContentsView.java
```diff
@@ -52,9 +55,37 @@ public static ActionButton getDeriveSamplesButton(@NotNull Container container,
...
+
+        linkToStudyURL.deleteParameters();
```
labkey-klum: no need to delete parameters since you just created the URL above with no params

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
```diff
@@ -0,0 +1,299 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import static org.labkey.api.util.PageFlowUtil.urlProvider;
+
+@RequiresPermission(InsertPermission.class)
+public class SampleTypePublishConfirmAction extends AbstractPublishConfirmAction<SampleTypePublishConfirmAction.SampleTypePublishConfirmForm>
+{
+    private SamplesSchema _sampleTypeSchema;
+    private ExpSampleType _sampleType;
```
labkey-klum, line 54: what's the difference between `_sampleType` on the action and the one on the form? It's probably better to have one version of this rather than trying to pass it back and forth (maybe just initialize the action scoped variable in `initializeFieldKeys`)

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
```diff
@@ -0,0 +1,299 @@
...
+
+    /**
+     * Names need to match those found in PublishConfirmForm.DefaultValueSource
+     */
+    public enum DefaultValueSource
+    {
+        PublishSource
+                {
+                    @Override
+                    public FieldKey getParticipantIDFieldKey(FieldKey participantId)
```
labkey-klum, line 97: I don't think we need to pass in a parameter for these methods, just return either null or the saved member variable. It looks like from your implementation you are just returning the same value you pass in.

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    private void initializeFieldKeys(ExpSampleType sampleType)
+    {
+        if (null == _participantId || (null == _visitId && null == _date))
+        {
+            String visitURI = PropertyType.VISIT_CONCEPT_URI;
+            String particpantURI = PropertyType.PARTICIPANT_CONCEPT_URI;
+
+            UserSchema userSchema = QueryService.get().getUserSchema(getUser(), getContainer(), "exp.materials");
```
labkey-klum: don't use exp.materials, use SamplesSchema.SCHEMA_NAME instead, exp.materials may work but it's not the official name of the sample type schema.

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
```diff
@@ -0,0 +1,299 @@
...
+
+            for (ColumnInfo ci : columnInfoList)
+            {
+                String columnURI = ci.getConceptURI();
+                String columnName = ci.getName();
+                FieldKey fieldKey = ci.getFieldKey();
+
+                if (null != columnURI && columnURI.equals(visitURI))
+                {
+                    if (PropertyType.xsdDouble.getURI().equals(ci.getRangeURI()))
```
labkey-klum: you can accomplish this much easier by using: `ci.getJdbcType().isReal()` or `ci.getJdbcType().isDateOrTime()`

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
```diff
@@ -0,0 +1,299 @@
...
+
+            UserSchema userSchema = QueryService.get().getUserSchema(getUser(), getContainer(), "exp.materials");
+            TableInfo tableInfo = userSchema.getTable(sampleType.getName());
+            if (tableInfo == null)
+                throw new IllegalStateException(String.format("Sample Type %s not found", sampleType.getName()));
+            List<ColumnInfo> columnInfoList = tableInfo.getColumns();
+
+            for (ColumnInfo ci : columnInfoList)
+            {
+                String columnURI = ci.getConceptURI();
```
labkey-klum, line 143: This loop is why you get the last column annotated with a particular concept URI, since you are not checking if is non null (like you do the the fall back method of checking on column names).

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
```diff
@@ -0,0 +1,299 @@
...
+    {
+        initializeFieldKeys(form.getSampleType());
+        return super.handlePost(form, errors);
+    }
+
+    @Override
+    protected SamplesSchema getUserSchema(SampleTypePublishConfirmForm form)
+    {
+        _sampleTypeSchema = (SamplesSchema) QueryService.get().getUserSchema(form.getUser(), getContainer(), SamplesSchema.SCHEMA_SAMPLES);
+        return _sampleTypeSchema;
```
labkey-klum: does this need to be a class variable, it doesn't appear that we access it elsewhere

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
```diff
@@ -0,0 +1,299 @@
...
+
+    @Override
+    protected QuerySettings getQuerySettings(SampleTypePublishConfirmForm form)
+    {
+        String queryName = form.getSampleType().getName();
+        QuerySettings qs = new QuerySettings(form.getViewContext(), QueryView.DATAREGIONNAME_DEFAULT, queryName);
+        qs.setSchemaName(SamplesSchema.SCHEMA_NAME);
+        qs.setQueryName(queryName);
+
+        getUserSchema(form);
```
labkey-klum: does this call need to happen?

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
```diff
@@ -0,0 +1,299 @@
...
+        for (PublishKey publishKey : dataKeys.values())
+        {
+            Map<String, Object> dataMap = new HashMap<>();
+
+            Container targetStudyContainer = targetStudy;
+            if (publishKey.getTargetStudy() != null)
+                targetStudyContainer = publishKey.getTargetStudy();
+            assert targetStudyContainer != null;
+
+            String sourceLSID = new Lsid("Data", String.valueOf(publishKey.getDataId())).toString();
```
labkey-klum: We can't make up an Lsid here, the sourceLSID is supposed to represent the LSID of the object that the published rows originated from. For assays, this would be the run, but for samples it should be the sample type LSID. Probably should use : `_sampleType.getLSID()`

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
```diff
@@ -0,0 +1,299 @@
...
+        ExpSampleType sampleType = form._sampleType;
+        StudyPublishService.get().checkForAlreadyCopiedRows(getUser(), Pair.of(Dataset.PublishSource.SampleType, sampleType.getRowId()), errors, rowIdsByTargetContainer);
+        if (!errors.isEmpty())
+        {
+            return null;
+        }
+
+        return StudyPublishService.get().publishData(getUser(), form.getContainer(), targetStudy, sampleType.getName(),
+                Pair.of(Dataset.PublishSource.SampleType, sampleType.getRowId()),
+                dataMaps, "RowId", errors);
```
labkey-klum: use the rowid constant you defined

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    @Override
+    public ModelAndView getView(SampleTypePublishConfirmForm form, boolean reshow, BindException errors) throws Exception
+    {
+        if (form.getDefaultValueSource().equals("Assay"))
+            form.setDefaultValueSource("SampleType");
```
labkey-klum: We shouldn't be touching the default value source, the view will default and initialize to `PublishSource` these values you have don't map to valid enums.

experiment/src/org/labkey/experiment/publish/SampleTypePublishStartAction.java
```diff
@@ -0,0 +1,143 @@
...
+
+    public void setSampleType(ExpSampleType sampleType)
+    {
+        _sampleType = sampleType;
+    }
+
+    @Override
+    protected ActionURL getSuccessUrl(SampleTypePublishStartForm form)
+    {
+        ActionURL url = urlProvider(ExperimentUrls.class).getLinkToStudyConfirmURL(getContainer(), _sampleType);
```
labkey-klum: don't need to declare a variable

experiment/src/org/labkey/experiment/controllers/exp/SampleTypeContentsView.java
RosalineP: Done!

experiment/src/org/labkey/experiment/controllers/exp/SampleTypeContentsView.java
RosalineP: Done!

experiment/src/org/labkey/experiment/controllers/exp/SampleTypeContentsView.java
RosalineP: Done!

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
RosalineP: `SampleTypePublishStartForm`'s `_sampleTypeIds` is determining whether or not we're coming from selecting an entire sample type (`true`) or just selecting sample type rows (`false`)
There's a note on `getDataIDs()` now around checking `form.isSampleTypeIds()`. In a later story, if it is an entire sample type, we'd update `getDataIDs()` to populate `List<Integer> _sampleTypeIds` on the action accordingly

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
RosalineP: Done!

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
RosalineP: Nope!

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
RosalineP: Nope, removed!

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
RosalineP: (Discussed, add comment)

RosalineP: Done!

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
RosalineP: Done!

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
RosalineP: Done!

experiment/src/org/labkey/experiment/publish/SampleTypePublishStartAction.java
RosalineP: Done!

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
RosalineP: Done!

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
RosalineP: Removed stale lines!

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
RosalineP: I'm not sure if I'm parsing correctly—is this resolved by the change of using `ci.getJdbcType()`? Not sure on what page 'last column' is referring to here.

(The reason why checks like `_visitId == null` weren't included on setting `_visitId`, etc is because if a column named 'participantid' is set, and then a column with a type of `PARTICIPANT_CONCEPT_URI` is set, I overwrite it with the latter column so I can do my processing in one pass)

api/src/org/labkey/api/study/publish/PublishConfirmForm.java
```diff
@@ -16,11 +16,14 @@
...
+     * Controls the source of the subject and timepoint values for the publish results query view
+     */
     public enum DefaultValueSource
     {
-        Assay,
-        Specimen,
-        UserSpecified,
+        PublishSource,  // the publish source results table
+        Specimen,       // for assays only, specimen linked to assay results
+        UserSpecified,  // user specified from the confirm view
```
RosalineP: Extremely minor: these three comments read like a contiguous sentence to me at first. I'd recommend sentence-casing the first word to avoid confusion for future readers

study/src/org/labkey/study/assay/StudyPublishManager.java
```diff
@@ -1123,4 +1135,106 @@ public ExpProtocol ensureStudyPublishProtocol(User user, Container container, St
...
+                if (protocol != null)
+                    event.setProtocol(protocol.getRowId());
+                event.setTargetStudy(def.getStudy().getContainer().getId());
+                event.setDatasetId(def.getDatasetId());
+
+                AuditLogService.get().addEvent(user, event);
+            }
+
+            case SampleType -> {
+
```
RosalineP: Maybe add intermediary comment indicating this is to be completed in another story?

api/src/org/labkey/api/assay/AbstractAssayProvider.java
```diff
@@ -1518,7 +1518,8 @@ public AssayRunAsyncContext createRunAsyncContext(AssayRunUploadContext context)
         StudyPublishService svc = StudyPublishService.get();
 
         if (AuditLogService.get().isViewable() && null != svc)
-            result.add(new NavTree("view copy-to-study history", svc.getPublishHistory(viewContext.getContainer(), protocol, containerFilter)));
+            result.add(new NavTree("view link-to-study history", svc.getPublishHistory(viewContext.getContainer(),
```
RosalineP: Just checking, the hyphenated `link-to-study` convention is used just for the history-related text, in order to group the wording sets better, right? Since it seems to be consistent I've updated a string accordingly in 5724733. Let me know though if we want to get rid of the hyphen usage altogether if it's not what we want to go with anymore

study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -2696,14 +2696,14 @@ protected void populateButtonBar(DataView view, ButtonBar bar)
                 }
             }
             else
-                view.addView(new HtmlView("The Dataset does not exist."));
+                view.addView(new HtmlView(HtmlString.unsafe("The Dataset does not exist.")));
```
RosalineP: Why not use `of` instead of `unsafe` here?

study/src/org/labkey/study/query/DatasetQueryView.java
```diff
@@ -434,8 +433,8 @@ protected void populateButtonBar(DataView view, ButtonBar bar)
             {
                 if (publishSource != null)
                 {
-                    ExpProtocol protocol = (ExpProtocol)publishSource;
-//                  do we ever add the assay import buttons to the dataset query view?
+//                    ExpProtocol protocol = (ExpProtocol)publishSource;
+//                    do we ever add the assay import buttons to the dataset query view?
```
RosalineP: Move text comment above lines of (newly commented-out) code

RosalineP (commented):
Had some small comments, but saw nothing blocking! Would approve if it wasn't my own PR.

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
labkey-klum: I think that makes sense for that scenario you described but if you have more than one field annotated as `PARTICIPANT_CONCEPT_URI` then you will always get the last one annotated which is somewhat different from how we resolve conflicts. You might need two loops, one to find the concept URI and another to fall back on the special names if nothing has been found yet.

RosalineP: Change pushed!

experiment/src/org/labkey/experiment/publish/SampleTypePublishConfirmAction.java
```diff
@@ -0,0 +1,312 @@
...
+        for (PublishKey publishKey : dataKeys.values())
+        {
+            Map<String, Object> dataMap = new HashMap<>();
+
+            Container targetStudyContainer = targetStudy;
+            if (publishKey.getTargetStudy() != null)
+                targetStudyContainer = publishKey.getTargetStudy();
+            assert targetStudyContainer != null;
+
+            String sourceLSID = new Lsid("Data", _sampleType.getLSID()).toString();
```
labkey-klum: I think we literally want to use the sample type lsid, the code you have will mutate the lsid namespace and we wont be able to identify the original source.

```suggestion
            String sourceLSID = sampleType.getLSID();
```

RosalineP: Use sample type LSID

api/src/org/labkey/api/study/publish/PublishConfirmForm.java
labkey-klum: done

study/src/org/labkey/study/assay/StudyPublishManager.java
labkey-klum: done.

api/src/org/labkey/api/assay/AbstractAssayProvider.java
labkey-klum: I agree that the non-hyphenated version is more consistent, I've changed this here so it may make sense to change it on the sample details panel too.

study/src/org/labkey/study/controllers/StudyController.java
labkey-klum: good point, nothing to unsafe 👍

study/src/org/labkey/study/query/DatasetQueryView.java
labkey-klum: also changed this to a TODO.

labkey-klum (approved):

api/src/org/labkey/api/assay/AbstractAssayProvider.java
RosalineP: (As discussed all hyphenated versions will be changed to non-hyphenated in the nomenclature story)

PR #2172, opened 2021-04-10
Issue 42924: Stable URL fails to URL encode filter parameters

labkey-jeckels:
#### Rationale
Stable URLs are meant for easy sharing, but when we don't fully encode parameter it's much more likely they'll get mangled

#### Changes
* URL encode filter GET parameter names and values


labkey-kevink (approved):

labkey-martyp (approved):
Nice, thanks

PR #2173, opened 2021-04-12
Item 8735: Enable file/attachment field data type for Sample Type and Data Class (i.e. Sources)

cnathe:
#### Rationale
The SM app v1 explicitly hid the file/attachment column data type from the field editors in the app because we didn't properly display the file in the details view and edit details panel view. This PR fixes that to allow for files to be POSTed with the insertRows and updateRows command and handled in the QueryController actions.

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/94
* https://github.com/LabKey/labkey-ui-components/pull/498
* https://github.com/LabKey/platform/pull/2173
* https://github.com/LabKey/sampleManagement/pull/544
* https://github.com/LabKey/biologics/pull/853
* https://github.com/LabKey/inventory/pull/220

#### Changes
* Update core package.json @labkey/api and @labkey/components package versions
* Add support for insertRows / updateRows to include file attachments 
* Add metrics for counts of file and attachment column usages in sample type and data class domains


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -3957,6 +3974,31 @@ protected JSONObject executeJson(JSONObject json, CommandType commandType, boole
...
+        private void addRowAttachments(Map<String, Object> rowMap, int rowIndex)
+        {
+            if (getFileMap() != null)
+            {
+                for (Map.Entry<String, MultipartFile> fileEntry : getFileMap().entrySet())
+                {
+                    // allow for the fileMap key to include the row index for defining which row to attach this file to
+                    // ex: "FileField::0", "FieldField::1"
+                    String fieldKey = fileEntry.getKey();
+                    int delimIndex = fieldKey.indexOf(ROW_ATTACHMENT_INDEX_DELIM);
```
labkey-jeckels:
```suggestion
                    int delimIndex = fieldKey.lastIndexOf(ROW_ATTACHMENT_INDEX_DELIM);
```

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -3778,6 +3779,21 @@ else if ("tsv".equalsIgnoreCase(form.getFormat()))
         public static final String PROP_COMMAND = "command";
         private static final String PROP_ROWS = "rows";
 
+        private JSONObject _json;
+
+        @Override
+        public void validateForm(ApiSaveRowsForm apiSaveRowsForm, Errors errors)
+        {
+            _json = apiSaveRowsForm.getJsonObject();
+            if (_json == null)
```
labkey-jeckels: Please add a comment about when this if() will be useful and a null check for the parameter value itself. It looks like it would NPE in the JSONObject constructor if it's absent

labkey-jeckels (approved):
Two suggestions but no need for additional review

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -3957,6 +3974,31 @@ protected JSONObject executeJson(JSONObject json, CommandType commandType, boole
...
         }
 
+        private void addRowAttachments(Map<String, Object> rowMap, int rowIndex)
+        {
+            if (getFileMap() != null)
+            {
+                for (Map.Entry<String, MultipartFile> fileEntry : getFileMap().entrySet())
+                {
+                    // allow for the fileMap key to include the row index for defining which row to attach this file to
+                    // ex: "FileField::0", "FieldField::1"
```
labkey-kevink, line 3994: It probably won't collide with attachments, but we use the same syntax for the auto-generated pivot columns:  "Column::Value"

labkey-jeckels: Cory and I discussed that this afternoon. He thought it was unlikely to collide and I agreed.

query/src/org/labkey/query/controllers/QueryController.java
cnathe: Done

query/src/org/labkey/query/controllers/QueryController.java
cnathe: Done

PR #2174, opened 2021-04-13
Issue 42897: remove getDefaultSampleTypeMaterialLsidPrefix

labkey-kevink:
#### Rationale
The default "Unspecified" SampleType has been deprecated since 19.1, but the `ExperimentService.getExpMaterials()` method could create new samples using the default SampleType's LSID prefix.  This PR removes `getExpMaterials` method as well as the public `SampleTypeService` methods related to the default SampleType.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2140
* https://github.com/LabKey/commonAssays/pull/320

#### Changes
- remove `SampleTypeService.getDefaultSampleTypeLsid()` and `getDefaultSampleTypeMaterialLsidPrefix()`
- remove `ExperimentService.getExpMaterials()` and usages in favor of `findExpMaterial()`
- unresolved samples during matrix assay import will be created in a SampleType, if possible

### DIFF
--- a/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
+++ b/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
@@ -3440,12 +3440,6 @@ public class ExperimentController extends SpringActionController
         public ModelAndView getView(DeleteForm deleteForm, boolean reshow, BindException errors)
         {
             List<ExpSampleType> sampleTypes = getSampleTypes(deleteForm);
-            String defaultSampleType = SampleTypeServiceImpl.get().getDefaultSampleTypeLsid();
-            if (sampleTypes.stream().anyMatch(ss -> defaultSampleType.equals(ss.getLSID())))
-            {
-                throw new RedirectException(ExperimentUrlsImpl.get().getShowSampleTypeListURL(getContainer(), "You cannot delete the default sample type."));
-            }
-
             if (!ensureCorrectContainer(sampleTypes))
             {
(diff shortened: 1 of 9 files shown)


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -3440,7 +3440,7 @@ protected void deleteObjects(DeleteForm deleteForm)
         public ModelAndView getView(DeleteForm deleteForm, boolean reshow, BindException errors)
         {
             List<ExpSampleType> sampleTypes = getSampleTypes(deleteForm);
-            String defaultSampleType = SampleTypeService.get().getDefaultSampleTypeLsid();
+            String defaultSampleType = SampleTypeServiceImpl.get().getDefaultSampleTypeLsid();
             if (sampleTypes.stream().anyMatch(ss -> defaultSampleType.equals(ss.getLSID())))
```
labkey-jeckels: Is this still relevant? If so, why are we checking for a match when deleting but no longer checking for editing?

labkey-jeckels: and if we can delete it, does that mean we can delete the couple of deprecated items that are now private?

labkey-jeckels (approved):
A couple of questions but no further review required

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-kevink: I went back and forth on removing all of the checks.  Since any default SampleTypes have been deleted, I went ahead and removed the last bits.

PR #2177, opened 2021-04-13
Issue42527: Summary View 'isPrimaryKey' column 

RosalineP:
#### Rationale
Pull in SummaryView fix that displays the 'Is Primary Key' column only for List Domain Designers.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/501
* https://github.com/LabKey/sampleManagement/pull/547
* https://github.com/LabKey/biologics/pull/855

#### Changes
* Bump ui-components version


cnathe (approved):
Thank you for bumping all of them, but just an FYI for future updates, if your package changes only involve field editor changes then we can just bump the core/package.json (since we have consolidated all of the field editor entryPoints into that module for build perf reasons).

PR #2183, opened 2021-04-14
Skip stored procedure check when LabKey connects to LabKey via JDBC

labkey-adam:
#### Rationale
When connecting to LabKey via PostgreSQL JDBC, don't bother querying for stored procedures

#### Related Pull Requests
* https://github.com/LabKey/connectors/pull/74

### DIFF
--- a/api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
+++ b/api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
@@ -1904,4 +1904,11 @@ public abstract class PostgreSql91Dialect extends SqlDialect
         if (null != _adminWarning)
             warnings.add(_adminWarning);
     }
+
+    @Override
+    public boolean isProcedureExists(DbScope scope, String schema, String name)
+    {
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
```diff
@@ -1904,4 +1904,11 @@ public void addAdminWarningMessages(Warnings warnings)
         if (null != _adminWarning)
             warnings.add(_adminWarning);
     }
+
+    @Override
+    public boolean isProcedureExists(DbScope scope, String schema, String name)
```
labkey-jeckels, line 1909: isProcedureExists()? From the king of grammar and spelling fix commits?

labkey-jeckels (approved):

api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
labkey-adam: This one was courtesy Tony

PR #2186, opened 2021-04-14
Issue 42948: Remove use of inefficient HTMLContentExtractor

labkey-adam:
#### Rationale
Our full-text search indexer has special handling that parses HTML files using `HTMLContentExtractor`. This class extends the JDK's `HTMLEditorKit.ParserCallback`, which is very inefficient with large HTML files. Tika seems to have a perfectly fine HTML parser (which uses TagSoup), so we remove our special handling and just let Tika do its thing. See [Issue 42948: org.labkey.api.util.HTMLContentExtractor uses JDK's very inefficient HTML parser](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42948)

Also taking the opportunity to remove special handling for certain text files; this code doesn't have the problems of HTMLContentExtractor, but we might as well delegate to Tika here as well. I had to add a metadata hint for "content encoding = UTF-8" for text/plain parsing, otherwise, Tika sometimes guesses IBM500 (!!) for short text bodies. See https://issues.apache.org/jira/browse/TIKA-2771

### DIFF
--- a/api/src/org/labkey/api/util/HTMLContentExtractor.java
+++ b/api/src/org/labkey/api/util/HTMLContentExtractor.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.io.Reader;
 import java.io.StringReader;
 
+// Warning: This is not an efficient parser. See Issue 42948.
 public abstract class HTMLContentExtractor extends HTMLEditorKit.ParserCallback
 {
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
Probably worth leaving a comment behind in HTMLContentExtractor with a warning about its memory usage perhaps with a pointer to this bug.

PR #2189, opened 2021-04-15
Issue 42944: Avoid NPE when study service is not available.

labkey-susanh:
#### Rationale
[Issue 42944](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42944).  When exporting a LKSM folder within a distribution that does not include the study module, we get an NPE.

#### Changes
* Add NPE check


labkey-jeckels (approved):
I saw this NPE but I couldn't figure it out. I'm so used to the study module always being present. I guess the code is too.

PR #2191, opened 2021-04-15
42606: Assay results LSID required to attach provenance

labkey-klum:
#### Rationale
This is a fix for this issue: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42606

Details:
- The ProvenanceService was at some point refactored so that if a provider was not registered, a default noop implementation was returned
- Added a @NotNull annotation to the `get` method
- Removed all null checks for the service, since they are no longer possible
- In the cases where we need to know if the provenance module is present we use the `ServiceRegistry` instead so we can short circuit the code (this is the main bug fix).

### DIFF
--- a/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
+++ b/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
@@ -59,6 +59,7 @@ import org.labkey.api.reader.ColumnDescriptor;
 import org.labkey.api.reader.DataLoader;
 import org.labkey.api.reader.TabLoader;
 import org.labkey.api.security.User;
+import org.labkey.api.services.ServiceRegistry;
 import org.labkey.api.study.ParticipantVisit;
 import org.labkey.api.study.Study;
 import org.labkey.api.study.StudyService;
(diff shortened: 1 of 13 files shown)


api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
```diff
@@ -525,7 +523,7 @@ public void importRows(ExpData data, User user, ExpRun run, ExpProtocol protocol
 
     private void addAssayResultRowsProvenance(Container container, ExpRun run, List<Map<String, Object>> insertedData, Map<Integer, String> rowIdToLsid)
     {
-        ProvenanceService pvs = ProvenanceService.get();
+        ProvenanceService pvs = ServiceRegistry.get().getService(ProvenanceService.class);
```
labkey-kevink: Getting the service directly to short circuit feels a bit too subtle. Should we add a `ProvenanceService.isProvenanceSupported()` or something to make a distinction between the no-op implementation and the actual service?

labkey-kevink (approved):
Looks fine, but we may want to make it more obvious when the provenance implementation is no-op and can be skipped.

api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
labkey-klum: good idea

PR #2192, opened 2021-04-15
Issue 42937: Assay results grid loading performance can degrade with a large number of "copied to study" columns

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42937
Limit the default visible "copied to ..." study columns to 3, after that we don't add any to the default view and rely on users to customize their views themselves.

#### Changes
* AssayProtocolSchema.addCopiedToStudyColumns to limit default visibileColumnNames to 3

### DIFF
--- a/api/src/org/labkey/api/assay/AssayProtocolSchema.java
+++ b/api/src/org/labkey/api/assay/AssayProtocolSchema.java
@@ -874,9 +874,13 @@ public abstract class AssayProtocolSchema extends AssaySchema
 
                 table.addColumn(studyCopiedColumn);
 
-                visibleColumnNames.add(studyCopiedColumn.getName());
+                // Issue 42937: limit default visible columns to 3 for a given assay protocol
+                if (datasetIndex > 3)
+                    visibleColumnNames.clear();
+                else
+                    visibleColumnNames.add(studyCopiedColumn.getName());
             }
-            if (setVisibleColumns)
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/assay/AssayProtocolSchema.java
```diff
@@ -874,9 +874,13 @@ else if (!first)
 
                 table.addColumn(studyCopiedColumn);
 
-                visibleColumnNames.add(studyCopiedColumn.getName());
+                // Issue 42937: limit default visible columns to 3 for a given assay protocol
+                if (datasetIndex > 3)
```
labkey-jeckels, line 878: This approach is fine, but it might be clearer to set setVisibleColumns = false instead

labkey-jeckels (approved):
Suggested an alternate approach to consider. Your call on which variant you like more

api/src/org/labkey/api/assay/AssayProtocolSchema.java
cnathe: Using `setVisibleColumns` to `false` would skip the if statement that does the `table.setDefaultVisibleColumns`, but we do need to also clear the `visibleColumnNames` set which is returned from this method and used to append to `visibleColumns` in the callers.

labkey-jeckels: Indeed. OK, never mind!

PR #2193, opened 2021-04-15
42756: Disable bulk update for Submitters+Readers

labkey-klum:
#### Rationale
To address this issue : https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=42756
We want to ensure the bulk update import option is properly hidden if the user doesn't have update permissions. We also need to ensure the update will fail on the same permission check on execute.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/509

#### Changes
- clean up some of the hardcoded URL param string names
- add a permission check on `ImportViewBean.setShowImportOptions`
- introduce `canInsert` and `canUpdate` methods and make the appropriate check against the selected insert option

### DIFF
--- a/api/src/org/labkey/api/query/AbstractQueryImportAction.java
+++ b/api/src/org/labkey/api/query/AbstractQueryImportAction.java
@@ -49,6 +49,7 @@ import org.labkey.api.reader.TabLoader;
 import org.labkey.api.resource.Resource;
 import org.labkey.api.security.User;
 import org.labkey.api.security.permissions.InsertPermission;
+import org.labkey.api.security.permissions.UpdatePermission;
 import org.labkey.api.util.CPUTimer;
 import org.labkey.api.util.FileStream;
 import org.labkey.api.util.PageFlowUtil;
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/query/AbstractQueryImportAction.java
```diff
@@ -268,12 +275,52 @@ public final ApiResponse execute(FORM form, BindException errors) throws Excepti
...
+
+            switch (insertOption)
+            {
+                case MERGE, REPLACE, UPSERT -> {
+                    if (!canUpdate(user))
+                        errors.reject(SpringActionController.ERROR_MSG, "User does not have permission to update rows");
+                }
+                default -> {
+                    if (!canInsert(user))
+                        errors.reject(SpringActionController.ERROR_MSG, "User does not have permission to update rows");
```
labkey-adam:
```suggestion
                        errors.reject(SpringActionController.ERROR_MSG, "User does not have permission to insert rows");
```

labkey-adam (changes requested):
Does the insert validation message needs to be updated?

labkey-adam (approved):

PR #2194, opened 2021-04-15
Item 8618: Creating Aliquots in the UI 

XingY:
#### Rationale
Sample name can be auto generated during creation used the configured naming pattern. Aliquots are a group of special samples that needs special naming that depend own its parent name. 

#### Related Pull Requests
* 

#### Changes
* generate aliquot name as ParentSampleName-Incrementing Value


api/src/org/labkey/api/exp/query/ExpSchema.java
```diff
@@ -434,6 +434,18 @@ public TableInfo getLookupTableInfo()
...
     }
 
+    public ForeignKey getMaterialForeignKey(ContainerFilter cf, String targetColumnName)
+    {
+        return new LookupForeignKey(targetColumnName)
+        {
+            @Override
+            public TableInfo getLookupTableInfo()
+            {
+                return getTable(TableType.Materials.toString(), ContainerFilter.EVERYTHING);
```
labkey-susanh: Is this really supposed to be ignoring the passed-in `ContainerFilter`?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -978,4 +980,44 @@ public void addAuditEvent(User user, Container container, String comment, ExpMat
         event.setInventoryUpdateType(updateType);
         AuditLogService.get().addEvent(user, event);
     }
+
+    @Override
+    public ExpMaterial getMaterialByName(@NotNull String sampleName, @NotNull String sampleTypeLsid, Container container)
+    {
+        SimpleFilter filter = SimpleFilter.createContainerFilter(container);
+        filter.addCondition(FieldKey.fromParts("Name"), sampleName);
+        filter.addCondition(FieldKey.fromParts("cpastype"), sampleTypeLsid);
```
labkey-susanh: It surprises me we wouldn't have had such a method already.  I wonder why it never came up before.

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -752,20 +756,29 @@ public DataIterator getDataIterator(DataIteratorContext context)
...
+        final Container _container;
+        final String _sampleTypeLsid;
+        final int _batchSize;
         boolean first = true;
 
         String generatedName = null;
         String generatedLsid = null;
 
-        _GenerateNamesDataIterator(ExpSampleTypeImpl sampletype, MapDataIterator source, DataIteratorContext context)
+        private Map<String, DbSequence> _aliquotSequences = new HashMap<>();
```
labkey-susanh, line 768: Are both `_aliquotSequences` and `alliquotDBSequences` used?

labkey-susanh (approved):

PR #2199, opened 2021-04-20
Issue 42872: Account for aliased field names when constructing detailed audit log diffs

labkey-susanh:
#### Rationale
When field names contain special characters, the existing row we use for computing audit log diffs uses the aliased field name (that is, the field name used in the database) as the key but the row provided by the user and has the display names as keys and we want to use the display names when showing the differences.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/503
* https://github.com/LabKey/sampleManagement/pull/552
* https://github.com/LabKey/biologics/pull/862

#### Changes
* Find the column name based on the column alias in `getOldAndNewRecordForMerge`


labkey-matthewb (commented):
This fix makes sense, but I'm wondering if I should investigate fixing up the column names that are created for the existing record?

api/src/org/labkey/api/audit/AuditHandler.java
```diff
@@ -40,27 +42,30 @@ void addAuditEvent(User user, Container c, TableInfo table, @Nullable AuditBehav
...
 
+        List<ColumnInfo> columns = table == null ? Collections.emptyList() : table.getColumns();
         // Iterate through existingRow keys since these have the casing we want
         // and we won't convert sample type and data class names into lower case.
         for (Map.Entry<String, Object> entry : existingRow.entrySet())
         {
             String key = entry.getKey();
-            boolean isExtraAuditField = extraFieldsToInclude != null && extraFieldsToInclude.contains(key);
-            if (!excludedFromDetailDiff.contains(key) && row.containsKey(key))
+            String nameFromAlias = columns
```
labkey-matthewb, line 59: This is usually fine, but I've seen cases where this n^2 kind of loop has blown up (e.g. accidentally calling resolveColumn() in table constructor etc).  Some tables have a lot of columns.  Consider a map?

labkey-susanh: There is, in some cases an aliasMap available, but it didn't seem available always and is not exposed from a general tableInfo.  Maybe it could be?

labkey-susanh:
> This fix makes sense, but I'm wondering if I should investigate fixing up the column names that are created for the existing record?

Fixing up the column names seemed like it would have further-reaching effects than I wanted for this change, but if you want to investigate what makes sense, I'm all for it.

labkey-matthewb (approved):

PR #2201, opened 2021-04-20
Fix for provenance exception during assay import

labkey-klum:
#### Rationale
This change fixes the exception : `java.lang.IllegalStateException: Assay results LSID required to attach provenance` seen during assay run import. The previous refactor was causing the rowId to Lsid map to be empty even if the provenance module was present.


labkey-martyp (approved):
Thanks Karl, looks like that has fixed those tests

PR #2206, opened 2021-04-21
Issue 42990: System Maintenance task update to all for non-configurable, single-use tasks to be registered

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42990

This PR adds a new option to maintenance tasks so that we can register special / single-use type of tasks with the system maintenance UI which will not be included in the configurable set of daily maintenance tasks. 
Screen Shot 2021-04-21 at 4 35 08 PM

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/326
* https://github.com/LabKey/platform/pull/2206

#### Changes
* MaintenanceTask.isConfigurable(), default to true
* Update SystemMaintenanceJob to skip over any non-configurable tasks
* Configure System Maintenance page display updates to split configurable vs non-configurable tasks into separate panels

### DIFF
--- a/api/src/org/labkey/api/util/SystemMaintenance.java
+++ b/api/src/org/labkey/api/util/SystemMaintenance.java
@@ -265,5 +265,14 @@ public class SystemMaintenance
         {
             return false;
         }
+
+        /**
+         * Specify if the task is meant to be configurable and therefore shown in the group of maintenance tasks
+         * that can be enabled / disabled as part of the daily/nightly schedule.
+         */
+        default boolean isConfigurable()
+        {
+            return true;
+        }
     }
 }
--- a/api/src/org/labkey/api/util/SystemMaintenance.java
+++ b/api/src/org/labkey/api/util/SystemMaintenance.java
@@ -267,10 +267,10 @@ public class SystemMaintenance
         }
 
         /**
-         * Specify if the task is meant to be configurable and therefore shown in the group of maintenance tasks
-         * that can be enabled / disabled as part of the daily/nightly schedule.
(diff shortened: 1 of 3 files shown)


labkey-jeckels (approved):
One naming suggestion to consider.

labkey-jeckels:
@labkey-adam you might be interested in this too. Cory and I discussed this use case, where a pipeline job is triggered by an upgrade script but needs to be manually rerun on a single server. In general, I thought you'd be a fan of doing more Java-based upgrades in pipeline jobs when possible.

labkey-adam:
Right, seems like an alternative to the "hidden" upgrade actions we sometimes add... like `MigrateAuthenticationConfigurationsAction`

api/src/org/labkey/api/util/SystemMaintenance.java
```diff
@@ -265,5 +265,14 @@ default boolean hideFromAdminPage()
         {
             return false;
         }
+
+        /**
+         * Specify if the task is meant to be configurable and therefore shown in the group of maintenance tasks
+         * that can be enabled / disabled as part of the daily/nightly schedule.
+         */
+        default boolean isConfigurable()
```
labkey-jeckels: Consider isSchedulable() or isRecurring()

labkey-jeckels (approved):

api/src/org/labkey/api/util/SystemMaintenance.java
cnathe: Done. renamed to isRecurring()

PR #2207, opened 2021-04-21
Update StudyPublishTest to verify non-zero records in published datasets

labkey-chrisj:
#### Rationale
Issue [25892 ](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=25892)calls for verification that published datasets receive non-zero records, this adds verification that published datasets get at least one

#### Related Pull Requests
n/a


labkey-tchad (changes requested):
This should verify that the published study has participants as well. The original regression mentions that they were missing as well: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=25883

labkey-chrisj:
in this context, ensuring values in the MouseID column validates participants, right?  I can do that in the same dataTable right after ensuring non-zero records in it

study/test/src/org/labkey/test/tests/study/StudyPublishTest.java
```diff
@@ -383,7 +389,10 @@ protected void verifyPublishedStudy(@LoggedParam final String name, final String
         {
             pushLocation();
             clickAndWait(Locator.linkWithText(dataset));
-            new DatasetPropertiesPage(getDriver()).clickViewData();
+            DataRegionTable dataTable = new DatasetPropertiesPage(getDriver()).clickViewData().getDataRegion();
+            assertTrue("Expect > 0 records", dataTable.getDataRowCount() > 0 );
+            assertThat("expect all participant data to be present",
+                    dataTable.getColumnDataAsText("Mouse Id"), allOf(notNullValue()));
```
labkey-tchad: This doesn't work as you expect. `allOf` combines multiple matchers; it doesn't apply a matcher to each member of a collection. This will just verify that `getColumnDataAsText` returns a non-null value (which it always does).
I think you want `not(hasItem(""))`; I don't think `getColumnDataAsText` returns null values for empty cells.

labkey-tchad (changes requested):

study/test/src/org/labkey/test/tests/study/StudyPublishTest.java
```diff
@@ -899,6 +908,7 @@ private void rePublishStudy(String parentName, String newName)
         waitForElement(Locator.xpath("//div[@class = 'labkey-nav-page-header'][text() = 'Folder Objects']"));
         verifyPublishWizardSelectedCheckboxes(StudyPublishWizardGrid.folderObjects,
                 "Container specific module properties",
+                "ETL Definitions",
```
labkey-tchad: Is this a fix for a test failure? I don't think "ETL Definitions" will show up in community suites.

labkey-chrisj: ok, that explains it.  Also, I kinda wish we didn't have all these things you 'just have to know' about why tests in a standard work environment behave so differently from the test environment.

labkey-tchad (approved):

PR #2209, opened 2021-04-23
Item 8789: Initial implementation for client side metric counts

cnathe:
#### Rationale
This PR adds the server side implementation for the initial light weight client side metric tracking. This implementation adds a new API action in the core controller and an in memory map to track the counts for various client side metrics grouped by feature area so that those counts can be included with the JSON metric data supplied by the core module. 

Here is an example of what that new JSON data will look like:

```
            "Core": {
                "authenticationConfigurations": {
                    "CAS": 1,
                    "Database": 1
                },
                "buildInfo": {
                    "vcsTag": "Unknown",
                    "buildTime": "Apr 12, 2021, 9:18:25 AM",
                    "vcsBranch": "Unknown",
                    "buildNumber": "Unknown",
                    "vcsUrl": "Unknown",
                    "version": "21.001",
                    "vcsRevision": "Unknown"
                },
                "clientSideMetricCounts": {
                    "Freezer Manager": {
                        "copyFreezerDefinition": 1
                    },
                    "Area1": {
                        "metric1": 2,
                        "metric2": 3
                    }
                },
                ...
            }
```

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/514
* https://github.com/LabKey/platform/pull/2209
* https://github.com/LabKey/inventory/pull/229

#### Changes
* core/package.json @labkey/components version update
* add new CoreController.IncrementClientSideMetricCountAction
* register ClientSideMetricManager usage metrics in CoreModule
* ClientSideMetricManager class to track metric count map by feature area and add to core module JSON metric data

### DIFF
--- a/core/src/org/labkey/core/metrics/ClientSideMetricManager.java
+++ b/core/src/org/labkey/core/metrics/ClientSideMetricManager.java
@@ -6,25 +6,23 @@ import org.labkey.api.util.UsageReportingLevel;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 public class ClientSideMetricManager
 {
     private static final ClientSideMetricManager instance = new ClientSideMetricManager();
-    private static final Map<String, Integer> METRIC_COUNTS = new HashMap<>();
+    private static final Map<String, Map<String, AtomicInteger>> FEATURE_AREA_METRIC_COUNTS = new ConcurrentHashMap<>();
 
     public static ClientSideMetricManager get()
     {
         return instance;
     }
 
-    public Integer increment(String metricName)
+    public int increment(String featureArea, String metricName)
     {
-        if (!METRIC_COUNTS.containsKey(metricName))
-            METRIC_COUNTS.put(metricName, 0);
-
-        Integer count = METRIC_COUNTS.get(metricName) + 1;
-        METRIC_COUNTS.put(metricName, count);
-        return count;
+        FEATURE_AREA_METRIC_COUNTS.computeIfAbsent(featureArea, k -> new HashMap<>());
+        return FEATURE_AREA_METRIC_COUNTS.get(featureArea).computeIfAbsent(metricName, s -> new AtomicInteger(0)).incrementAndGet();
     }
 
     public void registerUsageMetrics(String moduleName)
@@ -32,7 +30,7 @@ public class ClientSideMetricManager
         UsageMetricsService svc = UsageMetricsService.get();
         if (null != svc)
         {
-            svc.registerUsageMetrics(UsageReportingLevel.MEDIUM, moduleName, () -> Collections.singletonMap("clientSideMetricCounts", METRIC_COUNTS));
+            svc.registerUsageMetrics(UsageReportingLevel.MEDIUM, moduleName, () -> Collections.singletonMap("clientSideMetricCounts", FEATURE_AREA_METRIC_COUNTS));
         }
     }
 }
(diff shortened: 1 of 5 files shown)


labkey-susanh (commented):
These changes seem fine to me, but I wonder if we will always want all of these metrics to appear under the Core module? At the moment, of course, it's a small set of metrics, but it could get somewhat unwieldy, and using separate keys for the applications could make for easier querying.  We are already providing the application name when submitting the metric, so we could change the structure of the metric to have another layer under the Core.clientSideMetricCounts, or each module with an application could register with the manager (if it moved to an API package) and get its own clentSideMetricCounts.  An argument for leaving it with Core is that we might want to gather the same count across multiple modules (if a UI component is shared, for example), but that seems like the data might get a bit muddier.

core/src/org/labkey/core/metrics/ClientSideMetricManager.java
```diff
@@ -0,0 +1,38 @@
...
+{
+    private static final ClientSideMetricManager instance = new ClientSideMetricManager();
+    private static final Map<String, Integer> METRIC_COUNTS = new HashMap<>();
+
+    public static ClientSideMetricManager get()
+    {
+        return instance;
+    }
+
+    public Integer increment(String metricName)
```
labkey-jeckels: This can return int instead of Integer.

core/src/org/labkey/core/metrics/ClientSideMetricManager.java
```diff
@@ -0,0 +1,38 @@
...
+    {
+        if (!METRIC_COUNTS.containsKey(metricName))
```
labkey-jeckels: Suggest using an AtomicInteger instead. It'll be thread safe, and will let you do something like:

return METRIC_COUNTS.computeIfAbsent(metricName, s -> new AtomicInteger(0)).incrementAndGet();

core/src/org/labkey/core/metrics/ClientSideMetricManager.java
```diff
@@ -0,0 +1,38 @@
...
+import org.labkey.api.util.UsageReportingLevel;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ClientSideMetricManager
+{
+    private static final ClientSideMetricManager instance = new ClientSideMetricManager();
+    private static final Map<String, Integer> METRIC_COUNTS = new HashMap<>();
```
labkey-jeckels: Multiple HTTP requests may be accessing this concurrently, so this should be synchronized in some way. Swapping in ConcurrentHashMap should be enough. It doesn't support null for keys or values, but that's fine for this scenario.

labkey-jeckels (changes requested):

labkey-jeckels:
Cory, I know we talked about possible approaches for a nested JSON tree for these metrics or a flat namespace, but if you're receiving the metric API request with two separate strings I'd vote for reflecting that in the JSON structure that's submitted.

core/src/org/labkey/core/metrics/ClientSideMetricManager.java
cnathe: done

core/src/org/labkey/core/metrics/ClientSideMetricManager.java
cnathe: done

core/src/org/labkey/core/metrics/ClientSideMetricManager.java
cnathe: done

labkey-jeckels (approved):

labkey-susanh (approved):

PR #2212, opened 2021-04-25
Issue 42613 - Combine low and medium usage reporting levels

labkey-ankurj:
#### Rationale
We want to reduce the complexity for implementing usage metrics and in choosing the configuration for a given installation. Right now we have three options - off, low, and medium.

We want to switch to just off and on. The on level will send the same content as the medium level does today.

During upgrade, servers that are currently at low or medium will be switch to the "on" setting.

#### Related Pull Requests
* 

#### Changes
* Remove "MEDIUM" and "LOW" usage levels
* sql upgrade script to update "MEDIUM" and "LOW" to "ON"

### DIFF
--- a/api/src/org/labkey/api/security/AuthenticationManager.java
+++ b/api/src/org/labkey/api/security/AuthenticationManager.java
@@ -1369,7 +1369,7 @@ public class AuthenticationManager
 
     public static void registerMetricsProvider()
     {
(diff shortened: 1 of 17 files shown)


labkey-jeckels (approved):

labkey-jeckels:
I know I just approved all of the different PRs, but what do you think about removing the level argument from the registerUsageMetrics() method? It's no longer adding any value - either it's off, in which case the metrics won't be gathered, or it's on and it will.

labkey-jeckels (approved):

PR #2216, opened 2021-04-27
8802: sample publish/recall audit events

labkey-kevink:
#### Rationale
Create an aggregate publish event when linking a group of SampleType rows to a study.  Add sample timeline events for each SampleType row as well.  Similarly, add recall events when removing SampleType rows from a study dataset.

#### Related Pull Requests
* https://github.com/LabKey/provenance/pull/67
#### Changes
* Rename AssayAuditProvider to PublishAuditProvider and support SampleType publish/recall events
* Conditionally add the "Link to Study" and "Recall" buttons only if study is available
* Teach the publish history views to support SampleType events
* Create provenance run linking the published SampleType -> dataset rows


api/src/org/labkey/api/exp/api/ExperimentUrls.java
```diff
@@ -32,6 +33,11 @@
  */
 public interface ExperimentUrls extends UrlProvider
 {
+    static ExperimentUrls get()
+    {
+        return PageFlowUtil.urlProvider(ExperimentUrls.class);
+    }
+
```
labkey-klum, line 40: nice

labkey-klum (approved):
I saw you added code to create a provenance run for samples. Are we cleaning up that run if the source gets deleted or is that something that needs to be handled later?

labkey-klum:
@labkey-susanh if you are interested in reviewing how publish/recall events are incorporated into `SampleTimeline` take a look at the changes in `StudyPublishManager`

labkey-kevink:
> I saw you added code to create a provenance run for samples. Are we cleaning up that run if the source gets deleted or is that something that needs to be handled later?

Good catch.  We were cleaning up the provenance data when the dataset rows were recalled, but not when the sample rows were deleted.  We will now cleanup the provenance data and StudyPublish run if the source samples are deleted.  I've refactored the provenance delete code to handle dataset, assay result, and sample LSIDs.

PR #2222, opened 2021-04-28
Issue 43009: ETL fails when aliasing built-in columns

labkey-martyp:
#### Rationale
For built-in columns like ParticipantId, if they are aliased in the source query we are passing through propertyURIs of the original column which can cause duplicate matches based on propertyURI alone.  This update is scoped to ETLs and matches on the following:
- Name and PropertyURI - Top match (This is really just necessary for ParticipantId)
- Name - Second match
- Vocabulary match - Same as before

Ran this across most of the core tests and dataintegration.

#### Related Pull Requests
* https://github.com/LabKey/dataintegration/pull/105

#### Changes
* Update DataIteratorUtil._matchColumns as described above

### DIFF
--- a/api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
+++ b/api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
@@ -191,6 +191,14 @@ public class DataIteratorUtil
         ArrayList<Pair<ColumnInfo,MatchType>> matches = new ArrayList<>(input.getColumnCount()+1);
         matches.add(null);
 
+        boolean isEtl = false;
+        if (input instanceof SimpleTranslator)
+        {
+            String source = ((SimpleTranslator) input)._context.getDataSource();
+            if (null != source)
+                isEtl = source.equals(ETL_DATA_SOURCE);
+        }
+
         // match columns to target columninfos (duplicates StandardDataIteratorBuilder, extract shared method?)
         for (int i=1 ; i<=input.getColumnCount() ; i++)
         {
@@ -201,20 +209,11 @@ public class DataIteratorUtil
                 continue;
             }
 
-            // Match by name first
-            Pair<ColumnInfo,MatchType> to = targetMap.get(from.getName());
-
-            // If name matches, check if property URI matches for higher priority match type
-            if (null != to && null != from.getPropertyURI())
-            {
-                // Built-in columns aliased in source query will not match. Just stick with name match.
-                if (from.getPropertyURI().equals(to.first.getPropertyURI())) {
-                    Pair<ColumnInfo,MatchType> toUri = targetMap.get(from.getPropertyURI());
-                    if (null != toUri)
-                        to = toUri;
-                }
-            }
-
+            Pair<ColumnInfo,MatchType> to = null;
+            if (null != from.getPropertyURI() && !isEtl)
+                to = targetMap.get(from.getPropertyURI());
+            if (null == to)
+                to = targetMap.get(from.getName());
             if (null == to)
             {
                 // Check to see if the column i.e. propURI has a property descriptor and vocabulary domain is present
--- a/api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
+++ b/api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
@@ -210,6 +210,7 @@ public class DataIteratorUtil
             }
 
             Pair<ColumnInfo,MatchType> to = null;
+            // Only do propertyURI matching if not an ETL
             if (null != from.getPropertyURI() && !isEtl)
                 to = targetMap.get(from.getPropertyURI());
             if (null == to)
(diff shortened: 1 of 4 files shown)


study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
```diff
@@ -251,13 +251,18 @@ else if (match.getPropertyType() == PropertyType.FILE_LINK)
...
-        }
-        catch (ValidationException e)
+
+        // If not an ETL then get ParticipantId translate column
+        if (context.getDataSource() == null || !context.getDataSource().equals("etl"))
         {
-            context.getErrors().addRowError(e);
+            try
+            {
+                translatedIndexPTID = it.translatePtid(indexPTIDInput, user);
```
labkey-matthewb: Wouldn't it me more consistent to just avoid the propertyURI matching business in findColumnInMap()?  This would skip alias handing which could be a problem.

labkey-matthewb: I'm not following this.  translatePtid() manages participant aliases as in alternate ids for the same participant.

labkey-martyp: @labkey-matthewb Got it. I didn't follow that all the way through.  Thought this was for translating participantId column name that has been changed in the study properties.  I reverted this change conditionalizing the translating.  

However, does the columnInfo created in translatePtid have to be named ParticipantId?  It doesn't seem to be using the column name anywhere.  Instead it uses the index of the column when it uses it.  The issue for this name matching solution is having two columns named ParticipantId when trying to do column matching after this in the StandardDataIteratorBuilder.  

I pushed a commit trying out renaming the translatePtid columnInfo to ParticipantAliasId, to run some tests.  Not sure if we have test coverage for ParticipantId aliasing.  I'll look for that.

labkey-martyp:
@labkey-matthewb I tried out several other solutions for this but didn't come up with another one that handles source query aliasing of built-in columns, participantId alias table and importing participantId by label. I ended going back to my original solution for this but scoping it to just ETLs.  Basically it's matching columns by
- Name and PropertyURI - Top match (This is really just necessary for ParticipantId)
- Name - Second match
- Vocabulary match - Same as before

As discussed before, using something like `SELECT ParticipantId, ParticipantId as AliasId FROM Dataset` as a source query will result in two columns in the dataiterator with the same PropertyURI, so that alone is not a sufficient match.  Also since we are adding the translatePtid column, which has to be named "ParticipantId" (but does not have a propertyURI) in order to match statement parameters, there can be multiple columns in the dataiterator with the name "ParticipantId", so name alone is not sufficient to match for ParticipantId.  Thus this method of matching both Name and PropertyURI as the top match is really the only way with the current logic to match ParticipantId columns.

I do think it would be worth considering making the matchColumns a little more customizable for different data storage types or import processes.  For instance, DatasetDataiteratorBuild.findColumnInMap has different matching than DataiteratorUtil.matchColumns, and now possibly different matching for ETLs.  Not sure if that would mean pushing matchColumns into the dataiterators or somewhere else.

study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
labkey-martyp: Renaming that column does not work as it does not match with statement parameters.

api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
```diff
@@ -200,11 +205,34 @@
...
+                // Match by name first
                 to = targetMap.get(from.getName());
+
+                // If name matches, check if property URI matches for higher priority match type
+                if (null != to && null != from.getPropertyURI())
+                {
+                    // Renamed built-in columns (ex. LSID, ParticipantId) in source queries will not match propertyURI with
+                    // target. In that case, just stick with name match. Otherwise check for propertyURI match for higher priority match
+                    // (this is primarily for ParticipantId which can have two columns with same name in the dataiterator)
+                    if (from.getPropertyURI().equals(to.first.getPropertyURI())) {
```
labkey-matthewb, line 221: I understand matching by name first (higher priority), that was one of my earliest suggestions.  If we are only matching by name in this isETL branch, then the nested if (propertyURI.equals) only makes sense if there are multiple columns in the source with the same name??? Is that a thing that happens?

labkey-matthewb (approved):

api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
labkey-martyp: Yes, just ParticipantId.  Because of the participantId alias table functionality, we're adding a [columnInfo hard-coded to ParticipantId](https://github.com/LabKey/platform/blob/8a26f73a7644cd729cf73a562dba9ad58fd4b755/study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java#L615) name, so two participantId named columnInfos is likely with datasets.  I tried changing that hard-coded name because it's not used in the column matching or in calling the translator, but apparently it is needed for matching the statement parameters.  

The participantId alias columnInfo is actually performing a second purpose, although I'm not sure it's intentional.  In the validation recently added in StandardDataIteratorBuilder, having this column there allows imports by ParticipantId label (Participant ID) to pass validation of required columns, since our normal column matching does not consider labels but this participantId alias column is always there to meet required column validation.  Anyway, that's not totally related but makes it difficult to change the current participantId alias behavior.

PR #2224, opened 2021-04-29
core-apiTest.view action update to "Eval" button text -> "Eval JSON Response"

cnathe:
#### Rationale
The core-apiTest.view action can be used to test api actions and view the response information. On this page there is a button with text "Eval". This button will evaluate the validity of the JSON response object from the api call. This PR updates that button text to make it a bit more clear what it will do.

### DIFF
--- a/query/src/org/labkey/query/view/apitest.jsp
+++ b/query/src/org/labkey/query/view/apitest.jsp
@@ -105,7 +105,7 @@
     </tr>
     <tr>
         <td width="100%">
-            <input type="button" id="btnEval" value="Eval JSON Response" onclick="evalResponse()"/>&nbsp;
+            <input type="button" id="btnEval" value="Validate JSON Response" onclick="evalResponse()"/>&nbsp;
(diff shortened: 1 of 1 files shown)


query/src/org/labkey/query/view/apitest.jsp
```diff
@@ -105,7 +105,7 @@
     </tr>
     <tr>
         <td width="100%">
-            <input type="button" id="btnEval" value="Eval" onclick="evalResponse()"/>&nbsp;
+            <input type="button" id="btnEval" value="Eval JSON Response" onclick="evalResponse()"/>&nbsp;
```
labkey-susanh: This is definitely better.  Perhaps "Validate JSON Response"? 'Eval' suggests to me that it might do something like "evaluate expressions".

labkey-susanh (approved):

query/src/org/labkey/query/view/apitest.jsp
cnathe: Good suggestion. Changed to "Validate JSON Response"

PR #758, opened 2020-01-21
getQueryDetails performance

labkey-matthewb:
avoid redundant calls to createTable() and other table constructors

PR #2229, opened 2021-04-30
Issue 43050: Lookup with bad fkColumnName reference passes through API

labkey-jeckels:
#### Rationale
We used to detect this case and use the lookup target's PK instead. We accidentally (I think) lost that behavior with the changes to JsonWriter in the related PR

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/758

#### Changes
* Fall back to use lookup's PK when the referenced column doesn't exist
* Log warnings so the incorrect lookup config can be corrected

### DIFF
--- a/api/src/org/labkey/api/data/JsonWriter.java
+++ b/api/src/org/labkey/api/data/JsonWriter.java
@@ -16,6 +16,8 @@
 package org.labkey.api.data;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONArray;
 import org.json.JSONObject;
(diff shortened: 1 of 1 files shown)


labkey-matthewb (approved):
This seems fine.  It does go a little against the attempt to do less resolving/construction of objects during meta-data description calls.  This change itself doesn't affect performance, but the expectation that we return null for invalid configurations enforces the idea that we have to check, and this is a performance road-bump in for query metadata API (in particular that we have to construct the target table in order to validate column existence and casing).

labkey-kevink (approved):

labkey-jeckels:
@labkey-matthewb yes, I agree. If it hadn't already been constructing the table and checking if the column existed I wouldn't propose adding this code. But it seemed like as long as we're in a position to catch this, it's better to help identify the problem. We could also make it blow up instead of transforming it to the PK

PR #2235, opened 2021-05-05
Link to Study Improvements

labkey-klum:
#### Rationale
The PR introduces some improvements to link to study as part of the previous work to support linking sample records to studies.

#### Changes
- Lineage column support for samples : parent sample columns can be traversed to pull in sources for subject/timepoint during the link to study process. The parent fields must be annotated with the proper concept URIs (no magic field name matching will occur)
- Assay sample match : if the assay has a sample column we will attempt to resolve the subject/timepoint from that sample in a similar way that specimen matching is handled. The sample must have been previously linked to the study for the resolution to occur.
- Match status indicator : similar to the specimen match indicator on the publish confirm page, we will display a check mark if the sample ID has been matched to a sample in the target study. Unlike the specimen match, we will only try to match via sample ID, and are not trying to determine if there is a sample which matches the subject/timepoint entered on the page. Sample matching only occurs if there is no specimen ID field in the assay results and there is a sample ID field.


api/src/org/labkey/api/study/assay/SampleParticipantVisitResolver.java
```diff
@@ -0,0 +1,89 @@
...
+        super(runContainer, targetStudyContainer, user);
+    }
+
+    @Override
+    protected @NotNull ParticipantVisit resolveParticipantVisit(String sampleId, String participantID, Double visitID, Date date, Container targetStudyContainer)
+    {
+        ParticipantVisitImpl originalInfo = new ParticipantVisitImpl(sampleId, participantID, visitID, date, getRunContainer(), targetStudyContainer);
+        if (targetStudyContainer != null)
+        {
+            Integer id = Integer.valueOf(sampleId);
```
labkey-kevink, line 44: Should this method check the `_resolvedSamples` local cache?  Should we cash misses as well?

api/src/org/labkey/api/study/query/PublishResultsQueryView.java
```diff
@@ -460,6 +475,19 @@ public Object getUserDate(RenderContext ctx, boolean includeTimestamp)
...
         }
 
+        private Object getColumnValue(ColumnInfo col, RenderContext ctx)
+        {
+            DisplayColumn dc = col.getRenderer();
+            if (dc instanceof IMultiValuedDisplayColumn)
+            {
+                // support for lineage and multivalue columns
+                List<Object> values = ((IMultiValuedDisplayColumn)dc).getDisplayValues(ctx);
+                if (values.size() == 1)
```
labkey-kevink, line 485: Should we warn or throw if there is more than one value?

api/src/org/labkey/api/study/query/PublishResultsQueryView.java
```diff
@@ -670,6 +698,36 @@ else if (_specimenIDCol != null && _specimenIDCol.getValue(ctx) != null)
...
 
+        public Pair<Boolean, String> getSampleMatchStatus(RenderContext ctx)
+        {
+            Container targetStudy = getUserTargetStudy(ctx);
+            if (targetStudy == null)
+                targetStudy = _targetStudyContainer;
+
+            // Bail early if no match can be made.
+            if (targetStudy == null)
+                return new Pair<>(false, "<p>No target study selected for this row.</p>");
```
labkey-kevink, line 711: Ack, stringly html!

study/src/org/labkey/study/assay/AssayPublishConfirmAction.java
```diff
@@ -203,28 +211,53 @@ protected FieldKey getObjectIdFieldKey(AssayPublishConfirmForm form)
...
+        additionalCols.put(ExtraColFieldKeys.SpecimenVisit, new FieldKey(new FieldKey(specimenIDFieldKey, "Specimen"), "SequenceNum"));
+        additionalCols.put(ExtraColFieldKeys.SpecimenDate, new FieldKey(new FieldKey(specimenIDFieldKey, "Specimen"), "DrawTimestamp"));
+
+        UserSchema userSchema = getUserSchema(form);
+        QueryView view = new QueryView(userSchema, getQuerySettings(form), null);
+        DataView dataView = view.createDataView();
+        Map<FieldKey, ColumnInfo> selectColumns = dataView.getDataRegion().getSelectColumns();
+
+        // If there are any sample columns in the assay, we can try to match existing samples in the target study
+        List<ColumnInfo> sampleCols = selectColumns.values().stream()
```
labkey-kevink, line 243: This might not work on old assay definitions that predate our use of the sample concept URI.  We have some older code that inspects the lookup target schema is "samples".

labkey-kevink (approved):
Looks good, just minor comments.  We will definitely need an automated test to cover this scenario of including lineage columns when performing link-to-study.

api/src/org/labkey/api/study/assay/SampleParticipantVisitResolver.java
labkey-klum: 👍

api/src/org/labkey/api/study/query/PublishResultsQueryView.java
labkey-klum: I added a warning

study/src/org/labkey/study/assay/AssayPublishConfirmAction.java
labkey-klum: Good point, I can check with Hannah to see if we need to add that support.

PR #2237, opened 2021-05-05
Issue 43096: Dynamic warning providers should suppress warnings for non-admins

labkey-adam:
#### Rationale
[Issue 43096: Dynamic warning providers should suppress warnings for non-admins](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43096)

#### Changes
* Show missing views warning to site admins only
* Refactor to ensure non-null ViewContext, User, and Request

### DIFF
--- a/core/src/org/labkey/core/view/template/bootstrap/CoreWarningProvider.java
+++ b/core/src/org/labkey/core/view/template/bootstrap/CoreWarningProvider.java
@@ -16,13 +16,13 @@
 package org.labkey.core.view.template.bootstrap;
 
 import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.NotNull;
 import org.labkey.api.admin.AdminUrls;
 import org.labkey.api.data.ConnectionWrapper;
 import org.labkey.api.data.DbScope;
 import org.labkey.api.module.JavaVersion;
 import org.labkey.api.module.ModuleHtmlView;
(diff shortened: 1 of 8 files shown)


core/src/org/labkey/core/view/template/bootstrap/CoreWarningProvider.java
```diff
@@ -69,41 +69,37 @@ public void addStaticWarnings(Warnings warnings)
     }
 
     @Override
-    public void addDynamicWarnings(Warnings warnings, ViewContext context)
+    public void addDynamicWarnings(@NotNull Warnings warnings, @NotNull ViewContext context)
     {
-        if (context != null && context.getRequest() != null)
+        if (context.getUser().hasSiteAdminPermission())
```
labkey-jeckels, line 74: It looks like app admins don't have control over any of the reasons for these warnings. I did a quick review, and assume you did the same.

labkey-jeckels (approved):

core/src/org/labkey/core/view/template/bootstrap/CoreWarningProvider.java
labkey-adam: This was a pre-existing permissions check (easiest to view while ignoring whitespace), but yes, I concur. Site admin might not be able to fix these issues either, of course, but gotta tell someone...

PR #2238, opened 2021-05-05
Improve audit log message when deleting an assay row

labkey-jeckels:
#### Rationale
Admins can enable editing and deleting individual rows in an assay's result data. Deletions are being audited but don't fully identify the row. This is a minimal change to add more context.

#### Changes
* Include the RowId of the newly deleted row in the message


labkey-kevink (approved):

labkey-susanh (commented):
Seems like it won't break anything, but row ids don't seem that useful, especially when they don't exist anymore.

labkey-jeckels:
@labkey-susanh Yes, but slightly better than nothing and potentially useful (if still painfully) for ETL purposes

PR #2243, opened 2021-05-07
Item 8709: Add support for Sample Picklists Definition CRUD

labkey-susanh:
#### Rationale
We are building support for user-defined picklists of samples within Sample Manager and eventually within Biologics as well. These lists are built on top of our list implementation, but using a new domain kind with some properties specific to picklists. 
We add a new permission class `ManagePicklistsPermission` because we want to enable Editors in the application to be able to create picklists without having permission to create lists in general.

Users should be able to see all public picklists and their own private picklists.  Administrators should be able to see and delete  their own lists as well as public picklists.  Within LKS, users with DesignListPermission can update picklist domains just as other domains, but Editors cannot.

Picklists can be imported and exported, along with other lists, but, just as other lists with references to sampleIds, are generally not portable between containers.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/522
* https://github.com/LabKey/inventory/pull/234
* https://github.com/LabKey/sampleManagement/pull/565

#### Changes
* Add new 'category' field for list metadata table to distinguish regular lists from public picklists from private picklists
* Add new permission class for managing picklists and assign it to editors and administrators
* Update DeleteListDefinitionAction to allow for passing in multiple listIds to be deleted
* Update methods for getting lists to allow for showing only public or owned picklists as appropriate.


api/src/org/labkey/api/exp/list/ListService.java
```diff
@@ -42,9 +42,10 @@ static void setInstance(ListService ls)
     }
 
     Map<String, ListDefinition> getLists(Container container);
+    Map<String, ListDefinition> getLists(Container container, User user, boolean includeAllLists);
     boolean hasLists(Container container);
     ListDefinition createList(Container container, String name, ListDefinition.KeyType keyType);
-    ListDefinition createList(Container container, String name, ListDefinition.KeyType keyType, @Nullable TemplateInfo templateInfo);
+    ListDefinition createList(Container container, String name, ListDefinition.KeyType keyType, @Nullable TemplateInfo templateInfo, ListDefinition.@Nullable Category category);
```
cnathe: What is the `ListDefinition.@Nullable` for the category param? I think that might be a copy-paste error.

list/src/org/labkey/list/ListModule.java
```diff
@@ -154,7 +158,10 @@ public void startupAfterSpringConfig(ModuleContext moduleContext)
         {
             svc.registerUsageMetrics(getName(), () -> {
                 Map<String, Object> metric = new HashMap<>();
-                metric.put("listCount", new SqlSelector(DbSchema.get("exp", DbSchemaType.Module), "SELECT COUNT(*) FROM exp.list").getObject(Long.class));
+                DbSchema dbSchema = DbSchema.get("exp", DbSchemaType.Module);
+                metric.put("listCount", new SqlSelector(dbSchema, "SELECT COUNT(*) FROM exp.list").getObject(Long.class));
+                metric.put("publicPicklistCount", new SqlSelector(dbSchema, "SELECT COUNT(*) FROM exp.list WHERE category ='" + ListDefinition.Category.PublicPicklist.toString() + "'").getObject(Long.class));
+                metric.put("privatePicklistCount", new SqlSelector(dbSchema, "SELECT COUNT(*) FROM exp.list WHERE category ='" + ListDefinition.Category.PrivatePicklist.toString() + "'").getObject(Long.class));
```
cnathe, line 164: Should these use `?` for the category = ...?

list/src/org/labkey/list/controllers/ListController.java
```diff
@@ -275,34 +291,68 @@ public void addNavTrail(NavTree root)
...
+                        ListDef listDef = ListManager.get().getList(c, listId);
+                        boolean isPicklist = listDef.getCategory() != null;
+                        if (isPicklist)
+                        {
+                            boolean isOwnPicklist = listDef.getCreatedBy() == getUser().getUserId();
+                            boolean canDeletePicklist = isOwnPicklist && c.hasPermission(getUser(), ManagePicklistsPermission.class) ||
+                                    (listDef.getCategory() == ListDefinition.Category.PublicPicklist && (
+                                                    getUser().hasApplicationAdminPermission() ||
+                                                    getUser().hasSiteAdminPermission() ||
+                                                    getContainer().hasPermission(getUser(), AdminPermission.class)
```
cnathe: Shouldn't this just be a check for if the user has `DesignListPermission` on the container? i.e. `getContainer().hasPermission(getUser(), DesignListPermission.class)`. Or what is the need for the site level admin permissions check?

list/src/org/labkey/list/model/ListDomainKind.java
```diff
@@ -259,9 +261,9 @@ public DbScope getScope()
         return Collections.emptySet(); // TODO: Allow this to return the Key Column
     }
 
-    public static Lsid generateDomainURI(String name, Container c, ListDefinition.KeyType keyType)
+    public static Lsid generateDomainURI(String name, Container c, KeyType keyType, ListDefinition.@Nullable Category category)
```
cnathe: Ok maybe the other usage wasn't a typo. How is `ListDefinition.@Nullable` different from `@Nullable`?

list/src/org/labkey/list/model/ListManagerTable.java
```diff
@@ -46,6 +56,92 @@ public ListManagerTable(ListManagerSchema userSchema, TableInfo table, Container
         addWrapColumn(_rootTable.getColumn(FieldKey.fromParts("CreatedBy")));
         addWrapColumn(_rootTable.getColumn(FieldKey.fromParts("Modified")));
         addWrapColumn(_rootTable.getColumn(FieldKey.fromParts("ModifiedBy")));
+        addWrapColumn( _rootTable.getColumn(FieldKey.fromParts("Category")));
+        MutableColumnInfo sharingCol = addWrapColumn("Sharing", _rootTable.getColumn(FieldKey.fromParts("Category")));
```
cnathe, line 59: I guess I would have expected this to be an AliasedColumn, but this seems to work

list/src/org/labkey/list/model/ListManagerTable.java
```diff
@@ -37,7 +46,8 @@ public ListManagerTable(ListManagerSchema userSchema, TableInfo table, Container
     {
         super(table, userSchema, cf);
 
-        addWrapColumn(_rootTable.getColumn(FieldKey.fromParts("ListID"))).setHidden(true);
+        MutableColumnInfo listIdCol = addWrapColumn(_rootTable.getColumn(FieldKey.fromParts("ListID")));
+        listIdCol.setHidden(true);
```
cnathe: was the previous way this `ListId` column was being `setHidden` not working? I don't see any downstream usages of `listIdCol`.

list/src/org/labkey/list/model/ListManagerTable.java
```diff
@@ -46,6 +56,92 @@ public ListManagerTable(ListManagerSchema userSchema, TableInfo table, Container
...
+
+                @Override
+                public @NotNull HtmlString getFormattedHtml(RenderContext ctx)
+                {
+                    return HtmlString.of(getDisplayValue(ctx));
+                }
+            }
+        });
+
+        MutableColumnInfo countCol = addWrapColumn("ItemCount", _rootTable.getColumn(FieldKey.fromParts("ListID")));
```
cnathe, line 98: again I would have expected this to be an AliasedColumn

list/src/org/labkey/list/view/ListsWebPart.java
```diff
@@ -85,7 +85,7 @@ protected void renderView(ViewContext model, PrintWriter out) throws Exception
 
     private void renderNarrowView(ViewContext model, PrintWriter out)
     {
-        Map<String, ListDefinition> lists = ListService.get().getLists(model.getContainer());
+        Map<String, ListDefinition> lists = ListService.get().getLists(model.getContainer(), model.getUser(), false);
```
cnathe, line 88: Does the `false` here mean that we don't want to show any picklists in the narrow lists webpart? Just curious.

list/src/org/labkey/list/view/listsWebPart.jsp
```diff
@@ -37,9 +37,8 @@
     assert null != me;
     Container c = getContainer();
     User user = getUser();
-    Map<String, ListDefinition> lists = ListService.get().getLists(c);
+    Map<String, ListDefinition> lists = ListService.get().getLists(c, getUser(), false);
```
cnathe, line 40: Same question as above, does the `false` here mean we don't include any pick lists?

cnathe (changes requested):

api/src/org/labkey/api/exp/list/ListService.java
labkey-susanh: It's IntelliJ being super intelligent.  Same as `@Nullable ListDefinition.Category`.

list/src/org/labkey/list/ListModule.java
labkey-susanh: I guess it might be safer for portability, but string concatenation seems OK here.

list/src/org/labkey/list/controllers/ListController.java
labkey-susanh: The intent is that public picklists should be able to be deleted by any admin.  DesignListPermission is given to non-admins (e.g., AssayDesigners), so that didn't seem appropriate.  Perhaps there's some overkill in these permission checks.

list/src/org/labkey/list/model/ListDomainKind.java
labkey-susanh: It's not.  The annotations are really just for IntelliJ and it did this weird syntax when I added a `@Nullable` parameter to this method.

list/src/org/labkey/list/model/ListManagerTable.java
labkey-susanh: It was working fine.  This was just some initial modifications during my implementation iteration. Reverted.

list/src/org/labkey/list/model/ListManagerTable.java
labkey-susanh: `addWrapColumn` creates an `AliasedColumn` under the covers.

list/src/org/labkey/list/view/ListsWebPart.java
labkey-susanh: It means we shouldn't show ALL lists, only those the user has access to (public and own picklists and regular lists).

list/src/org/labkey/list/view/listsWebPart.jsp
labkey-susanh: Same answer as above. :-)

list/src/org/labkey/list/model/ListManagerTable.java
labkey-susanh: Yup, it's an aliased column.

cnathe (approved):

PR #2244, opened 2021-05-07
Sample + Study Integration: Sample Type Polish

RosalineP:
#### Rationale
Following the implementation of the bulk of the link-to-study functionality, there are a few remaining features to implement within this polish story. These include:

- Track sample rows that have been linked to a study
- Cap number of columns that we add for the "linked to study" list
- Auto link to study on import
- Clean up dataset rows on sample delete
- Metrics tracking

#### Related Pull Requests
* https://github.com/LabKey/provenance/pull/68
* https://github.com/LabKey/commonAssays/pull/329

#### Changes
* Modularize `addLinkedToStudyColumns` to be usable by Assay and Sample Type, and move it into `StudyPublishService`
* Within `SamplesSchema`'s `createTable()`, add linked to study columns which indicate if a row has been linked into a particular study
* Nomenclature change to distinguish non-modularized `autoLinkAssayResults` and `autoLinkSampleType` within `StudyPublishService`
* Update sample type dataIterator to auto-link new rows if auto-link target is set, and the necessary columns are populated
* Update within `GetMaterialDeleteConfirmationDataAction` to warn user when they are deleting sample type rows that will also delete linked study sample type dataset rows
* When deleting sample type that is linked to study, delete corresponding study sample type dataset. User is also warned of this on relevant page given by `DeleteProtocolByRowIdsAction`
* Add metrics


study/src/org/labkey/study/query/SampleDatasetTable.java
```diff
@@ -110,6 +111,10 @@ private TableInfo getSamplesTable()
             UserSchema userSchema = QueryService.get().getUserSchema(_userSchema.getUser(), sampleType.getContainer(), SamplesSchema.SCHEMA_NAME);
             if (userSchema != null)
                 _sampleTable = userSchema.getTable(sampleType.getName());
+
+            // Hide 'linked' column for Sample Type Datasets
+            if (userSchema instanceof AbstractExpSchema)
```
RosalineP: I think it should always be a `AbstractExpSchema`, but should we 
- keep the if as a double-check, or 
- perhaps add an assert?

experiment/src/org/labkey/experiment/ConfirmDeleteView.java
```diff
@@ -142,11 +142,6 @@ public Integer getSingleObjectRowId()
         }
     }
 
-    public ConfirmDeleteView(String objectType, Class<? extends Controller> detailAction, List<? extends ExpObject> objects, DeleteForm form)
```
RosalineP, line 145: Was unused

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -285,6 +289,129 @@ public boolean next() throws BatchValidationException
...
+                Object date = _dateCol != null ? _dateCol.get() : null;
+                Object visit = _visitIdCol.get();
+                Object lsid = _lsidCol.get();
+                int rowId = ((Number) _rowIdCol.get()).intValue();
+
+                // Only link rows that have a participant and a visit/date. Return if this is not the case
+                if (participantId == null || (date == null || visit == null))
+                    return true;
+
+                Study study = StudyService.get().getStudy(sampleType.getAutoLinkTargetContainer());
```
labkey-klum: we should be able cache the study variable since it's not going to change for the sample type auto link setting

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -285,6 +289,129 @@ public boolean next() throws BatchValidationException
...
+                Float visitId = null;
+                Date dateId = null;
+
+                // 13647: Conversion exception in auto link to study
+                if (study.getTimepointType().isVisitBased())
+                {
+                    visitId = Float.parseFloat(visit.toString());
+                } else {
```
labkey-klum: brace styling

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -3755,6 +3754,65 @@ private void deleteProtocolInputs(Container c, String protocolIdsInClause)
...
 
+    /**
+     * For set of rows selected for deletion, find get all linked-to-study datasets that will be affected by the delete
+     * and warn user
+     */
+    public static ArrayList<Map<String, Object>> includeLinkedToStudyText(List<ExpMaterialImpl> allMaterials, Set<Integer> deletable, User user, Container container)
+    {
+        ArrayList<Map<String, Object>> associatedDatasets = new ArrayList<>();
+        StudyPublishService studyPublishService = StudyPublishService.get();
+        if (studyPublishService != null && allMaterials.size() > 0)
```
labkey-klum: `!allMaterials.isEmpty()` is more readable

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -3755,6 +3754,65 @@ private void deleteProtocolInputs(Container c, String protocolIdsInClause)
...
+                    linkedColumnNames.add(column.getName());
+            }
+
+            // Obtain, for selected rows, the ids of datasets that are linked
+            Set<Integer> linkedDatasetsBySelectedRow = new HashSet<>();
+
+            // Over each selected row
+            SimpleFilter filter = new SimpleFilter().addInClause(FieldKey.fromParts(ExpMaterialTable.Column.RowId.toString()), deletable);
+            TableSelector rowIdsFromTableSelector = new TableSelector(tableInfo, new HashSet<>(linkedColumnNames), filter, null);
+            Collection<Map> selectedRow = rowIdsFromTableSelector.getCollection(Map.class);
```
labkey-klum: `Collection<Map<String, Object>> selectedRow = rowIdsFromTableSelector.getMapCollection()` would give you a better typed result

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -3755,6 +3754,65 @@ private void deleteProtocolInputs(Container c, String protocolIdsInClause)
...
+
+            // Over each column of name 'dataset<N>'
+            for (Map selectedColumn : selectedRow)
+            {
+                // Check if each cell is populated (that is, if the given row was linked to a certain study)
+                for (Object cell : selectedColumn.entrySet())
```
labkey-klum: if you used : `Map.Entry<String, Object> entry : selectedColumn.entrySet()` you wouldn't have to do the weird casting below.

study/src/org/labkey/study/assay/ExperimentListenerImpl.java
```diff
@@ -47,4 +62,36 @@ public void afterResultDataCreated(Container container, User user, ExpRun run, E
...
+                    if (!dataset.canDelete(user))
+                    {
+                        throw new UnauthorizedException("Cannot delete rows from dataset " + dataset);
+                    }
+
+                    UserSchema schema = QueryService.get().getUserSchema(user, dataset.getContainer(), "study");
+                    TableInfo tableInfo = schema.getTable(dataset.getName());
+
+                    SimpleFilter filter = new SimpleFilter(FieldKey.fromParts(ExpMaterialTable.Column.RowId.toString()), material.getRowId());
+                    Collection<String> lsids = new TableSelector(tableInfo, singleton("LSID"), filter, null).getCollection(String.class);
```
labkey-klum: Shouldn't this always return a single row? If so you could use `getObject` which will blow up if more than one row is returned.

study/src/org/labkey/study/assay/StudyPublishManager.java
```diff
@@ -1071,6 +1079,66 @@ public ActionURL autoLinkResults(ExpProtocol protocol, ExpRun run, User user, Co
...
+                {
+                    LOG.debug(String.format("Resolved target study in container %s for auto-linking with %s from container %s", targetContainerPath, sampleTypeName, containerPath));
+                    List<Map<String, Object>> dataMaps = new ArrayList<>();
+
+                    final String pidName = StudyPublishService.PARTICIPANTID_PROPERTY_NAME;
+                    final String dateName = StudyPublishService.DATE_PROPERTY_NAME;
+                    final String visitName = StudyPublishService.SEQUENCENUM_PROPERTY_NAME;
+                    final String rowIdName = ExpMaterialTable.Column.RowId.toString();
+
+                    for (Map<String, Object> result : results) {
```
labkey-klum: brace styling

study/src/org/labkey/study/assay/StudyPublishManager.java
```diff
@@ -1071,6 +1079,66 @@ public ActionURL autoLinkResults(ExpProtocol protocol, ExpRun run, User user, Co
...
+                        user,
+                        container,
+                        targetContainer,
+                        sampleTypeName,
+                        Pair.of(Dataset.PublishSource.SampleType, sampleType.getRowId()),
+                        dataMaps,
+                        ExpMaterialTable.Column.RowId.toString(),
+                        publishErrors
+                    );
+                } else {
```
labkey-klum: styling

study/src/org/labkey/study/query/SampleDatasetTable.java
labkey-klum: I would change the check to : `instanceof SamplesSchema` since that's what you really want, and if it is false throw an `IllegalStateException`. You should also delete the code at 112-113, we don't need to fetch the table twice.

labkey-klum (approved):

study/src/org/labkey/study/assay/StudyPublishManager.java
RosalineP: Resolved

study/src/org/labkey/study/assay/StudyPublishManager.java
RosalineP: Resolved

experiment/src/org/labkey/experiment/ExpDataIterators.java
RosalineP: Resolved

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
RosalineP: Resolved

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
RosalineP: Ah, thanks! Resolved

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
RosalineP: Oddly I get a intellij red squiggle and then a compilation error when I attempt to make the type  `Map.Entry<String, Object>`, and `Object` is the dictated type when I try to fix it—let me know if this doesn't show up on your machine.
I've changed the cast structure a little bit to look a tad less gross, though

study/src/org/labkey/study/query/SampleDatasetTable.java
RosalineP: Resolved

study/src/org/labkey/study/assay/ExperimentListenerImpl.java
RosalineP: `deleteDatasetRows()` below requires a `Collection<String> lsids` parameter, which is why it's currently like that. Thoughts on leaving it as is, or getting with `getObject`, and creating a singleton collection with it for `deleteDatasetRows()`?

experiment/src/org/labkey/experiment/ExpDataIterators.java
RosalineP: Like this, right? (Double-checking this is what you meant)

study/src/org/labkey/study/assay/ExperimentListenerImpl.java
labkey-klum: The advantage of using getObject is if you are only expecting a single row to be returned, then there is built in validation and you can possibly avoid side effects on subsequent calls and then as you mentioned use a singleton collection to convert.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-klum: I was thinking you could declare a member variable in the class and lazily set it, since that value shouldn't change in a bulk operation?
```
                if (_study == null)
                    _study = StudyService.get().getStudy(sampleType.getAutoLinkTargetContainer());
```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-klum: Wouldn't something like this work?
```
            Collection<Map<String, Object>> selectedRow = rowIdsFromTableSelector.getMapCollection();

            // Over each column of name 'dataset<N>'
            for (Map<String, Object> selectedColumn : selectedRow)
            {
                // Check if each cell is populated (that is, if the given row was linked to a certain study)
                for (Map.Entry<String, Object> entry : selectedColumn.entrySet())
                {
                    String key = entry.getKey();
                    Object value = entry.getValue();
                    if (value instanceof Integer && linkedColumnNames.contains(key))
                    {
                        linkedDatasetsBySelectedRow.add((Integer) value);
                    }
                }
            }
```

RosalineP: Oop that works great! Not sure what I encountered earlier, this makes sense.

study/src/org/labkey/study/assay/ExperimentListenerImpl.java
RosalineP: Gotcha! Done.

experiment/src/org/labkey/experiment/ExpDataIterators.java
RosalineP: Ah gotcha! Done

PR #2245, opened 2021-05-08
Fix AssayTest check of audit message

labkey-jeckels:
#### Rationale
Update test to match new logging message with more detail

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2238
* https://github.com/LabKey/testAutomation/pull/718

#### Changes
* Update to revised audit log wording
* Don't mix/match lower-casing between DB and Java to avoid matching problem

### DIFF
--- a/study/test/src/org/labkey/test/tests/study/AssayTest.java
+++ b/study/test/src/org/labkey/test/tests/study/AssayTest.java
@@ -284,7 +284,7 @@ public class AssayTest extends AbstractAssayTest
                 "Specimen ID changed from 'AAA07XK5-05' to 'EditedSpecimenID'",
                 "Visit ID changed from '601.0' to '601.5",
                 "testAssayDataProp5 changed from blank to '514801'",
-                "Deleted data row, id ");
+                "Deleted data row.");
     }
 
     /**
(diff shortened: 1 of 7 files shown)


labkey-jeckels:
This fixes the current failure in AssayTest that I introduced in https://github.com/LabKey/platform/pull/2238

There's a separate downstream failure that I don't fully understand that was introduced with https://github.com/LabKey/platform/pull/2216. It looks like a problem with the SourceLSID value that's being stashed in the study dataset not matching the filter when querying the audit log

study/test/src/org/labkey/test/tests/study/AssayTest.java
```diff
@@ -443,41 +443,42 @@ private void uploadRuns(String folder, String asUser)
...
+//        _customizeViewsHelper.removeColumn("Run/Batch/testAssaySetProp2");
+//        _customizeViewsHelper.removeColumn("testAssayDataProp4");
         _customizeViewsHelper.applyCustomView();
 
         assertTextPresent("Blood (Whole)", 4);
 
-        Locator.XPathLocator trueLocator = Locator.xpath("//table[contains(@class, 'labkey-data-region')]//td[text() = 'true']");
-        int totalTrues = getElementCount(trueLocator);
-        assertEquals(4, totalTrues);
+//        Locator.XPathLocator trueLocator = Locator.xpath("//table[contains(@class, 'labkey-data-region')]//td[text() = 'true']");
```
labkey-kevink: Should these commented out sections be restored?

labkey-kevink (approved):

PR #2246, opened 2021-05-09
use assay run's LSID as publish audit event sourceLsid

labkey-kevink:
#### Rationale
Fix EvaluationContentTest.validateAssayFolder and AssayTest.  For assay results, use the assay run's LSID as the publish audit event's sourceLsid.  For samples, continue to use the SampleType's LSID as the sourceLsid.

#### Related Pull Requests
* #2216


labkey-kevink:
Merged into #2245

PR #2254, opened 2021-05-12
Remove unnecessary main.d.ts file

cnathe:
#### Rationale
See rationale from the related PR 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/523

#### Changes
* Remove main.d.ts file


labkey-alan (changes requested):
Shouldn't this PR also bump the ui-components version in package.json?

cnathe:
> Shouldn't this PR also bump the ui-components version in package.json?

I don't believe so. In develop, the pipeline/package.json is on 2.29.0, which is the version in labkey-ui-components where I replaced the LABKEY const with getServerContext().

labkey-alan (approved):

PR #2255, opened 2021-05-12
allow searching on any part of the displayName or email address

labkey-klum:
#### Rationale
Contains the fix for: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42611

#### Changes
- In the store model, create a field which contains the concatenated display name and email address of the user.
- Use the concatenated field as the display value, to allow searching on it, we override the rendering of the select anyway so the field contents are not visible in the UI
- All the comboBox to match on any characters instead of just from the start


internal/webapp/Impersonate.js
```diff
@@ -84,8 +85,17 @@ Ext4.define('LABKEY.Security.ImpersonateUser', {
...
                     {name: 'email', type: 'string'},
+                    {name: 'active', type: 'boolean'},
                     {name: 'displayName', type: 'string'},
-                    {name: 'active', type: 'boolean'}
+                    {name: 'concatenatedName', type: 'string',
+                        convert : function(v, record) {
+                            // concatenate both displayName and email so they can both be used
+                            // to search for users
+                            if (record && record.raw && record.raw.email && record.raw.displayName)
+                                return record.raw.email + ' ' + record.raw.displayName;
```
cnathe, line 96: I forget the exact format we use, but some other places in the UI we put the second naming string in parenthesis (something like "email (displayName)" or "displayName (email)". Do we want to do that here?

cnathe (approved):

internal/webapp/Impersonate.js
labkey-klum: Yes, that is the format, although this field is solely for the purpose to allow for querying and is never shown in the pick list. On around line 62 we introduce a custom template to render in the format you described and render it differently if the user is inactive.

cnathe: Ah ok, great. That makes sense. Thanks.

PR #2260, opened 2021-05-17
Issue 40747: Guests should not be offered the ability to save custom grid views

labkey-ankurj:
#### Rationale
This PR hides the "Save" Button in custom grid views for guests and displays correct error response of user not logged in.


labkey-adam (changes requested):
Your code changes look good for my original suggestion, but I think there's a better functional change we should make. Sorry... see issue for details.

labkey-adam (approved):
Well that's nice and simple!

PR #2262, opened 2021-05-17
Avoid error "A JSP is attempting to render a string!"

labkey-matthewb:
#### Rationale
Avoid error, it looks like HasHtmlString causes the LabKeyJspWriter.print(String) overload to be called?

java.lang.IllegalStateException: A JSP is attempting to render a string! For help rectifying this problem, review this page https://www.labkey.org/Documentation/21.3/wiki-page.view?name=premServerEncoding or contact your LabKey Account Manager.
       at org.labkey.api.jsp.LabKeyJspWriter.throwException(LabKeyJspWriter.java:59)
       at org.labkey.api.jsp.LabKeyJspWriter.print(LabKeyJspWriter.java:40)

### DIFF
--- a/api/src/org/labkey/api/view/template/PageConfig.java
+++ b/api/src/org/labkey/api/view/template/PageConfig.java
@@ -424,7 +424,7 @@ public class PageConfig
     }
 
     // For now, gives a central place to render messaging
-    public HasHtmlString renderSiteMessages(ViewContext context)
+    public HtmlString renderSiteMessages(ViewContext context)
     {
         HtmlStringBuilder messages = HtmlStringBuilder.of();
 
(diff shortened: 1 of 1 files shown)


labkey-adam:
Uh, what's the repro for this? I don't see how print(String) would ever get invoked.

That said, the current return value is not technically correct. `HasHtmlString` doesn't mean `SafeToRender`. I think the right fix is just to return `HtmlStringBuilder`, but I want to understand the repro.

labkey-matthewb:
Probably a rebuild problem???  Not clear to me why HasHtmlString is a natural return type for this method.

labkey-adam:
Yeah, it's not the natural return type. It's not causing any problems, since `HtmlStringBuilder` implements `SafeToRender` as well, but I'll fix it on a different branch for clarity.

PR #2263, opened 2021-05-17
DDIB should not output two subject columns.

labkey-matthewb:
#### Rationale
DatasetDataIteratorBuilder should not output two subject columns.
Currently it outputs the original column and the translated column, this depends on subtle behavior in matchColumns() to not break things.  Subtle is bad and we don't need both columns.

#### Related Pull Requests
https://github.com/LabKey/premium/pull/246

### DIFF
--- a/api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
@@ -54,7 +54,7 @@ public abstract class ExistingRecordDataIterator extends WrapperDataIterator
     {
         super(in);
 
-        // NOTE in might get wrapped with a LoggingDataIterator, so remember the original DataIterator
+        // NOTE it might get wrapped with a LoggingDataIterator, so remember the original DataIterator
         this._unwrapped = useMark ? (CachingDataIterator)in : null;
 
         this.target = target;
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -52,6 +53,10 @@
     ExistingRecordDataIterator(DataIterator in, TableInfo target, @Nullable Set<String> keys, boolean useMark)
     {
         super(in);
+
+        // NOTE in might get wrapped with a LoggingDataIterator, so remember the original DataIterator
```
labkey-adam:
```suggestion
        // NOTE it might get wrapped with a LoggingDataIterator, so remember the original DataIterator
```

labkey-adam (approved):

labkey-martyp (commented):
This is great, I looked into something like this as well.  I think there's one more issue though that needs to be resolved for this to work.  If you try to import ParticipantId by it's label ["Participant ID"](https://github.com/LabKey/premium/blob/develop/test/sampledata/testDatasetAppend.tsv), like the FileWatcherDatasetTest, it will fail [validation](https://github.com/LabKey/platform/blob/43fe753e50991f22716232e6390f933dc393dcee/api/src/org/labkey/api/dataiterator/StandardDataIteratorBuilder.java#L231) if that ParticipantId translate column is not always there.  Which really seems like an unintentional validation.  So not sure if we should be matching by column label as well, or if importing by column label should not be allowed and the test data just needs to be fixed.

I kicked off a bunch of tests on this branch.  Here's the failing [FileWatcher tests](https://teamcity.labkey.org/viewLog.html?buildId=1414081&tab=buildResultsDiv&buildTypeId=LabkeyTrunk_GitPostgres&branch_LabKey_Trunk_Premium=translatePtid). So far that's all that's failing.

labkey-matthewb:
I think the test is wrong.  This dataset has a column (name="ParticipantId" label="Participant Id") and a column (name ="Participant Id" and label="Participant Id").  Notice the space.  The code correctly matches the input column "Participant Id" preferentially to the column named "Participant Id" rather than the one labeled "Participant Id".

labkey-martyp (approved):
Got it. Looks good.

PR #2264, opened 2021-05-18
Issue43149: LDAP "Test" button breaks when "No Question Marks in URLs" enabled 

RosalineP:
#### Rationale
If the Experimental Feature "No Question Marks in URLs" was enabled, the 'Test' button within an LDAP configuration would lead to a malformed url.
Thanks goes to Nick for the best practice explanation and implementation here!

#### Related Pull Requests
* n/a

#### Changes
* Ensure use of '?' char after after .view using `URL`


labkey-adam (approved):
Looks okay to me, but please be sure Nick reviews it as well, given the strange language you're using here :)

labkey-nicka (approved):

PR #2267, opened 2021-05-18
matchColumns() cleanup

labkey-matthewb:
#### Rationale
Address Issue 43009: ETL fails when aliasing built-in columns
The primary way we match import columns is by name, however there is code to also match propertyuri-propertyuri (todo verify this is still used?).  The problem is that this match is treated as _higher_ priority than a name match, and is somewhat susceptible to generating duplicate/ambiguous matches (see issue).

The proposed fix is to treat a propertyuri-propertyuri match as low priority.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/2222
https://github.com/LabKey/dataintegration/pull/105
https://github.com/LabKey/platform/pull/2263


api/schemas/tableInfo.xsd
```diff
@@ -996,14 +996,23 @@
                     These properties define how the column data is imported, and the target ontology.
                 </xs:documentation>
             </xs:annotation>
-            <xs:element name="labelColumn" type="xs:string" minOccurs="0">
+            <xs:element name="subtree" type="xs:string" minOccurs="0" maxOccurs="1">
+                <xs:annotation>
+                    <xs:documentation>
+                        This is a slash ('/') separated path of concepts codes.  The values in this column are expected
```
labkey-adam, line 1002:
```suggestion
                        This is a slash ('/') separated path of concept codes.  The values in this column are expected
```

api/schemas/expTypes.xsd
```diff
@@ -589,6 +589,15 @@
 					These properties define how the column data is imported, and the target ontology.
 				</xs:documentation>
 			</xs:annotation>
+			<xs:element name="subtree" type="xs:string" minOccurs="0" maxOccurs="1">
+				<xs:annotation>
+					<xs:documentation>
+						This is a slash ('/') separated path of concepts codes.  The values in this column are expected
```
labkey-adam, line 595:
```suggestion
						This is a slash ('/') separated path of concept codes.  The values in this column are expected

```

labkey-adam (approved):

api/schemas/expTypes.xsd
labkey-matthewb: Ugh.  Must have had an unpushed commit???  I dunno, this was only supposed to have changes for DataIteratorUtil.java.

Will fix.

labkey-adam: I guess I should have asked the question...

labkey-matthewb:
created clean version

PR #2268, opened 2021-05-19
matchColumns() cleanup

labkey-matthewb:
#### Rationale
Address Issue 43009: ETL fails when aliasing built-in columns
The primary way we match import columns is by name, however there is code to also match propertyuri-propertyuri (todo verify this is still used?).  The problem is that this match is treated as _higher_ priority than a name match, and is somewhat susceptible to generating duplicate/ambiguous matches (see issue).

The proposed fix is to treat a propertyuri-propertyuri match as low priority.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/2222
https://github.com/LabKey/dataintegration/pull/105
https://github.com/LabKey/platform/pull/2263

### DIFF
--- a/api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
+++ b/api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
@@ -52,6 +52,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Spliterator;
 import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


labkey-matthewb:
Sorry, here is the clean version without the cross-branch contamination.

labkey-adam (approved):

PR #2273, opened 2021-05-20
Issue43149 Test Outages: Update snapshots and unit tests

RosalineP:
#### Rationale
I merged an Issue fix having erroneously remembered that the Core JS TC module auto-ran as a pre-merge check when the core .tsx files were touched.   🥲 👎

The fix allowed for the `<ImageAndFileAttachmentForm/>` to receive: 
`imageUrl={this.props.headerLogoUrl}`
instead of 
`imageUrl={ActionURL.getBaseURL(true) + this.props.headerLogoUrl}`. 

This led to snapshot outages since `<ImageAndFileAttachmentForm/>` had previously always been passing at least `ActionURL.getBaseURL(true)` for the imageUrl prop. 

I've scanned and verified the snapshot updates make sense, and also added a "With images attached" unit test to pair with the "No images attached" test.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2264

#### Changes
* Update snapshot
* Add "With images attached" test

### DIFF
--- a/core/src/client/components/SSOFields.tsx
+++ b/core/src/client/components/SSOFields.tsx
@@ -3,7 +3,6 @@ import { FileAttachmentForm } from '@labkey/components';
 
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faImage, faTimesCircle } from '@fortawesome/free-solid-svg-icons';
-import { ActionURL } from '@labkey/api';
 
 class NoImageSelectedDisplay extends PureComponent {
     render () {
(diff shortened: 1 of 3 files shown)


core/src/client/components/SSOFields.tsx
```diff
@@ -3,6 +3,7 @@ import { FileAttachmentForm } from '@labkey/components';
 
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faImage, faTimesCircle } from '@fortawesome/free-solid-svg-icons';
+import { ActionURL } from '@labkey/api';
```
labkey-alan: Unused import can be removed.

labkey-alan (approved):
Found an unused import that should be removed, I'm going to approve the PR, just make sure to remove that import before merging.

I think it's also worth noting that in our experience (in ui-components in particular) snapshot tests have proven to be particularly flaky, and instead it's recommended to use Enzyme and write tests that query the DOM for expected state. If you have the time I would consider converting these tests to enzyme instead of snapshots.

core/src/client/components/SSOFields.tsx
RosalineP: Removed 👍

PR #2276, opened 2021-05-20
Sample + Study Integration: Study Polish 

RosalineP:
#### Rationale
Following the implementation of the bulk of the link-to-study functionality, there are a few remaining features to implement within this polish story. These include:

- Folder export/import for sample datasets
- Sample Types export/import ‘Auto-Link Data to Study’ property 
- QC state support 
- Export Study
    - Nomenclature
    - Issue43186: In Study Export, user cannot obtain subset of dataset types with row data
- Issue43119: Layout problems on Link to Study page

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/730
* https://github.com/LabKey/compliance/pull/113

#### Changes
_Rosaline's Changes_
* Issue43119: Fix UI juddering on `publishChooseStudy.jsp` in assay run case when screen is horizontally resized
* Ensure ‘Auto-Link Data to Study’ property on Sample Type Design is preserved during Study export + import
* Update nomenclature to clarify that `DatasetDataWriter` writes dataset data while `DatasetDefinitionWriter` writes dataset design definitions
* Replace low-hanging nomenclature changes for 'CRF Datasets' to 'Study Datasets' within code and UI
* Issue43186: Support more granular selection within datasets for Study export options (more detail on this located in spec)
* Automation test changes to reflect nomenclature updates


study/src/org/labkey/study/writer/InternalStudyWriter.java
```diff
@@ -31,4 +34,8 @@ default boolean includeWithTemplate()
     {
         return true;
     }
+
+    default void write(StudyImpl study, StudyExportContext ctx, VirtualFile root, Set<String> dataTypes) throws Exception
```
RosalineP: This was added in order to call my new `write()` function in `DatasetDataWriter`, which I need in `StudyWriter` on line 71 since I'm working with a `InternalStudyWriter` and not a `DatasetDataWriter`. I wasn't sure whether the best approach here would be to refactor in this default method, or to cast on line 71 and do an `instanceof` check.

study/src/org/labkey/study/writer/SampleTypeDatasetData.java
```diff
@@ -0,0 +1,21 @@
+package org.labkey.study.writer;
+
+import org.jetbrains.annotations.Nullable;
+import org.labkey.api.writer.VirtualFile;
+import org.labkey.study.model.StudyImpl;
+
+// DatasetDataWriter actually writes all dataset data (study, assay, and sample type).  This is a do-nothing writer to get the
```
RosalineP, line 7: There are a bunch of these comments—the original one is in `AssayDatasetDefinition.java` and I copied the format, since it is confusing that these writers are around that don't write anything. I thought that repetitious comments that are closer to the actual confusing file would be better than doing a more general, modularized comment around maybe `StudySerializationRegistryImpl.getInternalStudyWriters()` but I'm open to thoughts on this.

api/src/org/labkey/api/study/publish/publishChooseStudy.jsp
```diff
@@ -71,6 +71,26 @@
                 .build());
     }
 %>
+
+<%-- Issue43119: <labkey:input type="checkbox"/> elements cause juddering and overlapping --%>
```
RosalineP, line 75: I asked about this on frontend and the advice was to not change the `<labkey:input type="checkbox"/>`, but to override styles and shift around the html. I was able to obtain a raw css solution 👍

study/src/org/labkey/study/writer/StudyWriter.java
```diff
@@ -59,15 +59,28 @@ public void write(StudyImpl study, StudyExportContext ctx, VirtualFile vf) throw
...
+                definitionWriter.write(study, ctx, vf);
+            }
 
             // Call all the writers defined in the study module.
             for (Writer<StudyImpl, StudyExportContext> writer : StudySerializationRegistryImpl.get().getInternalStudyWriters())
             {
                 String text = writer.getDataType();
 
-                if (null == text || dataTypes.contains(text) || (exportDatasets && text.endsWith("Datasets")))
+                if (null == text || dataTypes.contains(text))
```
labkey-klum, line 83: much cleaner!

study/src/org/labkey/study/writer/InternalStudyWriter.java
labkey-klum: Isn't the list of dataTypes already available in the `StudyExportContext`? Or is this a different set of data types.

labkey-klum (approved):

PR #2277, opened 2021-05-21
Item 8784: Support CRUD operations for picklist data

labkey-susanh:
#### Rationale
For LKSM we are adding the ability to create picklists of samples and various actions on these collections of samples. We need to update some permission checks for this.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/576
* https://github.com/LabKey/labkey-ui-components/pull/531

#### Changes
* Add some permission checks for operations on picklists

### DIFF
--- a/list/src/org/labkey/list/model/ListQueryUpdateService.java
+++ b/list/src/org/labkey/list/model/ListQueryUpdateService.java
@@ -53,6 +53,7 @@ import org.labkey.api.query.ValidationException;
 import org.labkey.api.reader.DataLoader;
 import org.labkey.api.security.User;
 import org.labkey.api.util.UnexpectedException;
+import org.labkey.api.view.UnauthorizedException;
 import org.labkey.api.writer.VirtualFile;
 import org.labkey.list.view.ListItemAttachmentParent;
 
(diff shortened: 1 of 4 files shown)


list/src/org/labkey/list/model/ListQueryUpdateService.java
```diff
@@ -396,6 +418,9 @@ private boolean validateProperty(DomainProperty prop, Object value, Map<String,
     @Override
     protected Map<String, Object> deleteRow(User user, Container container, Map<String, Object> oldRowMap) throws InvalidKeyException, QueryUpdateServiceException, SQLException
     {
+        if (!_list.isVisible(user))
+            throw new UnauthorizedException("You do not have permission to delete data from this table.");
```
cnathe, line 422: these two lines appear several times. might be overkill, but thoughts on having them in a `checkIfListVisible(user)` helper method?

cnathe (approved):

list/src/org/labkey/list/model/ListQueryUpdateService.java
labkey-susanh: The error messaging is slightly different in the various cases, so I will opt for no change here.

PR #2281, opened 2021-05-24
Fb sample type file watcher

labkey-klum:
#### Rationale
This introduces and registers a new file watcher to import data into a sample type from a data file per this specification:

https://docs.google.com/document/d/189GBFubU5QceCRN9_5G-uxyrv54qqlWh8PZMmPXWBvI/edit#

#### Changes
- Add a new file watcher that will either import into an existing sample type or create a new sample type by inferring the fields from the data file
- Accepts tsv, csv, xlsx, xls formatted data files
- Supports both name and ID capture groups to resolve to an existing sample type


RosalineP (approved):

RosalineP:
Reviewed and approved ✅

PR #2282, opened 2021-05-25
Issue 43200: Update default audit log behavior for dataset table

labkey-susanh:
#### Rationale
When truncating a dataset, our default audit log level is NONE but the expectation is that we will at least create a summary record.  This change overrides the default from NONE to SUMMARY.  The audit log will appear in the Query Update Events instead.  Attempts to set this audit behavior on the configParameters within `truncateRows` were not successful because we still call `getAuditBehavior` to figure out which AuditHandler to return.

#### Changes
* set default audit behavior for `DatasetDefinition`


labkey-susanh:
Gonna have to try something else.

PR #2287, opened 2021-05-25
Import performance fixes

labkey-matthewb:
#### Rationale
1. incrementSampleCounts - cache the DbSequence objects, and use getPreallocatingSequence()
2. unescapeJava - unescapeJava() is _really_ slow. Don't use if there are no backslashes.
3. queue search items in background - Don't need to synchronously expand the list of searchable items

### DIFF
--- a/api/src/org/labkey/api/reader/TabLoader.java
+++ b/api/src/org/labkey/api/reader/TabLoader.java
@@ -309,7 +309,7 @@ public class TabLoader extends DataLoader
         value = StringUtils.trimToEmpty(value);
         if ("\\N".equals(value))
             return _preserveEmptyString ? null : "";
-        if (_unescapeBackslashes)
+        if (_unescapeBackslashes && value.indexOf('\\') >= 0)   // unescapeJava() is really slow
         {
             try
             {
--- a/api/src/org/labkey/api/data/NameGenerator.java
+++ b/api/src/org/labkey/api/data/NameGenerator.java
@@ -45,6 +45,7 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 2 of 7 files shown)


labkey-matthewb:
Do we rely on unescapeJava() to do anything other than handle "\\"?

labkey-kevink:
> Do we rely on unescapeJava() to do anything other than handle "\\"?

No, it is just for handling backslash escapes for tabs, newlines, unicode characters.  This would be a breaking change, but I wonder if we should flip the default to unescapeBackslashes=false unless the tsv file is known to include the escaped characters.  We don't backslash-escape any characters on tsv export.  PostgreSQL `COPY TO` backslash escapes with the [text format](https://www.postgresql.org/docs/current/sql-copy.html#id-1.9.3.55.9.2) but doesn't for the CSV format.

labkey-kevink (approved):
Looks good

api/src/org/labkey/api/reader/TabLoader.java
```diff
@@ -309,7 +309,7 @@ protected String parseValue(String value)
         value = StringUtils.trimToEmpty(value);
         if ("\\N".equals(value))
             return _preserveEmptyString ? null : "";
-        if (_unescapeBackslashes)
+        if (_unescapeBackslashes && value.indexOf('\\') >= 0)   // unescapeJava() is really slow
```
labkey-jeckels, line 312: This change seems fairly safe. Out of curiosity though, how did you discover this? I've had methods like this show up as being very slow when profiling in instrumentation mode, but they're actually really fast in practice.

labkey-susanh (approved):

api/src/org/labkey/api/reader/TabLoader.java
labkey-matthewb: I've seen odd profiling artifacts with this before too.  In this case it was the IDEA built in profiler.  Either way, I don't have to look at it any more!

labkey-nicka:
Just curious, what kind of performance gains were you seeing here versus your benchmarks?

labkey-matthewb:
> Just curious, what kind of performance gains were you seeing here versus your benchmarks?

Depends on what you're measuring.  We were spending a measurable amount of time queuing search objects in the foreground thread.  Moved to background.  So same amount of work, but the user doesn't need to wait for that.

Also the DbSequence stuff was pretty significant (tens of percent?) of foreground thread time (after removing the indexing).

This was all based on profiling.  I did not do a clock time comparison, but it certainly seemed faster from the browser.

PR #2289, opened 2021-05-26
Issue43218: Allow MultiAuthConfigs to specify "dictateFieldVisibility"

RosalineP:
#### Rationale
We now have a second checkbox on the LDAP authentication configuration page ("Read LDAP attributes") that needs to reveal additional properties when checked. The original code was only configured to reveal additional properties for one checkbox—this update allows for any number of 'dictateFieldVisibility' checkboxes.

#### Related Pull Requests
* https://github.com/LabKey/ldap/pull/46

#### Changes
* Determine whether a field should be displayed according to relevant field's 'dictateFieldVisibility' setting

### DIFF
--- a/core/src/client/components/DynamicConfigurationModal.tsx
+++ b/core/src/client/components/DynamicConfigurationModal.tsx
@@ -31,7 +31,7 @@ export default class DynamicConfigurationModal extends PureComponent<Props, Part
         const {authConfig, modalType} = this.props;
         const fieldValues: any = {};
         this.props.modalType.settingsFields.forEach(field => {
-            const value = this.props.authConfig[field.name] != null ? this.props.authConfig[field.name] : "";
+            const value = this.props.authConfig[field.name] != null ? this.props.authConfig[field.name] : '';
             fieldValues[field.name] = field.name in this.props.authConfig ? value : field.defaultValue;
         });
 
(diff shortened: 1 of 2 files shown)


core/src/client/components/DynamicConfigurationModal.tsx
```diff
@@ -31,7 +31,8 @@ export default class DynamicConfigurationModal extends PureComponent<Props, Part
         const {authConfig, modalType} = this.props;
         const fieldValues: any = {};
         this.props.modalType.settingsFields.forEach(field => {
-            fieldValues[field.name] = field.name in this.props.authConfig ? this.props.authConfig[field.name] : field.defaultValue;
+            const value = this.props.authConfig[field.name] != null ? this.props.authConfig[field.name] : "";
```
RosalineP: Avoids React warning about rendering `null` in a text field

labkey-alan: Yeah I really wish react would just do this for you.

core/src/client/components/DynamicConfigurationModal.tsx
```diff
@@ -66,7 +67,10 @@ export default class DynamicConfigurationModal extends PureComponent<Props, Part
         }
 
         Object.keys(this.state.fieldValues).map(item => {
-            form.append(item, this.state.fieldValues[item]);
+            const itemValue = this.state.fieldValues[item];
+            if (itemValue !== null && itemValue !== "") {
```
labkey-alan: You may want to do `itemValue.trim() !== ''`

labkey-alan (approved):
Looks good, left one minor note.

core/src/client/components/DynamicConfigurationModal.tsx
RosalineP: Good point; done!

PR #2295, opened 2021-05-26
Item #8853: Ontology, Form insert and update

Sigmonia:
#### Rationale
Item #8853: Ontology, Form insert and update

#### Related Pull Requests
* https://github.com/LabKey/ontology/pull/43

#### Changes
* Adds the ConceptSubtree DomainProperty so we can apply a specific ontology hierarchy path to a field


api/src/org/labkey/api/data/ColumnRenderPropertiesImpl.java
```diff
@@ -865,64 +867,79 @@ public void setPrecision(int precision)
...
-    public String getConceptLabelColumn()
+    public String getSourceOntology()
     {
-        return _conceptLabelColumn;
+        return _sourceOntology;
     }
 
     @Override
-    public void setConceptLabelColumn(String conceptLabelColumn)
+    public void setSourceOntology(String sourceOntology)
```
cnathe, line 904: I noticed that several of these ontology related setters don't have the `assert _checkLocked();` like the others. Not likely part of this PR, but is that intentional for any reason that you know of? or should they be added?

api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -1564,15 +1564,47 @@ private static PropertyDescriptor ensurePropertyDescriptor(PropertyDescriptor pd
...
+            out[0] = (new SqlSelector(getExpSchema(), reselect).getObject(PropertyDescriptor.class));
+            return rowcount;
+        }
+        catch(SQLException sqlx)
+        {
+            throw ExceptionFramework.Spring.translate(getExpSchema().getScope(), "insertPropertyIfNotExists", sqlx);
+        }
+
+
+/*        TableInfo t = get();
```
cnathe: There is a lot of commented out code left here. Is that to be removed at this point?

experiment/resources/schemas/dbscripts/sqlserver/exp-21.004-21.005.sql
```diff
@@ -0,0 +1 @@
+--ALTER TABLE exp.PropertyDescriptor ADD ConceptSubtree NVARCHAR(MAX) NULL;
```
cnathe: I assume we do want to support this in sql server. Is there a reason this is commented out?

cnathe (changes requested):
I reviewed the changes and made a few comments, but @labkey-matthewb I think it would be good if you took a look at the changes for `OntologyManager.java` and the related classes. I am not familiar with that code.

cnathe:
> I reviewed the changes and made a few comments, but @labkey-matthewb I think it would be good if you took a look at the changes for `OntologyManager.java` and the related classes. I am not familiar with that code.

Looking at the commits, maybe you were the one that made those changes @labkey-matthewb. Is that correct?

labkey-matthewb (approved):

cnathe (approved):

api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -2355,88 +2271,95 @@ public static PropertyDescriptor insertOrUpdatePropertyDescriptor(PropertyDescri
     static final String parameters = "propertyuri,name,description,rangeuri,concepturi,label," +
```
Sigmonia, line 2286: @mbellew There are some fields that are no longer being included in the insert statement. Should the be added here?

Sigmonia: facetingbehaviortype,
phi,
redactedText,
excludefromshifting,
mvindicatorstoragecolumnname,
defaultscale,

labkey-matthewb: This was an attempt to make the code _more_ consistent instead of repeating the columns in three places.  I didn't realize that insertPropertyIfNotExists() handled more columns than insertPropertyDescriptors().  I can't imagine that's intentional.  I assume that's genetic drift?

PR #2296, opened 2021-05-27
Fix sample type scoping issues for link to study

labkey-klum:
#### Rationale
This [issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43233) uncovered some problems in the sample type integration to study that were related to whether a sample type definition was scoped at the project or shared folder level AND data was uploaded and linked from a different folder.

- Error in publish start view, could not resolve the sample type
- Recall audit events were scoped to the wrong container (sample type definition vs data)
- Publish history view couldn't always resolve the source sample type lookup

#### Changes
- When resolving a sample type, use this API instead: `SampleTypeService..getSampleType(container, user, rowId)` since this variant will also check project and shared folders in addition to current folder.
- To resolve the sourceLsid container for samples, we now pass in an optional sourceRowId which represents the source sample row and translates to an `ExpMaterial` this gives us the container that the data was linked from instead of the container the that sample type definition lives in.
- Expand the container filter when resolving sample type lookups since they won't necessarily live in the same container that the sample data has been uploaded into.


api/src/org/labkey/api/study/Dataset.java
```diff
@@ -124,7 +125,7 @@ public boolean hasUsefulDetailsPage(Integer publishSourceId)
                     }
 
                     @Override
-                    public @Nullable Container resolveSourceLsidContainer(String sourceLsid)
+                    public @Nullable Container resolveSourceLsidContainer(String sourceLsid, Integer sourceRowId)
                     {
                         // for assays the source lsid is the run
```
labkey-klum, line 130: For assays we don't need to do anything additional with the source row ID since the run is properly scoped to the source container.

study/src/org/labkey/study/controllers/publish/SampleTypePublishConfirmAction.java
```diff
@@ -59,15 +57,12 @@ public ExpSampleType getSampleType()
...
-                    _sampleType = SampleTypeService.get().getSampleType(getContainer(), getRowId());
+                    _sampleType = SampleTypeService.get().getSampleType(getContainer(), getUser(), getRowId());
                 else
                     throw new NotFoundException("Sample Type ID not specified.");
 
                 if (_sampleType == null)
                     throw new NotFoundException("Sample Type " + getRowId() + " does not exist in " + getContainer().getPath());
-
-                if (!_sampleType.getContainer().hasPermission(getViewContext().getUser(), ReadPermission.class))
-                    throw new UnauthorizedException();
```
labkey-klum, line 70: We don't need to permission check any longer since the variant of the `getSampleType` we are using will check for read access on all of the containers it considers.

study/src/org/labkey/study/query/SampleDatasetTable.java
```diff
@@ -113,7 +113,7 @@ private TableInfo getSamplesTable()
             // Hide 'linked' column for Sample Type Datasets
             if (userSchema instanceof SamplesSchema)
             {
-                _sampleTable = ((SamplesSchema) userSchema).getSampleTable(sampleType, null);
+                _sampleTable = ((SamplesSchema) userSchema).getSampleTable(sampleType, ContainerFilter.EVERYTHING);
```
labkey-klum, line 116: `ContainerFilter.EVERYTHING` drops all container where clauses

study/src/org/labkey/study/assay/query/PublishAuditProvider.java
```diff
@@ -127,7 +127,7 @@ else if (COLUMN_NAME_SAMPLE_TYPE_ID.equalsIgnoreCase(col.getName()))
                 {
                     // lookup to SampleType by ID
                     col.setLabel("Sample Type ID");
-                    col.setFk(QueryForeignKey.from(getUserSchema(), cf).schema(ExpSchema.SCHEMA_NAME).table(ExpSchema.TableType.SampleSets));
+                    col.setFk(QueryForeignKey.from(getUserSchema(), ContainerFilter.EVERYTHING).schema(ExpSchema.SCHEMA_NAME).table(ExpSchema.TableType.SampleSets));
```
labkey-klum, line 131: We need this in the case where we are trying to resolve a sample type that is defined in a different folder from that data that was linked.

study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -2827,15 +2827,19 @@ public boolean handlePost(DeleteDatasetRowsForm form, BindException errors)
             // Need to handle this by groups of source lsids -- each assay or SampleType container needs logging
             MultiValuedMap<String,Pair<String,Integer>> sourceLsidToLsidPair = new ArrayListValuedHashMap<>();
             List<Map<String,Object>> data = def.getDatasetRows(getUser(), allLsids);
+            Integer sourceRowId = null;
             for (Map<String,Object> row : data)
             {
                 String sourceLSID = (String)row.get("sourcelsid");
                 String datasetRowLsid = (String)row.get("lsid");
-                Integer sourceRowId = (Integer)row.get("rowId");
+                Integer rowId = (Integer)row.get("rowId");
```
RosalineP: Is there a place we could pull these string constants from, or is raw strings typically the way to go for these values?

RosalineP (approved):
Minor comment, otherwise all is looking good!

study/src/org/labkey/study/controllers/StudyController.java
labkey-klum: 👍 found some in `StudyPublishService`

PR #2303, opened 2021-05-28
ParameterMapStatement exceptions

labkey-matthewb:
#### Rationale
Make ParameterMapStatement easier to use and more compatible with SqlExecutor by using ExceptionFramework.

#### Related Pull Requests
* 

#### Changes
*


api/src/org/labkey/api/data/ParameterMapStatement.java
```diff
@@ -42,70 +43,108 @@
...
+        _exceptionFramework = from._exceptionFramework;
+        try
+        {
+            _stmt = createStatement(_conn, _sqlf);
+            for (int i = 0; i < from._parameters.length; i++)
+                _parameters[i] = from._parameters[i].copy(_stmt);
+        }
+        catch (SQLException x)
+        {
+            throw _exceptionFramework.translate(_scope, "Duplicate statement", x);
```
labkey-kevink, line 99: Is this the right task description?

labkey-kevink (approved):
Nice cleanup of SQLExceptions

api/src/org/labkey/api/data/ParameterMapStatement.java
labkey-matthewb: That does look odd.  That's "duplicate" as in "copy"...

labkey-matthewb: Shouldn't really ever fail, since we've already successfully created one statement using the same SQL.

PR #2310, opened 2021-06-01
43241: display of dates from multi-value columns

labkey-kevink:
#### Rationale
The fix in PR #2298 worked for standard MultiValueFKs but not the lineage MVFK.  The inner display column needs to have `prepare()` called to apply the container's date and number formats.  In addition, we were rewrapping MultiValueDisplayColumn with another MultiValueDisplayColumn when calling `getRenderer()` which resulted in the values being rendered as nested arrays in the JSON response.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2298

#### Changes
- Don't wrap the display column with a MultiValuedDisplayColumn twice
- Call DisplayColumnDecorator.prepare() to apply container number and date formats

### DIFF
--- a/api/src/org/labkey/api/data/DisplayColumnDecorator.java
+++ b/api/src/org/labkey/api/data/DisplayColumnDecorator.java
@@ -605,6 +605,12 @@ public class DisplayColumnDecorator extends DisplayColumn
     }
 
     @Override
+    public void prepare(Container c)
+    {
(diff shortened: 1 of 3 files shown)


labkey-matthewb (approved):

labkey-susanh (approved):
Looks good.  Issue verified in this branch.

PR #2311, opened 2021-06-01
Fb assay file watcher

labkey-klum:
#### Rationale
This PR adds support for importing assay data via a file watcher. This initial development adds the ability to import data into a GPAT assay as well as module based assays (file-based). We will support three data representations:
- Results data only via : tsv, csv, xlsx and xls files.
- Multi-sheet excel format where results, batch and run properties are specified in separate sheets.
- Zip archive format where results, batch, run properties, and plate metadata are furnished in separate files.

Details can be found in the spec [here](https://docs.google.com/document/d/1OqCKqUWbPUjYkSKxDL4hnrb67ztJs3ZgtmzOAoFgpns/edit#).

#### Changes
- Extend the `AssayImportRunTask` to support the above scenarios
- Register a file analysis task which uses the same `AssayImportRunTask` that script based pipelines use but add a different factory : `AssayImportRunTask$FileAnalysisFactory` to handle the parsing and processing of the file watcher specific data file types.
- Add custom file watcher UI to capture the assay protocol (and existing UI to capture the protocol name), these both leverage the existing capability of the `AssayImportRunTask` to resolve protocol and providers.
- Add the ability to resolve the protocol and provider through a name capture group in the file pattern regex, both name and ID are supported.


api/src/org/labkey/api/assay/TsvDataHandler.java
```diff
@@ -63,7 +63,7 @@
 
     static
     {
-        FileType fileType = new FileType(Arrays.asList(".tsv", ".csv", ".xls", ".xlsx", ".txt", ".fna", ".fasta"), ".tsv");
+        FileType fileType = new FileType(Arrays.asList(".tsv", ".csv", ".xls", ".xlsx", ".txt", ".fna", ".fasta", ".zip"), ".tsv");
```
labkey-klum, line 66: Don't really like this since we didn't natively add the ability to process .zip files through the UI, but this was needed to allow for the zip archive scenario.

api/src/org/labkey/api/reader/ExcelLoader.java
```diff
@@ -437,16 +437,20 @@ public void close()
             XSSFReader xssfReader = new XSSFReader(xlsxPackage);
             StylesTable styles = xssfReader.getStylesTable();
             XSSFReader.SheetIterator iter = (XSSFReader.SheetIterator) xssfReader.getSheetsData();
-            if (iter.hasNext())
+            while (iter.hasNext())
```
labkey-klum, line 440: This was needed to add support for parsing a named sheet for XLSX formats, we already supported this for XLS formats.

assay/src/org/labkey/assay/pipeline/AssayImportRunTask.java
```diff
@@ -102,6 +115,358 @@ public TaskFactory create(TaskId taskId, TaskType xobj, Path taskDir)
...
+            }
+            catch (IOException e)
+            {
+                throw new PipelineJobException(e);
+            }
+        }
+
+        private Map<String, Object> loadProperties(DataLoader loader) throws PipelineJobException
+        {
+            Map<String, Object> properties = new HashMap<>();
```
labkey-kevink: This should probably be a CaseInsensitiveHashMap

labkey-kevink (approved):
Looks good.  I tested it out with a multisheet xlsx file and it worked the first time. :)

assay/src/org/labkey/assay/pipeline/AssayImportRunTask.java
labkey-klum: 👍

PR #2312, opened 2021-06-02
Project locking compliance feature

labkey-adam:
#### Rationale
This introduces a project locking feature for archival and compliance purposes. Administrators can manually lock a project, which makes it inaccessible to non-administrators (i.e., non-admin role assignments in every folder of the project are ignored). Administrators can unlock a project to re-enable all the role assignments.

#### Related Pull Requests
* https://github.com/LabKey/compliance/pull/116

#### Changes
* Add `LockState` and `ExpirationDate` columns to `core.Containers` table, `Container`, and `ContainerFactory`
* During permission checks, invoke current `LockedProjectHandler` if the project's `LockState` is non-null*

### DIFF
--- a/api/src/org/labkey/api/data/ContainerManager.java
+++ b/api/src/org/labkey/api/data/ContainerManager.java
@@ -729,6 +729,18 @@ public class ContainerManager
         _removeFromCache(container);
     }
 
+    public static void updateLockState(Container container, LockState lockState, User user)
+    {
+        //For some reason there is no primary key defined on core.containers
+        //so we can't use Table.update here
+        StringBuilder sql = new StringBuilder("UPDATE ");
+        sql.append(CORE.getTableInfoContainers());
+        sql.append(" SET LockState = ? WHERE RowID = ?");
+        new SqlExecutor(CORE.getSchema()).execute(sql, lockState, container.getRowId());
+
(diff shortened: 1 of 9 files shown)


api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -727,6 +729,18 @@ public static void updateSearchable(Container container, boolean searchable, Use
...
     }
 
+    public static void updateLockState(Container container, LockState lockState, User user)
+    {
+        //For some reason there is no primary key defined on core.containers
+        //so we can't use Table.update here
+        StringBuilder sql = new StringBuilder("UPDATE ");
+        sql.append(CORE.getTableInfoContainers());
+        sql.append(" SET LockState = ? WHERE RowID = ?");
+        new SqlExecutor(CORE.getSchema()).execute(sql, lockState, container.getRowId());
```
labkey-klum, line 739: Should we be doing permission checking, or is that planned when we hook it to a controller action?

labkey-klum (approved):

api/src/org/labkey/api/data/ContainerManager.java
labkey-klum: nvm, just looked at the compliance code 👍

PR #2314, opened 2021-06-02
AuditTypeProvider.getDomainKind() perf

labkey-matthewb:
#### Rationale
Some code paths call getDomainKind() a lot (sample import).  This can get a little expensive mostly due to the implementation of AbstractAuditDomainKind.createPropertyDescriptor().  We can avoid all this just by holding a singleton DomainKind object.


labkey-matthewb:
This should have only had the *AuditTypeProvider.java changes.  (forgot to checkout develop before creating branch)

labkey-klum (approved):

PR #2317, opened 2021-06-02
Issue43234: Support '(Data import folder)' auto-link option for Sample Types

RosalineP:
#### Rationale
Previously, it was not possible to create a shared Sample Type, select the auto-import option equivalent to 'link the values into the study the Sample Type is currently in', (this option has been renamed to "(Data import folder)"), add rows to the Sample Type, and see the new Sample Type Dataset appear. This functionality has now been implemented, along with some minor nomenclature changes.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2296

#### Changes
* Update `AUTO_LINK_TARGET_ASSAY_IMPORT_FOLDER` => `AUTO_LINK_TARGET_IMPORT_FOLDER`, as the  shared container is used by both assay and sample type use cases
* Set `_study` appropriately in `ExpDataIterators` for '(Data import folder)' case
* Make similar change as above bullet point in `StudyPublishManager.autoLinkResults()`
* Change "(Assay import folder)" to "(Data import folder)" and update tests


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -353,6 +354,9 @@ protected AutoLinkToStudyDataIterator(DataIterator di, DataIteratorContext conte
                 if (_study == null && targetContainer != null && studyService != null) {
                     _study = studyService.getStudy(targetContainer);
                 }
+                // Issue43234: Support '(Data import folder)' auto-link option
+                if (_study == null && targetContainer == ContainerManager.getSharedContainer() && studyService != null)
```
labkey-klum: replace with `StudyPublishService.AUTO_LINK_TARGET_IMPORT_FOLDER`?

labkey-klum (approved):

experiment/src/org/labkey/experiment/ExpDataIterators.java
RosalineP: Good call, done!

PR #2319, opened 2021-06-03
batch updates for material.setLastIndexed()

labkey-matthewb:
#### Rationale
On bulk insert/update we reindex affected samples.  Avoid updating lastindexed property one row at a a time.


api/src/org/labkey/api/util/JobRunner.java
```diff
@@ -126,6 +126,11 @@ public void execute(Runnable command)
      * @param command
      * @param delay
      */
+    public void execute(long delay, Runnable command)
```
labkey-susanh, line 129: Do we have to have two methods with just a change in the order of the parameters?

experiment/src/org/labkey/experiment/api/ExpMaterialImpl.java
```diff
@@ -372,11 +374,32 @@ public WebdavResource createIndexDocument()
...
             public void setLastIndexed(long ms, long modified)
             {
-                ExperimentServiceImpl.get().setMaterialLastIndexed(getRowId(), ms);
+                // setLastIndexed() can get called very rapidly after a bulk insert/update
+                // so we'll collect these instead of updating one at a time
+                synchronized (updateLastIndexedList)
+                {
+                    boolean wasEmpty = updateLastIndexedList.isEmpty();
+                    updateLastIndexedList.add(new Pair<>(getRowId(), ms));
+                    if (wasEmpty)
```
labkey-susanh, line 383: Clever.

labkey-susanh (approved):

api/src/org/labkey/api/util/JobRunner.java
labkey-matthewb: The original method pre-dates lambdas.  Methods with a lambda really work best if it is the last parameter.  (could go find all the old calls...)

PR #2321, opened 2021-06-03
Issues 43021, 43022: Changes to better support the use of import parent aliases for samples and name expressions

labkey-susanh:
#### Rationale
We allow users to specify import aliases for samples, which allows them to avoid the clumsy technical syntax such as MaterialInputs/Lab in when creating samples.  When users try to incorporate the names of the parents in a name expression using the usual syntax such as `DataInputs:first`, they will either get an error or a name that is missing data if the import actually uses the import alias.  The only way to get this name expression to work is to revert to not using the import alias when creating the samples.  Also, the use of `DataInputs` and `MaterialInputs` is not intuitive for users who have provide import aliases.  It is more natural for them to be able to provide the import aliases in their name expressions.  These changes allow for both possibilities. 
 
* [Issue 43021](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43021): LKSM: Name expression to create names directly from particular Sources or Parent Samples
* [Issue 43022](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43022): Using Source alias and ${DataInputs:first} for name expression will cause an error 

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/753

#### Changes
* Change `NameGenerator `to be able to take into account an optional set of import aliases
* Update `UploadSamplesHelper` to provide the parent import aliases to the name generator
* Update some error messaging to give users a bit more actionable text.


experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -827,7 +830,20 @@ else if (aliquotedFromObj instanceof Number)
                     generatedName = aliquotName;
                 }
                 else
-                    generatedName = nameGen.generateName(nameState, map);
+                {
+                    Supplier<Map<String, Object>> extraPropsFn = () -> {
+                        try
+                        {
+                           return Map.of(PARENT_IMPORT_ALIAS_MAP_PROP, sampletype.getImportAliasMap());
```
labkey-kevink: Given that the sample type is the same for all the rows, we could parse the import alias map once in the _GenerateNamesDataIterator constructor.

labkey-kevink (approved):

PR #2325, opened 2021-06-04
Issue 43246: Lineage query NPE while processing an UploadedFile

labkey-kevink:
Prior to fixing Issue 41675, it was possible that a UploadedFile exp.data
imported into an assay could create a new exp.data and exp.object with
a different LSID and leave the old exp.object in the edge table.  The exp.edge
table and the run ended up out of sync.

This PR finds all orphaned UploadedFile exp.object that are outputs of a run,
rebuilds the runs, and removes the orphaned exp.object rows.

#### Related Issues:
- Issue 41675: exp.object deleted when detaching exp.data from run
- Issue 41675: allow updating the LSID of an existing exp.data

#### Related Pull Requests:
- #1741
- #1758

#### Changes:
- upgrade script to remove orphaned UploadedFile exp.objects and rebuild edge table
- add experiment-verifyRunEdges.api utility action

### DIFF
--- a/api/src/org/labkey/api/exp/OntologyManager.java
+++ b/api/src/org/labkey/api/exp/OntologyManager.java
@@ -811,7 +811,7 @@ public class OntologyManager
     }
 
 
-    public static int deleteOntologyObjects(DbSchema schema, SQLFragment sub, Container c, boolean deleteOwnedObjects)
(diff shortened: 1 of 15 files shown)


labkey-jeckels (approved):
Changes look OK, but these kinds of upgrade scripts are fragile when the underlying schema changes and the Java code is updated to the new variant, but the code runs before the schema has been completely migrated.

Given the potentially slow run time, what do you think about moving this to a pipeline job that's queued by the upgrade? The server should be useable while it's running, correct?

PR #2327, opened 2021-06-07
Issue43219: Move "System Default Domain" to Authentication page

RosalineP:
#### Rationale
The second property on the "Site Settings" page is called "System default domain (default domain for user log in)". This is really an authentication property and should be on the Authentication configuration page. Moving it there would make it more discoverable.

#### Related Pull Requests
* https://github.com/LabKey/ldap/pull/51
* https://github.com/LabKey/itn-tslk-modules/pull/11

#### Changes
* Refactor calls of `AppProps.getDefaultDomain()` to use `AuthenticationManager.getDefaultDomain()`
* Create `AuthenticationManager.getDefaultDomain()`
* Refactor calls of `WriteableAppProps.setDefaultDomain()` to `AuthenticationManager.setDefaultDomain()`
* Create `AuthenticationManager.setDefaultDomain()`
* Add deferred Java upgrade code that reads the old property value and saves it to the new property set
* Update front-end


api/src/org/labkey/api/security/AuthenticationManager.java
```diff
@@ -257,19 +272,27 @@ public static void saveAuthSetting(User user, String key, boolean value)
...
+        Map<String, Object> changed = map.entrySet().stream()
+                .filter(e->!e.getValue().equals(props.get(e.getKey())))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
 
         if (!changed.isEmpty())
         {
             changed.forEach((k, v)->{
-                props.put(k, Boolean.toString(v));
-                addAuthSettingAuditEvent(user, k, v ? "enabled" : "disabled");
+                boolean vIsBoolean = v instanceof Boolean;
```
RosalineP: Not very neat but my shot at wrapping the save into saveAuthSettings(). Alternatively, I could pull out saving the string into a separate function?

api/src/org/labkey/api/settings/AppPropsImpl.java
```diff
@@ -626,7 +620,6 @@ public static void populateSiteSettingsWithStartupProps()
                 LOG.debug("Setting site settings config property '" + prop.getName() + "' to '" + prop.getValue() + "'");
                 switch (prop.getName())
                 {
-                    case DEFAULT_DOMAIN_PROP -> writeable.setDefaultDomain(prop.getValue());
```
RosalineP, line 629: Is this still needed in this location, actually, since it's still a 'site setting' that gets populated here?

core/src/client/components/GlobalSettings.tsx
```diff
@@ -22,17 +24,52 @@ const ROW_TEXTS = [
...
+class DefaultDomainField extends PureComponent<DefaultDomainProps> {
+    render() {
+        const { globalAuthOnChange, defaultDomain } = this.props;
+
+        return (
+            <div className="global-settings__default-domain">
+                System Default Domain
+                <LabelHelpTip title="Tip">
+                    <div>
+                        <div> Default domain for user log in.</div>
```
RosalineP: Note to Adam: Please check in the UI that the tooltip helptext looks good to you

core/src/org/labkey/core/CoreUpgradeCode.java
```diff
@@ -201,4 +206,49 @@ public void migrateSecondaryAuthenticationConfigurations(ModuleContext context)
...
+        {
+            // Removed from AppPropsImpl
+            final String DEFAULT_DOMAIN_PROP = "defaultDomain";
+
+            String defaultDomainFromAppProps = lookupStringValue(DEFAULT_DOMAIN_PROP, "");
+            AuthenticationManager.setDefaultDomain(defaultDomainFromAppProps);
+        }
+    }
+
+    public Map<String, String> getProperties(Container c)
```
RosalineP: I aggressively moved all necessary calls to help functions into the method—let me know if this is not what you meant when we talked about this!

RosalineP:
@labkey-adam 
I checked around and it looks like the other core executeJavaUpgradeCodes don't bump a value in `getSchemaVersion()` within `CoreModule` as is done with general upgrade scripts for other modules. Is the core module a special case?

**Edit:**
Nevermind, Josh cleared this up for me by pointing out the `SchemaVersion` setting in `core/module.properties`!

core/src/client/AuthenticationConfiguration/AuthenticationConfiguration.tsx
```diff
@@ -95,21 +101,20 @@ export class App extends PureComponent<{}, Partial<State>> {
...
-        const newState = {...this.state.globalSettings, [id]: !oldState };
-
-        this.setState(
-            prevState => ({
-                ...prevState,
-                globalSettings: { ...newState },
-                dirty: !isEquivalent(newState, globalSettingsDirtinessData)
-            })
-        );
+        const newState = { ...this.state.globalSettings, [id]: value };
```
labkey-alan: This should be computed inside the this.setState call below and should reference `prevState` instead of `this.state` when doing so, that way you are guaranteed to be calculating state on the most up to date state.

labkey-alan (approved):
From a frontend perspective it looks good, though there is a setState call that should probably be changed to compute state via the state object passed to setState just to be safe.

core/src/client/AuthenticationConfiguration/AuthenticationConfiguration.tsx
RosalineP: Oops, thanks for the catch. Done!

api/src/org/labkey/api/security/AuthenticationManager.java
labkey-adam: Eh, I wouldn't worry about it... seems fine to me

api/src/org/labkey/api/settings/AppPropsImpl.java
labkey-adam: Good catch. Our syseng team is likely setting this property on hosted instances, so we need to coordinate with them. We need to email them know that, as of 21.7, there's a new way to set the default domain via startup properties (and we should verify that it works). It would be nice if this code would remain backward compatible for a while... it should accept the old property name but call the new AuthenticationManager setter AND log a warning stating its going away and listing the correct approach.

core/src/org/labkey/core/CoreUpgradeCode.java
labkey-adam: Since this is a `@DeferredUpgrade`, it's safe to call into `AppPropsImpl`... but you'd have to make lookupStringValue() `public`. Oh, and AppPropsImpl isn't public. Here's my suggestion (not tested, but should work):

```
        String defaultDomain = (new AbstractWriteableSettingsGroup(){
            @Override
            protected String getGroupName()
            {
                return "site settings";
            }

            @Override
            protected String getType()
            {
                return SITE_CONFIG_NAME;
            }

            private String getDefaultDomain()
            {
                return lookupStringValue(DEFAULT_DOMAIN_PROP, "");
            }
        }).getDefaultDomain();
```

core/src/org/labkey/core/admin/customizeSite.jsp
```diff
@@ -177,13 +177,9 @@ Click the Save button at any time to accept the current settings and continue.</
 </tr>
 
 <tr>
-    <td colspan=2>Set default domain for user sign in and base server url (<%=bean.helpLink%>)</td>
+    <td colspan=2>Set base server url (<%=bean.helpLink%>)</td>
```
labkey-adam:
```suggestion
    <td colspan=2>URL settings (<%=bean.helpLink%>)</td>

```

labkey-adam (changes requested):

core/src/client/components/GlobalSettings.tsx
labkey-adam:
```suggestion
                        <div> Default domain for user sign in.</div>
```

core/src/org/labkey/core/CoreUpgradeCode.java
RosalineP: Done!

api/src/org/labkey/api/settings/AppPropsImpl.java
RosalineP: Done!

RosalineP:
Misclicked and accidentally re-requested from Alan; feel free to ignore that.

api/src/org/labkey/api/settings/AppPropsImpl.java
```diff
@@ -626,7 +621,10 @@ public static void populateSiteSettingsWithStartupProps()
                 LOG.debug("Setting site settings config property '" + prop.getName() + "' to '" + prop.getValue() + "'");
                 switch (prop.getName())
                 {
-                    case DEFAULT_DOMAIN_PROP -> writeable.setDefaultDomain(prop.getValue());
+                    case DEFAULT_DOMAIN_PROP -> {
+                        AuthenticationManager.setDefaultDomain(prop.getValue());
+                        LOG.warn("DEFAULT_DOMAIN_PROP here will be deprecated in 21.11 in favor of using DEFAULT_DOMAIN in AuthenticationManager.populateSettingsWithStartupProps()");
```
labkey-adam:
```suggestion
                        LOG.warn("Support for the \"SiteSettings.defaultDomain\" startup property will be removed shortly; use \"Authentication.DefaultDomain\" instead.");

```

core/src/org/labkey/core/login/LoginController.java
```diff
@@ -1931,7 +1932,7 @@ public boolean handlePost(SetPasswordForm form, BindException errors) throws Exc
                         {
                             String defaultDomain = userEmailAddress.substring(atSign + 1);
                             WriteableAppProps appProps = AppProps.getWriteableInstance();
```
labkey-adam: No need to touch appProps... remove this line

core/src/org/labkey/core/login/LoginController.java
```diff
@@ -1931,7 +1932,7 @@ public boolean handlePost(SetPasswordForm form, BindException errors) throws Exc
...
-                            appProps.setDefaultDomain(defaultDomain);
+                            AuthenticationManager.setDefaultDomain(defaultDomain);
                             appProps.save(null);
```
labkey-adam: ...and remove this line

labkey-adam (changes requested):
A bit of cleanup...

core/src/org/labkey/core/login/LoginController.java
RosalineP: Done!

core/src/org/labkey/core/login/LoginController.java
RosalineP: Done!

labkey-adam (approved):

PR #2330, opened 2021-06-08
Issue 43279: XML Metadata: GUI editor loses previous changes

labkey-jeckels:
#### Rationale
TableInfos should generally be created via UserSchema.getTable() which applies XML metadata by default, but callers can opt-out when we're wanting to diff the configuration against the "real" table as opposed to the customized table

Sample types and data classes were always applying the XML customizations, messing up the diffing, and making subsequent edits destructive of prior changes

#### Changes
* Let getTable() control how XML metadata is applied
* Make it harder to call methods that are for internal schema implementation

### DIFF
--- a/api/src/org/labkey/api/exp/query/SamplesSchema.java
+++ b/api/src/org/labkey/api/exp/query/SamplesSchema.java
@@ -132,18 +132,19 @@ public class SamplesSchema extends AbstractExpSchema
         if (st == null)
             return null;
 
+        AbstractTableInfo tableInfo = (AbstractTableInfo) createSampleTable(st, cf);
+
         // Get linked to study columns
         StudyPublishService studyPublishService = StudyPublishService.get();
         if (studyPublishService != null)
         {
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/exp/query/SamplesSchema.java
```diff
@@ -132,18 +132,19 @@ public TableInfo createTable(String name, ContainerFilter cf)
...
 
+        AbstractTableInfo tableInfo = (AbstractTableInfo) createSampleTable(st, cf);
+
         // Get linked to study columns
         StudyPublishService studyPublishService = StudyPublishService.get();
         if (studyPublishService != null)
         {
-            AbstractTableInfo tableInfo = (AbstractTableInfo) getSampleTable(st, cf);
-            int rowId = getSampleTypes().get(tableInfo.getName()).getRowId();
+            int rowId = st.getRowId();
```
RosalineP, line 141: Ah, thanks for updating this

RosalineP (approved):
Reviewed, and functionality verified.

PR #2336, opened 2021-06-09
Fix S3ImportTTestt test outage

RosalineP:
#### Rationale
Importing a folder from a template should not retain dataset data info, but as of an update to folder export, dataset data was being included within 'Folder Management > Export > Folder objects to export'. 

#### Related Pull Requests
* n/a

#### Changes
* Specifically within folder template export, do not add any 'Dataset Data' writers

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -4690,7 +4690,7 @@ public class AdminController extends SpringActionController
                     for (org.labkey.api.writer.Writer child : childWriters)
                     {
                         dataType = child.getDataType();
-                        if (dataType != null)
+                        if (dataType != null && !dataType.endsWith("Dataset Data"))
                             registeredFolderWriters.add(dataType);
                     }
                 }
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -4690,7 +4690,7 @@ public class AdminController extends SpringActionController
                     for (org.labkey.api.writer.Writer child : childWriters)
                     {
                         dataType = child.getDataType();
-                        if (dataType != null && !dataType.endsWith("Dataset Data"))
+                        if (dataType != null)
                             registeredFolderWriters.add(dataType);
                     }
                 }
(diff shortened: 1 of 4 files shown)


RosalineP:
NOTE:
So unfortunately, if I instead want to write something like `(dataType != null && !dataType.equals(StudyArchiveDataTypes.ASSAY_DATASET_DATA) && [...])` as the conditional, then `StudyArchiveDataTypes`, which is within `platform/study`, has to be visible from `AdminController` which is in `platform/core`. I tried moving `StudyArchiveDataTypes` to the api, which worked fine, but touched a ton of files.
In this situation, would it be better to use this PR's approach, or the move-StudyArchiveDataTypes-to-API approach?

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -4690,7 +4690,7 @@ public URLHelper getSuccessURL(ImportFolderForm importFolderForm)
                     for (org.labkey.api.writer.Writer child : childWriters)
                     {
                         dataType = child.getDataType();
-                        if (dataType != null)
+                        if (dataType != null && !dataType.endsWith("Dataset Data"))
```
RosalineP: NOTE:
So unfortunately, if I instead want to write something like `(dataType != null && !dataType.equals(StudyArchiveDataTypes.ASSAY_DATASET_DATA) && [...])` as the conditional, then `StudyArchiveDataTypes`, which is within `platform/study`, has to be visible from `AdminController` which is in `platform/core`. I tried moving `StudyArchiveDataTypes` to the api, which worked fine, but touched a ton of files.
In this situation, would it be better to use this PR's approach, or the move-StudyArchiveDataTypes-to-API approach?

labkey-klum: Seems like this could work, although would it make more sense to override the `includeWithTemplate` method for:
- `AssayDatasetData`
- `StudyDatasetData`
- `SampleTypeDatasetData`

It appears that `FolderWriter.getChildren` already tries to pass a parameter indicating whether it is for a template, and `StudyWriterFactory` is already handling the filtering by asking all of the internal writers whether they should be included when used as a template.

RosalineP: Ah, that's a much nicer solution. Hadn't made the connection between that `!excludeForTemplate` and the dummy writers. Implemented 👍

labkey-klum (approved):

PR #2343, opened 2021-06-10
Item 9012: Update application navigation menu behavior

labkey-susanh:
#### Rationale
We are simplifying the product/application navigation menu to better suit the majority use case for this menu. In particular, we are removing the option to navigate to products in a different container from this menu, with the only exception being navigating to the Home folder.  When in the home folder for LKS, we will still list the current container under the LKS header in the menu, but that item will no longer be clickable.

We also add a new Look & Feel setting to allow the menu to be hidden from non-admin users.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/555
* https://github.com/LabKey/biologics/pull/909
* https://github.com/LabKey/sampleManagement/pull/594

#### Changes
* Add new Look & Feel setting for disabling application menu for non-admins and add metric for this setting
* Update use of ProductNavigationMenu so the link to the current container is disabled.

### DIFF
--- a/core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
+++ b/core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
@@ -49,7 +49,6 @@
     boolean themeNameInherited = !c.isRoot() && laf.isThemeNameInherited();
     boolean canUpdate = !c.isRoot() || c.hasPermission(getUser(), ApplicationAdminPermission.class);
     boolean hasPremiumModule = ModuleLoader.getInstance().hasModule("Premium");
-    boolean isSMHostedOnly = !hasPremiumModule && ModuleLoader.getInstance().hasModule("SampleManagement");
 %>
 <%=formatMissedErrors("form")%>
 <labkey:form name="preferences" method="post" id="form-preferences">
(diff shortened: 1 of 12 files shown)


core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
```diff
@@ -102,14 +104,35 @@
...
+    <% if (hasPremiumModule)
+    {
+    %>
+<tr>
+    <%
+        FolderDisplayMode currentMenuDisplayMode = laf.getApplicationMenuDisplayMode();
+    %>
+    <td class="labkey-form-label">
+        Show Application Selection Menu
+        <div id="app-menu-warning" class="labkey-error" style=<%=currentMenuDisplayMode == FolderDisplayMode.ADMIN ? q("display:block;"): q("display:none;")%>>Users will not be able to navigate between applications and LabKey Server when this menu is hidden.</div>
```
cnathe: having this warning/error message in the label area seems odd to me. thoughts on putting it after the radio text for "Only for Administrators" when that option is selected? or below the radio inputs?

cnathe (approved):

core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
labkey-susanh: I tried it in various places.  Below the radio inputs doesn't really work because it's not clear which option it applies to (the one above or the one below) without lots of extra text.  I'll try it after the "Only for Administrators" text.

PR #2346, opened 2021-06-11
Always include required modules in the container.getActiveModules() set

cnathe:
#### Rationale
We have recently been adding some more server level context to the LABKEY JS object using the "api" module for use in the client side via the LABKEY.moduleContext.api. Some examples include allowing the shared components code to be able to tell which modules are available on the given server (i.e. do we have the "premium" module? do we have the "ontology" module). We also use this for the new session invalid behavior property for compliance settings as a way to make sure the property value is available in the client side code for all LK containers.

However, it looks like there are some scenarios which create a LK project or container via an API call which can results in the "API" module not being considered part of the active modules set. This PR make sure that the required modules are always added to the container.getActiveModules() set (they are required after all).

#### Changes
* Add required modules to the container.getActiveModules() set

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -931,7 +931,7 @@ public class Container implements Serializable, Comparable<Container>, Securable
 
     public Set<Module> getRequiredModules()
     {
-        Set<Module> requiredModules = new HashSet<>(getFolderType().getActiveModules());
+        Set<Module> requiredModules = new HashSet<>(getRequiredModulesForFolderType(getFolderType()));
         requiredModules.add(ModuleLoader.getInstance().getModule("API"));
         requiredModules.add(ModuleLoader.getInstance().getModule("Internal"));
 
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/Container.java
```diff
@@ -1164,6 +1164,10 @@ public boolean getAsBoolean()
         }
 
         Set<Module> modules = new HashSet<>();
+
+        // always put the required modules in the set
+        modules.addAll(getRequiredModules());
```
labkey-jeckels, line 1187: getRequiredModules() blindly adds everything from getFolderType().getActiveModules(), whereas the code a little lower down in this method is much more defensive.

Seems like we should move the defensive code out of this method and into getRequiredModules()

labkey-jeckels (changes requested):

api/src/org/labkey/api/data/Container.java
cnathe: It looks like the cod within `getRequiredModules()` that is adding from the `child.getFolderType().getActiveModules()` is only for workbook child folders. I don't know a lot about workbooks, but does that change anything about this recommendation?

labkey-jeckels: It's the current folder too:

https://github.com/LabKey/platform/blob/2443dfb3713412c5924ddbd3c3122bce2616b3bb/api/src/org/labkey/api/data/Container.java#L934

Based on the comment where it's doing the filtering now, I would expect that we should do the same filtering on everything, whether it's a workbook child or not

cnathe: Got it. Wow, I can't believe I didn't see that line after your initial comment. I'll make the suggested refactor and resubmit for review. Thanks.

labkey-jeckels (approved):

PR #2347, opened 2021-06-11
Standard support for "with timezone" JDBC types

labkey-adam:
#### Rationale
Add `Types.TIME_WITH_TIMEZONE` as an alias for `JdbcType.Time` and `Types.TIMESTAMP_WITH_TIMEZONE` as an alias for `JdbcType.TIMESTAMP`.

[https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=42311](https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=42311)

#### Related Pull Requests
* https://github.com/LabKey/redshift/pull/22

### DIFF
--- a/api/src/org/labkey/api/data/JdbcType.java
+++ b/api/src/org/labkey/api/data/JdbcType.java
@@ -217,6 +217,13 @@ public enum JdbcType
             assert !(d instanceof Time);
             return new Time(d.getTime());
         }
+
+        @Override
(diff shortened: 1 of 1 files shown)


labkey-matthewb (approved):
I could imagine that we might actually want different behavior for TIMESTAMP v TIMESTAMP_WITH_TIMEZONE, but good for now...

PR #2348, opened 2021-06-11
43366: Users without AddUserPermission can still create new users from the manage group UI

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=43366

#### Changes
The UI allows adding existing as well as new users to the group, spin through the user list to validate permissions and either fail or succeed the entire (user operation).

### DIFF
--- a/core/src/org/labkey/core/security/SecurityController.java
+++ b/core/src/org/labkey/core/security/SecurityController.java
@@ -834,12 +834,15 @@ public class SecurityController extends SpringActionController
 
                 // issue 43366 : users without the AddUserPermission are still allowed to create new users through the
                 // manage group UI.
-                for (ValidEmail email : addEmails)
+                if (!container.hasPermission(getUser(), AddUserPermission.class))
                 {
-                    if (!UserManager.userExists(email) && !container.hasPermission(getUser(), AddUserPermission.class))
+                    for (ValidEmail email : addEmails)
                     {
-                        errors.reject(ERROR_MSG, "You do not have permissions to create new users.");
-                        break;
+                        if (!UserManager.userExists(email))
+                        {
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/security/SecurityController.java
```diff
@@ -832,7 +832,18 @@ public boolean handlePost(UpdateMembersForm form, BindException errors) throws E
                     }
                 }
 
-                if (addGroups.size() > 0 || addEmails.size() > 0)
+                // issue 43366 : users without the AddUserPermission are still allowed to create new users through the
+                // manage group UI.
+                for (ValidEmail email : addEmails)
+                {
+                    if (!UserManager.userExists(email) && !container.hasPermission(getUser(), AddUserPermission.class))
```
labkey-adam: Not critical, but doing the same permissions check on every non-existent user isn't really needed. Perhaps hoist the `hasPermissions()` call out of the loop, assign to a local, and use that below?

labkey-adam (approved):

core/src/org/labkey/core/security/SecurityController.java
labkey-klum: 👍

core/src/org/labkey/core/security/SecurityController.java
```diff
@@ -832,7 +832,21 @@ public boolean handlePost(UpdateMembersForm form, BindException errors) throws E
                     }
                 }
 
-                if (addGroups.size() > 0 || addEmails.size() > 0)
+                // issue 43366 : users without the AddUserPermission are still allowed to create new users through the
+                // manage group UI.
+                if (!container.hasPermission(getUser(), AddUserPermission.class))
```
labkey-adam, line 837: 👍 Even better than my suggestion

PR #2350, opened 2021-06-14
Issue 43368: Infinite redirect when hitting assay upload wizard

labkey-jeckels:
Happens when you have a batchId parameter and no batch properties in design

#### Changes
* Redirect to run page of the wizard as intended. Must be a long-standing copy/paste error


labkey-kevink (approved):

labkey-jeckels:
@labkey-kevink My first approach didn't work quite as I hoped. The original code wasn't as broken as it seemed

PR #2353, opened 2021-06-15

vagisha:
#### Rationale
FileImporter, that is used with the Panorama Public copy pipeline,  creates rows in exp.data for each file that is copied to the new folder. I found two problems with this:
1. If there is a '+' character in the file name, it is replaced with a space in the name that is saved to the 'Name' column of exp.data
2. If there is a '%' character in the file name,  an exception is thrown when trying to save a row for the file. Exp.data rows are not saved for any of the remaining files

This is described here: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43374


labkey-jeckels:
Replaced with this alternative approach: https://github.com/LabKey/platform/pull/2374

PR #2354, opened 2021-06-15
43373: Link to study : Assay's don't support fields with Visit concept URIs

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43373

#### Changes
- In the `PublishResultsQueryView` if either the visit or date columns have not resolved, search the columns in the current view that have the VISIT concept URI and add if appropriate, this is the current behavior for the participant ID.
- Those fields can originate from a sample type and don't need to be on the base assay results table, but do need to be pulled into the default data view.

### DIFF
--- a/study/src/org/labkey/study/assay/AssayPublishConfirmAction.java
+++ b/study/src/org/labkey/study/assay/AssayPublishConfirmAction.java
@@ -247,7 +247,7 @@ public class AssayPublishConfirmAction extends AbstractPublishConfirmAction<Assa
         if (sampleCols.size() == 1)
             additionalCols.put(ExtraColFieldKeys.SampleId, sampleCols.get(0).getFieldKey());
 
-        if ((additionalCols.get(ExtraColFieldKeys.Date) == null))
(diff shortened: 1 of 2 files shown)


study/src/org/labkey/study/assay/AssayPublishConfirmAction.java
```diff
@@ -247,7 +247,7 @@ protected FieldKey getObjectIdFieldKey(AssayPublishConfirmForm form)
         if (sampleCols.size() == 1)
             additionalCols.put(ExtraColFieldKeys.SampleId, sampleCols.get(0).getFieldKey());
 
-        if ((additionalCols.get(ExtraColFieldKeys.Date) == null))
+        if (!selectColumns.containsKey(additionalCols.get(ExtraColFieldKeys.Date)))
```
labkey-klum, line 250: Needed to check the selectColumns map instead because the `FieldKey` in the additionalCols map is always non-null

RosalineP (approved):
Looks good! Just pending TC validation.

PR #2355, opened 2021-06-15
43297: Import Samples via File Watcher, File Extension was kept as part of the Sample name

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43297

#### Changes
- When creating a new sample type from a file watcher file, just use the base name and exclude the file extension (if any)


RosalineP (approved):
👍

PR #2367, opened 2021-06-18
ontology field

labkey-matthewb:
#### Rationale
create a search field "ontology" that acts as an identifier


search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
```diff
@@ -611,6 +612,7 @@ else if (owner instanceof User)
             assert StringUtils.isNotEmpty((String)props.get(PROPERTY.categories.toString()));
 
             addTerms(doc, FIELD_NAME.searchCategories, Field.Store.YES, terms(PROPERTY.categories, props, null));
+            addTerms(doc, FIELD_NAME.ontology, Field.Store.YES, terms(PROPERTY.ontology, props, null));
```
labkey-adam: You need to store it?

labkey-adam (approved):

search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
labkey-matthewb: I don't think so??? The concept picker needs to be able to do "+ontology:NCIT aspirin".  Do we need to be able to retrieve the value?

labkey-adam: If you just need to query it then you don't need to store the value. That's only for things like the summary, URL, container/resource ID. Not sure why we're storing searchCategory.

PR #2369, opened 2021-06-18
Fix default domain migration code

labkey-adam:
#### Rationale
The "default domain" property moved from the "SiteConfig" property map to the "Authentication" property map. Unfortunately, the upgrade code that retrieved the old value was incorrect (`getGroupName()` and `getType()` were reversed). This fixes the upgrade.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2327


RosalineP (approved):
Works as intended!

PR #2374, opened 2021-06-21

labkey-jeckels:
#### Rationale
We support files in folder export/import, but are mangling the URLs we generate to connect them with exp.data rows during the import process

#### Changes
* URI encode the relative path part of the file name


filecontent/src/org/labkey/filecontent/FileQueryUpdateService.java
```diff
@@ -549,7 +551,7 @@ private WebdavResource getWebdavUrlFromAbsoluteFilePath(@NotNull String absolute
                 if (offset.startsWith("/"))
                     offset = offset.substring(1);
 
-                String davUrl = rootDavUrl + "/" + offset;
+                String davUrl = rootDavUrl + "/" + URLEncoder.encode(offset, StringUtilsLabKey.DEFAULT_CHARSET);
```
labkey-susanh: Any chance that offset includes "/"'s that are not meant to be encoded?

labkey-susanh (approved):

filecontent/src/org/labkey/filecontent/FileQueryUpdateService.java
labkey-jeckels: Yes! I hate URI encoding!

labkey-jeckels: Interestingly, the encoded slashes ended up working correctly. I have a vague, unsettled feeling that means some other code isn't doing the right thing with the URIs, but I'll save that for another bug.

Change pushed to retain the slashes

PR #2375, opened 2021-06-21
small fixes to query-import.api

labkey-kevink:
#### Changes
* `path` parameter resolves files under pipeline root
* return error message as top-level `exception` property for client api libraries

### DIFF
--- a/api/src/org/labkey/api/action/ExtFormResponseWriter.java
+++ b/api/src/org/labkey/api/action/ExtFormResponseWriter.java
@@ -145,10 +145,13 @@ public class ExtFormResponseWriter extends ApiJsonWriter
     @Override
     public void writeAndClose(Errors errors) throws IOException
     {
+        String message = null;
         JSONObject jsonErrors = new JSONObject();
-        for(ObjectError error : errors.getAllErrors())
+        for (ObjectError error : errors.getAllErrors())
         {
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/action/ExtFormResponseWriter.java
```diff
@@ -159,6 +162,8 @@ public void writeAndClose(Errors errors) throws IOException
 
         JSONObject root = new JSONObject();
         root.put("success", false); //used by Ext forms
+        // used by client api for FormApiActions (e.g. query-import.api and ImportDataCommand in java clientapi)
+        root.put("exception", message == null ? "(No error message)" : message);
```
labkey-matthewb, line 166: Do we document the expected/typical/best-practice format for JSON error response anywhere???

labkey-matthewb (approved):

api/src/org/labkey/api/action/ExtFormResponseWriter.java
labkey-kevink: We do!

https://www.labkey.org/Documentation/wiki-page.view?name=implementActions#error

labkey-matthewb: 🥳

PR #2377, opened 2021-06-22
Issue 43420: Swap jTidy --> jSoup

Sigmonia:
#### Rationale
jTidy appears abandoned and doesn't support common HTML5 tag elements

[Issue #43420: Replace jTidy with jSoup...or similar](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43420)

#### Related Pull Requests
* https://github.com/LabKey/server/pull/95

#### Changes
* Swapped jSoup library in place of jTidy
* TODOs
  * Additional code clean up
  * Expand unit tests
  * Use experimental flag/deprecation scheme?

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -81,7 +81,7 @@ List others = [
         "org.apache.xmlgraphics:fop:${fopVersion}",
         "flyingsaucer:core-renderer:${flyingsaucerVersion}",
         "com.google.guava:guava:${guavaVersion}",
-        "net.sf.jtidy:jtidy:${jtidyVersion}",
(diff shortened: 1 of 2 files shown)


labkey-susanh (commented):
Don't know much about jtidy or jsoup, but the switch seems like it will simplify things.  Are the default configurations more aligned with what we want or does it just not allow as much configuration.  There are obviously some tests that will need to be updated and perhaps others added (e.g., to cover the svg issue that initiated this?).

labkey-jeckels:
Updated draft PR to get fresh test results: https://github.com/LabKey/platform/pull/4542

PR #2378, opened 2021-06-22
testGetUsersWithPermissions()

labkey-matthewb:
#### Rationale
test getUsersWithPermissions() and getUsersWithOneOf()
moved TestCase to bottom of .java file


labkey-adam (approved):
GitHub gave me the worst diff ever here... but pulling it into IntelliJ made it sane again

PR #2379, opened 2021-06-22
QC State Filter Standardization

RosalineP:
#### Rationale
There are a handful of places where the QC State selection is not being observed such as 
- Create chart
- Create Report
- Summary stats dialog
- Column chart
due to the fact that QC States are not applied as traditional filters. This work refactors QC States to use standard url filters on grid views.

#### Related Pull Requests
* n/a

#### Changes
* Make QC Filter dropdown options apply standard filter to url href
* Make QC Filter dropdown sensitive to standard filters appleid on QC State column
* Remove defunct usage of legacy QC url param in `StudyUrlsImpl.getDatasetURL`
* Ensure navigation page upon successfully applying QC State to rows navigates to page with standard url filter

#### Issues Fixed
- [QC State : remove more of the legacy QC state code](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=43448)
- [QC State : ParticipantNavView doesn't respect the new QC state filters](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=43466)


study/src/org/labkey/study/query/DatasetQueryView.java
```diff
@@ -504,17 +513,100 @@ private ActionButton createFilterButton()
...
+                .collect(Collectors.toList());
+        String filterValue = new SimpleFilter.InClause(FieldKey.fromParts(""), qcLabels).toURLParam("").getValue();
+        return filterValue != null ? filterValue : "";
+    }
+
     private MenuButton createQCStateButton(QCStateSet currentSet)
     {
         List<QCStateSet> stateSets = QCStateSet.getSelectableSets(getContainer());
         MenuButton button = new MenuButton("QC State");
+        String dataRegionName = this.getSettings().getDataRegionName();
```
RosalineP, line 489: The section below is long, and I considered pulling parts into helper functions, but wasn't sure if that would just disrupt the logical flow of readability or if it would actually help someone in parsing the code.

study/src/org/labkey/study/query/DatasetQueryView.java
```diff
@@ -504,17 +513,100 @@ private ActionButton createFilterButton()
...
+                }
+            }
+
+            NavTree setItem = new NavTree(set.getLabel(), urlHelper);
             setItem.setId("QCState:" + set.getLabel());
-            if (set.equals(currentSet))
-                setItem.setSelected(true);
+
+            // When QC State Column gets filtered, detect the change and update QC State dropdown selection accordingly
+            boolean urlHasEqFilter = getViewContext().cloneActionURL().getParameter(eq) != null;
```
labkey-klum: You don't need to clone the `ActionURL` unless you are planning to mutate it, you can probably just use `getActionURL` if you are just going to read params. 

Also can we move this block of boolean outside of the loop? It looks like they wouldn't vary from one state to another.

study/src/org/labkey/study/view/overview.jsp
```diff
@@ -288,8 +291,8 @@
                 defaultReportURL.addParameter(DatasetDefinition.DATASETKEY, dataset.getDatasetId());
                 if (selectedCohort != null && bean.cohortFilter != null)
                     bean.cohortFilter.addURLParameters(study, defaultReportURL, "Dataset");
-                if (bean.qcStates != null)
-                    defaultReportURL.addParameter("QCState", bean.qcStates.getFormValue());
+                if (bean.qcStates != null && StringUtils.isNumeric(bean.qcStates.getFormValue()))
+                    defaultReportURL.replaceParameter(eq, QCStateManager.getInstance().getQCStateForRowId(container, Integer.parseInt(bean.qcStates.getFormValue())).getLabel());
```
labkey-klum, line 300: I think this will only work for matching a single QC state, but not for multiple states like private or public data. You probably need to generate the same URL you do using the in clause. Maybe move the code in `DatasetQueryView` into the `QCStateSet` class?

labkey-klum (changes requested):
we'll need to address the links on the overview page as well as remove the code in `DatasetQueryView` which silently adds the filter.

study/src/org/labkey/study/view/overview.jsp
RosalineP: Resolved!

study/src/org/labkey/study/query/DatasetQueryView.java
RosalineP: Fixed the `ActionURL` issue. 

After creating the helper functions in`QCStateSet`, there are some constants defined within loops that would be slightly better efficiency-wise if they were defined outside of loops, but I think for readability and not having duplicate code hanging around, the (very minor) efficiency decrease is a worthwhile trade-off.

RosalineP: For example, note the constants 'eq' and 'in' within `selectedQCStateLabelFromUrl()`

RosalineP:
Issues resolved as of new Review Request:

- Issue 43448: Karl removed legacy QC State Code
- Issue 43441: Dataset QC state status above grid doesn't update
- Issue 43438: Menu item filters getting combined incorrectly. When clicking to 'Public/approved data' from the QC State menu, to a private state option, no rows were displayed. Now, Clicking from 'Public/approved data' to a private state or 'Private/non-approved data' to a public state correctly shows to full set of the latter chosen state.
- Issue 43436: Default data visibility configuration doesn't play well with new QC state filters. Now, setting 'Default visibility' to 'Public Data' affects all links to datasets that pass through the `DefaultDatasetReportAction` will be filtered automatically to 'Public/approved data'. The returnUrl of the 'Manage Dataset QC States' page will also behave correctly.


- CR Comment: On Overview page, setting dropdown to 'Public/approved data' or 'Private/non-approved data' will generate correctly filtered links. Even if 'Default visibility' is set to to 'Public Data'

study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -3730,6 +3752,13 @@ public ActionURL getRedirectURL(Object o)
                     else
                         url.addParameter(DATASET_VIEW_NAME_PARAMETER_NAME, defaultView);
                 }
+
+                if (StudyManager.getInstance() != null && url.getParameter("cancelRedirect") != null && !url.getParameter("cancelRedirect").equals("1"))
```
labkey-klum: The expression doesn't look quite right, the `cancelRedirect` parameter is only on URLs from the overview page (and missing on other pages) so I don't think this would work from the datasets page or clinical and assay data. You should be able to safely assume that the `StudyManager` is present in the module, so maybe we can simplify the expression to something like:

`if (!"1".equals(url.getParameter("cancelRedirect")))`

I would also suggest a different param name since we aren't really cancelling the redirect, maybe something like `skipDataVisibility` or something that communicates why we are bypassing that bit of code.

labkey-klum (approved):
Just need to address the feedback in `StudyController:3756`

study/src/org/labkey/study/controllers/StudyController.java
RosalineP: Done!

PR #2380, opened 2021-06-22
Item 9065: Support linking samples to study in LKSM 

labkey-susanh:
#### Rationale
When the study module is present along with sample manager, we want to be able to link samples to study datasets from within the application.  We also want to link into LKSM when the source type's container has that module enabled.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/607
* https://github.com/LabKey/labkey-ui-components/pull/564
* https://github.com/LabKey/biologics/pull/919

#### Changes
* Update links to the source type of a sample data set to go to LKSM when the source container has that module enabled.
* Use newly parameterized SampleTypeDesigner

### DIFF
--- a/api/src/org/labkey/api/study/Dataset.java
+++ b/api/src/org/labkey/api/study/Dataset.java
@@ -35,6 +35,7 @@ import org.labkey.api.exp.api.ExperimentService;
 import org.labkey.api.exp.api.ExperimentUrls;
 import org.labkey.api.exp.api.SampleTypeService;
 import org.labkey.api.exp.property.Domain;
+import org.labkey.api.module.ModuleLoader;
 import org.labkey.api.reports.model.ViewCategory;
 import org.labkey.api.security.User;
 import org.labkey.api.security.UserPrincipal;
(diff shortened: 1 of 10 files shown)


RosalineP (approved):
Looks good!

api/src/org/labkey/api/study/Dataset.java
```diff
@@ -204,15 +202,35 @@ protected String getAuditMessageSourceType()
                     {
                         return "sample type";
                     }
+
+                    @Override
+                    public ActionURL getSourceActionURL(ExpObject sourceObject, Container container)
+                    {
+                        ActionURL url;
+                        if (container.getActiveModules().contains(ModuleLoader.getInstance().getModule("sampleManagement")))
```
cnathe, line 211: may be worth a comment as to why this if/else has been added

core/src/client/SampleTypeDesigner/SampleTypeDesigner.tsx
```diff
@@ -157,6 +157,7 @@ export class App extends React.PureComponent<any, State> {
                     includeDataClasses={true}
                     useTheme={true}
                     appPropertiesOnly={false}
+                    showLinkToStudy={true}
```
cnathe, line 160: same comment as LKSM case, I that `true` should be the default for this new prop, so that you can drop this from the common use cases.

api/src/org/labkey/api/study/Dataset.java
labkey-susanh: Good thought.  Done.

labkey-klum (approved):

PR #2381, opened 2021-06-23
Item 9807: domain properties API

labkey-kevink:
#### Rationale
Introduce a few APIs for working with vocabulary domain properties.

#### Related Pull Requests
* https://github.com/LabKey/labbook/pull/110
* https://github.com/LabKey/labkey-api-js/pull/103

#### Changes
* Annotate the magic "Properties" column with conceptURI of `http://www.labkey.org/types#properties` so the client-side grid can use special rendering.

* When querying the magic "Properties" column on run, dataclass, or samples, include a new top-level `additionalFieldMetadata` in the response with field metadata for the properties encountered.  Unfortunately, we can't know the properties before streaming and rendering the response so the additional field metadata will appear at the end.

* property-getProperties.api - Previously only supported getting properties via propertyId or propertyURI.  Added the ability to find properties by domain, domain kind, a search term (over name, label, description, importAliases), or arbitrary property descriptor fields.  In addition, pagination is possible with `maxRows` and `offset` parameters.

* property-updateDomain.api - Add, update, remove properties on a domain one at a time, unlike the property-saveDomain.api which requires all fields in a domain to be included.

* property-propertyUsages.api - Find objects using a property in the ontology store, including the recently added support for null values.  Limit the number of objects serialized in the response with the `maxUsageCount` parameter.


api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -2078,6 +2078,101 @@ public static PropertyDescriptor getPropertyDescriptor(String propertyURI, Conta
...
+            @Nullable Set<String> domainKinds,
+            @Nullable String searchTerm,
+            @Nullable SimpleFilter propertyFilter,
+            @Nullable String sortColumn,
+            @Nullable Integer maxRows, @Nullable Long offset)
+    {
+        final FieldKey propertyIdKey = FieldKey.fromParts("propertyId");
+
+        // To filter by domain kind, we query the exp.DomainProperty table and filter by domainId.
+        // To construct a PropertyDescriptor, we will need to traverse the lookup to exp.PropertyDescriptor and select all of it's columns.
```
labkey-jeckels:
```suggestion
        // To construct a PropertyDescriptor, we will need to traverse the lookup to exp.PropertyDescriptor and select all of its columns.

```

api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -2078,6 +2078,101 @@ public static PropertyDescriptor getPropertyDescriptor(String propertyURI, Conta
...
+        var colMap = QueryService.get().getColumns(getTinfoPropertyDomain(), fields);
+
+        var filter = new SimpleFilter();
+        if (propertyFilter != null)
+        {
+            filter.addAllClauses(propertyFilter);
+        }
+
+        filter.addCondition(new FieldKey(propertyIdKey, "container"), c.getId());
+        if (domainIds != null && !domainIds.isEmpty())
```
labkey-jeckels, line 2110: Should we throw an exception if you pass both domainIds and domainKinds as arguments? domainIds win and causes us to ignore domainKinds

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -1457,7 +1625,33 @@ else if (form.getPropertyURIs() != null && !form.getPropertyURIs().isEmpty())
...
-                throw new ApiUsageException("Expected propertyIds or propertyURIs");
+                SimpleFilter filter = null;
+                if (form.getFilters() != null && !form.getFilters().isEmpty())
+                {
+                    filter = new SimpleFilter();
+
+                    // Reparent the URL filter columns to be relative from the "propertyId" lookup used
+                    // in the query (e.g. from "query.column~op=value" to "query.propertyId/column~op=value"),
+                    // create filter clauses, and them to the filter collection.
+                    form.getFilters()
```
labkey-jeckels, line 1661: Can we use SimpleFilter.addUrlFilters() to do the parsing for us?

labkey-jeckels: Obviously we'd have to transform it still, but it would hopefully be less string manipulation

labkey-jeckels (approved):
Some suggestions but overall looks OK to me

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
labkey-kevink: That was my first attempt, but trying to fixup the FieldKey after the filter clause is constructed proved trickier than rewriting it before constructing the filter.  Ideally, we'd have a `CompareClause.remapFieldKeys` method similar to `ForeignKey.remapFieldKeys`

api/src/org/labkey/api/exp/OntologyManager.java
labkey-kevink: I went back and forth on this.  I changed this to allow both domainIds and domainKinds to be specified.  I also added validation of the other form parameters

PR #2383, opened 2021-06-23
Issue38536: Refactor Button.java

RosalineP:
#### Rationale
Disentangle use of `iconCls` and `tooltip` attributes on buttons.

Note that this change should not (and does not) affect any UI. As previously, some buttons that relate to Issues (such as on url `issues-list.view?issueDefName=<name of issue def>`) have a search icon instead of any text, and the text 'Search' in a hover-over tooltip.

#### Related Pull Requests
* n/a

#### Changes
* Remove conditional dependency on `iconOnly`
* Update buttons that use `iconCls` accordingly


api/src/org/labkey/api/util/Button.java
```diff
@@ -231,7 +231,7 @@ public HtmlString getHtmlString()
     {
         boolean iconOnly = getIconCls() != null;
         String submitId = GUID.makeGUID();
-        final HtmlString tip = (null != tooltip ? HtmlString.of(tooltip) : (!iconOnly ? null : html));
+        final HtmlString tip = HtmlString.of(tooltip);
```
labkey-adam, line 234: Previously, `tip` would have ended up null if `tooltip` were null (and `!iconOnly`). Now it'll end up an empty `HtmlString`. Is that correct? Will we have empty tooltips now?

RosalineP: Nope! An empty string collapses into a non-existent attribute. 

With `tooltip`, relevant sections underlined in red:
WithToolTip

Without `tooltip`:
WithoutToolTip

This behavior is the same whether there is a `iconCls` or not on the button builder.

labkey-adam: Okay!

labkey-adam (approved):

PR #2386, opened 2021-06-23
fix reported npe

labkey-matthewb:
#### Rationale
fix reported npe

### DIFF
--- a/api/src/org/labkey/api/data/BaseColumnInfo.java
+++ b/api/src/org/labkey/api/data/BaseColumnInfo.java
@@ -67,6 +67,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


XingY (approved):

api/src/org/labkey/api/data/BaseColumnInfo.java
```diff
@@ -1272,12 +1273,10 @@ else if (xmlCol.isSetKeyVariable())
     @Override
     public void setSortFieldKeysFromXml(String xml)
     {
-        List<FieldKey> keys = new ArrayList<>();
-        for (String key : xml.split(","))
-        {
-            keys.add(FieldKey.fromString(key));
-        }
-
+        List<FieldKey> keys = Arrays.stream(StringUtils.split(xml, ','))
```
Sigmonia, line 1276: Possible NPE if xml is null

PR #2389, opened 2021-06-23
Issue 43261:  Convert workflow priority and status lists to enum tables

XingY:
#### Rationale
Sample manager workflow uses container scoped lists to reference certain job and task properties. We don't have a way to mark those special list as 'system' lists and they might get edited or deleted by users with edit permissions. Switching the implementation to enum tables will help prevent such errors.  

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2389
* https://github.com/LabKey/sampleManagement/pull/609
* https://github.com/LabKey/biologics/pull/922

#### Changes
* add exp.WorkflowTaskStatus and exp.WorkflowJobPriority enum tables


XingY:
@labkey-susanh I don't think those tables need to live in exp. They probably suit SampleManagementSchema better. What do you think?

api/src/org/labkey/api/exp/query/WorkflowTaskStatusEnumTableInfo.java
```diff
@@ -0,0 +1,116 @@
...
+import org.labkey.api.query.ExprColumn;
+import org.labkey.api.query.UserSchema;
+
+import java.util.Arrays;
+import java.util.EnumSet;
+
+public class WorkflowTaskStatusEnumTableInfo extends EnumTableInfo
+{
+    public static final String WORKFLOW_TASK_STATUS_TABLE_NAME = "WorkflowTaskStatus";
+    public static final String WORKFLOW_JOB_PRIORITY_TABLE_NAME = "WorkflowJobPriority";
```
labkey-susanh: It's strange to me that the priority table and enum live in this class.  I think I would move them into their own class or just in the Schema class itself, and yes, I think these may be better suited to the SampleManager schema.

labkey-susanh (changes requested):

XingY:
Moved code to sampleManagement, closing PR

PR #2390, opened 2021-06-24
Issue38536: Fix button tooltip related test outages

RosalineP:
#### Rationale
The previous merge overlooked the existence of buttons setting their icon class using `setIconCls`. For all instances where this is the case, the tooltip must be set explicitly.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2383

#### Changes
* Add necessary calls to `setTooltip`


labkey-adam:
Looks like a similar failure on `DomainDesignerTest.showHideFieldOnInsertGridView`

core/src/org/labkey/core/workbook/WorkbookQueryView.java
```diff
@@ -65,6 +66,7 @@ public ActionButton createInsertButton()
         final ActionButton insertButton = super.createInsertButton();
         insertButton.setCaption("Create Workbook");
```
labkey-adam: Delete this line? Does setting caption on an icon button do anything now?

PR #2391, opened 2021-06-24
Miscellaneous cleanups

labkey-adam:
#### Rationale
Simplify and streamline some code

### DIFF
--- a/api/src/org/labkey/api/security/AuthenticationManager.java
+++ b/api/src/org/labkey/api/security/AuthenticationManager.java
@@ -94,6 +94,7 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/security/AuthenticationManager.java
```diff
@@ -256,7 +256,7 @@ public static String getDefaultDomain()
 
     public static void setDefaultDomain(User user, String value)
     {
-        saveAuthSetting(user, DEFAULT_DOMAIN, value);
+        saveAuthSetting(user, DEFAULT_DOMAIN, StringUtils.trimToEmpty(value));
```
labkey-matthewb: move trimToEmpty() into saveAuthSetting()?

labkey-matthewb (approved):

RosalineP (approved):

api/src/org/labkey/api/security/AuthenticationManager.java
labkey-adam: Better yet... I've in-lined saveAuthSettings(), since this was the only caller

PR #2393, opened 2021-06-24
Fix WebDav copy action NPE

XingY:
#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/core/src/org/labkey/core/webdav/DavController.java
+++ b/core/src/org/labkey/core/webdav/DavController.java
@@ -5991,7 +5991,7 @@ public class DavController extends SpringActionController
      * during the copy operation
      * @param destPath Destination path
      */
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/webdav/DavController.java
```diff
@@ -5999,7 +5999,7 @@ private WebdavStatus copyResource(WebdavResource src, Map<Path,WebdavStatus> err
 
         if (src.isCollection())
         {
-            if (!dest.getFile().mkdir())
+            if (dest.getFile() == null || !dest.getFile().mkdir())
```
labkey-matthewb: This whole method too "file system" specific.  That should be improved eventually.  For this line of code maybe a better change is to call " if (!dest.createCollection(getUser())"?  See class MkcolAction.

labkey-matthewb (changes requested):
see comment

labkey-matthewb (approved):

PR #2395, opened 2021-06-24
Update .gitignore location for generated files

labkey-tchad:
#### Rationale
Moved due to '@labkey/build' v4.0.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,12 +17,8 @@ core/resources/web/clientapi
 core/resources/web/core/css
 */resources/web/gen
 */resources/views/gen
-issues/resources/web/issues/gen
-list/resources/web/list/gen
-query/resources/web/query/gen
-study/resources/web/study/gen
 
-*/resources/credits/jars.txt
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -17,27 +15,14 @@ core/resources/web/clientapi
 
 # Auto-generated files
 core/resources/web/core/css
-core/resources/web/core/gen
-core/resources/views/gen
+*/resources/web/gen
+*/resources/views/gen
+issues/resources/web/issues/gen
```
labkey-nicka: It's fine to ignore these but really they shouldn't be there any longer. My guess is they just weren't cleaned up for the "final" time before switching to the new gen directory paradigm so they're still there.

labkey-nicka (approved):

.gitignore
labkey-tchad: Ahh. I did a clean build to make sure I didn't stop ignoring anything I shouldn't but those files must have been left over from a previous build.

PR #2397, opened 2021-06-25
Issue 43349: SampleTypeUpdateServiceDI.getExistingRows() performance

XingY:
#### Rationale
Sample import with merge is slow due to lineage queries (one for parent material and one for parent data) that's done for each row of data in SampleTypeUpdateServiceDI.getExistingRows(). Those queries are needed for sample detailed audit log, which is used by sample timeline. This PR merges data and material parent into one query, and additionally handle 50 rows in a single query, which should reduce the number of lineage queries by ~99%. 

#### Related Pull Requests
* 

#### Changes
* new ExperimentServiceImpl.getParentMaterialAndDataMap function that gets both parent data and parent material for a set of sample lsid in a single query
* SampleTypeUpdateServiceDI.getExistingRows to use getParentMaterialAndDataMap
* Update ExistingDataIteratorsGetRows to get 50 rows a time

### DIFF
--- a/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
+++ b/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
@@ -70,8 +70,8 @@ import org.springframework.util.StringUtils;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -569,7 +569,6 @@ public class SampleTypeUpdateServiceDI extends DefaultQueryUpdateService
             }
         }
 
-
         if (!lsidRowNumMap.isEmpty())
         {
             Filter filter = new SimpleFilter(FieldKey.fromParts(ExpMaterialTable.Column.LSID), lsidRowNumMap.keySet(), CompareType.IN);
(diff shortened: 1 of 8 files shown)


experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -516,10 +521,68 @@ else if (parent instanceof ExpMaterial)
...
+
+    private Map<Integer, Map<String, Object>> getMaterialMapsWithInput(Map<Integer, Map<String, Object>> keys, User user, Container container)
             throws QueryUpdateServiceException
     {
-        return getRows(user, container, keys, true /* this is for detailed audit logging */);
+        Map<Integer, Map<String, Object>> sampleRows = new LinkedHashMap<>();
+        Set<Identifiable> lsids = new LinkedHashSet<>(keys.size());
+        Map<Integer, String> rowLsid = new HashMap<>();
+
+        for (Map.Entry<Integer, Map<String, Object>> keyMap : keys.entrySet())
```
labkey-matthewb, line 540: The remaining work is to unroll this loop and use using and IN() filter instead of =
(can be separate PR)

labkey-kevink (approved):
The single lineage query for both parents and children for a group of samples looks good.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -516,10 +521,68 @@ else if (parent instanceof ExpMaterial)
...
+                filter = new SimpleFilter(FieldKey.fromParts(ExpMaterialTable.Column.LSID), lsid);
+            else
+                throw new QueryUpdateServiceException("Either RowId or LSID is required to get Sample Type Material.");
+
+            Map<String, Object> sampleRow = new TableSelector(getQueryTable(), filter, null).getMap();
+            if (null != sampleRow)
+            {
+                String sampleLsid = lsid != null ? lsid : (String) sampleRow.get("lsid");
+
+                Identifiable id = LsidManager.get().getObject(sampleLsid);
```
labkey-matthewb: We really shouldn't call LsidManager.get().getObject(sampleLsid);  This causes two more queries per row.  One for findType() and one for getObject().  We already know the type and we already have the row in memory.

labkey-matthewb (changes requested):
I think it would be worth looking at the LsidManager.get().getObject(sampleLsid) issue and deciding whether to change in this PR or fix later.

labkey-matthewb (approved):

PR #2399, opened 2021-06-25
Fb issue 43456

labkey-klum:
#### Rationale
We create a single StudyPublishProtocol for the provenance run when either assay or sample data is linked to a study. While we ensure that a single protocol exists for the site, the protocol is created in the container that it is first requested from. If we try to export and reimport that folder into a new folder the import will fail because the new protocol doesn't match the existing protocol's container.

- https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43456

#### Changes
- In the future create the protocol in the shared folder, the XarReader will allow usage if the existing protocol container is in the shared folder.
- Upgrade existing StudyPublishProtocols and move them to the shared folder.

### DIFF
--- a/experiment/resources/schemas/dbscripts/postgresql/exp-21.007-21.008.sql
+++ b/experiment/resources/schemas/dbscripts/postgresql/exp-21.007-21.008.sql
@@ -1,0 +1,4 @@
+-- move the StudyPublishProtocol container to the shared folder
+UPDATE exp.protocol
+    SET container = (SELECT entityid FROM core.containers WHERE name = 'Shared')
+    WHERE lsid LIKE 'urn:lsid:labkey.org:Protocol:StudyPublishProtocol%';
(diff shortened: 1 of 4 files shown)


experiment/resources/schemas/dbscripts/postgresql/exp-21.007-21.008.sql
```diff
@@ -0,0 +1,4 @@
+-- move the StudyPublishProtocol container to the shared folder
+UPDATE exp.protocol
+    SET container = (SELECT entityid FROM core.containers WHERE name = 'Shared')
```
labkey-kevink, line 3: This gave me pause, but I think referring to 'Shared' ok for both upgrade and bootstrap given that the protocol won't exist yet.

labkey-kevink (approved):

experiment/resources/schemas/dbscripts/postgresql/exp-21.007-21.008.sql
labkey-klum: Thanks, I'll test the bootstrap case just to be sure

PR #2401, opened 2021-06-25
CreatedModified Abstract Entity

labkey-nicka:
#### Rationale
Introduces an abstract `CreatedModified` class that can be extended by Jackson serialized models to support our standard CBMB fields.

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/924
* https://github.com/LabKey/sampleManagement/pull/611
* https://github.com/LabKey/labkey-ui-components/pull/567
* https://github.com/LabKey/platform/pull/2401

#### Changes
* Introduces abstract `CreatedModified` class for use by LK entity models.

### DIFF
--- a/api/src/org/labkey/api/data/CreatedModified.java
+++ b/api/src/org/labkey/api/data/CreatedModified.java
@@ -1,0 +1,78 @@
+package org.labkey.api.data;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.json.JSONObject;
+import org.labkey.api.security.User;
+
+import java.util.Date;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/CreatedModified.java
```diff
@@ -0,0 +1,78 @@
...
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public abstract class CreatedModified
+{
+    private Long _created;
+    private User _createdBy;
+    private Long _modified;
+    private User _modifiedBy;
+
+    @JsonProperty("created")
```
labkey-susanh, line 19: Is this explicit property annotation necessary? Seems like this would be the default.

labkey-susanh (approved):

api/src/org/labkey/api/data/CreatedModified.java
labkey-nicka: IIRC this is necessary because there are two different setters for different types. Annotating the getter of the type that it should choose allows it to work.

PR #2407, opened 2021-06-28
Issue 43440: Sample Type Dataset shows superfluous Linked To Study column

labkey-jeckels:
#### Rationale
Once a sample type is linked to a study dataset, we don't want to show the columns with link info on the dataset side

#### Changes
* Expose a way to get a sample type table without the linked columns added (or XML metadata from the sample type side of things applied)


RosalineP (approved):
Looks good, thanks Josh!

PR #2408, opened 2021-06-28
Issue 43374: + and % in container paths not properly encoded when creating exp.data rows in FileImporter

labkey-jeckels:
#### Rationale
A previous fix addressed special characters in the files that are part of the folder import, but there's a related problem when the target folder has special characters in its own name

#### Changes
* Hacky encoding for import pathway to hold us over until we can uniformly encode the webdav URL

### DIFF
--- a/filecontent/src/org/labkey/filecontent/FileContentServiceImpl.java
+++ b/filecontent/src/org/labkey/filecontent/FileContentServiceImpl.java
@@ -1372,6 +1372,12 @@ public class FileContentServiceImpl implements FileContentService
                 filesRoot = rootPathVal;
 
             String rootDavUrl = (String) child.get("webdavURL");
+
+            // Hack for issue 43374 - encode special characters in container paths. Need to push this encoding
+            // into FilesWebPart._getRootPath(), but other codepaths are doing their own compensation so it's a more
+            // involved change
+            rootDavUrl = rootDavUrl.replace("%", "%25").replace("+", "%2B");
+
             WebdavResource resource = getResource(rootDavUrl);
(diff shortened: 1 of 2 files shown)


filecontent/src/org/labkey/filecontent/FileContentServiceImpl.java
```diff
@@ -1372,6 +1372,12 @@ public void ensureFileData(@NotNull ExpDataTable table)
                 filesRoot = rootPathVal;
 
             String rootDavUrl = (String) child.get("webdavURL");
+
+            // Hack for issue 43374 - encode special characters in container paths. Need to push this encoding
+            // into FilesWebPart._getRootPath(), but other codepaths are doing their own compensation so it's a more
+            // involved change
+            rootDavUrl = rootDavUrl.replace("%", "%25").replace("+", "%2B");
```
labkey-susanh, line 1379: Seems OK as long as we know that `child.get("webdavURL")` is not yet been through any encoding (such that a `%` is present from an encoding and thus might get encoded twice).

labkey-susanh (approved):

filecontent/src/org/labkey/filecontent/FileContentServiceImpl.java
labkey-jeckels: @labkey-susanh yes, it's always unencoded. See FilesWebPart._getRootPath() where it calls c.getPath() instead of c.getEncodedPath(). I think ultimately that call should change and do the encoding, but other places in the client code are doing their own encoding right now.

PR #2412, opened 2021-06-30
Automation test coverage for issue 42937 

labkey-sweta:
#### Rationale
Automation test coverage for issue 42937 

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/782

#### Changes
Added a test case "testLinkedColumnNotDisplayedCase" in study/test/src/org/labkey/test/tests/study/AutoLinkToStudyTest.java

### DIFF
--- a/study/test/src/org/labkey/test/tests/study/AutoLinkToStudyTest.java
+++ b/study/test/src/org/labkey/test/tests/study/AutoLinkToStudyTest.java
@@ -10,6 +10,7 @@ import org.labkey.test.TestFileUtils;
 import org.labkey.test.TestTimeoutException;
 import org.labkey.test.categories.Assays;
 import org.labkey.test.categories.DailyC;
+import org.labkey.test.components.CustomizeView;
 import org.labkey.test.util.DataRegionTable;
 import org.labkey.test.util.StudyHelper;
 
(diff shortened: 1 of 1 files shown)


labkey-chrisj (approved):

study/test/src/org/labkey/test/tests/study/AutoLinkToStudyTest.java
```diff
@@ -54,14 +62,81 @@ public void testAutoLinkInSameFolder()
...
+
+    /*
+        Test coverage for : Issue 42937: Assay results grid loading performance can degrade with a large number of "copied to study" columns
+     */
+    @Test
+    public void testLinkedColumnNotDisplayedCase()
+    {
+        String runName = "Link to study run";
+        log("Creating the study projects");
+        _containerHelper.createProject("Study 1", "Study");
```
cnathe: In case someone runs this test locally without cleanup, I might suggest a slightly more unique project name for these 3 study projects (maybe something like getProjectName() + "Study 1", etc.). Or you could probably create them as 3 study subfolders of the current test project. That way they would get deleted with the project delete as well.

study/test/src/org/labkey/test/tests/study/AutoLinkToStudyTest.java
```diff
@@ -54,14 +62,81 @@ public void testAutoLinkInSameFolder()
...
-                .clickSave();
+        clickAndWait(Locator.linkWithText(ASSAY_NAME));
+        clickAndWait(Locator.linkWithText(runName));
+        DataRegionTable runsTable = DataRegionTable.DataRegion(getDriver()).withName("Data").waitFor();
+        checker().verifyFalse("Linked column for Study 1 should not be present",
+                runsTable.getColumnNames().contains("linked_to_Study_1_Study"));
+        checker().verifyFalse("Linked column for Study 2 should not be present",
+                runsTable.getColumnNames().contains("linked_to_Study_2_Study"));
+        checker().verifyFalse("Linked column for Study 3 should not be present",
+                runsTable.getColumnNames().contains("linked_to_Study_3_Study"));
```
cnathe: These checks are good and verify the issue. I might also consider adding a check after these lines to make sure that the 3 or 4 linked study columns can be added to the grid via customize view (i.e. verifies that they aren't in the default view but can still be added).

cnathe (commented):
@labkey-sweta test case looks good. a couple of suggestions here.

study/test/src/org/labkey/test/tests/study/AutoLinkToStudyTest.java
labkey-sweta: Got it, i am deleting these projects in cleanup. But will change the name.

labkey-sweta: done

study/test/src/org/labkey/test/tests/study/AutoLinkToStudyTest.java
labkey-sweta: Done

cnathe (approved):

PR #2413, opened 2021-06-30
Switch to java.sql.Date due to Jackson serialization issues with LocalDate

labkey-adam:
#### Rationale
Our version of Jackson can't serialize `LocalDate`

#### Related Pull Requests
* https://github.com/LabKey/compliance/pull/121

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -1661,13 +1661,26 @@ public class Container implements Serializable, Comparable<Container>, Securable
         _lockState = lockState;
     }
 
-    public @Nullable LocalDate getExpirationDate()
+    @Deprecated
+    public @Nullable LocalDate getExpirationDateLD()
     {
         return _expirationDate;
     }
 
-    public void setExpirationDate(LocalDate expirationDate)
+    @Deprecated
+    public void setExpirationDateLD(LocalDate expirationDate)
     {
         _expirationDate = expirationDate;
+    }
+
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/data/Container.java
```diff
@@ -1661,13 +1661,26 @@ public void setLockState(LockState lockState)
         _lockState = lockState;
     }
 
-    public @Nullable LocalDate getExpirationDate()
+    @Deprecated
```
labkey-nicka: Alternatively, we could tell Jackson which getter/setter to use for `expirationDate`. Something like this:

```java
@Deprecated
@JsonIgnore
public @Nullable LocalDate getExpirationDateLD()
{
    return _expirationDate;
}

// TODO: Convert to LocalDate once we fix Jackson serialization of LocalDate
@JsonProperty("expirationDate")
public java.sql.Date getExpirationDate()
{
    return java.sql.Date.valueOf(_expirationDate);
}

@JsonIgnore // expirationDate is serialized as java.sql.Date
public void setExpirationDate(LocalDate expirationDate)
{
}

public void setExpirationDate(java.sql.Date expirationDate)
{
    _expirationDate = expirationDate.toLocalDate();
}
```

labkey-adam: I had marked the member var a while back to workaround one Jackson problem, but obviously that wasn't enough. At this point (one day shy of branching and about to get on a plane) I'm leaning toward keeping it simple and not messing with annotations.

labkey-nicka (approved):

PR #2302, opened 2021-05-27
Issue32017: Capitalize 'default' in default view

RosalineP:
#### Rationale
To align with LK's general aesthetics, update the 'default' in default view of grids to be 'Default,' and update tests.

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/338
* https://github.com/LabKey/testAutomation/pull/741

#### Changes
* 'default' => 'Default' on grid views
* update tests to align with capitalization change

PR #2417, opened 2021-07-01
Issue32017: Make 'Default' reserved view name

RosalineP:
#### Rationale
Previously, the default grid title was changed from "default" to "Default". But "Default" is not a reserved view name like "default" is. This allows you to have two views named "Default". 
This change updates the reserved view name.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2302

#### Changes
* Change reserved view name casing
* Changing the casing caused the error message to display with cut-off text. Update text to fit within popover

### DIFF
--- a/internal/webapp/internal/ViewDesigner/Designer.js
+++ b/internal/webapp/internal/ViewDesigner/Designer.js
@@ -134,7 +134,7 @@ Ext4.define('LABKEY.internal.ViewDesigner.Designer', {
                             else if (trimmedValue.length > 50)
                                 return "The grid view name must be less than 50 characters long";
                             else if (LABKEY.internal.ViewDesigner.Designer.RESERVED_VIEW_NAMES.indexOf(trimmedValue) > -1) {
-                                return "The grid view name '" + value + "' is not allowed";
+                                return "The name '" + value + "' is not allowed";
                             }
                             return true;
                         },
--- a/internal/webapp/internal/ViewDesigner/Designer.js
+++ b/internal/webapp/internal/ViewDesigner/Designer.js
@@ -130,11 +130,11 @@ Ext4.define('LABKEY.internal.ViewDesigner.Designer', {
                         validator: function(value) {
                             var trimmedValue = value.trim();
                             if (trimmedValue.length == 0)
-                                return "Blank grid view name is not allowed";
+                                return "Blank grid view name is not allowed.";
                             else if (trimmedValue.length > 50)
-                                return "The grid view name must be less than 50 characters long";
+                                return "The grid view name must be less than 50 characters long.";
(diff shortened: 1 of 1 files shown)


RosalineP:
@cnathe If you see a way to adjust the size of the popover rather than adjusting the text, please let me know! I was not able to find it among the ext4, though.

cnathe:
> @cnathe If you see a way to adjust the size of the popover rather than adjusting the text, please let me know! I was not able to find it among the ext4, though.

Yeah, Ext4 is a pain with these types of things. It looks like the alert message box for some reason hard codes in a height so that it only shows the first line of text. The "fix" would be to move away from an Ext4.Msg.alert() to a Ext4.Msg.show, see https://docs.sencha.com/extjs/4.2.1/#!/api/Ext.MessageBox-method-show

```
                            Ext4.Msg.show({
                                title:"Invalid grid view name",
                                msg: nameField.getErrors(),
                                buttons: Ext4.Msg.OK,
                                icon: Ext4.Msg.ERROR
                            });
```

internal/webapp/internal/ViewDesigner/Designer.js
```diff
@@ -134,7 +134,7 @@ Ext4.define('LABKEY.internal.ViewDesigner.Designer', {
                             else if (trimmedValue.length > 50)
                                 return "The grid view name must be less than 50 characters long";
                             else if (LABKEY.internal.ViewDesigner.Designer.RESERVED_VIEW_NAMES.indexOf(trimmedValue) > -1) {
-                                return "The grid view name '" + value + "' is not allowed";
+                                return "The name '" + value + "' is not allowed";
```
cnathe: minor: can we get periods added to the end of these error message sentences?

cnathe (approved):

labkey-tchad (commented):
Should probably merge this to release21.7-SNAPSHOT

internal/webapp/internal/ViewDesigner/Designer.js
RosalineP: Done!

internal/webapp/internal/ViewDesigner/Designer.js
```diff
@@ -205,7 +205,12 @@ Ext4.define('LABKEY.internal.ViewDesigner.Designer', {
                         var nameField = win.down('#nameFieldContainer').items.get(1);
                         if (!nameField.isValid())
                         {
-                            Ext4.Msg.alert("Invalid grid view name", nameField.getErrors());
+                            Ext4.Msg.show({
```
cnathe, line 208: Looks like there is another usage of `Ext4.Msg.alert` below which might be worth converting to this.

PR #2424, opened 2021-07-02
Fb dataset notifications 9061

labkey-klum:
#### Rationale
We want to expand the options available when configuring notifications for report and dataset changes in a study. With this change we now support selection on a per dataset basis which will be mutually exclusive from the other previous options:
- All
- By category

In addition the notification sent out will contain a link (for datasets) to the audit changes that occurred for that notification. 

#### Related Documents
- [Specification](https://docs.google.com/document/d/1_7mF64TIT5b5x28RSw3S1G8xrmj_avaJyrWGDi481Dk/edit)

#### Changes
- Add a new option to the notification config UI to set on a per dataset basis
- Refactor the Ext4 code to be more consistent with other components that we have written in Ext4. Extend the panel class instead of acting like a factory class.
- Update the email template and formatting (styles were not included properly)
- Simplify the HTML generation of the report and notification email template grid. Switch from HTML generation in java to rendering using a JSP.
- Add an audit link to the manage datasets page
- Collect metrics for the number of notification configurations on the server


study/src/org/labkey/study/controllers/DatasetController.java
```diff
@@ -102,7 +103,7 @@ protected void addExtraNavTrail(NavTree root)
         }
     }
 
-    @RequiresPermission(AdminPermission.class)
+    @RequiresPermission(CanSeeAuditLogPermission.class)
```
labkey-klum, line 107: Checked with Susan, no reason that this view should be restricted to admin only.

query/src/org/labkey/query/reports/view/ReportAndDatasetChangeDigestEmailTemplate.java
```diff
@@ -111,45 +112,36 @@ public void init(Container c, Map<ViewCategory, List<NotificationInfo>> reports)
...
+
+    private String buildReportAndDatasetList(Container c, Map<ViewCategory, List<NotificationInfo>> reports)
+    {
+        try
+        {
+            NotificationBean bean = new NotificationBean();
+            bean.setReports(reports);
 
-        return sb.toString();
+            ViewContext context = ViewContext.getMockViewContext(User.getSearchUser(), c, new ActionURL(), true);
```
labkey-adam: Can you use a `JspTemplate` here instead?

query/src/org/labkey/query/reports/view/manageNotifications.jsp
```diff
@@ -33,62 +32,33 @@
...
-        var manager = Ext4.create('LABKEY.ext4.ManageReportNotifications');
-        var panel = manager.getManageReportPanel({
-                    renderTo : 'manageNotificationsDiv',
-                    minWidth : 750
-                },
-                categories, returnUrl, notifyOption);
+        Ext4.create('LABKEY.ext4.ReportNotificationPanel', {
+            title : 'Choose Notification Option',
+            categories : <%=HtmlString.unsafe(jsonMapper.writeValueAsString(form.getCategories()))%>,
+            datasets : <%=HtmlString.unsafe(jsonMapper.writeValueAsString(form.getDatasets()))%>,
```
labkey-adam: I think you could use `<%=toJsonArray()%>` or `<%=toJsonObject()%>` here. More succinct and their output is already marked `SafeToRender`.

labkey-adam (changes requested):
Just a couple suggestions to consider

labkey-adam:
Did you want to target `release21.7-SNAPSHOT`?

query/src/org/labkey/query/reports/view/manageNotifications.jsp
labkey-klum: 👍

query/src/org/labkey/query/reports/view/ReportAndDatasetChangeDigestEmailTemplate.java
labkey-klum: ah right, for the life of me I couldn't remember the name of that class.

query/src/org/labkey/query/reports/view/ReportAndDatasetChangeDigestEmailTemplate.java
```diff
@@ -111,45 +112,36 @@ public void init(Container c, Map<ViewCategory, List<NotificationInfo>> reports)
...
+            JspTemplate<NotificationBean> view = new JspTemplate<>("/org/labkey/query/reports/view/reportAndDatasetList.jsp", bean);
```
labkey-adam:
```suggestion
            return new JspTemplate<>("/org/labkey/query/reports/view/reportAndDatasetList.jsp", bean).render();

```

labkey-adam: You don't need the `MockViewContext`, etc. -- `JspTemplate.render()` will handle it for you

labkey-adam (changes requested):
Simplify `JspTemplate` handling

labkey-adam (approved):

PR #2427, opened 2021-07-05
Issue43511: Update System Default Domain tooltip href

RosalineP:
#### Rationale
Previously, the System Default Domain tooltip pointed to the `configAdmin` page. Now, it correctly points to `authenticationModule#dom`.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2327

#### Changes
* Update href

### DIFF
--- a/core/src/client/components/GlobalSettings.tsx
+++ b/core/src/client/components/GlobalSettings.tsx
@@ -39,7 +39,7 @@ class DefaultDomainField extends PureComponent<DefaultDomainProps> {
                 <LabelHelpTip title="Tip">
                     <div>
                         <div> Default domain for user sign in.</div>
-                        <a href="https://www.labkey.org/Documentation/wiki-page.view?name=configAdmin">More info</a>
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
I'm guessing we don't have a TypeScript helper function to generate help links? The equivalent of org.labkey.api.util.HelpTopic on the Java side? It's nothing fancy, but it helps us avoid hardcoding the URL prefix in lots of different places

labkey-jeckels (approved):
Looking at HelpTopic closer, I think we should add these to the <a>:

target="labkeyHelp" rel="noopener noreferrer"

labkey-susanh:
> I'm guessing we don't have a TypeScript helper function to generate help links? The equivalent of org.labkey.api.util.HelpTopic on the Java side? It's nothing fancy, but it helps us avoid hardcoding the URL prefix in lots of different places

There is the method `helpLinkNode` in ui-components if that helps.

PR #2429, opened 2021-07-06
Issue 43516: "Resolve lookup by value" experimental feature flag does not work for the "Derive Samples" insert form

cnathe:
#### Rationale
Issue [43516](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43516): "Resolve lookup by value" experimental feature flag does not work for the "Derive Samples" insert form

This PR removes the experimental feature for "Resolve lookup by value" to make it the default behavior for the existing locations where it is implemented. It also adds support for this lookup value resolution to the "Derive Samples" form code path.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/788

#### Changes
* Remove "resolve lookups by value" experimental feature 
* Use cache.remap() to resolve lookup value in Derive Samples form code path

### DIFF
--- a/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
+++ b/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
@@ -1013,9 +1013,10 @@ public class DefaultAssayRunCreator<ProviderType extends AbstractAssayProvider> 
     public static List<ValidationError> validateColumnProperties(ContainerUser context, Map<ColumnInfo, String> properties)
     {
         List<ValidationError> errors = new ArrayList<>();
+        RemapCache cache = new RemapCache();
         for (Map.Entry<ColumnInfo, String> entry : properties.entrySet())
         {
-            validateProperty(context, entry.getKey(), entry.getValue(), errors);
+            validateProperty(context, entry.getKey(), entry.getValue(), cache, errors);
         }
         return errors;
     }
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
```diff
@@ -1011,30 +1014,55 @@ protected void savePropertyObject(ExpObject object, Container container, Map<Dom
...
+                lookupSchemaName, lookupQueryName, type.getJavaType(), errors);
+    }
+
+    private static void validateProperty(ContainerUser context, List<ColumnValidator> validators, String value,
+                                         String label, Boolean required, String lookupSchemaName, String lookupQueryName,
+                                         Class type, List<ValidationError> errors)
     {
         boolean missing = (value == null || value.length() == 0);
         int rowNum = 0;
+        RemapCache cache = new RemapCache();
```
labkey-kevink: Can the remap cache be shared across columns and rows?

labkey-kevink (approved):

api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
cnathe: Yes. Fixed.

labkey-klum (approved):

PR #2430, opened 2021-07-06
support VALUES in LabKey SQL

labkey-matthewb:
#### Rationale
https://www.postgresql.org/docs/13/queries-values.html

### DIFF
--- a/query/src/org/labkey/query/sql/QValuesTable.java
+++ b/query/src/org/labkey/query/sql/QValuesTable.java
@@ -1,5 +1,6 @@
 package org.labkey.query.sql;
 
+import org.apache.logging.log4j.LogManager;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.data.BaseColumnInfo;
@@ -8,6 +9,7 @@ import org.labkey.api.data.JdbcType;
 import org.labkey.api.data.SQLFragment;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.query.FieldKey;
+import org.labkey.api.query.QueryParseException;
 import org.labkey.data.xml.ColumnType;
 
 import java.util.LinkedHashMap;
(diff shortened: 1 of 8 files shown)


query/src/org/labkey/query/sql/QValuesTable.java
```diff
@@ -0,0 +1,305 @@
...
+            {
+                methodName = (QExpr)expr.childList().get(0);
+                if (null == methodName.getFieldKey())
+                    methodName = null;
+            }
+
+            QExpr ret = (QExpr) expr.clone();
+            for (QNode child : expr.children())
+            {
+                //
```
labkey-susanh: Stray empty comment

query/src/org/labkey/query/sql/QValuesTable.java
```diff
@@ -0,0 +1,305 @@
...
+        public SQLFragment getFromSql()
+        {
+            SqlBuilder ret = new SqlBuilder(query.getSchema().getDbSchema().getSqlDialect());
+            ret.append("(");
+            appendSql(ret);
+            ret.append(") ");
+            ret.append(getAlias());
+            ret.append(" (");
+            String comma = "";
+            for (int i=1 ; i<= countOfColumns ; i++)
```
labkey-susanh, line 272: Why does this start at 1 instead of 0? Maybe add a comment if that's intended?

query/src/org/labkey/query/sql/SqlBase.g
```diff
@@ -563,19 +573,34 @@ annotation
...
 	: logicalOrExpression
 	;
 
+
+// public entry point (testing)
+parseConstantExpression
+	: simpleScalarExpression EOF!;
+
+
+// This is parses the same as 'expression', but leaves an marker node in the tree to indicate to SqlParser that
```
labkey-susanh:
```suggestion
// This parses the same as 'expression', but leaves a marker node in the tree to indicate to SqlParser that
```

labkey-susanh (approved):
I don't feel entirely qualified to pass judgment here, but it all seems reasonable enough.

PR #2433, opened 2021-07-07
Clear selection only for deleted items

labkey-kevink:
#### Rationale
The change in #2423 clears all selected items after deleting samples even if the samples can't be deleted.  The `SampleTypeLineageTest.testDeleteSamplesSomeWithDerivedSamples` was relying on this behavior, even if strange.  To not break the test, the selection is only cleared for samples that can be deleted.

#### Related Pull Requests
* #2423
* https://github.com/LabKey/testAutomation/pull/790

### DIFF
--- a/experiment/webapp/experiment/confirmDelete.js
+++ b/experiment/webapp/experiment/confirmDelete.js
@@ -87,7 +87,7 @@ LABKEY.experiment.confirmDelete = function(dataRegionName, schemaName, queryName
                             Ext4.Msg.hide();
                         }
                         else if (btn === 'ok') {
-                            let canDelete = response.data.canDelete;
+                            const canDelete = response.data.canDelete;
                             Ext4.Ajax.request({
                                 url: LABKEY.ActionURL.buildURL('query', 'deleteRows'),
                                 method: 'POST',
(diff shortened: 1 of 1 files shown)


experiment/webapp/experiment/confirmDelete.js
```diff
@@ -87,20 +87,25 @@ LABKEY.experiment.confirmDelete = function(dataRegionName, schemaName, queryName
                             Ext4.Msg.hide();
                         }
                         else if (btn === 'ok') {
+                            let canDelete = response.data.canDelete;
```
labkey-susanh: Could be `const` instead of `let` I believe.

experiment/webapp/experiment/confirmDelete.js
```diff
@@ -87,20 +87,25 @@ LABKEY.experiment.confirmDelete = function(dataRegionName, schemaName, queryName
...
                                     schemaName: schemaName,
                                     queryName: queryName,
-                                    rows: response.data.canDelete,
+                                    rows: canDelete,
                                     apiVersion: 13.2
                                 },
                                 success: LABKEY.Utils.getCallbackWrapper(function(response)  {
+                                    // clear the selection only for the rows that were deleted
+                                    // TODO: support clearing selection in query-deleteRows.api using a selectionKey
+                                    let ids = canDelete.map((row) => row.RowId);
```
labkey-susanh:
```suggestion
                                   const ids = canDelete.map((row) => row.RowId);
```

experiment/webapp/experiment/confirmDelete.js
```diff
@@ -87,20 +87,25 @@ LABKEY.experiment.confirmDelete = function(dataRegionName, schemaName, queryName
...
                                     let dr = LABKEY.DataRegions[dataRegionName];
                                     if (dr) {
-                                        dr.clearSelected({
-                                            selectionKey: selectionKey,
+                                        dr.setSelected({
+                                            ids: ids,
```
labkey-susanh:
```suggestion
                                            ids,
```

labkey-susanh (approved):
One reason for keeping the selection of undeleted samples is it allows you to effectively iterate up the lineage graph in order to delete a full set of samples.  I'm not sure how common that scenario is, but I've found it very handy.  This is, of course, just another example of our "very sticky selections" that I fear we'll never get consensus on.

PR #2437, opened 2021-07-07
More Double -> BigDecimal sequence nums

labkey-adam:
#### Rationale
`BigDecimal` is our preferred representation for sequence numbers; migrate some more cases


labkey-adam:
In addition to all the BVTs passing, all relevant tests in the five PostgreSQL Daily suites are green. (Just some unrelated, existing Luminex test failures.) The Dailies showed failures previously, pointing out the need for better BigDecimal normalization.

labkey-matthewb (approved):

PR #2442, opened 2021-07-09
Display trigger information in schema browser

labkey-kevink:
#### Rationale
To help developers with debugging triggers, list the triggers attached to a table in the schema browser.  Both Java and JavaScript triggers are supported.

Screen Shot 2021-07-09 at 12 28 45 AM

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/935
* https://github.com/LabKey/labkey-api-js/pull/105

#### Changes
- add `includeTriggers` to query-getQueryDetails.api
- render grid of triggers and the supported events
- add trigger information to the DataIterator trace logging

### DIFF
--- a/api/src/org/labkey/api/data/TableInfo.java
+++ b/api/src/org/labkey/api/data/TableInfo.java
@@ -408,6 +408,14 @@ public interface TableInfo extends TableDescription, HasPermission, SchemaTreeNo
         }
     }
 
+    enum TriggerMethod
+    {
+        init, complete,
+        beforeInsert, afterInsert,
+        beforeUpdate, afterUpdate,
+        beforeDelete, afterDelete
+    }
+
 
     /**
      * Queries may have named parameters, SELECT queries (the only kind we have right now) may
--- a/api/src/org/labkey/api/data/TableInfo.java
+++ b/api/src/org/labkey/api/data/TableInfo.java
@@ -408,6 +408,9 @@ public interface TableInfo extends TableDescription, HasPermission, SchemaTreeNo
         }
     }
 
+    /**
+     * Enumeration of trigger methods that may be implemented in trigger scripts.
+     */
     enum TriggerMethod
     {
         init, complete,
--- a/query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
+++ b/query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
@@ -34,6 +34,7 @@ import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.DisplayColumn;
 import org.labkey.api.data.JsonWriter;
 import org.labkey.api.data.TableInfo;
+import org.labkey.api.data.triggers.Trigger;
 import org.labkey.api.exp.property.Domain;
 import org.labkey.api.exp.property.DomainKind;
 import org.labkey.api.exp.property.PropertyService;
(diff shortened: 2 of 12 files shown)


api/src/org/labkey/api/data/TableInfo.java
```diff
@@ -408,6 +408,14 @@ public String getMethodName()
         }
     }
 
+    enum TriggerMethod
+    {
+        init, complete,
+        beforeInsert, afterInsert,
+        beforeUpdate, afterUpdate,
+        beforeDelete, afterDelete
```
cnathe, line 419: I think this makes sense here, but might be worth a comment pointing to or somehow referencing that these match the methods from Trigger.java interface.

query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
```diff
@@ -214,6 +215,18 @@ public ApiResponse execute(Form form, BindException errors)
...
 
+        if (form.isIncludeTriggers() && tinfo instanceof AbstractTableInfo)
+        {
+            Collection<Trigger> triggers = ((AbstractTableInfo)tinfo).getTriggers(container);
+            resp.put("triggers", triggers.stream().map(t -> new JSONObject()
+                    .put("name", t.getName())
+                    .put("description", t.getDescription())
+                    .put("module", t.getModuleName())
+                    .put("source", t.getSource())
+                    .put("events", t.getEvents())
```
cnathe: minor, but would it make sense to have this as a toJSON method on the Trigger class itself to keep it closer to the source?

query/webapp/query/browser/Caches.js
```diff
@@ -146,6 +146,8 @@ Ext4.define('LABKEY.query.browser.cache.QueryDetails', {
                 schemaName: '' + schemaName, // stringify LABKEY.SchemaKey
                 queryName: queryName,
                 fk: fk,
+                // Only fetch trigger scripts for the top-level table or query
+                includeTriggers: fk === undefined,
```
cnathe, line 150: Good call!

cnathe (approved):
Nice. This will be great to show on the schema browser. 

I did try pulling down the FB locally to take a look but was getting a build error (looks like TC is as well). I assume it just needs a merge in from develop to the fb or something along those lines, so I'll approve the PR now.

labkey-jeckels:
This looks nice! A few questions on the UI itself based on the screenshot here in the PR:

Will there ever be a "Description" value for triggers? If not, can we trim that column to save space?

For the non-JS entries in the screenshot, what are those? Java class names? Could we show the fully qualified name instead?

How are they ordered? The sequence in which they'll be executed?

labkey-kevink:
> This looks nice! A few questions on the UI itself based on the screenshot here in the PR:
> 
> Will there ever be a "Description" value for triggers? If not, can we trim that column to save space?
>

I've added some descriptions to the Biologics triggers.  I don't yet have any thoughts about how to add descriptions to the JavaScript triggers.

> 
> For the non-JS entries in the screenshot, what are those? Java class names? Could we show the fully qualified name instead?
>

They are the Java class simple name.  We could show the full class name, but I don't think it's necessary to identify the trigger implementation.  I've included the full class name in a "source" property, if we need it.  I'd like to add a link to the JavaScript triggers to launch the module editor using the module name and the source.

> 
> How are they ordered? The sequence in which they'll be executed?

Yes, they are listed in execution order.

PR #2234, opened 2021-05-04
Filewatcher UI

labkey-alan:
#### Rationale
This PR adds FormSchema and related classes to allow the server to instruct the client on how to render a form (via AutoForm in the related ui-components branch), as well as refactors FileWatchers to send FormSchema objects to the client.

I have added relevant methods to PipelineJobService, FileAnalysisTaskPipelineSettings, and FileAnalysisTaskPipeline to serialize FormSchema objects to the client. I have also added to the pipelineTasks.xsd XML Schema to allow Script Pipelines to define custom fields.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/519
* https://github.com/LabKey/testAutomation/pull/713
* https://github.com/LabKey/premium/pull/245

#### Changes
* Add FormSchema, AbstractField, TextField, TextAreaField, NumberField, CheckboxField, SelecField, RadioField, Option classes
* Add getFormSchema to PipelineJobService
* Add getFormSchema, getCustomFieldsFormSchema to FileAnalysisTaskPipeline
* Add getCustomFields to FileAnalysisTaskPipelineSettings
* Add CreatePipelineTrigger React component
* Refactor createPipelineTrigger.jsp to use CreatePipelineTrigger component
* Add custom fields to study, specimen, and list FileWatchers

PR #2443, opened 2021-07-09
Issue 43526: Add experimental feature to test usages of QuerySelect component in place of DataColumn renderSelectFormInput

cnathe:
#### Rationale
Issue [43526](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43526): Use the React QuerySelect / SelectInput components on the row insert/update form to allow better UX for lookups with >10,000 options/rows

#### Changes
* Add experimental feature EXPERIMENTAL_USE_QUERYSELECT_COMPONENT
* Add core module entryPoint for wrapper around React QuerySelect component
* Conditionally use renderQuerySelectFormInput() instead of renderSelectFormInput() in DataColumn.java


labkey-nicka (approved):
This is great. As we discussed a few things to consider in the future:

- Consider overlap of this and AutoForm. See https://github.com/LabKey/platform/pull/2234.
- Using `<SelectInput />` for non-lookup selects.

labkey-kevink (approved):
This is great!

PR #2444, opened 2021-07-12
Item 8788: Find samples using sample Ids or barcodes

labkey-susanh:
#### Rationale
When scanning barcodes or coming to the application with a set of known ids, users want to be able to find these samples quickly and act on them in bulk. To support this, we need to provide a way to search for samples across all types for Sample Ids and search for barcodes in all of the Unique ID fields for all sample types.  The sample IDs (which are the names of the samples, not their rowIds) are not unique across sample types.  The Unique IDs are globally unique but any sample type may have more than one (or none) such fields.  Furthermore, it is important that the ids provided are handled in the order in which they are provided so that if scanning the samples to be put into a freezer, they are assigned to spots in the storage box in the order provided.  For this, we require an ordering column attached to our session query used to select the various samples.

#### Related Pull Requests
* 

#### Changes
* Add `ExpMaterialUniqueIdUnionTableInfo` virtual table for gathering together all data from the 
* Add index for UniqueId columns in sample types
* Add `SaveOrderedSamplesQueryAction` to `ExperimentController`


experiment/src/org/labkey/experiment/api/ExpMaterialUniqueIdUnionTableInfo.java
```diff
@@ -0,0 +1,103 @@
...
+import org.labkey.api.query.QueryService;
+import org.labkey.api.query.UserSchema;
+import org.labkey.api.security.User;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.labkey.api.exp.api.ExperimentJSONConverter.ROW_ID;
+
+public class ExpMaterialUniqueIdUnionTableInfo extends VirtualTable
```
labkey-matthewb: This doesn't really need to be TableInfo at all, it's just a helper class with one method getSql().

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -6712,4 +6714,191 @@ public void addNavTrail(NavTree root)
...
+            int index = 1;
+            SQLFragment sampleIdValuesSql = new SQLFragment();
+            SQLFragment uniqueIdValuesSql = new SQLFragment();
+            for (String id : form.getIds())
+            {
+                if (id.startsWith(SAMPLE_ID_PREFIX))
+                {
+                    sampleIdValuesSql.append(sampleIdComma).append("\t(").append(index);
+                    sampleIdValuesSql.append(", ");
+                    sampleIdValuesSql.append("'").append(id.substring(SAMPLE_ID_PREFIX.length())).append("'");
```
labkey-matthewb: DialectStringHandler.quoteStringLiteral()

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -6712,4 +6714,191 @@ public void addNavTrail(NavTree root)
...
+                    sampleIdValuesSql.append(")");
+                    sampleIdComma = "\n,";
+                }
+                else if (id.startsWith(UNIQUE_ID_PREFIX))
+                {
+                    uniqueIdValuesSql.append(uniqueIdComma).append("\t(").append(index);
+                    uniqueIdValuesSql.append(", ");
+                    uniqueIdValuesSql.append("'").append(id.substring(UNIQUE_ID_PREFIX.length())).append("'");
```
labkey-matthewb: DialectStringHandler.quoteStringLiteral()

experiment/src/org/labkey/experiment/api/property/DomainImpl.java
```diff
@@ -785,7 +785,7 @@ private void ensureUniqueIdValues(List<DomainProperty> propsAdded) throws SQLExc
 
         List<ColumnInfo> uniqueIndexCols = new ArrayList<>();
         table.getUniqueIndices().values().forEach(idx -> {
-            uniqueIndexCols.addAll(idx.second);
+            idx.second.stream().filter(col -> !col.isUniqueIdField()).forEach(uniqueIndexCols::add);
```
labkey-matthewb, line 790: comments?  Got lost on what's happening here.

internal/src/org/labkey/experiment/api/SampleTypeDomainKind.java
```diff
@@ -238,7 +234,13 @@ public ActionURL urlEditDefinition(Domain domain, ContainerUser containerUser)
     @Override
     public Set<PropertyStorageSpec.Index> getPropertyIndices(Domain domain)
     {
-        return INDEXES;
+        Set<PropertyStorageSpec.Index> indexes = Sets.newLinkedHashSet(Arrays.asList(
+                new PropertyStorageSpec.Index(true, "lsid")
+        ));
+        domain.getProperties().stream().filter(DomainProperty::isUniqueIdField).forEach(prop -> {
```
labkey-matthewb: If we index on (prop,"RowId") then query engine should be able to execute the select directly using the index w/o touching the base table.  Once the engine finds a matching row in the index for "INNER JOIN exp.materials M ON _ordered_ids_.column2 = M.Name", it can do the "SELECT M.RowId" with the index entry it has in hand.

labkey-matthewb (changes requested):

internal/src/org/labkey/experiment/api/SampleTypeDomainKind.java
labkey-susanh: So, are you suggesting adding "RowId" as one of the columns?  Since these uniqueId values are, in fact, unique independent of the rowId, I thought it redundant, but you seem to be saying adding that column will be a hint at even more efficient behavior, right?

experiment/src/org/labkey/experiment/api/ExpMaterialUniqueIdUnionTableInfo.java
labkey-susanh: Ah, good thought.  That will be better.

experiment/src/org/labkey/experiment/api/property/DomainImpl.java
labkey-susanh: Done.

internal/src/org/labkey/experiment/api/SampleTypeDomainKind.java
labkey-susanh: Hmm.  Well, that didn't work, but maybe I've misunderstood the intent. "RowId" isn't a column on the sample type provisioned table, only on the exp.materials table.  I could use "lsid" here together with `prop.name()`.  Would that achieve the same thing @labkey-matthewb?

labkey-matthewb (approved):

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-nicka: We're generating LabKey SQL here so this should use `LabKeySql.quoteString()`.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-nicka: We're generating LabKey SQL here so this should use `LabKeySql.quoteString()`.

PR #2445, opened 2021-07-12
Issue 43470: Cannot import the output file from a provenance run into an assay

labkey-kevink:
#### Rationale
LabKey models assay runs as producing the data file that is imported into the server as the *output* of the assay run.  In some cases, a lab may want to transform the data file using an external process (e.g., normalization) prior to importing the assay into the server and model the transformation as an experiment run as well.  Unfortunately, we run into the LabKey constraint that requires a data file to only be produced by a single source run.

This patch, adds a `allowCrossRunFileInputs` flag to the assay import process.  When importing from the filebrowser, the user will be prompted to confirm the would like to import the file if it has been created by another run.  Once confirmed, the file will be added as an input into the assay run and create attach a new exp.data as the assay output. The result is a run graph similar to:

Screen Shot 2021-07-12 at 8 53 46 AM

#### Related Pull Requests
* 

#### Changes
* link to other exp.data with the same dataFileUrl from the exp.data details page
* add `AssayRunUploadContext.isAllowCrossRunFileInputs` flag to allow importing a file already claimed by another run
* junit test


api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
```diff
@@ -831,7 +890,13 @@ protected void addOutputDatas(AssayRunUploadContext<ProviderType> context, Map<E
             else
                 dataType = context.getProvider().getDataType();
 
-            ExpData data = DefaultAssayRunCreator.createData(context.getContainer(), file, file.getName(), dataType, context.getReRunId() == null);
+            // Reuse existing exp.data as the assay output file unless we are reimporting the
+            // file again (via the rerun mechanism) or the output file already one of the input files.
```
labkey-jeckels:
```suggestion
            // file again (via the rerun mechanism) or the output file is already one of the input files.

```

assay/src/org/labkey/assay/actions/PipelineDataCollectorRedirectAction.java
```diff
@@ -125,28 +166,80 @@ public ModelAndView getView(UploadRedirectForm form, BindException errors)
...
-        throw new RedirectException(AssayService.get().getProvider(form.getProtocol()).getImportURL(container, form.getProtocol()));
+
+        ActionURL url = AssayService.get().getProvider(form.getProtocol()).getImportURL(container, form.getProtocol());
+        if (form.isAllowCrossRunFileInputs())
+            url.addParameter("allowCrossRunFileInputs", "true");
+
+        throw new RedirectException(url);
+    }
+
+    private class FakeConfirmAction extends ConfirmAction<Object>
```
labkey-jeckels, line 184: This confused me for a while. It's just a way to get a little configuration to the JSP? If so, a comment to that effect would be nice

labkey-jeckels (approved):

assay/src/org/labkey/assay/actions/PipelineDataCollectorRedirectAction.java
labkey-kevink: Yes, since we are POSTing to this action when importing from the filecontent browser, I can't use the standard ConfirmAction.  I added a comment with some details.

PR #2447, opened 2021-07-12
Disallow overriding email template replacement parameters

labkey-adam:
#### Rationale
`EmailTemplate` seems to let subclasses override replacement parameters, but the behavior is inconsistent (e.g., duplicate parameters appear on the customize page). This just seems like a bad idea, so remove existing overrides and add a check to prevent others in the future. Should fix `MessagesLongTest`, which flagged a related change in behavior.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2439

#### Changes
* `AnnouncementDigestProvider`: remove redundant override of `folderName`; just use `folderPath`
* `ReportAndDatasetChangeDigestEmailTemplate`: remove redundant overrides of `folderUrl` and `folderPath`
* Add assert that disallows adding a parameter that matches an existing parameter name for that template

### DIFF
--- a/announcements/src/org/labkey/announcements/model/AnnouncementDigestProvider.java
+++ b/announcements/src/org/labkey/announcements/model/AnnouncementDigestProvider.java
@@ -138,7 +138,7 @@ public class AnnouncementDigestProvider implements MessageDigest.Provider
 
     public static class DailyDigestEmailTemplate extends EmailTemplate
     {
-        protected static final String DEFAULT_SUBJECT = "New posts to ^folderName^";
+        protected static final String DEFAULT_SUBJECT = "New posts to ^folderPath^";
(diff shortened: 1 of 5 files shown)


labkey-jeckels (approved):
Are we case-insensitive when doing the substitutions into the template? If not, we probably should make the change.

labkey-adam:
The two places that enumerate `getAllReplacements()` resolve names via `equalsIgnoreCase()`. Would be nice if these were held in a (case-insensitive) map instead of constantly enumerating, but I'm done making changes for now.

labkey-adam:
MessagesLongTest passed with this change: https://teamcity.labkey.org/viewLog.html?buildId=1474524&buildTypeId=bt20&tab=buildResultsDiv&branch_LabkeyTrunk_DailySuites=foldName_vs_folderPath

PR #2449, opened 2021-07-13
Issues list related field - fix for multi value FK to not attempt to resolve lookup by value

cnathe:
#### Rationale
See related PR. The change their removed the experimental feature that was conditionally attempting to resolve a lookup by value in the event of a conversion error (i.e. the lookup points to an integer PK but the user is providing the display value) on data entry. This ended up breaking the Issues list related issues field (as caught by the IssuesTest case). This PR fixes this case by skipping the "resolve by value" code in the multi value FK case.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2429

#### Changes
* TableViewForm update to if statement to skip "resolve lookup by display value" if not fk.allowImportByAlternateKey()

### DIFF
--- a/api/src/org/labkey/api/data/TableViewForm.java
+++ b/api/src/org/labkey/api/data/TableViewForm.java
@@ -474,7 +474,7 @@ public class TableViewForm extends ViewForm implements DynaBean, HasBindParamete
 
                 // Attempt to resolve lookups by display value
                 ColumnInfo col = getColumnByFormFieldName(propName);
-                if (col != null && col.getFk() != null && !(col.getFk() instanceof MultiValuedForeignKey))
+                if (col != null && col.getFk() != null && col.getFk().allowImportByAlternateKey())
                 {
                     ForeignKey fk = col.getFk();
                     Container container = fk.getLookupContainer() != null ? fk.getLookupContainer() : getContainer();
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -474,7 +474,7 @@ else if (_validateRequired && null != _tinfo)
 
                 // Attempt to resolve lookups by display value
                 ColumnInfo col = getColumnByFormFieldName(propName);
-                if (col != null && col.getFk() != null)
+                if (col != null && col.getFk() != null && !(col.getFk() instanceof MultiValuedForeignKey))
```
cnathe: @labkey-kevink do you think we need to add a check like this for other usages of the RemapCache cases?

labkey-kevink: I'm not completely sure, but you may be able to use `fk.allowImportByAlternateKey()` in this case.  The `SimpleTranslator.createConvertColumn()` sets up a `RemapPostConvertColumn` around like 1292 and then checks that the foreign key is multivalued around line 1296 so it's likely we support remapping with multivalued FKs.

cnathe: Yup, that looks to have worked. It looks like the ExpRunTableImpl.updateRow() case already had that `fk.allowImportByAlternateKey()` check.

labkey-kevink (approved):

PR #2457, opened 2021-07-14
Issue 43557: JSON date format doesn't parse in Safari

labkey-jeckels:
#### Rationale
When we changed our JSON date representation to include milliseconds, we ended up adopting a variant that doesn't parse with Safari's standard Date JS object

#### Changes
* Use a format closer to ISO that includes a T between the date and time, which appears to parse in all browsers and hopefully other client API languages

### DIFF
--- a/api/src/org/labkey/api/util/DateUtil.java
+++ b/api/src/org/labkey/api/util/DateUtil.java
@@ -32,6 +32,8 @@ import org.labkey.api.pipeline.PipelineJobService;
 import org.labkey.api.settings.DateParsingMode;
 import org.labkey.api.settings.FolderSettingsCache;
 import org.labkey.api.settings.LookAndFeelProperties;
+import org.labkey.api.view.HttpView;
+import org.labkey.api.view.ViewContext;
 
 import javax.xml.datatype.DatatypeConfigurationException;
 import javax.xml.datatype.DatatypeFactory;
@@ -51,6 +53,7 @@ import java.util.TimeZone;
 import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -959,14 +959,14 @@ public static String getJsonDateTimeFormatString()
...
-        // Without millisecond portion here, datetime comparision would show these datetimes as equals,
+        // Without millisecond portion here, datetime comparison would show these datetimes as equals,
         // ex: 2019/01/01 11:11:11.117 and 2019/01/01 11:11:11.399 - since milliseconds would be ignored, this would result in No Work.
 
         // 2. Separately, adding .SSS to the existing string (yyyy/MM/dd HH:mm:ss) worked on Chrome (and MS Edge), but not on Firefox.
         // Strangely, Firefox seem to adhere strictly to ECMA Specification of date string interchange format (i.e. date with
         // hyphens instead of forward slashes) in this particular case with milliseconds. Hence had to modify the previous string from yyyy/MM/dd to yyyy-MM-dd.
         // Chrome seem to behave as expected with this change (and so does MS Edge, but extensive testing has not been done on this browser).
-        return "yyyy-MM-dd HH:mm:ss.SSS";
+        return "yyyy-MM-dd'T'HH:mm:ss.SSS";
```
labkey-bpatel, line 978: Does T have to be within single quotes?

labkey-jeckels: Yes, I think so. https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html

Within date and time pattern strings, unquoted letters from 'A' to 'Z' and from 'a' to 'z' are interpreted as pattern letters representing the components of a date or time string. Text can be quoted using single quotes (') to avoid interpretation. "''" represents a single quote

Here's a JSON snippet with this change in use:

   "Created" : "2021-07-14T15:25:12.674",

labkey-bpatel: ok, good to know! thanks!

labkey-bpatel (approved):

api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -1110,12 +1119,32 @@ public static boolean isSpecialNamedFormat(String dateFormat)
...
+        // Issue 43557 - Safari needs a 'T' between the date and time. Instead of needing to rev all client APIs to parse
+        // the new variant, conditionally send Safari its preferred format (which works fine in other browsers but not
+        // in all other parsing code)
+        boolean isSafari = false;
+        if (HttpView.hasCurrentView())
+        {
+            ViewContext context = HttpView.currentContext();
+            if (context != null && context.getRequest() != null)
+            {
+                String userAgent = context.getRequest().getHeader("User-Agent");
```
labkey-nicka: You could consider adding this check for `isSafari` to `HttpUtil.java`. There are other methods that do similar checks.

labkey-nicka (approved):

api/src/org/labkey/api/util/DateUtil.java
labkey-jeckels: Done.

PR #2458, opened 2021-07-15
Item 9204: Sample type parent import alias inclusion in download template and display in details panel

cnathe:
#### Rationale
Sample type definitions allow for parent import aliases to be set to make it easier for users to map columns in the sample import files to the parent samples / sources. This set of PRs exposes those parent import aliases in the sample type details panel and download template column headers.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/581
* https://github.com/LabKey/platform/pull/2458
* https://github.com/LabKey/sampleManagement/pull/628
* https://github.com/LabKey/biologics/pull/941

#### Changes
* Add SampleTypes listing table aliased columns for displaying MaterialInputImportAliases and DataInputImportAliases
* Fix createdBy and modifiedBy lookups by removing the core.Users <fk>
* Show "Parent Import Alias(es)" in Sample Type Properties panel
* Include sample type parent import aliases in download template file for importSamples action


experiment/resources/schemas/exp.xml
```diff
@@ -1,1146 +1,1106 @@
...
-      <column columnName="EntityId"/>
-      <column columnName="Created">
-        <description>Contains the time at which this run group was created</description>
-      </column>
-      <column columnName="CreatedBy">
-        <fk>
-          <fkColumnName>UserId</fkColumnName>
-          <fkTable>Users</fkTable>
-          <fkDbSchema>core</fkDbSchema>
-        </fk>
```
labkey-susanh, line 121: These `fk` changes don't seem related to this PR.  Are they intentional?

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -919,4 +922,24 @@ public DataIteratorBuilder persistRows(DataIteratorBuilder data, DataIteratorCon
...
+    {
+        List<Pair<String, String>> templates = new ArrayList<>();
+        ActionURL url = PageFlowUtil.urlProvider(QueryUrls.class).urlCreateExcelTemplate(ctx.getContainer(), getPublicSchemaName(), getName());
+        url.addParameter("headerType", ColumnHeaderType.DisplayFieldKey.name());
+        try
+        {
+            if (getSampleType() != null && !getSampleType().getImportAliasMap().isEmpty())
+            {
+                for (String aliasKey : getSampleType().getImportAliasMap().keySet())
+                    url.addParameter("includeColumn", aliasKey);
```
labkey-susanh, line 960: I wonder how many of these aliases people define and if we'll have issues with URLs that get to be too long.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -652,6 +652,19 @@ public ModelAndView getView(ExpObjectForm form, BindException errors)
                 detailsView.getDataRegion().addDisplayColumn(parentCol);
             }
 
+            try
+            {
+                SimpleDisplayColumn importAliasCol = new SimpleDisplayColumn();
+                importAliasCol.setCaption("Parent Import Alias(es):");
+                if (!_sampleType.getImportAliasMap().isEmpty())
+                    importAliasCol.setDisplayHtml(PageFlowUtil.filter(StringUtils.join(_sampleType.getImportAliasMap().keySet(), ", ")));
```
labkey-susanh, line 665: I wonder if we want to show the name of the sample type being aliased as well?

labkey-susanh (approved):

experiment/resources/schemas/exp.xml
cnathe: Yes, intentional. I noticed in the apps that the createdBy and modifiedBy values for the sample type listing grid wasn't showing as clickable user display names but the source type listing grid was. It turned out that there were a few tables in this schema with the same issue. I can revert this change and open as a separate PR if you think it might have unrelated consequences.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
cnathe: Good question. Do you think this is a "wait and see if we get feedback / issues" or if this is something we want to try and address now?

experiment/resources/schemas/exp.xml
labkey-susanh: No need for a separate PR.  Just checking.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
cnathe: Yeah, I thought that too, but my assumption is that these aliases are likely to be pretty self explanatory in practice. Something like "BLD" instead of "Blood" or even just setting an alias as "Blood" so that you don't need to use an import column of "materialInput/Blood". 

Happy to add the target sample type / source type name if you think otherwise. @labkey-hannah Thoughts?

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
labkey-susanh: This is affected, of course, by our addition of all the storage columns ad "includeColumn" values as well for SM.  My guess is the number of aliases is small so I think we can afford to wait.  An obvious modification might be to change the name of the parameter to something shorter than "includeColumn".  That's nicely descriptive but takes up 13 characters for each one.  Of course, it's been that way for a good long while.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-hannah: I think being overly clear is always helpful - so if it's easy to show the sample type name let's do it

cnathe: Sure thing. Do you like display the target sample/source type in parenthesis or with an arrow? @labkey-susanh @labkey-hannah 

Screen Shot 2021-07-16 at 3 28 01 PM
Screen Shot 2021-07-16 at 3 26 09 PM

labkey-hannah: I like the arrows, since it's a little more clear that my column name wouldn't be Blood Parent Alias (Blood), but rather just Blood Parent Alias

labkey-susanh: I wonder if we want to be less cryptic here and say something like "Blood Alias (Sample Type: Blood)" with the link only on the actual sample type name.  I'm not sure how often the name of the alias will be the same as or very similar to the sample type name and is added just to get away from having to provide the clunky syntax required otherwise, so it won't be clear what's on the left and what's on the right of the arrow. I also wonder if a little help tip for these might be called for.

(Sorry, should have thought of this earlier.)

labkey-hannah: That's a good suggestion - the links on both seemed confusing to me, too. I don't want this to drastically increase scope too much, so if we can address this in what seems like a reasonable effort lets do it - otherwise we're still improving behavior as-is

cnathe: Makes sense to me. Thoughts on something like this? We can put in the "Sample Type" or "Source Type" text in place of "Alias Target" if you'd like (I just didn't do that in the mockup as it was a bit more work). Also, is the clickable part better in this case? @labkey-susanh @labkey-hannah 

Screen Shot 2021-07-16 at 4 31 37 PM

labkey-susanh: Yes, I like that.  I might change the wording to "Alias for" instead of "Alias Target".

PR #2460, opened 2021-07-15
Item 9252: Add action to save find Ids to HTTP session instead of browser session

labkey-susanh:
#### Rationale
The browsers sessionStorage does not get cleared until a user closes the browser, but an https session key will be cleared when the session expires, so it's a better candidate for storing these ids.

#### Related Pull Requests
* #2444 
* https://github.com/LabKey/labkey-ui-components/pull/582
* https://github.com/LabKey/sampleManagement/pull/629
* https://github.com/LabKey/inventory/pull/277
* https://github.com/LabKey/biologics/pull/942

#### Changes
* Add new `SaveFindIdsAction` in `ExperimentController`
* Update `SaveOrderedSamplesQueryAction` to retrieve ids from session instead of from query paraemter

### DIFF
--- a/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
+++ b/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
@@ -44,6 +44,7 @@ import org.labkey.api.action.ReadOnlyApiAction;
 import org.labkey.api.action.ReturnUrlForm;
 import org.labkey.api.action.SimpleApiJsonForm;
 import org.labkey.api.action.SimpleErrorView;
+import org.labkey.api.action.SimpleResponse;
 import org.labkey.api.action.SimpleViewAction;
 import org.labkey.api.action.SpringActionController;
 import org.labkey.api.assay.AssayFileWriter;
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -6771,10 +6832,26 @@ public Object execute(OrderedSamplesForm form, BindException errors) throws Exce
                     "   </table>\n" +
                     "</tables>";
             QueryDefinition def = QueryService.get().saveSessionQuery(getViewContext(), getContainer(), ExperimentServiceImpl.get().getExpSchema().getName(), select.getSQL(), metadata);
-            return success("Session query created", def.getName());
+            return success("Session query created", Map.of("queryName", def.getName(), "ids", _ids));
         }
 
-        private SQLFragment getOrderedRowsSql(OrderedSamplesForm form)
+
+        private List<String> getFindIdsFromSession(String sessionKey)
```
cnathe, line 6839: This may not be very helpful to share/refactor, but this type of code exists within the SaveFindIdsAction.execute() method as well. Any reason to share it?

cnathe (approved):

PR #2463, opened 2021-07-15
Issue 43567: Add an option to set the default folder type on the Folder Types admin settings page

vagisha:
#### Rationale
We would like to add an admin setting on PanoramaWeb that will let us change the default folder type that is chosen when users create a new folder. The default option is "Collaboration" but most of the time users want a "Panorama" folder on PanoramaWeb and are confused when they are unable to upload Skyline documents to a new folder or don't see the familiar folder layout because they forgot to select "Panorama" when they created the folder.


labkey-jeckels:
@vagisha I created a copy of your feature branch in the LabKey copy of the platform repo to get TeamCity to run builds and tests. The code looks good. I'll approve after the tests complete.

labkey-jeckels (approved):
Looks good.

PR #2472, opened 2021-07-19
Issue 43278- display forbidden exception message in error handling page

labkey-ankurj:
#### Rationale
Display forbidden exception message in error page instead of the general permission exception message.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2452

### DIFF
--- a/api/src/org/labkey/api/action/PermissionCheckableAction.java
+++ b/api/src/org/labkey/api/action/PermissionCheckableAction.java
@@ -41,7 +41,6 @@ import org.labkey.api.security.roles.RoleManager;
 import org.labkey.api.util.ConfigurationException;
 import org.labkey.api.util.HttpUtil;
 import org.labkey.api.view.BadRequestException;
-import org.labkey.api.view.ForbiddenProjectException;
 import org.labkey.api.view.NotFoundException;
 import org.labkey.api.view.RedirectException;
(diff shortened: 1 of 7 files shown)


labkey-ankurj:
@labkey-adam I couldn't find the server side callers to LoadLibraryAction and LogClientExceptionAction. The ones I found were client side and the caller to those actions doesn't pass the container.

labkey-adam (changes requested):
A key change mentioned in the issue is to remove the `@IgnoresForbiddenProjectCheck` annotation on both actions. And test. I had to add the annotation was because these actions were getting called in a specific folder/project instead of root. Since you've made no changes to the callers I suspect they're still NOT root-based URLs.

labkey-adam:
Actually, hold on... let me pull this down locally. Something is strange with the isForbiddenProject handling.

labkey-adam:
Sorry, @labkey-ankurj , I was mistaken. Turns out the "forbidden project" check was NOT failing originally because the invocation was to a different project. It forbid an admin impersonating in a project from accessing root, which is arguably too strict, but that's what's it's doing. Let's forget about removing the annotation... your code is good!

But I'd like to make my follow-on change to this branch... I'll let you know when my change is ready.

labkey-adam (approved):
Looks good. Let me make my follow-on commit here.

PR #2477, opened 2021-07-20
Feature Request 43224: TypeAheadSelectDisplayColumn to render select input using React QuerySelect

cnathe:
#### Rationale
The related PR added an experimental feature to toggle the LabKey insert/update forms to use the React QuerySelect component in place of a standard HTML select input. This experimental feature is nice for testing, but is an "all or nothing" approach. This PR adds a new TypeAheadSelectDisplayColumn factory which can be set for a table's XML metadata to opt-in to using this React component for given columns.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2443

#### Changes
* factor out TypeAheadSelectDisplayColumn from DataColumn 
* use TypeAheadSelectDisplayColumn in DataColumn.renderInputHtml() when experimental feature enabled
* allow for "maxRows" value to be passed as XML `<properties>` for the `<displayColumnFactory>` definition

PR #2480, opened 2021-07-21
Feature Request 43224: TypeAheadSelectDisplayColumn factory which renders React QuerySelect component

cnathe:
#### Rationale
The related PR added a new TypeAheadSelectDisplayColumn factory which can be set for a table's XML metadata to opt-in to using the React QuerySelect component for given columns in the LK insert/update forms. Since the client is planning to install 21.7 soon, this PR backports the core/entryPoint.js code and the TypeAheadSelectDisplayColumn class to 21.7 so that it can be used on their deployment.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2477

#### Changes
* Minor DataColumn.java refactor to pull out a getStringValue() method for use in subclasses
* TypeAheadSelectDisplayColumn to add custom rendering of React entryPoint for QuerySelect component
* core/entryPoint for QuerySelectInput wrapper over QuerySelect component to pass props from columnInfo

### DIFF
--- a/api/src/org/labkey/api/view/TypeAheadSelectDisplayColumn.java
+++ b/api/src/org/labkey/api/view/TypeAheadSelectDisplayColumn.java
@@ -78,7 +78,7 @@ public class TypeAheadSelectDisplayColumn extends DataColumn
         sb.append(" });\n");
         sb.append("});\n");
         sb.append("</script>\n");
-        sb.append("<div id='").append(renderId).append("'></div>");
+        sb.append("<div id=").append(PageFlowUtil.jsString(renderId)).append("></div>");
         out.write(sb.toString());
 
         // disabled inputs are not posted with the form, so we output a hidden form element:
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/view/TypeAheadSelectDisplayColumn.java
```diff
@@ -0,0 +1,118 @@
...
+        public Factory(MultiValuedMap<String, String> map)
+        {
+            String maxRowsStr = getProperty(map, "maxRows");
+            _maxRows = maxRowsStr != null ? Integer.parseInt(maxRowsStr) : null;
+        }
+
+        private String getProperty(MultiValuedMap<String, String> map, String propertyName)
+        {
+            Collection<String> values = map == null ? Collections.emptyList() : map.get(propertyName);
+            if (!values.isEmpty())
```
labkey-susanh: Is it possible that the value in the map is null?

labkey-susanh (approved):

api/src/org/labkey/api/view/TypeAheadSelectDisplayColumn.java
cnathe: In practice, it doesn't look like it can be....but doesn't hurt to add a check. Done.

PR #2485, opened 2021-07-21
Issue 43428: Finding user-defined queries

labkey-jeckels:
#### Rationale
Now that we show a unified list of both LabKey SQL queries and tables, users have requested that we make it easier to show only certain categories of entries.

Note that because ExtJS trees don't filter as advertised, it was most pragmatic to reload the whole page when the filtering changes (as we already did for the show/hide of hidden queries and schemas). Because the currently selected query may disappear or reappear based on the filter, I didn't spend a lot of time trying to preserve the current selection.

#### Changes
* Checkboxes to show/hide user-defined queries, module-defined queries, and system tables
* New icons to differentiate between user- and module-defined queries
* Show icons in query details pane as well


query/webapp/query/browser/Tree.js
```diff
@@ -111,17 +143,14 @@ Ext4.define('LABKEY.query.browser.Tree', {
...
         var params = LABKEY.ActionURL.getParameters();
-        params.showHidden = showHidden;
+        params[propertyName] = value;
 
         var url = LABKEY.ActionURL.buildURL('query', 'begin', null, params);
-        if (showHidden)
-            window.location.href = url + window.location.hash;
-        else
-            window.location.href = url;
+        window.location.href = url;
```
cnathe: I think we can preserve the users current selection in the schema browser tree by putting the url hash back onto the new URL:
`window.location.href = url + window.location.hash;`

query/webapp/query/browser/Tree.js
```diff
@@ -15,13 +15,21 @@ Ext4.define('LABKEY.query.browser.Tree', {
     constructor : function(config) {
 
         this.showHidden = LABKEY.ActionURL.getParameters().showHidden === 'true';
+        this.showUserDefined = LABKEY.ActionURL.getParameters().showUserDefined !== 'false';
+        this.showModuleDefined = LABKEY.ActionURL.getParameters().showModuleDefined !== 'false';
+        this.showSystemTables = LABKEY.ActionURL.getParameters().showSystemTables !== 'false';
```
cnathe: Using the URL params for this to match `showHidden` makes sense. I could see in the future using the browser session storage as a way to stash these props/settings so that they could follow the user as they navigate to other folders or to and from the schema browser.

query/webapp/query/browser/Tree.js
```diff
@@ -43,7 +51,7 @@ Ext4.define('LABKEY.query.browser.Tree', {
                     {name: 'text'},
                     {name: 'icon', convert: function(_, rec) {
                         if (rec.raw.table === undefined) { return undefined; }
-                        return LABKEY.ActionURL.getContextPath() + '/reports/' + (rec.raw.table ? 'grid.gif' : 'grid-sql.png');
+                        return LABKEY.ActionURL.getContextPath() + '/reports/' + (rec.raw.table ? 'grid.gif' : (rec.raw.fromModule ? 'grid-sql.png' : 'grid-sql-editable.png'));
```
cnathe, line 54: I like the different icon's for the 3 different types. The green makes the "user defined" queries stand out nicely in the longer list of queries.

query/webapp/query/browser/view/QueryDetails.js
```diff
@@ -597,14 +597,15 @@ Ext4.define('LABKEY.query.browser.view.QueryDetails', {
             children.push({
                 tag: 'span',
                 style: 'cursor: default;',
-                html: 'LabKey SQL query' + (queryDetails.moduleName ? ' defined in ' + Ext4.htmlEncode(queryDetails.moduleName) + ' module' : '')
+                html: '<img src="' + LABKEY.ActionURL.getContextPath() + '/reports/grid-sql' + (queryDetails.moduleName? '' : '-editable') + '.png" height="16" width="16" alt="' + (queryDetails.moduleName ? 'Module' : 'User') + '-defined query"/>' +
+                        '&nbsp;LabKey SQL query' + (queryDetails.moduleName ? ' defined in ' + Ext4.htmlEncode(queryDetails.moduleName) + ' module' : '')
```
cnathe: I like the inclusion of the icon here so that it matches between the tree view on the left and the filter/checkbox options. However, the text between the checkbox options and this text is different (which is a bit misleading). Both user defined and module queries are a type of "LabKey SQL query" so maybe this is fine.

query/webapp/query/browser/view/QueryDetails.js
```diff
@@ -597,14 +597,15 @@ Ext4.define('LABKEY.query.browser.view.QueryDetails', {
...
+                html: '<img src="' + LABKEY.ActionURL.getContextPath() + '/reports/grid-sql' + (queryDetails.moduleName? '' : '-editable') + '.png" height="16" width="16" alt="' + (queryDetails.moduleName ? 'Module' : 'User') + '-defined query"/>' +
+                        '&nbsp;LabKey SQL query' + (queryDetails.moduleName ? ' defined in ' + Ext4.htmlEncode(queryDetails.moduleName) + ' module' : '')
             });
         }
         else {
             children.push({
                 tag: 'span',
                 style: 'cursor: default;',
-                html: 'Built-in table'
+                html: '<img src="' + LABKEY.ActionURL.getContextPath() + '/reports/grid.gif" height="16" width="16" alt="System table"/>&nbsp;Built-in table'
```
cnathe, line 608: I like this text "Built-in table" better then the checkbox/filter text of "System tables". I'd suggest making these two match.

cnathe (changes requested):

cnathe:
I did some manual testing of this change as well. It looks good and I think that a page reload is fine (given the Ext tree filtering restrictions). If we add back the url hash for the user selection, I think that will improve the UX for the reload case.

query/webapp/query/browser/Tree.js
labkey-jeckels: We have some strange interactions between the GET parameters that are put on the URL when you navigate to the schema browser through the nav trail, but overall it seems OK to retain the hash. You may see the details from a query that no longer meets your filter criteria but I think that's fine. It doesn't show in the tree

query/webapp/query/browser/view/QueryDetails.js
labkey-jeckels: "LabKey SQL query" is too long to fold into the checkbox test, but I added a "User-defined" prefix for non-module queries in the details pane to make it more consistent

query/webapp/query/browser/view/QueryDetails.js
labkey-jeckels: Done

cnathe (approved):

PR #2487, opened 2021-07-22
43615: Can't create or load a dataset via a file watcher when the dataset name has a number in it.

labkey-klum:
#### Rationale
This fixes the problem detailed in this issue:
- https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43615

There are some scenarios where trying to create a new dataset using a file watcher will fail if the triggered file has a number in it's name.

#### Changes
- update the regular expression used in the file watcher schema reader so that all characters (in the name) are accepted.

### DIFF
--- a/study/src/org/labkey/study/pipeline/DatasetInferSchemaReader.java
+++ b/study/src/org/labkey/study/pipeline/DatasetInferSchemaReader.java
@@ -45,7 +45,8 @@ import java.util.stream.Collectors;
  */
 public class DatasetInferSchemaReader extends DatasetFileReader implements SchemaReader
 {
-    protected Pattern _filePattern = Pattern.compile("(^.*).(?:tsv|txt|xls|xlsx)$");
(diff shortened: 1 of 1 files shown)


study/src/org/labkey/study/pipeline/DatasetInferSchemaReader.java
```diff
@@ -45,7 +45,7 @@
  */
 public class DatasetInferSchemaReader extends DatasetFileReader implements SchemaReader
 {
-    protected Pattern _filePattern = Pattern.compile("(^\\D*).(?:tsv|txt|xls|xlsx)$");
+    protected Pattern _filePattern = Pattern.compile("(^.*).(?:tsv|txt|xls|xlsx)$");
```
labkey-adam:
```suggestion
    protected Pattern _filePattern = Pattern.compile("^(.*)\.(?:tsv|txt|xls|xlsx)$");

```

labkey-adam (approved):
minor suggestions to improve regex

PR #2466, opened 2021-07-16
fix Luminex re-run import to reuse existing exp.data

labkey-kevink:
#### Rationale
Fix various Luminex tests by restoring previous behavior when importing re-run:
  - create new exp.data when importing a re-run with assay provider supports ReRunReplace
  - related output data already claimed by another run are not added as outputs

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2445
* https://github.com/LabKey/commonAssays/pull/352

PR #2488, opened 2021-07-23
delete cross-run data outputs

labkey-kevink:
#### Rationale
When deleting a run, delete any duplicated exp.data outputs created when importing a file into an assay that are already claimed as an output of a run.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2466

#### Changes
* Clean up the duplicate exp.data identified with the "cross run input" role and the same dataFileUrl path
* When truncating a table, select non-null LSID values to delete ontology objects

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -3788,6 +3788,15 @@ public class ExperimentServiceImpl implements ExperimentService
                     {
                         for (ExpDataImpl output : datasToDelete)
                         {
+                            // Don't delete the exp.data output if it is being used in other runs
+                            List<? extends ExpRun> otherUsages = ExperimentService.get().getRunsUsingDatas(List.of(output));
+                            otherUsages.remove(run);
+                            if (!otherUsages.isEmpty())
+                            {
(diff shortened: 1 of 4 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -3780,6 +3781,21 @@ public void deleteExperimentRunsByRowIds(Container container, final User user, @
...
+                    List<ExpDataImpl> crossRun = new ArrayList<>();
+                    List<ExpDataImpl> inputData = run.getInputDatas(DefaultAssayRunCreator.CROSS_RUN_DATA_INPUT_ROLE, null);
+                    if (!inputData.isEmpty())
+                    {
+                        for (ExpDataImpl output : datasToDelete)
+                        {
+                            for (ExpDataImpl input : inputData)
+                            {
+                                if (input.getDataFileUrl().equals(output.getDataFileUrl()))
+                                    crossRun.add(output);
```
labkey-jeckels, line 3803: Should we check if this data has any other references? Deleting if orphaned seems good, but could it have been picked up by something else?

labkey-jeckels (approved):
One question added. If you think more discussion or review would be helpful, send a rereview request.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-kevink: Excellent point.  I'll check it hasn't been used elsewhere.

labkey-jeckels (approved):
Looks good. Thanks for the nice debug logging

PR #2489, opened 2021-07-23
Fix SMAliquotImportExportTest.testCreateRootAliquotAndSubAliquot

XingY:
#### Rationale
A bug is revealed for subaliquot import by test automation. Specifically, if a sample, its aliquots and subaliquots are imported at the same time, the subalqiuots might not have the correct root sample populated.

Sample import processes lineage in a batch. In the case when a sample, its aliquots, and subaliquots are imported/created together and processed in the same batch, the subaliquots's parent aliquot might not have AliquotedFromLSID set yet. This is resulting in the subaliquots to not correctly finding its root sample. We should not rely on AliquotedFromLSID solely to determine its aliquot's root sample during import.
                   
#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/637

#### Changes
* Check _aliquotRootCache for aliquot parents that are newly created

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -6452,7 +6452,11 @@ public class ExperimentServiceImpl implements ExperimentService
                 if (rec._action.getActionSequence() == SIMPLE_PROTOCOL_CORE_STEP_SEQUENCE)
                 {
                     ExpMaterial parent = rec._runRecord.getAliquotInput();
-                    boolean isParentRootMaterial = StringUtils.isEmpty(parent.getAliquotedFromLSID());
+
+                    // in the case when a sample, its aliquots, and subaliquots are imported/created together, the subaliquots's parent aliquot might not have AliquotedFromLSID yet.
+                    // Use cache to double check detemine subaliquots's root
+                    boolean isParentRootMaterial = StringUtils.isEmpty(parent.getAliquotedFromLSID()) && !_aliquotRootCache.containsKey(parent.getLSID());
+
                     for (ExpMaterial outputAliquot : rec._runRecord.getAliquotOutputs())
                     {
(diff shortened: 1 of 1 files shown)


labkey-danield (commented):
It looks ok to me but I will leave it to Susan to approve.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -6452,7 +6452,11 @@ private void saveExpMaterialAliquotOutputs(List<ProtocolAppRecord> protAppRecord
                 if (rec._action.getActionSequence() == SIMPLE_PROTOCOL_CORE_STEP_SEQUENCE)
                 {
                     ExpMaterial parent = rec._runRecord.getAliquotInput();
-                    boolean isParentRootMaterial = StringUtils.isEmpty(parent.getAliquotedFromLSID());
+
+                    // in the case when a sample, its aliquots, and subaliquots are imported/created together, the subaliquots's parent aliquot might not have AliquotedFromLSID yet.
+                    // Use cache to double check detemine subaliquots's root
+                    boolean isParentRootMaterial = StringUtils.isEmpty(parent.getAliquotedFromLSID()) && !_aliquotRootCache.containsKey(parent.getLSID());
```
labkey-susanh, line 6458: How do we know that the root is processed first and is in the cache?

XingY: The cache consist of 2 sets of data. First, it's initialized to query all Root for all "AliquotedFrom". Then, as each record gets processed, each aliquot and their root is put in the cache. If a subaliquot is created before aliquot, or if aliquot is imported before a root sample, the import will be rejected since the parent is invalid.

labkey-susanh: Go it.  Thanks for the explanation.

labkey-susanh (approved):

PR #2490, opened 2021-07-23
Issue 43508: Improve visibility of which user ID enabled which ETLs

labkey-jeckels:
#### Rationale
Disabling or deleting users may have important side effects. We want to preview these for admins

#### Changes
* Let UserListeners give feedback on what will happen if a user is deleted or deactivated
* Don't force UserListener implementations to implement every method

### DIFF
--- a/core/src/org/labkey/core/user/deactivateUsers.jsp
+++ b/core/src/org/labkey/core/user/deactivateUsers.jsp
@@ -20,6 +20,8 @@
 <%@ page import="org.labkey.api.view.HttpView" %>
 <%@ page import="org.labkey.api.view.JspView" %>
 <%@ page import="org.labkey.core.user.DeactivateUsersBean" %>
+<%@ page import="org.labkey.api.security.UserManager" %>
+<%@ page import="org.labkey.api.util.HtmlString" %>
 <%@ taglib prefix="labkey" uri="http://www.labkey.org/taglib" %>
 <%@ page extends="org.labkey.api.jsp.JspBase" %>
 <%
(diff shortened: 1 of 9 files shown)


core/src/org/labkey/core/user/deactivateUsers.jsp
```diff
@@ -35,8 +37,16 @@
     <%
         for (User user : bean.getUsers())
         {
-            %><li><%=h(user.getDisplayName(currentUser))%></li><%
-        } 
+            %><li><%=h(user.getDisplayName(currentUser))%><%
+            if (!bean.isActivate()) { %>
+                <ul>
+                    <% for (HtmlString message : UserManager.previewUserAccountDeactivated(user)) { %>
+                    <li><%= message %></li>
```
labkey-martyp, line 44: Does this need to be HTML encoded?

core/src/org/labkey/core/user/deleteUsers.jsp
```diff
@@ -36,10 +38,16 @@
...
-        for (User user : bean.getUsers())
-        {
-            %><li><%=h(user.getDisplayName(currentUser))%></li><%
+        <%
+            for (User user : bean.getUsers())
+            {
+        %><li><%=h(user.getDisplayName(currentUser))%>
+        <ul>
+            <% for (HtmlString message : UserManager.previewUserAccountDeleted(user)) { %>
+            <li><%= message %></li>
```
labkey-martyp, line 47: HTML encoded?

core/src/org/labkey/core/user/deactivateUsers.jsp
labkey-jeckels: Thanks for asking!

These are HtmlString, which handle the encoding work. Because I wanted to have links to the containers with ETLs in the messages, I used HtmlString to make it clear who is responsible for the encoding.

labkey-martyp (approved):
Looks like a nice feature and cleanup

core/src/org/labkey/core/user/deactivateUsers.jsp
labkey-martyp: Got it. I suspected that but thought I would ask

PR #2493, opened 2021-07-26
43558: Exported Dataset File cannot be reimported via Filewatcher

labkey-klum:
#### Rationale
In the scenario where we create a new dataset from a file watcher, the logic for detecting the built-in columns is fairly strict and tries to match the names exactly. This is a little different from creating a new dataset through the UI where we provide a column mapping UI to map subject ID and timepoint fields from the source to the target.

While this change doesn't add that full flexibility (due to the lack of a UI), we can do a slightly better job of matching the built in columns by stripping off any spaces in the inferred column names.

More details here:
- https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43558


labkey-klum:
@cnathe since you are familiar with the `DatasetColumnMappingPanel.tsx`, assigning the code review to you just to make sure we aren't inconsistent in the way we try to match the preferred subject ID column.

study/src/org/labkey/study/pipeline/DatasetInferSchemaReader.java
```diff
@@ -145,7 +145,8 @@ private void initialize()
                     for (ColumnDescriptor col : columns)
                     {
                         // filter out the built-in types
-                        if (DatasetDefinition.isDefaultFieldName(col.getColumnName(), _study))
+                        // issue 43558 : remove spaces from the column names (they may be exported column captions)
+                        if (DatasetDefinition.isDefaultFieldName(col.getColumnName().replace(" ", ""), _study))
```
cnathe, line 152: looking in the body of `isDefaultFieldName` it looks like it uses `study.getSubjectColumnName()` which is user defined right? In that case, is it possible that the subject column name could be configured with spaces in it?
In that case, should this DatasetInferSchemaReader check both with and without spaces?

cnathe (changes requested):

study/src/org/labkey/study/pipeline/DatasetInferSchemaReader.java
cnathe: As you mentioned, we do this similar type of replacement in the DatasetColumnMappingPanel.tsx, so if we test this with the subject column name defined with a space in the name and it doesn't work with this change, then we should likely get the same check updated in that React component case. I can take a look at that part if you'd like.

labkey-klum: Good point Cory, I can modify the code so that we do the check twice with and without spaces removed. And yes, if you  wouldn't mind taking a look at the React component to make sure we are consistent in our behavior.

cnathe: Ok. Upon closer review and local testing, it looks like we do allow the study property for Subject Column Name to include spaces, but we then strip those when creating the column in the study.Participant table. It looks like the code in the DatasetColumnMappingPanel.tsx is accounting for the matching based on both with and without spaces.

cnathe (approved):

PR #2496, opened 2021-07-27
Accommodate UserReg-WS JSP file packaged at /com/hphc/...

labkey-adam:
#### Rationale
Teach JspFinder about JSP files packaged at /com/hphc


labkey-susanh (approved):
Seems fine on the surface, but are we effectively not filtering out any jsps?  

(BTW, I'm pretty sure that `shouldInclude` method can now be removed since that funny JSP in Biologics no longer exists.)

labkey-adam:
> Seems fine on the surface, but are we effectively not filtering out any jsps?
> 
> (BTW, I'm pretty sure that `shouldInclude` method can now be removed since that funny JSP in Biologics no longer exists.)

This ensures that JSPs live in /org/ or /com/. This is filtering out nothing at this point, but future JSPs could appear in strange places. Honestly, I don't remember what problem this was trying to solve... we could just take every file path and either there's a matching reference or not.

I removed `shouldInclude()`, etc.

PR #2499, opened 2021-07-28
Issue 43648: Study dataset reorder page jumps to top when there are enough datasets to make the page include a scrollbar

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43648

Client reported that with a 21.3.4 change it got much more difficult to reorder study datasets when there are a large number of datasets in the study (i.e. in their example, see support ticket details, they have >200 datasets). In addition to fixing the "page jump" issue, this PR adds in new "move to top" and "move to bottom" buttons to help with the client scenario.

#### Changes
* update datasetDisplayOrder.jsp to use insertBefore() and appendChilde() to move select option items 
* remove "submit" from move buttons to prevent page jump on click
* add "move to top" and "move to bottom" buttons
* similar button styling updates for Manage Visit Order and Project Display Order pages

### DIFF
--- a/study/src/org/labkey/study/view/datasetDisplayOrder.jsp
+++ b/study/src/org/labkey/study/view/datasetDisplayOrder.jsp
@@ -23,12 +23,10 @@
 <%@ page extends="org.labkey.study.view.BaseStudyPage" %>
 <%@ taglib prefix="labkey" uri="http://www.labkey.org/taglib" %>
 <script>
-function saveList(listName)
-{
+function saveList() {
     var itemList = "";
     var itemSelect = document.reorder.items;
-    for (var i = 0; i < itemSelect.length; i++)
-    {
+    for (var i = 0; i < itemSelect.length; i++) {
         itemList += itemSelect.item(i).value;
         if (i < itemSelect.length - 1)
             itemList += ",";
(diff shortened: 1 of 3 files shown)


study/src/org/labkey/study/view/datasetDisplayOrder.jsp
```diff
@@ -115,9 +100,11 @@ function orderModule(down)
                 %>
                 </select>
             </td>
-            <td align="center" valign="center">
-                <%= button("Move Up").submit(true).onClick("return orderModule(0)") %><br><br>
-                <%= button("Move Down").submit(true).onClick("return orderModule(1)") %>
+            <td align="center" valign="center" style="padding-left: 10px;">
+                <%= button("Move Up").onClick("return moveDatasetItem('up')") %><br>
+                <%= button("Move Down").onClick("return moveDatasetItem('down')") %><br><br>
+                <%= button("Move to Top").onClick("return moveDatasetItem('top')") %><br>
```
labkey-jeckels: Should we make this the first button?

labkey-jeckels (approved):

study/src/org/labkey/study/view/datasetDisplayOrder.jsp
cnathe: I went back and forth on that. Here is a screenshot of the current implementation. I decided to group the buttons for move up/move down and then move to top / move to bottom. Happy to change it as you are suggesting if you think that is more intuitive.

Screen Shot 2021-07-28 at 10 11 10 AM

cnathe: Updated display based on chat conversation with Josh:

Screen Shot 2021-07-28 at 1 17 21 PM

labkey-jeckels: Breathtaking!

PR #2506, opened 2021-08-02
Dataset change Email Notification Automation test

labkey-sweta:
#### Rationale
Dataset change email notifications automation test

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/824

#### Changes
study/test/src/org/labkey/test/tests/study/StudyDatasetChangeEmailNotificationTest.java

### DIFF
--- a/study/test/src/org/labkey/test/tests/study/StudyDatasetChangeEmailNotificationTest.java
+++ b/study/test/src/org/labkey/test/tests/study/StudyDatasetChangeEmailNotificationTest.java
@@ -1,6 +1,7 @@
 package org.labkey.test.tests.study;
 
 import org.jetbrains.annotations.Nullable;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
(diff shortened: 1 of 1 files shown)


study/test/src/org/labkey/test/tests/study/StudyDatasetChangeEmailNotificationTest.java
```diff
@@ -0,0 +1,200 @@
...
+        setFormElement(Locator.name("quf_SequenceNum"), "1");
+        clickButton("Submit");
+
+        log("Execute the script to send the email");
+        executeScript("LABKEY.Ajax.request({ url: '/labkey/home/reports-sendDailyDigest.view', method: 'POST' });");
+
+        goToModule("Dumbster");
+        EmailRecordTable notifications = new EmailRecordTable(this);
+        while (notifications.getEmailCount() < emailCount)
+            refresh();
```
labkey-klum, line 92: you may want to limit this to a certain number of refreshes to avoid an infinite loop on an error.

labkey-klum (changes requested):
See comments about capping the number of times you refresh the mail record page. Another approach would be to hit the daily digest API at the end of `doCreateSteps` which should reset the notifications so you don't have to account for that initial email from creation, you could also clear out the mail records at that start of each test by unchecking the `Record email messages sent` checkbox on the webpart, that way you are only looking for a single email in each test.

labkey-sweta:
> Record email messages sent

Updated the test.

labkey-klum (approved):
if we fail to send an email then the code at line : 179 will just loop continuously, not great but I guess the test will time out at the 3 minute mark?

labkey-sweta:
> if we fail to send an email then the code at line : 179 will just loop continuously, not great but I guess the test will time out at the 3 minute mark?

Yes, test will timeout.

PR #2507, opened 2021-08-02
Fix StudyMergeParticipantsTest

labkey-danield:
#### Rationale
Fix StudyMergeParticipantsTest by updating call to deprecated clickButton function.

#### Related Pull Requests
* None

#### Changes
* Removed call to deprecated clickButton function and updated it to use the appropriate/current clickButton method.


labkey-sweta (approved):
Shouldn't we wait for other element on that page? Just page load is enough?

PR #2509, opened 2021-08-03
Page & action to (re-)invoke upgrade code methods

labkey-adam:
#### Rationale
[Issue 43665: Provide action to invoke arbitrary upgrade code methods](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43665)

image

### DIFF
--- a/core/src/org/labkey/core/admin/sql/SqlScriptController.java
+++ b/core/src/org/labkey/core/admin/sql/SqlScriptController.java
@@ -215,10 +215,10 @@ public class SqlScriptController extends SpringActionController
 
             StringBuilder html = new StringBuilder("<table>");
 
+            html.append("<tr><td colspan=4>");
+
             if (AppProps.getInstance().isDevMode())
             {
-                html.append("<tr><td colspan=4>");
-
                 if (form.isManagedOnly())
                     html.append(PageFlowUtil.link("show all modules").href(new ActionURL(ScriptsAction.class, ContainerManager.getRoot())));
                 else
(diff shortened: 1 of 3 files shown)


labkey-klum (approved):

core/src/org/labkey/core/admin/sql/SqlScriptController.java
```diff
@@ -1300,6 +1308,111 @@ public void addNavTrail(NavTree root)
...
+            return _module;
+        }
+
+        public String getMethod()
+        {
+            return _method;
+        }
+    }
+
+    @RequiresPermission(AdminOperationsPermission.class)
```
labkey-jeckels, line 1344: Should we make this restricted to Site Admin instead? I can't imagine wanting to have hosted customers invoke this on their own.

labkey-jeckels (approved):

labkey-susanh (approved):

core/src/org/labkey/core/admin/sql/SqlScriptController.java
labkey-adam: Sure... I had it site admin at first, but noticed that Cory had annotated all the other script actions with AdminOperations. (Though most of the links aren't shown in production mode.)

labkey-adam: Wait... isn't `AdminOperationsPermission` available to site administrators only?

labkey-jeckels: Hmm... You're right. I guess don't understand the difference between SiteAdminPermission and AdminOperationsPermission, but it looks like both are granted only to Site Admins

labkey-adam: Yeah, me neither

PR #2521, opened 2021-08-05
Delete cross run input exp.data

labkey-kevink:
#### Rationale
As a part of the recently added support for importing a exp.data into an assay that is already marked as being produced by another run (#2445), we will delete any output exp.data if it matches the input exp.data's dataFileURL and if it isn't used elsewhere. The intent is to avoid creating duplicate exp.data for the same file path that end up being orphaned.  This worked well when deleting assay runs from newest to oldest.  See #2488.

This PR cleans up any input exp.data that would be orphaned when deleting assay runs from oldest to newest.  We won't delete the input exp.data if it is used in another run or if it has file content domain properties.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2488

#### Changes
* Delete cross run input exp.data that are not being used and have no file-content properties.


labkey-jeckels (changes requested):
This diff looks like a patch got applied incorrectly and the build is failing on TeamCity. Can you double-check?

PR #2522, opened 2021-08-05
Delete unused cross run input without file-content properties

labkey-kevink:
#### Rationale
As a part of the recently added support for importing a exp.data into an assay that is already marked as being produced by another run (#2445), we will delete any output exp.data if it matches the input exp.data's dataFileURL and if it isn't used elsewhere. The intent is to avoid creating duplicate exp.data for the same file path that end up being orphaned.  This worked well when deleting assay runs from newest to oldest.  See #2488.

This PR cleans up any input exp.data that would be orphaned when deleting assay runs from oldest to newest.  We won't delete the input exp.data if it is used in another run or if it has file content domain properties.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2488

#### Changes
* Delete cross run input exp.data that are not being used and have no file-content properties.

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -76,6 +76,7 @@ import org.labkey.api.exp.query.ExpSchema;
 import org.labkey.api.exp.query.SamplesSchema;
 import org.labkey.api.exp.xar.LsidUtils;
 import org.labkey.api.exp.xar.XarConstants;
+import org.labkey.api.files.FileContentService;
 import org.labkey.api.gwt.client.model.GWTDomain;
 import org.labkey.api.gwt.client.model.GWTIndex;
 import org.labkey.api.gwt.client.model.GWTPropertyDescriptor;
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -3781,26 +3782,49 @@ public void deleteExperimentRunsByRowIds(Container container, final User user, @
...
+                                        break;
+                                    }
+
+                                    crossRunOutputs.add(output);
+                                }
+                            }
+
+                            // If there are comments or file properties attached to the exp.data don't delete it.
+                            // TODO: We don't want to delete exp.data for uploaded files since we track created/createdby and other metadata... is there a better way to identify these exp.data we don't want to delete?
+                            // CONSIDER: move these properties to the matching output exp.data with the same dataFileUrl?
```
labkey-jeckels, line 3815: Yes, I think merging the properties is the right way to go long-term. We could retain the oldest Created/CreatedBy pair and the newest Modified/ModifiedBy pair. Other properties are harder to merge if there are different values on the different data entries

labkey-jeckels (approved):

PR #2523, opened 2021-08-05
Folder export updates for better handling of SM Workflow jobs and preventing duplicate audit entries

labkey-susanh:
#### Rationale
Testing of folder import and export for Sample Manager folders revealed a few things that don't round trip well.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/652

#### Changes
* Filter the ExperimentRuns included in the sample and dataset xar to only derivation runs. Other runs with samples as inputs will be included in the experiment xar.
* Add special handling for the AssayTypes property attached to workflow job tasks so they work well for folder import and export
* Make the DATE property type recognized for export
* Don't export picklists 
* Check for presence of the StudyService before getting datasets


experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
```diff
@@ -170,6 +170,8 @@ public void write(Container object, ImportContext<FolderDocument.Folder> ctx, Vi
 
         if (!datasToExport.isEmpty())
             exportedRuns.addAll(ExperimentService.get().getRunsUsingDatas(datasToExport));
+        // only want the sample derivation runs; other runs will get included in the experiment xar.
+        exportedRuns = exportedRuns.stream().filter(run -> run.getProtocol().getLSID().equals(ExperimentService.SAMPLE_DERIVATION_PROTOCOL_LSID)).collect(Collectors.toSet());
```
XingY: what about SAMPLE_ALIQUOT_PROTOCOL_LSID?

list/src/org/labkey/list/model/ListWriter.java
```diff
@@ -89,6 +89,12 @@
     public boolean write(Container c, User user, VirtualFile listsDir, ImportContext ctx) throws Exception
     {
         Map<String, ListDefinition> lists = ListService.get().getLists(c, user, true);
+        // We exclude picklists because they contain just sampleIds, which will be different in the new container.
+        // Picklists are meant to be transient, so not including them in the export makes sense.
```
XingY, line 92: Should we add a new ListService.getLists with boolean param to exclude picklist instead?

experiment/src/org/labkey/experiment/XarExporter.java
```diff
@@ -1270,6 +1272,21 @@ private PropertyCollectionType getProperties(Map<String, ObjectProperty> propert
...
+                        // This property stores rowIds of assay designs; we need to translate them to LSIDs for export
+                        // TODO perhaps this property should hold protocol strings instead of rowIds
+                        else if (value.getPropertyURI().endsWith(":WorkflowTask#AssayTypes"))
+                        {
+                            String[] assayIds = value.getStringValue().split(",");
+                            List<String> protocolStrings = new ArrayList<>();
+                            List<ExpProtocol> protocols = AssayService.get().getAssayProtocols(value.getContainer());
+                            for (String assayId : assayIds)
+                            {
+                                int assayRowId = Integer.parseInt(assayId);
```
XingY: Perhaps only parse if it's numeric, to allow the TODO scenario mentioned?

experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
labkey-susanh: Good catch!  Thanks!

list/src/org/labkey/list/model/ListWriter.java
labkey-susanh: Did that, but then it became a slightly bigger refactor because of my poor choice of parameter names earlier.

experiment/src/org/labkey/experiment/XarExporter.java
labkey-susanh: Good thought.

XingY (approved):

PR #2524, opened 2021-08-05
Issue 43647: SM: creating aliquots for a sample type with a required field gives an error

XingY:
#### Rationale
RequiredValidator fails import/creation for aliquots when the sample type has require fields. This PR modifies the import so that aliquots can bypass the required field validator

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/597
* https://github.com/LabKey/platform/pull/2524
* https://github.com/LabKey/sampleManagement/pull/653

#### Changes
*  skip RequiredValidation for sample fields that are required, if the sample is an aliquot

### DIFF
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -218,7 +218,19 @@ public class ExpDataIterators
             if (!(v instanceof RequiredValidator) || _aliquotedFromColInd < 0)
                 return super.validate(v, rowNum, value, data);
 
-            String aliquotedFromValue = (String) data.get(_aliquotedFromColInd);
+            String aliquotedFromValue = null;
+            Object aliquotedFromObj = data.get(_aliquotedFromColInd);
+            if (aliquotedFromObj != null)
+            {
+                if (aliquotedFromObj instanceof String)
+                {
+                    aliquotedFromValue = (String) aliquotedFromObj;
+                }
+                else if (aliquotedFromObj instanceof Number)
+                {
+                    aliquotedFromValue = aliquotedFromObj.toString();
+                }
+            }
 
             // skip required field check for aliquots since aliquots properties are inherited
             if (!StringUtils.isEmpty(aliquotedFromValue))
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -16,6 +16,7 @@
 package org.labkey.experiment;
 
 import org.apache.commons.beanutils.ConvertUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONArray;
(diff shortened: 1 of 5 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -189,6 +194,67 @@ public DataIterator getDataIterator(DataIteratorContext context)
...
+            {
+                if (aliquotedFromObj instanceof String)
+                {
+                    aliquotedFromValue = (String) aliquotedFromObj;
+                }
+                else if (aliquotedFromObj instanceof Number)
+                {
+                    aliquotedFromValue = aliquotedFromObj.toString();
+                }
+            }
```
labkey-susanh, line 233: Do we need both string and number to account for "123" and 123 as row Ids?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -158,7 +161,13 @@ public int importRows(User user, Container container, DataIteratorBuilder rows,
     public DataIteratorBuilder createImportDIB(User user, Container container, DataIteratorBuilder data, DataIteratorContext context)
     {
         assert _sampleType != null : "SampleType required for insert/update, but not required for read/delete";
-        DataIteratorBuilder dib = super.createImportDIB(user, container, data, context);
+
+        DataIteratorBuilder dib = new ExpDataIterators.ExpMaterialDataIteratorBuilder(getQueryTable(), data, container, user);
```
labkey-susanh, line 165: I think you also need this block from `super.createImportDIB`:
```
        if (_enableExistingRecordsDataIterator)
        {
            // some tables need to generate PK's, so they need to add ExistingRecordDataIterator in persistRows() (after generating PK, before inserting)
            dib = ExistingRecordDataIterator.createBuilder(dib, getQueryTable(), getSelectKeys(context));
        }
```

labkey-susanh (changes requested):

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: No, we support "123" as SampleID, aliquotedFrom is always sampleId, not rowId

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: I don't think so, SampleTypeUpdateServiceID is the subclass that specifically disable this, see
`// we do this in ExpMaterialTableImpl.persistRows() via ExpDataIterators.PersistDataIteratorBuilder
        _enableExistingRecordsDataIterator = false;`

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Got it.  Thanks.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
labkey-susanh: Ah ha.

labkey-susanh (approved):

PR #2526, opened 2021-08-06
Issue43480: Sample timeline shows sample link to study before registration

RosalineP:
#### Rationale
Currently, SM Timeline shows the "Sample was linked to a study" event before the "Sample was registered" event. Rearranging the order of the data iterators resolves this problem.

#### Related Pull Requests
* n/a

#### Changes
* move dataiterator

### DIFF
--- a/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
+++ b/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
@@ -159,23 +159,19 @@ public class SampleTypeUpdateServiceDI extends DefaultQueryUpdateService
     {
         assert _sampleType != null : "SampleType required for insert/update, but not required for read/delete";
         DataIteratorBuilder dib = super.createImportDIB(user, container, data, context);
-        LOG.warn("Temp: inside of createImportDIB");
-        if (InventoryService.get() != null)
+        UserSchema userSchema = getQueryTable().getUserSchema();
+        if (InventoryService.get() != null && userSchema != null)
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -159,16 +159,17 @@ public DataIteratorBuilder createImportDIB(User user, Container container, DataI
...
-        if (InventoryService.get() != null)
+        UserSchema userSchema = getQueryTable().getUserSchema();
+        if (InventoryService.get() != null && userSchema != null)
         {
             ExpSampleType sampleType = ((ExpMaterialTableImpl) getQueryTable()).getSampleType();
-            UserSchema userSchema = getQueryTable().getUserSchema();
-            if (userSchema != null)
-                return LoggingDataIterator.wrap(InventoryService.get().getPersistStorageItemDataIteratorBuilder(dib, userSchema.getContainer(), userSchema.getUser(), sampleType.getMetricUnit()));
-            return dib;
+            dib = LoggingDataIterator.wrap(InventoryService.get().getPersistStorageItemDataIteratorBuilder(dib, userSchema.getContainer(), userSchema.getUser(), sampleType.getMetricUnit()));
```
labkey-susanh, line 166: I think you'll still want the check for `userScheme != null` before this line.

labkey-susanh (approved):
Thanks for doing this.  One NPE protection check suggested, but no need for re-review.

PR #825, opened 2020-02-10
Update Module

labkey-ivyc:
(Created for Matt because he is OOO)

PR #2528, opened 2021-08-06
Delete unused JSP and its commented reference

labkey-jeckels:
#### Rationale
This JSP has been orphaned for 17+ months. No need to keep it around and cluttering the report of unused JSPs

#### Changes
* Remove unused JSP and now no-op overridden method that once referenced it


labkey-tchad:
@labkey-matthewb Near the end of the PR when that test was introduced (#825), there was a commit disabling it "temporarily": https://github.com/LabKey/platform/pull/825/commits/5f988307ebd20c649ea9c5a5e1c4e50a15ec6c82

labkey-matthewb (commented):
Not sure why this is was commented out.  Is SimpleModule.canBeEnabled() tested elsewhere?

labkey-tchad:
It doesn't look like it is being tested anywhere else.

labkey-tchad:
Let's see what happens if we enable it: #2531

labkey-tchad:
The test found an issue, so we should probably fix that and enable the test. Should probably fix the bug and close this PR.

labkey-tchad:
Test is needed. Updating and enabling in #2531

PR #2530, opened 2021-08-09
25741: Support for trigger scripts on UI bulk insert

labkey-klum:
#### Rationale
- https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=25741

Also represents an opportunity to clean up some of the variety of legacy code paths which insert data into datasets so that they are funneled through the QUS in order to get more of a uniform behavior.
 
#### Changes
- User `BatchValidationException` instead of `List<String>` for errors
- Refactor `DatasetDefinition.importDatasetData` to go through the `QueryUpdateService` to load data.
- Consolidate code in `StudyManager.importDatasetData` to go through common code paths and push more options into `DataIteratorContext`


study/src/org/labkey/study/assay/StudyPublishManager.java
```diff
@@ -423,7 +428,17 @@ else if (!datasetPublishSourceId.equals(publishSource.second) || dataset.getPubl
             if (defaultQCStateId != null)
                 defaultQCState = QCStateManager.getInstance().getQCStateForRowId(targetContainer, defaultQCStateId.intValue());
 
-            datasetLsids = StudyManager.getInstance().importDatasetData(user, dataset, convertedDataMaps, errors, DatasetDefinition.CheckForDuplicates.sourceAndDestination, defaultQCState, null, false);
+            BatchValidationException validationException = new BatchValidationException();
+            if (!targetContainer.hasPermission(user, AdminPermission.class) && targetContainer.hasPermission(user, InsertPermission.class))
```
labkey-matthewb, line 462: targetContainer.hasPermissions(Set.of(...))

study/src/org/labkey/study/assay/StudyPublishManager.java
```diff
@@ -423,7 +428,17 @@ else if (!datasetPublishSourceId.equals(publishSource.second) || dataset.getPubl
...
+            {
+                // we allow linking data to a study even if the study security is set to read-only datasets, since the
+                // underlying insert uses the QUS, we pass in a contextual role to allow the insert to succeed
+                Set<Role> contextualRoles = new HashSet<>(user.getStandardContextualRoles());
```
labkey-matthewb, line 466: Set.of(user.getStandardContextualRoles(),RoleManager.getRole(FolderAdminRole.class)) is shorter

study/src/org/labkey/study/model/StudyManager.java
```diff
@@ -3452,63 +3452,78 @@ public void batchValidateExceptionToList(BatchValidationException errors, List<S
...
-        parseData(user, def, loader, columnMap);
         DataIteratorContext context = new DataIteratorContext(errors);
+
         context.setInsertOption(insertOption);
         context.setAllowImportLookupByAlternateKey(importLookupByAlternateKey);
+
         Map<Enum, Object> options = new HashMap<>();
         options.put(DetailedAuditLogDataIterator.AuditConfigs.AuditBehavior, auditBehaviorType);
-
+        options.put(DatasetUpdateService.Config.ForUpdate, Boolean.valueOf(false));
```
labkey-matthewb: why not Boolean.FALSE?  (several usages in patch)

study/src/org/labkey/study/query/DatasetUpdateService.java
```diff
@@ -95,7 +95,10 @@
         // NOTE: There really has to be better way to handle the functionality of StudyImportContext.getTableIdMap()
         // NOTE: Could this be handled by a method on StudySchema or something???
         // see StudyImportContext.getTableIdMapMap()
-        StudyImportMaps       // expected: Map<String,Map<Object,Object>>
+        StudyImportMaps,        // expected: Map<String,Map<Object,Object>>
+
+        KeyList,                // expected: List<String>
+        ForUpdate               // expected: Boolean
```
labkey-matthewb: Should we add an new value to the InsertOption enum so we don't need a similar looking ForUpdate?

labkey-matthewb (changes requested):
I'm really happy we're continuing the conversion to QUS!

study/src/org/labkey/study/assay/StudyPublishManager.java
labkey-klum: can't really use that since the check is if you can insert but aren't already an admin.

study/src/org/labkey/study/assay/StudyPublishManager.java
labkey-klum: Hmmm, this didn't seem to work since the first param is `Set<Role>` and the second is `Role`, it ends up returning `Set<Object>`

study/src/org/labkey/study/model/StudyManager.java
labkey-klum: sounds good, done

study/src/org/labkey/study/query/DatasetUpdateService.java
labkey-klum: I generally like that idea although looking closer at when we use that flag doesn't map very cleanly to any of the existing `InsertOption` enums. We set it to both true or false for either INSERT or IMPORT and maybe it's wrong in it's current state. 

Looking at `DatasetDataIteratorBuilder:145` it appears to be used to control whether we allow a managed key to be imported. Maybe just renaming this Enum to make it clearer is the better route since it doesn't appear to be related to an insert option.

labkey-matthewb (approved):

PR #2539, opened 2021-08-11
43716: Sample types with name expression values don't round trip via folder archive as expected for SM app

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43716

This seems to be a long standing bug with XAR export/import but we weren't properly setting up the idcol1 field in materialSource when a name expression was provided. 

#### Changes
This change aligns the XAR export/import code with the code which creates a sample type in the `SampleTypeService` with respect to name expressions and the idCol1 field. This ensures that a name expression will work only if idCol1 is set to "name".

### DIFF
--- a/experiment/src/org/labkey/experiment/XarExporter.java
+++ b/experiment/src/org/labkey/experiment/XarExporter.java
@@ -605,7 +605,8 @@ public class XarExporter
         {
             xSampleSet.setNameExpression(sampleType.getNameExpression());
         }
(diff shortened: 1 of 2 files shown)


cnathe (approved):

cnathe:
@labkey-klum FYI, I create a PR in SM to add this `<KeyField>` prop to the SM trial eval folder archive:
https://github.com/LabKey/sampleManagement/pull/657

PR #2541, opened 2021-08-12
Enable import for Folder/Study archives when paired with the cloud module

Sigmonia:
#### Rationale
Item #9160 - [Labkey R&D-EI - Reload Study from S3 root via Folder Management UI](https://docs.google.com/document/d/1Fe9tyqIdvMTyu7gFYm4PJrfFMXMf1oQbpVXXdvRzqn8/)

#### Related Pull Requests
* https://github.com/LabKey/cloud/pull/87
* https://github.com/LabKey/dataintegration/pull/122
* https://github.com/LabKey/testAutomation/pull/847
* https://github.com/LabKey/commonAssays/pull/360
* https://github.com/LabKey/targetedms/pull/423
* https://github.com/LabKey/immport/pull/42
* https://github.com/LabKey/moduleEditor/pull/45

#### Changes
* Refactors to replace some usages of File with nio.Path
  * Deprecated several API methods in preference for Path based versions
  * Fixed some typos & NPEs along the way
* Added cloud service task to download study archive files
  * Adjusted context and root Job properties to account for cloud --> temp directory path change during processing


api/src/org/labkey/api/admin/FolderImportContext.java
```diff
@@ -83,44 +104,40 @@ public synchronized FolderDocument getDocument() throws ImportException, Invalid
...
     }
 
-    // Assume file was referenced in folder.xml file   // TODO: Context should hold onto the root -- shouldn't have to pass it in
-    public File getFolderFile(File root, File dir, String name) throws ImportException
-    {
-        return getFolderFile(root, dir, name, _folderXml.getName());
-    }
-
-    public File getFolderFile(File root, File dir, String name, String source) throws ImportException
+//    // Assume file was referenced in folder.xml file   // TODO: Context should hold onto the root -- shouldn't have to pass it in
```
labkey-klum: is there a reason to keep this code around?

api/src/org/labkey/api/pipeline/PathSerialization.java
```diff
@@ -71,14 +74,24 @@ protected Deserializer()
         public Path deserialize(JsonParser parser, DeserializationContext context) throws IOException
```
labkey-klum: add a `Nullable` annotation?

api/src/org/labkey/api/pipeline/PathSerialization.java
```diff
@@ -71,14 +74,24 @@ protected Deserializer()
...
+//                return Path.of(str);
+                CloudStoreService css = CloudStoreService.get();
+                if (css != null)
+                {
+                    //TODO this will likely work only for cloud paths that include access ids eg: s3://<pub-access-key>@s3.amazonaws.com/<my bucket>/...
+                    return css.getPathFromUrl(str);
+                    //TODO need LKS container to pull the appropriate S3 credentials/FileSystem representation (in the event of interrupted job)
+                }
+
+                return Path.of(str); //TODO this will cause issues...
```
labkey-klum: would it be better to throw then?

api/src/org/labkey/api/pipeline/PipelineProvider.java
```diff
@@ -249,7 +249,7 @@ public Module getOwningModule()
 
     public boolean supportsCloud()
     {
-        return false;
+        return true;
```
labkey-klum: Are we really saying that all providers now support cloud storage? Wouldn't it be safer to keep this unchanged and just override.

api/src/org/labkey/api/study/StudyService.java
```diff
@@ -179,7 +180,13 @@ static StudyService get()
 
     Set<? extends Study> getAllStudies(Container root);
 
-    boolean runStudyImportJob(Container c, User user, @Nullable ActionURL url, File studyXml, String originalFilename, BindException errors, PipeRoot pipelineRoot, ImportOptions options);
+    @Deprecated //prefer the Path version
+    default boolean runStudyImportJob(Container c, User user, @Nullable ActionURL url, File studyXml, String originalFilename, BindException errors, PipeRoot pipelineRoot, ImportOptions options)
```
labkey-klum: Is there any code that references this method?

api/src/org/labkey/api/util/FileType.java
```diff
@@ -605,16 +638,35 @@ public String toString()
         return (_dir == null || !_dir.booleanValue() ? _suffixes.toString() : _suffixes + "/");
     }
 
+//    @NotNull
+//    public static List<FileType> findTypes(@NotNull List<FileType> types, @NotNull List<File> files)
```
labkey-klum: delete?

api/src/org/labkey/api/admin/InvalidFileException.java
```diff
@@ -22,14 +22,21 @@
 import org.labkey.api.writer.VirtualFile;
 
 import java.io.File;
+import java.nio.file.Path;
 
 public class InvalidFileException extends ImportException
 {
+    @Deprecated // prefer the Path version
```
labkey-susanh, line 29: Would we want to have this forward to the Path version?
```
this(root.toPath(), file.toPath(), t);
```

study/src/org/labkey/study/pipeline/StudyReloadSourceJob.java
```diff
@@ -141,4 +152,46 @@ public boolean isMerge()
     {
         return false;
     }
+
+    @Override
+    public void downloadCloudArchive(Path studyXml, BindException errors)
```
labkey-klum: This method is nearly identical to the version in `StudyImportJob`, we should try to consolidate otherwise it will probably bite us down the road.

api/src/org/labkey/api/pipeline/PathSerialization.java
```diff
@@ -71,14 +74,24 @@ protected Deserializer()
         public Path deserialize(JsonParser parser, DeserializationContext context) throws IOException
         {
             String str = parser.getValueAsString();
+//            return Path.of(URI.create(str));
```
labkey-susanh: remove commented code?

api/src/org/labkey/api/pipeline/PathSerialization.java
```diff
@@ -71,14 +74,24 @@ protected Deserializer()
...
+//                // TODO: problem is that we need a container to map a URL string to an S3 path, because we have a prefix derived form the container (#35865)
+//                // TODO: one possibility is to tease out what config/container matches the bucket/prefix we find, but there could be more than 1 match
+//                return Path.of(str);
+                CloudStoreService css = CloudStoreService.get();
+                if (css != null)
+                {
+                    //TODO this will likely work only for cloud paths that include access ids eg: s3://<pub-access-key>@s3.amazonaws.com/<my bucket>/...
+                    return css.getPathFromUrl(str);
+                    //TODO need LKS container to pull the appropriate S3 credentials/FileSystem representation (in the event of interrupted job)
+                }
```
labkey-susanh, line 85: All of these TODOs make me wonder if this does work.  If it works in limited cases, do we need or have some corresponding documentation?

api/src/org/labkey/api/pipeline/PipeRoot.java
```diff
@@ -69,6 +69,20 @@
     @Nullable
     File resolvePath(String relativePath);
 
+    //TODO implement this properly
```
labkey-susanh: Properly?  Do we know what that means here?  Something to do now?

api/src/org/labkey/api/pipeline/PipelineJob.java
```diff
@@ -123,6 +123,11 @@ public void clearActionSet(ExpRun run)
         _actionSet = new RecordedActionSet();
     }
 
+    public void updateWorkingRoot(Path importRoot) throws PipelineJobException
+    {
+        throw new PipelineJobException("Unable to update working root. Updates not supported by job");
```
labkey-susanh:
```suggestion
        throw new PipelineJobException("Unable to update working root. Updates not supported by job.");

```

study/src/org/labkey/study/importer/StudyImportInitialTask.java
```diff
@@ -62,6 +65,13 @@ public RecordedActionSet run() throws PipelineJobException
         StudyJobSupport support = job.getJobSupport(StudyJobSupport.class);
         StudyImportContext ctx = support.getImportContext();
 
+        boolean isCloudRoot = job.getPipeRoot().isCloudRoot();
+        if (isCloudRoot)
```
labkey-klum: `StudyFolderImporter` also checks whether the study dir has a cloud scheme. Should we do the same here?

api/src/org/labkey/api/pipeline/PipelineJob.java
```diff
@@ -1086,7 +1091,7 @@ public void run()
     // Should be called in run()'s finally by any class that overrides run(), if class uses LocalDirectory
     protected void finallyCleanUpLocalDirectory()
     {
-        if (null != _localDirectory)
+        if (null != _localDirectory & isDone())
```
labkey-susanh, line 1096: when is this "finallyCleanUp" called when the job is not done?  Is this protection against an error case?  Or a case when nothing happened in the first place?

api/src/org/labkey/api/pipeline/file/FileAnalysisJobSupport.java
```diff
@@ -76,6 +79,25 @@
      */
     File findInputFile(String name);
 
+    default Path findInputPath(String filepath)
+    {
+        //TODO This needs implementation in derived classes...
```
labkey-susanh: Is this to be done in this PR?

api/src/org/labkey/api/writer/FileSystemFile.java
```diff
@@ -55,31 +56,42 @@ private FileSystemFile()
...
+    public FileSystemFile(Path root)
+    {
+        try
+        {
+            ensureWriteableDirectory(root);
+            _root = root;
+        }
+        catch (IOException e)
+        {
+            //TODO do more...
```
labkey-susanh: More?  What did you have in mind?

api/src/org/labkey/api/writer/FileSystemFile.java
```diff
@@ -162,42 +173,63 @@ public XmlObject getXmlBean(String filename) throws IOException
...
+        {
+            return null == files ? Collections.emptyList() :
+                    files.filter(filter)
+                         .map(Path::getFileName)
+                         .map(Path::toString)
+                         .collect(Collectors.toList());
+        }
+        catch (IOException e)
+        {
+            //TODO should do more here...
```
labkey-susanh: Do more?  Is there some way to recover?

api/src/org/labkey/api/writer/FileSystemFile.java
```diff
@@ -162,42 +173,63 @@ public XmlObject getXmlBean(String filename) throws IOException
...
-        File file = new File(_root, makeLegalName(filename));
-        return file.delete();
+        Path file = _root.resolve(makeLegalName(filename));
+        try
+        {
+            return Files.deleteIfExists(file);
+        }
+        catch (IOException e)
+        {
+            //TODO do more here...
```
labkey-susanh: Same comment here.  Let's either remove this TODO or make the comment more meaningful or do whatever "more" is in this case.

api/src/org/labkey/api/writer/PrintWriters.java
```diff
@@ -51,7 +54,31 @@ public static PrintWriter getPrintWriter(OutputStream out)
...
     public static PrintWriter getPrintWriter(File file) throws FileNotFoundException
     {
-        return new StandardPrintWriter(file);
+        try
+        {
+            return getPrintWriter(file.toPath());
+        }
+        catch (FileNotFoundException e)
+        {
+            // TODO Retain old behavior
```
labkey-susanh: What is the old behavior that needs to be retained?  I'm guessing we don't want to change the behavior.

experiment/src/org/labkey/experiment/api/ExpDataTableImpl.java
```diff
@@ -937,7 +938,7 @@ public void testWebDavColumns() throws Exception
                 try (ExcelWriter excel = view.getExcelWriter(ExcelWriter.ExcelDocumentType.xlsx))
                 {
                     VirtualFile f = new MemoryVirtualFile();
-                    excel.write(f.getOutputStream("excel.xlsx"));
+                    excel.write(f.getOutputStream("excel.xlsx")); //TODO should this outputStream be closed?
```
labkey-susanh: Closing the stream seems a good idea unless that happens elsewhere.

pipeline/src/org/labkey/pipeline/PipelineDirectoryImpl.java
```diff
@@ -32,6 +32,7 @@
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
```
labkey-susanh: Needed?

pipeline/src/org/labkey/pipeline/api/PipeRootImpl.java
```diff
@@ -413,7 +420,9 @@ public File getImportDirectory()
         return new File(root, PipelineService.UNZIP_DIR);
     }
 
+    //TODO why do we have both of these? 1
```
labkey-susanh: Are both referenced?

pipeline/src/org/labkey/pipeline/api/PipelineManager.java
```diff
@@ -805,68 +814,100 @@ else if (!pipeRoot.isCloudRoot() && !pipeRoot.isUnderRoot(archiveFile))     // T
...
-            catch (IOException e)
+
+            if (folderDoc.getFolder().isSetStudy())
             {
-                errors.reject(ERROR_MSG, "This file does not appear to be a valid .zip file.");
+                importDir = importDir.resolve(folderDoc.getFolder().getStudy().getDir());
             }
         }
 
+        return importDir.toAbsolutePath().resolve(xmlFileName);
```
labkey-susanh, line 892: Can `xmlFileName` be null?

pipeline/src/org/labkey/pipeline/importer/FolderImportTask.java
```diff
@@ -137,6 +145,11 @@ public RecordedActionSet run() throws PipelineJobException
         return new RecordedActionSet();
     }
 
+    private boolean hasCloudRoot(PipelineJob job)
+    {
+        return job.getPipeRoot().isCloudRoot(); //TODO should this check to target file's protocol?
```
labkey-susanh: I don't know the answer to the TODO question.  Is there a way to do that check here?  How would we have `isCloudRoot` return true but the protocol is not as expected?

study/api-src/org/labkey/api/specimen/pipeline/AbstractSpecimenTask.java
```diff
@@ -188,7 +198,12 @@ public ImportHelper getImportHelper()
 
     public interface ImportHelper
     {
-        VirtualFile getSpecimenDir(PipelineJob job, SimpleStudyImportContext ctx, @Nullable File inputFile) throws IOException, ImportException, PipelineJobException;
+//        default VirtualFile getSpecimenDir(PipelineJob job, SimpleStudyImportContext ctx, @Nullable File inputFile) throws IOException, ImportException, PipelineJobException
+//        {
+//            return getSpecimenDir(job, ctx, inputFile !=  null ? inputFile.toPath() : null);
+//        }
```
labkey-susanh: Remove?

study/src/org/labkey/study/importer/StudyImportJob.java
```diff
@@ -96,6 +96,65 @@ public StudyImportJob(Container c, User user, ActionURL url, File studyXml, Stri
...
+    public void downloadCloudArchive(Path studyXml, BindException errors)
+    {
+        //check if cloud based pipeline root, and study xml hasn't been downloaded already
+        if (!studyXml.startsWith(getPipeRoot().getImportDirectory().toPath().toAbsolutePath()))
+        {
+            if (CloudStoreService.get() != null)   //proxy of is Cloud Module enabled for the current job/container
+            {
+                try
+                {
+                    Path importRoot = CloudStoreService.get().copyExpandedStudyArchiveLocally(studyXml, _ctx.getXml(), _ctx.getRoot().getLocation(), getPipeRoot(), _ctx.getLoggerGetter().getLogger(), errors);
```
labkey-susanh: Probably a naive question at this stage, but is there a chance that the study in the cloud is going to be too big to copy locally?

study/src/org/labkey/study/importer/StudyImporterFactory.java
```diff
@@ -104,14 +106,22 @@ public void process(PipelineJob job, ImportContext<FolderDocument.Folder> ctx, V
                 User user = ctx.getUser();
                 BindException errors = new NullSafeBindException(c, "import");
 
+                //TODO unify with other download location
```
labkey-susanh: Is this to be done now?

study/src/org/labkey/study/importer/StudyImporterFactory.java
```diff
@@ -121,7 +131,14 @@ public void process(PipelineJob job, ImportContext<FolderDocument.Folder> ctx, V
...
 
-                StudyImportContext studyImportContext = new StudyImportContext(user, c, studyDoc, ctx.getDataTypes(), ctx.getLoggerGetter(), studyDir);
+                StudyImportContext studyImportContext = new StudyImportContext.Builder(user,c)
+                        .withDocument(studyDoc)
+                        .withDataTypes(ctx.getDataTypes())
+                        .withLogger(ctx.getLoggerGetter())
+                        .withRoot(useLocalImportDir ? new FileSystemFile(job.getPipeRoot().getImportDirectory()) : studyDir)
+                        .build();
+
+//                new StudyImportContext(user, c, studyDoc, ctx.getDataTypes(), ctx.getLoggerGetter(), studyDir);
```
labkey-susanh: remove?

study/src/org/labkey/study/importer/StudyReload.java
```diff
@@ -178,13 +182,24 @@ public ReloadStatus attemptReload(ImportOptions options, String source) throws I
             return new ReloadStatus("Reload failed");
         }
 
+        private void userCanReloadStudy(User user, StudyImpl study)
+        {
+            if (user == null || !user.isActive())
+            {
+                throw new UnauthorizedException("Reload user is inactive");
+            }
+            if (!study.hasPermission(user, UpdatePermission.class))  //TODO double check required permission
```
labkey-susanh: The permission seems correct to me, but you could verify by impersonating an editor and seeing if reload is available.

study/src/org/labkey/study/importer/StudyReload.java
```diff
@@ -202,36 +217,27 @@ public ReloadStatus reloadStudy(StudyImpl study, ImportOptions options, String s
...
-                    }
+                    studyXml = Files.list(options.getAnalysisDir())
+                            .filter(p -> p.getFileName().toString().equalsIgnoreCase("study.xml"))
+                            .findFirst()
+                            .orElse(null);
                 }
                 else
                 {
-                    studyXml = root.resolvePath("study.xml");
+                    studyXml = root.resolveRelativePath("study.xml");  //TODO this is likely wrong...
```
labkey-susanh: Do we know what the right answer is or why it is likely wrong?  Would be worth adding a comment about why it seems wrong, at least.

pipeline/src/org/labkey/pipeline/importer/FolderImportTask.java
```diff
@@ -59,15 +61,21 @@ public RecordedActionSet run() throws PipelineJobException
         VirtualFile vf;
         FolderImportContext importContext;
         PipelineJob job = getJob();
-        boolean isFileAnalysisJob = FileAnalysisJobSupport.class.isInstance(job); // File watcher triggered job
 
+        if (hasCloudRoot(job) && CloudStoreService.get() != null)
+        {
+            Path importRoot = CloudStoreService.get().downloadFolderArchive(job);
```
labkey-klum: I'm confused by the abstraction of the code which handles the downloading of bits from a cloud store. In this case we are interfacing directly with the `CloudStoreService` to download the archive. In the study import and reload cases the pipeline jobs which implement `StudyJobSupport.downloadCloudArchive` are called on to handle the download. Is there a reason why the pipeline job in the folder import case can't do the same thing?

Additionally the code in `CloudFolderImporter` and `CloudStudyImporter` could be pushed together as they already share some duplicate code. It seems like the only distinction is one downloads a folder archive and the other downloads an expanded archive?

labkey-klum: If we did push more of this code into the pipeline job then `updateWorkingRoot` could be handled inside of the job as part of the download operation. Then we don't need to add the method to the abstract base since it seems like calling this method only makes sense in the context of downloading into a temp location (the study pipeline jobs seem to be doing this in `generateImportContext`).

labkey-klum (changes requested):
It would be nice to clean up the abstraction of the code which invokes and implements the downloading of either a compressed or expanded study or folder archive.

api/src/org/labkey/api/admin/FolderImportContext.java
Sigmonia: removed

api/src/org/labkey/api/pipeline/PathSerialization.java
Sigmonia: added

api/src/org/labkey/api/pipeline/PathSerialization.java
Sigmonia: If the string path isn't viable it will cause an exception, the existing message is fairly clear (at least as clear as I would make it) rather than catch and rethrow, I think it makes sense to leave it as is and let the parser throw if needed.

api/src/org/labkey/api/pipeline/PipelineProvider.java
Sigmonia: switched

api/src/org/labkey/api/study/StudyService.java
Sigmonia: I'm a little concerned an external reference may exist, but I removed the method for now.

api/src/org/labkey/api/util/FileType.java
Sigmonia: deleted

study/src/org/labkey/study/pipeline/StudyReloadSourceJob.java
Sigmonia: consolidated

study/src/org/labkey/study/importer/StudyImportInitialTask.java
Sigmonia: yes, I think...

api/src/org/labkey/api/admin/InvalidFileException.java
Sigmonia: fixed

api/src/org/labkey/api/pipeline/PathSerialization.java
Sigmonia: removed

api/src/org/labkey/api/pipeline/PipeRoot.java
Sigmonia: this TODO is a left over implementation thought.

I think using the implemented getRootNioPath will work correctly and gives the Root Impl the authority of how to relativize the Path.

api/src/org/labkey/api/pipeline/PipelineJob.java
Sigmonia: added

study/src/org/labkey/study/importer/StudyImportJob.java
Sigmonia: Absolutely it is possible. Currently jClouds would be the generator of the Exception on the download attempt. We don't currently check for size constraints but we probably should. I think this might be worth breaking out as a bug to follow up with after the main story is solidified. We also dont do this for the TargetedMS imports AFICT

study/src/org/labkey/study/importer/StudyImporterFactory.java
Sigmonia: removed

study/src/org/labkey/study/importer/StudyReload.java
Sigmonia: The thought was that I was using the incorrect resolution method, but I think it is actually correct. Looking at the previous method AFICT they resolve the same.

api/src/org/labkey/api/pipeline/PipelineJob.java
Sigmonia: This is protection against an error case -- specifically when there is an error and there are notifications to clean up. See line ~1813 where this method is called. The method references the log file, but if this method executes before that then the log file has moved--so it causes a secondary FileNotFoundException.

api/src/org/labkey/api/writer/FileSystemFile.java
Sigmonia: Was thinking some additional logging, but there isn't an obvious logger available.

Removed

api/src/org/labkey/api/writer/FileSystemFile.java
Sigmonia: Was thinking some additional logging, but there isn't an obvious logger available.

Removed.

api/src/org/labkey/api/writer/FileSystemFile.java
Sigmonia: Was thinking some additional logging, but there isn't an obvious logger available.

Removed.

api/src/org/labkey/api/writer/PrintWriters.java
Sigmonia: It was just throwing, removed TODO, and move try block to Path method to catch the new IOException.

api/src/org/labkey/api/pipeline/file/FileAnalysisJobSupport.java
Sigmonia: Short answer is no. Some of these are complicated, and also not part of LabKey's managed code

experiment/src/org/labkey/experiment/api/ExpDataTableImpl.java
Sigmonia: done.

pipeline/src/org/labkey/pipeline/PipelineDirectoryImpl.java
Sigmonia: removed

pipeline/src/org/labkey/pipeline/api/PipeRootImpl.java
Sigmonia: not any more...

pipeline/src/org/labkey/pipeline/api/PipelineManager.java
Sigmonia: Not as far as I can tell, I trace it back to static strings from PipelineController:~1319

pipeline/src/org/labkey/pipeline/importer/FolderImportTask.java
Sigmonia: Consolidated to use the FileUtil to check URI protocol.  (My thought was if we ever allowed GCS) I think by using the helper method here we can ensure that that would be the case.

study/api-src/org/labkey/api/specimen/pipeline/AbstractSpecimenTask.java
Sigmonia: done.

study/src/org/labkey/study/importer/StudyImporterFactory.java
Sigmonia: done.

study/src/org/labkey/study/importer/StudyReload.java
Sigmonia: verified.

pipeline/src/org/labkey/pipeline/api/PipelineManager.java
```diff
@@ -805,68 +814,94 @@ else if (!pipeRoot.isCloudRoot() && !pipeRoot.isUnderRoot(archiveFile))     // T
...
+        //Downloading expanded archive will be handled later in the archive processing...
+        //We don't really have the job context here
+
+        return xmlFile;
+    }
+
+    public static @NotNull Path getXmlFilePathFromArchive(Path importDir, Path archiveFile, @NotNull String xmlFileName) throws InvalidFileException
+    {
+        // when importing a folder archive for a study, the study.xml file may not be at the root
+        if ("study.xml".equals(xmlFileName) && archiveFile.getFileName().toString().endsWith(".folder.zip"))
```
labkey-jeckels: Case sensitivity

pipeline/src/org/labkey/pipeline/api/PipelineManager.java
```diff
@@ -805,68 +814,94 @@ else if (!pipeRoot.isCloudRoot() && !pipeRoot.isUnderRoot(archiveFile))     // T
...
+    }
 
-                xmlFile = new File(importDir, xmlFileName);
-            }
-            catch (FileNotFoundException e)
+    private static Path getImportXmlFile(@NotNull PipeRoot pipelineRoot, @NotNull Path archiveFile, @NotNull String xmlFileName, BindException errors) throws InvalidFileException
+    {
+        Path xmlFile = archiveFile;
+
+        if (archiveFile.getFileName().toString().endsWith(".zip"))
```
labkey-jeckels: Case sensititivity

study/api-src/org/labkey/api/specimen/pipeline/AbstractSpecimenTask.java
```diff
@@ -197,22 +207,22 @@ public ImportHelper getImportHelper()
...
                 for (SpecimenTransform transformer : SpecimenService.get().getSpecimenTransforms(ctx.getContainer()))
                 {
-                    if (transformer.getFileType().isType(inputFile))
+                    if (transformer.getFileType().isType(inputFile.getFileName().toString()))
                     {
                         if (job != null)
                             job.setStatus("TRANSFORMING " + transformer.getName() + " DATA");
-                        specimenArchive = ARCHIVE_FILE_TYPE.getFile(inputFile.getParentFile(), transformer.getFileType().getBaseName(inputFile));
-                        transformer.transform(job, inputFile, specimenArchive);
+                        specimenArchive = ARCHIVE_FILE_TYPE.getFile(inputFile.getParent().toFile(), transformer.getFileType().getBaseName(inputFile));
```
labkey-jeckels: Won't this blow up if it's S3-backed?

study/api-src/org/labkey/api/specimen/pipeline/AbstractSpecimenTask.java
```diff
@@ -250,18 +260,38 @@ public void afterImport(SimpleStudyImportContext ctx)
...
+                                try
+                                {
+                                    if (Files.deleteIfExists(p))
+                                    {
+                                        log.info("Deleting " + p);
+                                    }
+                                }
+                                catch (IOException e)
+                                {
+                                    log.error("Unable to delete file: " + p);
```
labkey-jeckels: Include exception here like a few lines lower?

study/src/org/labkey/study/importer/StudyImportProvider.java
```diff
@@ -55,12 +58,21 @@ public void updateFileProperties(ViewContext context, PipeRoot pr, PipelineDirec
...
     {
         @Override
-        public boolean accept(File file)
+        public boolean accept(Path entry) throws IOException
         {
-            return file.getName().endsWith("study.xml") || file.getName().endsWith(".study.zip") || file.getName().endsWith(".folder.zip");
+            // Path.endsWith checks the final path piece (eg, /root/filename.xml checks filename.xml).
+            // So, we want to check the string value instead
+            String pathString = entry.toString();
+            return pathString.endsWith("study.xml") || pathString.endsWith(".study.zip") || pathString.endsWith(".folder.zip");
```
labkey-jeckels: Case sensitivity

labkey-jeckels (commented):
Smaller suggestions.

Also, there aren't a lot of places here where we check if CloudStoreService.get() == null, but eventually we'll probably want a dummy implementation that lets us avoid doing null checks everywhere, throwing UnsupportedOperationExceptions or no-opping as appropriate.

pipeline/src/org/labkey/pipeline/api/PipelineManager.java
Sigmonia: fixed

pipeline/src/org/labkey/pipeline/api/PipelineManager.java
Sigmonia: fixed

study/api-src/org/labkey/api/specimen/pipeline/AbstractSpecimenTask.java
Sigmonia: Made a change here to use nio.Path

study/api-src/org/labkey/api/specimen/pipeline/AbstractSpecimenTask.java
Sigmonia: Adjusted to use FileUtil.dirDelete

study/src/org/labkey/study/importer/StudyImportProvider.java
Sigmonia: fixed

pipeline/src/org/labkey/pipeline/importer/FolderImportTask.java
Sigmonia: So I did unify the download code into the `CloudArchiveImportSupport` which is in turn implemented by the jobs. This is also where the updateWorkingRoot method is now implemented. I think this allows us to isolate the cloud support to those jobs that implement that job supporter.

labkey-klum (approved):

PR #2544, opened 2021-08-16
43669: R report option groupings won't expand on Safari

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43669

#### Changes
- override `FieldSet.createLegendCt` which creates the fieldset expand/collapse and title components, so that for Safari we render a div instead of a legend element.

### DIFF
--- a/api/webapp/ext-4.2.1/ext-patches.js
+++ b/api/webapp/ext-4.2.1/ext-patches.js
@@ -1104,12 +1104,6 @@ if (Ext4.isSafari || Ext4.isGecko) {
     });
 }
 
-/**
- * Issue 43669: R report option groupings won't expand on Safari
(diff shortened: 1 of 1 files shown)


api/webapp/ext-4.2.1/ext-patches.js
```diff
@@ -1103,3 +1103,47 @@ if (Ext4.isSafari || Ext4.isGecko) {
...
+                    xtype: 'container',
+                    baseCls: me.baseCls + '-header',
+                    id: me.id + '-legend',
+                    items: items,
+                    ownerCt: me,
+                    shrinkWrap: true,
+                    ownerLayout: me.componentLayout
+                };
+
+        if (Ext4.isSafari && me.collapsible) {
```
cnathe, line 1128: add comment with issue number

cnathe (approved):

PR #2547, opened 2021-08-16
Create "Smoke" suite

labkey-tchad:
#### Rationale
Need an extra-quick test suite to verify modules.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/838

#### Changes
* Add `TestWhen.When.SMOKE` for targeting basic module structure


labkey-chrisj (approved):
This looks reasonable enough, I'm assuming there are other changes elsewhere to handle the case where a test has the "SMOKE" `@TestWhen` annotation

PR #2551, opened 2021-08-17
Issue 43751: Support databaseName property on jTDS connection strings

labkey-adam:
#### Rationale
[Issue 43751: Support databaseName property on jTDS connection strings](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43751)

jTDS docs are unclear, but the driver seems to support URLs with a "databaseName" property instead of "/\<database>" immediately following the host name. This upgrades our jTDS JDBC URL parser to look for this property if "/\<database>" isn't found. It also logs and displays the parsed database name in several places to highlight any parsing issues in the future.

The "raw schema metadata" page fails to display the list of tables on SQL Server data sources because our custom `getTables()` implementation doesn't treat a `null` table name pattern as "all tables." (The JDBC docs are unclear, but the PostgreSQL and jTDS drivers both treat `null` as "all tables".) Also, `getTables()` is supposed to return tables sorted by type and name.

#### Changes
* Support "databaseName" property on jTDS connection strings
* Log/display database name at first connection, on the data sources page, on the server info page, and in diagnostics export
* Accept `null` as "all tables" in `getTables()` implementation. Return results sorted by table type and name.


labkey-jeckels (approved):

bigiron/src/org/labkey/bigiron/mssql/BaseMicrosoftSqlServerDialect.java
```diff
@@ -994,21 +994,35 @@ public JdbcHelper getJdbcHelper()
 
         jdbc:jtds:sqlserver://host:1433/database
         jdbc:jtds:sqlserver://host/database;SelectMethod=cursor
+        jdbc:jtds:sqlserver://host:1433;SelectMethod=cursor;databaseName=database
```
labkey-matthewb, line 997: I think we'll be happier if we just go ahead and write/find a JdbcUrlParser that returns prefix, host, port, path, properties.

labkey-matthewb: Then again, I don't think they format is consistent across drivers e.g.  ;prop1=value1;prop2=value2; VS. ?prop1=value1&prop2=value2

Using regexp might help reduce the space-sensitivity of this code.

labkey-matthewb (approved):

PR #2557, opened 2021-08-19
Issue43583: Change message boards notification filtering

RosalineP:
#### Rationale
Previously, whether or not a user was able to view a secure message thread was based off of whether they were included in the initial post's notify list. As of this change, adding a user to a secure message thread allows them to see the contents of the entire thread. 

#### Related Pull Requests
* n/a

#### Changes
* Change permissions for viewing threads such that if a user is on the notify list of the most recent post on the thread, that user can see the entire thread
* Update getting member list to reflect that 'source of truth' member list resides on the most recent post on the thread


announcements/src/org/labkey/announcements/model/SecureMessageBoardPermissions.java
```diff
@@ -83,7 +83,20 @@ public SimpleFilter getThreadFilter()
 
         // Filter for non-editors
         if (!hasPermission(SecureMessageBoardReadPermission.class))
-            filter.addWhereClause("RowId IN (SELECT MessageId FROM " + CommSchema.getInstance().getTableInfoMemberList() + " WHERE UserId = ?)", new Object[]{_user.getUserId()});
+        {
+            filter.addWhereClause(
+            "RowId IN (" +
+                    "SELECT OldestRowId FROM " + CommSchema.getInstance().getTableInfoUserList() +
```
labkey-jeckels: FWIW, reviewing AnnouncementManager.getLatestPostId(), it looks like the comm.threads view already does effectively the same thing

announcements/src/org/labkey/announcements/model/AnnouncementManager.java
```diff
@@ -622,7 +616,8 @@ private static synchronized void saveMemberList(User user, List<Integer> userIds
     // Most recent member list is attached to the most recent approved post
     static List<Integer> getMemberList(AnnouncementModel ann)
     {
-        SQLFragment sql = new SQLFragment("SELECT UserId FROM " + _comm.getTableInfoMemberList() + " WHERE MessageId = ?", ann.getRowId());
+        SQLFragment sql = new SQLFragment("SELECT UserId FROM " + _comm.getTableInfoMemberList() + " WHERE MessageId = ?",
```
labkey-jeckels, line 619: Ideally we'd do this in a single query instead of two successive calls, but since AnnouncementModel.getMemberListIds() caches the result it's not a big deal

labkey-jeckels (approved):

announcements/src/org/labkey/announcements/model/SecureMessageBoardPermissions.java
RosalineP: Good catch! Updated this as of 586c4a7 for the cleanliness bump

PR #2561, opened 2021-08-23
Predefined Dataset Category Automation test

labkey-sweta:
#### Rationale
Predefine assay design to study category automation test

#### Related Pull Requests
https://github.com/LabKey/platform/pull/2561

#### Changes
Additional test case in study/test/src/org/labkey/test/tests/study/AutoLinkToStudyTest.java.


labkey-danield (approved):

labkey-danield (approved):
You have build failures that should be addressed (see failed tests).
Although I'm unclear why a change to this test would trigger some of the builds it did.

labkey-klum (approved):
Can we add an additional check? If we have an existing dataset with a category already assigned to it, a subsequent category assignment through the link to study workflow would not cause the existing category to be overwritten.

labkey-klum (approved):

PR #2568, opened 2021-08-26
Issue 43802: Export runs to XAR via "Write to ... pipeline" fails

labkey-jeckels:
#### Rationale
We've clearly not been testing the ability to export XARs to the pipeline directory, and want it to be a viable option

#### Changes
* Fix serialization error for pipeline job by storing run RowIds instead of full ExpRun objects
* Fix problem when trying to write TSV of assay data in background thread due to lack of active ViewContext
* Redirect to job status page instead of folder home page to monitor progress
* Fail job when export fails, and log errors

### DIFF
--- a/api/src/org/labkey/api/pipeline/PipelineJobNotificationProvider.java
+++ b/api/src/org/labkey/api/pipeline/PipelineJobNotificationProvider.java
@@ -112,7 +112,7 @@ public interface PipelineJobNotificationProvider
                 Integer jobId = PipelineService.get().getJobId(user, job.getContainer(), job.getJobGUID());
                 if (jobId != null)
                 {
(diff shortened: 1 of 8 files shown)


labkey-susanh (commented):
Does this affect only the pipeline exports?  Otherwise, the use of only runIds seems like it wouldn't be portable.

labkey-jeckels:
@labkey-susanh it affects both the browser-download and write-to-pipeline behavior. We're already using RowIds as the way to retain what other objects are included in the export (samples, etc). It's used within the server to know what to write into the XAR file. The RowId doesn't get written to the XAR itself.

labkey-susanh (approved):
Thanks for the clarification, @labkey-jeckels.

PR #2569, opened 2021-08-26
Look up from Panorama to sample type if we find matching sample IDs

labkey-jeckels:
#### Rationale
Users want to link data from their Skyline documents to sample metadata, especially in LKSM

#### Changes
* Make it possible to search for materials by name across containers
* Create LazyForeignKey to make it easy to avoid expensive initialization when it's not required
* Avoid rendering sample URLs when we don't have a matching RowId to use as a parameter

### DIFF
--- a/api/src/org/labkey/api/data/LazyForeignKey.java
+++ b/api/src/org/labkey/api/data/LazyForeignKey.java
@@ -1,0 +1,105 @@
+package org.labkey.api.data;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.labkey.api.collections.NamedObjectList;
+import org.labkey.api.query.FieldKey;
+import org.labkey.api.util.StringExpression;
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/data/LazyForeignKey.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Supplier;
+
+/**
+ * A foreign key implementation that waits until it's asked for something and then uses the Supplier to create the
+ * real foreign key. This lets us postpone potentially expensive initialization of the lookup, and in many cases
+ * avoid needing to do it at all if the lookup isn't being actively used in a given codepath.
+ */
+public class LazyForeignKey implements ForeignKey
```
labkey-susanh, line 18: Nice!

labkey-susanh (approved):

PR #2570, opened 2021-08-27
Issue 43823: admin-showPrimaryLog.view breaks when catalina.base differs from catalina.home

labkey-jeckels:
#### Rationale
In most installs, catalina.base is the same as catalina.home. However, catalina.base is the right place to look for log files.

#### Changes
* Prefer catalina.base for resolving log file location

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -2465,23 +2465,36 @@ public class AdminController extends SpringActionController
         }
     }
 
+    private static File getTomcatBaseDir()
+    {
+        // Issue 43823 - catalina.base and catalina.home typically match, but we should prefer catalina.base
+        String baseDir = System.getProperty("catalina.base");
+        if (baseDir == null)
+        {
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -2465,23 +2465,36 @@ public void export(Object o, HttpServletResponse response, BindException errors)
...
+    {
+        // Issue 43823 - catalina.base and catalina.home typically match, but we should prefer catalina.base
+        String baseDir = System.getProperty("catalina.base");
+        if (baseDir == null)
+        {
+            baseDir = System.getProperty("catalina.home");
+        }
+        if (baseDir == null)
+        {
+            throw new IllegalStateException("Neither catalina.base nor catalina.home system properties were set");
```
labkey-susanh:
```suggestion
            throw new IllegalStateException("Neither catalina.base nor catalina.home system property was set");

```

labkey-susanh (approved):

PR #2572, opened 2021-08-27
Issues 43760 and 43788: Do not mark Barcode fields as not user editable and prevent NPE for folder import 

labkey-susanh:
#### Rationale
`AutoGeneratedDetailsCustomView.getFieldKeysToShow` first filters out all fields that are not user editable, so we need to not set useEditable to false on all fields we want to show in the details view (even if the fields are not meant to be edited by the user; the `readOnly` field provides that protection).

We also don't like NPEs, even if we don't think they should be able to happen.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/616
* https://github.com/LabKey/sampleManagement/pull/679

#### Changes
* Prevent NPE when an archive has a null "WorkflowTask#AssayTypes" value.
* Keep STORAGE_UNIQUE_ID fields userEditable since otherwise the are hidden from details views


experiment/src/org/labkey/experiment/XarExporter.java
```diff
@@ -1281,27 +1281,33 @@ private PropertyCollectionType getProperties(Map<String, ObjectProperty> propert
...
                         {
-                            String[] assayIds = value.getStringValue().split(",");
-                            List<String> protocolStrings = new ArrayList<>();
-                            List<ExpProtocol> protocols = AssayService.get().getAssayProtocols(value.getContainer());
-                            for (String assayId : assayIds)
+                            String assayIdsString = value.getStringValue();
+                            if (!StringUtils.isEmpty(assayIdsString))
                             {
-                                try
+                                String[] assayIds = value.getStringValue().split(",");
```
cnathe: could use `assayIdsString` here (i.e. `assayIdsString.split(",")`)

cnathe (approved):

PR #2573, opened 2021-08-27
43444: ParticipantDetailsWebPart doesn't respect the QC state setting

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=43444

#### Changes
Removes the code in the web part customization page to specify QC public/private data for the participant details web part. This view was never wired up to respect the QC state.


RosalineP (approved):
Looks good!

PR #2574, opened 2021-08-30
Stop warning about JDK 17. Add preliminary JDK 17 JavaDoc link.

labkey-adam:
#### Rationale
JDK 17 will be released in a couple weeks and we've done enough testing that we can claim to support it


labkey-jeckels (approved):

labkey-susanh (approved):
Seems a little premature given that 17 isn't released yet, but OK.

labkey-adam:
Final release candidate is available, so it's effectively released

PR #2576, opened 2021-08-31
Issue 43661: Excessive logging when indexing a .log file containing "\u" that confuses TabLoader

labkey-jeckels:
#### Rationale
Search indexing is treating pipeline job log files, where each line's timestamp contains a comma, as a CSV. That means we treat "\u" as a Unicode escape sequence and log for every instance where it's not the expected Unicode syntax, which can quickly overwhelm the log

#### Changes
* When indexing, don't treat \u as a Unicode escape for sniffed CSV files

### DIFF
--- a/api/src/org/labkey/api/reader/TabLoader.java
+++ b/api/src/org/labkey/api/reader/TabLoader.java
@@ -98,7 +98,7 @@ public class TabLoader extends DataLoader
     public static class CsvFactory extends AbstractDataLoaderFactory
     {
         @NotNull @Override
-        public DataLoader createLoader(File file, boolean hasColumnHeaders, Container mvIndicatorContainer) throws IOException
+        public TabLoader createLoader(File file, boolean hasColumnHeaders, Container mvIndicatorContainer) throws IOException
         {
             TabLoader loader = new TabLoader(file, hasColumnHeaders, mvIndicatorContainer);
             loader.parseAsCSV();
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-adam After I investigated how a CSV-style TabLoader is getting chosen here it seemed safest to disable the Unicode decoding altogether. I could have gone with the message suppression after the first N lines per file instead. Let me know if you'd prefer that approach.

api/src/org/labkey/api/reader/TabLoader.java
```diff
@@ -107,7 +107,7 @@ public DataLoader createLoader(File file, boolean hasColumnHeaders, Container mv
 
         @NotNull @Override
         // A DataLoader created with this constructor does NOT close the reader
```
labkey-adam:
```suggestion
        // A TabLoader created with this constructor does NOT close the reader
```

api/src/org/labkey/api/reader/TabLoader.java
```diff
@@ -121,21 +121,26 @@ public DataLoader createLoader(InputStream is, boolean hasColumnHeaders, Contain
...
+        private TabLoader configParsing(TabLoader loader)
+        {
             loader.setInferTypes(false);
+            // Issue 43661 - Excessive logging when indexing a .log file containing backslash followed by "u" that confuses TabLoader
+            loader.setUnescapeBackslashes(false);
             return loader;
         }
 
         @NotNull @Override
         // A DataLoader created with this constructor does NOT close the reader
```
labkey-adam:
```suggestion
        // A TabLoader created with this constructor does NOT close the reader
```

labkey-adam (approved):
Couple comment update suggestions

PR #2578, opened 2021-08-31
Issue 43821: Escape all pattern parameters in JDBC metadata APIs

labkey-adam:
#### Rationale
JDBC metadata methods are inconsistent with their parameters. The schema and table parameters are sometimes LIKE patterns and sometimes simple strings. When a table name containing LIKE wild cards ("_" or "%") is passed to a pattern parameter, the metadata call can return results for multiple tables, leading to broken queries. See [43821](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43821) for details.

#### Related Pull Requests
* https://github.com/LabKey/synonym/pull/9

#### Changes
* Adjust LabKey's `JdbcMetaDataLocator` (which holds and provides key parameters to the JDBC DatabaseMetaData methods) so callers explicitly indicate:
   * Their filtering intent: multiple schemas vs. single schema, multiple tables vs. single table
   * Whether the metadata call expects a pattern or a simple string
   * Understanding these intents allows the code to escape wild card characters appropriately
* Stop using a transaction to share the connection since we now share connections on each thread by default
* Update our custom SQL Server `getTables()` and `getColumns()` implementations to correctly handle pattern parameters (and therefore the escaping we're introducing here)

### DIFF
--- a/api/src/org/labkey/api/data/dialect/BaseJdbcMetaDataLocator.java
+++ b/api/src/org/labkey/api/data/dialect/BaseJdbcMetaDataLocator.java
@@ -15,9 +15,10 @@
  */
 package org.labkey.api.data.dialect;
 
-import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.NotNull;
 import org.labkey.api.data.ColumnInfo.ImportedKey;
 import org.labkey.api.data.DbScope;
+import org.labkey.api.data.TableInfo;
 
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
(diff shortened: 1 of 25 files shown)


api/src/org/labkey/api/data/dialect/BaseJdbcMetaDataLocator.java
```diff
@@ -75,12 +115,28 @@ public String getCatalogName()
     @Override
     public String getSchemaName()
     {
+        assert null != _schemaName;
```
labkey-jeckels: I suggest making these asserts into IllegalStateExceptions with a message about singleTable() and allTables() not having been called so that we don't end up with unpredictable behavior on production servers

labkey-jeckels (approved):

labkey-jeckels (approved):

PR #2582, opened 2021-08-31
Fix EHR test failures in trigger scripts after refactor to use loadRows()

labkey-jeckels:
#### Rationale
Two categories of problems post-refactor:
- Transaction boundary issues in DatasetDefinition when it hit a problematic row but the calling code wanted to continue gathering additional row errors
- Numeric values being added to the row map in the JS trigger code were coming through as Doubles and not converting to Longs even though they were whole numbers

#### Changes
* Don't push into the transaction stack is one is already active, allowing callers to keep processing without having the connection closed out from under them
* Convert values like "6.0" to Integer and Long without error. Strings or floating points that include decimal values will still error out
* Be more consistent in rejecting decimal values during conversions

### DIFF
--- a/api/src/org/labkey/api/data/ConvertHelper.java
+++ b/api/src/org/labkey/api/data/ConvertHelper.java
@@ -159,8 +159,8 @@ public class ConvertHelper implements PropertyEditorRegistrar
         _register(new NullSafeConverter(new SqlTimeConverter()), java.sql.Time.class);
         _register(new LenientTimestampConverter(), java.sql.Timestamp.class);
         _register(new LenientDateConverter(), java.util.Date.class);
-        _register(new NullSafeConverter(new LongConverter()), Long.class);
-        _register(new LongConverter(), Long.TYPE);
+        _register(new NullSafeConverter(new _LongConverter()), Long.class);
+        _register(new _LongConverter(), Long.TYPE);
         _register(new NullSafeConverter(new LongArrayConverter()), long[].class);
         _register(new ReportIdentifierConverter(), ReportIdentifier.class);
         _register(new RoleConverter(), Role.class);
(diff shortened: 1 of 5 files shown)


study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -2131,13 +2130,22 @@ private void checkConnection(Transaction transaction)
         }
     }
 
+    private Transaction ensureTransaction()
```
labkey-martyp, line 2133: Would this be useful in DbScope for other use cases?  Seems like just a variation of DbScope.ensureTransaction.

api/src/org/labkey/api/data/ConvertHelper.java
```diff
@@ -159,8 +159,8 @@ protected void register()
         _register(new NullSafeConverter(new SqlTimeConverter()), java.sql.Time.class);
         _register(new LenientTimestampConverter(), java.sql.Timestamp.class);
         _register(new LenientDateConverter(), java.util.Date.class);
-        _register(new NullSafeConverter(new LongConverter()), Long.class);
```
labkey-martyp, line 162: Cleanup import

labkey-martyp (approved):
Looks good to me.  @labkey-matthewb or @labkey-klum might have other feedback

labkey-klum (approved):

study/src/org/labkey/study/model/DatasetDefinition.java
labkey-jeckels: I'm open to expanding its availability, but in most cases we'll be inviting other problems by encouraging developers to try unsuccessfully to recover from SQLExceptions which can put the connection into a bad state. I'll leave it as-is for this PR but as we discover more use cases we can centralize and figure out how to describe its proper use

PR #2583, opened 2021-08-31
Issue 43754: SampleTypeDomainKind.hasNullValues override to filter out aliquot rows in "can field be made required" check

cnathe:
### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43754

If a sample type has aliquots, those rows inherit the field values from the parent sample. However in the check that determines if a field can be set as "required" it compares the total sample/row count agains the non-nulls. Since field values in the DB for aliquot samples will always be null, this results in all fields for that sample type not being able to be set as required, even if all parent samples have values for that field. This PR takes aliquot sample rows into account when getting the total row count (i.e. filtering out aliquots).

#### Changes
* Override hasNullValues in SampleTypeDomainKind to add where clause filtering out aliquot sample rows from total row count query

### DIFF
--- a/internal/src/org/labkey/experiment/api/SampleTypeDomainKind.java
+++ b/internal/src/org/labkey/experiment/api/SampleTypeDomainKind.java
@@ -32,6 +32,7 @@ import org.labkey.api.data.JdbcType;
 import org.labkey.api.data.PropertyStorageSpec;
 import org.labkey.api.data.RuntimeSQLException;
 import org.labkey.api.data.SQLFragment;
+import org.labkey.api.data.SqlSelector;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.UpdateableTableInfo;
 import org.labkey.api.exp.ExperimentException;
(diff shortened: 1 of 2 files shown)


internal/src/org/labkey/experiment/api/SampleTypeDomainKind.java
```diff
@@ -517,4 +519,28 @@ public SampleTypeDomainKindProperties getDomainKindProperties(GWTDomain domain,
...
+    public boolean hasNullValues(Domain domain, DomainProperty prop)
+    {
+        SQLFragment allRowsSQL = new SQLFragment();
+        SQLFragment nonBlankRowsSQL = new SQLFragment();
+
+        if (getTotalAndNonBlankSql(domain, prop, allRowsSQL, nonBlankRowsSQL))
+        {
+            // Issue 43754: Don't include aliquot rows in the null value check (see ExpMaterialTableImpl.createColumn for IsAliquot)
+            String table = domain.getStorageTableName();
+            allRowsSQL = new SQLFragment("SELECT * FROM exp.material WHERE LSID IN (")
```
labkey-susanh: Would it be better to call this `nonAliquotRowsSQL`?  It is quite specifically not "all" rows.

labkey-susanh (approved):

XingY (approved):

internal/src/org/labkey/experiment/api/SampleTypeDomainKind.java
cnathe: Very good point! Changed.

PR #2585, opened 2021-08-31
verbose exception for fail FK join

labkey-matthewb:
#### Rationale
Help track down exception with more verbose message


labkey-matthewb:
https://github.com/LabKey/platform/pull/2849

PR #2586, opened 2021-08-31
Issue43657: Option for setting file root wording

RosalineP:
#### Rationale
On the page 'Admin > Folder > Project Settings > Files', the option 'Use a default based on the project-level root' should instead read 'Use a default based on the site-level root' when in a project, and 'Use a default based on the file root of the parent container:' when in a folder or subfolder.

#### Related Pull Requests
* n/a

#### Changes
* Update wording


core/src/org/labkey/core/admin/view/filesProjectSettings.jsp
```diff
@@ -118,7 +118,7 @@
                                 type="radio" name="fileRootOption" id="optionSiteDefault" value="<%=FileRootProp.siteDefault%>"
                                 <%=checked(FileRootProp.siteDefault.name().equals(bean.getFileRootOption()))%>
                                 onclick="updateSelection(<%=h(!FileRootProp.siteDefault.name().equals(bean.getFileRootOption()))%>);">
-                            Use a default based on the project-level root
+                            Use a default based on the <%=h(getContainer().isProject() ? "site-level file root" : "file root of the parent container")%>:
```
labkey-jeckels: I see that Molly recommended using "container" here, but it's a term I try to avoid using outside of developer contexts. I think it's worth a bit more effort to further conditionalize and use "parent project" or "parent folder" as appropriate.

labkey-jeckels (changes requested):

core/src/org/labkey/core/admin/view/filesProjectSettings.jsp
labkey-susanh: Please opt for "parent folder" here as we still consider the use of the term "project" in various contexts.

RosalineP: Resolved!

labkey-jeckels (approved):

PR #2594, opened 2021-09-02
Issue 43818: Indicate uninterruptible upgrade script state with a "lockfile"

labkey-jeckels:
#### Rationale
Some upgrade steps can be missed if Tomcat is shut down in the midst of the upgrade process. We want to avoid this and catch potential problems up front. We're doing that through a lock file that automation can monitor to easily understand the server's status, and the server will refuse to continue if it detects that a previous upgrade attempt was interrupted.

#### Changes
* Write a labkeyUpgradeLockFile to the deployment root (sibling file to the ./modules directory)
* Refuse to start up if a labkeyUpgradeLockFile already exists
* Delete labkeyUpgradeLockFile when startup is incomplete

### Notes
* This makes the critical section larger than may be strictly required for many startups but should be guaranteed to handle all upgrades, deferred and not
* This does NOT currently monitor pipeline jobs that may get queued to import or migrate data


labkey-stuartm (approved):

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -610,6 +618,8 @@ public void addStaticWarnings(@NotNull Warnings warnings)
         startNonCoreUpgradeAndStartup(User.getSearchUser(), execution, coreRequiredUpgrade);  // TODO: Change search user to system user
 
         _log.info("LabKey Server startup is complete; " + execution.getLogMessage());
+
+        lockFile.delete();
```
labkey-adam: In most cases, upgrade is not complete at this point. The rest of the message is usually "modules are being upgraded and initialized in the background".

I think the bottom of `afterUpgrade()` is closer to the right place. Need to verify that this gets called in all cases (including non-upgrade startup).

labkey-adam (changes requested):

api/src/org/labkey/api/module/ModuleLoader.java
labkey-jeckels: You're correct - I was misunderstanding the threading behavior. afterUpgrade() appears to be called in all startups in my testing.

Note that this change may be annoying on dev machines as it takes a while to get that far into the startup process. If so, we could pull it back by checking if we have any deferred upgrades, or perhaps even if there are any version changes, and delete the lock file much earlier if there won't be any upgrade work to do.

labkey-adam: So the dev machine annoyance is the fact that restarting before startup is complete means a ConfigurationException and need to manually delete the lock file? We could tone down the warning in dev mode... e.g., log a loud warning but delete the file and continue upgrade.

labkey-jeckels: I also thought about that approach but hadn't implemented it. Assuming other devs have similar build/restart patterns as me, it seems worth special casing. Update pushed.

curiositycasualty: Would it make sense, or be possible to conceal this behind a feature flag we could use to keep this disabled for dev scenarios?

labkey-jeckels: @curiositycasualty I pushed a change that logs a warning when you're running in dev mode (an existing flag) but doesn't halt the startup process as it does in production mode.

labkey-adam: Dev mode seems best to me since we want to make a fuss about interrupted upgrades on all production servers, including those that don't use our automation (and hence don't use any special feature flags).

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -629,6 +637,34 @@ private void checkForRenamedModules()
...
+                _log.warn("Bravely continuing because this server is running in Dev mode.");
+            }
+            else
+            {
+                throw new ConfigurationException("Lock file " + FileUtil.getAbsoluteCaseSensitiveFile(result) + " already exists - a previous upgrade attempt may have left the server in an indeterminate state. Proceed with extreme caution as the database may not be properly upgraded. To continue, delete file and restart Tomcat.");
+            }
+        }
+        try (BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(new FileOutputStream(result), StringUtilsLabKey.DEFAULT_CHARSET)))
+        {
+            writer.write("LabKey instance beginning initialization at " + new Date());
```
labkey-adam: We try not to output `Date.toString()`...
```suggestion
            writer.write("LabKey instance beginning initialization at " + DateUtil.formatDateTimeISO8601(new Date()));

```

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -629,6 +637,34 @@ private void checkForRenamedModules()
...
+        if (result.exists())
+        {
+            if (AppProps.getInstance().isDevMode())
+            {
+                _log.warn("Lock file " + FileUtil.getAbsoluteCaseSensitiveFile(result) + " already exists - a previous upgrade attempt may have left the server in an indeterminate state.");
+                _log.warn("Bravely continuing because this server is running in Dev mode.");
+            }
+            else
+            {
+                throw new ConfigurationException("Lock file " + FileUtil.getAbsoluteCaseSensitiveFile(result) + " already exists - a previous upgrade attempt may have left the server in an indeterminate state. Proceed with extreme caution as the database may not be properly upgraded. To continue, delete file and restart Tomcat.");
```
```suggestion
                throw new ConfigurationException("Lock file " + FileUtil.getAbsoluteCaseSensitiveFile(result) + " already exists - a previous upgrade attempt may have left the server in an indeterminate state. Proceed with extreme caution as the database may not be properly upgraded. To continue, delete the file and restart Tomcat.");

```

labkey-adam (approved):
Couple minor suggestions

PR #2595, opened 2021-09-02
Update testZipOfFiles to catch error - avoid popup on test case

labkey-jeckels:
#### Rationale
As written this TargetedMSQCTest.testZipOfFiles is missing an alert that pops up: "Relative path of file - first.zipme is incorrect"

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/419

#### Changes
* Avoid pop up in check that's relevant for the real-world directory zipping cases but not important for our artificial but testable single-file upload


labkey-jeckels:
TargetedMS tests are now passing with the more stringent check for errors

labkey-nicka (approved):

PR #2540, opened 2021-08-11
Link to Study Dataset Category Assignment

RosalineP:
#### Rationale
It is desired that we add functionality to be able to categorize the dataset resulting from a link to study at the assay design level, for the auto-link case. For the manual link to study case, the dataset category assignment can now happen at the initial 'publishStart' page.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/604

#### Changes
* Roundtrip new 'autoLinkCategory' property
* Support dataset category assignment for auto-link, manual link, and manual link using pipeline
* Prevent entering Dataset Category name in Assay Design that's longer than 300 characters
* Add new 'Specify Linked Dataset Category' text input on 'publishStart' page

PR #2596, opened 2021-09-02
Link to Study Dataset Category Assignment: Sample Types

RosalineP:
#### Rationale
It is desired that we add functionality to be able to categorize the dataset resulting from a link to study at the sample type design level, for the auto-link case. For the manual link to study case, the dataset category assignment can now happen at the initial 'publishStart' page.
This functionality was initially implemented for Assay link to study, and these changes will allow for functionality parity on Sample Types.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2540
* https://github.com/LabKey/labkey-ui-components/pull/622

#### Changes
* Support roundtrip of new sample type property, 'AutoLinkCategory'
* Add call to ensureViewCategory() upon automatic or manual link to study cases to generate or locate dataset category
* Slightly update publishChooseStudy.jsp wording
* Reorder where 'Auto Link Target Container' appears on `showSampleType` view of sample type within Sample Type Properties


api/src/org/labkey/api/study/publish/publishChooseStudy.jsp
```diff
@@ -217,8 +217,8 @@
     <%
         if (!autoLinkEnabled)
         {
-            String autoLinkCategoryTip = "Specify the desired category for the Assay Dataset that will be created (or appended to) in the target study when rows are linked. " +
-                    "If the category you specify does not exist, it will be created. If the Assay Dataset already exists, this setting will not overwrite a previously assigned category. " +
+            String autoLinkCategoryTip = "Specify the desired category for the Dataset that will be created (or appended to) in the target study when rows are linked. " +
```
RosalineP, line 232: It was a little bit of a cheat to just change 'Assay Dataset' to 'Dataset' in order to be generic to both link to study types. 
The alternative is to pass a way to tell whether we're coming from an assay or a sample type, or to use the existing values passed to figure it out. I was a fan of just changing the text to 'Dataset' because I think that that's still clear and informative, but I'm open to other thoughts!

api/src/org/labkey/api/study/publish/StudyPublishService.java
```diff
@@ -88,10 +88,6 @@ ActionURL publishData(User user, Container sourceContainer, Container targetCont
                           Pair<Dataset.PublishSource, Integer> publishSource,
```
RosalineP, line 88: Note that the publishData() starting on line 87 (only difference in signature is that it doesn't include the 'targetContainer' param) has only one usage. I waffled on whether to change that usage to the publishData() with the 'targetContainer' param set to null, and delete the function signature. I ultimately thought I'd leave it as-is to make it a little clearer that Category setting in link-to-study is not supported in (what seems to be) Vaccine Study Design study generation

labkey-klum: sounds right, good call.

api/src/org/labkey/api/study/publish/publishChooseStudy.jsp
labkey-klum: This is a reasonable change, we can always revisit if users find it unclear but I'm guessing this will be sufficient.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -569,13 +569,14 @@ public ExpSampleTypeImpl createSampleType(Container c, User u, String name, Stri
...
-                parentCol, nameExpression, templateInfo, null, null, null, null);
+                parentCol, nameExpression, templateInfo, null, null, null, null, null);
     }
 
     @NotNull
     @Override
     public ExpSampleTypeImpl createSampleType(Container c, User u, String name, String description, List<GWTPropertyDescriptor> properties, List<GWTIndex> indices, int idCol1, int idCol2, int idCol3, int parentCol,
-                                              String nameExpression, @Nullable TemplateInfo templateInfo, @Nullable Map<String, String> importAliases, @Nullable String labelColor, @Nullable String metricUnit, @Nullable Container autoLinkTargetContainer)
+                                              String nameExpression, @Nullable TemplateInfo templateInfo, @Nullable Map<String, String> importAliases, @Nullable String labelColor, @Nullable String metricUnit,
+                                              @Nullable Container autoLinkTargetContainer, @Nullable String autoLinkCategory)
```
labkey-klum, line 579: lol so many parameters, at some point we might consider moving to a builder pattern.

labkey-klum (changes requested):
Just wanted to verify that the new property is round-tripped in folder export. It looks like schema changes were made to the xsd but I didn't notice any changes to `XarReader` or `XarExporter`.

RosalineP:
> Just wanted to verify that the new property is round-tripped in folder export. It looks like schema changes were made to the xsd but I didn't notice any changes to `XarReader` or `XarExporter`.

Oop, I missed that because it wasn't in my tasks. I've implemented it as of the last commit e83c627!

labkey-klum (approved):

experiment/resources/schemas/dbscripts/sqlserver/exp-21.010-21.011.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE exp.MaterialSource ADD AutoLinkCategory VARCHAR(200) NULL;
```
XingY, line 1: @labkey-adam Should we stick with NVARCHAR for sqlserver, or does it not matter?

labkey-adam: Yes, I would stick with NVARCHAR

PR #2597, opened 2021-09-02
Issue 43850: Need a way to convert from text to JSON and JSONB in PG

labkey-jeckels:
#### Rationale
We introduced pass throughs for Postgres JSON operators and functions but don't support casting a VARCHAR or other data type to JSON or JSONB

#### Changes
* Introduce parse_json and parse_jsonb LabKey SQL functions to do the casting

### DIFF
--- a/query/src/org/labkey/query/sql/Method.java
+++ b/query/src/org/labkey/query/sql/Method.java
@@ -1481,7 +1481,7 @@ public abstract class Method
             }
         });
 
-        // Special functions to cast an argument to the JSON or JSONB data types
+        // Special functions to cast an argument to the JSON or JSONB data types without needing to support as official datatype in CAST
         postgresMethods.put("parse_json", new ParseJSONMethod("json"));
         postgresMethods.put("parse_jsonb", new ParseJSONMethod("jsonb"));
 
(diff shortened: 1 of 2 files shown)


query/src/org/labkey/query/sql/Method.java
```diff
@@ -1571,4 +1575,27 @@ private static void addJsonPassthroughMethod(String name, JdbcType type, int min
...
+        ParseJSONMethod(String targetType)
+        {
+            super(JdbcType.OTHER, 1, 1);
+            _targetType = targetType;
+        }
+
+        @Override
+        public MethodInfo getMethodInfo()
+        {
+            return new AbstractMethodInfo(JdbcType.OTHER)
```
labkey-matthewb, line 1591: Almost time to add a new JdbcType!

labkey-matthewb (changes requested):
Maybe add a test to Query.postgresOnlyFunctions

query/src/org/labkey/query/sql/Method.java
labkey-jeckels: I know, I considered it but didn't feel brave enough to plumb it through

labkey-jeckels:
@labkey-matthewb thanks for the pointer, I added a few simple test cases

labkey-matthewb (approved):

labkey-jeckels:
Wrong branch name for 21.7. Will open a new one

PR #2598, opened 2021-09-02
 Issue 43850: Need a way to convert from text to JSON and JSONB in PG

labkey-jeckels:
#### Rationale
We introduced pass throughs for Postgres JSON operators and functions but don't support casting a VARCHAR or other data type to JSON or JSONB

#### Changes
* Introduce parse_json and parse_jsonb LabKey SQL functions to do the casting

### DIFF
--- a/query/src/org/labkey/query/sql/Query.java
+++ b/query/src/org/labkey/query/sql/Query.java
@@ -1801,6 +1801,9 @@ public class Query
         new SqlTest("SELECT 'similar' WHERE similar_to('abc','%(b|d)%')", 1, 1),
         new SqlTest("SELECT 'similar' WHERE similar_to('abc','(b|c)%')", 1, 0),
         new SqlTest("SELECT 'similar' WHERE similar_to('abc|','abc\\|', '\\')", 1, 1),
+        new SqlTest("SELECT parse_jsonb('{\"a\":1, \"b\":null}')", 1, 1),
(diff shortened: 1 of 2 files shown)


labkey-jeckels:
If someone else is able to approve this first, Matt already reviewed and approved the same change in a different PR where I messed up the branch name.

https://github.com/LabKey/platform/pull/2597

query/src/org/labkey/query/sql/Query.java
```diff
@@ -1801,6 +1801,9 @@ private void verifyType(ColumnInfo column, JdbcType expectedType, @Nullable Stri
         new SqlTest("SELECT 'similar' WHERE similar_to('abc','%(b|d)%')", 1, 1),
         new SqlTest("SELECT 'similar' WHERE similar_to('abc','(b|c)%')", 1, 0),
         new SqlTest("SELECT 'similar' WHERE similar_to('abc|','abc\\|', '\\')", 1, 1),
+        new SqlTest("SELECT parse_jsonb('{\"a\":1, \"b\":null}')", 1, 1),
```
labkey-susanh, line 1804: No test for `parse_json`?

labkey-susanh (approved):

query/src/org/labkey/query/sql/Query.java
labkey-jeckels: I'll add one. It doesn't look like we exhaustively test pass-throughs and I didn't add any for the original JSON support, but I just added a couple to make sure that JSON works, not just JSONB

PR #2604, opened 2021-09-07
Item 9422 : Naming patterns for Aliquots

XingY:
#### Rationale
Our existing naming pattern function is incapable to create aliquot names that's usually a parent name followed by a count, such as parent-1, parent-2. An advanced "XML Counter Columns" configuration is needed to achieve such goals. To allow generating aliquot names following client specified conventions, "AliquotNameExpression" field is added to sample types, and additionally, a new ${nameprefix:withCounter} expression syntax is introduced to allow generating columns-count style names.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2604
* https://github.com/LabKey/labkey-ui-components/pull/627
* https://github.com/LabKey/sampleManagement/pull/688
* https://github.com/LabKey/biologics/pull/990

#### Changes
* Added AliquotNameExpression to MaterialSource
* Update sample creation to generate aliquot name based on AliquotNameExpression
* Support parsing and evaluation of the new ${nameprefix:withCounter} expression


api/src/org/labkey/api/exp/api/SampleTypeService.java
```diff
@@ -74,7 +74,7 @@ ExpSampleType createSampleType(Container container, User user, String name, Stri
 
     @NotNull
     ExpSampleType createSampleType(Container container, User user, String name, String description, List<GWTPropertyDescriptor> properties, List<GWTIndex> indices, int idCol1, int idCol2, int idCol3, int parentCol,
-                                   String nameExpression, @Nullable TemplateInfo templateInfo, @Nullable Map<String, String> importAliases, @Nullable String labelColor, @Nullable String metricUnit, @Nullable Container autoLinkTargetContainer)
+                                   String nameExpression, String aliquotNameExpression, @Nullable TemplateInfo templateInfo, @Nullable Map<String, String> importAliases, @Nullable String labelColor, @Nullable String metricUnit, @Nullable Container autoLinkTargetContainer)
```
labkey-susanh: Wow.  That's a crazy number of parameters for this method.

experiment/resources/schemas/dbscripts/sqlserver/exp-21.010-21.011.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE exp.MaterialSource ADD AliquotNameExpression NVARCHAR(200) NULL;
+GO
```
labkey-susanh: Is the GO needed?  I always forget...

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -288,7 +325,7 @@ public void close()
             _rowNumber = -1;
         }
 
-        private String nextName(Map<String, Object> rowMap, Set<ExpData> parentDatas, Set<ExpMaterial> parentSamples, @Nullable Supplier<Map<String, Object>> extraPropsFn)
+        private String nextName(Map<String, Object> rowMap, Set<ExpData> parentDatas, Set<ExpMaterial> parentSamples, @Nullable Supplier<Map<String, Object>> extraPropsFn, FieldKeyStringExpression altExpression)
```
labkey-susanh: `@Nullable FieldKeyStringExpression`?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -359,7 +397,9 @@ private String genName(@NotNull Map<String, Object> rowMap,
             // Add extra context variables
             Map<String, Object> ctx = additionalContext(rowMap, parentDatas, parentSamples, sampleCounts, extraProps);
 
-            String name = _parsedNameExpression.eval(ctx);
+            // allow using alternative expression for evaluation.
+            // for example, use AliquotNameExpression instead of NameExpression if sample is aliquot
+            String name = (altExpression != null ? altExpression : _parsedNameExpression).eval(ctx);
```
labkey-susanh: I wonder if it would be cleaner to initialize the NameGenerator with the other expression instead of having this "alt" parameter.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -532,6 +572,230 @@ else if (parts[0].equalsIgnoreCase(ExpMaterial.MATERIAL_INPUT_PARENT))
...
+    /**
+     *  Same as FieldKeyExpression, but supports :withCounter syntax
+     */
+    public static class NameGenerationExpression extends FieldKeyStringExpression
+    {
+        private final Container _container;
+        private Function<String, Long> _getNonConflictCountFn;
+        private String _counterSeqPrefix;
+
+        public NameGenerationExpression(String source, boolean urlEncodeSubstitutions, NullValueBehavior nullValueBehavior, boolean allowSideEffects, Container container)
```
labkey-susanh: Do these need/want to be public given the static `create` methods?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -532,6 +572,230 @@ else if (parts[0].equalsIgnoreCase(ExpMaterial.MATERIAL_INPUT_PARENT))
...
+                String numberFormat = counterMatcher.group(3);
+                if (!StringUtils.isEmpty(startIndStr))
+                {
+                    try
+                    {
+                        startInd = Integer.valueOf(startIndStr);
+                    }
+                    catch (NumberFormatException e)
+                    {
+                        // ignore illegal startInd
```
labkey-susanh, line 631: It will probably be too verbose to log an error here, right?  I think when we tackle the story of doing some better validation of name expressions, we'll want to attach this error somewhere so it can be surfaced to the user.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -532,6 +572,230 @@ else if (parts[0].equalsIgnoreCase(ExpMaterial.MATERIAL_INPUT_PARENT))
...
+                        break;
+
+                    if (openCount < 0)
+                        throw new IllegalArgumentException("Illegal expression: open and close tags are not matched.");
+                }
+
+                if (openCount == 0)
+                {
+                    String sub = _source.substring(openIndex + 2, subInd - 1);
+                    StringExpressionFactory.StringPart part = parsePart(sub);
```
labkey-susanh, line 688: Is this going to do a recursive parsing of `sub` then?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -532,6 +572,230 @@ else if (parts[0].equalsIgnoreCase(ExpMaterial.MATERIAL_INPUT_PARENT))
...
+        }
+
+        @Override
+        public String getValue(Map map)
+        {
+            String prefix = _parsedNameExpression.eval(map);
+            long existingCount = -1;
+
+            if (_getNonConflictCountFn != null)
+                existingCount = _getNonConflictCountFn.apply(prefix);
```
labkey-susanh: Is this going to be called for each value?  Since it does a select each time that's probably going to be more expensive than we want it to be.  We should consider if the function can cache a value that will prevent the query from happening each time, particularly when importing a file.  The `GenerateUniqueDataIterator` might be helpful in this regard.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -532,6 +572,230 @@ else if (parts[0].equalsIgnoreCase(ExpMaterial.MATERIAL_INPUT_PARENT))
...
+                _counterSequences.put(prefix, newSequence);
+            }
+
+            long count = _counterSequences.get(prefix).next();
+
+            Object countStr = String.valueOf(count);
+            if (_counterFormat != null)
+            {
+                countStr = _counterFormat.format(count);
+            }
```
labkey-susanh: Could be done as:
```
Object countStr;
if (_counterFormat != null) 
  countStr = _counterForma.format(count);
else
  countStr = String.valueOf(count);
```
Saves the conversion to string in some cases.

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -747,4 +804,38 @@ public void setImportAliasMap(Map<String, String> aliasMap)
...
+                try
+                {
+                    long id = Long.parseLong(sampleId);
+                    if (id > max)
+                        max = id;
+                }
+                catch (NumberFormatException ignored) {
+                    ;
+                }
+            }
```
labkey-susanh: It might be slightly more efficient to filter out the non-number suffixes, just because throwing exceptions is relatively expensive. You would have to walk the list twice in this case, of course.

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -747,4 +804,38 @@ public void setImportAliasMap(Map<String, String> aliasMap)
...
+        return (sampleNamePrefix) ->
+        {
+            long max = 0;
+
+            SimpleFilter filter = SimpleFilter.createContainerFilter(container);
+            filter.addCondition(FieldKey.fromParts("cpastype"), sampleTypeLsid);
+            filter.addCondition(FieldKey.fromParts("Name"), sampleNamePrefix, STARTS_WITH);
+
+            TableSelector selector = new TableSelector(ExperimentServiceImpl.get().getTinfoMaterial(), Collections.singleton("Name"), filter, null);
+            final List<String> sampleIds = new ArrayList<>();
```
labkey-susanh: Maybe call this something like `nameSuffixes` since it's not the full sampleId involved here.

experiment/src/org/labkey/experiment/api/ExpDataClassImpl.java
```diff
@@ -367,4 +375,38 @@ public String getDocumentId()
...
+        return (dataNamePrefix) ->
+        {
+            long max = 0;
+
+            SimpleFilter filter = SimpleFilter.createContainerFilter(container);
+            filter.addCondition(FieldKey.fromParts("CpasType"), dataclassLsid);
+            filter.addCondition(FieldKey.fromParts("Name"), dataNamePrefix, STARTS_WITH);
+
+            TableSelector selector = new TableSelector(ExperimentServiceImpl.get().getTinfoData(), Collections.singleton("Name"), filter, null);
+            final List<String> dataNames = new ArrayList<>();
```
labkey-susanh: `nameSuffixes`?

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -747,4 +804,38 @@ public void setImportAliasMap(Map<String, String> aliasMap)
     {
         _object.setMaterialParentImportAliasMap(SampleTypeServiceImpl.get().getAliasJson(aliasMap, _object.getName()));
     }
+
+    @Override
+    public Function<String, Long> getMaxSampleCounterFunction()
```
labkey-susanh, line 820: This function is very much like the one for data classes. Can we factor it out so it can actually be shared?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -614,6 +614,11 @@ public ExpSampleTypeImpl createSampleType(Container c, User u, String name, Stri
         if (nameExpression != null && nameExpression.length() > nameExpMax)
             throw new ExperimentException("Name expression may not exceed " + nameExpMax + " characters.");
 
+        // Validate the aliquot name expression length
+        int aliquotNameExpMax = materialSourceTable.getColumn("AliquotNameExpression").getScale();
+        if (aliquotNameExpression != null && aliquotNameExpression.length() > aliquotNameExpMax)
+            throw new ExperimentException("Aliquot Name expression may not exceed " + aliquotNameExpMax + " characters.");
```
labkey-susanh:
```suggestion
            throw new ExperimentException("Aliquot naming pattern may not exceed " + aliquotNameExpMax + " characters.");

```

internal/src/org/labkey/experiment/api/SampleTypeDomainKind.java
```diff
@@ -353,6 +353,10 @@ public void validateOptions(Container container, User user, SampleTypeDomainKind
         if (StringUtils.isNotBlank(options.getNameExpression()) && options.getNameExpression().length() > nameExpMax)
             throw new IllegalArgumentException("Value for Name Expression field may not exceed " + nameExpMax + " characters.");
 
+        int aliquotNameExpMax = materialSourceTI.getColumn("AliquotNameExpression").getScale();
+        if (StringUtils.isNotBlank(options.getAliquotNameExpression()) && options.getAliquotNameExpression().length() > aliquotNameExpMax)
+            throw new IllegalArgumentException("Value for Aliquot Name Expression field may not exceed " + nameExpMax + " characters.");
```
labkey-susanh:
```suggestion
            throw new IllegalArgumentException("Value for Aliquot Naming Pattern field may not exceed " + nameExpMax + " characters.");

```

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -570,6 +834,272 @@ public String getName()
             return _name;
         }
     }
+
+    public static class TestCase extends Assert
```
labkey-susanh, line 839: Thanks for adding these tests!  These cover non-error cases mostly. It would also be useful to have test for error cases so we can "document" the expected behaviors.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -570,6 +834,272 @@ public String getName()
...
+                StringExpression se = NameGenerationExpression.create("${d:date('& yy-MM-dd'):htmlEncode}", false);
+                String s = se.eval(m);
+                assertEquals("&amp; 11-12-03", s);
+            }
+
+            {
+                StringExpression se = NameGenerationExpression.create("${d:date('ISO_ORDINAL_DATE')}", false);
+                String s = se.eval(m);
+                assertEquals("2011-337", s);
+            }
```
labkey-susanh, line 873: Should there be tests added for error cases? Are non-dates a possibility?  What happens for non-date-formats?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -570,6 +834,272 @@ public String getName()
...
+                        "${${AliquotedFrom}.${SourceMeta}.:withCounter}", false, NullValueBehavior.ReplaceNullWithBlank, true, c, null);
+
+                ArrayList<StringExpressionFactory.StringPart> parsedExpressions = se.getParsedExpression();
+
+                assertEquals(1, parsedExpressions.size());
+                assertEquals(4, se.getDeepParsedExpression().size());
+                assertTrue(parsedExpressions.get(0) instanceof NameGenerator.CounterExpressionPart);
+
+                String s = se.eval(m);
+                assertEquals("S100.mouse1.1", s);
```
labkey-susanh, line 1075: Might be good to include a test for generating more than one, so you can see the counter increment properly. Does it increment when "mouse1" becomes "mouse2", for example?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -570,6 +834,272 @@ public String getName()
...
+                StringExpression se = NameGenerationExpression.create("${d:number('0.0000')}", false);
+                String s = se.eval(m);
+                assertEquals("123456.7890", s);
+            }
+
+            {
+                StringExpression se = NameGenerationExpression.create("${d:number('000000000')}", false);
+                String s = se.eval(m);
+                assertEquals("000123457", s);
+            }
```
labkey-susanh, line 910: What happens with non-numbers in the number format?

api/src/org/labkey/api/data/NameGenerator.java
XingY: Because of State (used to increment sample counts, such as dailySampleCount), a single NameGenerator needs to be used for both sample name and aliquot name. NameGenerator needs pick expression per data record.

api/src/org/labkey/api/data/NameGenerator.java
XingY: Yeah it would log one per sample, probably too much. Agree it should be handled by the validation story.

api/src/org/labkey/api/data/NameGenerator.java
XingY: No, parsePart is not recursive. If the expression does contain recursive expression other than :withCounter, it would generate unexpected names.

api/src/org/labkey/api/data/NameGenerator.java
XingY: I moved the _getNonConflictCountFn call to inside  if (!_counterSequences.containsKey(prefix)), that way _counterSequences only queries for each prefix once. It was left outside the check by mistake.

api/src/org/labkey/api/data/NameGenerator.java
labkey-susanh: Got it.  OK.

api/src/org/labkey/api/data/NameGenerator.java
XingY: good suggestion, added.

api/src/org/labkey/api/data/NameGenerator.java
XingY: interestingly non-dates and invalid date-formats both silently parses, and generate an output.

api/src/org/labkey/api/data/NameGenerator.java
XingY: added tests for both invalid number format as well as non-numbers

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
XingY: refactored

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
XingY: good point, changed to use regex

experiment/resources/schemas/dbscripts/sqlserver/exp-21.010-21.011.sql
XingY: I recall Adam saying it doesn't matter since rollup scripts does the right thing..

labkey-adam: It's never needed as the last statement in a script, since each script gets run in its own `execute()`. At consolidation time, we'll add GO statements if they're needed (e.g., if the next script references AliquotNameExpression).

api/src/org/labkey/api/data/NameGenerator.java
labkey-susanh: We'll want to be sure to call this out in the documentation (and try to remember to flag this when doing the validation that I really hope we get to do).

api/src/org/labkey/api/data/NameGenerator.java
labkey-susanh: well, now we know...  I would not have predicted much of this behavior.

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
labkey-susanh: Nice.  Thanks.

labkey-susanh (approved):

PR #2608, opened 2021-09-08
Postgres 9.6 doesn't support direct casting from JSONB to numeric

labkey-jeckels:
#### Rationale
TeamCity has informed us that Postgres 9.6 doesn't support directly casting from JSONB and JSON to INTEGER and other numeric types. We're deprecating support for 9.6 so this isn't really an important functional difference but this will clean up test failures.

https://teamcity.labkey.org/buildConfiguration/LabKey_217Release_Internal_InternalSuites_OldDependenciesPostgres/1530613?expandedTest=build%3A%28id%3A1530613%29%2Cid%3A14693

#### Changes
* Do a VARCHAR-based validation instead
* Fill in JSON variant that was intended for original merge


labkey-tchad:
Verifying pg 9.6 on TeamCity: https://teamcity.labkey.org/buildConfiguration/LabKey_217Release_Internal_InternalSuites_OldDependenciesPostgres/1531435

labkey-tchad (approved):

PR #2611, opened 2021-09-09
Issue 43853: Update GetContainersAction to add container user permissions check before adding results to response

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=43853

#### Changes
* Add check that user has read access to a container before adding info to the response object

### DIFF
--- a/core/src/org/labkey/core/portal/ProjectController.java
+++ b/core/src/org/labkey/core/portal/ProjectController.java
@@ -1495,7 +1495,8 @@ public class ProjectController extends SpringActionController
                 List<Map<String, Object>> containerJSON = new ArrayList<>();
                 for (Container c : containers)
                 {
-                    containerJSON.add(getContainerJSON(c, user, propertiesToSerialize));
+                    if (c != null)
+                        containerJSON.add(getContainerJSON(c, user, propertiesToSerialize));
                 }
                 resultMap = Collections.singletonMap("containers", containerJSON);
             }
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/portal/ProjectController.java
```diff
@@ -1495,17 +1495,20 @@ public ApiResponse execute(GetContainersForm form, BindException errors)
                 List<Map<String, Object>> containerJSON = new ArrayList<>();
                 for (Container c : containers)
                 {
-                    if (c != null)
+                    if (c != null && c.hasPermission(user, ReadPermission.class)) // 43853
```
labkey-jeckels: We should at least consider the desired behavior when a user has no access to the project but has read permission to a subfolder.

getContainerJSON() calls to getVisibleChildren() which does this kind of filtering and recursion. Do we need to pop this up a level?

labkey-jeckels (changes requested):

core/src/org/labkey/core/portal/ProjectController.java
cnathe: Good point. I updated this PR so that they check for if the user has container read perm is done within the getContainerJSON. This should account for the scenario you mentioned of the user having perm to a child container but not the parent / requested container.

labkey-jeckels (approved):

PR #2614, opened 2021-09-10
Fix NPE in Mascot tests

Sigmonia:
#### Rationale
Missed a method override which caused some of the derrived classes tests to fail 

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/365

### DIFF
--- a/internal/src/org/labkey/api/exp/AbstractFileXarSource.java
+++ b/internal/src/org/labkey/api/exp/AbstractFileXarSource.java
@@ -88,15 +88,18 @@ public abstract class AbstractFileXarSource extends XarSource
     }
 
     @Override
+    @Nullable
+    @Deprecated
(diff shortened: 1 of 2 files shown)


internal/src/org/labkey/api/exp/AbstractFileXarSource.java
```diff
@@ -88,15 +88,18 @@ public ExperimentArchiveDocument getDocument() throws XmlException, IOException
     }
 
     @Override
+    @Nullable
```
labkey-susanh, line 91: Do the things that are consuming this know this is nullable and treat it properly?

It looks like the MS2Manager implementation requires it to be not null.  I'm confused.

labkey-susanh (approved):

PR #2616, opened 2021-09-10
Issue43761: Deadlock calling ViewCategoryManager.ensureViewCategory() within transaction 

RosalineP:
#### Rationale
Previously, calling ViewCategoryManager.ensureViewCategory() would occasionally result in a deadlock, as uncovered by integration tests. This change resolves the issue causing the intermittent deadlocks, and re-instates desired changes in link-to-study category assignment.

#### Related Pull Requests
* n/a

#### Changes
* Push down category name string rather than a ViewCategory down to _publishData
* Wrap saveCategory() in transaction and move deadlock-causing calls to clear cache


labkey-klum (approved):
yay, great to see that this approach worked

PR #2618, opened 2021-09-13
Rotate labkey-errors.log (43686), show logger info in Admin console (43787)

labkey-jeckels:
#### Rationale
* Runaway error logging can fill the disk via labkey-errors.log, further compounding server problems
* Users like to know what loggers may have debug logging of interest. They can currently find some guidance at https://www.labkey.org/Documentation/wiki-page.view?name=loggers
* Issue 43787: Add use-case guidance to Loggers page of admin console
* Issue 43686: Improve log rotation on labkey-errors.log to avoid filling disk

#### Changes
* Rotate labkey-errors.log at 100MB
* Retain last 3 files, plus ensure that we archive the error log with the first error logging from the current startup in a separate dated file
* Let code include a short note about a logger's usage
* Show the note in the Admin Console's Loggers page
* Add various logger notes

### DIFF
--- a/api/src/org/labkey/api/data/ConnectionWrapper.java
+++ b/api/src/org/labkey/api/data/ConnectionWrapper.java
@@ -31,6 +31,7 @@ import org.labkey.api.util.LoggerWriter;
 import org.labkey.api.util.MemTracker;
 import org.labkey.api.util.Pair;
 import org.labkey.api.util.SimpleLoggerWriter;
+import org.labkey.api.util.logging.LogHelper;
 import org.labkey.api.view.ViewServlet;
 
 import java.lang.reflect.Method;
@@ -68,7 +69,7 @@ import java.util.concurrent.atomic.AtomicLong;
  */
 public class ConnectionWrapper implements java.sql.Connection
 {
-    private static final Logger LOG = LogManager.getLogger(ConnectionWrapper.class);
+    private static final Logger LOG = LogHelper.getLogger(ConnectionWrapper.class, "All JDBC meta data and SQL execution calls being made");
     private static final Map<ConnectionWrapper, Pair<Thread, Throwable>> _openConnections = Collections.synchronizedMap(new IdentityHashMap<>());
     private static final Set<ConnectionWrapper> _loggedLeaks = new HashSet<>();
     private static final Logger _logDefault = LogManager.getLogger(ConnectionWrapper.class);
--- a/api/src/org/labkey/api/data/ConnectionWrapper.java
+++ b/api/src/org/labkey/api/data/ConnectionWrapper.java
@@ -69,7 +69,8 @@ import java.util.concurrent.atomic.AtomicLong;
  */
 public class ConnectionWrapper implements java.sql.Connection
 {
-    private static final Logger LOG = LogHelper.getLogger(ConnectionWrapper.class, "All JDBC meta data and SQL execution calls being made");
+    private static final Logger LOG = LogHelper.getLogger(ConnectionWrapper.class, "All JDBC metadata and SQL execution calls being made");
+
     private static final Map<ConnectionWrapper, Pair<Thread, Throwable>> _openConnections = Collections.synchronizedMap(new IdentityHashMap<>());
     private static final Set<ConnectionWrapper> _loggedLeaks = new HashSet<>();
     private static final Logger _logDefault = LogManager.getLogger(ConnectionWrapper.class);
--- a/api/src/org/labkey/api/data/Table.java
+++ b/api/src/org/labkey/api/data/Table.java
@@ -89,7 +89,7 @@ public class Table
     public static final int ERROR_DELETED = 10002;
     public static final int ERROR_TABLEDELETED = 10003;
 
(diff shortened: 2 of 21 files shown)


api/src/org/labkey/api/data/ConnectionWrapper.java
```diff
@@ -68,7 +69,7 @@
  */
 public class ConnectionWrapper implements java.sql.Connection
 {
-    private static final Logger LOG = LogManager.getLogger(ConnectionWrapper.class);
+    private static final Logger LOG = LogHelper.getLogger(ConnectionWrapper.class, "All JDBC meta data and SQL execution calls being made");
```
labkey-susanh:
```suggestion
    private static final Logger LOG = LogHelper.getLogger(ConnectionWrapper.class, "All JDBC metadata and SQL execution calls being made");

```

api/src/org/labkey/api/data/Table.java
```diff
@@ -89,7 +89,7 @@
     public static final int ERROR_DELETED = 10002;
     public static final int ERROR_TABLEDELETED = 10003;
 
-    private static final Logger _log = LogManager.getLogger(Table.class);
+    private static final Logger _log = LogHelper.getLogger(Table.class, "SQL generation and execution, some TableInfo-scoped DB scoped operations");
```
labkey-susanh: I find this difficult to parse.  Is the second "scoped" meant to be there?

api/src/org/labkey/api/dataiterator/LoggingDataIterator.java
```diff
@@ -34,7 +34,7 @@
  */
 public class LoggingDataIterator extends AbstractDataIterator implements ScrollableDataIterator, MapDataIterator
 {
-    static Logger _staticLog = LogManager.getLogger(LoggingDataIterator.class);
+    static Logger _staticLog = LogHelper.getLogger(LoggingDataIterator.class, "Shows transformations and mappings during many types of data imports and ETLs");
```
labkey-susanh:
```suggestion
    static Logger _staticLog = LogHelper.getLogger(LoggingDataIterator.class, "Transformations and mappings during many types of data imports and ETLs");
```
(Just for consistency.)

api/src/org/labkey/api/util/logging/ErrorLogRotator.java
```diff
@@ -0,0 +1,88 @@
...
+import java.nio.file.Path;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Helper class to retain labkey-errors.log file content if it's danger of rotating out and potentially losing
```
labkey-susanh:
```suggestion
 * Helper class to retain labkey-errors.log file content if it's in danger of rotating out and potentially losing
```

api/src/org/labkey/api/util/logging/ErrorLogRotator.java
```diff
@@ -0,0 +1,88 @@
...
+        // Narrow to just the error log files
+        paths = paths.stream().filter(p -> p.getPath().toString().contains("labkey-errors.log")).collect(Collectors.toList());
+
+        List<PathWithAttributes> result = new ArrayList<>();
+
+        // Look for the first file from the current set of logging to move it away instead of rotating it
+        PathWithAttributes logToRetain = null;
+        long bestTimeVersusStartup = Long.MAX_VALUE;
+
+        for (int i = MAX_RETAINED; i < paths.size(); i++)
```
labkey-susanh, line 55: Do we know the paths are sorted?  Is that why starting at `MAX_RETAINED` is ok?

api/src/org/labkey/api/util/logging/ErrorLogRotator.java
```diff
@@ -0,0 +1,88 @@
...
+
+            try
+            {
+                Files.move(logToRetain.getPath(), target);
+                // Don't need to mark it for deletion, as it's already been moved
+                result.remove(logToRetain);
+            }
+            catch (IOException e)
+            {
+                LOG.warn("Failed to retain labkey-errors.log file", e);
```
labkey-susanh: Do you want to use the actual name of the file here?

labkey-susanh (approved):

api/src/org/labkey/api/util/logging/ErrorLogRotator.java
labkey-jeckels: Yes. I've added a comment.

api/src/org/labkey/api/util/logging/ErrorLogRotator.java
labkey-jeckels: I've added it. I expect it'll be in the message from the exception too, which is why I didn't include it originally.

api/src/org/labkey/api/data/Table.java
labkey-jeckels: Nope!

PR #2624, opened 2021-09-16
Dataset "Delete All Rows" fix for confirm dialog so it doesn't truncate message text

cnathe:
#### Rationale
I noticed that for a manage dataset page the confirm dialog truncates the message in the display when you click the "Delete All Rows" button. This PR fixes that by making the dialog wider.

BEFORE
Screen Shot 2021-09-16 at 5 48 15 PM

AFTER
Screen Shot 2021-09-16 at 5 48 33 PM

#### Changes
* convert from Ext4.Msg.confirm() to Ext.Msg.show() which allows for control over the width of the dialog

### DIFF
--- a/study/src/org/labkey/study/view/datasetDetails.jsp
+++ b/study/src/org/labkey/study/view/datasetDetails.jsp
@@ -260,9 +260,17 @@ if (!pipelineSet)
                 %>msg +="<b>This will delete data in sub-folders that use this dataset.</b><br>";<%
         } %>
         msg += "This action cannot be undone and will result in an empty dataset.";
-        Ext4.Msg.confirm("Confirm Deletion", msg, function(button){
-            if (button === 'yes') {
-                truncate();
+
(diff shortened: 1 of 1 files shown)


cnathe:
@RosalineP I believe I recall you making a similar Ext4.Msg box change recently. Do you mind code reviewing this PR?

labkey-klum (approved):
Was this a general problem with the `confirm` function? This fix seems fine but wondering whether we might have add an ext patches fix if this crops up in other places.

cnathe:
> Was this a general problem with the `confirm` function? This fix seems fine but wondering whether we might have add an ext patches fix if this crops up in other places.

Good question. I think this is very similar to the issue Rosaline fixed recently with another "confirm" usage, so maybe this is something we should check to see if it is a wider issue. I'll take a look at usages. Thanks, Karl.
-Cory

PR #2629, opened 2021-09-20
Remove search tests from daily suite

labkey-tchad:
#### Rationale
Search tests are very sensitive to data left on the server by unrelated tests. We will run them in a more isolated environment on TeamCity to avoid such collisions.

#### Changes
* Remove `SearchTestDefault` and `DataClassSearchTest` from daily suite


labkey-danield (approved):
This looks ok to me. The failure in the MSSQL run is interesting, but not sure it is related to this change.

labkey-sweta (approved):

PR #2630, opened 2021-09-20
Dataset trigger script fix

labkey-klum:
#### Rationale
Since the refactor to support trigger scripts for bulk dataset actions, some EHR tests have been failing regularly. Josh has investigated a failure in the ONPRC_EHRTest2.testBirthStatusApi test and the cause of the failure is that the triggers are firing twice from a saveRows() update API call. The first is a beforeUpdate and the second is a beforeInsert.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43946

#### Changes
`AbstractQueryUpdateService.updateRows` invokes the abstract method `updateRow`. For datasets, as has always been the case, we eventually call into `StudyManager.importDatasetData`. The previous refactor consolidated all of the bulk update code to use the QUS so that the trigger script data iterator will get set up. The problem is that `AbstractQueryUpdateService.updateRows:666` fires the row triggers directly and then we see the triggers again getting fired from the `TriggerDataIterator`.

This is somewhat of an un-elegant solution that basically threads a setting into `DatasetDefinition.insertData` to indicate whether the row triggers have been fired and if so, just grab the dataset dataiterator and invoke it directly which is what the code used to do prior to the refactor.

Another approach that I thought about was to conditionally support the trigger scripts at a lower level in `_importRowsUsingDIB`, but would obviously have a larger footprint outside of datasets.

### DIFF
--- a/api/src/org/labkey/api/query/AbstractQueryUpdateService.java
+++ b/api/src/org/labkey/api/query/AbstractQueryUpdateService.java
@@ -300,18 +300,25 @@ public abstract class AbstractQueryUpdateService implements QueryUpdateService
             context.setDataSource((String) extraScriptContext.get(DataIteratorUtil.DATA_SOURCE));
         }
 
-        in = preTriggerDataIterator(in, context);
-
+        boolean skipTriggers = context.getConfigParameterBoolean(ConfigParameters.SkipTriggers);
         boolean hasTableScript = hasTableScript(container);
         TriggerDataBuilderHelper helper = new TriggerDataBuilderHelper(getQueryTable(), container, user, extraScriptContext, context.getInsertOption().useImportAliases);
-        if (hasTableScript)
-            in = helper.before(in);
+        if (!skipTriggers)
+        {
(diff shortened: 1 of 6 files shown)


labkey-jeckels (commented):
I'll defer to @labkey-matthewb on this, as I'm not sure what the preferred interaction is between QUS and DI these days.

labkey-matthewb:
Hmm, there's a lot I could say about how this code evolved like this.  But for now, I'll just say I think I prefer that the config parameter name reflects what it does.  In this case something like QueryUpdateService.ConfigParameters.SkipTriggers instead of DatasetUpdateService.Config.CalledFromQUS.  Then the skip triggers logic can be moved to where the triggers are actually setup and then you wouldn't need that ugly "if" in DatasetDefinition.insertData().

It would also be nice to avoid the deprecated call to StudyManager.importDatasetData(), if possible, but maybe that's for another time.

labkey-matthewb (approved):

PR #2633, opened 2021-09-21
Update JDK 17 JavaDoc link

labkey-adam:
#### Rationale
With JDK 17's release last week, the 17 JavaDocs have been published to a new location; need to update our base URL to match.

#### Changes
* Update JDK 17 JavaDocs base URL
* Remove obsolete comment (the method adds the `revision` parameter, so callers shouldn't)

### DIFF
--- a/api/src/org/labkey/api/util/DateUtil.java
+++ b/api/src/org/labkey/api/util/DateUtil.java
@@ -837,7 +837,7 @@ validNum:       {
     }
 
 
-    public static long parseDateTime(String s, MonthDayOption md)
+    private static long parseDateTime(String s, MonthDayOption md)
     {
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -837,7 +837,7 @@ public static long parseDateTime(Container c, String s)
     }
 
 
-    public static long parseDateTime(String s, MonthDayOption md)
+    private static long parseDateTime(String s, MonthDayOption md)
```
labkey-susanh, line 840: This change seems unrelated, but as long as the code is ok with it, so am I.

labkey-susanh (approved):

PR #2636, opened 2021-09-22
Item 9192: Initial modifications to support sample status

labkey-susanh:
#### Rationale
We are developing support for adding a status field to samples, which will determine which actions can be taken on the samples. Here we make changes to various APIs that will check the status of samples in order to determine if the action is permitted or not.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/639
* https://github.com/LabKey/inventory/pull/306
* https://github.com/LabKey/sampleManagement/pull/703
* https://github.com/LabKey/biologics/pull/1006
* https://github.com/LabKey/premium/pull/273
* https://github.com/LabKey/ehrModules/pull/238
* https://github.com/LabKey/snd/pull/100
* https://github.com/LabKey/premium/pull/273

#### Changes
* Add experimental feature flag for sample status tracking
* Rename `core.qcstate` to `core.dataStates` for more general use
* Add `sampleState` column to `exp.materials` table for tracking status
* Rename and refactor various QCState classes and methods to use DataState instead
* Update actions for deleting and modifying samples, uploading assay data, linking samples to studies and recalling from studies to check for status when applicable.


api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
```diff
@@ -1096,6 +1097,22 @@ else if (provenanceInputs instanceof Collection)
...
+                .collect(Collectors.toList());
+        if (!lockedSamples.isEmpty())
+        {
+            String message;
+            if (lockedSamples.size() == 1)
+                message = "Sample " + lockedSamples.get(0).getName() + " has status " + lockedSamples.get(0).getStateLabel() + ", which prevents";
+            else if (lockedSamples.size() <= 10)
+                message = "Samples " + lockedSamples.stream().map(ExpMaterial::getNameAndStatus).collect(Collectors.joining(", ")) + " have statuses that prevent";
+            else
+                message = lockedSamples.size() + " samples have statuses that prevent";
```
cnathe: After reviewing the other PRs in this set, I am starting to wonder if any sort of central helper methods for these two things might be helpful to reduce some of the "oh so close to duplicate" code: 1) a helper to take in a list of ExpMaterial objects and a SampleOperation and return the list where the operation is not permitted and 2) a helper that takes in a sample list and an operation and gives back the error message string with this one vs a few vs many check. Thoughts?

api/src/org/labkey/api/data/CoreSchema.java
```diff
@@ -153,9 +155,16 @@ public TableInfo getTableInfoNotifications()
         return getSchema().getTable("Notifications");
     }
 
+    // Prefer getTableInfoDataStates
+    @Deprecated (since = "21.10")
     public TableInfo getTableInfoQCState()
     {
-        return getSchema().getTable("QCState");
+        return getSchema().getTable("DataStates");
```
cnathe:
```suggestion
        return getTableInfoDataStates();
```

core/resources/schemas/core.xml
```diff
@@ -642,19 +642,22 @@
     </columns>
   </table>
 
-  <table tableName="QCState" tableDbType="TABLE">
-    <description>Contains one row per quality control state.</description>
+  <table tableName="DataStates" tableDbType="TABLE">
+    <description>Contains one row per data state.</description>
```
cnathe: could leave in the "example" usages if that would be helpful, something like:
```suggestion
    <description>Contains one row per data state (ex. quality control state, sample state).</description>
```

core/resources/schemas/dbscripts/postgresql/core-21.006-21.007.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE core.qcstate RENAME TO DataStates;
+
+ALTER TABLE core.DataStates ADD COLUMN StateType VARCHAR(20);
```
cnathe, line 3: in the core.xml file it says that "by default, the state will be a QC state". Is that determined just by the value being null/empty?

core/src/org/labkey/core/query/CoreQuerySchema.java
```diff
@@ -624,9 +627,24 @@ protected TableInfo getFilesTable()
...
     {
-        return new QCStateTableInfo(this);
+        return new DataStatesTableInfo(this);
+    }
+
+    public TableInfo getQCStatesTable()
+    {
+        TableInfo dataStatesTable = getDataStatesTable();
+        FilteredTable table = new FilteredTable<>(dataStatesTable, this);
+        SQLFragment sql = new SQLFragment("(stateType IS NULL)");
```
cnathe, line 639: nevermind my other question in the SQL upgrade script. here is my answer!

experiment/src/org/labkey/experiment/api/ExpMaterialImpl.java
```diff
@@ -177,6 +179,56 @@ public String getAliquotedFromLSID()
...
+        DataState state = getSampleState();
+        if (state == null)
+            return null;
+        return state.getLabel();
+    }
+
+    @Override
+    public boolean isOperationPermitted(ExperimentService.SampleOperations operation)
+    {
+        if (!SampleTypeService.isSampleStatusEnabled()) // permit everything if feature not enabled
```
cnathe, line 214: I could be wrong (it is all blurring together), but I thought I remembered a check in one of the other PRs which was doing this `isSampleStatusEnabled` check before calling `isOperationPermitted`. If so, I guess it isn't doing any hard, but just unnecessary since it is being done here.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -22,25 +22,7 @@
...
-import org.labkey.api.data.ImportAliasable;
-import org.labkey.api.data.JdbcType;
-import org.labkey.api.data.MutableColumnInfo;
-import org.labkey.api.data.PHI;
-import org.labkey.api.data.RenderContext;
-import org.labkey.api.data.SQLFragment;
-import org.labkey.api.data.Sort;
-import org.labkey.api.data.TableInfo;
-import org.labkey.api.data.UnionContainerFilter;
+import org.labkey.api.data.*;
```
cnathe: was this change to `*` intentional or just an IntelliJ change?

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -580,6 +578,22 @@ protected ContainerFilter getLookupContainerFilter()
             defaultCols.add(FieldKey.fromParts(ExpMaterialTable.Column.SampleSet));
 
         addColumn(ExpMaterialTable.Column.Flag);
+
+        if (SampleTypeService.isSampleStatusEnabled())
+        {
+            var statusColInfo = addColumn(ExpMaterialTable.Column.SampleState);
```
cnathe, line 603: I'm curious why some of these columns look to duplicate a bunch of these settings in the `createColumn` and the `populateColumns` methods?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -631,13 +632,23 @@ public ExpMaterialImpl getExpMaterial(int rowid)
...
         List<ExpMaterialImpl> materials = getExpMaterials(c, u, List.of(rowId), sampleType);
-        if (materials.isEmpty())
+        if (materials == null || materials.isEmpty())
             return null;
         if (materials.size() > 1)
             throw new IllegalArgumentException("Expected 0 or 1 samples, got: " + materials.size());
         return materials.get(0);
     }
 
+    public List<Integer> findIdsNotPermittedForOperation(List<ExpMaterialImpl> candidates, SampleOperations operation)
```
cnathe: oh, there is a helper here for this filtering of candidates. Any reason more usages don't call this instead of doing their own filtering?

study/src/org/labkey/study/controllers/publish/SampleTypePublishConfirmAction.java
```diff
@@ -206,6 +209,13 @@ protected ActionURL linkToStudy(SampleTypePublishConfirmForm form, Container tar
     {
         List<Map<String, Object>> dataMaps = new ArrayList<>();
         Map<Container, Set<Integer>> rowIdsByTargetContainer = new HashMap<>();
+        List<? extends ExpMaterial> samples = ExperimentService.get().getExpMaterials(dataKeys.keySet());
+        String unlinkablesMsg = samples.stream()
+                .filter(sample -> !sample.isOperationPermitted(ExperimentService.SampleOperations.LinkToStudy))
+                .map(ExpMaterial::getNameAndStatus)
+                .collect(Collectors.joining(", "));
+        if (!StringUtils.isEmpty(unlinkablesMsg))
+            errors.add("The following samples have a status that prevents linking to the study: " + unlinkablesMsg);
```
cnathe: here we don't do the <=10 check. This makes me think a few helpers for the "find me samples in my collection that are no permitted" and "get me the error message" might be helpful.

cnathe (approved):

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
labkey-susanh: Just IntelliJ doing things I did not intend it to. I've updated my settings to make this revert.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
labkey-susanh: I believe they are called from different code paths, but I also found it curious.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: This one is specifically wanting the RowIds, not the materials (since that's what we need for the delete confirmation response.  I imagine this will get refactored and reused with the next story that beefs up the confirmation API to include other operations.

experiment/src/org/labkey/experiment/api/ExpMaterialImpl.java
labkey-susanh: I'm leaving it in a few key places, to escape early and to cover various code paths.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -940,6 +954,42 @@ else if (o instanceof Number)
...
+
+        Set<String> lockedRelativeLsids = lockedRelatives.stream().map(ExpMaterial::getLSID).collect(Collectors.toSet());
+        Set<String> newRelativeLsids = currentSampleRelatives.stream().map(ExpMaterial::getLSID).collect(Collectors.toSet());
+        // check if all the locked relatives are still in the current list
+        Set<ExpMaterial> removedLocked = lockedRelatives.stream().filter(sample -> !newRelativeLsids.contains(sample.getLSID())).collect(Collectors.toSet());
+        if (!removedLocked.isEmpty())
+        {
+            String message = String.format("One or more existing %s of sample %s has a status that prevents the updating of lineage", relationPlural, sampleName);
+            if (removedLocked.size() <= 10)
+                message += ": " + removedLocked.stream().map(ExpMaterial::getNameAndStatus).collect(Collectors.joining(", "));
```
cnathe, line 972: Does it make sense for the message generation in this method to go through `getOperationNotPermittedMessage`?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -631,13 +632,23 @@ public ExpMaterialImpl getExpMaterial(int rowid)
...
         return materials.get(0);
     }
 
+    public List<Integer> findIdsNotPermittedForOperation(List<? extends ExpMaterial> candidates, SampleTypeService.SampleOperations operation)
+    {
+        if (!SampleTypeService.isSampleStatusEnabled())
+            return Collections.emptyList();
+
+        return SampleTypeService.get().getSamplesNotPermitted(candidates, operation)
+                .stream().filter(material -> !material.isOperationPermitted(operation))
```
cnathe: isn't `getSamplesNotPermitted` doing the filtering for `isOperationPermitted`? curious why this is streaming and filtering again.

cnathe (approved):

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: I chose not to because of the variability in the prefix (parents or children).

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Oops.  Yes.

PR #2639, opened 2021-09-22
Add option to hide SSO login links from header area

bbimber:
This is related to:

https://www.labkey.org/ONPRC/Support%20Tickets/issues-details.view?issueId=43965
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43979

The proposal here is to create one global setting that toggles whether header icons are shown for SSO providers or not.

### DIFF
--- a/api/src/org/labkey/api/security/AuthenticationManager.java
+++ b/api/src/org/labkey/api/security/AuthenticationManager.java
@@ -197,7 +197,7 @@ public class AuthenticationManager
             .forEach(cp-> {
                 switch(cp.getName())
                 {
-                    case SELF_REGISTRATION_KEY, AUTO_CREATE_ACCOUNTS_KEY, SELF_SERVICE_EMAIL_CHANGES_KEY -> saveAuthSetting(null, cp.getName(), Boolean.parseBoolean(cp.getValue()));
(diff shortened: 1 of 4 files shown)


bbimber:
@labkey-adam @labkey-jeckels: I apparently cant select a reviewer, but what do you think of this proposal for the auth header link? This should have no change on existing systems, unless the admin opt-in to disabling the header links.

labkey-adam:
See latest comment on #43979; issue has been estimated and priority will be reviewed relative to other requested enhancements. We're leaning toward the approach of suppressing the link entirely if no logo has been specified.

bbimber:
@labkey-adam OK. I dont have particularly strong preferences on the specific implementation. As far as ensuring something along these lines gets into the next major release: can I do anything to advance that? I appreciate there is more work to a feature than the code, such as updating docs, etc. I'm not in a great position to write docs, but can I submit a PR that conforms to your ideas on the feature, update test cases, etc., that would get this into the next release?

labkey-adam:
The issue was reviewed and prioritized for 21.11 (next major release), so I don't think you need to do anything, @bbimber . The "hide if logo not present" part is easy and we've already discussed a couple implementation approaches. The upgrade code to fill in missing logos should be straightforward as well... especially since I wrote all of the SSO infrastructure.

bbimber:
@labkey-adam OK, I'm perfectly fine doing nothing...

PR #2646, opened 2021-09-27
Remove isStatic() requirement from upgrade code filter

labkey-adam:
#### Rationale
The upgrade code action was blissfully unaware of a bunch of upgrade code methods because it filtered out all non-static methods. Turns out upgrade code can be implemented as either static or instance methods.


labkey-jeckels (approved):
Thanks. I wondered if there was a need for it to be static earlier but failed to followup.

PR #487, opened 2019-10-01
Issue 38450: Sample Set Lookup 

RosalineP:
Hi Josh!
I took your advice in [this](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=38450) issue -- this fix is working, and after some tweaks everything is running green in TC.

PR #2647, opened 2021-09-27
Issue 43572: Display fields on lookups to issues

RosalineP:
#### Rationale
Previously, after creating a lookup field to issueslistdef with a DisplayColumnName of issueid (or, in general, creating a lookup field where the DisplayColumnName is the PK), upon attempting to insert a new row into the data structure or update a row within it, the dropdown on the lookup field did not show the DisplayColumnName. 
Now, in the above example, when updating a row or inserting a row, the dropdown correctly shows text from issueid, not the title column.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/487

#### Changes
* Catch case in which display column is the same as PK column


labkey-matthewb (approved):

labkey-klum (approved):
that's simple!

PR #2648, opened 2021-09-27
Issue 43844: Better handling for script running failures

labkey-adam:
#### Rationale
Holding static lists of scripts makes iterative development of drop/create scripts impossible. Switch to constructing instances of `SqlScriptRunner` as needed.

### DIFF
--- a/api/src/org/labkey/api/data/SqlScriptRunner.java
+++ b/api/src/org/labkey/api/data/SqlScriptRunner.java
@@ -22,7 +22,6 @@ import org.jetbrains.annotations.Nullable;
 import org.labkey.api.cache.CacheManager;
 import org.labkey.api.module.Module;
 import org.labkey.api.module.ModuleLoader;
-import org.labkey.api.security.User;
 import org.labkey.api.util.logging.LogHelper;
 
 import java.util.ArrayList;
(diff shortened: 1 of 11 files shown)


api/src/org/labkey/api/data/SqlScriptRunner.java
```diff
@@ -38,12 +38,29 @@
...
+    }
+
+    public SqlScriptRunner(@Nullable User user)
+    {
+        _user = user;
+    }
+
+    private User getUser()
+    {
+        return null != _user ? _user : ModuleLoader.getInstance().getUpgradeUser();
```
labkey-jeckels: Is this @NotNull?

labkey-jeckels (approved):

api/src/org/labkey/api/data/SqlScriptRunner.java
labkey-adam: Not guaranteed... upgrade user can't be constructed and set until after core is bootstrapped/upgraded. (Construction code path touches core tables like groups and principals.)

labkey-adam: Thinking about it further, I got rid of the "upgrade user" entirely, since it's an antiquated concept.

PR #2650, opened 2021-09-29
Time out when acquiring locks during transaction setup instead of deadlocking

labkey-jeckels:
#### Rationale
Deadlocking the server is really bad. Similar to how we switched to timing out with an exception when DB connections aren't available from the pool, we can time out when trying to acquire (most) lock objects passed to DbScope.ensureTransaction()

#### Changes
* Time out after 2 minutes of failing to get a lock
* Unwind already acquired locks when one fails
* New test coverage
* Enable test that now seems to pass
* More descriptions for loggers

### DIFF
--- a/api/src/org/labkey/api/cache/BlockingCache.java
+++ b/api/src/org/labkey/api/cache/BlockingCache.java
@@ -20,6 +20,7 @@ import org.jetbrains.annotations.Nullable;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.labkey.api.util.DeadlockPreventingException;
 import org.labkey.api.util.Filter;
 
 import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


labkey-adam (changes requested):
Looks like a connection leak in `testLockException()`

labkey-jeckels:
@labkey-adam Indeed. Fixed with my latest commit. I'll kick off the dailies to make sure this doesn't introduce other subtle problems.

labkey-jeckels:
https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/1553901

labkey-adam (approved):

labkey-jeckels:
@labkey-adam I've made a number of followup changes here but they're not as exciting as the original change. Up to you if you want to review again. I'll merge tomorrow if I don't hear objections.

labkey-adam:
Seem fine to me

PR #2651, opened 2021-09-29
Issue 43522: MetadataUnavailableException for no such column

ankurjuneja:
#### Rationale
Don't blow up when editing the metadata of a table/query for the columns that get added through TableCustomizer.

### DIFF
--- a/query/src/org/labkey/query/MetadataTableJSON.java
+++ b/query/src/org/labkey/query/MetadataTableJSON.java
@@ -72,7 +72,7 @@ public class MetadataTableJSON extends GWTDomain<MetadataColumnJSON>
     private boolean _userDefinedQuery;
     /** If metadata is not stored in the current container, the folder path where it is stored */
     private String _definitionFolder;
-    private static final Logger log = LogHelper.getLogger(MetadataTableJSON.class, "Table metadata");
+    private static final Logger log = LogHelper.getLogger(MetadataTableJSON.class, "Visual editor support for table/query metadata");
 
     @Override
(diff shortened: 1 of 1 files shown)


query/src/org/labkey/query/MetadataTableJSON.java
```diff
@@ -69,6 +72,7 @@
     private boolean _userDefinedQuery;
     /** If metadata is not stored in the current container, the folder path where it is stored */
     private String _definitionFolder;
+    private static final Logger log = LogHelper.getLogger(MetadataTableJSON.class, "Table metadata");
```
labkey-jeckels:
```suggestion
    private static final Logger log = LogHelper.getLogger(MetadataTableJSON.class, "Visual editor support for table/query metadata");
```

query/src/org/labkey/query/MetadataTableJSON.java
```diff
@@ -178,7 +182,7 @@ public MetadataTableJSON saveMetadata(String schemaName, User user, Container co
                 }
                 else
                 {
-                    throw new MetadataUnavailableException("No such column: " + metadataColumnJSON.getName());
+                    log.info("No such column: " + metadataColumnJSON.getName());
```
labkey-jeckels: Please also log the schema/query and container to give context

labkey-jeckels (approved):
Small suggestions, but no need for additional review

query/src/org/labkey/query/MetadataTableJSON.java
ankurjuneja: done

PR #1851, opened 2021-01-05
Issue 7580: inconsistent use of returnUrl parameter

labkey-kevink:
#### Rationale
Remove last vestiges of 'srcURL' and 'returnURL' in favor of the 'returnUrl' parameter.  While testing the PR on TeamCity, throw an exception to find any other usages.  We'll switch this to only throw in devmode before merging.

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/83
* https://github.com/LabKey/labkey-ui-components/pull/426
* https://github.com/LabKey/premium/pull/225

#### Changes
* log an ERROR and throw an exception when using 'returnURL' or 'srcURL' instead of 'returnUrl'
* switch last usages of 'returnURL' and 'srcURL' to 'returnUrl'
* change usages of `ReturnUrlForm.getReturnUrl` string to use `.getReturnActionURL` instead
* use ActionURL.Param.returnUrl enum in java and jsp code
* use LABKEY.ActionURL.getReturnUrl() in client-side code

PR #2652, opened 2021-09-29
Issue 44023: Query Report created through Clinical and Assay tab of study appends comma in URL

labkey-martyp:
#### Rationale
Two returnURL parameters (one blank) being passed as parameters from this jsp. Binding both puts a comma in the returnURL before the blank returnURL. There has been some refactoring on this page since the original hidden returnURL field was added (2013). I don't see a use case where it is needed any longer.  I've run a lot of tests and no use cases are showing up.

#### Changes
* Remove hidden field

### DIFF
--- a/query/src/org/labkey/query/reports/view/createQueryReport.jsp
+++ b/query/src/org/labkey/query/reports/view/createQueryReport.jsp
@@ -113,7 +113,7 @@
                 description : true,
                 shared      : true
             },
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
This param used to be called srcURL until this PR:

https://github.com/LabKey/platform/pull/1851

I don't see anywhere that the hidden param was being set with either the old or new name, so the change looks fine to me.

PR #2655, opened 2021-09-30
Handle `InvalidPathException` in `PipeRootImpl.resolveToNioPath`

labkey-tchad:
#### Rationale
Should prevent errors such as this one triggered by the crawler:
```
java.nio.file.InvalidPathException: Illegal char <>> at index 68: D:\teamcity\work\c86b103d373f4c06\server\testAutomation\data\flow\-->">'>'"<\script><img src="xss" onerror="alert('8(')">
	[...]
	at org.labkey.pipeline.api.PipeRootImpl.resolveToNioPath(PipeRootImpl.java:377) ~[pipeline-21.10-SNAPSHOT.jar:?]
	at org.labkey.api.pipeline.browse.PipelinePathForm.getValidatedPaths(PipelinePathForm.java:155) ~[api-21.10-SNAPSHOT.jar:?]
	at org.labkey.api.pipeline.browse.PipelinePathForm.getValidatedSinglePath(PipelinePathForm.java:223) ~[api-21.10-SNAPSHOT.jar:?]
	at org.labkey.api.pipeline.browse.PipelinePathForm.getValidatedSingleFile(PipelinePathForm.java:217) ~[api-21.10-SNAPSHOT.jar:?]
	at org.labkey.flow.controllers.executescript.AnalysisScriptController$ImportAnalysisFromPipelineAction.getView(AnalysisScriptController.java:524) ~[flow-21.10-SNAPSHOT.jar:?]
	at org.labkey.flow.controllers.executescript.AnalysisScriptController$ImportAnalysisFromPipelineAction.getView(AnalysisScriptController.java:518) ~[flow-21.10-SNAPSHOT.jar:?]
```

#### Related Pull Requests
* #2541 

#### Changes
* Handle `InvalidPathException` in `PipeRootImpl.resolveToNioPath`

### DIFF
--- a/pipeline/src/org/labkey/pipeline/api/PipeRootImpl.java
+++ b/pipeline/src/org/labkey/pipeline/api/PipeRootImpl.java
@@ -356,7 +356,7 @@ public class PipeRootImpl implements PipeRoot
     @Nullable
     public Path resolveToNioPath(String path)
     {
-        if (null == path)
+        if (StringUtils.isBlank(path))
             throw new NotFoundException("Must specify a file path");
 
         try
--- a/pipeline/src/org/labkey/pipeline/api/PipeRootImpl.java
+++ b/pipeline/src/org/labkey/pipeline/api/PipeRootImpl.java
@@ -47,6 +47,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
+import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


labkey-jeckels (commented):
This seems like a reasonable change, but we could also return null as we do in other codepaths here. What do you think, Ian?

pipeline/src/org/labkey/pipeline/api/PipeRootImpl.java
```diff
@@ -358,23 +359,30 @@ public Path resolveToNioPath(String path)
         if (null == path)
```
Sigmonia: This should probably be StringUtils.isBlank or similar, seems like both null and "" would be equally ambiguous...though the former will cause an error

pipeline/src/org/labkey/pipeline/api/PipeRootImpl.java
```diff
@@ -358,23 +359,30 @@ public Path resolveToNioPath(String path)
...
+                File file = resolvePath(path);
+                return null != file ? file.toPath() : null;
+            }
         }
-        else
+        catch (InvalidPathException e)
         {
-            File  file = resolvePath(path);
-            return null != file ? file.toPath() : null;
+            throw new NotFoundException("Must specify a valid file path", e);
```
Sigmonia, line 385: Responding to Josh's comment:  I think throwing the error here is better as returning `null` (at least in this case) doesn't get us much, as the same error is just thrown one frame up after a null check--we also lose most of the context of the initiating error's cause.

Sigmonia (approved):

PR #2661, opened 2021-10-02
Issue 43898: Study dataset with subject column name set to "SubjectID" has issues when a field of that same name is also added to the dataset

labkey-martyp:
#### Rationale
Don't allow a dataset to have a user defined field that matches the Study Subject Column in study properties. Validated when creating/updating dataset designs and when updating study properties.  Applied to all dataset types.

#### Changes
* Update reserved names for datasets
* Validation on ManageStudyPropertiesAction
* Add automated test

### DIFF
--- a/study/src/org/labkey/study/controllers/StudyController.java
+++ b/study/src/org/labkey/study/controllers/StudyController.java
@@ -69,6 +69,7 @@ import org.labkey.api.exp.PropertyDescriptor;
 import org.labkey.api.exp.api.ExpProtocol;
 import org.labkey.api.exp.api.ExpSampleType;
 import org.labkey.api.exp.property.Domain;
+import org.labkey.api.exp.property.DomainProperty;
 import org.labkey.api.gwt.client.AuditBehaviorType;
 import org.labkey.api.module.ModuleHtmlView;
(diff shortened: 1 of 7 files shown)


study/src/org/labkey/study/model/DatasetDomainKind.java
```diff
@@ -263,6 +263,21 @@ DatasetDefinition getDatasetDefinition(String domainURI)
         return StudyManager.getInstance().getDatasetDefinition(domainURI);
     }
 
+    protected Set<String> getStudySubjectReservedName(Domain domain)
```
cnathe, line 267: suggest adding a comment indicating the issue number for reference

study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -1564,6 +1565,35 @@ public void addNavTrail(NavTree root)
             root.addChild("Study Properties");
         }
 
+        @Override
+        public void validateForm(TableViewForm form, Errors errors)
+        {
+            // Validate that the subject column name is not a user defined field in one of the datasets
```
cnathe: suggest adding a comment indicating the issue number for reference

cnathe (approved):

PR #2663, opened 2021-10-04
Enable using File Watchers for cloud based Pipeline Roots

Sigmonia:
#### Rationale
https://docs.google.com/document/d/1adCI1KteCV-vNuCJ6Okz1lVcsO6AgZIKXZeSSg2cKQ0/edit?pli=1#

#### Related Pull Requests
* https://github.com/LabKey/cloud/pull/88
* https://github.com/LabKey/commonAssays/pull/373
* https://github.com/LabKey/MaxQuant/pull/40
* https://github.com/LabKey/premium/pull/276
* https://github.com/LabKey/targetedms/pull/435
* https://github.com/LabKey/nlp/pull/199
* https://github.com/LabKey/testAutomation/pull/883

#### Changes
* Lots of File --> nio.Path refactors
* Additional Cloud api interfaces
* new CloudWatcherThread to go along with the FileSystemWatcherImpl


api/src/org/labkey/api/pipeline/PipelineJob.java
```diff
@@ -433,7 +433,6 @@ public PipeRoot getPipeRoot()
     public void setLogFilePath(Path logFile)
     {
         _logFilePathName = FileUtil.pathToString(logFile);
-//        _logJobStopStart.info("LogFilePath set to " + _logFilePathName);
```
labkey-klum, line 436: delete?

api/src/org/labkey/api/pipeline/PipelineProvider.java
```diff
@@ -133,19 +151,20 @@ public boolean accept(File f, boolean checkSiblings)
...
-                                    {
-                                        return false;
-                                    }
-                                }
+                                        ft.isType(sibling.getFileName().toString()) &&
+                                        ft.getIndexMatch(sibling) < indexMatch);
+                            }
+                            catch (IOException e)
+                            {
                             }
```
labkey-klum, line 173: should we be throwing an exception or at the least log an error?

api/src/org/labkey/api/files/FileSystemWatcherImpl.java
```diff
@@ -314,6 +319,68 @@ private void close()
...
+                {
+                        CloudStoreService.get().executeWatchJob(cloudWatcherJob);
+                });
+            }
+            catch (Throwable e)  // Make sure throwables don't kill the background thread
+            {
+                ExceptionUtil.logExceptionToMothership(null, e);
+            }
+
+            Thread.sleep(10000);
```
labkey-klum: Similar to the `FileSystemWatcherThread` would it be possible to move from looping/waiting every 10 seconds? I don't know whether we can get notified by the SQS service and placing some object in a blocking queue.

labkey-klum (approved):
see inline comments

api/src/org/labkey/api/pipeline/PipelineJob.java
Sigmonia: deleted.

api/src/org/labkey/api/pipeline/PipelineProvider.java
Sigmonia: Added Logging and returning false, as I think it better matches the previous behavior.

api/src/org/labkey/api/files/FileSystemWatcherImpl.java
Sigmonia: This is the difference between SQS and SNS. SQS has no notification system, only polling I did switch it to an explicit timer task. It is now using the same period constant as the pollingQueue.

PR #2667, opened 2021-10-05
Item 9533: App support for module defined sample assay results view configs to show in sample assays tabbed grid

cnathe:
#### Rationale
We want to allow for non-standard assays (i.e. those defined in customModules but not part of the assay designer infrastructure) to have a chance to participate in the Sample > Assays tabbed grid view within the LKB and LKSM apps. This PR adds some registries / services to allow for modules to register these SampleAssayResultsConfig objects. It also has a registry so that the LKB and LKSM apps can provider overrides for ExperimentUrl interface methods as a way for ActionURLs within LKS to opt in to generating URL that will bring the user into the relevant application based on a given context (i.e. for a sample type or for a sample Id).

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/649
* https://github.com/LabKey/platform/pull/2667
* https://github.com/LabKey/targetedms/pull/436
* https://github.com/LabKey/sampleManagement/pull/719
* https://github.com/LabKey/biologics/pull/1019
* https://github.com/LabKey/dataintegration/pull/129

#### Changes
* Add UrlProviderService to allow modules to register their UrlProvider interface overrides for generating app-based ActionURLs
* Add UrlProviderOverrideHandler to check for UrlProvider overrides based on the given interface method being invoked
* Add SampleAssayResultsService to allow for modules to register a query/view and related config information to be used within the app to show sample related assay results for that module
* Update ExperimentUrls interface to provide default "return null" implementations for all methods
* Update Dataset getSourceActionURL method to use new URLProvider overrides check

### DIFF
--- a/api/src/org/labkey/api/action/UrlProviderService.java
+++ b/api/src/org/labkey/api/action/UrlProviderService.java
@@ -28,7 +28,7 @@ public class UrlProviderService
 {
     private static final UrlProviderService instance = new UrlProviderService();
     private static final Map<Class, Class<? extends UrlProvider>> _urlProviderToImpl = new HashMap<>();
-    private static final Map<Class<? extends UrlProvider>, List<Pair<Module, Class<? extends UrlProvider>>>> _urlProviderToOverrideImpls = new HashMap<>();
+    private static final Map<Class<? extends UrlProvider>, List<Pair<Module, UrlProvider>>> _urlProviderToOverrideImpls = new HashMap<>();
 
     public static UrlProviderService getInstance()
     {
@@ -75,9 +75,9 @@ public class UrlProviderService
      * @param impl the override URLProvider implementation class
      * @param module the module providing the override
      */
-    public void registerUrlProviderOverride(Class<? extends UrlProvider> inter, Class<? extends UrlProvider> impl, Module module)
+    public void registerUrlProviderOverride(Class<? extends UrlProvider> inter, UrlProvider impl, Module module)
     {
-        List<Pair<Module, Class<? extends UrlProvider>>> impls = new ArrayList<>();
+        List<Pair<Module, UrlProvider>> impls = new ArrayList<>();
         if (_urlProviderToOverrideImpls.containsKey(inter))
             impls = _urlProviderToOverrideImpls.get(inter);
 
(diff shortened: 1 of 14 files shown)


api/src/org/labkey/api/action/UrlProviderService.java
```diff
@@ -0,0 +1,92 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class UrlProviderService
+{
+    private static final UrlProviderService instance = new UrlProviderService();
+    private static final Map<Class, Class<? extends UrlProvider>> _urlProviderToImpl = new HashMap<>();
```
labkey-jeckels, line 30: Should we stash UrlProvider instances instead of their implementation classes? They should be stateless (I hope, at least), so this would slightly improve performance and also make sure that if there's a problem, it's reported when it's being registered instead of every time it's attempted to be used.

api/src/org/labkey/api/action/UrlProviderService.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Register an implementation class to use for overrides to a URLProvider interface.
+     * @param inter the URLProvider interface
+     * @param impl the override URLProvider implementation class
+     * @param module the module providing the override
+     */
+    public void registerUrlProviderOverride(Class<? extends UrlProvider> inter, Class<? extends UrlProvider> impl, Module module)
+    {
+        List<Pair<Module, Class<? extends UrlProvider>>> impls = new ArrayList<>();
```
labkey-jeckels: I believe this approach means that the UrlProvider override is responsible for deciding if it should be enabled in a given container, correct? Probably worth making that explicit in the docs here

api/src/org/labkey/api/assay/sample/SampleAssayResultsConfig.java
```diff
@@ -0,0 +1,93 @@
+package org.labkey.api.assay.sample;
+
+import org.jetbrains.annotations.Nullable;
+import org.json.JSONObject;
+import org.labkey.api.data.ContainerFilter;
+
+public class SampleAssayResultsConfig
```
labkey-jeckels, line 12: A comment about the use for this class?

labkey-jeckels (approved):
Some suggestions but looking good overall.

api/src/org/labkey/api/action/UrlProviderService.java
cnathe: Good call. Done.

api/src/org/labkey/api/action/UrlProviderService.java
cnathe: That is correct. I went back on for on where that "is module enabled in the container should go". If you have other suggestions on that let me know. I have added a comment here regarding the current implementation.

api/src/org/labkey/api/assay/sample/SampleAssayResultsConfig.java
cnathe: done

PR #2671, opened 2021-10-06
Issue 44079: Perf regression with large domain create/update operations

labkey-jeckels:
#### Rationale
We're accidentally doing too much work when persisting exp.PropertyDescriptors. We're also being inefficient in how we manage large collections of commit tasks. This regressed as part of https://github.com/LabKey/platform/pull/2295

#### Changes
* Annotate as @Transient PropertyDescriptors.isVocabulary() and other getter methods that take significant work but don't correspond to columns on the exp.PropertyDescriptor table
* Use maps instead of sets to track commit tasks, making it fast to find an existing matching task

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -1827,7 +1827,7 @@ public class DbScope
         PRECOMMIT
         {
             @Override
-            protected Set<Runnable> getRunnables(TransactionImpl transaction)
+            protected Map<Runnable, Runnable> getRunnables(TransactionImpl transaction)
             {
                 return transaction._preCommitTasks;
             }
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/data/DbScope.java
```diff
@@ -1884,21 +1884,18 @@ public void run(TransactionImpl transaction)
 
         public <T extends Runnable> T add(TransactionImpl transaction, T task)
         {
-            T addedObj = task;
-            boolean added = getRunnables(transaction).add(task);
-            for(Runnable r : getRunnables(transaction))
+            Map<Runnable, Runnable> runnables = getRunnables(transaction);
+            T result = (T)runnables.get(task);
```
labkey-matthewb: runnables.putIfAbsent()

labkey-matthewb (approved):

PR #2672, opened 2021-10-07
Issue 43464: Session timeout dialog updates

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43464
- add back close button/option for modal in the non-compliance case
- add modal body content to give user option to "log in via another browser tab"

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2672
* https://github.com/LabKey/compliance/pull/139

#### Changes
* WebSocket.js updates to allow modal close for non-compliant case and update modal body text/link
* WebSocket.js updates to re-establish WebSocket after each close event, not just the first

### DIFF
--- a/api/webapp/clientapi/dom/WebSocket.js
+++ b/api/webapp/clientapi/dom/WebSocket.js
@@ -52,6 +52,10 @@ LABKEY.WebSocket = new function ()
 
     function websocketOnClose(evt) {
         if (evt.wasClean) {
+            var modalContent = 'Please reload the page or '
+                + '<a href="login-login.view?" target="_blank" rel="noopener noreferrer">'
+                + 'log in via another browser tab</a> to continue.'
+
             // first chance at handling the event goes to any registered callback listeners
             if (_callbacks[evt.code]) {
                 _callbacks[evt.code].forEach(function(cb){cb(evt)});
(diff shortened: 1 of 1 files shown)


api/webapp/clientapi/dom/WebSocket.js
```diff
@@ -173,9 +180,7 @@ LABKEY.WebSocket = new function ()
                 $('#lk-websocket-reload').on('click', function() {
                     window.location.reload();
                 });
-
-                openWebsocket(); // re-establish the websocket connection for the new guest user session
-            }, null, true, true);
+            }, null, true, isSessionInvalidBackgroundHideEnabled());
```
labkey-susanh, line 183: Wow.  That function name is a real mouthful.

api/webapp/clientapi/dom/WebSocket.js
```diff
@@ -135,6 +138,8 @@ LABKEY.WebSocket = new function ()
     }
 
     function hideModal() {
+        openWebsocket(); // re-establish the websocket connection for the new user session
+
         toggleBackgroundVisible(false);
```
labkey-susanh: This isn't part of this PR, but it's confusing to me that this call to `toggleBackgroundVisible` with the argument `false` results in the background being visible (not blurred).  Would updating the name to `toggleBackgroundBlurred` make sense?

labkey-susanh (approved):

api/webapp/clientapi/dom/WebSocket.js
cnathe: Good point. Done.

PR #2681, opened 2021-10-11
Add Workflow Task column to Exp Run table

labkey-alan:
#### Rationale
We need a Workflow Task column (FK to Exp ProtocolApplication) on the Exp Run table in order to associate Workflow Tasks with Assay Runs.

#### Related Pull Requests
* n/a

#### Changes
* DB Scripts to add WorkflowTask column to exp.ExperimentRun
* Add WorkflowTask column to ExpRunTableImpl
* Add getter/setter for workflowTaskId to ExperimentRun
* Add getter/setter for WorkflowTask to ExpRunImpl
* Roundtrip WorkflowTask during export/import

### DIFF
--- a/experiment/src/org/labkey/experiment/XarReader.java
+++ b/experiment/src/org/labkey/experiment/XarReader.java
@@ -1070,6 +1070,8 @@ public class XarReader extends AbstractXarImporter
                     {
                         throw new ExperimentException(e);
                     }
+                } else {
+                    getLog().warn("Could not find ProtocolApplication with LSID containing: " + lsidPart);
                 }
             }
         }
--- a/experiment/src/org/labkey/experiment/XarReader.java
+++ b/experiment/src/org/labkey/experiment/XarReader.java
@@ -58,6 +58,7 @@ import org.labkey.api.exp.api.ExpExperiment;
 import org.labkey.api.exp.api.ExpMaterial;
 import org.labkey.api.exp.api.ExpMaterialProtocolInput;
 import org.labkey.api.exp.api.ExpProtocol;
+import org.labkey.api.exp.api.ExpProtocolApplication;
 import org.labkey.api.exp.api.ExpProtocolInput;
 import org.labkey.api.exp.api.ExpProtocolInputCriteria;
 import org.labkey.api.exp.api.ExpRun;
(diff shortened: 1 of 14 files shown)


experiment/src/org/labkey/experiment/XarReader.java
```diff
@@ -1024,6 +1042,38 @@ private void loadExperimentRun(ExperimentRunType a, List<ExpMaterial> startingMa
...
+     */
+    private void saveRunWorkflowTaskIds() throws ExperimentException
+    {
+        for (ExpRun run : _loadedRuns)
+        {
+            String lsidPart = _runWorkflowTaskMap.get(run.getRowId());
+
+            if (lsidPart != null)
+            {
+                ProtocolApplication protocolApplication = _loadedProtocolApplications.stream()
```
labkey-jeckels: Instead of stashing the protocol applications as we load them, I think we should query the DB for a match. It's possible that the runs were exported in multiple XARs, so we may not have imported everything that's available.

I'd also suggest blowing up if we find multiple matches. I'm worried that not all protocol applications will use a GUID as the ObjectId in their LSID and it's possible we'd have an ambiguous reference.

experiment/src/org/labkey/experiment/XarReader.java
```diff
@@ -1024,6 +1042,38 @@ private void loadExperimentRun(ExperimentRunType a, List<ExpMaterial> startingMa
...
+                        .filter(p-> p.getLSID().endsWith(lsidPart)).findFirst()
+                        .orElse(null);
+
+                if (protocolApplication != null)
```
labkey-jeckels: I suggest logging a warning if we didn't find a match, but not failing the import

labkey-jeckels (changes requested):
A couple of suggestions for how we resolve the matching ProtocolApplication.

experiment/src/org/labkey/experiment/XarReader.java
labkey-alan: I tried to go the route of querying, but could not find an obvious way to do it. There is no container column on the ProtocolApplication table, so I am not able to filter to the current container, resulting in my queries returning results for the container I exported from. Suggestions welcome.

experiment/src/org/labkey/experiment/XarReader.java
labkey-alan: Done.

experiment/src/org/labkey/experiment/XarReader.java
labkey-jeckels: Pull the latest in this FB. I added a new ExperimentService method, getExpProtocolApplicationsByObjectId(), that should help. We have to join to the run table to get the container. It's completely untested so adjust if needed, but should be very close.

labkey-alan: Your change works. I am concerned we may be missing some imports in the case that runs are imported over multiple XARs. If a folder can be exported across multiple XARs wouldn't it be possible that I could import a Run that references a ProtocolApplication that is in a different XAR, thus we'd not be able to wire up the FK reference, and we wouldn't have it when importing the second XAR because that reference is only stored on the Run. Does that make sense?

labkey-jeckels: As long as the run that has the referenced workflow task (aka protocol application) has already been imported (either via the current XAR or a previously imported XAR) it should resolve with this approach (unlike your initial implementation, which would only resolve if it was in the same XAR).

If you haven't already imported the referenced run, you're correct, it will never get wired up (even if it's later imported). If that's an important sequence to handle, we'd have to somewhere stash the unmatched values and try to resolve them during all future imports. I'd guess that's not worth the effort, but don't really know the expected pattern of export/import either.

labkey-alan: I spoke with Karl about this and it is unlikely to happen, and would not be worth the effort to make it work at this time.

labkey-jeckels (changes requested):
The updated code looks good.

It looks like you're not yet handling delete. I suspect you'll get a SQLException if someone tries to delete a run that's referenced via a workflow task. Should we null out the FK reference here or disallow the delete?

labkey-alan:
I had the constraint set the value to NULL on delete. I don't believe we'll want to prevent users from deleting tasks, even if they are referenced by an assay.

labkey-jeckels (approved):
Got it - I didn't scroll far enough to the right to see that was part of the script.

I'll mark this as approved for now so you're not blocked by my previous request, but feel free to request additional review if you make further edits.

PR #2682, opened 2021-10-11
Unique names for data classes and sample types

labkey-klum:
#### Rationale
For this [story ](https://docs.google.com/document/d/1N6xfULI198pbU7xYdijYi4jmEg8n1a9s5PcPc5Tb4w0/edit) we want to support an administrative ability to set an optional prefix for data classes and sample types on a per container basis. 

We also will want to give administrators the ability to disable manually inserting/updating a name for sample types and data classes. This should make is so the name field is removed from all insert/update forms for LKS, LKB, LKSM and raise an error if a name comes in through an API endpoint.

#### Related Pull Requests
- https://github.com/LabKey/sampleManagement/pull/723
- https://github.com/LabKey/labkey-ui-components/pull/652
- https://github.com/LabKey/biologics/pull/1023

#### Changes
- Define a `NameExpressionOptionService` with a default NoOp implementation. An actual implementation will be provided in the `SampleManagementModule`
- If users are not allowed to manually insert names
  - Remove the name field for insert/update views
  - Add validation in the name expression `DataIterator` to fail if detected
  - Disallow removing the name expression from the domain editors
  - For samples, if the `UpdateExisting` option is specified, fail only if a specified name doesn't exist in the database

### DIFF
--- a/api/src/org/labkey/api/dataiterator/NameExpressionDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/NameExpressionDataIterator.java
@@ -39,6 +39,7 @@ public class NameExpressionDataIterator extends WrapperDataIterator
     private Container _container;
     private Function<String, Long> _getNonConflictCountFn;
     private String _counterSeqPrefix;
+    private boolean _allowUserSpecifiedNames = true;        // whether manual names specification is allowed or only name expression generation
 
     public NameExpressionDataIterator(DataIterator di, DataIteratorContext context, @Nullable TableInfo parentTable, @Nullable Container container, Function<String, Long> getNonConflictCountFn, String counterSeqPrefix)
     {
@@ -56,6 +57,12 @@ public class NameExpressionDataIterator extends WrapperDataIterator
         _getNonConflictCountFn = getNonConflictCountFn;
         _counterSeqPrefix = counterSeqPrefix;
 
+    }
+
+    public NameExpressionDataIterator setAllowUserSpecifiedNames(boolean allowUserSpecifiedNames)
+    {
+        _allowUserSpecifiedNames = allowUserSpecifiedNames;
+        return this;
     }
 
     MapDataIterator getInput()
@@ -93,7 +100,11 @@ public class NameExpressionDataIterator extends WrapperDataIterator
                 curName = StringUtils.isEmpty((String)curName) ? null : curName;
 
             if (curName != null)
+            {
+                if (!_allowUserSpecifiedNames)
+                    getErrors().addRowError(new ValidationException("Manual entry of names has been disabled for this folder. Only naming pattern generated names are allowed."));
                 return curName;
+            }
 
             Map<String, Object> currentRow = getInput().getMap();
 
--- a/api/src/org/labkey/api/dataiterator/NameExpressionDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/NameExpressionDataIterator.java
@@ -102,7 +102,7 @@ public class NameExpressionDataIterator extends WrapperDataIterator
             if (curName != null)
             {
                 if (!_allowUserSpecifiedNames)
-                    getErrors().addRowError(new ValidationException("Manual entry of names has been disabled for this folder. Only naming pattern generated names are allowed."));
+                    getErrors().addRowError(new ValidationException("Manual entry of names has been disabled for this folder. Only naming-pattern-generated names are allowed."));
                 return curName;
             }
 
--- a/api/src/org/labkey/api/exp/api/NameExpressionOptionService.java
+++ b/api/src/org/labkey/api/exp/api/NameExpressionOptionService.java
@@ -85,7 +85,7 @@ public interface NameExpressionOptionService
         @Override
         public String createPrefixedExpression(Container c, String nameExpression, boolean isAliquotNameExpression)
         {
-            return null;
+            return nameExpression;
         }
     }
 }
--- a/api/src/org/labkey/api/exp/api/NameExpressionOptionService.java
+++ b/api/src/org/labkey/api/exp/api/NameExpressionOptionService.java
@@ -1,0 +1,77 @@
+package org.labkey.api.exp.api;
+
+import org.jetbrains.annotations.NotNull;
+import org.labkey.api.data.Container;
+import org.labkey.api.security.User;
+import org.labkey.api.services.ServiceRegistry;
+
+import javax.annotation.Nullable;
+
+/**
+ * Service which exists to provide mostly application specific functionality around how name expressions work in
+ * data classes and sample types.
+ */
+public interface NameExpressionOptionService
+{
+    NameExpressionOptionService NO_OP_IMPL = new NoOpService();
+
+    @NotNull
+    static NameExpressionOptionService get()
+    {
+        NameExpressionOptionService impl = ServiceRegistry.get().getService(NameExpressionOptionService.class);
+        return impl != null ? impl : NO_OP_IMPL;
+    }
+
+    static void setInstance(NameExpressionOptionService impl)
+    {
+        ServiceRegistry.get().registerService(NameExpressionOptionService.class, impl);
+    }
+
+    /**
+     * Returns the optional name expression prefix configured for this container
+     */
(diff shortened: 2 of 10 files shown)


cnathe:
@labkey-klum I'm not sure if it was part of this story or not, but Rosaline and I were talking about some changes in the Field editor related to this container prefix and the per sample type naming pattern. She was demoing the container prefix part to me and mentioned that if the sample type had a naming pattern and then the container prefix was updated, then that updated prefix would override and change the per sample type naming pattern. Is this as expected? For example, if my sample type had a naming pattern of "PREFIX-Blood-${genId}" and the container prefix was changed to "PRX", I would not have expected the sample type naming pattern to then be set to "PRX". I would see it being updated to "PRX-Blood-${genId}". Just wanted to check in on this. Thanks.

api/src/org/labkey/api/dataiterator/NameExpressionDataIterator.java
```diff
@@ -93,7 +100,11 @@ public Object get(int i)
                 curName = StringUtils.isEmpty((String)curName) ? null : curName;
 
             if (curName != null)
+            {
+                if (!_allowUserSpecifiedNames)
+                    getErrors().addRowError(new ValidationException("Manual entry of names has been disabled for this folder. Only naming pattern generated names are allowed."));
```
labkey-susanh:
```suggestion
                    getErrors().addRowError(new ValidationException("Manual entry of names has been disabled for this folder. Only naming-pattern-generated names are allowed."));
```
Or, if you're anti-hyphen, "Only names generated by naming patterns are allowed."

api/src/org/labkey/api/exp/api/NameExpressionOptionService.java
```diff
@@ -0,0 +1,91 @@
...
+    String getExpressionPrefix(Container c);
+
+    /**
+     * Set the name expression prefix for the specified container. Setting the prefix to null
+     * will clear out an existing prefix.
+     */
+    void setExpressionPrefix(Container c, User user, @Nullable String prefix) throws Exception;
+
+    /**
+     * Returns whether user specified names for samples and dataclasses are permitted for this folder.
```
labkey-susanh:
```suggestion
     * Returns whether user-specified names for samples and dataclasses are permitted for this folder.
```

api/src/org/labkey/api/exp/api/NameExpressionOptionService.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    boolean allowUserSpecifiedNames(Container c);
+
+    /**
+     * Set whether user specified names for samples and dataclasses are allowed for this folder. If set to false
```
labkey-susanh:
```suggestion
     * Set whether user-specified names for samples and dataclasses are allowed for this folder. If set to false
```

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -205,6 +206,13 @@ public MutableColumnInfo createColumn(String alias, Column column)
                 c.setNullable(nameExpression != null);
                 String desc = ExpMaterialTableImpl.appendNameExpressionDescription(c.getDescription(), nameExpression);
                 c.setDescription(desc);
+
+                // shut off this field in insert and update views if user specified names are not allowed
```
labkey-susanh:
```suggestion
                // shut off this field in insert and update views if user-specified names are not allowed
```

api/src/org/labkey/api/exp/api/NameExpressionOptionService.java
```diff
@@ -0,0 +1,91 @@
...
+
+        @Override
+        public void setAllowUserSpecifiedNames(Container c, User user, boolean allowNames) throws Exception
+        {
+        }
+
+        @Override
+        public String createPrefixedExpression(Container c, String nameExpression, boolean isAliquotNameExpression)
+        {
+            return null;
```
labkey-susanh: Should this return `nameExpression`?

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -856,6 +869,22 @@ else if (aliquotedFromObj instanceof Number)
...
+                Object currNameObj = map.get("Name");
+                if (currNameObj != null && !_allowUserSpecifiedNames)
+                {
+                    if (StringUtils.isNotBlank(currNameObj.toString()))
+                    {
+                        if (_context.getInsertOption().equals(QueryUpdateService.InsertOption.MERGE))
+                        {
+                            // don't flag rows that already exist if the option is set to update existing
+                            if (!rowExists(currNameObj.toString()))
+                                addRowError("Manual entry of names has been disabled for this folder. Only naming pattern generated names (or existing names) are allowed.");
```
labkey-susanh:
```suggestion
                                addRowError("Manual entry of names has been disabled for this folder. Only naming-pattern- generated names (or existing names) are allowed.");

```

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -856,6 +869,22 @@ else if (aliquotedFromObj instanceof Number)
...
+                        }
+                        else
+                            addRowError("Manual entry of names has been disabled for this folder. Only naming pattern generated names are allowed.");
```
labkey-susanh:
```suggestion
                            addRowError("Manual entry of names has been disabled for this folder. Only naming-pattern-generated names are allowed.");

```

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -856,6 +869,22 @@ else if (aliquotedFromObj instanceof Number)
...
 
+                Object currNameObj = map.get("Name");
+                if (currNameObj != null && !_allowUserSpecifiedNames)
+                {
+                    if (StringUtils.isNotBlank(currNameObj.toString()))
+                    {
+                        if (_context.getInsertOption().equals(QueryUpdateService.InsertOption.MERGE))
+                        {
+                            // don't flag rows that already exist if the option is set to update existing
+                            if (!rowExists(currNameObj.toString()))
```
labkey-susanh, line 880: This seems like the sort of thing the ExistingRecordDataIterator was designed for, but I don't think it's in the mix here. This stashed set of names is probably sufficient, even if there's an outside chance the name won't be in the selected set.

labkey-susanh (approved):

api/src/org/labkey/api/exp/api/NameExpressionOptionService.java
labkey-klum: yes good catch

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -627,6 +630,17 @@ public ExpSampleTypeImpl createSampleType(Container c, User u, String name, Stri
...
 
+        NameExpressionOptionService svc = NameExpressionOptionService.get();
+        if (!svc.allowUserSpecifiedNames(c))
+        {
+            if (nameExpression == null)
+                throw new ExperimentException(NAME_EXPRESSION_REQUIRED_MSG);
+
+            // automatically apply the configured prefix to the name expression
+            nameExpression = svc.createPrefixedExpression(c, nameExpression, false);
+            aliquotNameExpression = svc.createPrefixedExpression(c, aliquotNameExpression, true);
```
labkey-susanh, line 644: I'm not sure the applying of prefixes should happen only when we don't allow user-specified names.  It seems like these settings are meant to be independent of each other.

labkey-susanh (changes requested):

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-klum: This is true, I've split it out from that conditional

labkey-susanh (approved):

PR #2689, opened 2021-10-14
Issue 44126: CommandTasks fail on remote pipeline servers

labkey-jeckels:
#### Rationale
Remote pipeline servers fail to run CommandTasks because we've introduced code that only works when invoked on the web server where we have the DB connection available

#### Changes
* Check if we're on the web server before setting up script replacement tokens and auto-creating an API key
* Debug logging to help track down future problems in finding MS2 conversion tools
* Improve error messages when invoking methods that don't work on remote servers

### DIFF
--- a/api/src/org/labkey/api/security/SecurityManager.java
+++ b/api/src/org/labkey/api/security/SecurityManager.java
@@ -646,6 +646,9 @@ public class SecurityManager
         {
             return _apikey;
         }
+
+        @Override
+        public abstract void close();
     }
 
     private static final class HttpSessionTransformSession extends TransformSession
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -646,6 +646,9 @@ public String getApiKey()
         {
             return _apikey;
         }
+
+        @Override
+        public abstract void close();
```
labkey-adam, line 651: This is causing a build break in `CromwellGctTask` (MacCossLabModules)... need to remove `catch (IOException)` there

labkey-adam (approved):

PR #2696, opened 2021-10-18
Defer DisplayColumnFactory instantiation error logging until column is actually used

labkey-adam:
#### Rationale
We want to relax the error logging when a custom `DisplayColumnFactory` can't be constructed. [Issue 44103: Announcements module has an undeclared dependency on Wiki module](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44103)

#### Changes
* Defer logging of instantiation errors until the column is used

### DIFF
--- a/api/src/org/labkey/api/data/BaseColumnInfo.java
+++ b/api/src/org/labkey/api/data/BaseColumnInfo.java
@@ -51,6 +51,8 @@ import org.labkey.data.xml.DbSequenceType;
 import org.labkey.data.xml.PropertiesType;
 
 import java.beans.Introspector;
+import java.io.IOException;
+import java.io.Writer;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.sql.ResultSet;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/BaseColumnInfo.java
```diff
@@ -1253,12 +1248,26 @@ else if (xmlCol.isSetKeyVariable())
...
+                // Substitute a factory that provides a renderer that displays and logs the error at render time
+                LOG.debug(message, e);
+                _displayColumnFactory = colInfo -> {
+                    LOG.error(message, e);
+                    return new SimpleDisplayColumn()
+                    {
+                        @Override
+                        public void renderGridCellContents(RenderContext ctx, Writer out) throws IOException
+                        {
+                            out.write(PageFlowUtil.filter("Error: " + message));
```
labkey-susanh, line 1267: This is going to be quite ugly for the user to see in their grid, which will, of course, make it obvious that what they see there is not their data, but is pretty unfriendly for those who know nothing about DisplayColumnFactories and the like.  Would it be sufficient to just display a short message such as "Display error." here and leave the rest in the logs?

labkey-susanh (approved):

api/src/org/labkey/api/data/BaseColumnInfo.java
labkey-adam: Yes, it's noisy, but this should never happen IRL. The schema would have to exist but not the code it references. (Which won't be the case in LKSM.) Noisy means users are more likely to alert admins... unclear how often admins proactively check the logs for errors like this.

But sure... we can render whatever we want. If you prefer a generic (and shorter) message like that I can do that.

RosalineP (approved):
Very cool 👍

api/src/org/labkey/api/data/BaseColumnInfo.java
labkey-jeckels: If it should never happen IRL, can we crank this up to error level logging so tests will fail if it does happen?

labkey-adam: The logging at render time is error level; if any code requests the column's display column it'll log an ERROR and the tests should fail.

The only real change was to dial down the logging level at XML parsing time, to address the LKSM case, where the parsing always takes place but wiki columns won't be referenced.

Let me know if I'm not understanding your request...

labkey-jeckels: Nope, my mistake. I saw the debug level logging and stopped looking for me. Looks good.

labkey-adam: @labkey-susanh and I discussed the noisy message. Apathy and inertia have convinced us to leave it the way it is.

PR #2697, opened 2021-10-18
44123: Issue list "Estimate" lookup doesn't resolve to description in details view

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44123

The estimate lookup was not rendering properly because the domain property had been setup to exclude that field from the default view:

image

When we create the render context for that view we populate the context using the values from the default view, so for the lookup only the key is returned (but not the title column).

#### Changes
- When we setup the render context for the various issue views, ask for all of the fields on the table so that rendering can proceed regardless of domain property configurations.
- The domain property configurations were already being respected and were independent of the render context prep.
- I'll add the field back to the default view on labkey.org until we can deploy this fix for 21.11.


labkey-adam:
Failing tests?

labkey-adam (approved):

ankurjuneja (approved):

PR #2698, opened 2021-10-19
Item 9584: ManageSampleStatusesPanel for sample statuses CRUD within a given container

cnathe:
#### Rationale
Recent stories have added a Sample Status column and support for status change checks for permitted operations. This PR adds the UI component to manage the sample statuses available for a given container. It will be shown in LKS, LKSM, and LKB. In addition, this story adds in a SampleStatusService registered from the SM module so that by default the sample status support is not "enabled" unless the SM module is included in the build and registers the service implementation to support it.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/653
* https://github.com/LabKey/platform/pull/2698
* https://github.com/LabKey/sampleManagement/pull/726
* https://github.com/LabKey/biologics/pull/1026
* https://github.com/LabKey/inventory/pull/323

#### Changes
* remove ExpSchema and SampleStateManager isOperationPermitted() check in favor of calls to SampleStatusService method
* SampleStatusService and default provider for community distribution scenario
* AbstractManageQCStateAction validation checks for empty label value
* DataStatesTableInfo updates for QueryUpdateService validation checks for valid state change and empty label value
* SampleTypeWebPart update to conditionally add "Manage Sample Statuses" button to grid header when supportsSampleStatus() is true


api/src/org/labkey/api/qc/SampleStatusService.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.List;
+
+/**
+ * Service to assign a status/state to samples. Sample states can be created for a few different types which will
+ * allow for certain permitted operations based on that state (i.e. Available, Consumed, Locked). The default
+ * implementation/provider will permit all operations for all samples regardless of their status value.
+ */
+public interface SampleStatusService
+{
+    List<SampleStatusService> _providers = new ArrayList<>();
```
labkey-susanh, line 20: Why is this a list if you can only ever register one?

api/src/org/labkey/api/qc/SampleStatusService.java
```diff
@@ -0,0 +1,111 @@
...
+    SampleStatusService _defaultProvider = new DefaultSampleStatusService();
+
+    static void registerProvider(SampleStatusService provider)
+    {
+        if (_providers.isEmpty())
+            _providers.add(provider);
+        else
+            throw new RuntimeException("An implementation of SampleStatusService has previously been registered");
```
labkey-susanh:
```suggestion
            throw new RuntimeException("An implementation of SampleStatusService has previously been registered.");
```

api/src/org/labkey/api/qc/DataStateHandler.java
```diff
@@ -15,16 +15,28 @@
...
     boolean isBlankStatePublic(Container container);
+
+    /**
+     * Check if a given state allows for changes based on things like if it is in-use, etc.
+     * @param container
+     * @param state The QC state being changed
+     * @param rowUpdates The map of row changes for this state
+     * @return Error message to show the user if the change is not allowed
+     */
+    @Nullable String isStateChangeAllowed(Container container, DataState state, Map<String, Object> rowUpdates);
```
labkey-susanh: From the name I would expect this to return a boolean. Perhaps `getStateChangeError(...)`?

core/src/org/labkey/core/query/DataStatesTableInfo.java
```diff
@@ -80,9 +91,36 @@ private boolean validateQCStateNotInUse(Map<String, Object> oldRowMap, Container
...
+                if (errorMsg != null)
+                    return errorMsg;
+            }
+            return null;
+        }
+
+        private boolean validateLabel(Map<String, Object> row)
+        {
+            String label = (String) row.get("label");
+            return !StringUtils.isBlank(label);
```
labkey-susanh, line 111: Should we have a `trim()` in here?

core/src/org/labkey/core/query/DataStatesTableInfo.java
```diff
@@ -80,9 +91,36 @@ private boolean validateQCStateNotInUse(Map<String, Object> oldRowMap, Container
...
+        }
+
         @Override
         protected Map<String, Object> updateRow(User user, Container container, Map<String, Object> row, @NotNull Map<String, Object> oldRow, boolean allowOwner, boolean retainCreation) throws InvalidKeyException, ValidationException, QueryUpdateServiceException, SQLException
         {
+            if (!validateLabel(row))
+                throw new QueryUpdateServiceException("State label cannot be blank.");
```
labkey-susanh:
```suggestion
                throw new QueryUpdateServiceException("Label cannot be blank.");
```
This avoids using "State" when the user sees nothing but "status".

core/src/org/labkey/core/query/DataStatesTableInfo.java
```diff
@@ -96,6 +134,9 @@ private boolean validateQCStateNotInUse(Map<String, Object> oldRowMap, Container
         @Override
         protected Map<String, Object> insertRow(User user, Container container, Map<String, Object> row) throws DuplicateKeyException, ValidationException, QueryUpdateServiceException, SQLException
         {
+            if (!validateLabel(row))
+                throw new QueryUpdateServiceException("State label cannot be blank.");
```
labkey-susanh:
```suggestion
                throw new QueryUpdateServiceException("Label cannot be blank.");
```

labkey-susanh (approved):

api/src/org/labkey/api/qc/SampleStatusService.java
cnathe: Two reasons:
1) I was following the pattern from another service interface
2) If I used a SampleStatusService variable and tried to set it in the registerProvider method, I get the following: `Cannot assign a value to final variable`

api/src/org/labkey/api/qc/SampleStatusService.java
cnathe: Done

api/src/org/labkey/api/qc/DataStateHandler.java
cnathe: Good suggestion. Done.

core/src/org/labkey/core/query/DataStatesTableInfo.java
cnathe: Done.

core/src/org/labkey/core/query/DataStatesTableInfo.java
cnathe: Done.

core/src/org/labkey/core/query/DataStatesTableInfo.java
cnathe: Not here, it looks like the StringUtils.isBlank check already handles that. But I have added a trim() on the client side before calling to insert/update rows.

PR #2704, opened 2021-10-20
log row values when exception during script execution

ankurjuneja:
#### Rationale
Log the row values when there is an exception during server side trigger script execution. 

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/256

### DIFF
--- a/api/src/org/labkey/api/data/triggers/ScriptTrigger.java
+++ b/api/src/org/labkey/api/data/triggers/ScriptTrigger.java
@@ -195,7 +195,7 @@ import java.util.Objects;
         }
         catch (NoSuchMethodException | ScriptException e)
         {
-            throw new UnexpectedException(e);
+            throw UnexpectedException.wrap(e, "Script execution failed for row with values: " + Arrays.toString(args));
         }
 
         return null;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/triggers/ScriptTrigger.java
```diff
@@ -195,7 +195,7 @@ protected void invokeTableScript(TableInfo table, Container c, User user, String
         }
         catch (NoSuchMethodException | ScriptException e)
         {
-            throw new UnexpectedException(e);
+            throw UnexpectedException.wrap(e, "Script execution failed for row with values: " + Arrays.toString(args));
```
labkey-jeckels: I like this approach, but we can do better when including the arguments into the error.

First, the init and complete functions will pass a single argument (the toString() of the TableInfo.TriggerType). Second, beforeUpdate and afterUpdate will pass both the new and old row maps and it won't be clear which is which.

I think we should pass a Function argument to _invokeTableScript() that knows how to render the args to the error message in a more descriptive way. init and complete don't have anything to include, and the insert/update/delete variants can include the row maps.

labkey-jeckels (changes requested):

labkey-adam (approved):

ankurjuneja:
@labkey-jeckels thanks for looking into this, github is not letting me merge until you approve the requested chnages.

labkey-jeckels (approved):
I merged release21.7-SNAPSHOT into the ehrModules FB with the same name so that we can get the test results prior to merging.

I'm not sure that the ehrModules change will be that valuable with this new, centralized logging, but it shouldn't do any harm either so I approved it too.

PR #2708, opened 2021-10-21
Item 9544: Restrict operations for selections of samples based on status

labkey-susanh:
#### Rationale
We are adding support for samples having status values that will restrict the types of operations that can be done on the samples.  The changes here are to support restrictions and warnings when a user has selected a number of samples that  may have different statuses, so we generalize the method we previously had for confirming deletion of samples to cover other types of operations as well.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/654
* https://github.com/LabKey/biologics/pull/1030
* https://github.com/LabKey/inventory/pull/325
* https://github.com/LabKey/sampleManagement/pull/728

#### Changes
* Add status check for adding assay data for samples that are either run or batch properties
* Remove RemoveFromPicklist operation since all samples can be removed from picklists and it seems unlikely we would restrict this. 
* Generalize and rename `GetMaterialDeleteConfirmationDataAction` and its relatives to apply to all operations

### DIFF
--- a/api/src/org/labkey/api/exp/api/SampleTypeService.java
+++ b/api/src/org/labkey/api/exp/api/SampleTypeService.java
@@ -53,7 +53,6 @@ public interface SampleTypeService
         UpdateStorageMetadata("updating storage metadata"),
         RemoveFromStorage("removing from storage"),
         AddToPicklist("adding to a picklist"),
-        RemoveFromPicklist("removing from a picklist"),
         Delete("deleting"),
         AddToWorkflow("adding to a workflow"),
         RemoveFromWorkflow("removing from a workflow"),
--- a/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
+++ b/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
@@ -3294,9 +3294,9 @@ public class ExperimentController extends SpringActionController
         public void validateForm(OperationConfirmationForm form, Errors errors)
         {
             if (form.getDataRegionSelectionKey() == null && form.getRowIds() == null)
-                errors.reject(ERROR_REQUIRED, "You must provide either a set of rowIds or a dataRegionSelectionKey");
+                errors.reject(ERROR_REQUIRED, "You must provide either a set of rowIds or a dataRegionSelectionKey.");
             if (form.getSampleOperation() == null)
-                errors.reject(ERROR_REQUIRED, "An operation type must be provided");
+                errors.reject(ERROR_REQUIRED, "An operation type must be provided.");
         }
 
         @Override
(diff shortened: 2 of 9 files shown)


api/src/org/labkey/api/exp/api/SampleTypeService.java
```diff
@@ -52,7 +52,6 @@
         UpdateStorageMetadata("updating storage metadata"),
         RemoveFromStorage("removing from storage"),
         AddToPicklist("adding to a picklist"),
-        RemoveFromPicklist("removing from a picklist"),
```
cnathe, line 55: I assume these means that we will just always allow for removing samples from picklists, regardless of status (i.e. it just isn't an operation we care about for status checking)? If so, does the "remove from workflow" fall into that same category?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -3318,6 +3285,59 @@ public void setRowIds(Set<Integer> rowIds)
...
+
+    @Marshal(Marshaller.Jackson)
+    @RequiresPermission(ReadPermission.class)
+    public class GetMaterialOperationConfirmationDataAction extends ReadOnlyApiAction<OperationConfirmationForm>
+    {
+        @Override
+        public void validateForm(OperationConfirmationForm form, Errors errors)
+        {
+            if (form.getDataRegionSelectionKey() == null && form.getRowIds() == null)
+                errors.reject(ERROR_REQUIRED, "You must provide either a set of rowIds or a dataRegionSelectionKey");
```
cnathe:
```suggestion
                errors.reject(ERROR_REQUIRED, "You must provide either a set of rowIds or a dataRegionSelectionKey.");
```

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -3318,6 +3285,59 @@ public void setRowIds(Set<Integer> rowIds)
...
+            if (form.getSampleOperation() == null)
+                errors.reject(ERROR_REQUIRED, "An operation type must be provided");
```
cnathe:
```suggestion
                errors.reject(ERROR_REQUIRED, "An operation type must be provided.");
```

cnathe (approved):

api/src/org/labkey/api/exp/api/SampleTypeService.java
labkey-susanh: It could be that we'd always want to allow removing from workflow, but for now, we restrict that for locked samples.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: Done.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: Done.

PR #2709, opened 2021-10-21
Add Support for WorkflowTask to ImportRunApiAction

labkey-alan:
#### Rationale
Workflow Task can be set via updateRows since my last PR, however it was not properly wired up to get set during assay import, which is where it is needed. This PR adds a workflowTask param to the ImportAPIAction.

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/113

#### Changes
* Add Support for WorkflowTask to ImportRunApiAction

### DIFF
--- a/api/src/org/labkey/api/assay/AssayRunDatabaseContext.java
+++ b/api/src/org/labkey/api/assay/AssayRunDatabaseContext.java
@@ -25,6 +25,7 @@ import org.labkey.api.exp.ObjectProperty;
 import org.labkey.api.exp.api.ExpData;
 import org.labkey.api.exp.api.ExpExperiment;
 import org.labkey.api.exp.api.ExpProtocol;
+import org.labkey.api.exp.api.ExpProtocolApplication;
 import org.labkey.api.exp.api.ExpRun;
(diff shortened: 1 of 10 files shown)


labkey-alan:
I wasn't sure exactly who should do the review for this, so I just tagged the folks who I felt would be most interested in the change. Let me know if there is someone more appropriate to do the review and I'll add them as a reviewer.

labkey-alan:
I apparently broke the build for [`LabDevKitModules`](https://github.com/LabKey/LabDevKitModules), which makes me wonder if I took the right approach. The fix would be easy, but not sure if we want to be breaking 3rd party modules?

labkey-adam:
`LabDevKitModules` isn't really 3rd party... we maintain these modules, which are EHR dependencies. You should be able to commit to this repo to accommodate your change.

labkey-jeckels (approved):

labkey-jeckels:
@labkey-alan I doubt there are too many implementations of AssayRunUploadContext in non-LK repos, but you could give a default implementation of the new method that just returns null

labkey-susanh (approved):

PR #2710, opened 2021-10-21
Show error message if study/folder/list export-to-zip process fails

labkey-adam:
#### Rationale
As discussed in [Issue 44152: Stop exporting corrupted archives](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44152), the two "pipeline" options on the folder/study export page will show the error page if an exception occurs during the export process. The "Browser as zip file" option does not show an error page, because it proactively configures the response to stream a zip file; the end result is a corrupted zip file with no warning (other than an exception stack trace in the log).

This change makes the "Browser as zip file" option behave like the pipeline options; if an error occurs during export the user will see a standard error page and the archive won't be sent to the browser. The downside is potential delay between clicking the Export button and seeing the file downloading in the browser. In Chrome, I see a spinner on the browser tab and a "Waiting for cache..." message below where downloads appear, but this may not be enough feedback for some users. We could consider client-side changes to provide better feedback (e.g., disable the Submit button as we do with the pipeline options or display some kind of modal spinny thing). This PR does none of that.

Current commits handle the folder/study export case. Once reviewed and approved, I'll make a similar change for the export list archive case.

#### Changes
* Export folder/study archives to a temporary file before streaming to the browser so export exceptions are presented by our normal error page

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -4371,7 +4371,7 @@ public class AdminController extends SpringActionController
                     {
                         ContainerManager.checkContainerValidity(container); // TODO: Why isn't this called in the other two cases?
 
-                        // Export to a temporary file first so any exceptions are displayed by the standard error page, Issue #44152
+                        // Export to a temporary file first so exceptions are displayed by the standard error page, Issue #44152
                         // Same pattern as ExportListArchiveAction
                         Path tempDir = FileUtil.getTempDirectory().toPath();
                         Path tempZipFile = exportFolderToFile(tempDir, container, writer, ctx, errors);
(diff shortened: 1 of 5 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -4366,26 +4366,27 @@ else if (root.isCloudRoot())
...
+                        Path tempZipFile = exportFolderToFile(tempDir, container, writer, ctx, errors);
+
+                        // No exceptions, so stream the resulting zip file to the browser
+                        try (OutputStream os = ZipFile.getOutputStream(getViewContext().getResponse(), tempZipFile.getFileName().toString()))
                         {
-                            writer.write(container, ctx, zip);
+                            Files.copy(tempZipFile, os);
                         }
+
+                        Files.delete(tempZipFile);
```
labkey-jeckels: Shouldn't this go in a finally block in case there's an IOException streaming the file back to the browser?

labkey-jeckels (approved):
One requested change, but no need for further review

core/src/org/labkey/core/admin/AdminController.java
labkey-adam: Sure. I didn't want to delete it in the case of an export exception (keep the temp file around for investigation purposes), but much less reason to keep it if a streaming exception hits.

PR #2712, opened 2021-10-22
Add some debug logging to folder type and dataset importers

labkey-tchad:
#### Rationale
`StudySimpleExportTest` fails intermittently when validating that the default number format is imported correctly.
```
org.junit.ComparisonFailure: expected:<[#.000]> but was:<[]>
  at org.junit.Assert.assertEquals(Assert.java:117)
  at org.junit.Assert.assertEquals(Assert.java:146)
  at org.labkey.test.tests.study.StudySimpleExportTest.verifyDefaultDatasetFormats(StudySimpleExportTest.java:298)
```
The imported `folder.xml` does contain `defaultNumberFormat="#.000"` as expected. Adding debug logging will allow us to verify whether the folder importer is setting that property correctly or, maybe, getting overwritten by the dataset importer.

#### Changes
* Add debug logging to `FolderTypeImporter` and `DatasetDefinitionImporter`
* Fix condition for importing extra date-time pattern

### DIFF
--- a/core/src/org/labkey/core/admin/importer/FolderTypeImporterFactory.java
+++ b/core/src/org/labkey/core/admin/importer/FolderTypeImporterFactory.java
@@ -38,6 +38,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * User: cnathe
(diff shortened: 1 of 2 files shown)


core/src/org/labkey/core/admin/importer/FolderTypeImporterFactory.java
```diff
@@ -124,10 +131,11 @@ public void process(PipelineJob job, ImportContext<FolderDocument.Folder> ctx, V
                 }
             }
 
-            if (folderXml.isSetExtraDateParsingPattern())
+            if (folderXml.isSetExtraDateTimeParsingPattern())
```
labkey-adam, line 134: Good catch!

labkey-adam (approved):

PR #2713, opened 2021-10-22
Fix RReport Junit tests for premium

XingY:
#### Rationale
The check for rstudio service needs to also check if rstudio is configured, in addition to the presence of rstudio module

#### Related Pull Requests
* 

#### Changes
*


labkey-danield (approved):
Looks ok to me.

PR #2714, opened 2021-10-22
Support additional configuration for exporting subfolders

labkey-nicka:
#### Rationale
Adds additional configuration for exporting folders in support of what is needed for creating cross-folder ELN data snapshots.

#### Related Pull Requests
* https://github.com/LabKey/labbook/pull/118
* https://github.com/LabKey/biologics/pull/1031
* https://github.com/LabKey/labkey-ui-components/pull/655
* https://github.com/LabKey/platform/pull/2714

#### Changes
* Allow implementors of `ImportContext` to specify the necessary permissions needed to export subfolders. Defaults to `FolderExportPermission`.
* Pass through any `XarExportContext` to subfolders being exported.
* Bug. Prevent `null` being placed into the collection for resolved `ColumnInfo` models coming from Vocabulary properties.

### DIFF
--- a/api/src/org/labkey/api/admin/ImportContext.java
+++ b/api/src/org/labkey/api/admin/ImportContext.java
@@ -62,7 +62,7 @@ public interface ImportContext<XmlType extends XmlObject> extends ContainerUser
     }
 
     // Used to determine the level of permissions the user needs to have in order to export a subfolder.
-    default Class<? extends Permission> getSubfolderPermission()
+    default Class<? extends Permission> getSubfolderExportPermission()
     {
         return FolderExportPermission.class;
     }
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/admin/ImportContext.java
```diff
@@ -60,6 +61,12 @@ default AuditBehaviorType getAuditBehaviorType() throws Exception
         return null;
     }
 
+    // Used to determine the level of permissions the user needs to have in order to export a subfolder.
+    default Class<? extends Permission> getSubfolderPermission()
```
labkey-susanh: If this applies only to export, does it belong in the `ImportContext` class?  (Oh, I see below that this is the type used in the `write` signature, so I guess so.) Should it use `Export` in its name (e.g., `getSubfolderExportPermission`)?

labkey-susanh (approved):

api/src/org/labkey/api/admin/ImportContext.java
labkey-nicka: Renamed to `getSubfolderExportPermission `.

PR #2715, opened 2021-10-25
Item 9574: Store aliquot count and total amount data in exp.materials table

XingY:
#### Rationale
Aliquot rollup columns (counts and volume) are computed columns which resulted in slow query time for large sample sets. This set of PRs implements the change to store the calculated rollup values in exp.material table, so they don't have to be computed at query time. 

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2715
* https://github.com/LabKey/labkey-ui-components/pull/656
* https://github.com/LabKey/inventory/pull/326
* https://github.com/LabKey/sampleManagement/pull/729
* https://github.com/LabKey/biologics/pull/1032

#### Changes
* add rollup fields to exp.material table
* wire up rollup recompute on sampletype update, sample insert/delete
* fix Issue 44039: SM: Cannot import a LKSM project if the project contains sub-aliquots


experiment/src/org/labkey/experiment/XarReader.java
```diff
@@ -1897,7 +1914,24 @@ else if (!diffs.isEmpty())
         }
         else
         {
-            protocol = ExperimentServiceImpl.get().saveProtocol(getUser(), xarProtocol);
+            if (xarProtocol.getLSID().equals(ExperimentService.SAMPLE_DERIVATION_PROTOCOL_LSID) || xarProtocol.getLSID().equals(ExperimentService.SAMPLE_ALIQUOT_PROTOCOL_LSID))
+            {
+                // create derivation and aliquot protocol using shared folder
```
labkey-susanh, line 1917: Is the shared folder where these normally reside?

experiment/src/org/labkey/experiment/api/Material.java
```diff
@@ -33,6 +33,12 @@
     private String aliquotedFromLSID;
     private Integer sampleState;
 
+    //rollup columns
```
labkey-susanh:
```suggestion
    // aliquot rollup columns
```

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -445,11 +449,14 @@ protected Domain getDomain()
                     throw new QueryUpdateServiceException(String.format("Sample with RowID %d cannot be deleted due to its current status (%s)", rowId, dataState));
                 }
 
+                if (map.containsKey("RootMaterialLSID") && !StringUtils.isEmpty(map.get("RootMaterialLSID")))
+                    aliquotParents.add((String) map.get("RootMaterialLSID"));
```
labkey-susanh: Seems like we might want a constant for this `RootMaterialLSID` field.  We use it a lot.

experiment/src/org/labkey/experiment/XarReader.java
XingY: Yes, importing folders containing derivatives or aliquots would fail if those protocols don't reside on shared folder. A repro would be to import from an archive containing derivatives/aliquots into a clean DB. Then subsequent import using a different import archive would fail. In that case, a upgrade/fix script to set the protocols to shared folder is needed to fix those dbs.

experiment/resources/schemas/dbscripts/postgresql/exp-21.016-21.017.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE exp.Material ADD COLUMN RecomputeRollup BOOL NULL DEFAULT(FALSE);
+ALTER TABLE exp.Material ADD COLUMN AliquotCount INTEGER NULL;
+ALTER TABLE exp.Material ADD COLUMN AliquotVolume FLOAT NULL;
+ALTER TABLE exp.Material ADD COLUMN AliquotUnit VARCHAR(10) NULL;
+
+UPDATE exp.Material SET RecomputeRollup=TRUE WHERE rowId IN
+                                                   (SELECT distinct parent.rowId
+                                                    FROM exp.material aliquot
+                                                             JOIN exp.material parent ON aliquot.rootMaterialLsid = parent.lsid
+                                                   );
```
labkey-matthewb: Simpler to do lsid IN ()?
WHERE lsid IN (SELECT rootMaterialLsid FROM exp.material)

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -348,7 +349,47 @@ public TableInfo getLookupTableInfo()
...
+                ret.setShownInDetailsView(false);
+                ret.setShownInInsertView(false);
+                ret.setShownInUpdateView(false);
+                return ret;
+            }
+
+            case AliquotCount:
+            {
+                var ret = wrapColumn(alias, _rootTable.getColumn("AliquotCount"));
+                ret.setUserEditable(false);
```
labkey-matthewb: Is is possible to set these properties in exp.xml?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -4298,6 +4320,15 @@ private void deleteMaterialByRowIds(User user, Container container,
                     if (!stDeleteFrom.getLSID().equals(material.getCpasType()))
                         throw new IllegalArgumentException("Error deleting '" + stDeleteFrom.getName() + "' sample: '" + material.getName() + "' is in the sample type '" + material.getCpasType() + "'");
                 }
+
+                if (isAliquotRollupSupported && !isTruncate && !StringUtils.isEmpty(material.getRootMaterialLSID()))
+                {
+                    ExpSampleType sampleType = material.getSampleType();
+                    if (!sampleTypeAliquotParents.containsKey(sampleType))
+                        sampleTypeAliquotParents.put(sampleType, new HashSet<>());
+                    sampleTypeAliquotParents.get(sampleType).add(material.getRootMaterialLSID());
```
labkey-matthewb: I think you can do this:
  sampleTypeAliquotParents.computeIfAbsent(sampleType, (k) -> new HashSet<>()).add(material.getRootMaterialLSID());

or look at org.apache.commons.collections.map.MultiValueMap

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -4398,6 +4429,29 @@ private void deleteMaterialByRowIds(User user, Container container,
...
+                    for (Map.Entry<ExpSampleType, Set<String>> sampleTypeParents: sampleTypeAliquotParents.entrySet())
+                    {
+                        ExpSampleType parentSampleType = sampleTypeParents.getKey();
+                        Set<String> parentLsids = sampleTypeParents.getValue();
+
+                        List<ExpMaterialImpl> parentSamples = getExpMaterialsByLsid(parentLsids);
+                        Set<Integer> parentSampleIds = new HashSet<>();
+                        parentSamples.forEach(p -> parentSampleIds.add(p.getRowId()));
+
+                        InventoryService.get().recomputeSamplesRollup(parentSampleIds, parentSampleType, container);
```
labkey-matthewb: Since you ae grouping by SampleType, I assume that means that exp.material by itself does not have all the information required by recomputeSamplesRollup().  It would be interesting to figure out if we could make this self contained.  (I haven't read that code yet)

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -424,6 +427,7 @@ protected Domain getDomain()
         else
```
labkey-matthewb: super.deleteRows() will call deleteRow().  Does that need to be updated?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -445,11 +449,14 @@ protected Domain getDomain()
...
 
+                if (map.containsKey("RootMaterialLSID") && !StringUtils.isEmpty(map.get("RootMaterialLSID")))
+                    aliquotParents.add((String) map.get("RootMaterialLSID"));
+
                 ids.add(rowId);
                 result.add(map);
             }
             // TODO check if this handle attachments???
-            ExperimentServiceImpl.get().deleteMaterialByRowIds(user, container, ids, true, _sampleType, false);
+            ExperimentServiceImpl.get().deleteMaterialByRowIds(user, container, ids, true, _sampleType, false, false);
```
labkey-matthewb, line 491: That probably should have been migrated to SampleTypeServiceImpl.  oops.

labkey-matthewb (changes requested):

experiment/resources/schemas/dbscripts/postgresql/exp-21.016-21.017.sql
XingY: good call, updated

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
XingY: yes, I did not know that

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: It's mainly to save a query to sample type for the configured sample type unit. Refactored

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: It's handled in deleteMaterialByRowIds

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -6610,10 +6667,20 @@ private void saveExpMaterialAliquotOutputs(List<ProtocolAppRecord> protAppRecord
...
                     }
                 }
             }
+
+            // mark aliquot parents RecomputeRollup=true
+            if (!parentIds.isEmpty())
+            {
+                SQLFragment sql = new SQLFragment("UPDATE ").append(tableInfo, "").
+                        append(" SET RecomputeRollup = ? WHERE RowId ");
+                sql.add(true);
```
labkey-susanh, line 6684: Does this need to be the True for the SQL dialect?

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: I think if you use append then need to use dialect.getBooleanTRUE(), but with add it's not needed? I'm not sure, but the test plan called out sql server testing and they seem to pass

labkey-matthewb (approved):

PR #2717, opened 2021-10-26
Issue 43523: MetadataUnavailableException - multiple document elements on table name rename

ankurjuneja:
#### Rationale
Query metadata editor allows the user to edit the tableName i.e. the queryName or tableName which causes problems when editing columns metadata through visual editor as multiple XML docs get generated one with old table name and one with new table name along with changed column's metadata.

#### Changes
* add error message when tableName attribute is changed in text metadata editor

Screen Shot 2021-10-25 at 5 58 03 PM

### DIFF
--- a/query/src/org/labkey/query/controllers/QueryController.java
+++ b/query/src/org/labkey/query/controllers/QueryController.java
@@ -1056,7 +1056,8 @@ public class QueryController extends SpringActionController
                                 {
                                     if (!Objects.equals(tableType.getTableName(), form.getQueryName()))
                                     {
-                                        errors.reject(ERROR_MSG, "Table name must not be modified");
+                                        errors.reject(ERROR_MSG, "Table name in the XML metadata must match the table/query name: " + form.getQueryName());
+
                                     }
 
                                     TableType.Columns tableColumns = tableType.getColumns();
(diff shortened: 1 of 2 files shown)


labkey-jeckels (changes requested):
Can we make it do the right thing automatically, and swap whatever name the user might have included in the XML with the correct name for the table/query? It's unfortunate that the XML schema requires the name at all for this use case.

I can imagine other repro scenarios where a user creates a query through the schema browser, sets some XML metadata on it, then renames the query and goes back to edit the XML. It would be nice for this to just work for them instead of giving them an error message of any kind.

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -1054,6 +1054,11 @@ public void validateForm(SourceForm form, Errors errors)
                             {
                                 if (null != tableType)
                                 {
+                                    if (!Objects.equals(tableType.getTableName(), form.getQueryName()))
+                                    {
+                                        errors.reject(ERROR_MSG, "Table name must not be modified");
```
labkey-jeckels:
```suggestion
                                        errors.reject(ERROR_MSG, "Table name in the XML metadata must match the table/query name: " + form.getQueryName());

```

labkey-jeckels (approved):

PR #2719, opened 2021-10-26
Hide archived assay designs

RosalineP:
#### Rationale
The purpose of this story is to provide a way for admins to archive an assay design. To this ends, we added a new 'Status' checkbox input to the Assay Designer Properties panel. When status is 'off', the assay design is considered to be archived, and does not appear in certain areas of the UI, such as the mega-menu or the new 'Active' tab of the Assays List page.

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/379
* https://github.com/LabKey/sampleManagement/pull/732
* https://github.com/LabKey/biologics/pull/1036
* https://github.com/LabKey/labkey-ui-components/pull/659

#### Changes
* (For Karl to populate)


assay/src/org/labkey/assay/AssayDomainServiceImpl.java
```diff
@@ -204,6 +204,7 @@ private GWTProtocol getAssayTemplate(AssayProvider provider, Pair<ExpProtocol, L
         result.setName(protocol.getName());
         result.setProviderName(provider.getName());
         result.setDescription(protocol.getDescription());
+        result.setStatus(protocol.getStatus() != null ? protocol.getStatus().name() : ExpProtocol.Status.Active.name());
```
RosalineP, line 207: A perhaps slightly hacky addition of mine to get create-case working

api/src/org/labkey/api/assay/AssayProvider.java
```diff
@@ -138,7 +138,7 @@ default String getLabel()
     @NotNull
     AssayTableMetadata getTableMetadata(@NotNull ExpProtocol protocol);
 
-    ExpProtocol createAssayDefinition(User user, Container container, String name, String description) throws ExperimentException;
+    ExpProtocol createAssayDefinition(User user, Container container, String name, String description, @Nullable ExpProtocol.Status status) throws ExperimentException;
```
labkey-nicka, line 141: @labkey-klum I guess I'm a little surprised this needed to bubble up all the way to the `AssayProvider` interface. Isn't there other properties that can be set on the `ExpProtocol` that aren't configured via this top-level creation method?

labkey-nicka (approved):

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
labkey-nicka: This is essentially how we configure the default value to be "Active". Seems fine.

api/src/org/labkey/api/assay/AssayProvider.java
labkey-klum: Not that I found, we could set it directly on the returned `ExpProtocol` object but that would have required a second save to the database.

PR #2723, opened 2021-10-28
Issue 43924: Assure samples that are deleted are removed from all picklists

labkey-susanh:
#### Rationale
When samples are deleted, they currently leave holes in the picklists that had included them.  While functionally this doesn't seem to cause a problem it does make for a somewhat odd user experience.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/735

#### Changes
* Add ExperimentListener for Picklists to samples that are about to be deleted can also be removed from picklists

### DIFF
--- a/list/src/org/labkey/list/PicklistMaterialListener.java
+++ b/list/src/org/labkey/list/PicklistMaterialListener.java
@@ -1,11 +1,12 @@
 package org.labkey.list;
 
 import org.labkey.api.data.Container;
-import org.labkey.api.data.SQLFragment;
-import org.labkey.api.data.SqlExecutor;
+import org.labkey.api.data.SimpleFilter;
+import org.labkey.api.data.Table;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.exp.api.ExpMaterial;
 import org.labkey.api.exp.api.ExperimentListener;
+import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.FilteredTable;
 import org.labkey.api.security.User;
 import org.labkey.list.model.ListDef;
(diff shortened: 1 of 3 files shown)


list/src/org/labkey/list/PicklistMaterialListener.java
```diff
@@ -0,0 +1,40 @@
...
+        List<Integer> materialIds = materials.stream().map(ExpMaterial::getRowId).collect(Collectors.toList());
+        picklists.forEach(picklist -> {
+            ListQuerySchema listQuerySchema = new ListQuerySchema(user, container);
+            TableInfo table = listQuerySchema.getTable(picklist.getName());
+            if (table != null)
+            {
+                SQLFragment sql = new SQLFragment("DELETE FROM ")
+                        .append(((FilteredTable) table).getRealTable(), "")
+                        .append(" WHERE SampleID ").appendInClause(materialIds, table.getSqlDialect());
+                new SqlExecutor(table.getSchema()).execute(sql);
```
cnathe: any reason that `Table.delete()` wasn't used here with the IN clause as a filter?

cnathe: also, any need for a transaction wrapping any of this?

cnathe (changes requested):

list/src/org/labkey/list/PicklistMaterialListener.java
labkey-susanh: It's a long and boring story of confusion, misdirection, and following bad patterns that exist in other places (doesn't work on the table itself since that's a filtered table, etc., etc.), but mainly it's because I forget there is such a thing as `Table.delete`.  That's been fixed now.

No need for a transaction because this is only called within the transaction that is deleting the samples.

cnathe (approved):

PR #2724, opened 2021-10-28
Issue 43984: Replace curly quotes with regular quotes in name expressions

labkey-susanh:
#### Rationale
When naming patterns are constructed via cut and paste and Microsoft somehow gets into the mix, curly quotes can get introduced.  We don't recognize these when parsing naming patterns and they generally just cause problems.  (See the [issue ](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43984) for more details.)

The implementation here is not sophisticated and is targeted to naming patterns.  This is deliberate so as not to introduce unnecessary processing in places where these characters a merely annoying and don't result in errors. If there's a more general place to include this that doesn't incur lots of overhead, please let me know.

#### Changes
* Add utility method for removing bad characters and apply that method before saving naming patterns.

### DIFF
--- a/api/src/org/labkey/api/util/StringUtilsLabKey.java
+++ b/api/src/org/labkey/api/util/StringUtilsLabKey.java
@@ -379,8 +379,36 @@ public class StringUtilsLabKey
         return s.length() > maxLength ? s.substring(0, maxLength - 3) + "..." : s;
     }
 
+    /**
+     * Replaces known bad characters (currently curly quotes) with reasonable replacements (non-curly quotes).
+     * @param original The string to be sanitized
+     * @return the sanitized string
+     */
+    public static String replaceBadCharacters(String original)
+    {
+        if (original == null)
+            return null;
+
+        return original.replaceAll("[\\u2018\\u2019]", "'")
+                        .replaceAll("[\\u201C\\u201D]", "\"");
+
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/util/StringUtilsLabKey.java
```diff
@@ -379,8 +379,36 @@ public static String truncate(@Nullable Object o, int maxLength)
         return s.length() > maxLength ? s.substring(0, maxLength - 3) + "..." : s;
     }
 
+    /**
+     * Replaces known bad characters (currently curly quotes) with reasonable replacements (non-curly quotes).
+     * @param original The string to be sanitized
+     * @return the sanitized string
+     */
+    public static String replaceBadCharacters(String original)
```
labkey-adam: Annotate parameter and return value with `@Nullable`?

labkey-adam (approved):
Given that this is the first time we've encountered curly quote issues in 18 years (AFAIK), I think a targeted fix for name expressions is fine

api/src/org/labkey/api/util/StringUtilsLabKey.java
labkey-susanh: Done.

PR #2725, opened 2021-10-28
StudyQuerySchema contextual role

labkey-matthewb:
#### Rationale
The study schema currently enforces non-readability of datasets by failing DatasetDefinition.getTableInfo().  If we want to allow control of readability by other means (other than just the study security configuration), we need to allow for the TableInfo to be checked for permissions later (e.g.  UserSchema.getTable()).

A side effect of the current implementation is that many tables did not correctly implement canRead() as they simply would never be returned to the user if canRead()==false.

This change generalizes how we elevate permissions in the StudySchema to use contextual roles and also correctly implements canRead() for tables in the study schema.

#### Related Pull Requests
* 

#### Changes
* StudyQuerySchema can take a contextual role in its constructor (replaces mustCheckPermissions flag)
* Study tables consistently implement canRead()
* The study schema is now responsible for hiding tables w/o read permission.

### DIFF
--- a/api/src/org/labkey/api/data/DataRegion.java
+++ b/api/src/org/labkey/api/data/DataRegion.java
@@ -59,6 +59,7 @@ import org.labkey.api.view.JspView;
 import org.labkey.api.view.NavTree;
 import org.labkey.api.view.NotFoundException;
 import org.labkey.api.view.PopupMenuView;
+import org.labkey.api.view.UnauthorizedException;
 import org.labkey.api.view.ViewContext;
 import org.labkey.api.visualization.VisualizationUrls;
 
@@ -765,7 +766,18 @@ public class DataRegion extends DisplayElement
     final public Results getResults(RenderContext ctx) throws SQLException, IOException
     {
         if (!ctx.getViewContext().hasPermission("DataRegion.getResults()", ReadPermission.class))
-            return null;
+        {
+            // this should usually be detected sooner!
+            _log.warn("Unauthorized thrown from DataRegion.getResults()");
+            throw new UnauthorizedException();
+        }
+
+        if (!getTable().hasPermission(ctx.getViewContext().getUser(), ReadPermission.class))
+        {
+            // this should usually be detected sooner!
+            _log.warn("Unauthorized thrown from DataRegion.getResults()");
+            throw new UnauthorizedException();
+        }
 
         DataRegion oldRegion = ctx.getCurrentRegion();
         if (oldRegion != this)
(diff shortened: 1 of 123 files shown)


labkey-adam:
Build failing in `nircEHRModules` (newer repo) due to `canReadSchema()`: "attempting to assign weaker access privileges; was public"

api/src/org/labkey/api/data/DataRegion.java
```diff
@@ -776,10 +777,11 @@ final public Results getResults(RenderContext ctx) throws SQLException, IOExcept
 
         try
         {
+            TableInfo tinfoMain = getTable();
```
labkey-adam, line 780: Why pull this up? Is it doing a required permissions check or something?

api/src/org/labkey/api/data/DataRegion.java
```diff
@@ -2003,7 +2003,11 @@ private void renderUpdateForm(RenderContext ctx, Writer out) throws IOException
             }
             else
             {
-                Collection<Map<String, Object>> maps = new TableSelector(getTable(), selectKeyMap.values(), new PkFilter(getTable(), viewForm.getPkVals()), null).getMapCollection();
+                if (!hasPermission(ctx, ReadPermission.class))
+                    throw new UnauthorizedException();
+
+                TableInfo tinfoMain = getTable();
+                Collection<Map<String, Object>> maps = new TableSelector(tinfoMain, selectKeyMap.values(), new PkFilter(getTable(), viewForm.getPkVals()), null).getMapCollection();
```
labkey-adam:
```suggestion
                Collection<Map<String, Object>> maps = new TableSelector(tinfoMain, selectKeyMap.values(), new PkFilter(tinfoMain, viewForm.getPkVals()), null).getMapCollection();
```

labkey-adam:
@labkey-matthewb Bunch of "use DatasetTableImpl.hasPermission()" exceptions in the BVT-EHR runs

study/src/org/labkey/study/query/StudyQuerySchema.java
```diff
@@ -221,9 +248,11 @@ public static StudyQuerySchema createSchemaWithoutStudy(Container c, User u)
 
     protected Set<String> getSubSchemaNames()
     {
-        if (AppProps.getInstance().isExperimentalFeatureEnabled(EXPERIMENTAL_STUDY_SUBSCHEMAS))
-            return new LinkedHashSet<>(Arrays.asList("Datasets","Design","Specimens"));
-        return Collections.emptySet();
+        if (!AppProps.getInstance().isExperimentalFeatureEnabled(EXPERIMENTAL_STUDY_SUBSCHEMAS))
+            return Collections.emptySet();
+        if (null != ModuleLoader.getInstance().getModule("Specimen"))
```
labkey-adam, line 251: You can check `null != SpecimenService.get()` instead

labkey-adam (changes requested):
Everything reviewed and seems very close. Please review comments and get EHR, etc. tests in reasonable shape and I'll approve.

api/src/org/labkey/api/data/DataRegion.java
labkey-matthewb: I think that was an intermediate change before I switched this.hasPermission()

labkey-matthewb: changed in 21.11_fb_readerrole

PR #2707, opened 2021-10-20
reduce dataset id collisions

labkey-matthewb:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43491
21.11 version of this pull request https://github.com/LabKey/platform/pull/2701

#### Related Pull Requests
* 

#### Changes
*

PR #2727, opened 2021-10-29
Fix PipelineTest, SharedStudyTest, and PanoramaPublicTests failures

Sigmonia:
#### Rationale
Fix failing tests

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2663

#### Changes
* Prevent setting updating logFile after Job completion for non-cloud jobs (For cloud jobs: log is moved to cloud working dir)
* Fix Job LocalDirectory overwriting log file location for split-able jobs causing error
* Fix incorrect expected value (StudyTest

### DIFF
--- a/study/test/src/org/labkey/test/tests/study/SharedStudyTest.java
+++ b/study/test/src/org/labkey/test/tests/study/SharedStudyTest.java
@@ -59,7 +59,7 @@ public class SharedStudyTest extends BaseWebDriverTest
     private static final String STUDY1 = "Study001";
     private static final String STUDY2 = "Study002";
     private static final String SHARED_DEMOGRAPHICS = "P_One_Shared";
-    private static final String SHARED_DEMOGRAPHICS_ID = "5001";
+    private static final String SHARED_DEMOGRAPHICS_ID = "10001";
     private static final String STUDY2_DATASET = "Extra Dataset";
     private static final String STUDY2_DATASET_ID = "999";
     private static final String[] STUDY1_PTIDS = {"1000", "1001", "1002", "1003"};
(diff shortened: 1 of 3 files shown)


study/test/src/org/labkey/test/tests/study/SharedStudyTest.java
```diff
@@ -59,7 +59,7 @@
     private static final String STUDY1 = "Study001";
     private static final String STUDY2 = "Study002";
     private static final String SHARED_DEMOGRAPHICS = "P_One_Shared";
-    private static final String SHARED_DEMOGRAPHICS_ID = "5001";
+    private static final String SHARED_DEMOGRAPHICS_ID = "10001";
```
labkey-jeckels, line 62: This seems unrelated.

labkey-jeckels (approved):

study/test/src/org/labkey/test/tests/study/SharedStudyTest.java
Sigmonia: Agreed, and I cant find the originating change. But it was needed to unblock the SharedStudyTest's setup method.

Sigmonia:
https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_ModuleSuites_MacCossLabModulesPostgres/1588748?buildTab=overview

study/test/src/org/labkey/test/tests/study/SharedStudyTest.java
labkey-adam: I'll bet it was related to the dataset ID collisions change: https://github.com/LabKey/platform/pull/2707

PR #2729, opened 2021-10-29
Issue 44151: Add site warning message for failed WebSocket connection

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44151

As we make increasing use of websocket connections for notifications, it's more important to make sure that they're working as expected. This PR adds a site warning message for admin users to see when the WebSocket Connection Failed. This is checked on the client side and in the failure case, calls the server API to track the failure, add a metric flag, and display the warning accordingly (similar to warnings for locked projects, downgraded modules, etc.).

Screen Shot 2021-10-29 at 9 47 16 AM

#### Related Pull Requests
* n/a

#### Changes
* add websocket connection success/failure metric tracking
* core controller WebSocketConnectionAction to set session attribute for a failed connection and add metric
* CoreWarningProvider update to add site warning message based on the websocket connection metrics
* WebSocket.js client side check to catch failed connection and call API action to notify the server
* fix for LABKEY.Utils.modal close button in LKSM app use-case

### DIFF
--- a/core/src/org/labkey/core/CoreController.java
+++ b/core/src/org/labkey/core/CoreController.java
@@ -203,6 +203,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 import static org.labkey.api.view.template.WarningService.SESSION_WARNINGS_BANNER_KEY;
+import static org.labkey.core.view.template.bootstrap.CoreWarningProvider.WEBSOCKET_CONNECTION_KEY;
 
 /**
  * User: jeckels
@@ -2415,6 +2416,53 @@ public class CoreController extends SpringActionController
             }
 
             return new ApiSimpleResponse(json);
+        }
+    }
+
+    @RequiresLogin
+    public class WebSocketConnectionAction extends MutatingApiAction<WebSocketWarningForm>
+    {
+        @Override
+        public ApiResponse execute(WebSocketWarningForm form, BindException errors)
+        {
+            // update session attribute so warnings will display via CoreWarningProvider.java
+            ViewContext context = getViewContext();
+            HttpSession session = context.getRequest().getSession(true);
+            Object previousValue = session.getAttribute(WEBSOCKET_CONNECTION_KEY);
+            session.setAttribute(WEBSOCKET_CONNECTION_KEY, form.isConnected());
+
+            JSONObject json = new JSONObject();
+            json.put("success", true);
+
+            // if the value has changed, recheck for warning messages
+            Object updatedValue = session.getAttribute(WEBSOCKET_CONNECTION_KEY);
+            if (previousValue == null || !previousValue.equals(updatedValue))
+            {
+                Warnings warnings = WarningService.get().getWarnings(context);
+                if (!warnings.isEmpty())
+                    json.put("warningsHtml", WarningService.get().getWarningsHtml(warnings, context).toString());
+
+                // if the value changed to "not connected", set client side metric
+                if (updatedValue != null && !(boolean) updatedValue)
+                    ClientSideMetricManager.get().setFlag("Core", "webSocketConnectionFailed", true);
+            }
+
+            return new ApiSimpleResponse(json);
+        }
+    }
+
+    public static class WebSocketWarningForm
+    {
+        private boolean _connected;
+
+        public boolean isConnected()
+        {
+            return _connected;
+        }
+
+        public void setConnected(boolean connected)
+        {
+            _connected = connected;
         }
     }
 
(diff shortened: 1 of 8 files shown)


cnathe:
@josheckels let me know your thoughts on the text to include in the site warning banner and if the link to the "Tomcat Configuration" doc page is the correct target.

core/src/org/labkey/core/view/template/bootstrap/CoreWarningProvider.java
```diff
@@ -168,6 +172,14 @@ private void getModuleErrorWarnings(Warnings warnings, ViewContext context)
         }
     }
 
+    private void getWebSocketConnectionWarnings(Warnings warnings, ViewContext context)
+    {
+        HttpSession session = context.getRequest().getSession(true);
+        Object connected = session.getAttribute(WEBSOCKET_CONNECTION_KEY);
+        if (SHOW_ALL_WARNINGS || (connected != null && !(boolean) connected))
+            addStandardWarning(warnings, "The WebSocket connection failed.", "configTomcat#websocket", "Tomcat Configuration");
```
labkey-jeckels:
```suggestion
            addStandardWarning(warnings, "The WebSocket connection failed. LabKey Server uses WebSockets to send notifications and alerts users when their session ends.", "configTomcat#websocket", "Tomcat Configuration");

```

core/src/org/labkey/core/CoreController.java
```diff
@@ -2418,6 +2419,53 @@ public ApiResponse execute(Object o, BindException errors)
...
+            ViewContext context = getViewContext();
+            HttpSession session = context.getRequest().getSession(true);
+            Object previousValue = session.getAttribute(WEBSOCKET_CONNECTION_KEY);
+            session.setAttribute(WEBSOCKET_CONNECTION_KEY, form.isConnected());
+
+            JSONObject json = new JSONObject();
+            json.put("success", true);
+
+            // if the value has changed, recheck for warning messages
+            Object updatedValue = session.getAttribute(WEBSOCKET_CONNECTION_KEY);
```
labkey-jeckels: This mixed approach of tracking at both the session level and then setting the metric server-side seems a bit odd, but perhaps OK. Though the client will only report when it fails to connect, right? So if a single client fails anytime, the server will be thinking that it's generally not working correctly?

labkey-jeckels (approved):

core/src/org/labkey/core/view/template/bootstrap/CoreWarningProvider.java
cnathe: Done.

core/src/org/labkey/core/CoreController.java
cnathe: Yes, I agree it is a bit odd but I couldn't think of a different/better way to track the metric. You are correct that the client will only report on failed connections. So basically if this metric shows up it would tell us that "at least one user on this server hit a failed connection during this server's last reporting window". I was initially using the incrementing client side metrics for this, but that number didn't seem very useful as it would increment for each user session that hit a failed connection.  Thoughts?

Basically the code in WebSocket.js is currently calling the API endpoint on each page load when the connection fails. Then the server side code flips the session boolean and the metric boolean the first time it sees this and each additional time just checks the current value and says "yeah yeah I already know".

cnathe: @labkey-jeckels the updated implementation we discussed has been pushed here for review. This takes the session attribute tracking out and replaces it with the success/failure counters. So this is currently, via WebSocket.js, making an API call on every page load for a logged in user to increment the counter for if the websocket connection was successful or failed. The WebSocketConnectionAction will increment one of the two counters for each call. Was that when you were thinking? Or were you thinking that this was the action that tracked the failures, but the successes would be counted in some other place or via some other existing action / endpoint?

labkey-jeckels: I think we should do it in NotificationEndpoint and other places that are accepting WebSocket connections. No need to introduce an extra AJAX call on every successful page load.

cnathe: Yup, that looks to effectively do the same without making the extra API (i.e. call the success increment on NotificationEndPoint.onOpen). This is ready for re-review.

core/src/org/labkey/core/metrics/WebSocketConnectionManager.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.labkey.api.usageMetrics.UsageMetricsProvider;
+
+import java.util.Map;
+
+public class WebSocketConnectionManager implements UsageMetricsProvider
+{
+    private static final WebSocketConnectionManager _instance = new WebSocketConnectionManager();
+
+    private int successCounter = 0;
```
labkey-jeckels: Let's use AtomicInteger for these to maximize thread safety (though the risk here is very low). See ViewServlet._requestCount for an example.

labkey-jeckels (approved):
One small followup suggestion, but no further review needed. Thanks for the iteration on this - I think this should be a useful way to track the availability of websockets.

core/src/org/labkey/core/metrics/WebSocketConnectionManager.java
cnathe: Done

PR #2730, opened 2021-10-29
Issue44091: Race condition prevents study dataset loading 

RosalineP:
#### Rationale
If a user, upon loading a page to view a dataset, hovered rapidly over the still-loading icons that generate helpDivs, it was possible for the page to blank-out indefinitely. 
This is because a call to document.write() while the document was assumed to be open, but was actually closed was replacing the contents of the page's <body/> with the contents of the helpDiv. (This being a known behavior of document.write() when a document is closed.)
As it is an edge case to be creating and hiding helpDivs while actively loading a page, a solution is to not render helpDivs until the page is fully loaded.

#### Related Pull Requests
* N/A

#### Changes
* Remove clause that uses `document.write()` when document is open

### DIFF
--- a/internal/webapp/labkey.js
+++ b/internal/webapp/labkey.js
@@ -151,8 +151,6 @@ if (typeof LABKEY == "undefined")
                 elem.innerHTML = html;
                 document.body.appendChild(elem.firstChild);
             }
-            else
-                document.write(html);
         };
 
         //private. used to append additional module context objects for AJAXd views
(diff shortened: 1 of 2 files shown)


labkey-nicka (approved):
Rather than having our own `addMarkup` method I think we should consider taking that out of `labkey.js` and switching the help div mechanism to use jQuery instead. It can handle the safety of all this and add listeners, etc. Not for this PR but would get us out of this business.

labkey-jeckels (approved):

labkey-jeckels:
@RosalineP note that we're not doing regular merges from the 21.3 branch anymore. You'll need to request a special merge to get this into develop. @labkey-sweta is currently on merge duty.

PR #2733, opened 2021-11-01
Hide sample status column for a sample type display/insert/update when container doesn't have statuses defined

cnathe:
#### Rationale
After merging the final set of changes for the Sample Status feature, I noticed that in LKS if you have a sample type but the container does not have any sample statuses defined, the status input / dropdown for the insert and update forms is empty. Also the column in the default sample type grid is shown but is always empty. This PR checks for defined sample statuses in the given container when decided to show/hide the status column for the sample type display/insert/update views.

#### Related Pull Requests
* https://github.com/LabKey/premium/pull/282

#### Changes
* ExpMaterialTableImpl update to include "does this container have sample statuses defined" in check to hide/show the sample status column

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
@@ -332,7 +332,7 @@ public class ExpMaterialTableImpl extends ExpRunItemTableImpl<ExpMaterialTable.C
                 return (BaseColumnInfo) createPropertiesColumn(alias);
 
             case SampleState:
-                boolean statusEnabled = SampleStatusService.get().supportsSampleStatus();
+                boolean statusEnabled = SampleStatusService.get().supportsSampleStatus() && SampleStatusService.get().getStates(getContainer()).size() > 0;
                 var ret = wrapColumn(alias, _rootTable.getColumn(column.name()));
                 ret.setLabel("Status");
                 ret.setHidden(!statusEnabled);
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -333,7 +333,7 @@ public StringExpression getURL(ColumnInfo parent)
 
             case SampleState:
             {
-                boolean statusEnabled = SampleStatusService.get().supportsSampleStatus();
+                boolean statusEnabled = SampleStatusService.get().supportsSampleStatus() && SampleStatusService.get().getStates(getContainer()).size() > 0;
```
labkey-susanh, line 336: This also means that if all statuses are intentionally removed, we won't show the column. I don't know if we have precedence for that, but it seems OK as long as the Manage Statuses button remains on the sample type grid.

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
cnathe: Correct. The Manage Statuses button remains so they can always add some back and "turn on" the feature.

cnathe:
@labkey-susanh TC looks happy with this change. But after thinking about it a bit, I am debating if this should be targeting 21.11? Without any sample statuses defined in a container, it is a little odd to have a useless column and select input showing up for the sample type insert/update form. Thoughts?

labkey-susanh:
> @labkey-susanh TC looks happy with this change. But after thinking about it a bit, I am debating if this should be targeting 21.11? Without any sample statuses defined in a container, it is a little odd to have a useless column and select input showing up for the sample type insert/update form. Thoughts?

Yeah, I think you're right since otherwise some folks will wait for 22.3 to get this update.

PR #2734, opened 2021-11-01
Explicitly set dataset ID to test dataset shadowing

labkey-tchad:
#### Rationale
Automatically generated dataset IDs start at 10001 to avoid collisions but some test cases depend on that collision. Explicitly setting the ID should get the test back to passing.

#### Related Pull Requests
* #2707 
* #2727 

#### Changes
* Explicitly set dataset ID to test dataset shadowing


labkey-tchad:
Verifying on TeamCity: https://teamcity.labkey.org/buildConfiguration/LabKey_2111Release_Community_BvtAPostgres/1591454

labkey-sweta (approved):

PR #2735, opened 2021-11-01
Issue #42992: Set default ontology filter tab

Sigmonia:
#### Rationale
Issue #42992: concept filter dialog defaults

#### Changes
* Adjusted the default filter tab for ontology columns

### DIFF
--- a/api/webapp/clientapi/ext3/FilterDialog.js
+++ b/api/webapp/clientapi/ext3/FilterDialog.js
@@ -69,7 +69,7 @@ LABKEY.FilterDialog = Ext.extend(Ext.Window, {
                 handler: this.closeDialog,
                 scope: this
             }],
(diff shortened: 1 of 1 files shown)


api/webapp/clientapi/ext3/FilterDialog.js
```diff
@@ -322,6 +322,11 @@ LABKEY.FilterDialog = Ext.extend(Ext.Window, {
         return filters;
     },
 
+    getDefaultTab: function() {
+        return this.column.conceptURI === CONCEPT_CODE_CONCEPT_URI && this.hasOntologyModule ?
```
cnathe: looks like this check is done in a few places in this file. could the `this.column.conceptURI === CONCEPT_CODE_CONCEPT_URI && this.hasOntologyModule` part be moved to a helper function?

cnathe (approved):
this approach looks good to me. one suggestion but no need to re-review.

PR #2740, opened 2021-11-02
npm run start-link fix issue with @webpack-cli/serve version

cnathe:
#### Rationale
The npm run start-link was hitting some TypeError issues with mismatches in the package versions in the LK module node_modules and the versions used by the latest @labkey/components package. This PR fixes that by updating the @labkey/components version and rehydrating the package-lock.json file. However, the rehydrate brought in v1.6.0 of @webpack-cli/serve which causes the following issue: [webpack-cli] Unable to load '@webpack-cli/serve' command. This is fixed by downgrading the package-lock.json version of @webpack-cli/serve to v1.5.2. 

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2740
* https://github.com/LabKey/puppeteer/pull/17
* https://github.com/LabKey/provenance/pull/78
* https://github.com/LabKey/commonAssays/pull/381
* https://github.com/LabKey/moduleEditor/pull/48
* https://github.com/LabKey/tutorialModules/pull/72

#### Changes
* update @labkey/components and @labkey/build versions
* rehydrate package-lock.json
* downgrade @webpack-cli/server from v1.6.0 to 1.5.2

### DIFF
--- a/assay/package-lock.json
+++ b/assay/package-lock.json
@@ -5,17 +5,17 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.14.5",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.14.5.tgz",
-      "integrity": "sha512-9pzDqyc6OLDaqe+zbACgFkb6fKMNG6CObKpnYXChRsvYGyEdc7CA2BaqeOM+vOtCS5ndmJicPJhKAwYRI6UfFw==",
+      "version": "7.16.0",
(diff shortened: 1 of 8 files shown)


Sigmonia (approved):

cnathe:
@labkey-alan @labkey-nicka Do you think we should be pinning the @webpack-cli/serve version to `1.5.2` somewhere because of this? If so, where? Without doing that, then a rehydrate of one of these module would likely result in the same npm run start-link issues if it bumps to v1.6.0.

PR #2741, opened 2021-11-02
Issue 44151: Let admins know when there's a websocket connection issue - help link

labkey-jeckels:
#### Rationale
When a help link is targeting a fragment within the doc page, we're mangling the URL by appending a "referrer=inPage" or similar GET parameter after the #

Also, we can improve our logging and error messages to help users zero in on problems

#### Changes
* Correctly assemble help links that target a fragment
* Include string value when in error message about values being too long, truncating to keep the output manageable
* Debug-level logging to help diagnose failure to report exceptions/usage metrics
* Be specific about column that failed to return a value when we can't do query logging

### DIFF
--- a/api/src/org/labkey/api/data/LoggingResultSetWrapper.java
+++ b/api/src/org/labkey/api/data/LoggingResultSetWrapper.java
@@ -133,7 +133,7 @@ public class LoggingResultSetWrapper extends ResultSetWrapper
             {
                 Object obj = getObject(dataLoggingColumn.getAlias());
                 if (null == obj)
-                    throw new UnauthorizedException("Unable to read expected data logging column.");
+                    throw new UnauthorizedException("Unable to read expected data logging column for " + dataLoggingColumn.getFieldKey() + " with alias " + dataLoggingColumn.getAlias());
                 _dataLoggingValues.add(obj);
             }
         }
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/data/LoggingResultSetWrapper.java
```diff
@@ -133,7 +133,7 @@ protected void updateQueryLogging() throws SQLException
             {
                 Object obj = getObject(dataLoggingColumn.getAlias());
                 if (null == obj)
-                    throw new UnauthorizedException("Unable to read expected data logging column.");
+                    throw new UnauthorizedException("Unable to read expected data logging column for " + dataLoggingColumn.getFieldKey() + " with alias " + dataLoggingColumn.getAlias());
```
labkey-susanh, line 136: Any chance alias is null here?

labkey-susanh (approved):

cnathe (approved):

api/src/org/labkey/api/data/LoggingResultSetWrapper.java
labkey-jeckels: It's possible, though unlikely. If so, I'd like to know because we haven't been able to repro the problem that inspired this additional detail.

PR #2742, opened 2021-11-02
44112: Gather more metrics on study

labkey-klum:
#### Rationale
Add additional study metrics per this [issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44112).

#### Changes
Additional metrics for:
- Number of studies with REDCap or CDISC repositories configured
- Number of studies created via Publish study
- Number of datasets of the following: demographics, "regular", and third key (split into managed third key vs user provided)
- Number of datasets created from assays
- Number of datasets created from samples
- Number of studies where alternate participant IDs and aliases are configured
- Number of studies where Study products are configured
- Number of studies where Treatments are configured
- Number of studies where Assay Schedule are configured

### DIFF
--- a/study/src/org/labkey/study/StudyModule.java
+++ b/study/src/org/labkey/study/StudyModule.java
@@ -40,6 +40,7 @@ import org.labkey.api.data.SqlSelector;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.TableSelector;
 import org.labkey.api.data.UpgradeCode;
+import org.labkey.api.data.dialect.SqlDialect;
 import org.labkey.api.data.views.DataViewService;
 import org.labkey.api.exp.LsidManager;
 import org.labkey.api.exp.api.ExperimentService;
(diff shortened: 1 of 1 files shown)


study/src/org/labkey/study/StudyModule.java
```diff
@@ -454,6 +455,29 @@ protected void startupAfterSpringConfig(ModuleContext moduleContext)
...
+
+                metric.put("assayScheduleCount", new SqlSelector(StudySchema.getInstance().getSchema(), "SELECT COUNT(DISTINCT(container)) FROM study.AssaySpecimen").getObject(Long.class));
+
+                SQLFragment sql = new SQLFragment("SELECT COUNT(DISTINCT(x.container)) ")
+                        .append("FROM (")
+                        .append("SELECT container FROM study.Participant WHERE AlternateId IS NOT NULL ")
+                        .append("UNION ")
+                        .append("SELECT container FROM study.Study WHERE AlternateIdPrefix IS NOT NULL ")
+                        .append(") x");
+                metric.put("alternateParticipantIdCount", new SqlSelector(StudySchema.getInstance().getSchema(), sql).getObject(Long.class));
```
labkey-klum: I wasn't 100% sure on the behavior of the alternate participant ID but it seems like we have to look at both the participant and study tables. If a prefix is configured the `participant.alternateId` field is not populated until an export occurs but if an alias map is uploaded then it is populated immediately.

study/src/org/labkey/study/StudyModule.java
```diff
@@ -514,6 +540,16 @@ protected void startupAfterSpringConfig(ModuleContext moduleContext)
...
+                        TableInfo products = schema.getTable(StudyQuerySchema.PRODUCT_TABLE_NAME);
+                        long productCount = new TableSelector(products).getRowCount();
+                        if (productCount > 0)
+                            hasProducts.increment();
+
+                        TableInfo treatments = schema.getTable(StudyQuerySchema.TREATMENT_TABLE_NAME);
+                        long treatmentCount = new TableSelector(treatments).getRowCount();
+                        if (treatmentCount > 0)
+                            hasTreatments.increment();
+
```
labkey-klum, line 551: These tables are provisioned so can't execute direct SQL to get the counts.

study/src/org/labkey/study/StudyModule.java
```diff
@@ -454,6 +455,29 @@ protected void startupAfterSpringConfig(ModuleContext moduleContext)
                 metric.put("linkedSampleTypeDatasetCount", new SqlSelector(StudySchema.getInstance().getSchema(), "SELECT COUNT(PublishSourceType) FROM study.dataset WHERE PublishSourceType = 'SampleType'").getObject(Long.class));
                 metric.put("linkedAssayDatasetCount", new SqlSelector(StudySchema.getInstance().getSchema(), "SELECT COUNT(PublishSourceType) FROM study.dataset WHERE PublishSourceType = 'Assay'").getObject(Long.class));
 
+                metric.put("redcapCount", new SqlSelector(PropertySchema.getInstance().getSchema(), "SELECT COUNT(*) FROM prop.PropertySets WHERE Category = 'RedcapConfigurationSettings'").getObject(Long.class));
+                metric.put("publishStudyCount", new SqlSelector(PropertySchema.getInstance().getSchema(), "SELECT COUNT(DISTINCT(destination)) FROM study.StudySnapshot WHERE Type = 'publish'").getObject(Long.class));
+
```
labkey-adam, line 463: This might not have been on the list, but shouldn't we count the "ancillary" studies as well?

study/src/org/labkey/study/StudyModule.java
labkey-adam: On my machine, the second condition (AlternateIdPrefix IS NOT NULL) captures a lot more containers than the first. I think that's because the alias map and prefix are copied in the publish process, so every published study has them whereas only source studies have AlternateIds in the Participants table. You might check with Hannah, but I think it'd be best to separate these into distinct metrics to avoid confusion.

labkey-adam (changes requested):
A couple suggestions... not strictly required, but worth considering

study/src/org/labkey/study/StudyModule.java
labkey-klum: Spoke to Hannah, and she does want to track both (but separately)

https://github.com/LabKey/platform/pull/2742/commits/d637eddabde2dbe37684ddd9aa242760d51e5662

study/src/org/labkey/study/StudyModule.java
labkey-klum: Yes, agreed https://github.com/LabKey/platform/pull/2742/commits/d637eddabde2dbe37684ddd9aa242760d51e5662

labkey-adam (approved):

PR #2746, opened 2021-11-04
Study schema contextual roles

labkey-matthewb:
#### Rationale
Flush out TableInfo.hasPermission(ReadPermission) implementations
use contextual roles for permission elevation (study schema)

Also, back-port WebSocket warning change so @cnathe doesn't have to

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/study/src/org/labkey/study/StudyUnionTableInfo.java
+++ b/study/src/org/labkey/study/StudyUnionTableInfo.java
@@ -75,7 +75,7 @@ public class StudyUnionTableInfo extends VirtualTable
 
     public StudyUnionTableInfo(StudyImpl study, Collection<DatasetDefinition> defs, User user, boolean crossContainer)
     {
-        super(StudySchema.getInstance().getSchema(), "StudyData");
+        super(StudySchema.getInstance().getSchema(), "StudyData", null);
         _study = study;
         _user = user;
         _crossContainer = crossContainer;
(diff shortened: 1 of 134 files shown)


study/src/org/labkey/study/StudyUnionTableInfo.java
```diff
@@ -360,4 +368,10 @@ protected ColumnInfo resolveColumn(String name)
 
         return result;
     }
+
+    @Override
+    public boolean hasPermission(@NotNull UserPrincipal user, @NotNull Class perm)
+    {
+        return super.hasPermission(user, perm);
+    }
```
labkey-adam, line 376: Why this override?

study/src/org/labkey/study/assay/StudyPublishManager.java
```diff
@@ -1394,7 +1394,7 @@ public ExpProtocol ensureStudyPublishProtocol(User user, Container container, St
             {
                 // Don't enforce permissions for the current user - we still want to tell them if the data
```
labkey-adam, line 1396: Bad comment? Aren't we enforcing permissions now?

labkey-adam (approved):

study/src/org/labkey/study/assay/StudyPublishManager.java
labkey-matthewb: Thanks

PR #2722, opened 2021-10-28
Item 9537: Sample Status, Remove experimental feature flag

cnathe:
#### Rationale
We are ready to enable the sample status feature for the LKB, LKSM, and LKFM apps. This set of PRs removes that experimental feature flag and turns on the sample status column/field for various grid displays, insert/update forms, etc.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/661
* https://github.com/LabKey/platform/pull/2722
* https://github.com/LabKey/sampleManagement/pull/734
* https://github.com/LabKey/biologics/pull/1037
* https://github.com/LabKey/inventory/pull/327
* https://github.com/LabKey/testAutomation/pull/893

#### Changes
* remove sample status experimental feature flag
* update check in sample confirm delete modal to check for SM module instead of experimental feature flag
* add sample status column to picklist default view

PR #2747, opened 2021-11-04
DataStatesTableInfo QUS fix for validate label check in updateRows case

cnathe:
#### Rationale
This issue was hit by the python API integration tests. If the API makes a call to update rows for the core.DataStates table and is only updating the state's description, the QUS validation check was saying that "Label cannot be blank". This fix is so that the validate label check doesn't apply in the updateRows case when the label is not being changed.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2722

#### Changes
* DataStatesTableInfo validateLabel() param for `allowMissing` in the updateRows case

### DIFF
--- a/core/src/org/labkey/core/query/DataStatesTableInfo.java
+++ b/core/src/org/labkey/core/query/DataStatesTableInfo.java
@@ -105,16 +105,16 @@ public class DataStatesTableInfo extends FilteredTable<CoreQuerySchema>
             return null;
         }
 
-        private boolean validateLabel(Map<String, Object> row)
+        private boolean validateLabel(Map<String, Object> row, boolean allowMissing)
         {
             String label = (String) row.get("label");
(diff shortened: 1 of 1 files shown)


labkey-susanh (approved):
Change seems fine, but do we want this to target 21.11 perhaps?  There's a workaround (provide the label in the API call) and the use case is probably not too common, but it's a small enough change that it might be easier to put in the release and avoid the chance of a need for the workaround until 22.3 for folks.

cnathe:
> Change seems fine, but do we want this to target 21.11 perhaps? There's a workaround (provide the label in the API call) and the use case is probably not too common, but it's a small enough change that it might be easier to put in the release and avoid the chance of a need for the workaround until 22.3 for folks.

Good point. I'll change this to target release21.11 and re-request review. Thanks.

PR #2750, opened 2021-11-04
Item 9090: Naming pattern improvements - lineage lookup

XingY:
#### Rationale
Parent (data or material) fields are both supported in name expressions, but only the name value of the parents can be used and not other properties. A single level of looks up is also supported in name expressions, but that doesn't apply to parent data. This PR add support to use parent data's non name properties in name expression.

#### Related Pull Requests
* 

#### Changes
* Support lineage lookup expression
* Update collection SubstitutionFormat to not error out if input value is not a collection


api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -407,6 +477,143 @@ private String genName(@NotNull Map<String, Object> rowMap,
...
+                    return parentObject.getCreated();
+                case "modified":
+                    return parentObject.getModified();
+                case "createdby":
+                    return parentObject.getCreatedBy();
+                case "modifiedby":
+                    return parentObject.getModifiedBy();
+                default:
+                {
+                    Map<String, Object> properties = new CaseInsensitiveHashMap<>();
```
labkey-susanh, line 508: Do you need the map?  Can you filter `getObjectProperties()` to find the one where `field.equalsIgnoreCase(prop.getName())`?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -407,6 +477,143 @@ private String genName(@NotNull Map<String, Object> rowMap,
...
+                {
+                    if (parentObject instanceof ExpMaterial)
+                        return ((ExpMaterial) parentObject).getDescription();
+                    else if (parentObject instanceof ExpData)
+                        return ((ExpData) parentObject).getDescription();
+                }
+                case "created":
+                    return parentObject.getCreated();
+                case "modified":
+                    return parentObject.getModified();
```
labkey-susanh, line 499: I assume the date values get formatted elsewhere, right?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -407,6 +477,143 @@ private String genName(@NotNull Map<String, Object> rowMap,
...
+                    return parentObject.getCreatedBy();
+                case "modifiedby":
+                    return parentObject.getModifiedBy();
+                default:
+                {
+                    Map<String, Object> properties = new CaseInsensitiveHashMap<>();
+                    parentObject.getObjectProperties().values().forEach(prop -> {
+                        properties.put(prop.getName(), prop.getObjectValue());
+                    });
+                    return properties.get(field);
```
labkey-susanh, line 527: If we don't find this property and return null here, will that be handled gracefully somewhere else?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -407,6 +477,143 @@ private String genName(@NotNull Map<String, Object> rowMap,
...
+                return;
+
+
+            boolean hasTypeLookup = false;
+            if (_expLineageLookupFields.containsKey("Inputs"))
+                hasTypeLookup = true;
+
+            if (!hasTypeLookup && isMaterialParent)
+            {
+                if (_expLineageLookupFields.containsKey("MaterialInputs"))
```
labkey-susanh: Use constant here and below?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -437,50 +644,102 @@ private String genName(@NotNull Map<String, Object> rowMap,
...
                     String[] parts = colName.split("/", 2);
+
                     if (parts.length == 2)
                     {
-                        addInputs(parts, colName, value, inputs, parentImportAliases);
+                        if (_exprHasLineageInputs)
+                            addInputs(parts, colName, value, inputs, parentImportAliases);
+                        if (_exprHasLineageLookup)
+                        {
+                            boolean isParent = parts[0].equalsIgnoreCase(ExpMaterial.MATERIAL_INPUT_PARENT) || parts[0].equalsIgnoreCase(ExpData.DATA_INPUT_PARENT);
```
labkey-susanh: A small optimization, but you could avoid doing the compare to `MATERIAL_INPUT_PARENT` twice:
```
bool isMaterialParent = parts[0].equalsIgnoreCase(ExpMaterial.MATERIAL_INPUT_PARENT);
bool isDataParent = parts[0].equalsIgnoreCase(ExpData.DATA_INPUT_PARENT);

if (isMaterialParent || isDataParent) 
...
```

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -437,50 +644,102 @@ private String genName(@NotNull Map<String, Object> rowMap,
...
+                            if (isParent)
+                            {
+                                boolean isMaterialParent = parts[0].equalsIgnoreCase(ExpMaterial.MATERIAL_INPUT_PARENT);
+                                for (String parent : parentNames(value, colName))
+                                    addLineageLookupContext(parts[1], parent, isMaterialParent, parentImportAliases, inputLookupValues);
+                            }
+                        }
```
labkey-susanh: Same logic as above, so could be refactored.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -407,6 +477,143 @@ private String genName(@NotNull Map<String, Object> rowMap,
...
+                case "modified":
+                    return parentObject.getModified();
+                case "createdby":
+                    return parentObject.getCreatedBy();
+                case "modifiedby":
+                    return parentObject.getModifiedBy();
+                default:
+                {
+                    Map<String, Object> properties = new CaseInsensitiveHashMap<>();
+                    parentObject.getObjectProperties().values().forEach(prop -> {
```
labkey-susanh, line 509: It looks like this is going to make a query for every name we're generating, which could slow things down considerably.  Is the `materialCache` helping us out here somehow?  Otherwise, could we cache the properties? I'm thinking of the case when many samples derived from the same source or aliquoted from the same parent are being generated.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -141,6 +164,21 @@ else if (value instanceof Collection)
                 .filter(s -> !s.isEmpty());
     }
 
+    private boolean isLineageInput(Object token, @Nullable Map<String, String> importAliases)
```
labkey-susanh: This seems like a method that might be generally useful.  We do the comparison to DataInputs and MaterialInputs in various places anyway, so it might deserve to be a static method somewhere else.  The other places where we do this kind of comparison may not care about the `Inputs` case, but that could be parameterized.  Just a thought, though.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -141,6 +164,21 @@ else if (value instanceof Collection)
...
+    {
+        String sTok = token.toString().toLowerCase();
```
labkey-susanh: Again, just a small optimization, but since you're using `equalsIgnoreCase` below, the `toLowerCase` is superfluous. It will do that again in `equalIgnoreCase`.

api/src/org/labkey/api/data/NameGenerator.java
XingY: Yep, it's formatted at a later step

labkey-susanh (approved):

PR #2760, opened 2021-11-08
Item 9575: Update folder import and export to support sample status

labkey-susanh:
#### Rationale
We have introduced the ability for users to set statuses on samples to control the actions that are allowed on the samples.  Here we assure that the statuses of the samples as well as the configured possible SampleState values are included in folder import and export. 

#### Related Pull Requests
#2722 

#### Changes
* Update `qcStates.xsd` to include stateType field and allow for use without some of the qc state parameters
* Add `getRemapMissingBehavior` field to ColumnInfo so columns can override the default behavior and, in particular, throw an error when a status name that is not currently defined is associated with a sample being imported  
* Update labeling in Folder Export UI from "QC State Settings" to "Sample Status and QC State Settings" when sample status is supported
* Rename some classes from QC* to DataState*
* Update `SampleTypeAndDataClassFolderWriter` to write a new file for each sample type to record the status and add `SampleStatusImporter` to import these status values after all other data has been imported


api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -1255,19 +1255,20 @@ public int addConvertColumn(String name, int fromIndex, JdbcType toType, @Nullab
...
     }
 
     public SimpleConvertColumn createConvertColumn(@NotNull ColumnInfo col, int fromIndex, boolean useOriginalValueOnRemapFailure)
     {
-        return createConvertColumn(col, fromIndex, NO_MV_INDEX, null, col.getPropertyType(), col.getJdbcType(), useOriginalValueOnRemapFailure);
+        return createConvertColumn(col, fromIndex, NO_MV_INDEX, null, col.getPropertyType(), col.getJdbcType(), useOriginalValueOnRemapFailure, null);
     }
 
-    private SimpleConvertColumn createConvertColumn(@NotNull ColumnInfo col, int fromIndex, int mvIndex, @Nullable PropertyDescriptor pd, @Nullable PropertyType pt, @Nullable JdbcType type, boolean useOriginalValueOnRemapFailure)
+    private SimpleConvertColumn createConvertColumn(@NotNull ColumnInfo col, int fromIndex, int mvIndex, @Nullable PropertyDescriptor pd, @Nullable PropertyType pt, @Nullable JdbcType type, boolean useOriginalValueOnRemapFailure, @Nullable RemapMissingBehavior remapMissingBehavior)
```
XingY: It feels weird to have both useOriginalValueOnRemapFailure and remapMissingBehavior params. Is it possible to refactor/remove useOriginalValueOnRemapFailure and pass in remapMissingBehavior = RemapMissingBehavior.OriginalValue?

experiment/src/org/labkey/experiment/XarReader.java
```diff
@@ -1361,43 +1385,53 @@ private ExpMaterial loadMaterial(MaterialBaseType xbMaterial,
...
         String description = output.getClass().getSimpleName();
         String lsid = output.getLSID();
         boolean changed = false;
+        boolean isAliquot = aliquotedFromLSID != null;
+        boolean hadSource = false;
+        boolean hadRun = false;
 
         getLog().debug("Found an existing entry for " + description + " LSID " + lsid + ", not reloading its values from scratch");
 
+        // if the output is an aliquot, its source and run have already been wired up during the import of the tsv file
```
XingY: Would import for aliquot be broken if "AliquotedFrom" is absent? My understanding is that the aliquot relationship is defined by the runs, and the runs and inputs/outputs should round trip correctly, much like how derivatives can be exported/imported without extra info. 

In the case of a single csv import, "AliquotedFrom" is required due to the absence of the runs. But I don't see the need here for folder import.

experiment/src/org/labkey/experiment/XarReader.java
```diff
@@ -1407,9 +1441,10 @@ private void updateSourceInfo(RunItem output, Integer sourceApplicationId,
                 getLog().debug("Updating " + description + " with aliquot root LSID");
 
                 String newRootLsid = rootMaterial != null ? rootMaterial.getLSID() : rootMaterialLSID;
+                // When importing over existing samples, the LSIDs will never match, so we only log an info message here and don't update.
                 if (((Material) output).getRootMaterialLSID() != null && !((Material) output).getRootMaterialLSID().equals(rootMaterialLSID))
                 {
-                    throw new XarFormatException(description + " with LSID '" + lsid + "' already has aliquot root material LSID of " + ((Material) output).getRootMaterialLSID() + "; cannot set it to " + newRootLsid);
+                    getLog().info(description + " with LSID '" + lsid + "' already has root material LSID of " + ((Material) output).getRootMaterialLSID() + "; not updating to " + newRootLsid + ".");
```
XingY, line 1469: With this change, will the scenario of importing S1-1 (an existing aliquot, whose parent used to be S-1), but update its parent to S1 error out?

experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
```diff
@@ -285,6 +277,52 @@ private void writeDataClassDataFiles(Set<ExpDataClass> dataClasses, ImportContex
...
+        try (TSVGridWriter tsvWriter = new TSVGridWriter(factory))
+        {
+            tsvWriter.setApplyFormats(false);
+            tsvWriter.setColumnHeaderType(ColumnHeaderType.FieldKey);
+            PrintWriter out = dir.getPrintWriter(baseName + ".tsv");
+            tsvWriter.write(out);
+        }
+    }
+
+    private ColumnInfo getAliquotedFromNameColumn(TableInfo tinfo)
```
XingY, line 292: Again, I don't think this is needed/should be added.

experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
```diff
@@ -285,6 +277,52 @@ private void writeDataClassDataFiles(Set<ExpDataClass> dataClasses, ImportContex
...
+
+        // SampleState
+        // substitute the Label value for the RowId lookup value
+        FieldKey statusFieldKey = FieldKey.fromParts(ExpMaterialTable.Column.SampleState.name(), "Label");
+        Map<FieldKey, ColumnInfo> select = QueryService.get().getColumns(tinfo, Collections.singletonList(statusFieldKey));
+        ColumnInfo statusAlias = new AliasedColumn(tinfo, ExpMaterialTable.Column.SampleState.name(), select.get(statusFieldKey));
+
+        columns.add(statusAlias);
+
+        // In order to update status values for Aliquots, the AliquotedFrom field, which is the
```
XingY, line 319: Why is AliquotedFromLSID not sufficient?

XingY (commented):
I'm not clear why AliquotedFrom is needed in export. Is there a scenario that the import fails to create aliquots correctly? If that's the case,  I think the fix should be on the import end. Issue 44277 addressed one case when aliquots are created without AliquotedFrom. I wouldn't be surprised if there are other places that needs additional work.

experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
labkey-susanh: This is definitely needed for the import of the status values, which happens after the samples are created and after all the assay run and storage data have been imported.  You can't current update samples without the `AliquotedFrom` field in place.

XingY: I think probably the current behavior should be updated to also accept AliquotedFromLSID. Is AliquotedFromLSID column included in the folder export file? I can look into allowing either AliquotedFromLSID or AliquotedFrom in my api branch, if you prefer.

experiment/src/org/labkey/experiment/XarReader.java
labkey-susanh: This wasn't quite the right fix, but I believe it's been address now with commit 77828846c64d93144f99dc4588de7fb21a0eda1c.  

The original import succeeds without the `AliquotedFrom` field, but a subsequent re-import, or an import over existing aliquots would fail when trying to read the tsv included in the folder archive.

experiment/src/org/labkey/experiment/XarReader.java
labkey-susanh: It will not produce an error anymore, but it will also not update the parent and will tell you about that, which seems like proper behavior to me, particularly since this now allows you to do a reimport that includes aliquots.

experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
labkey-susanh: AliquotedFromLSID is also not in the TSV.  Do you see the inclusion of the AliquotedFrom column that uses the name as problematic or error-prone?

experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
labkey-susanh: This is not the field the import is currently looking for, and it seemed less complicated to use the name in the column for the export file than to change the import code to also handle the LSID field.

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
labkey-susanh: Ah yes, much better now.  Thanks.  See fe3a19bc00528b31625a7d8d5e2f4e54eb56c5d7.

experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderWriter.java
XingY: I don't have strong objection to AliquotedFrom, it just seemed redundant since the import zip already have all the runs info. Also I thought the export already included material.AliquotedFromLSID, which is just as good as AliquotedFrom, but seems that's not the case. I'm fine with AliquotedFrom.

XingY (approved):

api/schemas/qcStates.xsd
```diff
@@ -56,22 +58,28 @@
...
                                             <xs:documentation>
-                                                 Indicates whether this dataset contains demographic data.
+                                                 Indicates whether data in this QC state is publicly available or not.
+                                            </xs:documentation>
+                                        </xs:annotation>
+                                    </xs:attribute>
+                                    <xs:attribute name="type" type="qc:stateTypeEnum">
+                                        <xs:annotation>
+                                            <xs:documentation>
+                                                The type of the data state.
```
cnathe, line 82: this refers to "data state" but everything else above refers to "QC state". not sure which should be updated.

api/schemas/qcStates.xsd
```diff
@@ -82,5 +90,13 @@
...
 			</xs:all>
 		</xs:complexType>
-	</xs:element>
+    </xs:element>
+
+    <xs:simpleType name="stateTypeEnum">
+        <xs:restriction base="xs:string">
+            <xs:enumeration value="Available"/>
+            <xs:enumeration value="Consumed"/>
+            <xs:enumeration value="Locked"/>
```
cnathe, line 99: Should we put a comment in ExpSchema.SampleStateType to say that this xsd info needs to stay in sync with the types on the java side?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -709,7 +709,7 @@ public boolean next() throws BatchValidationException
             if (hasNext)
             {
                 String lsid = (String) get(_lsidCol);
-                if (_aliquotParentCol > -1)
+                if (_aliquotParentCol > -1 && !_context.getConfigParameterBoolean(SampleTypeService.ConfigParameters.DeferAliquotRuns))
```
cnathe, line 712: should this / can this use the `_skipAliquot` variable instead of getting the context config again?

experiment/src/org/labkey/experiment/XarReader.java
```diff
@@ -1361,15 +1385,19 @@ private ExpMaterial loadMaterial(MaterialBaseType xbMaterial,
...
     {
         String description = output.getClass().getSimpleName();
         String lsid = output.getLSID();
         boolean changed = false;
+        boolean isAliquot = aliquotedFromLSID != null;
 
         getLog().debug("Found an existing entry for " + description + " LSID " + lsid + ", not reloading its values from scratch");
 
+        // if the output is an aliquot, its source and run have already been wired up during the import of the tsv file
+        // by virtue of the AliquotedFrom column being included in that file.  For importing legacy folder imports where the
```
cnathe: is this comment still accurate after your latest changes to defer the aliquot connection part until the run's import?

experiment/src/org/labkey/experiment/samples/SampleStatusFolderImporter.java
```diff
@@ -0,0 +1,134 @@
...
+                    try
+                    {
+                        typesXarSource.init();
+                    }
+                    catch (Exception e)
+                    {
+                        log.error("Failed to initialize types XAR source", e);
+                        throw(e);
+                    }
+                    XarReader typesReader = new XarReader(typesXarSource, job);
```
cnathe: all of this handling of the `typesXarFile`, `typesXarSource`, and `typesReader` seems duplicated from the SampleTypeAndDataClassFolderImporter.java. Is there any chance they could share that part of the code? No problem if that isn't feasible.

cnathe: I just saw that this class does extend `SampleTypeAndDataClassFolderImporter` so I'm hopeful it isn't too difficult to refactor and share some of that parsing.

experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderImporter.java
```diff
@@ -256,6 +261,7 @@ private void importTsvData(ImportContext ctx, String schemaName, List<String> ta
                                     {
                                         log.error("Unable to get audit behavior for import. Default behavior will be used.");
                                     }
+                                    options.put(SampleTypeService.ConfigParameters.DeferAliquotRuns, true);
```
cnathe, line 276: this defer implementation looks to have worked out well

cnathe (approved):
A few comments and questions but overall looks good.

api/schemas/qcStates.xsd
labkey-susanh: Since no one seemed to notice that this referred to datasets previously, I'm going to switch them all to data state and not worry any further.

api/schemas/qcStates.xsd
labkey-susanh: Done

experiment/src/org/labkey/experiment/XarReader.java
labkey-susanh: Updated. Thanks.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Maybe, but this seemed safer and not expensive.  `skipAliquot` comes via a different pathway.

experiment/src/org/labkey/experiment/samples/SampleStatusFolderImporter.java
labkey-susanh: Did one further bit of refactoring.

labkey-susanh (commented):
I forgot to submit my comments, so I do it now for posterity.

PR #2769, opened 2021-11-11
Update dataset shadowing test to match new behavior

labkey-tchad:
#### Rationale
We now prevent creation of a dataset with an ID that collides with the ID of a shared dataset. This update explicitly checks for that error. We do not prevent dataset ID collisions when creating a shared dataset however, so we can verify the shadowing warnings by creating the datasets in the opposite order.

#### Related Pull Requests
* #2764 

#### Changes
* Update dataset shadowing test to match new behavior


labkey-tchad:
Verifying fix: https://teamcity.labkey.org/buildConfiguration/LabKey_2111Release_Community_BvtAPostgres/1608186

labkey-tchad:
Fixing the fix: https://teamcity.labkey.org/buildConfiguration/LabKey_2111Release_Community_BvtAPostgres/1608287

labkey-chrisj (approved):

PR #2778, opened 2021-11-15
44297: email links to announcements aren't properly encoded

labkey-klum:
#### Rationale
This change addresses the encoding issue found in the Biologics/announcements email notification:

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44297

#### Changes
Instead of the original, more general change, to encode links. The issue detail link is encoded specifically in the issue update template and marked as an HTML content type so it won't get escaped later.


labkey-alan (approved):
Looks good

PR #2793, opened 2021-11-22
Fix sample status import when sample type has required and MV fields

labkey-susanh:
#### Rationale
During a folder import, the sample TSV files are imported use the MERGE option for insert.  Normally, this means that we check for the presence of required fields and the missing value indicators, if applicable.  We've recently split the sample type import into two files, one containing just the fields needed for adding status to the samples after all other data in the import has been registered.  We do not need to check for the required fields in this second TSV file, so we add a new data iterator config parameter to allow us to skip those checks.

#### Related Pull Requests
* #2760 

#### Changes
* Indicate that checks for required fields should be skipped during status TSV import.


cnathe (approved):
Nice. This is very clean and seems to have worked out well from the options we discussed!

PR #2800, opened 2021-11-23
Issue 44307: ReplacedByRun not getting set in folder export/import

labkey-jeckels:
#### Rationale
We want to include run versioning info in the export/import process

#### Changes
* Include replacement info in XAR schema
* Write during export
* Retain during import and resolve references when possible after all runs have been created

### DIFF
--- a/experiment/src/org/labkey/experiment/XarReader.java
+++ b/experiment/src/org/labkey/experiment/XarReader.java
@@ -123,6 +123,8 @@ public class XarReader extends AbstractXarImporter
     private final Set<String> _experimentLSIDs = new HashSet<>();
     private final Map<String, Integer> _propertyIdMap = new HashMap<>();
     private final Map<Integer, String> _runWorkflowTaskMap = new HashMap<>();
+    /** Retain replacement info so we can wire them up after all runs have been imported */
+    private final Map<Integer, String> _runReplacedByMap = new HashMap<>();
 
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/XarReader.java
```diff
@@ -428,6 +432,29 @@ else if (Files.exists(path))
...
+                if (run == null)
+                {
+                    throw new XarFormatException("Could not find previously imported run with RowId " + entry.getKey());
+                }
+                run.setReplacedByRun(replacedByRun);
+                run.save(getUser());
+            }
+            else
+            {
+                getLog().warn("Could not resolve replacement run " + entry.getValue() + " so it will not be referenced.");
```
labkey-susanh, line 453: Is there any value in distinguishing the cases of not finding the run at all and not finding it in this container?

labkey-susanh (approved):

experiment/src/org/labkey/experiment/XarReader.java
labkey-jeckels: I considered it but ultimately decided to not bother, as the usual scenario includes the current folder's RowId in the LSID so it's unlikely it would resolve to anything but the current container. And now I've spent as much time explaining my reasoning here as I would have added a second log message. :)

PR #2803, opened 2021-11-23
44085: Large assay results link-to-study in UI fails

labkey-klum:
#### Rationale
When users try to link large numbers of either assay or sample type data to a study, if they exceed the 10,000 Tomcat POST parameter limit, the UI can fail in a number of ways. Rather than try to increase the current limit of around 9950 rows or try to refactor the UI to allow an arbitrary number of rows, the decision is to set a max ceiling and don't allow the user to proceed any further while displaying a message with a help link that will show them options to help work around this limit.

More details in this and linked issues:
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44085

#### Changes
- Add a hardcoded 9950 row limit
- For individual rows, if the limit is hit, they won't get past the study selection page and will see an error message
- For selected runs, if the limit is hit, the auto link check box is selected. If they toggle off the check box then won't be allowed to proceed and the same error message will be shown.


api/src/org/labkey/api/study/publish/publishChooseStudy.jsp
```diff
@@ -182,6 +202,15 @@
 </script>
 
 <%
+    if (exceedsMaxRows)
+    {
+%>
+    <span class="labkey-error" id="row-exceeded-msg" <%=!bean.getBatchIds().isEmpty() ? HtmlString.unsafe("style = 'display:none;'") : HtmlString.unsafe("")%>><h4>
+        The maximum number of rows (<%=AbstractPublishStartAction.MAX_ROWS_TO_LINK%>) that can be linked at one time has been exceeded. You can review this <%=new HelpTopic("publishAssayData#trouble").getSimpleLinkHtml("help topic")%> for more information on how to work around this limit.
```
cnathe:
```suggestion
        The maximum number of rows, i.e. (<%=AbstractPublishStartAction.MAX_ROWS_TO_LINK%>), that can be linked at one time has been exceeded. You can review this <%=new HelpTopic("publishAssayData#trouble").getSimpleLinkHtml("help topic")%> for more information on how to work around this limit.
```

api/src/org/labkey/api/study/publish/publishChooseStudy.jsp
```diff
@@ -127,6 +131,22 @@
             }
         };
 
+        // invoked when the auto link checkbox is changed, if the number of rows has been exceeded we only allow
+        // the auto link route and show/hide the max rows error message.
+        toggleAutoLink = function() {
+            if (exceedsMaxRow) {
+                let errorMsg = $("span[id='row-exceeded-msg']");
+                if ($("input[id='autoLink']").prop('checked')){
```
cnathe, line 139: does the `onChange` handler pass through the checkbox state to the function here? if so then we wouldn't need to get it again here.

cnathe (approved):
Looks good, just a couple of minor comments/questions.

api/src/org/labkey/api/study/publish/publishChooseStudy.jsp
labkey-klum: It doesn't get passed into the function, but the selector : `$("input[id='autoLink']").prop('checked')` gets the checked state of the checkbox. Perhaps I could have been clearer with a variable declaration rather than using the inline selector.

PR #2806, opened 2021-11-24
Item 9704: Sample Status Polish - Dashboard stacked bar chart and minor styling/wording updates

cnathe:
#### Rationale
LKSM and LKB app dashboards are adding a stacked bar chart for viewing the sample counts of sample types by status. This feature needed a stacked bar chart geom in the D3 vis API library.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2806
* https://github.com/LabKey/biologics/pull/1071
* https://github.com/LabKey/inventory/pull/335
* https://github.com/LabKey/sampleManagement/pull/762
* https://github.com/LabKey/labkey-ui-components/pull/680

#### Changes
* D3 vis API support for stacked bar plot geom with example in demo page
* related changes in plot.js to calculated y-axis domain max for the stacked bar chart case


internal/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -2926,6 +2926,146 @@ LABKEY.vis.internal.D3Renderer = function(plot) {
...
+        data = stackData.reduce(function(prev, current) {
+            // keep track of which data points are from the last set, for the showValues case below
+            for (var index in prev) {
+                prev[index].showValue = false;
+            }
+            for (var index in current) {
+                current[index].showValue = true;
+            }
+            return prev.concat(current);
+        }, []);
```
labkey-alan: Instead of iterating through all of the previous values on every pass, couldn't we just detect when we're on the last set, and then set the values?
```suggestion
        data = stackData.reduce(function(prev, current, index) {
            // Only showValue for the last set
            if (index === stackData.length - 1) {
                prev.forEach(function (item) {
                    item.showValue = false;
                });
    
                current.forEach(function (item) {
                   item.showValue = true;
                });
            }
            return prev.concat(current);
        }, []);

```

internal/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -2926,6 +2926,146 @@ LABKEY.vis.internal.D3Renderer = function(plot) {
...
+            return geom.getXSub(d) - offsetWidth;
+        };
+
+        yAcc = function(d){
+            return geom.getY(d)
+        };
+
+        colorAcc = geom.fill;
+        if (geom.colorAes && geom.colorScale)
+        {
```
labkey-alan: nit: inconsistent brace placement here.

internal/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -2926,6 +2926,146 @@ LABKEY.vis.internal.D3Renderer = function(plot) {
...
+                .attr('x', function(d) {
+                    return xAcc(d);
+                })
+                .attr('y', function(d) {
+                    var verticalOffset = yAcc({ value: d.y0 });
+                    return verticalOffset - heightFn(d);
+                })
+                .attr('width', barWidth).attr('height', heightFn)
+                .attr('stroke', geom.color).attr('stroke-width', geom.lineWidth)
+                .attr('fill', colorAcc).attr('fill-opacity', geom.opacity);
```
labkey-alan, line 3038: I feel like we should be more consistent with how we place calls to `.attr` but I don't know if that's how the styling is elsewhere in this code since it's been a long while for me.
```suggestion
                .attr('width', barWidth)
                .attr('height', heightFn)
                .attr('stroke', geom.color)
                .attr('stroke-width', geom.lineWidth)
                .attr('fill', colorAcc)
                .attr('fill-opacity', geom.opacity);

```

internal/webapp/vis/src/geom.js
```diff
@@ -484,6 +484,7 @@ LABKEY.vis.Geom.DataspaceBoxPlot.prototype.render = function(renderer, grid, sca
  * @param {String} [config.fill] (Optional) A string value used for the fill color in the bar plot. Defaults to white (#ffffff)
  * @param {Number} [config.lineWidth] (Optional) A used to set the width of the lines used in the bar plot. Defaults to 1.
  * @param {Number} [config.opacity] (Optional) A number between 0 and 1 used to set the opacity of the bar plot. Defaults to 1.
+ * @param {Boolean} [config.stacked] (Optional) True to show stacked bars for the individual bar groupings.
```
labkey-alan: Should we maybe add some more details about what else is required in order to set stacked to true?

labkey-alan (approved):
Looks good, I added a few minor pieces of feedback.

internal/webapp/vis/src/internal/D3Renderer.js
cnathe: yeah that might work well. I'll give it a try

internal/webapp/vis/src/internal/D3Renderer.js
cnathe: done

internal/webapp/vis/src/internal/D3Renderer.js
cnathe: yeah, doesn't look like it is very consistent across this file but I will at least clean it up for this new function

internal/webapp/vis/src/geom.js
cnathe: done

internal/webapp/vis/src/internal/D3Renderer.js
cnathe: done

PR #2808, opened 2021-11-24
Pass through container when getting old and new record maps for audit logs

labkey-susanh:
#### Rationale
In order to record the label of the sample status when creating the audit log, we need to have the container.  We also need the map we look for this value in to be case-insensitive.  Because the insert of rows does not go through the `setOldAndNewMapsForUpdate` method, the row that comes through when inserting has not had its keys lower-cased. This is strange to me, but investigating further is beyond the scope of the current PR. 

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/759

#### Changes
* Pass through container to `setNewRecordMap` and `setOldRecordMap`.
* Use CaseInsensitiveHashMap for better checking of sampleState values.


cnathe (approved):
Looks safe enough to me.

PR #2816, opened 2021-11-29
Issue #44384: Folder w/ Sample Type can not be used as template

Sigmonia:
#### Rationale
Change to use Path objects fails to resolve the relative directory when using an existing project Sample Type folder as template
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44384

#### Changes
* Changed from using the FileUtil.getPath to Path.of --This should be safe as we are already working with a VirtualFile object

### DIFF
--- a/experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderImporter.java
+++ b/experiment/src/org/labkey/experiment/samples/SampleTypeAndDataClassFolderImporter.java
@@ -76,7 +76,8 @@ public class SampleTypeAndDataClassFolderImporter implements FolderImporter
 
         if (xarDir != null)
         {
-            Path xarDirPath = FileUtil.getPath(ctx.getContainer(), FileUtil.createUri(xarDir.getLocation()));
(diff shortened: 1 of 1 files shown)


cnathe (approved):
I'll pull this down today in 21.11 to do manual testing on the fb.

labkey-jeckels (commented):
Won't this break when the VirtualFile is either ZipFile or MemoryVirtualFile?

cnathe:
> Won't this break when the VirtualFile is either ZipFile or MemoryVirtualFile?

From my local manual testing of this change, the sample type import looks to work as expected now for the following cases:
1. create folder from template
2. import zip archive onto folder
3. import folder archive onto folder from pipeline
4. import folder archive using template folder option

labkey-jeckels:
Digging deeper, we end up having a bogus Path for both typesXarFile and runsXarFile in SampleTypeAndDataClassFolderImporter, but we only end up using them to try to resolve log files which are optional. Overall, I'm finding it very hard to follow this code (which predates this PR, of course), but it seems functional.

PR #2818, opened 2021-11-29
Publishing a date-based study with compliance turned on results in an NPE

labkey-adam:
#### Rationale
Fix the NPE. `DatasetSchemaTableInfo` shouldn't be thinking about PHI. And its callers shouldn't ask it to.

[Issue 44284: Publishing a date-based study with compliance turned on results in an NPE](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44284)


labkey-jeckels (commented):
No objections here but Matt is probably in a better position to review.

labkey-matthewb (approved):

PR #2819, opened 2021-11-30
Lineage columns don't return the display value for 8.3 API responses

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44337

The R API only supports the 8.3 response format (non-extended). When the response is generated, only the display column method `getJsonValue` is called whereas both `getDisplayValue` and `getJsonValue` are called in the 9.1 response format. The lineage display columns are designed to return the `objectid` unless the display value is requested. 

I think we want to return the display value for the lineage DC by calling `getJsonValues` and joining them together if there are multiples. It does mean that the 9.1 response format will no longer return the `objectid` but it seems like that was never useful.

### DIFF
--- a/experiment/src/org/labkey/experiment/api/LineageDisplayColumn.java
+++ b/experiment/src/org/labkey/experiment/api/LineageDisplayColumn.java
@@ -2,6 +2,7 @@ package org.labkey.experiment.api;
 
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.json.JSONArray;
 import org.labkey.api.collections.ResultSetRowMapFactory;
 import org.labkey.api.data.AbstractTableInfo;
 import org.labkey.api.data.BaseColumnInfo;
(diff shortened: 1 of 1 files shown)


cnathe (approved):

experiment/src/org/labkey/experiment/api/LineageDisplayColumn.java
```diff
@@ -182,6 +182,13 @@ public Object getDisplayValue(RenderContext ctx)
         return getDisplayValues(ctx).stream().map(o -> o == null ? " " : o.toString()).collect(Collectors.joining(", "));
     }
 
+    @Override
+    public Object getJsonValue(RenderContext ctx)
+    {
+        // issue: 44337. Doesn't seem to be a reason to return the object ID, even in the extended API response
+        return getJsonValues(ctx).stream().map(o -> o == null ? " " : o.toString()).collect(Collectors.joining(", "));
+    }
+
```
labkey-matthewb, line 192: Is it more correct to just return this as a JSONArray?  see ApiJsonWriter.writeObject()

labkey-matthewb (approved):

PR #2820, opened 2021-11-30
Issue 44262: Fix container filter and cell display for group audit logs for proper sorting and exporting

labkey-susanh:
#### Rationale
The Group Audit log table has fields that reference the `core.Principals` table, but because of the application of a container filter for the current container, which is the root container when looking at the audit logs, the names of users and groups were not resolving properly.  This led to empty values in the export and the sort functionality using the ids instead of the display names.  

Additionally, I've removed some logic that would not display certain group names in the audit log table.  We had been displaying only the Administrators group and project groups, but this leads to audit entries that reference the name of a group in the comment but then don't show that group in the field.  I'm unclear why we need to not display the group names that are referenced in the comments.

#### Changes
* Create `PrincipalUserDisplayColumn` for use in displaying both the Group and the Member column. 
* Change container filter for the Principal lookups for groups and users
* Remove logic that filters the names of groups displayed in the Group column.
* Other misc. clean up

### DIFF
--- a/api/src/org/labkey/api/audit/provider/GroupAuditProvider.java
+++ b/api/src/org/labkey/api/audit/provider/GroupAuditProvider.java
@@ -378,11 +378,16 @@ public class GroupAuditProvider extends AbstractAuditTypeProvider implements Aud
                         {
                             String displayText = PageFlowUtil.filter(u.getDisplayName(loggedInUser));
                             ActionURL url = UserManager.getUserDetailsURL(ctx.getContainer(), loggedInUser, id);
-                            out.write("<a href=\"");
-                            out.write(PageFlowUtil.filter(url));
-                            out.write("\">");
-                            out.write(displayText);
-                            out.write("</a>");
+                            if (url != null)
+                            {
+                                out.write("<a href=\"");
+                                out.write(PageFlowUtil.filter(url));
+                                out.write("\">");
+                                out.write(displayText);
+                                out.write("</a>");
+                            }
+                            else
+                                out.write(displayText);
                         }
                         else
                         {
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/audit/provider/GroupAuditProvider.java
```diff
@@ -332,28 +336,29 @@ public GroupForeignKey(UserSchema userSchema)
         public TableInfo getLookupTableInfo()
         {
             TableInfo tinfoUsers = CoreSchema.getInstance().getTableInfoPrincipals();
-            FilteredTable ret = new FilteredTable<>(tinfoUsers, _userSchema);
+            FilteredTable<UserSchema> ret = new FilteredTable<>(tinfoUsers, _userSchema);
+            ret.setContainerFilter(ContainerFilter.EVERYTHING);
```
labkey-klum: Do we need to set the CF here? Doesn't the change in `PrincipleIdForeignKey` take care of the primary issue.

labkey-klum (approved):

api/src/org/labkey/api/audit/provider/GroupAuditProvider.java
labkey-susanh: We also need on the GroupForeignKey for the sorting to work properly.  We show the group name in the grid by virtue of the overridden `renderGridCellContents`, but sorting still uses the filtered values for the names, which means it's like not having sorting at all since none of the groups are defined in the root container.

labkey-susanh: Oh, but now I see this class is pretty much the same as `PrincipleIdForeignKey`, so I think I can get rid of it.

PR #2821, opened 2021-11-30
Issue 44390: HTTP connection pool leak when remote ETLs fail before starting to iterate through results

labkey-jeckels:
#### Rationale
We leak HTTP connection on remote ETLs if there's a problem after we create the DataIteratorBuilder but before we use it to create the DataIterator. For example, a problem truncating the target table. This leaks a connection from the HTTP pool which can lead to eventually exhausting the pool and causing threads to hang forever.

#### Changes
* Don't execute the HTTP request until we're ready for the DataIterator, which will be closed as part of its lifecycle

### DIFF
--- a/internal/src/org/labkey/remoteapi/SelectRowsStreamHack.java
+++ b/internal/src/org/labkey/remoteapi/SelectRowsStreamHack.java
@@ -41,12 +41,20 @@ public class SelectRowsStreamHack
 {
     public static DataIteratorBuilder go(Connection cn, String container, SelectRowsCommand cmd, Container targetContainer) throws IOException, CommandException
     {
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
Kicked off ETL tests:

https://teamcity.labkey.org/buildConfiguration/LabKey_2111Release_Premium_ModulesSuites_DataIntegrationSqlserver/1622655

https://teamcity.labkey.org/buildConfiguration/LabKey_2111Release_Premium_ModulesSuites_DataIntegrationPostgres/1622653

labkey-adam (approved):

labkey-martyp (approved):

PR #2828, opened 2021-12-05
Issue 44425: StudyManager.ensureVisits() hits duplicate key constraint violation

labkey-adam:
#### Rationale
`StudyManager.ensureVisits()` needs to remember visits that it creates, otherwise it may attempt to create duplicates

### DIFF
--- a/study/src/org/labkey/study/model/StudyManager.java
+++ b/study/src/org/labkey/study/model/StudyManager.java
@@ -1102,6 +1102,8 @@ public class StudyManager
             if (result.getRowId() == 0)
             {
                 createVisit(study, user, result, visits);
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
I don't know if you're planning to use one PR for the other items you mentioned in the issue, but the current change seems fine.

labkey-adam:
This seems like the most important change for 21.11. The validator & list changes could go to develop, IMO, since there's an easy workaround and there may be test impact. @labkey-jeckels LMK if you feel strongly either way.

labkey-jeckels:
@labkey-adam handling the others in develop sounds fine to me. I think these are all long-standing problems.

PR #2831, opened 2021-12-06
Add ManageVersion column to core.Modules query

labkey-adam:
#### Rationale
This column will be convenient for managing schema version bumps, etc.

#### Changes
* Add ManageVersion column to core.Modules query
* Add link to schema browser when in root

### DIFF
--- a/core/src/org/labkey/core/query/ModulesTableInfo.java
+++ b/core/src/org/labkey/core/query/ModulesTableInfo.java
@@ -27,6 +27,7 @@ import org.labkey.api.data.JdbcType;
 import org.labkey.api.data.RenderContext;
 import org.labkey.api.data.SQLFragment;
 import org.labkey.api.data.dialect.DialectStringHandler;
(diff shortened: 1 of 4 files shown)


core/src/org/labkey/core/query/ModulesTableInfo.java
```diff
@@ -172,6 +174,7 @@ public SQLFragment getFromSQL(String alias)
             appendStringLiteral(h, cte,",",module.getUrl());
             appendStringLiteral(h, cte,",",module.getAuthor());
             appendStringLiteral(h, cte,",",module.getMaintainer());
+            cte.append(",").append(module.shouldManageVersion());
```
labkey-matthewb: SqlDialect.getBooleanLiteral()

labkey-matthewb (changes requested):

labkey-matthewb (approved):

PR #2835, opened 2021-12-08
Item 9627: TextChoiceValidator and usage for rendering select input for LKS insert/update form and app update forms

cnathe:
#### Rationale
Part 1 of supporting a new field type for "Text Choice" as a light weight lookup which allows users to restrict data entry into text fields using a list of text choices. This PR adds the new TextChoiceValidator on the server side (very much like a RegEx validator) and then detects when a field has one of these validators so that it can render a select input in the LKS insert and update forms with the valid values for that field (based on the validator expression).

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2835
* https://github.com/LabKey/labkey-ui-components/pull/684
* https://github.com/LabKey/sampleManagement/pull/769
* https://github.com/LabKey/biologics/pull/1085

#### Changes
* add PropertyValidatorType.TextChoice and TextChoiceValidator class
* update DataColumn to render select input with valid values for field that has a TextChoiceValidator
* update field JSON writer to include valid values for a field that has a TextChoiceValidator
* PropertyService updates for helpers to getValidatorForColumn() and getTextChoiceValidatorOptions()


api/gwtsrc/org/labkey/api/gwt/client/model/PropertyValidatorType.java
```diff
@@ -70,6 +70,24 @@ public boolean isHidden()
         {
             return true;
         }
+    },
+    TextChoice
+    {
+        @Override
+        public HelpPopup createHelpPopup()
+        {
+            return new HelpPopup("Text Choice Validator", "Text Choice validators allow you to specify a set of text values which should be used to constraint values for a given domain field, like a light weight lookup.");
```
labkey-susanh:
```suggestion
            return new HelpPopup("Text Choice Validator", "Text Choice validators allow you to specify a set of text values that are used to constrain values for a given domain field, like a light-weight lookup.");

```

api/gwtsrc/org/labkey/api/gwt/client/model/PropertyValidatorType.java
```diff
@@ -70,6 +70,24 @@ public boolean isHidden()
...
+        }
+        @Override
+        public boolean isConfigurable()
+        {
+            return false;
+        }
+        @Override
+        public boolean isHidden()
+        {
+            return true;
```
labkey-susanh: Is this just a temporary thing, until we finish implementation?

experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
```diff
@@ -329,6 +335,37 @@ public void deleteValidatorsAndFormats(Container c)
...
+            {
+                if (validator.getTypeURI().equals(typeURI))
+                    return validator;
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public List<String> getTextChoiceValidatorOptions(IPropertyValidator validator, boolean sorted)
```
labkey-susanh: I'm not sure where this comment belongs, probably because this is something to be handled in a following story, but somewhere we need to handle deduplication of the values, even if they are read from XML.

experiment/src/org/labkey/experiment/api/property/TextChoiceValidator.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public String getTypeURI()
+    {
+        return createValidatorURI(PropertyValidatorType.TextChoice).toString();
+    }
+
+    @Override
+    public boolean validate(IPropertyValidator validator, ColumnRenderProperties field, @NotNull Object value,
```
labkey-susanh, line 46: I guess this is necessary to handle the case of values coming through the API or file import, but it's unfortunate that it will be a somewhat redundant check.  We should include in the test plan a case of someone specifying 200 options and importing a large file with values for that field. 

BTW, we aren't actually restricting the number of values anywhere currently.  Should we be?

api/gwtsrc/org/labkey/api/gwt/client/model/PropertyValidatorType.java
cnathe: Done

api/gwtsrc/org/labkey/api/gwt/client/model/PropertyValidatorType.java
cnathe: I was just matching some properties from the other types that seemed similar. However, looking at it now it doesn't look like isHidden or isConfigurable is used anywhere. Removing.

experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
cnathe: Yeah, I think this is the right place to remove duplicates, trim values, and sort. Done.

experiment/src/org/labkey/experiment/api/property/TextChoiceValidator.java
cnathe: Good point about the perf. I'll be sure to check that. And you are right that in the "data entry through the UI" case this validate() check is redundant. 

Regarding restricting the number of values. I was not planning to do that for the XML case, but just in the UI. If a user wants to setup a TextChoiceValidator via XML then I think it is up to them to add as many values as they see fit.

cnathe: FYI, I tried this with a set of > 200 text choices and importing >160,000 rows to a sample type and the field validation of those rows was very fast (no exact times, but it was < 2 seconds to import all 160,000 rows). Note that the TextChoiceValidator.validate() does have a cache it uses so that it only has to parse the expression one time for the set.

experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
```diff
@@ -329,6 +337,40 @@ public void deleteValidatorsAndFormats(Container c)
...
+        String expression = "";
+        if (validator != null && validator.getExpressionValue() != null)
+            expression = validator.getExpressionValue();
+
+        // split expression, trim choices, and remove duplicates
+        String[] choiceTokens = expression.split("\\|");
+        Set<String> choiceSet = Arrays.stream(choiceTokens).map(String::trim).collect(Collectors.toSet());
+
+        // remove empty strings and sort
+        return choiceSet.stream().filter(choice -> !"".equals(choice))
```
labkey-susanh: Maybe use `!StringUtils.isEmpty(choice)` (just reads nicer, I think).

experiment/src/org/labkey/experiment/api/property/TextChoiceValidator.java
labkey-susanh: Thanks for trying that.  Maybe add a note about this in the spec as well, at least for posterity.

labkey-susanh: And agreed about the not limiting this within XML.

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
cnathe: yup, much better. done.

PR #2837, opened 2021-12-08
Issue 44434: Add admin action for adjusting timestamps for Created and Modified fields.

labkey-susanh:
#### Rationale
System-created timestamp values are created using the server's timestamp.  When we adjust the timezone for the server to correspond to the client's usual timezone, existing timestamps remain in the original timezone.  This action provides a way to adjust these system-created timestamp values so they line up with the adjusted server timestamp.

#### Changes
* Add `AdjustSystemTimestampAction` to `AdminController`

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -10541,6 +10541,7 @@ public class AdminController extends SpringActionController
                 });
                 t.commit();
             }
+            logger.info("DONE Adjusting all " + toUpdate + " timestamp fields in all tables by " + form.getHourDelta() + " hours.");
             return true;
         }
 
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -10527,7 +10527,7 @@ public class AdminController extends SpringActionController
                 ModuleLoader.getInstance().getModules().forEach(module -> {
                     logger.info("==> Beginning update of timestamps for module: " + module.getName());
                     module.getSchemaNames().stream().sorted().forEach(schemaName -> {
-                        DbSchema schema = DbSchema.get(schemaName, DbSchemaType.Module); // Will these always be Module schemas ??
+                        DbSchema schema = DbSchema.get(schemaName, DbSchemaType.Module);
                         schema.getTableNames().forEach(tableName -> {
                             TableInfo tInfo = schema.getTable(tableName);
                             if (tInfo.getTableType() == DatabaseTableType.TABLE)
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -10487,33 +10487,40 @@ public class AdminController extends SpringActionController
             return new JspView<>("/org/labkey/core/admin/adjustTimestamps.jsp", form, errors);
         }
 
-        private void updateField(TableInfo tInfo, String fieldName, int delta)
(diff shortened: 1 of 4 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -10465,6 +10465,99 @@ private static boolean validateAndSaveFormat(String format, Runnable clearer, Fo
...
+        {
+            SQLFragment sql = new SQLFragment();
+            DbSchema schema = tInfo.getSchema();
+
+            ColumnInfo col = tInfo.getColumn(FieldKey.fromParts(fieldName));
+            if (col != null)
+            {
+                sql.append("UPDATE ").append(tInfo, "");
+                if (schema.getSqlDialect().isPostgreSQL())
+                    sql.append(String.format("SET %s = %s + interval '%d hour';\n", fieldName, fieldName, delta));
```
labkey-matthewb: just FYI, {fn timestampadd(SQL_TSI_HOUR, delta, field)} is cross database.  That is how we implement this in LabKey SQL.
https://docs.oracle.com/javadb/10.8.3.0/ref/rrefjdbc88908.html

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -10465,6 +10465,99 @@ private static boolean validateAndSaveFormat(String format, Runnable clearer, Fo
...
+                errors.reject(ERROR_MSG, "You must specify a non-zero value for 'Hour Delta'");
+        }
+
+        @Override
+        public ModelAndView getView(AdjustTimestampsForm form, boolean reshow, BindException errors) throws Exception
+        {
+            return new JspView<>("/org/labkey/core/admin/adjustTimestamps.jsp", form, errors);
+        }
+
+        private void updateField(TableInfo tInfo, String fieldName, int delta)
```
labkey-matthewb: Collection<String> fieldNames?

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -10465,6 +10465,99 @@ private static boolean validateAndSaveFormat(String format, Runnable clearer, Fo
...
+        {
+            SQLFragment sql = new SQLFragment();
+            DbSchema schema = tInfo.getSchema();
+
+            ColumnInfo col = tInfo.getColumn(FieldKey.fromParts(fieldName));
+            if (col != null)
```
labkey-matthewb: && col.getJdbcType() == JdbcType.TIMESTAMP

labkey-matthewb (changes requested):
1) Maybe expand what we consider built-in date columns .e.g Modified, Created, lastIndexed, diModified, diCreated, _ts, etc. (I don't think we need to tackle this but things like "opened", "resolved", and "closed" get interesting, you could imagine asking tableinfo/module what are your built-in timestamp columns?).  This can be one off, but BuiltInColumnTypes is a possible place to consolidate default adjudication of column 'types'.
2) I think on UPDATE per table would be nice

core/src/org/labkey/core/admin/AdminController.java
labkey-susanh: Done.

core/src/org/labkey/core/admin/AdminController.java
labkey-susanh: Done.

core/src/org/labkey/core/admin/AdminController.java
labkey-susanh: Done.

labkey-matthewb (approved):

PR #2839, opened 2021-12-08
canRead() -> hasPermission()

labkey-matthewb:
#### Rationale
fix for datasets.jsp

#### Related Pull Requests
* 

#### Changes
*


labkey-tchad (approved):
Does this mean anything still using `canRead` won't function properly with participant access control?

PR #2842, opened 2021-12-09
Deprecate Java 16, etc.

labkey-adam:
#### Rationale
We're moving on to Java 17. And other improvements.

#### Changes
* Deprecate Java 16 and update recommendation
* Add support for `LocalDate` and `LocalDateTime` to Table & Selector
* Add "database time" to Server Information page. Warn if server and database clocks are more than 10 seconds apart
* Better heuristic for cache stack traces: always save them for permanent caches, never save them for temporary caches
* Double capacity for cache that's "limited" on labkey.org


api/src/org/labkey/api/module/JavaVersion.java
```diff
@@ -110,7 +110,7 @@ public static String getJavaVersionDescription()
 
     public static String getRecommendedJavaVersion()
     {
-        return "AdoptOpenJDK 16 64-bit (x64) with HotSpot JVM";
+        return "Eclipse Temurin 17 64-bit (x64) with HotSpot JVM";
```
labkey-jeckels: My new MacBook has the M1 chipset (arm64, not x64). Do we need to include the "x64" here or is 64-bit enough?

core/src/org/labkey/core/admin/admin.jsp
```diff
@@ -80,6 +84,16 @@
...
+
+    LocalDateTime serverTime = LocalDateTime.now();
+    LocalDateTime databaseTime = new SqlSelector(DbScope.getLabKeyScope(), "SELECT CURRENT_TIMESTAMP").getObject(LocalDateTime.class);
+    long duration = Duration.between(serverTime, databaseTime).toSeconds();
+
+    // Warn if greater than this many seconds
+    long warningSeconds = 10;
+
+    HtmlString style = unsafe(duration > warningSeconds ? " style=\"color:red;\"" : "");
+    HtmlString warning = unsafe(duration > warningSeconds ? " - Warning: Server and database times differ by " + duration + " seconds!" : "");
```
labkey-jeckels:
```suggestion
    HtmlString warning = unsafe(duration > warningSeconds ? " - Warning: Web server and database times differ by " + duration + " seconds!" : "");

```

labkey-jeckels (approved):

api/src/org/labkey/api/module/JavaVersion.java
labkey-adam: Aren't you fancy. I suppose you want me to update the Supported Technologies page as well.

core/src/org/labkey/core/admin/admin.jsp
labkey-adam: Updated the labels and the warning in keeping with the spirit of this comment

labkey-jeckels: Cool, I thought about suggesting on those too. Nice to have the consistency.

PR #2844, opened 2021-12-09
Issue 44437: Permissions panel is taking minutes to render

labkey-jeckels:
#### Rationale
When you have ~10,000 workbook children, it takes a long time to ask them all about their enabled modules and if they have a study 

#### Changes
* We don't use workbooks for studies so don't interrogate workbook children


labkey-adam (approved):
Could consider moving this method to StudyService so this and the QCAnalyst role could share the impl

labkey-jeckels:
@labkey-adam I considered ways to share the little bit of code, but since one is checking if the Assay module is enabled I didn't think it was worth the effort.

labkey-adam:
Yeah, fair enough... not a lot to share there

PR #2855, opened 2021-12-13
Issue 43909: SubjectID custom column issues

labkey-martyp:
#### Rationale
The change that caused this issue was in effect for about 8 months so this tries to correct any cases where participantid is null and put a not null constraint on the column.  Additionally a check is put in place to NOT allow any more mutations of the data in these datasets until the subject id custom column is removed from the design.  The goal is to try to preserve the data but force the user to fix or create a new dataset.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2661

#### Changes
* Upgrade script that iterates studies and datasets, trying to correct any null value participantids and add a null constraint to the participantid column of all datasets
* Check in DatasetDataIteratorBuilder to throw an error if a dataset has a custom column of the same name as the subject alias for the study, and not allow any mutations of the data in the dataset

### DIFF
--- a/study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
+++ b/study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
@@ -410,7 +410,6 @@ public class DatasetDataIteratorBuilder implements DataIteratorBuilder
             setupError("Missing required field SequenceNum");
 
         // Issue 43909: Don't allow insert/update if subject id is overwritten with custom column (no longer allowed).
-        boolean subjectIdConflict = false;
         for (DomainProperty p : Objects.requireNonNull(_datasetDefinition.getDomain()).getProperties())
         {
             if (p.getName().equalsIgnoreCase(_datasetDefinition.getStudy().getSubjectColumnName()))
--- a/study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
+++ b/study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
@@ -407,6 +407,10 @@ public class DatasetDataIteratorBuilder implements DataIteratorBuilder
         if (timetype.isVisitBased() && null == it.indexSequenceNumOutput)
             setupError("Missing required field SequenceNum");
 
+        if (!inputMap.containsKey(DatasetDefinition.getStudyBaseURI() + "ParticipantId"))
+            setupError(_datasetDefinition.getStudy().getSubjectColumnName() + " is not mapped correctly, ensure the dataset " +
+                    "design does not overwrite this column with a custom column.");
+
         it.setInput(ErrorIterator.wrap(input, context, false, setupError));
         DataIterator ret = LoggingDataIterator.wrap(it);
 
--- a/study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
+++ b/study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
@@ -34,6 +34,7 @@ import org.labkey.api.dataiterator.LoggingDataIterator;
 import org.labkey.api.dataiterator.ScrollableDataIterator;
 import org.labkey.api.dataiterator.SimpleTranslator;
 import org.labkey.api.exp.PropertyType;
+import org.labkey.api.exp.property.DomainProperty;
 import org.labkey.api.qc.DataState;
 import org.labkey.api.qc.QCStateManager;
 import org.labkey.api.query.BatchValidationException;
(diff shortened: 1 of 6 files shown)


cnathe:
I have verified the upgrade script and the error check on data insert for this issue on both PG and MSSQL in the 21.11 feature branch. I saw the following in my IntelliJ console output on my PG DB:
```
INFO  SqlScriptRunner          2021-12-13T14:31:43,204           Module Upgrade : Running [study-21.004-21.005.sql]
INFO  SqlScriptManager         2021-12-13T14:31:43,205           Module Upgrade : start running script : study-21.004-21.005.sql
INFO  SqlScriptExecutor        2021-12-13T14:31:43,222           Module Upgrade : Executing StudyUpgradeCode.ensureParticipantIds(ModuleContext moduleContext)
INFO  StudyManager             2021-12-13T14:31:43,244           Module Upgrade : Ensuring study participantIds in all study datasets
INFO  StudyManager             2021-12-13T14:31:44,224           Module Upgrade : test updated 2 participantId rows.
INFO  StudyManager             2021-12-13T14:31:48,529           Module Upgrade : Dataset1 updated 1 participantId rows.
INFO  SqlScriptManager         2021-12-13T14:31:48,866           Module Upgrade : finished running script : study-21.004-21.005.sql
```

After the upgrade was complete, I verified that I saw the expected values in my SubjectID / PariticpantID column of my dataset and that I got the expected error when trying to insert a new dataset row before I removed the duplicate field. After removing the duplicate field for SubjectId, I was then able to insert new dataset rows as expected.

study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
```diff
@@ -407,6 +409,21 @@ else if (_datasetDefinition.getUseTimeKeyField())
...
+        for (DomainProperty p : Objects.requireNonNull(_datasetDefinition.getDomain()).getProperties())
+        {
+            if (p.getName().equalsIgnoreCase(_datasetDefinition.getStudy().getSubjectColumnName()))
+            {
+                subjectIdConflict = true;
+                break;
+            }
+        }
+
+        if (subjectIdConflict)
```
labkey-jeckels: Functionally equivalent, but you could move the setupError() call inside of the if statement() and eliminate the need for the subjectIdConflict variable completely

study/src/org/labkey/study/model/StudyManager.java
```diff
@@ -4717,6 +4718,76 @@ public void ensureDesignDomains(final ModuleContext context)
...
+                                    {
+                                        SQLFragment updateSql = new SQLFragment("UPDATE studydataset.").append(dom.getStorageTableName());
+                                        updateSql.append(" SET participantid = ").append(subjectColName);
+                                        updateSql.append(" WHERE participantid IS NULL");
+                                        int rows = new SqlExecutor(StudySchema.getInstance().getScope()).execute(updateSql);
+                                        if (rows > 0)
+                                            _log.info(dataset.getName() + " updated " + rows + " participantId rows.");
+                                    }
+                                    else
+                                    {   // set participantid to NA if it is null and doesn't match the above case. Only way known that it could
```
labkey-jeckels:
```suggestion
                                    {   // set participantid to Unknown if it is null and doesn't match the above case. Only way known that it could

```

study/src/org/labkey/study/model/StudyManager.java
```diff
@@ -4717,6 +4718,76 @@ public void ensureDesignDomains(final ModuleContext context)
...
+                                    // Try to correct datasets with subject id overwritten with custom column
+                                    List<? extends DomainProperty> dupeId = dom.getProperties().stream().filter(d -> d.getName().equalsIgnoreCase(subjectColName)).collect(Collectors.toList());
+                                    if (dupeId.size() > 0)
+                                    {
+                                        SQLFragment updateSql = new SQLFragment("UPDATE studydataset.").append(dom.getStorageTableName());
+                                        updateSql.append(" SET participantid = ").append(subjectColName);
+                                        updateSql.append(" WHERE participantid IS NULL");
+                                        int rows = new SqlExecutor(StudySchema.getInstance().getScope()).execute(updateSql);
+                                        if (rows > 0)
+                                            _log.info(dataset.getName() + " updated " + rows + " participantId rows.");
```
labkey-jeckels: For both logging lines, suggest adding the container path for the study.

labkey-jeckels (approved):

study/src/org/labkey/study/model/StudyManager.java
labkey-martyp: Added

study/src/org/labkey/study/model/StudyManager.java
labkey-martyp: Updated

study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
labkey-martyp: Yeah good call, that was refactored a couple times

PR #2856, opened 2021-12-13
Ignore auto-generated jars.txt from Experiment module

labkey-jeckels:
#### Rationale
Earlier this year, we switched to auto-generate this file. Thus, we don't need it committed to Git.

#### Changes
* Ignore the file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@ core/resources/web/core/css
 */resources/web/gen
 */resources/views/gen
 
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
A customer asked about this on his dev machine this morning. I told him to ignore it in the list of changed/added files, and then noticed I had one sitting around too.

labkey-susanh (approved):

PR #2857, opened 2021-12-14
ELN: Email Notification Improvements

RosalineP:
#### Rationale
We would like to update the email subject lines on emails sent when commenting on a Notebook to be more informative. Additionally, we would like to update who receives emails in various circumstances when commenting on a Notebook: users should generally not receive emails for comments they have just posted, and notebook authors should generally receive emails when their notebook receives a comment.

#### Related Pull Requests
* https://github.com/LabKey/labbook/pull/127

#### Changes
* When we know we are in a Notebook thread, 
    * Send notebook authors an email if their Email Notification Settings are set to 'my conversations', and a user comments in their notebook
    * Do not send a user an email if their Email Notification Settings are set to 'my conversations', and they have just made the most recent comment
* Add 'message subject' member that can overwrite the default email subject


announcements/src/org/labkey/announcements/model/AnnouncementManager.java
```diff
@@ -525,6 +527,22 @@ private static void sendNotificationEmails(final AnnouncementModel a, final Wiki
             IndividualEmailPrefsSelector sel = new IndividualEmailPrefsSelector(c);
             Set<User> recipients = sel.getNotificationUsers(a);
 
+            DiscussionSrcTypeProvider typeProvider = AnnouncementService.get().getDiscussionSrcTypeProvider(a.getDiscussionSrcEntityType());
+            if (typeProvider != null)
+            {
+                for (User u : typeProvider.getNotebookAuthors(c, user, a.getDiscussionSrcIdentifier()))
+                {
+                    Integer pref = sel.createEmailPrefsMap(c).get(u).getApplicablePreference(a).getEmailOptionId();
+                    if (EmailOption.MESSAGES_MINE.getValue() == pref && u.isActive())
```
RosalineP, line 537: Maybe I should add a short explanatory comment, "// Skip if user is inactive"? This is something we do in other places in the code to avoid giving users an opportunity to spam others who are not actually legitimate users.

labkey-alan: Can't hurt.

announcements/src/org/labkey/announcements/model/AnnouncementManager.java
```diff
@@ -525,6 +527,22 @@ private static void sendNotificationEmails(final AnnouncementModel a, final Wiki
...
+                for (User u : typeProvider.getNotebookAuthors(c, user, a.getDiscussionSrcIdentifier()))
+                {
+                    Integer pref = sel.createEmailPrefsMap(c).get(u).getApplicablePreference(a).getEmailOptionId();
+                    if (EmailOption.MESSAGES_MINE.getValue() == pref && u.isActive())
+                        recipients.add(u);
+                }
+
+                User createdByUser = UserManager.getUser(a.getCreatedBy());
+                Integer createdByPref = sel.createEmailPrefsMap(c).get(createdByUser).getApplicablePreference(a).getEmailOptionId();
+                if (EmailOption.MESSAGES_MINE.getValue() == createdByPref)
```
labkey-alan, line 544: I would add a comment that explains what we're doing here. Can be brief "Only email users that..." or "Don't email users that..."

announcements/src/org/labkey/announcements/model/AnnouncementManager.java
```diff
@@ -1164,6 +1182,9 @@ protected void addCustomReplacements(Replacements replacements)
             {
                 if (notificationBean == null)
                     return null;
+                if (notificationBean.messageSubject != null)
```
labkey-alan: Instead of doing this null check here I would have it setup in the constructor to have the messageSubject default to the code below. That way we can move most of the logic out of this and just return `notificationBean.messageSubject` which is similar to what we do for the other replacements.

api/src/org/labkey/api/announcements/api/DiscussionSrcTypeProvider.java
```diff
@@ -5,9 +5,22 @@
...
     @Nullable
     ActionURL getThreadURL(Container container, User user, int announcementRowId, String discussionSrcIdentifier);
+
+    @Nullable
+    default String getEmailSubject(Container container, User user, int announcementRowId, String discussionSrcIdentifier, String body, String title, String parentBody)
+    {
+        return null;
+    };
+
+    default Set<User> getNotebookAuthors(Container container, User user, String discussionSrcIdentifier)
```
labkey-alan: This should not be called `getNotebookAuthors`, as we don't want the concept of Notebooks leaking into other modules. Instead this should be called `getRecipients`. We need to support the same email scenarios for Workflow and it wouldn't make sense to have Workflow implement `getNotebookAuthors`.

labkey-alan (changes requested):
Overall looks good, I did leave some feedback that should be addressed though.

announcements/src/org/labkey/announcements/model/AnnouncementManager.java
RosalineP: Like so?

announcements/src/org/labkey/announcements/model/AnnouncementManager.java
```diff
@@ -1276,6 +1298,20 @@ public EmailNotificationBean(Container c,
...
+            {
+                String parentBody = null;
+                AnnouncementModel annParent = AnnouncementManager.getAnnouncement(c, a.getParent());
+                if (annParent != null)
+                    parentBody = annParent.getBody();
+                this.messageSubject = typeProvider.getEmailSubject(c, UserManager.getUser(a.getCreatedBy()), a.getRowId(), a.getDiscussionSrcIdentifier(), a.getBody(), a.getTitle(), parentBody);
+            }
+            else
+            {
+                this.messageSubject = StringUtils.trimToEmpty(isResponse ? "RE: " + parent.getTitle() : a.getTitle());
```
labkey-alan, line 1311: I don't think we want the default case in an else block, because getEmailSubject can return null. We have two options, I'm fine with either:

1. Initialize the value with the default before the if block above with the default value here. Then, only override the value if the typeProvide returns a non-null value.
2. Convert this else into a new, and separate, if block that checks `if (this.messageSubject == null)`

labkey-alan (approved):
Approved, but I did point out one minor thing that should be addressed so we don't accidentally have null subjects.

PR #2861, opened 2021-12-14
Issue 44110 - Support for displaying run object inputs and object outputs

ankurjuneja:
#### Rationale
This PR provides support to register run inputs and run outputs views with experiment service and display those views in run details view.

#### Related Pull Requests
* https://github.com/LabKey/provenance/pull/82

#### Changes
* View registration support


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -7710,6 +7714,30 @@ public boolean useUXDomainDesigner()
...
+    @Override
+    public void registerRunOutputsViewProvider(Set<QueryForm> queryForms)
+    {
+        _runOutputsQueryForms.addAll(queryForms);
+    }
+
+    @Override
+    public Set<QueryForm> getRunInputsQueries()
+    {
+        return _runInputsQueryForms;
```
labkey-jeckels: Please return an immutable wrapper over these collections via Collections.unmodifiableSet()

api/src/org/labkey/api/exp/api/ExperimentService.java
```diff
@@ -792,6 +793,14 @@ ExpRun derive(Map<ExpMaterial, String> inputMaterials, Map<ExpData, String> inpu
 
     SQLFragment generateExperimentTreeSQLLsidSeeds(List<String> lsids, ExpLineageOptions options);
 
+    Set<QueryForm> getRunInputsQueries();
+
+    Set<QueryForm> getRunOutputsQueries();
+
+    void registerRunInputsViewProvider(Set<QueryForm> queryForms);
```
labkey-jeckels: A comment here about where this is used and the expectation that the referenced query has a RunId column would be helpful.

labkey-jeckels (approved):
Some recommendations but no need for additional review.

api/src/org/labkey/api/exp/api/ExperimentService.java
ankurjuneja: done

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
ankurjuneja: done

PR #2862, opened 2021-12-14
Generate jars.txt file for Core, Search, Query, BigIron

Sigmonia:
#### Rationale
Migrate from a static credits file to one generated by the build 
https://docs.google.com/document/d/1JPZ8al_M6eQh813Q8y3ALhRNgTieveaKDXv5iFUbq7A/edit

#### Changes
* Dependency build method change to build.gradle
* deletion of jars.txt static file


labkey-jeckels (changes requested):
I just added the experiment module's jars.txt to .gitignore because we're autogenerating it. I assume we should do the same here and in the Biologics PR too.

https://github.com/LabKey/platform/pull/2856

core/build.gradle
```diff
@@ -1,27 +1,101 @@
...
+            "Commons Math",
+            "Apache",
+            "http://commons.apache.org/math/",
+            ExternalDependency.APACHE_2_LICENSE_NAME,
+            ExternalDependency.APACHE_2_LICENSE_URL,
+            "Lightweight, self-contained mathematics and statistics components"
+        )
+    )
+
+    //TODO should we add a variable for the version here?
```
labkey-susanh: A property for this version is not as useful as other places since I believe this dependency would only ever be used in one place, so I would remove this TODO.

query/build.gradle
```diff
@@ -9,22 +11,99 @@ configurations {
...
+//   external "pentaho:mondrian:${mondrianVersion}",
+//           {
+//              // exclude these because they cause interference with the tomcat versions
+//              exclude group: "javax.servlet", module:"jsp-api"
+//              exclude group: "javax.servlet", module:"servlet-api"
+//               // exclude this because it brings older version of log4j
+//              exclude group: "log4j", module:"log4j"
+//           }
+//
+//    external "org.olap4j:olap4j:${olap4jVersion}"
```
labkey-susanh: Remove commented out code.

query/build.gradle
```diff
@@ -9,22 +11,99 @@ configurations {
...
+
+    BuildUtils.addExternalDependency(
+        project,
+        new ExternalDependency(
+            "eigenbase:eigenbase-resgen:${eigenbaseResgenVersion}",
+            "Mondrian",
+            "Pentaho",
+            "http://mondrian.pentaho.com/",
+            "Eclipse Public License - v 1.0",
+            "http://www.apache.org/licenses/LICENSE-2.0",
```
labkey-susanh: Looks like the wrong license URL.

search/build.gradle
```diff
@@ -6,28 +9,207 @@ ext.luceneVersion="8.11.0"
...
+            "https://github.com/codelibs/jhighlight/blob/master/LICENSE_CDDL.txt",
+            "Parsers for Java, HTML, XHTML, XML and LZX"
+        )
+    )
+
+    BuildUtils.addExternalDependency(
+        project,
+        new ExternalDependency(
+            "org.apache.commons:commons-csv:1.9.0",
+            "Commons CSV ™",
```
labkey-susanh, line 166: Does the "TM" superscript actually come through?

labkey-susanh (changes requested):

labkey-susanh (approved):

labkey-jeckels (approved):
Thanks for the .gitignore updates. Looks like Susan has reviewed the other followup changes.

PR #2863, opened 2021-12-15
Issue 44494: QueryLogging errors not treated as a true error condition

labkey-jeckels:
#### Rationale
When we don't find a column that's required for logging purposes, we're stopping on the first one and throwing an UnauthorizedException. This achieves the goal of preventing users from seeing data where we aren't in a position to do the full audit logging. However, it doesn't up us track down these problems as UnauthorizedException is treated as a lack of permission instead of a failure to include the right columns in the data. 

#### Changes
* Use IllegalStateException to capture details in mothership reports and logging
* Concat all of the missing columns into the error

### DIFF
--- a/api/src/org/labkey/api/data/LoggingResultSetWrapper.java
+++ b/api/src/org/labkey/api/data/LoggingResultSetWrapper.java
@@ -15,14 +15,18 @@
  */
 package org.labkey.api.data;
 
+import org.apache.tomcat.util.buf.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.labkey.api.audit.AuditLogService;
 import org.labkey.api.view.UnauthorizedException;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/LoggingResultSetWrapper.java
```diff
@@ -15,14 +15,18 @@
  */
 package org.labkey.api.data;
 
+import org.apache.tomcat.util.buf.StringUtils;
```
labkey-adam: That's likely the wrong `StringUtils`

labkey-adam (changes requested):
StringUtils import

api/src/org/labkey/api/data/LoggingResultSetWrapper.java
labkey-jeckels: Sure is!

labkey-adam (approved):

PR #2865, opened 2021-12-15
Item 9092: Naming pattern improvements 

XingY:
#### Rationale
This PR adds support for validating naming patterns and providing auto generated name previews for samples and dataclasses.

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/118
* https://github.com/LabKey/labkey-ui-components/pull/689
* https://github.com/LabKey/platform/pull/2865
* https://github.com/LabKey/sampleManagement/pull/778
* https://github.com/LabKey/biologics/pull/1092

#### Changes
* Add GetDomainNamePreviewsAction and ValidateNameExpressionsAction
* Update components version to support validation and preview on sample and dataclass designers


api/src/org/labkey/api/exp/property/DomainKind.java
```diff
@@ -338,4 +339,14 @@ public void validateOptions(Container container, User user, T options, String na
         if (isUpdate && !canEditDefinition(user, domain))
             throw new UnauthorizedException("You don't have permission to edit this domain");
     }
+
+    public Tuple3<List<String>, List<String>, List<String>> validateNameExpressions(T options, GWTDomain domainDesign, Container container)
```
labkey-susanh: You could use a Java Record here instead of `Tuple3` and be very modern.

Otherwise, I think some Javadocs would be helpful here to say what the three parts of the Tuple are.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -670,17 +672,28 @@ public Domain getDomain()
...
+            sb.append("\n");
+        }
 
-        sb.append("If not provided, a unique name will be generated from the expression:\n");
+        sb.append("\nIf not provided, a unique name will be generated from the expression:\n");
         sb.append(nameExpression);
+        sb.append(".");
+        if (!StringUtils.isEmpty(nameExpressionPreview))
+        {
+            sb.append("\nExample name that will be generated from the current pattern: \n");
```
```suggestion
            sb.append("\nExample of name that will be generated from the current pattern: \n");

```

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -442,6 +443,77 @@ public Object execute(DomainApiForm form, BindException errors)
...
+        }
+
+        @Override
+        protected ObjectMapper createResponseObjectMapper()
+        {
+            return this.createRequestObjectMapper();
+        }
+
+        @Override
+        public void validateForm(DomainApiForm form, Errors errors)
```
labkey-susanh: Is the empty override needed?  If so, maybe add a comment in the body about why it's empty.

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -960,11 +1032,11 @@ public void setIncludeWarnings(boolean includeWarnings)
          * Method to validate form
          */
         @JsonIgnore
-        public void validate(Container container, User user)
+        public Tuple3<List<String>, List<String>, List<String>> validate(Container container, User user, boolean validateNameExpressionOnly)
```
labkey-susanh: Update to Javadocs for return value here would be good.

internal/src/org/labkey/api/exp/property/AbstractDomainKind.java
```diff
@@ -311,4 +314,21 @@ public boolean exceedsMaxLength(Domain domain, DomainProperty prop)
         int maxSize = new SqlSelector(ExperimentService.get().getSchema(), sql).getObject(Integer.class);
         return prop.getScale() < maxSize;
     }
+
+    @Override
+    public List<String> getDomainNamePreviews(String schemaName, String queryName, Container container, User user)
```
labkey-susanh, line 320: That this returns a list and the caller has to know which is first and which is second seems a little unsavory (though I get that usually there's only one pattern).

internal/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -163,13 +163,19 @@ else if (property.getLookup() != null)
     }
 
     @Nullable
-    public static GWTDomain<GWTPropertyDescriptor> getDomainDescriptor(User user, String typeURI, Container domainContainer)
+    public static GWTDomain<GWTPropertyDescriptor> getDomainDescriptor(User user, String typeURI, Container domainContainer, boolean skipPKCol /*for sample type and dataclass*/)
```
labkey-susanh: I don't immediately see why we have to skip these for sample types and data classes.  Can you say more in javadoc perhaps?

internal/src/org/labkey/experiment/api/SampleTypeDomainKind.java
```diff
@@ -428,6 +472,23 @@ public void validateOptions(Container container, User user, SampleTypeDomainKind
...
+            ValidationException errors = getNamePatternValidationResult(options.getNameExpression(), properties, aliasMap, container);
+            if (errors.hasErrors())
+                throw new IllegalArgumentException("Invalid Name Expression:" + errors.getMessage());
+        }
+
+        if (StringUtils.isNotBlank(options.getAliquotNameExpression()))
+        {
+            ValidationException errors = getNamePatternValidationResult(options.getAliquotNameExpression(), properties, aliasMap, container);
+            if (errors.hasErrors())
+                throw new IllegalArgumentException("Invalid Aliquot Name Expression:" + errors.getMessage());
```
labkey-susanh: Perhaps wait to throw so we can include both sets of error messages at once since these are independent of each other.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -141,6 +228,217 @@ public NameGenerator(@NotNull FieldKeyStringExpression nameExpression, @Nullable
...
+        for (SubstitutionValue subValue : SubstitutionValue.values())
+        {
+            String lcSub = subValue.name().toLowerCase();
+            int lcIndex = lcExpression.indexOf(lcSub);
+            if (lcIndex != -1)
+            {
+                 /*
+                 * also check that the part is not preceded by an alphabetic letter
+                 * - "unknown" contains "now", but should by pass reserved key word check
+                 * - "DataInputs" contains "Input", but should by pass "Input" check
```
labkey-susanh:
```suggestion
                 * - "unknown" contains "now", but should bypass reserved key word check
                 * - "DataInputs" contains "Input", but should bypass "Input" check
```

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -141,6 +228,217 @@ public NameGenerator(@NotNull FieldKeyStringExpression nameExpression, @Nullable
...
+                }
+                if (lcExpression.length() > (lcIndex + lcSub.length()))
+                {
+                    char postChar = lcExpression.charAt(lcIndex + lcSub.length());
+                    if (Character.isLetter(postChar))
+                        continue;
+                }
+                switch (subValue)
+                {
+                    case genId, randomId, batchRandomId, now, AliquotedFrom -> messages.addAll(SubstitutionFormat.validateNonFunctionalSyntax(subValue.name(), nameExpression, lcIndex));
```
labkey-susanh: Probably this is done elsewhere, but should this also include checks for `dailySampleCount`, `weeklySampleCount`, `monthlySampleCount`, `yearlySampleCount`?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -176,6 +474,93 @@ public static boolean isLineageInput(Object token, @Nullable Map<String, String>
...
+                return 1005;
+            case "name":
+            case "lsid":
+            case "description":
+                return "parent" + lookupField;
+            case "created":
+            case "modified":
+                try
+                {
+                    return new SimpleDateFormat("yyyy/MM/dd").parse("20210428");
```
labkey-susanh: I think we should use a slightly different date here than for the other date fields, so it could be more clear which is which if both are included.

I also think a constant for this parsed date is called for.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -224,73 +657,173 @@ private void initialize(@Nullable Map<String, String> importAliases)
...
+                                lineageLookupFields.computeIfAbsent(fieldParts.get(0) + "/" + fieldParts.get(1), (s) -> new ArrayList<>()).add(fieldParts.get(2));
+                                lookupValuePreview = getLineageLookupTokenPreview(fkTok, fieldParts.get(0), fieldParts.get(1), fieldParts.get(2), user);
+                            }
                         }
-                        else if (fieldParts.size() == 3)
+                        else
                         {
-                            // MaterialInputs/SampleType/lookup, DataInputs/DataClass/lookup
-                            lineageLookupFields.computeIfAbsent(fieldParts.get(0) + "/" + fieldParts.get(1), (s) -> new ArrayList<>()).add(fieldParts.get(2));
+                            String errorMsg = "Only one level of lookup supported for lineage input: " + fkTok + ".";
```
labkey-susanh:
```suggestion
                            String errorMsg = "Only one level of lookup is supported for lineage input: " + fkTok + ".";
```

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -224,73 +657,173 @@ private void initialize(@Nullable Map<String, String> importAliases)
...
+                            previewCtx.put(fkTok.toString(), lookupValuePreview);
+
+                        continue;
                     }
                     else if (fieldParts.size() > 2)
                     {
                         // for now, we only support one level of lookup: ${ingredient/name}
                         // future versions could support multiple levels
-                        throw new UnsupportedOperationException("Only one level of lookup supported: " + fkTok);
+                        String errorMsg = "Only one level of lookup supported: " + fkTok + ".";
```
labkey-susanh:
```suggestion
                        String errorMsg = "Only one level of lookup is supported: " + fkTok + ".";
```

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -224,73 +657,173 @@ private void initialize(@Nullable Map<String, String> importAliases)
...
+                            _syntaxErrors.add(errorMsg);
+                        else
+                            throw new UnsupportedOperationException(errorMsg);
                     }
 
-                    if (_parentTable == null)
-                        throw new UnsupportedOperationException("Parent table required for name expressions with lookups: " + fkTok);
+                    if (_parentTable == null && domainFields.isEmpty())
+                    {
+                        String errorMsg = "Parent table required for name expressions with lookups: " + fkTok + ".";
```
labkey-susanh:
```suggestion
                        String errorMsg = "Parent table is required for name expressions with lookups: " + fkTok + ".";
```
I worry a little about these error messages referring to "Parent table" and "lookup" when that isn't wording we use in LKSM.  And we're using "name expressions" here instead of "naming patterns".  But I don't have a better suggestion given inconsistencies in the products.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -224,73 +657,173 @@ private void initialize(@Nullable Map<String, String> importAliases)
...
+                            continue;
 
-                    List<ColumnInfo> pkCols = lookupTable.getPkColumns();
-                    if (pkCols.size() != 1)
-                        continue;
+                        List<ColumnInfo> pkCols = lookupTable.getPkColumns();
+                        if (pkCols.size() != 1)
+                        {
+                            if (_validateSyntax)
+                                _syntaxErrors.add("Look up field not supported on table with multiple PK fields: " + root);
```
labkey-susanh:
```suggestion
                                _syntaxErrors.add("Look up field not supported on table with multiple primary key fields: " + root + ".");
```

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -224,73 +657,173 @@ private void initialize(@Nullable Map<String, String> importAliases)
...
+
+                        if (fkSchema != null)
+                        {
+                            TableInfo lookupTable = fkSchema.getTable(col.getLookupQuery());
+                            if (lookupTable != null)
+                            {
+                                List<String> pkCols = lookupTable.getPkColumnNames();
+                                if (pkCols.size() != 1)
+                                {
+                                    _syntaxErrors.add("Look up field not supported on table with multiple PK fields: " + root);
```
labkey-susanh:
```suggestion
                                    _syntaxErrors.add("Lookup field not supported on table with multiple primary key fields: " + root + ".");
```

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -224,73 +657,173 @@ private void initialize(@Nullable Map<String, String> importAliases)
...
+                            }
+                        }
+                    }
+                }
+
+                if (_validateSyntax)
+                {
+                    if (!lookupExist)
+                    {
+                        _syntaxErrors.add("Look up field does not exist: " + fieldKey.toString());
```
labkey-susanh:
```suggestion
                        _syntaxErrors.add("Lookup field does not exist: " + fieldKey.toString() + ".");
```

labkey-susanh (approved):
Looks good!

api/src/org/labkey/api/exp/property/DomainKind.java
XingY: Good thought. That's what I initially planned when starting the implementation but rolled back due to it being not supported back then. Refactored to use record instead.

internal/src/org/labkey/api/exp/property/DomainUtil.java
XingY: Updated inline comment. Needed by bypass getTableInfo for name preview at table init time, because that causes stack overflow.

api/src/org/labkey/api/data/NameGenerator.java
XingY: Good catch. Fixed

PR #2875, opened 2021-12-16
44449 : Cross folder support for sample and dataclass prefixes

labkey-klum:
#### Rationale
In v21.11 we introduced a feature for users to specify an "ID/Name Prefix" that would be utilized across all Data Classes and Sample Types within a folder. Subsequently, in v21.12 we introduced the experimental capability to support cross-folder data in LKB. This PR is focused on getting the ID/Name Prefix feature to support cross-folder data in a way that aligns with our users expectations. Here is a brief outline of what is desired:
1. Folder administrators should be able to define and save an "ID/Name Prefix" that is specific to each folder (i.e. in the top-level folder and in each subfolder).
2. When data is inserted/imported into a folder, then the "ID/Name Prefix" specific to that folder should be applied to the data going into that folder.
The wrinkle here is supporting data in data types (i.e. Data Classes, Sample Types) whose definitions are shared cross-folder. The desire is that the prefix is obtained from the folder where the data is being written and not (inherently) from the prefix for the folder where the data type definition is defined. Currently, the prefix is explicitly prefixed onto the name expression for each data type domain. This is problematic for shared definitions because the prefix is domain-centric rather than data-centric.

https://docs.google.com/document/d/13P8gBu4kPHKqrWaR1S0AgXTx1SSbdGXymL1QCudOU7Y/edit

#### Related Pull Requests
* labkey/samplemanagement#783
* https://github.com/LabKey/biologics/pull/1097
* https://github.com/LabKey/labkey-ui-components/pull/691


api/src/org/labkey/api/exp/api/NameExpressionOptionService.java
```diff
@@ -15,6 +15,10 @@
 {
     String NAME_EXPRESSION_REQUIRED_MSG = "A Naming Pattern is required because manually specifying a name has been disabled for this folder.";
 
+    // Must match PREFIX_SUBSTITUTION_EXPRESSION in labkey-ui-components/.../domainproperties/constants.ts
+    String FOLDER_PREFIX_TOKEN = "folderPrefix";            // name expression substitution token for a configured prefix
```
XingY, line 19: Please add folderPrefix to SubstitutionValue enum, to support the newly added syntax validation and preview

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -776,7 +779,13 @@ public DataIterator getDataIterator(DataIteratorContext context)
             // sampleset.createSampleNames() + generate lsid
             // TODO does not handle insertIgnore
             DataIterator names = new _GenerateNamesDataIterator(sampletype, DataIteratorUtil.wrapMap(dataIterator, false), context, batchSize)
-                    .setAllowUserSpecifiedNames(NameExpressionOptionService.get().allowUserSpecifiedNames(sampletype.getContainer()));
+                    .setAllowUserSpecifiedNames(NameExpressionOptionService.get().allowUserSpecifiedNames(sampletype.getContainer()))
+                    .addExtraPropsFn(() -> {
+                        if (container != null)
```
XingY, line 784: minor: duplicate code in ExpDataClassDataTableImpl.java

api/src/org/labkey/api/exp/api/NameExpressionOptionService.java
labkey-klum: yes, done

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
labkey-klum: true, I was hoping that dataClasses and sampleTypes would share more common iterator code, but it didn't seem worthwhile to try to factor this out.

XingY (approved):

labkey-susanh (approved):

PR #2876, opened 2021-12-17
Item 9758: Field editor "Text Choice" data type and UI to create, update, delete field options/values

cnathe:
#### Rationale
The previous story in this epic added a TextChoiceValidator to the server side and updated the UI for the field insert/update forms to show a select input of the values. This set of PRs are part 1 of updating the field editor UI to add the new "Text Choice" data type and the UI in the field editor expanded row to define the set of values / choices for that field. 

This PR adds a few values to the domain field JSON response to make it easier for the field editor UI to display the existing values on domain update and to know the schema/query to query for distinct values when changing an existing field from "Text" to "Text Choice".

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/692
* https://github.com/LabKey/platform/pull/2876
* https://github.com/LabKey/sampleManagement/pull/784
* https://github.com/LabKey/biologics/pull/1098
* https://github.com/LabKey/commonAssays/pull/400

#### Changes
* add TEXT_CHOICE_CONCEPT_URI and initial metric to count columns of this type
* update domain field JSON response to include base table schema/query name and the validValues for an existing text choice validator


labkey-susanh (approved):

assay/src/org/labkey/assay/AssayManager.java
```diff
@@ -393,6 +395,22 @@ public AssaySchema createSchema(User user, Container container, @Nullable Contai
...
+        {
+            AssayProvider provider = getProvider(protocol);
+            AssayProtocolSchema schema = provider.createProtocolSchema(user, container, protocol, null);
+            for (String tableName : schema.getTableNames())
+            {
+                TableInfo table = schema.getTable(tableName, null, true, true);
+                if (table != null && table.getDomain() != null && table.getDomain().getTypeId() == domainId)
+                    return table;
+            }
+        }
```
labkey-susanh, line 409: It's unfortunate that isn't a query easily done with a join using some kind of "WHERE id = domainId".  This sort of update is not all that frequent, though, and mimics similar processing we do elsewhere.

cnathe: Yeah, I couldn't find any easier way. I borrow this "iterate over container assay protocols" approach from the AssayController serializeAssayDefinition() method.

PR #2883, opened 2021-12-21
Export ViewCategories in hierarchy order, etc.

labkey-adam:
#### Rationale
Export ViewCategories in a predictable order. No functional difference, but easier to debug.

Remove unused JSP file. Fix other miscellaneous IntelliJ warnings.


api/src/org/labkey/api/reports/model/ViewCategory.java
```diff
@@ -214,13 +217,58 @@ public int hashCode()
...
+    {
+        @Override
+        public int compare(ViewCategory vc1, ViewCategory vc2)
+        {
+            int parentDisplayOrder = Integer.compare(parentDisplayOrder(vc1), parentDisplayOrder(vc2));
+
+            if (0 != parentDisplayOrder)
+                return parentDisplayOrder;
+
+            int parentLabel = parentLabel(vc1).compareTo(parentLabel(vc2));
```
labkey-jeckels: Are we sure that parentLabel() and getLabel() below are never null? Can we annotate the methods as @NotNull

labkey-jeckels (approved):

api/src/org/labkey/api/reports/model/ViewCategory.java
labkey-adam: I can't guarantee it, so switching to `ObjectUtils.compare()`

PR #2885, opened 2021-12-21
ELN: Email Notification Improvements to Daily Digest

RosalineP:
#### Rationale
In the previous story, we updated subject lines on emails sent, when commenting on a Notebook, to be more informative. We also updated the recipients of the emails to include the notebook author.
In this story, we add that Notebook authors should receive in their 'Daily Digest of My Conversations' emails triggered by someone commenting in their notebook. Additionally, the subject lines of email threads within the digest have been updated to be more informative.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2857

#### Changes
* If email is sent form ELN Notebook, update subject lines in digest emails
* Remove defunct code from previous story's changes
* Edit logic of who receives what emails


announcements/src/org/labkey/announcements/model/EmailPrefsSelector.java
```diff
@@ -118,18 +121,38 @@ public boolean shouldSend(@Nullable AnnouncementModel ann, User user)
...
+                    return false;
+            }
+        }
+        else if (EmailOption.MESSAGES_MINE_DAILY_DIGEST.getValue() == emailPreference)
+        {
+            Set<User> authors = ann.getAuthors();
+            if (!authors.contains(user))
+            {
+                List<Integer> memberList = ann.getMemberListIds();
+                if (!memberList.contains(user.getUserId()) && !extraRecipients.contains(user))
```
RosalineP: As far as I could tell, on develop when users have their messages setting to 'Daily digest of my conversations', they get messages for emails even in threads they haven't participated in or are on a notify list for. This rectifies that, I think, unless I'm missing something. @labkey-adam Any thoughts on this?

Edit: Alan made the good point that users may be depending on the current behavior. If it's desired to not break current functionality, we could add a 'typeProvider != null' check on line 149.

announcements/src/org/labkey/announcements/model/AnnouncementDigestProvider.java
```diff
@@ -225,7 +228,20 @@ private void initPosts()
...
-                    sb.append(PageFlowUtil.filter(ann.getTitle())).append("</td></tr>");
+                    DiscussionSrcTypeProvider typeProvider = AnnouncementService.get().getDiscussionSrcTypeProvider(ann.getDiscussionSrcEntityType());
+                    if (typeProvider != null)
+                    {
+                        String parentBody = null;
+                        AnnouncementModel annParent = AnnouncementManager.getAnnouncement(ann.lookupContainer(), ann.getParent());
+                        if (annParent != null)
+                            parentBody = annParent.getBody();
+                        String emailSubject = typeProvider.getEmailSubject(ann.lookupContainer(), UserManager.getUser(ann.getCreatedBy()), ann.getRowId(), ann.getDiscussionSrcIdentifier(), ann.getBody(), ann.getTitle(), parentBody);
+                        sb.append(PageFlowUtil.filter(emailSubject)).append("</td></tr>");
```
RosalineP, line 239: One little suboptimal bit is that with this title change, you can have a structure in a digest like:

> (Subject: ) rosalinep replied in notebookName on thread: "This is the original post of the thread"
> 
> rosalinep responded at {time}: "Someone reply"
> 
> person1 responded at {time}: "Reply1"
> 
> person2 responded at {time}: "Reply2"

where the subject line only points out who started a new reply chain on the thread. Not sure if this is worth editing to something like, "New comment(s) on thread: "{original thread body up to five words}""

labkey-alan: I feel like this is fine, but this question may be better answered by someone else (e.g. Bernie or Adam).

labkey-alan (approved):
Looks good.

announcements/src/org/labkey/announcements/model/EmailPrefsSelector.java
labkey-adam: I haven't tested, but the code says you're correct. I'm not too surprised... we converted the prefs from int constants to enum constants. I suspect this used to be a bit-wise OR check that got botched in the migration. I'll do an actual code review tomorrow. Better to make this work correctly (and respect users' wishes) even if it means breaking some compatibility. The fact that no one has ever complained means probably not a lot of folks using this setting.

labkey-adam: Shouldn't these two arms be combined? As in:
```
    if (EmailOption.MESSAGES_MINE.getValue() == emailPreference || EmailOption.MESSAGES_MINE_DAILY_DIGEST.getValue() == emailPreference)
    {
        ...
```
In other words, why are we skipping the null and typeProvider checks in the case of daily digest - mine?

labkey-adam (changes requested):
See comment about duplicate code

announcements/src/org/labkey/announcements/model/EmailPrefsSelector.java
RosalineP: Good point—the nullcheck should be in there and there's some duplicate code to consolidate! But in the `MESSAGES_MINE_DAILY_DIGEST` case I think it makes more sense to include messages even when they're from the user receiving the digest. Otherwise, if user1 gets into a conversation with user2 in a thread, and user1 gets a daily digest with only user2's replies, it reads awkwardly.

I've updated the code accordingly

announcements/src/org/labkey/announcements/model/AnnouncementDigestProvider.java
labkey-adam: I'm okay with this

labkey-adam (approved):

PR #2886, opened 2021-12-22
Issue 44093: Admin option to update paths in the DB, separate from moving files

labkey-jeckels:
#### Rationale
It's a pain to fix up paths stored in the DB when they've been moved by something external to LabKey Server.

#### Changes
* New link on the Admin Console->Files page that lets an admin update paths across the various tables that store them
* Audit log under "Site Settings" heading that gives the old/new paths and the number of DB rows affected


labkey-adam (changes requested):
Need to update override in `OConnorExperiments`

labkey-adam (approved):

PR #2893, opened 2021-12-23
Fb issue 44363

labkey-klum:
#### Rationale
The problem is that when we reload the study, we're trying to adjust the key on the dataset before we've updated the fields. So if the key references a column that wasn't already present in the dataset, it blows up with this error.

Specifically, we're calling in StudyManager.importDatasetSchemas() we're calling manager.updateDatasetDefinition() around line 3607 before we call deleteAndSaveProperties() around line 3624.

This will be a little tricky to handle the various before/after states that are possible. We can't just move the manager.updateDatasetDefinition() call to always happen afterwards because the existing dataset may be currently using a key column that's being remove completely.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44363

#### Changes
- leverage the existing function `buildPropertySaveAndDeleteLists` which tracks the incoming (and existing) domains as well as properties which will be deleted, but modified it a bit to consolidate the data structure for each dataset
- call `buildPropertySaveAndDeleteLists` twice, once for existing datasets in order to check if the key has changed and whether the domain needs to be saved before the dataset is updated. It is called a second time for new and existing datasets where the key hasn't changed.


study/src/org/labkey/study/model/StudyManager.java
```diff
@@ -3607,7 +3624,36 @@ public boolean importDatasetSchemas(StudyImpl study, final User user, SchemaRead
             if (d.isNew)
                 manager.createDatasetDefinition(user, def);
             else if (d.isModified)
+            {
+                // issue 44363 : in certain situations the dataset domain will need to be saved earlier in order to support
+                // a change in the key column that may not be in the initial domain
+                if (manager.isKeyChanged(def) && !def.getUseTimeKeyField())
```
labkey-matthewb: It makes me nervous that we are changing the key definition here without ensuring that there are no rows in the table.  I don't think the alternative of updating the LSIDs immediately works since the key column might be a new (not populated) column.

labkey-klum: agreed, see thread on the issue but we decided to throw an error if the dataset keys are changing on a non-empty table.

labkey-matthewb (approved):

PR #2898, opened 2021-12-28
Add link in pipeline details page to show raw log file

bbimber:
This is related to issue: https://www.labkey.org/ONPRC/Support%20Tickets/issues-details.view?issueId=44525

The rationale is that pipeline logs can be long, and there are situations when viewing a plain text file can be useful. This adds a simple link next to 'show full log file', which loads the raw log via webdav in a new tab.

### DIFF
--- a/pipeline/src/org/labkey/pipeline/status/StatusDetailsBean.java
+++ b/pipeline/src/org/labkey/pipeline/status/StatusDetailsBean.java
@@ -2,15 +2,23 @@ package org.labkey.pipeline.status;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import edu.emory.mathcs.backport.java.util.Collections;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.log4j.Logger;
+import org.jetbrains.annotations.Nullable;
 import org.labkey.api.data.Container;
+import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.StringBuilderWriter;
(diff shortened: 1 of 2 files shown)


bbimber:
@keith-labkey  @labkey-jeckels I apparently cant assign reviewers on the LabKey org

bbimber:
A couple general comments:

1) I did not find a good place to re-use the code that translates file into webdav URL. There is a lot of experiment data-related code, but we are not starting with an ExpData and it didnt seem very clean to split apart that code; however, i could take another look if you feel like this duplicates too much.

2) I left two TODOs. One is a place where I think we could inject some kind of MAX_LINES logic when streaming the log. This is not directly related to showing the raw log, but rather would be a way to prevent loading large logs from crashing the server.

3) I think I'm testing for whether this log file exists in a reasonable Path/cloud-friendly way, using Files.exists(Path); however, it is possible there are cloud filesystem logs that this would not work properly with.

labkey-adam:
Thanks for the PR, @bbimber ! I've assigned review to Josh who's likely been thinking about this most... though hopefully not too much, since he's on vacation this week. I expect he'll get to it next week after his return.

Note that your branch is labeled "21.11_fb..." but you're merging to `develop`. Probably intentional, but I wanted to make sure.

labkey-jeckels:
I made updates to the original change so I'm closing this PR in favor of one backed by the LabKey repo FB.

PR #2903, opened 2021-12-29
Item 9782: Text Choice data type support for field editor updates to in use values

cnathe:
#### Rationale
For this round of changes to the Text Choice data type epic, we are adding support for allowing updates to in use text choice values from within the field editor UI along with locking those values that are in use by a "locked" item (i.e. sample with a locked status). This PR includes the server side updates to update the rows for any in use text choice values which are updated in the field editor UI. The DomainUtil class has helpers to use the query update server for the given domain table to call the updateRows() method, which will give us the detailed audit logging we want for this action.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/697
* https://github.com/LabKey/platform/pull/2903
* https://github.com/LabKey/biologics/pull/1109
* https://github.com/LabKey/sampleManagement/pull/789

#### Changes
* update core module @labkey/components version
* during the DomainUtil.updateDomainDescriptor() method, detect text choice validator value updates and call updateRows() to remap values in those rows

### DIFF
--- a/internal/src/org/labkey/api/exp/property/DomainUtil.java
+++ b/internal/src/org/labkey/api/exp/property/DomainUtil.java
@@ -1008,7 +1008,7 @@ public class DomainUtil
             if (domainTable != null && domainTable.getUpdateService() != null)
             {
                 // we need to make all of the row updates for this domain property at one time to prevent the
-                // double mapping if one choice value want from a -> b and another from b -> c
+                // double mapping if one choice value was changed from a -> b and another from b -> c
                 // (not sure why someone would do that though)
                 List<Map<String, Object>> rows = new ArrayList<>();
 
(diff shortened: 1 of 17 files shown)


internal/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -972,6 +993,48 @@ else if (prop == null)
...
+
+    private static void updateTextChoiceValueRows(Domain domain, User user, String propName, Map<String, Object> valueUpdates, ValidationException errors)
+    {
+        if (domain != null && domain.getDomainKind() != null)
+        {
+            TableInfo domainTable = domain.getDomainKind().getTableInfo(user, domain.getContainer(), domain);
+            if (domainTable != null && domainTable.getUpdateService() != null)
+            {
+                // we need to make all of the row updates for this domain property at one time to prevent the
+                // double mapping if one choice value want from a -> b and another from b -> c
```
labkey-susanh:
```suggestion
                // double mapping if one choice value went from a -> b and another from b -> c
```

labkey-susanh (approved):

internal/src/org/labkey/api/exp/property/DomainUtil.java
cnathe: done

PR #2904, opened 2021-12-29
44194: Biologics login modal is double encoding HTML error messages

labkey-adam:
#### Rationale
Passing server-generated HTML error messages is not a great idea, as seen in #44194

#### Changes
* Eliminate `FormattedError` usages and class
* Introduce `LabKeyErrorWithLink` class to convey additional advice text and link
* Pass `adviceText` and `adviceHref` in the JSON error object
* login.js: encode the message and add a link if adviceText and adviceHref are present


labkey-adam:
We still need to teach the app login pages about adviceText and adviceHref (biologics, sample manager, CDS?).

I opted to provide optional advice text and href, and force the client to create a proper `<a>` tag. This means the whole advice text becomes a link, whereas we previously linkified just a portion of that text. A more flexible solution would be to generate HTML server side and pass it to the client as `adviceHtml`. This is not strongly typed like `HtmlString` but the name makes it fairly obvious. We have exactly three error messages that use this mechanism, so the less flexible approach seems good enough.

labkey-klum (approved):

PR #2912, opened 2021-12-30
Item 9731: Naming pattern improvement - updates for genId

XingY:
#### Rationale
Add support to starting starting value, reset and update genIds

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/119
* https://github.com/LabKey/labkey-ui-components/pull/698
* https://github.com/LabKey/platform/pull/2912
* https://github.com/LabKey/sampleManagement/pull/791
* https://github.com/LabKey/biologics/pull/1111

#### Changes
* Added MinValueSubstitutionFormat to support ${numberField:minValue(xx)}
* Added Experiment-GetGenIdAction
* Added Experiment-SetGenIdAction


experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -464,6 +464,16 @@ public NameGenerator getAliquotNameGenerator()
...
     }
 
+    /*
+     * Check name expression and aliquot name expression for '${genId:minValue(100)}' syntax to return the specified startInd for genId
+     */
+    public long getMinGenId()
+    {
+        long nameGenMin = NameGenerator.getGenIdStartValue(_object.getNameExpression());
+        long aliquotGenMin = NameGenerator.getGenIdStartValue(_object.getAliquotNameExpression());
+        return Math.max(nameGenMin, aliquotGenMin);
```
labkey-susanh, line 474: If not already mentioned in the docs, we should probably call out that there is only one `genId` for both the sample names and aliquot names, so using `minValue` in both places will take the maximum of the minimums.

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -538,9 +548,41 @@ public String createSampleName(@NotNull Map<String, Object> rowMap,
...
+    }
+
+    @Override
+    public void ensureMinGenId(long newSeqValue, Container container) throws ExperimentException
+    {
+        DbSequence seq = DbSequenceManager.get(getContainer(), SEQUENCE_PREFIX, getRowId());
+        long current = seq.current();
+        if (newSeqValue < current)
+        {
+            if (getSamples(container).isEmpty())
```
labkey-susanh: Does this actually retrieve the samples only to check if there are any?  Seems like we might be able to do that more efficiently.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -7100,4 +7100,140 @@ public void setSessionKey(String sessionKey)
...
+    @RequiresPermission(ReadPermission.class) // actual permission checked later
+    public static class SetGenIdAction extends MutatingApiAction<GenIdForm>
+    {
+        @Override
+        public void validateForm(GenIdForm form, Errors errors)
+        {
+            String kindName = form.getKindName();
+
+            if (form.getRowId() == null || StringUtils.isEmpty(kindName))
+                errors.reject(ERROR_REQUIRED, "RowId or KindName must be provided");
```
labkey-susanh, line 7152:
```suggestion
                errors.reject(ERROR_REQUIRED, "RowId and KindName must be provided");

```

labkey-susanh (approved):

PR #2915, opened 2022-01-04
Initial Issues Service

labkey-klum:
#### Rationale
For HJF we eventually want to implement a sample request API on the server. The first step towards this will be to introduce a service interface to allow issues to be created via Java code.

We currently have 2 endpoints on the server to insert/update issues:
- `AbstractIssueAction` - this is the legacy form based endpoint that is behind the existing issue UI.
- `AbstractIssueApiAction` - this was created to support the now defunct assay request workflow in Biologics. We have at least one client that uses this API directly.

The code for both of these endpoints share some common code but most of it was written at different times and has diverged between the two. We will need to unify the existing code and push the majority of it into the new service implementation.

#### Related Pull Requests
- https://github.com/LabKey/hjfSampleRequests/pull/15

#### Changes
- Implement an `IssueService` interface and implementation
- Expose a new `Issue` interface which is the immutable form of the issue model
- Unify the existing code which validates and updates issues in the 2 server endpoints : `AbstractIssueAction` and `AbstractIssueApiAction`
- Update the above endpoints to use the new service code


issues/src/org/labkey/issue/IssueServiceImpl.java
```diff
@@ -0,0 +1,312 @@
...
+
+                IssueObject prevIssue = new IssueObject();
+                if (action != Issue.action.insert && issueObject.getIssueId() != 0)
+                {
+                    prevIssue = IssueManager.getIssue(container, user, issueObject.getIssueId());
+                }
+                IssueObject duplicateOf = null;
+                String resolution = issueObject.getResolution() != null ? issueObject.getResolution() : "Fixed";
+
+                switch (action)
```
labkey-adam, line 64: `switch (enum)` always bugs me... much cleaner and safer for each enum constant to implement its own handler. But in this case, I see the enum is part of the API, so I guess we're stuck with a switch.

issues/src/org/labkey/issue/IssueServiceImpl.java
```diff
@@ -0,0 +1,312 @@
...
+
+                // get previous related issue ids before updating
+                Set<Integer> prevRelatedIds = new HashSet<>();
+                if (prevIssue != null)
+                    prevRelatedIds = prevIssue.getRelatedIssues();
+
+                // update the duplicate issue
+                if (duplicateOf != null)
+                {
+                    duplicateOf.addComment(user, HtmlString.unsafe("<em>Issue " + issueObject.getIssueId() + " marked as duplicate of this issue.</em>"));
```
labkey-adam, line 109: Is it time to stop persisting all the comments as HTML? That was a terrible decision...

issues/src/org/labkey/issue/IssueServiceImpl.java
```diff
@@ -0,0 +1,312 @@
...
+                    IssueManager.saveIssue(user, container, duplicateOf);
+                }
+
+                Set<Integer> newRelatedIds = issueObject.getRelatedIssues();
+
+                // this list represents all the ids which will need related handling for a creating a relatedIssue entry
```
labkey-adam:
```suggestion
                // this list represents all the ids which will need related handling for creating a relatedIssue entry
```

issues/src/org/labkey/issue/IssueServiceImpl.java
```diff
@@ -0,0 +1,312 @@
...
+                }
+                if (!duplicateOf.lookupContainer().hasPermission(user, ReadPermission.class))
+                {
+                    errors.reject(ERROR_MSG, "User does not have Read permission for duplicate issue '" + issueObject.getDuplicate().intValue() + "'");
+                    return;
+                }
+            }
+        }
+
+        if (action == Issue.action.insert)
```
labkey-adam: Required permissions could certainly be pushed to the enum.

labkey-adam (approved):

issues/src/org/labkey/issue/IssueServiceImpl.java
labkey-klum: agreed

labkey-nicka (approved):

PR #2916, opened 2022-01-04
21.11 fb flow publish npe

labkey-matthewb:
#### Rationale
Fix code that assumes "ParticipantId" column exists (true for dataset.getSchemaTableInfo(), but no longer true for dataset.getTableInfo())

see issue 44594

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels (approved):
I updated the target branch.

PR #2917, opened 2022-01-04
Issue 44568: Respect allowImportLookupByAlternateKey when finding data and materials during import 

labkey-susanh:
#### Rationale
When importing samples and data objects, we have not been using the `allowImportLookupByAlternateKey` to guide our logic for finding, assuming instead that if the value given parsed as an integer then it is a rowId.  We now (as of not too long ago) allow sample and data object names to be just strings of digits, which then look very much like rowIds.  This PR updates the logic within the `findExpMaterial` and `findExpData` methods so we will first interpret the value give as a name and then, if not found using the name, attempt to parse it as an integer rowId when the `allowImportLookupByAlternateKey` parameter is true.  

The change here is a little odd because we had already been trying both the name and the rowId regardless of this property, but in the opposite order, so this flag is really being used as a way to say "prefer name over rowId".  Perhaps it is better to change it so the case where `allowImportLookupByAlternateKey` is false no longer attempts to find objects by searching by name.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/702
* https://github.com/LabKey/sampleManagement/pull/795
* https://github.com/LabKey/biologics/pull/1117

#### Changes
* Pass through the `allowImportLookupByAlternateKey` config property to methods for finding samples
* Update other usages of the find methods to also prefer finding by name first/allow searching by both name and id

### DIFF
--- a/api/src/org/labkey/api/data/NameGenerator.java
+++ b/api/src/org/labkey/api/data/NameGenerator.java
@@ -1286,20 +1286,25 @@ public class NameGenerator
                 Object value = rowMap.get(ALIQUOTED_FROM_FIELD_NAME);
                 if (value != null)
                 {
-                    User user = User.getSearchUser();
-                    ExpSampleType parentObjectType = _sampleTypes.computeIfAbsent((String) ctx.get(CURRENT_SAMPLE_TYPE_NAME), (name) -> SampleTypeService.get().getSampleType(_container, user, name));
-                    ExpMaterial material = null;
-                    String strValue = value.toString();
-                    try
+                    String strValue = value.toString().trim();
+                    // if the value isn't an integer, it can't be a rowId, so we don't need to disambiguate
+                    if (StringUtils.isNumeric(strValue))
                     {
-                        material = ExperimentService.get().findExpMaterial(_container, user, parentObjectType, (String) ctx.get(CURRENT_SAMPLE_TYPE_NAME), strValue, renameCache, materialCache, true);
+                        User user = User.getSearchUser();
+                        ExpSampleType parentObjectType = _sampleTypes.computeIfAbsent((String) ctx.get(CURRENT_SAMPLE_TYPE_NAME), (name) -> SampleTypeService.get().getSampleType(_container, user, name));
+                        ExpMaterial material = null;
+
+                        try
+                        {
+                            material = ExperimentService.get().findExpMaterial(_container, user, parentObjectType, (String) ctx.get(CURRENT_SAMPLE_TYPE_NAME), strValue, renameCache, materialCache, true);
+                        }
+                        catch (ValidationException e)
+                        {
+                            // if we haven't been able to find a sample using either the name or rowId, we'll use the value originally provided
+                        }
+                        if (material != null && !material.getName().equals(strValue)) // replace supplied value with the name of the material we found.
+                            ctx.put(ALIQUOTED_FROM_FIELD_NAME, material.getName());
                     }
-                    catch (ValidationException e)
-                    {
-                        // if we haven't been able to find a sample using either the name or rowId, we'll use the value originally provided
-                    }
-                    if (material != null && !material.getName().equals(strValue)) // replace supplied value with the name of the material we found.
-                        ctx.put(ALIQUOTED_FROM_FIELD_NAME, material.getName());
                 }
             }
 
--- a/api/src/org/labkey/api/data/NameGenerator.java
+++ b/api/src/org/labkey/api/data/NameGenerator.java
@@ -77,6 +77,8 @@ import java.util.stream.Stream;
 
 import static org.labkey.api.exp.api.ExpRunItem.INPUT_PARENT;
 import static org.labkey.api.exp.api.ExpRunItem.PARENT_IMPORT_ALIAS_MAP_PROP;
+import static org.labkey.api.exp.api.ExpSampleType.ALIQUOTED_FROM_FIELD_NAME;
+import static org.labkey.api.exp.api.ExpSampleType.CURRENT_SAMPLE_TYPE_NAME;
 import static org.labkey.api.util.SubstitutionFormat.dailySampleCount;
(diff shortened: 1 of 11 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -1452,27 +1452,66 @@ public ExpDataImpl getExpData(ExpDataClass dataClass, int rowId)
...
+            {
+                data = findExpDataByRowId(dataClass, dataName, dataCache);
+                if (data == null)
+                    throw e; // throw this since it is most likely the intent was to supply the name rather than the rowId
+                else
+                    return data;
+            }
+        }
+
+        data = findExpDataByRowId(dataClass, dataName, dataCache);
```
labkey-susanh, line 1484: Should this just be `return findExpDataByRowId(...)`?  We would change the default behavior, but this does at least make use of the `AlternateKey` flag consistent.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -1488,25 +1527,59 @@ public ExpData findExpData(Container c, User user,
...
+                material = findExpMaterialByRowId(c, user, sampleType, sampleName, materialCache);
+                if (material == null)
+                    throw ve; // throw this since it is most likely the intent was to supply the name rather than the rowId
+                else
+                    return material;
+            }
+        }
+        else
+        {
+            material = findExpMaterialByRowId(c, user, sampleType, sampleName, materialCache);
```
labkey-susanh, line 1558: Same question here.

experiment/src/org/labkey/experiment/samples/UploadSamplesHelper.java
```diff
@@ -361,6 +361,7 @@ public ExpMaterial getAliquotInput()
      *                will be incorporated into the resolved inputs and outputs
      * @param entityNamePairs set of (parent column name, parent value) pairs.  Parent values that are empty
      *                    indicate the parent should be removed.
+     * @param allowImportLookupByAlternateKey if true, we attempt to find the inputs and outpus by interpreting the values as names first, then ids
```
XingY, line 365: We are reusing allowImportLookupByAlternateKey for a different purpose than its original use. Here this param indicate look for sample/data by names first, then rowId. 

But I believe this flag originally meant to look up first by column's FK (which would be lookup table's PK), then uniq key, then other means (see RemapPostConvert). So if applied to samples/data as lookup, this flag when true, would lookup first by rowId, then by name. For this reason, I think a new context parameter (resolveDataByNameFirst) works better. 

Apart from the inconsistency with its original lookup order, this flag is also intended for lookup columns only. It somewhat makes sense to apply to lineage inputs or aliquotedFrom fields, but it's a new usage if we are applying it directly to the imported sample's self name.

labkey-susanh: I could change it so we don't try to look up by name if this flag is false, which would then be mostly consistent with other treatment.  (Probably I'm not considering the right use cases, but I'm having a hard time convincing myself that folks would use the "lookupByAlternateKey" and then provide rowIds, even for their samples, so searching by name first in this case makes sense to me.)  If I take out the rowId-then-name logic, that will change current behavior (essentially, requiring people to check the "use alternate key" checkbox in LKS UI when providing names instead of rowIds), but is not unreasonable in my opinion.  

I also don't know when we actually want to resolve by id first, except for the insert case, when this flag is always false. If a new parameter were added, it seems we would have to create UI for it to be set or else it will always be the same as the lookupByAlternateKey value.  What am I missing?

Is it explicitly only for lookup columns?  These lineage columns are, in effect, lookups, just not called that explicitly.

In short, I'm not coming up with a clean way to deal with this that doesn't put some extra burden on the user and/or require additional UI.

XingY: Perhaps RemapPostConvert.mappedValue should also be updated to resolve first by unique keys, and then by PK? Otherwise the inconsistency is introduced, that sample and dataclass names are resolved by UK first, but others are resolved by PK first.

It seems @labkey-klum is the last person to work around lookupByAlternateKey and added that resolve by PK first logic for https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=40909 .

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1274,6 +1279,30 @@ private void addLineageLookupContext(String parentTypeName, String parentName, b
                 }
             }
 
+            if (rowMap.containsKey(ALIQUOTED_FROM_FIELD_NAME))
+            {
+                // Issue 44568: We will pass through a rowId when creating aliquots via insertRows.
+                // We need to find the corresponding material so we can use its name in the name we generate instead of its rowId.
+                Object value = rowMap.get(ALIQUOTED_FROM_FIELD_NAME);
+                if (value != null)
```
XingY, line 1287: Can skip a query if aliquotedFrom cannot parseInt

XingY (commented):
is this still targeting release22.1-SNAPSHOT?

api/src/org/labkey/api/data/NameGenerator.java
labkey-susanh: Done. thanks.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Opted to not do this in order to support the case of using insertRows that creates parents referencing rows that are being inserted (and thus would only be able to be referenced by name).

XingY (approved):

labkey-susanh:
Different approach implemented in https://github.com/LabKey/platform/pull/2927

PR #2926, opened 2022-01-07
Add Query metadata validation check for TableInfo.getPublicSchemaName()

cnathe:
#### Rationale
In the related PR, I fixed an issue where 2 of the Luminex schema table's where overriding the public schema name based on legacy behavior. This PR adds a check in the query metadata validation code path to compare the TableInfo.getPublicSchemaName() against the schemaPath being validated.

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/413

#### Changes
* QueryManager update to validateQueryMetadata() for getPublicSchemaName() check

### DIFF
--- a/query/src/org/labkey/query/persist/QueryManager.java
+++ b/query/src/org/labkey/query/persist/QueryManager.java
@@ -703,6 +703,9 @@ public class QueryManager
         if (null == table)
             throw new IllegalArgumentException("The query '" + queryName + "' was not found in the schema '" + schemaPath.getName() + "'!");
 
+        if (table.isPublic() && table.getPublicSchemaName() != null && !schemaPath.toString().equalsIgnoreCase(table.getPublicSchemaName()))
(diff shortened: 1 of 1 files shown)


labkey-jeckels (commented):
Thanks for following up on this, Cory.

After a little more thought, I wonder if we should get rid of the ability to set or override the public schema name and always use the UserSchema's name. Poking around the code, I saw that GuideSetCurveFitTable overrides getPublicSchemaName(), which is incorrectly returning "assay" like the other Luminex tables you fixed in the recent PR.

@labkey-matthewb can you think of legitimate cases for there to be a mismatch?

labkey-jeckels (approved):
I'll go ahead and approve this as it's less work than the larger refactor and could provide useful feedback on whether there are legit mismatches

labkey-matthewb:
For history, I think this exists as a hint that this table being used as a lookup exists as a publicly available table under the given public schema/table name.  That is not always the case.  For any real table it should just be getUserSchema().getName()

PR #2927, opened 2022-01-10
Issue 44568: Reverse the order of precedence for looking for samples and data objects by rowId vs. name

labkey-susanh:
#### Rationale
In the fix for Issue 40302, we introduced the logic that will look up samples or data class objects by name if the lookup by RowId fails. This logic works just fine when there are no overlaps between rowIds and names, but if name contain only digits and thus parse like rowIds, we may find the wrong sample or data object when trying to hook up lineage.  By reversing the order of the lookup logic to first prefer the lookup by name, we will find the intended entities more often since most of the time users and applications are providing us with names instead of rowIds and we will correctly not remove leading 0s from names by parsing them as integers.  

In most cases, users will still be able to perform insert and update operations using either rowIds or entity names.  The one exception is when the names of the objects can overlap with the rowIds of objects of that type in the same container, in which case the use of rowIds will not work.

#### Changes
* reverse order of precedence when finding `ExpMaterial` and `ExpData` objects.

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -1457,78 +1457,79 @@ public class ExperimentServiceImpl implements ExperimentService
                             RemapCache cache, Map<Integer, ExpData> dataCache)
             throws ValidationException
     {
+        StringBuilder errors = new StringBuilder();
+        // Issue 44568, Issue 40302: Unable to use samples or data class with integer like names as material or data input
+        // Attempt to resolve by name first.
+        try
+        {
+            Integer rowId = cache.remap(ExpSchema.SCHEMA_EXP_DATA, dataClassName, user, c, ContainerFilter.Type.CurrentPlusProjectAndShared, dataName);
+            if (rowId != null)
+                return dataCache.computeIfAbsent(rowId, (x) -> getExpData(dataClass, rowId));
+        }
+        catch (ConversionException e2)
+        {
+            errors.append("Failed to resolve '").append(dataName).append("' into a data. ").append(e2.getMessage());
+            errors.append(" Use 'DataInputs/<DataClassName>' column header to resolve parents from a specific DataClass.");
+            errors.append(" ").append(e2.getMessage());
+        }
+
         try
         {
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -1457,78 +1457,79 @@ public ExpData findExpData(Container c, User user,
...
-        catch (ConversionException e1)
+        catch (ConversionException e2)
         {
-            // ignore
+            errors.append("Failed to resolve '").append(sampleName).append("' into a sample.");
+            if (sampleTypeName == null)
+            {
+                errors.append(" Use 'MaterialInputs/<SampleTypeName>' column header to resolve parent samples from a specific SampleType.");
+            }
+            errors.append(" ").append(e2.getMessage());
```
labkey-jeckels: Is this message always non-null?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -1457,78 +1457,79 @@ public ExpData findExpData(Container c, User user,
...
+            if (rowId != null)
+                return dataCache.computeIfAbsent(rowId, (x) -> getExpData(dataClass, rowId));
         }
-        catch (ConversionException e1)
+        catch (ConversionException e2)
         {
-            // ignore
+            errors.append("Failed to resolve '").append(dataName).append("' into a data. ").append(e2.getMessage());
+            errors.append(" Use 'DataInputs/<DataClassName>' column header to resolve parents from a specific DataClass.");
+            errors.append(" ").append(e2.getMessage());
```
labkey-jeckels: Is the message always non-null?

labkey-jeckels (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Probably, but I've added a check just the same.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Probably, but I've added a check just because I can.

PR #2930, opened 2022-01-11
Issue 44511: Ability to change a field's data type to known safe alternatives

labkey-jeckels:
#### Rationale
Users sometimes change their mind on the appropriate data type for a field. There are a number of transitions that we know will be safe, including changing an integer to a double or non-strings into strings.

#### Changes
* New StorageProvisioner support for changing a column's type
* Gate the allowed transitions via before/after type checks in DomainPropertyImpl
* Handle non-provisioned tables via updates to exp.ObjectProperty as well
* Preserve formatting for boolean values but not other types at this point
* Misc small code cleanup


api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
```diff
@@ -1099,10 +1076,23 @@ private String getResizeColumnStatement(TableChange change)
         sb.append( String.format("ALTER TABLE %s.%s ", change.getSchemaName(), change.getTableName()));
```
labkey-matthewb: makeTableIdentifier()?

api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
```diff
@@ -1099,10 +1076,23 @@ private String getResizeColumnStatement(TableChange change)
...
+            if (column.getJdbcType().isText())
+            {
+                //Using the common default max size to make type change to text
+                dbType = column.getSize() == -1 || column.getSize() > SqlDialect.MAX_VARCHAR_SIZE ?
+                        getSqlTypeName(JdbcType.LONGVARCHAR) :
+                        getSqlTypeName(column.getJdbcType()) + "(" + column.getSize().toString() + ")";
+            }
+            else if (column.getJdbcType().isDecimal())
+            {
+                dbType = getSqlTypeName(column.getJdbcType()) + "(" + column.getPrecision() + ", " + column.getSize() + ")";
```
labkey-matthewb: getScale()?

bigiron/src/org/labkey/bigiron/mssql/BaseMicrosoftSqlServerDialect.java
```diff
@@ -1135,22 +1112,30 @@ public boolean allowSortOnSubqueryWithoutLimit()
...
+            if (column.getJdbcType().isText())
+            {
+                //T-SQL will throw an error for nvarchar sizes >4000
+                //Use the common default max size to make type change to nvarchar(max)/text consistent
+                String size = column.getSize() == -1 || column.getSize() > SqlDialect.MAX_VARCHAR_SIZE ?
+                        "max" :
+                        column.getSize().toString();
+
+                statement = alterTableSegment + String.format(" ALTER COLUMN [%s] %s(%s) ",
+                        column.getName(),
```
labkey-matthewb: I'm surprised there's not more escaping of column names?  Maybe we always know column names don't have special chars at this level?

experiment/src/org/labkey/experiment/api/property/DomainPropertyImpl.java
```diff
@@ -803,6 +860,23 @@ else if (_pdOld != null)
...
     }
 
+    private void updateBooleanValue(String schemaTable, String column, String formatString)
+    {
+        column = OntologyManager.getExpSchema().getSqlDialect().makeLegalIdentifier(column);
+        BooleanFormat f = BooleanFormat.getInstance(formatString);
+        String trueValue = StringUtils.trimToNull(f.format(true));
+        String falseValue = StringUtils.trimToNull(f.format(false));
+        String nullValue = StringUtils.trimToNull(f.format(null));
+        SQLFragment sql = new SQLFragment("UPDATE ").append(schemaTable).append(" SET ").
```
labkey-matthewb, line 875: This deserves a comment.  updateBoolean only works on strings?  Does this mean we use strings today to store booleans in exp.objectproperty, but the string value is not stored consistently?

labkey-matthewb (changes requested):

api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
labkey-jeckels: No such thing! But I think I should use our hardcoded default of (15,4) instead. I extracted a constant.

api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
labkey-jeckels: Swapped

bigiron/src/org/labkey/bigiron/mssql/BaseMicrosoftSqlServerDialect.java
labkey-jeckels: I reused existing code that was there for resizing VARCHAR fields but did some extra testing.

This works in almost all cases because of the brackets. But not if you have a closing bracket. Switched to makeLegalIdentifer()

experiment/src/org/labkey/experiment/api/property/DomainPropertyImpl.java
labkey-jeckels: Comments added. Let me know if you think I should revise further.

labkey-matthewb (approved):

PR #2931, opened 2022-01-11
Issue 44579: XARs Allow Creation of Samples Without a Sample Type

labkey-jeckels:
#### Rationale
Don't allow XAR imports that reference sample types and data classes that don't exist and will cause problems later

#### Changes
* Switch warning dating back to 2008 to a hard error

### DIFF
--- a/experiment/src/org/labkey/experiment/xar/AbstractXarImporter.java
+++ b/experiment/src/org/labkey/experiment/xar/AbstractXarImporter.java
@@ -19,6 +19,7 @@ package org.labkey.experiment.xar;
 import org.apache.logging.log4j.Logger;
 import org.fhcrc.cpas.exp.xml.ExperimentArchiveType;
 import org.labkey.api.data.Container;
+import org.labkey.api.exp.ExperimentException;
 import org.labkey.api.exp.XarContext;
 import org.labkey.api.exp.XarSource;
 import org.labkey.api.exp.api.ExpData;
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/xar/AbstractXarImporter.java
```diff
@@ -46,38 +47,38 @@ public AbstractXarImporter(XarSource source, PipelineJob job)
...
 
-    protected void checkDataCpasType(String declaredType)
+    protected void checkDataCpasType(String declaredType) throws ExperimentException
     {
         if (declaredType != null && !ExpData.DEFAULT_CPAS_TYPE.equals(declaredType))
         {
             // check if this is a reference to a data class
             if (ExperimentService.get().getDataClass(declaredType) == null)
-                _job.getLogger().warn("Unrecognized CpasType '" + declaredType + "' loaded for Data object.");
+                throw new ExperimentException("Unrecognized CpasType '" + declaredType + "' loaded for Data object.");
```
labkey-susanh, line 56: Throwing an exception definitely seems more appropriate than just logging a warning here, but it does make me wonder about the original intent and whether this will render some Xar files impossible to import that previously would have resulted in most data in the right places (also not ideal, of course, just trying to think if we're introducing road blocks).

labkey-jeckels: @labkey-susanh it's certainly possible that we'll find use cases we want to support that are now treated as errors, but I'd rather encounter those errors and address them than end up with inconsistent data in the DB that can't be deleted.

I could be more conservative and only change to an error for the specific repro case (samples referring to an unknown sample type) but it seems like we're only postponing a related support ticket.

labkey-susanh: Seems fine, particularly given the undeletable status that results now.  Just wanted to check.

labkey-susanh (approved):

PR #2934, opened 2022-01-12
Preserve original IOException

bbimber:
This background is that we had an issue with the file root of a folder: the tomcat user didnt have R/W on it. This caused a cryptic error. Currently, the IOException is caught and rethrown as a RuntimeException; however, it didnt preserve the original IOException stack. This is a very minor change to keep that.


labkey-jeckels:
Change looks fine. I created a copy of the FB in our repo to trigger TeamCity validation.

labkey-jeckels (approved):

PR #2938, opened 2022-01-13
Roll back change from merge conflict.

labkey-danield:
#### Rationale
Bad merge conflict resolution from 21.11 into 22.1. Need to remove the reference to the log4j-1.2-api-2.17.1.jar.

#### Related Pull Requests
* None.

#### Changes
* Removed the reference to the log4j-1.2-api-2.17.1.jar.


labkey-adam (changes requested):
Not quite... the version was correct (2.17.1); that's what matches server/gradle.properties. All that needs to happen is `log4j-1.2-api-2.17.1.jar` line gets removed here, since that jar was eliminated on develop/release22.1.

labkey-adam (approved):
That looks right. Let's make sure the test is happy.

PR #2940, opened 2022-01-13
Eliminate useless ReloadTask and unused method

labkey-adam:
#### Rationale
`ReloadTask` began its life as a timer task then evolved to be a Quartz job, but we removed the scheduling feature when we introduced file watchers. ReloadTask never got the memo, though; it still implements Quartz Job but is never registered. And its `execute()` method is therefore never called. Sadly, I see no reason for ReloadTask to exist in the world any more...

I suspect more refactors of `StudyReload` will occur as we eliminate study archives, but this is a good checkpoint.


labkey-adam:
Code review tip: hide whitespace

labkey-jeckels (approved):

PR #2944, opened 2022-01-14
Fix to keep material and data lineage lookups separate

labkey-matthewb:
#### Rationale
The multi-valued lineage foreign key could misbehave if using both material and dataclass lookups in the same query.  This does not affect existing code since we don't do this.  However, planned features may use this.

Also an unrelated NPE fix.

#### Related Pull Requests
* 

#### Changes
*


labkey-susanh (approved):
Minimal testing on this branch suggests this is indeed a fix.  Thank you!

PR #2941, opened 2022-01-13
Issue 44540: More options for viewing the pipeline log

labkey-jeckels:
#### Rationale
Very large pipeline job logs cause problems for the detail page's log viewer

#### Changes
* Offer direct link to raw log file
* Centralize WebDav URL generation
* Cap detail page's log viewer at 10MB

PR #2946, opened 2022-01-15
Fix container path encoding problem introduced by PR 2941

labkey-jeckels:
#### Rationale
The TargetedMS suite caught a URL encoding problem but I didn't see it until I merged the platform PR.

https://teamcity.labkey.org/buildConfiguration/LabKey_2111Release_Community_ModuleSuites_TargettedMSSqlserve/1670482?buildTab=artifacts#%2FtomcatLogs.tar.gz

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2941

#### Changes
* Encode the container path too. Probably we should change PipeRootImpl.getWebdavUrl() so that it properly encodes, but there are bunch of callers and I haven't chased them all down to see if they'd double-encode yet.


labkey-adam (approved):

labkey-adam:
Looks like `api.ExpDataTableImpl$TestCase` is failing

PR #2949, opened 2022-01-17
Issue 43496: View names are not really case-insensitive

XingY:
#### Rationale
Custom views aren't case-insensitive but it's treated so on several UI areas.

#### Related Pull Requests
* 

#### Changes
* show all views for grid views dropdown and manage views

### DIFF
--- a/query/src/org/labkey/query/QueryServiceImpl.java
+++ b/query/src/org/labkey/query/QueryServiceImpl.java
@@ -779,7 +779,7 @@ public class QueryServiceImpl implements QueryService
     protected Map<String, CustomView> getCustomViewMap(@NotNull User user, Container container, @Nullable User owner, QueryDefinition qd,
                                                        boolean inheritable, boolean sharedOnly)
     {
-        Map<String, CustomView> views = new CaseInsensitiveHashMap<>();
+        Map<String, CustomView> views = new HashMap<>();
 
         // module query views have lower precedence, so add them first
         for (CustomView view : qd.getSchema().getModuleCustomViews(container, qd))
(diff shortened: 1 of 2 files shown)


query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -779,7 +779,7 @@ protected QueryDefinition getQueryDefMetadataOverride(User user, @NotNull Contai
     protected Map<String, CustomView> getCustomViewMap(@NotNull User user, Container container, @Nullable User owner, QueryDefinition qd,
                                                        boolean inheritable, boolean sharedOnly)
     {
-        Map<String, CustomView> views = new CaseInsensitiveHashMap<>();
+        Map<String, CustomView> views = new HashMap<>();
```
labkey-nicka, line 782: From a convention perspective it seems a bit odd to me that "schemas" and "queries" are case-insensitive but "views" are case-sensitive.  I would not expect this. Could we consider going the other way and preventing the persistence of views with the same name case-insensitively?

labkey-nicka (commented):
I'd like to discuss further.

query/src/org/labkey/query/QueryServiceImpl.java
labkey-adam: I tend to agree. We were trying (and apparently failing) to be case-insensitive here. In addition to schema & query consistency, I think case-insensitive will ensure consistency of view naming between PostgreSQL and Microsoft SQL Server.

XingY: Thanks both. updated ticket and closing this PR

PR #2954, opened 2022-01-18
Item 9875: Assay QC state option to require comment on QC state change for assay run

cnathe:
#### Rationale
The NIAID Response Team has a FANG ELISA assay which they would like to use LabKey for the Assay QC workflow. They will be using Assay run QC states to track passing and failing plate/sample runs. The initial QC state for each run will be set during the file watcher pipeline import task, but users with proper permissions will be allowed to update QC states for a run after review within the LabKey UI. This story adds a feature so that an admin can set, via the Manage QC states page, that a comment is required when changing an assay run QC state.

#### Related Pull Requests
* https://github.com/LabKey/niaidElisa/pull/1
* https://github.com/LabKey/premium/pull/305
* https://github.com/LabKey/platform/pull/2954
* https://github.com/LabKey/sampleManagement/pull/806
* https://github.com/LabKey/testAutomation/pull/969

#### Changes
* add "requireCommentOnQCStateChange" setting to qcStates.xsd and support round tripping that setting via folder export / import
* AssayQCService update to add getter and setter for RequireCommentOnQCStateChange flag
* Assay Manage QC States UI update to add new panel for setting the "require comment" flag
* UpdateQCStateAction update to validate required comment based on the container flag

### DIFF
--- a/api/schemas/qcStates.xsd
+++ b/api/schemas/qcStates.xsd
@@ -46,7 +46,7 @@
                 <xs:element name="requireCommentOnQCStateChange" type="xs:boolean" default="false" minOccurs="0">
                     <xs:annotation>
                         <xs:documentation>
-                            This setting determines whether a comment is required when updating an assay run QC state.
+                            This setting determines whether a comment is required when updating a QC state.
                         </xs:documentation>
                     </xs:annotation>
                 </xs:element>
--- a/api/schemas/qcStates.xsd
+++ b/api/schemas/qcStates.xsd
@@ -43,6 +43,13 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:element>
+                <xs:element name="requireCommentOnQCStateChange" type="xs:boolean" default="false" minOccurs="0">
+                    <xs:annotation>
+                        <xs:documentation>
+                            This setting determines whether a comment is required when updating an assay run QC state.
+                        </xs:documentation>
+                    </xs:annotation>
+                </xs:element>
                 <xs:element name="insertUpdateDefault" type="xs:string" minOccurs="0">
                     <xs:annotation>
                         <xs:documentation>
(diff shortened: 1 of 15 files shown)


api/schemas/qcStates.xsd
```diff
@@ -43,6 +43,13 @@
                         </xs:documentation>
                     </xs:annotation>
                 </xs:element>
+                <xs:element name="requireCommentOnQCStateChange" type="xs:boolean" default="false" minOccurs="0">
+                    <xs:annotation>
+                        <xs:documentation>
+                            This setting determines whether a comment is required when updating an assay run QC state.
```
labkey-klum: maybe not tie this specifically to assays, since the same option could one day be used for datasets as well

assay/src/org/labkey/assay/AssayController.java
```diff
@@ -1598,38 +1598,43 @@ public void addNavTrail(NavTree root)
...
                         break;
                 }
 
-                if (run != null)
+                if (_firstRun != null)
                 {
-                    DataState state = DataStateManager.getInstance().getStateForRowId(run.getProtocol().getContainer(), form.getState());
-                    if (state != null)
-                        svc.setQCStates(run.getProtocol(), getContainer(), getUser(), List.copyOf(form.getRuns()), state, form.getComment());
+                    if (AssayQCService.getProvider().isRequireCommentOnQCStateChange(_firstRun.getProtocol().getContainer()) && form.getComment() == null)
```
labkey-klum: should this be : `StringUtils.isBlank(form.getComment())`, or are we already trimming to null.

labkey-klum (approved):

api/schemas/qcStates.xsd
cnathe: done

assay/src/org/labkey/assay/AssayController.java
cnathe: good catch, I was not trimming it. Changed to use StringUtils.isBlank.

PR #2955, opened 2022-01-18
Issue 44686: Put Lucene index in a versioned subfolder if in dev mode

labkey-adam:
#### Rationale
Reduce churn when switching between branches with different Lucene versions

Fix exception when visiting export index action


labkey-jeckels (approved):
Code seems fine but there's a search test failure that could be related.

labkey-adam:
I'm actually amazed that so many of the search tests passed; they never pass. But a re-run is completely green so obviously my code is perfect.

PR #2960, opened 2022-01-19
Issue 44557: Optimize queries within Panorama QC folders for faster plot rendering

labkey-jeckels:
#### Rationale
Fast pages are good. QC folders in Panorama are slower than we want, and there are some platform changes to help improve the SQL we run.

Also, on my dev machine where I have many notifications queued up (4000+) every page render was taking 200ms to stream them all back into Java objects when all we cared about was the count.

#### Changes
* Make LazyForeignKey even lazier, meaning it's less likely to initialize the full FK
* Offer the execution plan for queries that have a CTE
* Expose a QueryService method to turn LabKey SQL into minimal TableInfos to be embedded in other, larger SQL queries

### DIFF
--- a/api/src/org/labkey/api/admin/notification/NotificationService.java
+++ b/api/src/org/labkey/api/admin/notification/NotificationService.java
@@ -69,6 +69,12 @@ public interface NotificationService
     List<Notification> getNotificationsByUser(Container container, int notifyUserId, boolean unreadOnly);
 
     /*
+     * Returns just the count of notifications for a specific user. At any given instant, will match the length of the
+     * list returned by getNotificationsByUser(), and is significantly more efficient to query.
+     */
+    long getNotificationCountByUser(Container container, int notifyUserId, boolean unreadOnly);
+
+    /*
      * Returns a list of notifications for a specific user based on the specified type.
      */
     List<Notification> getNotificationsByType(Container container, @NotNull String type, int notifyUserId, boolean unreadOnly);
--- a/api/src/org/labkey/api/data/Table.java
+++ b/api/src/org/labkey/api/data/Table.java
@@ -171,10 +171,11 @@ public class Table
             sqlLine = sqlLine.trim();
             if (!sqlLine.startsWith("--"))
             {
-                // First non-comment line must start with SELECT
+                // First non-comment line must start with SELECT or WITH (for a CTE)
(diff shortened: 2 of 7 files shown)


api/src/org/labkey/api/data/Table.java
```diff
@@ -171,10 +171,11 @@ public static boolean isSelect(String sql)
             sqlLine = sqlLine.trim();
             if (!sqlLine.startsWith("--"))
             {
-                // First non-comment line must start with SELECT
+                // First non-comment line must start with SELECT or WITH (for a CTE)
```
labkey-matthewb: Presence of WITH is does not mean that the whole query is a SELECT.  I would like to see this augmented to scan through and make sure that in addition a line starting with SELECT, that there is no line starting with "UPDATE", "DELETE" or "INSERT".  Having a SQLFragment version might make for more accurate/easier parsing, since the CTE's get broken out separately from the 'main' SQL statement.

api/src/org/labkey/api/admin/notification/NotificationService.java
```diff
@@ -68,6 +68,12 @@ Notification sendMessageForRecipient(Container c, User createdByUser, User recip
      */
     List<Notification> getNotificationsByUser(Container container, int notifyUserId, boolean unreadOnly);
 
+    /*
+     * Returns just the count of notifications for a specific user. At any given instant, will match the length of the
+     * list returned by getNotificationsByUser(), and is significantly more efficient to query.
+     */
+    long getNotificationCountByUser(Container container, int notifyUserId, boolean unreadOnly);
```
labkey-matthewb, line 75: 👍

labkey-matthewb (changes requested):

api/src/org/labkey/api/data/Table.java
labkey-jeckels: I will revert this change. I was focused on whether we show a link to get the execution plan in the Admin Console's Queries section where it's OK to be a little sloppy. However, the other usage in PostgresSql91Dialect is much more important to be sure, and without doing a full parse we be as confident (there's no guarantee that UPDATE or DELETE would be at the beginning of a line, for example)

labkey-matthewb (approved):

PR #2965, opened 2022-01-20
introduce Issue.Builder 

labkey-klum:
#### Rationale
We want to be able to support the ability to create/update issues in java using the `IssueService`. In order to make this possible (or easier) we've introduced a `builder` pattern that allows users to create an issue model that can then be passed around through the `IssueService`.

#### Related Pull Requests
- https://github.com/LabKey/hjfSampleRequests/pull/17

#### Changes
- Expose `Issue.Builder` class
- Allow updates to existing issues while passing in the sparse map of updates to perform
- Add a service method to locate an `IssueListDefinintion` by name


issues/api-src/org/labkey/api/issues/Issue.java
```diff
@@ -78,4 +86,323 @@ public void checkPermission(Container c, User user, Issue issue)
...
+        public static class IssueImpl extends Entity implements Issue
+        {
+            private Container _container;
+            private Integer _issueId;
+            private Integer _issueDefId;
+            private String _issueDefName;
+            private Map<String, Object> _properties = new CaseInsensitiveHashMap<>();
+            private String _related;
+
+            // for ObjectFactory
```
labkey-nicka, line 263: Tis a bit unfortunate. If this constructor wasn't here all items on this class could be `final`.

issues/api-src/org/labkey/api/issues/IssueService.java
```diff
@@ -28,4 +30,23 @@ static void setInstance(IssueService impl)
...
+     * be made to the existing issue. The returned issue can then be updated using any of the save methods
+     * in this interface.
+     * @param container
+     * @param user
+     * @param issueId - The id of the issue to fetch
+     * @param updates - The map which represents the changes to the existing issue that should be applied.
+     * @return
+     */
+    @Nullable
+    Issue getIssue(Container container, User user, Integer issueId, Map<String, Object> updates);
```
labkey-nicka: nit: This reads a little weird. Maybe `modifyIssue()` or `updateIssue()`? 

Another pattern could be to remove this variant of `getIssue()` and put an `update()` or `edit()` method directly on the `Issue` class that is built which returns a new `Issue` rebuilt using the ObjectFactory map.

labkey-nicka (approved):

PR #2970, opened 2022-01-25
44544: Sample type linked to study gives error on Manage Dataset > Edit Definition Page

labkey-klum:
#### Rationale
I think this was an early refactor of the code in preparation for supporting linking samples to study and it never got cleaned up. In the final versions, the publish source is always an `ExpObject` and you have to check to see if it's either an `Assay` or a `SampleType`

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44544


RosalineP (approved):
Looks good! Fix verified—thanks for cleaning this up!

PR #2971, opened 2022-01-25
Fixes for non-provisioned field type changes

labkey-jeckels:
#### Rationale
Issue 44723: Error while converting from INTEGER to DOUBLE for non-provisioned table
Issue 44724: Changing the boolean field to String is emptying the values in other columns for non provisioned domain

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/978

#### Changes
* Limit boolean formatting to just the property that's being edited
* Allow integer->decimal changes


labkey-sweta (approved):
Test is passing after the changes

PR #2973, opened 2022-01-25
Add some extra logging to DataSetTest

labkey-danield:
#### Rationale
Cleaning up a DataSetTest. Test continues to fail, but a little more logging has been added.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/980

#### Changes
* In StudyDatasetTest.verifyFilterPanelOnDemographics remove check for paging control, the check for the number of rows in the grid will catch the same error.
* Added some logging to help track down the failure.


labkey-tchad (commented):
I'm not sure this will give us much information. I'm working on a fix in DataRegionTable to make the facet panel more reliable.

labkey-danield:
@labkey-tchad 
Sure it may not add much but extra logging doesn't hurt.

This also:
- Removes the assert that has no info
- Removes the redundant check for number of rows returned after the filter is applied
- Addresses the timing issue where it assume the paging control will disappear immediately after the filter is click
- Reprioritizes the wait for a grid element to appear before checking if the number of rows in the grid is as expected.

labkey-tchad (approved):

PR #2981, opened 2022-01-26
Move some useful methods into new api class LabKeySql

labkey-matthewb:
#### Rationale
This code will be used to help clean up some code in the experiment module.  Also, it's just useful to put somewhere since we don't use the same tools to generate LabKey SQL as we use to generate 'database' SQL.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/sql/LabKeySql.java
+++ b/api/src/org/labkey/api/sql/LabKeySql.java
@@ -1,0 +1,36 @@
+package org.labkey.api.sql;
+
+import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.NotNull;
+
+
+public class LabKeySql
+{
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/sql/LabKeySql.java
```diff
@@ -0,0 +1,36 @@
+package org.labkey.api.sql;
+
+import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.NotNull;
+
+
```
labkey-adam, line 6: Please add a comment here about using these helpers only on LabKey SQL. I know it seems obvious to you, but people get confused...

labkey-adam (approved):
No need to re-review, but please add a comment

PR #2989, opened 2022-01-28

labkey-jeckels:
#### Rationale
Needing multiple DB connections in a single HTTP request thread is a recipe for deadlocks

#### Changes
* Reuse the Connection from the primary query to get aggregates
* Don't do a separate aggregate query when we only need the total row count and already know it


labkey-susanh (approved):
Seems reasonable to me, but I'm not completely versed in nuances here

labkey-adam (approved):

PR #2990, opened 2022-01-28
44748: GetIssueAction doesn't respect SeeUserAndGroupDetailsRole

labkey-klum:
#### Rationale
`Issues.getIssue.api` was leaking email and user details information to users who did not have the `SeeUserAndGroupDetailsRole` security role: 

https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=44748

#### Changes
- Check for `SecurityManager.canSeeUserDetails` in `User.getUserProps`
- Don't include properties : `notifyListEmail` and `notifyListUserEmail` in `issues.getIssue.api` is the user is not allowed to see user and group details.

### DIFF
--- a/api/src/org/labkey/api/security/User.java
+++ b/api/src/org/labkey/api/security/User.java
@@ -602,9 +602,12 @@ public class User extends UserPrincipal implements Serializable, Cloneable
         JSONObject props = new JSONObject();
 
         props.put("id", user.getUserId());
-        props.put("displayName", user.getDisplayName(currentUser));
-        props.put("email", user.getEmail());
-        props.put("phone", user.getPhone());
+        if (container == null || SecurityManager.canSeeUserDetails(container, currentUser))
+        {
+            props.put("displayName", user.getDisplayName(currentUser));
+            props.put("email", user.getEmail());
+            props.put("phone", user.getPhone());
+        }
         props.put("avatar", user.getAvatarThumbnailPath());
 
         if (includePermissionProps)
--- a/api/src/org/labkey/api/security/User.java
+++ b/api/src/org/labkey/api/security/User.java
@@ -604,10 +604,10 @@ public class User extends UserPrincipal implements Serializable, Cloneable
         props.put("id", user.getUserId());
         if (container == null || SecurityManager.canSeeUserDetails(container, currentUser))
         {
-            props.put("displayName", user.getDisplayName(currentUser));
             props.put("email", user.getEmail());
             props.put("phone", user.getPhone());
         }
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/security/User.java
```diff
@@ -602,9 +602,12 @@ public static JSONObject getUserProps(User user, User currentUser, @Nullable Con
         JSONObject props = new JSONObject();
 
         props.put("id", user.getUserId());
-        props.put("displayName", user.getDisplayName(currentUser));
-        props.put("email", user.getEmail());
-        props.put("phone", user.getPhone());
+        if (container == null || SecurityManager.canSeeUserDetails(container, currentUser))
+        {
+            props.put("displayName", user.getDisplayName(currentUser));
```
labkey-nicka: It seems a bit restrictive to not include a "displayName". This is how the user is identified by humans. There is logic inside of `getDisplayName()` about how we handle "sanitizing" the displayName although that logics seems somewhat deprecated these days and due for reassessment.

labkey-nicka (commented):
Additionally, and beyond the scope of this PR, it seems odd that I could see a user's email from one container but not another. Why is seeing user details related to a container? Shouldn't this be a site-wide permission? The container was only being passed into this method to assist with gathering permissions for that container and in a majority of use-cases for this method is `null`. @labkey-jeckels @labkey-adam

api/src/org/labkey/api/security/User.java
labkey-klum: Good point, that was not intentional, I've since modified that code to always include `displayName`

labkey-nicka (approved):

api/src/org/labkey/api/security/User.java
labkey-jeckels: As Nick mentioned, this may be out of scope for this PR, but before we make any changes to viewing email addresses we should carefully consider scenarios where unrelated groups use different projects on a single server.

Typically SeeUserDetailsPermission is indeed granted by a site-wide role. I'm seeing an EHR container-scoped role, EHRDataEntryRole, but would have to dig into history to re-derive the specific use case for it.

labkey-adam: All folder and project admins get this permission (implicitly), and are able to see details for the users in "their" folders. That's why you might see user details in one folder but not another. And why `Container` is passed into that method.

We added the site role for the special case where admins want to grant non-admins the ability to see all user details. This can make sense for sites that are not open to the public and whose collaborators know each other.

PR #2992, opened 2022-01-28
Stream documents to be indexed

labkey-matthewb:
#### Rationale
We currently buffer the tika converted documents in memory (the text version of whatever file is being indexed).  These files can be very large.  This change allows for these files to overflow to disk to limit the indexers demands on the java heap.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/util/FileUtil.java
+++ b/api/src/org/labkey/api/util/FileUtil.java
@@ -1353,38 +1353,37 @@ quickScan:
     public static class TempTextFileWrapper implements Closeable
     {
         final int characterLimitInMemory;
-        ByteBuffer _byteBuffer = null;
-        CharBuffer _charBuffer = null;
-//        int _maxPosition = 0;
+        final ByteBuffer _byteBuffer;
+        final CharBuffer _charBuffer;
         FileWriter _fileWriter = null;
         FileReader _fileReader = null;
         File _tmpFile = null;
+        boolean closed = false;             // so we can ignore multiple calls to close
 
-        Writer _writer;
-        Reader _reader;
+        Writer _writer = null;
+        Reader _reader = null;
 
         public TempTextFileWrapper(int characterLimitInMemory)
         {
             this.characterLimitInMemory = characterLimitInMemory;
+            this._byteBuffer = ByteBuffer.allocate(characterLimitInMemory * 2);
+            this._charBuffer = _byteBuffer.asCharBuffer();
         }
 
         public TempTextFileWrapper(CharBuffer charBuffer)
         {
-            this._charBuffer = charBuffer;
             this.characterLimitInMemory = charBuffer.capacity();
+            this._byteBuffer = null;
+            this._charBuffer = charBuffer;
         }
 
 
         public Writer getWriter()
         {
-            // should only ask for _writer once
-            assert null == _writer;
+            if (null != _writer || closed)
+                throw new IllegalStateException(closed ? "TempTextFileWrapper is closed" : "getWriter() called twice");
+
             // CONSIDER ByteBuffer.allocateDirect(), for now caller can pass in a direct buffer if desired
-            if (null == _charBuffer)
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -1329,6 +1342,168 @@ else if (!FileUtil.hasCloudScheme(path))
...
+            this.characterLimitInMemory = charBuffer.capacity();
+        }
+
+
+        public Writer getWriter()
+        {
+            // should only ask for _writer once
+            assert null == _writer;
+            // CONSIDER ByteBuffer.allocateDirect(), for now caller can pass in a direct buffer if desired
+            if (null == _charBuffer)
```
labkey-adam: Why not do this allocation in `TempTextFileWrapper(int)`, passing the `charBuffer` to `TempTextFileWrapper(CharBuffer)`?

api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -1329,6 +1342,168 @@ else if (!FileUtil.hasCloudScheme(path))
...
+     * direct to cache the first portion of the file to avoid hitting the
+     * file system if the file is smaller.
+     *
+     * The caller needs to call close() on this object or the Reader returned
+     * by getReader().  Calling close on both is OK.
+     */
+    public static class TempTextFileWrapper implements Closeable
+    {
+        final int characterLimitInMemory;
+        ByteBuffer _byteBuffer = null;
```
labkey-adam: `@Nullable`

api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -1329,6 +1342,168 @@ else if (!FileUtil.hasCloudScheme(path))
...
+        CharBuffer _charBuffer = null;
+//        int _maxPosition = 0;
```
labkey-adam: Remove?

api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -1329,6 +1342,168 @@ else if (!FileUtil.hasCloudScheme(path))
...
+                @Override
+                public void write(@NotNull char[] cbuf, int off, int len) throws IOException
+                {
+                    if (closed)
+                        throw new IOException("Writer is closed");
+                    if (_charBuffer.remaining() > 0)
+                    {
+                        var l = Math.min(_charBuffer.remaining(), len);
+                        _charBuffer.put(cbuf, off, l);
+//                        _maxPosition = Math.max(_maxPosition, _charBuffer.position());
```
labkey-adam: remove?

api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -1329,6 +1342,168 @@ else if (!FileUtil.hasCloudScheme(path))
...
+     * file system if the file is smaller.
+     *
+     * The caller needs to call close() on this object or the Reader returned
+     * by getReader().  Calling close on both is OK.
+     */
+    public static class TempTextFileWrapper implements Closeable
+    {
+        final int characterLimitInMemory;
+        ByteBuffer _byteBuffer = null;
+        CharBuffer _charBuffer = null;
```
labkey-adam: Could be `final` if initialized in the first constructor

labkey-adam (changes requested):
Couple suggestions for `TempTextFileWrapper`

api/src/org/labkey/api/util/FileUtil.java
labkey-matthewb: Moving the _charBuffer allocation is makes sense, but I want to make sure the caller only asks for the Writer/Reader once.  I don't want the interface to look like might be more than one writer (or reader) that separately track position.  Using null as a marker is simple.

labkey-adam (approved):

PR #2993, opened 2022-01-28
Item 9923: Add new storage role permissions classes

labkey-susanh:
#### Rationale
We are adding new roles and permissions specific to storage. These roles are not meant to be available to Admins by default.  The Storage Editors need to be able to manage picklists and jobs, but not assay data, samples or data classes. They are also allowed to update the status of samples when they are being discarded from storage (A later story will probably allow this more broadly.). The updating of sample status for stored samples is handled through an action in the `InventoryController`.  Here we add the permission class.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/723 (and related)

#### Changes
* Add `SampleWorkflowJobPermission` and `UpdateSampleStatusPermission`
* Allow workflow runs to be created, updated, and deleted for users with `SampleWorkflowJobPermission`.


experiment/src/org/labkey/experiment/api/ExpRunImpl.java
```diff
@@ -554,6 +555,15 @@ public void deleteProtocolApplications(List<ExpDataImpl> datasToDelete, User use
         ExperimentRunGraph.clearCache(getContainer());
     }
 
+    @Override
```
XingY, line 558: Should the check also be done in this class's "public void delete(User user)" method?

api/src/org/labkey/api/exp/api/ExpRun.java
```diff
@@ -123,4 +123,6 @@
     void setWorkflowTask(@Nullable ExpProtocolApplication workflowTask);
```
XingY, line 124: I don't see permission check updates in ExpRunTableImpl.RunTableUpdateService. Is the idea that workflow actions will be done exclusively through LKSM actions rather than query API?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -3005,17 +3075,17 @@ public ApiResponse execute(DeleteRunForm form, BindException errors)
             {
                 throw new NotFoundException("Could not find run with ID " + form.getRunId());
             }
-            if (!run.getContainer().hasPermission(getUser(), DeletePermission.class))
-            {
-                throw new UnauthorizedException("Not permitted");
-            }
+            if (!run.canDelete(getUser()))
+                throw new UnauthorizedException("You do not have permission to delete "
```
XingY, line 3079: " in this container." instead of " runs in this container", with an extra "runs"

XingY (changes requested):

experiment/src/org/labkey/experiment/api/ExpRunImpl.java
labkey-susanh: Good call.  Done.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: Done. Thanks.

api/src/org/labkey/api/exp/api/ExpRun.java
labkey-susanh: The ExperimentListener implementation in SM should do the checking we need since the UpdateService calls `run.save` when then invokes the listeners.

XingY (approved):

PR #3001, opened 2022-01-31
List: cross-folder domains and data

labkey-nicka:
#### Rationale
This PR updates LKS' List data structure to support cross-folder list domain resolution and cross-folder list data.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/725
* https://github.com/LabKey/biologics/pull/1139
* https://github.com/LabKey/platform/pull/3001
* https://github.com/LabKey/sampleManagement/pull/823
* https://github.com/LabKey/inventory/pull/372

#### Changes
* Support resolving Lists domain's cross-folder. An additional bit, `includeProjectAndShared`, has been added to `ListService.getLists()` to control this behavior. Defaults to true.
* `FolderListWriter` writes to each folder so results are resolved on a per-folder basis.
* (Unrelated to Lists) Data Class table explicit override of supplied container filter removed. Now always respects what is provided.


list/src/org/labkey/list/model/FolderListWriter.java
```diff
@@ -47,7 +47,7 @@ public void write(Container object, ImportContext<FolderDocument.Folder> ctx, Vi
     {
         Container c = ctx.getContainer();
 
-        if (ListService.get().hasLists(c))
+        if (ListService.get().hasLists(c, false))
```
labkey-jeckels: This is tricky. If a list is defined at the project level and you export a child folder, based on this diff, I'm guessing it won't include the list definition or any rows from the subfolder. This is different (I believe) from how things like assays and their runs behave, where I believe it includes the assay design and the runs from the current folder.

It's hard to say what the "correct" behavior is, but I wonder if we should try to be more consistent

labkey-jeckels (approved):

labkey-susanh (approved):

list/src/org/labkey/list/model/FolderListWriter.java
labkey-nicka: Thanks for the insight. I've iterated through the combinations to see what gets exported and I agree that switching to `true` is probably what we want. Things I noted with it being `true`:

1. Data from subfolders is only included if you check "Include subfolders" and the data is included in a separate .tsv within that exported folder.
2. Lists defined in the /Shared folder appear in each exported folder, however, any data from the /Shared folder in those lists is not exported. I'm not really sure what is expected when exporting /Shared data. Maybe we need an "Include /Shared" option upon export? What would that look like for import?

labkey-jeckels: 1. This sounds ideal.
2. This is extra tricky. Here's a proposal: don't export the list definition from /Shared as part of the archive. But do include a .tsv file for the current container's data rows (if any). And on import look for a match for the .tsv file, logging a WARN or INFO level message saying that we didn't find a list and are ignoring the data rows.

labkey-nicka: This behavior has since been reverted https://github.com/LabKey/platform/pull/3264.

PR #3003, opened 2022-01-31
Avoid NPEs when restarting a server while impersonating a role that is not present after the restart

labkey-susanh:
#### Rationale
If you happen to be impersonating a role that exists in one branch of the code and then rebuild your server for a different branch that doesn't have that role, when your server restarts with the new code, you can get NPEs in various places as it tries to determine the roles you are in.  

#### Changes
* Check for existence of roles in various places

### DIFF
--- a/api/src/org/labkey/api/security/impersonation/RoleImpersonationContextFactory.java
+++ b/api/src/org/labkey/api/security/impersonation/RoleImpersonationContextFactory.java
@@ -39,7 +39,9 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedList;
+import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * User: adam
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/security/impersonation/RoleImpersonationContextFactory.java
```diff
@@ -138,9 +140,11 @@ public void startImpersonating(ViewContext context)
...
     private String getRolesDisplayString(Set<String> roleNames)
     {
-        Set<String> roleDisplayNames = new HashSet<>();
-        for (String name : roleNames)
-            roleDisplayNames.add(RoleManager.getRole(name).getName());
+        Set<String> roleDisplayNames = roleNames.stream()
+                .map(RoleManager::getRole)
+                .filter(Objects::nonNull)
+                .map(Role::getName)
+                .collect(Collectors.toSet());
```
labkey-adam, line 147: Could `.collect(Collectors.joining(", ")` instead (and then tack on the prefix and suffix)

labkey-adam (approved):

api/src/org/labkey/api/security/impersonation/RoleImpersonationContextFactory.java
labkey-adam: (Not required, just a suggestion)

labkey-susanh: Yeah, might have done it if I thought of it first, but I'll avoid more churn here.

PR #3005, opened 2022-02-01
automated test fixes : a user should always be able to see their own details

labkey-klum:
#### Rationale
The previous change to not leak user details from the `GetIssueAction` caused some test failures by being too restrictive:

https://github.com/LabKey/platform/pull/2990

Added a slight change to allow users to see their own details even if they aren't in the `SeeUserAndGroupDetailsRole`.

### DIFF
--- a/api/src/org/labkey/api/security/User.java
+++ b/api/src/org/labkey/api/security/User.java
@@ -602,7 +602,8 @@ public class User extends UserPrincipal implements Serializable, Cloneable
         JSONObject props = new JSONObject();
 
         props.put("id", user.getUserId());
-        if (container == null || SecurityManager.canSeeUserDetails(container, currentUser))
+        // check for permission to see user details, users can always see their own details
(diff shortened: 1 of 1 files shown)


cnathe (approved):

labkey-susanh:
@labkey-klum This seems like more than just a test fix. Should we be targeting 22.2?

cnathe:
> @labkey-klum This seems like more than just a test fix. Should we be targeting 22.2?

@labkey-susanh FYI, Karl has backported this change to 22.2 with this PR: https://github.com/LabKey/platform/pull/3010

labkey-susanh:
> > @labkey-klum This seems like more than just a test fix. Should we be targeting 22.2?
> 
> @labkey-susanh FYI, Karl has backported this change to 22.2 with this PR: #3010

Ah, OK.  Thanks @cnathe.

PR #3006, opened 2022-02-02
Issue 44597: check status before deriving samples from LKS UI.

labkey-susanh:
#### Rationale
Issue 44597: When samples are locked, we shouldn't allow derivation of new samples. This code path is for LKS only and was missed in the original implementation of restrictions based on status.

#### Changes
* Check for sample status when deriving samples from LKS UI.


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -5022,9 +5022,14 @@ public boolean handlePost(DeriveMaterialForm form, BindException errors) throws
             List<ExpMaterial> materials = form.lookupMaterials();
 
             Map<ExpMaterial, String> inputMaterials = new LinkedHashMap<>();
+            List<ExpMaterial> lockedSamples = new ArrayList<>();
```
Sigmonia, line 5020: Should this be moved to the ValidateCommand method?

Sigmonia (approved):

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: Good suggestion.  Done.

PR #3007, opened 2022-02-02
Remove study import/reload. Improve generics throughout FolderImporter hierarchy.

labkey-adam:
#### Rationale
Study import/reload functionality has been redundant with folder import/reload (introduced 10 years ago). Removing study import/reload eliminates a large amount of duplicative code and allows type safety throughout the `FolderImporter` hierarchy and usages.

#### Changes
* Remove study import pipeline job, "Reload Study" file watcher, CheckForReload API action and support for "studyload.txt", and tests for these
* Remove type parameters from FolderImporter and its methods. FolderImporter is always typed as `<Folder>` and its always methods receive a FolderImportContext. Update all the implementations.
* Eliminate wonky `getImporterSpecificImportContext()`; pass appropriate context to `getChildrenDataTypes()` and let implementations create whatever context it needs internally.


study/webapp/WEB-INF/study/studyContext.xml
```diff
@@ -108,49 +86,6 @@
...
-                                <constructor-arg index="0" value="failForUndefinedVisits"/>
-                                <constructor-arg index="1" value="Fail for Undefined Visits"/>
-                                <constructor-arg index="2" value="false" />
-                                <constructor-arg index="3" value="false"/>
-                                <constructor-arg index="4" value="Enable to fail when undefined visits are encountered"/>
-                            </bean>
-                        </list>
-                    </property>
-                </bean>
             </list>
```
labkey-klum, line 89: What happens to the existing file watcher configurations? Aren't there clients out there who are using this file watcher. Or did it get moved somewhere else and I didn't notice?

labkey-klum (approved):

study/webapp/WEB-INF/study/studyContext.xml
labkey-adam: Existing study reload file watcher configurations will stop working. Account managers have indicated their clients are all clear (not using study import/reload). Recent metrics show no study import jobs having run in the last two weeks, assuming I'm reading this correctly: https://www.labkey.org/_mothership/query-executeQuery.view?schemaName=mothership&query.queryName=recentJsonMetricValues&query.SelectKey~contains=jobCounts&query.offset=100  @labkey-jeckels any suggestions for other metrics to check?

labkey-klum: Okay, I don't know if the USF server is reporting metrics but they were heavily using the legacy study reload at one point. We had an upgrade script to convert the legacy settings to file watcher configurations so they may not even be aware that they have file watchers setup. I don't know if there were others in this camp but USF sticks out in my mind.

labkey-adam: I had explicitly mentioned USF in my email to the AMs. But, if needed, they can switch over to a "Reload folder archive" watcher easily. I'll make sure the release notes explains how to migrate.

labkey-jeckels: I added the metrics for the pipeline job counts last month, so it's only available for servers that are running builds from develop or the newly branched 22.2. I don't know of other helpful metrics here, which is one reason I added them. https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=44605

PR #3011, opened 2022-02-03
Issue 44408: Updates for deactivated users

labkey-susanh:
#### Rationale
If users have become deactivated, we want to log them out when they make requests and not honor their credentials.


api/src/org/labkey/api/websocket/BrowserEndpoint.java
```diff
@@ -75,6 +75,11 @@ public void modifyHandshake(ServerEndpointConfig config, HandshakeRequest reques
         {
             HttpSession httpSession = (HttpSession) request.getHttpSession();
             User user = SecurityManager.getSessionUser(httpSession);
+            if (user != null && !user.isActive())
+            {
+                httpSession.invalidate();
+                user = null;
+            }
```
labkey-adam: Is this and the `NotificationEndpoint` change necessary? This just gets called at HTTP upgrade time... doesn't AuthFilter handle the inactive user before upgrade occurs?

Why not push the active user check into `SecurityManager.getSessionUser()` (returning null for inactive)? Seemed like a good chokepoint to avoid duplication and future mistakes.

labkey-susanh: Isn't this used for WebSocket notifications?  it may not be totally necessary, but probably we want to shut those down quickly too.

I tried putting this in `SecurityManager.getSessionUser`, but since I have only the session there and not the request, and the request can be of different types from these different callers, I can't call `SecurityManager.logoutUser`, which seems the cleanest thing to do for `HttpServletRequests`.   Moving the handling into the three distinct callers and letting them do the most natural thing seemed the next best answer.

labkey-adam: Yes, we use WebSockets for notifications and ideally we'd shut down these connections. But my _assumption_ is that `modifyHandshake()` gets called only at the point the socket is established... which I believe always begins with a normal HTTP request that then gets upgraded (and is subject to the inactive user check in that mode). But I know almost nothing about WebSockets... would be good for @labkey-matthewb and @cnathe to take a look, since they've worked with WS much more than I.

I would be in favor of pushing the inactive check into `SecurityManager.getSessionUser()` for consistency, even if it means separate method signatures to handle the different invalidation steps each request type.

As I said in standup, I think what's here addresses the key issue, so these changes aren't required.

cnathe: I did some testing of this locally. I pulled down this fb and commented out these changes in `BrowserEndpoint.java` and `NotificationEndpoint.java` and I get the expected behavior for a user that is logged in and viewing LKS or one of the apps and then is deactivated while they are logged in. In that case, the next time they make an API call or do a page load, the code change in `SecurityManager.java` calls `logoutUser` and the websocket listener in their case then sees that event and puts up the "you have been logged out" modal. So I don't think these changes are necessary (though likely not hurting anything).

As Adam mentioned, I believe that `modifyHandshake` is only called on the page load. So if a user was already sitting on a page or in the app when they were deactivated, this wouldn't get called for an API call they make.

labkey-susanh: Yes, if the user interacts with the app, they will make an HttpServletRequest and hit the other code.  Modifications in the other classes were for the presumed case of WebSocket listening (e.g., for Job updates) when the user does not need to interact with the browser to get additional info.  Definitely an edge case, and I don't know how to test that well.

All that said, I don't have a strong opinion here. @labkey-adam since you seem to have the strongest opinion, I'll make two signatures for `getSessionUser` as you suggest.

api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -483,10 +506,15 @@ public static User getSessionUser(HttpSession session)
...
-        User sessionUser = getSessionUser(session);
+        User sessionUser = getSessionUser(request);
 
         if (null != sessionUser)
         {
+            if (!sessionUser.isActive())
+            {
+                SecurityManager.logoutUser(request, sessionUser, null);
+                return null;
+            }
```
labkey-adam: This block is no longer necessary, right?

api/src/org/labkey/api/websocket/BrowserEndpoint.java
labkey-adam: What, me, opinions?? Thanks @labkey-susanh  and @cnathe .

labkey-adam (approved):
I think one code block can be removed, but no need for re-review

PR #3015, opened 2022-02-04
Lists: improve support for cross-folder lists

labkey-nicka:
#### Rationale
This PR addresses a few inconsistencies that came about with #3001 with regards to cross-folder support for lists. The main thing is that when creating or importing a list definition we will now only resolve against lists within the target folder. This restores the behavior where a user could define a list with the same name in a subfolder.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3001

#### Changes
* Lists: create/import list resolves only lists current folder
* ListService: telescope getLists() signatures
* ListWebPart: only display lists from current folder
* ListManager: display `CurrentPlusProjectAndShared` in the UI

### DIFF
--- a/api/src/org/labkey/api/exp/list/ListService.java
+++ b/api/src/org/labkey/api/exp/list/ListService.java
@@ -42,14 +42,16 @@ public interface ListService
     }
 
     Map<String, ListDefinition> getLists(Container container);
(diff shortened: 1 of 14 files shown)


labkey-susanh (approved):
Hard to keep track of all the booleans here, but this seems ok to me.

PR #3016, opened 2022-02-04
lineage sql

labkey-matthewb:
#### Rationale
Implement new LabKey SQL functionality "IN expDescendantsOf ()" and "IN expAncestorsOf ()" to efficiently support lineage filters required the new sample finder feature.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3016
* https://github.com/LabKey/commonAssays/pull/428

#### Changes
*


api/src/org/labkey/api/data/BaseColumnInfo.java
```diff
@@ -1049,7 +1049,12 @@ public void loadFromXml(ColumnType xmlCol, boolean merge)
         checkLocked();
 
         if (xmlCol.isSetConceptURI())
-            setConceptURI(xmlCol.getConceptURI());
+        {
+            String conceptURI = xmlCol.getConceptURI();
+            // User can not set this concepturi, it only applies to exp.object.objectid
+            if (!StringUtils.equalsIgnoreCase(conceptURI,BuiltInColumnTypes.EXPOBJECTID_CONCEPT_URI))
+                setConceptURI(xmlCol.getConceptURI());
```
XingY: can use conceptURI directly here

api/src/org/labkey/api/exp/api/ExpLineageOptions.java
```diff
@@ -114,6 +115,25 @@ public void setUseObjectIds(boolean useObjectIds)
         _useObjectIds = useObjectIds;
     }
 
+    /**
+     *  Use setOnlySelectObjectId() if you only care about the set of objectids returned form the lineage query.
+     *
+     * NOTE: that also means that there is no implicit container filter on the result.  The
```
XingY, line 121: The lack of container filtering should make supporting sample finder in a subfolder setup easier.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -794,22 +789,62 @@ private void addSampleTypeColumns(ExpSampleType st, List<FieldKey> visibleColumn
...
 
         SQLFragment sql = new SQLFragment();
+        sql.appendComment("<ExpMaterialTableImpl.getFromSQL()>", getSqlDialect());
         sql.append("(SELECT ");
         String comma = "";
         for (String dataCol : dataCols)
         {
-            sql.append(comma).append("m.").append(dataCol);
-            comma = ", ";
+            // don't need to generate SQL for columns that aren't selected
```
XingY, line 843: Is it true that a column is needed only if it's in the select? A query might not select RowId as a returned result, but uses rowId as a join condition to another table, or use it in the where filter. 

Ideally, LabKey would able to determine that selectedColumns should also include those dependency columns. But I have seen in testing that the absent of "RowId" is causing faceted filtering on any of those inventory columns added to samples grid to fail (Freeze/Thaw Count, StorageStatus, etc). Minimally "RowId" should probably be added to the wrappedFieldKeys set. But conservatively, we might want to continue to select all columns

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -794,22 +789,62 @@ private void addSampleTypeColumns(ExpSampleType st, List<FieldKey> visibleColumn
...
     public SQLFragment getFromSQL(String alias)
+    {
+        return getFromSQL(alias, null);
+    }
+
+
+    // These are mostly fields that are wrapped by fields with different names (see createColumn())
+    // we could handle each case separately, but this is easier
+    static final Set<FieldKey> wrappedFieldKeys = Set.of(
+            new FieldKey(null, "objectid"),
```
XingY, line 804: add RowId?

query/src/org/labkey/query/sql/QInLineage.java
```diff
@@ -0,0 +1,132 @@
...
+    {
+        return (_in ? " IN " : " NOT IN ") + (_parents ? "EXPANCESTORSOF " : "EXPDESCENDANTSOF " );
+    }
+
+    @Override
+    public void appendSql(SqlBuilder builder, Query query)
+    {
+        SQLTableInfo sqlti = new SQLTableInfo(query.getSchema().getDbSchema(), "_");
+        var children = childList();
+        var LHS = ((QExpr) getFirstChild());
```
XingY, line 54: LHS short for? (I'm guess left hand side but I might be way off... )

query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -376,10 +440,57 @@ private QExpr bind(QExpr expr, Map<FieldKey, ? extends ColumnInfo> columnMap)
...
+
+    private static final FieldKey expObjectIdFieldKey = new FieldKey(null, BuiltInColumnTypes.EXPOBJECTID_CONCEPT_URI);
+
+
+    /* It would be nice to put with class ChildOfMethod and class ParentOfMethod, but our build dependencies mean this goes here for now */
+    private static class InLineageOfClause extends WhereClause
+    {
+        static InLineageOfClause create(boolean parents, FieldKey fk, String value)
+        {
+            if (null != fk.getParent() || !StringUtils.startsWithIgnoreCase(value, "SELECT "))
```
XingY, line 458: Can consider allowing (Select ...). I see that the parentheses are added below, but doesn't hurt to double parentheses

XingY (changes requested):
Columns selection in ExpMaterialTableImpl needs some changes.

query/src/org/labkey/query/sql/QInLineage.java
labkey-matthewb: yes, left-hand side.

query/src/org/labkey/query/QueryServiceImpl.java
labkey-matthewb: I tried to make this work, but I think it's tricky.  I can check that the value starts and ends with parens.  I can't really check that they are matched parens though.

`(select col from table) AND (1=1)`

I could probably do something, but I don't want to think about it right now.

XingY: agree this is nice to have, but not important and can definitely do without for now.

XingY (approved):

PR #3017, opened 2022-02-04
Issue 44711: Field editor update to show confirm modal on change of data type for saved field

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44711

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/728
* https://github.com/LabKey/platform/pull/3017
* https://github.com/LabKey/biologics/pull/1147
* https://github.com/LabKey/sampleManagement/pull/829

#### Changes
* DomainPropertyImpl update to prevent file data type (i.e. Attachment and FileLink) from being converted


experiment/src/org/labkey/experiment/api/property/DomainPropertyImpl.java
```diff
@@ -777,6 +777,10 @@ else if (_pdOld != null)
                 }
             }
 
+            // Issue 44711: Prevent attachment and file field types from being converted to a different type
+            if ("file".equalsIgnoreCase(oldType.getInputType()) && oldType != newType)
```
labkey-jeckels:
```suggestion
            if (PropertyType.FILE_LINK.getInputType().equalsIgnoreCase(oldType.getInputType()) && oldType != newType)

```

I haven't tested this yet but it would remove the magic text value.

labkey-jeckels (approved):

experiment/src/org/labkey/experiment/api/property/DomainPropertyImpl.java
cnathe: Thank you. I meant to go back and update that magic text value to use the const....but forgot. Done.

PR #3019, opened 2022-02-05
44706 : Auditing for Participant Groups

labkey-klum:
#### Rationale
This change adds a new audit provider for participant groups. We will track CRUD operations for participant groups and categories and display relevant changes in the audit record:

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44706

#### Changes
- Implement and register a new `ParticipantGroupAuditProvider` 
- Add audit events when groups and categories are created, modified and deleted
- Revamped the group and category caches to be a single cache. Categories and groups can be retrieved by id, label etc. instead of only getting a `Collection` of objects. There was also some strange behavior where groups assigned to a category were computed outside of the cache, removed this side effecting. Unfortunately the objects returned from the cache can be mutated so we need to be aware of this. Possibly down the road we can fix this behavior.


study/src/org/labkey/study/audit/ParticipantGroupAuditProvider.java
```diff
@@ -0,0 +1,295 @@
...
+    @Override
+    public String getDescription()
+    {
+        return "Events related to the creation and modification of Participant Groups and Categories.";
+    }
+
+    @Override
+    public <K extends AuditTypeEvent> Class<K> getEventClass()
+    {
+        return null;
```
labkey-adam: Are you sure about returning `null`? Users of this method don't appear to be expecting `null`.

labkey-adam (changes requested):

study/src/org/labkey/study/audit/ParticipantGroupAuditProvider.java
labkey-klum: good point : [b0bfebd](https://github.com/LabKey/platform/pull/3019/commits/b0bfebd97de2abbfcf383128c8e887b799f6fa2d)

labkey-adam (approved):

PR #3029, opened 2022-02-09
Let Tomcat close websockets when sessions expire

labkey-alan:
#### Rationale
Tomcat automatically closes Websocket sessions for us when user sessions expire, this PR removes code that closes websocket connections during logout, in favor of letting Tomcat handle session close for us.

#### Related Pull Requests
* https://github.com/LabKey/labbook/pull/138
* https://github.com/LabKey/biologics/pull/1148

#### Changes
* NotificationService/Impl: Remove closeServerEvents
* NotificationEndpoint: Remove isSameHttpSession, doClose, close


api/src/org/labkey/api/security/AuthenticationManager.java
```diff
@@ -125,6 +126,7 @@
     private static final Logger _log = LogManager.getLogger(AuthenticationManager.class);
     // All registered authentication providers (DbLogin, LDAP, SSO, etc.)
     private static final List<AuthenticationProvider> _allProviders = new CopyOnWriteArrayList<>();
+    private static final List<LogoutHandler> _logoutHandlers = new ArrayList<>();
```
labkey-adam:
```suggestion
    private static final List<LogoutHandler> _logoutHandlers = new CopyOnWriteArrayList<>();

```
Using an `ArrayList` is not thread-safe

labkey-jeckels:
I don't know the full scenario driving this change, but consider piggybacking on the HttpSessionListener that I added last week:

https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=44761

https://github.com/LabKey/platform/pull/2999

It'll get invoked anytime a session ends, whether the user has logged out directly or the session has timed out from old age.

labkey-alan:
@labkey-jeckels yeah this looks good, I'll move the LogoutHandler registry to UserManager instead of AuthenticationManager, and I'll execute the handleLogout methods inside sessionDestroyed.

api/src/org/labkey/api/security/AuthenticationManager.java
labkey-alan: Ended up moving this code to UserManager, where I also updated it to use a CopyOnWriteArrayList

labkey-jeckels:
@labkey-alan cool, let me know if you hit snags. It's still pretty fresh in my mind.

labkey-alan:
@labkey-jeckels it went smoothly, was able to integrate it without issues, and it seems to be working according to my testing.

api/src/org/labkey/api/security/UserManager.java
```diff
@@ -437,11 +444,19 @@ public void sessionCreated(HttpSessionEvent event)
         @Override
         public void sessionDestroyed(HttpSessionEvent event)
         {
+            HttpSession session = event.getSession();
+            Integer userId = (Integer) event.getSession().getAttribute(SecurityManager.USER_ID_KEY);
             // Issue 44761 - track session duration for authenticated users
             if (event.getSession().getAttribute(SecurityManager.USER_ID_KEY) != null)
```
labkey-adam:
```suggestion
            if (userId) != null)
```

labkey-adam (approved):
Thanks for taking the suggestions. And for the rename (`SessionEventHandler`)! One more suggestion to use `userId` in the null check (since you have it). No need for further review.

labkey-adam:
Sorry, one more request to update the comment with the new names, just to avoid further confusion

labkey-alan:
@labkey-adam @labkey-jeckels it appears that Tomcat is very thoughtful and closes websockets connections for us automatically when sessions expire. This means by waiting until the session expiration event has been intercepted it is already too late, and we don't need to try to kill the websockets. This also means that our Websocket client code always renders a "Session Expired" modal instead of a "Logged Out" modal, because we're not sending our custom event that has the close reason set to "org.labkey.api.security.AuthNotify#SessionLogOut". Is this problematic?

labkey-alan:
In particular this is the line that will _always_ execute when a user logs out:

https://github.com/LabKey/platform/blob/develop/api/webapp/clientapi/dom/WebSocket.js#L93

Should we revert to my original implementation, where we close the sockets ourselves before Tomcat has a chance to, or should we just roll with this behavior, and allow Tomcat to handle this for us. My inclination is to continue to let Tomcat do this for us, and just remove all the session handler code.

labkey-jeckels:
Letting Tomcat handle this seems reasonable to me. We could always adjust the wording if we're worried people will be confused about timeouts vs explicit logouts, but it doesn't seem likely to me.

Based on one of our mothership metrics, users very seldom log out in most cases and just let the session timeout on its own. It's possible that's LKB or other areas have different usage patterns, of course.

labkey-adam:
Agreed that we let Tomcat do its thing here

labkey-nicka (approved):

labkey-adam (approved):

PR #3030, opened 2022-02-10
Improve error message when a DB connection can't be created

labkey-jeckels:
#### Rationale
toString() on DataSource is pretty useless, meaning we get a message like:

Can't create a database connection to org.apache.tomcat.dbcp.dbcp2.BasicDataSource@269d3b6b

#### Changes
* Include DataSource name and URL instead of useless toString()

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -1045,7 +1045,13 @@ public class DbScope
         }
         catch (SQLException e)
         {
-            throw new ConfigurationException("Can't create a database connection to " + getDataSource().toString(), e);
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-stuartm @labkey-matthewb @labkey-adam 

This change will show both the data source name like "jdbc/labkeyDataSource" and the URL, like "jdbc:postgresql://localhost/test"

The data source name shouldn't be especially sensitive, but I'd appreciate your input on whether we should omit the URL part. It'll be useful for admins but may leak more than we want about the deployment infrastructure to users, including potentially guests.

labkey-adam (approved):

PR #3031, opened 2022-02-10
Item 9998: Update permissions for StorageEditor and StorageDesigner roles

labkey-susanh:
#### Rationale
The StorageEditor and StorageDesigner roles were created recently to be able to remove some capabilities from admins.  These roles did not provide any read access, however, because in some cases those who manage samples in storage are not meant to be able to read data about all types of entities.  In particular, assay, data class (registry or source), media, and ELN data may contain IP that not everyone is meant to be able to see.  We introduce new specific Read permissions for these types of data that can be granted separately from a general Read permission (needed for reading just the basic data needed for interacting with LKS and the applications on top of it).

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/730
* https://github.com/LabKey/inventory/pull/376
* https://github.com/LabKey/biologics/pull/1150
* https://github.com/LabKey/labkey-api-js/pull/123
* https://github.com/LabKey/labbook/pull/142
* https://github.com/LabKey/recipe/pull/58
* https://github.com/LabKey/sampleManagement/pull/832

#### Changes
* Add new permission classes for Assays, Data classes, Notebooks, and Media
* Update interface for `ProductMenuProvider.getSections` so we can remove sections based on permissions


experiment/src/org/labkey/experiment/api/ExpTableImpl.java
```diff
@@ -44,6 +44,7 @@
 import org.labkey.api.query.UserSchema;
 import org.labkey.api.query.column.BuiltInColumnTypes;
 import org.labkey.api.security.UserPrincipal;
+import org.labkey.api.security.permissions.DataClassReadPermission;
```
Sigmonia: no longer used?

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -77,12 +77,16 @@
 import org.labkey.api.security.permissions.AbstractActionPermissionTest;
 import org.labkey.api.security.permissions.AdminOperationsPermission;
 import org.labkey.api.security.permissions.AdminPermission;
+import org.labkey.api.security.permissions.AssayReadPermission;
```
Sigmonia: Looks like these aren't being used

Sigmonia (approved):

PR #3033, opened 2022-02-10
44704: Data Views checkbox for "queries" should read "grid views"

labkey-klum:
#### Rationale
The data views option to show : queries only actually apply to gird views. To help clarify that, we are renaming the label to
- grid views
- grid views (inherited)

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44704

#### Changes
- Rename checkbox labels
- Add a tooltip for the data view type description, improved the inherited grid view description to help users understand better.
- Discovered that chrome was truncating Ext4 tooltips, updated ext-patches to apply the fix that was previously limited to `firefox` and `safari`

### DIFF
--- a/query/webapp/dataview/DataViewsPanel.js
+++ b/query/webapp/dataview/DataViewsPanel.js
@@ -1143,15 +1143,22 @@ Ext4.define('LABKEY.ext4.DataViewsPanel', {
             sizeItems = [];
 
         if (Ext4.isArray(data.types)) {
-            for (var i=0; i < data.types.length; i++) {
-                cbItems.push({
-                    boxLabel: data.types[i].name,
-                    name: data.types[i].name,
-                    checked: data.types[i].visible,
+            Ext4.each(data.types, function(type) {
+
+                var item = {
+                    boxLabel: type.name,
+                    name: type.name,
+                    checked: type.visible,
                     width: 150,
                     uncheckedValue : '0'
-                });
-            }
+                };
+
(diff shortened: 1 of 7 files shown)


query/webapp/dataview/DataViewsPanel.js
```diff
@@ -1143,15 +1143,22 @@ Ext4.define('LABKEY.ext4.DataViewsPanel', {
...
+                    checked: type.visible,
                     width: 150,
                     uncheckedValue : '0'
-                });
-            }
+                };
+
+                // include the description as a tooltip on the element box label
+                if (type.description.length > 0)
+                    item.afterBoxLabelTpl = '<a href="#" data-qtip="' + type.description + '"><span class="labkey-help-pop-up">?</span></a>';
```
cnathe, line 1158: The new / updated description for the "grid views (inherited)" checkbox is good. the ones for "datasets" and "reports" aren't very informative. It might be worth removing them or updating them at some point.

cnathe (approved):
Looks good. I pulled down the changes locally to verify that the Chrome hover tooltip truncate issue is resolved. Glad that change is a general one instead of having to track down individual cases.

query/webapp/dataview/DataViewsPanel.js
labkey-klum: good idea, I'll just remove them as part of this PR, description is not used anywhere else

PR #3036, opened 2022-02-11
Issue 44835: LKSM: Unable to use MaterialInputs/SampleTypeName in naming pattern

XingY:
#### Rationale
When support for lineage lookup in name expression was added, expression containing 2 parts can be interpreted as  both lookup or regular parent input: Inputs/Fields, Inputs/SampleType1. However, the change caused regression in recognizing Inputs/SampleType1 as parent input and rejected it as a lookup, because SampleType1 is not a sample property field. 

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/data/NameGenerator.java
+++ b/api/src/org/labkey/api/data/NameGenerator.java
@@ -24,6 +24,7 @@ import org.junit.Assert;
 import org.junit.Test;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
 import org.labkey.api.collections.CaseInsensitiveHashSet;
+import org.labkey.api.exp.Identifiable;
 import org.labkey.api.exp.PropertyType;
 import org.labkey.api.exp.api.ExpData;
 import org.labkey.api.exp.api.ExpDataClass;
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -573,11 +583,14 @@ public static boolean isLineageInput(Object token, @Nullable Map<String, String>
                 || aliasesInsensitive.containsKey(sTok);
     }
 
-    private Object getLineageLookupTokenPreview(FieldKey fkTok, String inputPrefix, @Nullable String inputDataType, String lookupField, User user)
+    /**
+     * @param inputDataTypeOrLookupField For syntax with 2 parts, the sencond part can be either a lookup field or a data type: Inputs/lookup, MaterialInputs/lookup, DataInputs/lookup, MaterialInputs/SampleType1
```
labkey-susanh: I'm confused by this comment.  Do we really support skipping over the sample type name entirely when using `MaterialInputs` (or `DataInputs`)? How do we find the lookup field if we don't know the sample type to look in?  I know we want to support `ImportAlias/lookup` as a substitute for `MaterialInputs/SampleType1/lookup`.  Is that what this is covering?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -622,6 +635,15 @@ private Object getLineageLookupTokenPreview(FieldKey fkTok, String inputPrefix,
             return null;
         }
 
+        // Inputs/SampleType1
+        if (inputDataType == null)
+        {
+            List<String> dataTypeNames = dataTypes.stream().map(Identifiable::getName).toList();
+
+            if (dataTypeNames.contains(inputDataTypeOrLookupField) || inputDataTypeOrLookupField.equals(currentDataType))
+                return isMaterial && isData ? SubstitutionValue.Inputs.getPreviewValue() : (isMaterial ? SubstitutionValue.MaterialInputs.getPreviewValue() : SubstitutionValue.DataInputs.getPreviewValue());
```
labkey-susanh: I think, at the very least, some parentheses are warranted here since the first part of this statement seems to be impossible (`isMaterial` *and* `isData`?)

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -1503,6 +1503,8 @@ public ExpData findExpData(Container c, User user,
         // First attempt to resolve by name.
         try
         {
+            // TODO, rowId is not found for samples newly created in the same import.
+            // This is causing name patterns containing lineage lookup to fail to generate the correct names if the child samples and their parents are created from the same import file
```
labkey-susanh, line 1507: Do we have any recourse here?  Do we need to document this or file an issue?

api/src/org/labkey/api/data/NameGenerator.java
XingY: We support ImportAlias/lookup, but we do also support skipping sample type name...  The assumption here is pooling is not that common (?), and that user might be using a field that's specific to one sample type. Just a way to provide shorthand in some scenarios.

labkey-susanh: How do we find the sample type to get to the actual lookup field if they don't tell us?  (And do we have documentation for this?)

api/src/org/labkey/api/data/NameGenerator.java
XingY: The scenario is when "Inputs" instead of "MaterialInputs" or "DataInputs" is used. Updated variable name to better match the scenario.

api/src/org/labkey/api/data/NameGenerator.java
XingY: The current implementation just iterate through all sample types until it finds a matching field. So this is not recommended to be used on a general field, such as "Name". Pinged docs team to clarify this.

labkey-susanh: Do we know of a use case where this shortcut is needed/wanted?  It seems like it would impact import performance at least a little.  Have we always supported that?  Maybe we want to deemphasize this option in the docs.

XingY: I did some refactoring to determine if a 2 parts lineage syntax refers to a sampletype or a sample field. That logic is only checked once per import, not one per row, so shouldn't impact performance much.

Agree that we should deemphasize this option in the docs.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: issue created https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44855

labkey-susanh (approved):

PR #3037, opened 2022-02-11
Help user understand why a clipboard copy failed

labkey-jeckels:
#### Rationale
navigator.clipboard APIs require HTTPS (or localhost, which makes this harder to test). While all servers should be using HTTPS, we can do better by giving feedback to the user

#### Changes
* Detect failure and show helpful text

### DIFF
--- a/pipeline/src/org/labkey/pipeline/status/details.jsp
+++ b/pipeline/src/org/labkey/pipeline/status/details.jsp
@@ -343,7 +343,11 @@
                     },
                     function () {
                         e.target.innerHTML = "Copy failed. Ensure the browser's security policy allows clipboard APIs and the page is using HTTPS.";
-                    });
+                    }).catch(function(error) {
+                        console.log(error);
+                        e.target.innerText = "Copy failed: " + error;
+                    }
+            );
         }
     }
 
--- a/pipeline/src/org/labkey/pipeline/status/details.jsp
+++ b/pipeline/src/org/labkey/pipeline/status/details.jsp
@@ -333,13 +333,17 @@
 
     function copyLogText(e) {
         if (logDataEl) {
-            navigator.clipboard.writeText(logDataEl.innerText).then(function () {
-                let orig = e.target.innerHTML;
-                e.target.innerHTML = "Copied!";
-                setTimeout(function () {
-                    e.target.innerHTML = orig;
-                }, 2000);
-            });
+            navigator.clipboard.writeText(logDataEl.innerText).then(
+                    function () {
+                        let orig = e.target.innerHTML;
+                        e.target.innerHTML = "Copied!";
+                        setTimeout(function () {
+                            e.target.innerHTML = orig;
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
I checked and the other places we're doing clipboard functionality we're not using navigator.clipboard, at least not directly. In some places we're doing the deprecated document.execCommand('copy') and in other places we're using clipboard.js. I was tempted to consolidate but ultimately decided to not venture out that far

labkey-martyp (approved):

pipeline/src/org/labkey/pipeline/status/details.jsp
```diff
@@ -333,13 +333,17 @@
...
+                    function () {
+                        let orig = e.target.innerHTML;
+                        e.target.innerHTML = "Copied!";
+                        setTimeout(function () {
+                            e.target.innerHTML = orig;
+                        }, 2000);
+                    },
+                    function () {
+                        e.target.innerHTML = "Copy failed. Ensure the browser's security policy allows clipboard APIs and the page is using HTTPS.";
+                    });
```
labkey-susanh: There should probably also be a `.catch` block attached after the then. I'm not sure if the same messaging would still be appropriate there (Maybe just "Copy failed. Please check the logs."), but at least logging the error would be helpful. 
```
navigator.clipboard.writeText(logDataEl.innerText).then( 
...
).catch(error -> {
   console.error(error)
   e.target.innerHTML = "Copy failed. So sorry, but have a nice day.";
});
```

labkey-susanh (changes requested):
One request, but this is certainly an improvement.

pipeline/src/org/labkey/pipeline/status/details.jsp
```diff
@@ -333,13 +333,21 @@
...
+                        e.target.innerHTML = "Copied!";
+                        setTimeout(function () {
+                            e.target.innerHTML = orig;
+                        }, 2000);
+                    },
+                    function () {
+                        e.target.innerHTML = "Copy failed. Ensure the browser's security policy allows clipboard APIs and the page is using HTTPS.";
+                    }).catch(function(error) {
+                        console.log(error);
+                        e.target.innerText = "Copy failed: " + error;
```
labkey-susanh, line 348: It seems likely this error message is going to be something fairly incomprehensible to the average user, but it should be a rare occurrence, so I leave it to you to decide whether to remove it or not since the error will get logged.

labkey-susanh (approved):

pipeline/src/org/labkey/pipeline/status/details.jsp
labkey-jeckels: Since it'll be logged to a place that a user is even less likely to find and relay to us, I'm OK keeping it as-is. I suppose I log it as a client exception to mothership. But since it seems unlikely to happen, I'm going to merge.

PR #3046, opened 2022-02-14
URLDisplayColumn updates to supported properties for finer grained control over click and popup behavior

cnathe:
#### Rationale
The URLDisplayColumn factory is very useful for showing thumbnails and hover tooltip displays for images and URL paths in the LK grid. This PR adds a few more finer grained options that can be set via the XML `<properties>` for further use-cases of how the click behavior and tooltip/hover display.

Here is an example usage which allows for an image path value to be used as a thumbnail display in the grid that will open in a new tab on click but not render a popup image on hover:
```
<tables xmlns="http://labkey.org/data/xml">
  <table tableName="ListWithAttachment" tableDbType="NOT_IN_DB">
    <columns>
      <column columnName="ImagePath">
        <displayColumnFactory>
          <className>org.labkey.api.data.URLDisplayColumn$Factory</className>
          <properties>
            <property name="thumbnailImageUrl">${ImagePath}</property>
            <property name="thumbnailImageWidth">200px</property>
            <property name="urlTarget">_blank</property>
            <property name="renderPopupImage">false</property>
            <property name="useThumbnailImageUrlOnClick">false</property>
          </properties>
        </displayColumnFactory>
      </column>
    </columns>
  </table>
</tables>
```

#### Changes
* Add URLDisplayColumn factory property support for useThumbnailImageUrlOnClick, renderPopupImage, and urlTarget


labkey-jeckels (approved):
One item to fix but no need for additional review unless you have questions

PR #3049, opened 2022-02-15
Issue 24827: Radeox logs StackOverflowError with square bracket wiki syntax

labkey-jeckels:
#### Rationale
No need to log a giant stack trace the stderr for a non-fatal situation

#### Changes
* Update Radeox to tone down logging
* Trim a couple of unneeded dependencies
* Add JUnit dependency required for compilation
* Fix accumulated compilation errors and warnings
* Bump version to 1.0.3, already uploaded to Artifactory
* Delete .class files


labkey-jeckels:
@labkey-susanh this is an improvement over the current behavior and what's currently checked into Git for Radeox. I know there's more we'd like to do, but I'd say this is a good pause point. See what you think.

wiki/radeox/Changes.txt
```diff
@@ -1,3 +1,7 @@
+1.0.3 2022-02-11
+
+- Stop logging StackOverflowErrors to std err and trim up some other System.out/System.err usage (LabKey issue 24827)
+
```
labkey-susanh, line 4: Good to make a new release at least every 18 years.

wiki/radeox/build.xml
```diff
@@ -27,7 +26,7 @@
      <property name="conf" value="./conf"/>
      <property name="jar" value="./lib"/>
      <property name="version" value="1.0"/>
-     <property name="status" value="-BETA-2"/>
+     <property name="status" value=".3"/>
```
labkey-susanh: Not our code, of course, but I think 'status' is meant to capture something like "alpha", "beta", etc, so I would probably change `version` to `1.0.3` and get rid of the `status` property (or set it to the empty string).

labkey-susanh (approved):
Definitely an improvement.  Thanks for doing this.

wiki/radeox/build.xml
labkey-jeckels: I thought about excising it, but ended up co-opting it instead. I pushed a commit that removes it completely

PR #3054, opened 2022-02-16
Issue 44869: Conditional format export can produce an invalid XML fragment

labkey-adam:
#### Rationale
Better to produce a valid XML document and log a warning if we encounter a bogus conditional format

I'd rather stick the warning in the pipeline log, but that would have been a lot of wiring for a rare scenario

### DIFF
--- a/api/src/org/labkey/api/data/ConditionalFormat.java
+++ b/api/src/org/labkey/api/data/ConditionalFormat.java
@@ -256,7 +256,7 @@ public class ConditionalFormat extends GWTConditionalFormat
         return result;
     }
 
-    public static void convertToXML(List<? extends GWTConditionalFormat> formats, ColumnType xmlColumn)
+    public static void convertToXML(List<? extends GWTConditionalFormat> formats, ColumnType xmlColumn, String tableName)
     {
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/data/ConditionalFormat.java
```diff
@@ -266,7 +266,10 @@ public static void convertToXML(List<? extends GWTConditionalFormat> formats, Co
         if (!formats.isEmpty())
         {
             ConditionalFormatsType xmlFormats = xmlColumn.addNewConditionalFormats();
-            addToXML(formats, xmlFormats);
+            if (!addToXML(formats, xmlFormats))
+            {
+                LOG.warn("One or more invalid conditional formats were discovered on column \"" + xmlColumn.getColumnName() + "\"");
```
labkey-jeckels: Can we pipe in the table name too? Looks like there are only two callers, both of which have easy access to it.

labkey-jeckels (approved):

api/src/org/labkey/api/data/ConditionalFormat.java
labkey-adam: Oh I suppose...

PR #3061, opened 2022-02-16
Issue 44873: Deleting Pipeline Jobs via "Select All" fails to delete some jobs

labkey-jeckels:
#### Rationale
If your job selection state goes beyond the current page of the grid, only the ones from the current page get deleted. It's because we're prematurely clearing the selection state prior to the delete itself.

#### Changes
* Don't clear the state until we're actually doing the delete
* Show the user the number of jobs to be deleted on the confirmation page

### DIFF
--- a/pipeline/src/org/labkey/pipeline/status/StatusController.java
+++ b/pipeline/src/org/labkey/pipeline/status/StatusController.java
@@ -743,7 +743,9 @@ public class StatusController extends SpringActionController
         @Override
         public void validateCommand(ConfirmDeleteStatusForm form, Errors errors)
         {
-            Set<String> runs = DataRegionSelection.getSelected(getViewContext(), true);
+            // Don't clear the state yet because we're just validating at this point. We'll clear it as part of the
+            // delete itself. See issue 44873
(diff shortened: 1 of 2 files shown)


pipeline/src/org/labkey/pipeline/status/StatusController.java
```diff
@@ -743,7 +743,9 @@ public void setDeleteRuns(boolean deleteRuns)
         @Override
         public void validateCommand(ConfirmDeleteStatusForm form, Errors errors)
         {
-            Set<String> runs = DataRegionSelection.getSelected(getViewContext(), true);
+            // Don't clear the state yet because we're just validating at this point. We'll clear it as part of the
+            // delete itself. See issue 44873
```
labkey-susanh, line 747: I assume clearing as part of the delete is already happening?

labkey-susanh (approved):

pipeline/src/org/labkey/pipeline/status/StatusController.java
labkey-jeckels: Yes. https://github.com/LabKey/platform/blob/develop/pipeline/src/org/labkey/pipeline/status/StatusController.java#L773

PR #3063, opened 2022-02-16

labkey-jeckels:
#### Rationale
I had to back out my original attempt to fix this as it caused intermittent failures on TeamCity, likely due to cancelled queries when the browser navigated away during render. I'll revisit how the connection gets reused. But this part of the change, which prevents the need for the separate aggregate query when we know the result set is complete is safe and makes it less likely we'll end up using two connections concurrently.

#### Changes
* Don't execute a separate aggregate query to get the count when we already know the number of rows
* Minor code cleanup


labkey-matthewb (approved):
Approve!

PR #3064, opened 2022-02-17

labkey-jeckels:
#### Rationale
Using multiple connections concurrently is bad because it can lead to deadlocks trying to get DB connections from the pool during heavy load. Better to use a single connection when possible, including for the main grid and its aggregate query

#### Changes
* Check if we have a Connection we can reuse from the primary ResultSet that's driving the grid's data


labkey-jeckels:
I found a simpler approach compared to my previous attempt. I've queued multiple instances of BVT-A for both Postgres and SQLServer to make sure there's not an intermittent failure lurking.

labkey-adam (approved):

PR #3066, opened 2022-02-18
Improve auto link to study for samples

labkey-klum:
#### Rationale
In 21.7 we introduced specific data types to samples to help with linking data to studies:
- `subject/participant`
- `visit date`
- `visit id`
These types are understood and used to resolve subject/timepoint fields when using the UI version of the link to study workflow. However, when the sample type is set to automatically link new data to a study we were reverting back to a set of hard-coded field names to resolve fields.

Additionally, automatic linking was not wired up to handle sample derivation either when using the UI or in the bulk import case. We also needed to unify the behavior between auto copying and linking via the UI to support this scenario:
- Subject/timepoint fields may sometimes be introduced from a parent sample
  - If the data types aren't found on the derived sample and the data types from a parent are pulled into the default view using the lineage `inputs` columns, then we will attempt to resolve those values.

#### Changes
- Rename and move `PublishResultsQueryView.ExtraColFieldKeys` to `StudyPublishService.LinkToStudyKeys`
- Fix for this issue: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44417 (change in `AssayTableMetadata`)
- Push functionality to handle sample data linking into `StudyPublshService` for both the derivation and non-derivation cases, and wire up the calls to handle auto linking in the data iterator as well as the controller action (for the UI derivation case). The derivation case requires creating a `QueryView` to get the columns in the default view as well as needing to use a `DisplayColumn` to resolve lineage values (`LineageDisplayColumn`)


labkey-adam (approved):
Looks good to me. Would be good for Matt to take a look as well.

PR #3077, opened 2022-02-21
Move more specimen code into specimen module

labkey-adam:
#### Rationale
Moving legacy code out of the study module

#### Changes
* Move 31 classes from study-api to specimen-main
* Remove all "specimen mode" references from StudyWizard.js, CreateChildStudyAction, CreateChildPipelineJob, etc.


cnathe (approved):

specimen/src/org/labkey/specimen/SpecimenRequestManager.java
```diff
@@ -30,15 +29,19 @@
 import org.labkey.api.query.ValidationException;
 import org.labkey.api.security.User;
 import org.labkey.api.settings.AppProps;
+import org.labkey.api.specimen.SpecimenManagerNew;
+import org.labkey.api.specimen.SpecimenQuerySchema;
+import org.labkey.api.specimen.SpecimenRequestException;
+import org.labkey.api.specimen.SpecimenRequestStatus;
+import org.labkey.api.specimen.SpecimenSchema;
+import org.labkey.api.specimen.Vial;
```
labkey-susanh, line 37: Are these classes going to move to the specimen module as well?

labkey-susanh (approved):

specimen/src/org/labkey/specimen/SpecimenRequestManager.java
labkey-adam: A bunch of these classes were intertwined with each other and ultimately with `SpecimenManager.deleteAllSpecimenData()` (which needs to stay in study, since study still manages this schema). But, it was fairly easy to break the cycle and move another 13 classes... thanks for the nudge!

Goal is to move all the classes from api.specimen into the specimen module, but some are very embedded in study.

PR #3078, opened 2022-02-21
Issue 44902: SelectRows 'includeMetadata' is ignored

RosalineP:
#### Rationale
In LABKEY.Query.selectRows, in the most recent version of the JS api, the configuration parameter 'includeMetadata' was not being respected.

#### Related Pull Requests
* N/A

#### Changes
* Add call to form.isIncludeMetadata() in most recent API version


labkey-klum (approved):
Gotta love a constructor with more than 10 parameters, as the comment on `ApiQueryResponse:88` mentions switch to a builder pattern.

PR #3080, opened 2022-02-22
EHR Sample Manager: lineage columns and sample metadata

labkey-martyp:
#### Rationale
As part of the EHR Sample Manager module, we want to foreign key to source data from a samples grid or detail.  We also want to apply metadata, such as a java customizer to all samples in a container.

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/334

#### Changes
* Add _SingleValueForeignKey to LineageForeignKey which will allow setting foreign keys on sources in a lineage. If the lineage has multiple sources of that type, then either a "not supported message" will be returned for that field (varchar columns) or null. (Note: if you join on a field returning "not supported" due to multiple sources, that will likely not match anything)
* Add a registration mechanism and interface to register sources that should be SingleValueForeignKey
* Similar to studyData.query.xml, add the ability to define a sampleData.query.xml which will apply metadata across all sample types in a container. For EHR Sample Manager module, this is applying a java customizer to all sample types.


api/src/org/labkey/api/data/MultiValuedLookupColumn.java
```diff
@@ -36,23 +36,39 @@
...
         _rightFk = fk;
         _junctionKey = junctionKey;
+        _singleValue = singleValue;
         copyAttributesFrom(display);
         copyURLFrom(display, parentPkColumn.getFieldKey(), null);
         // NOTE: Changing the type to a VARCHAR causes MultiValueRenderContext.get() type conversion to be skipped and we don't want that.
         //setJdbcType(JdbcType.VARCHAR);
     }
 
+    public MultiValuedLookupColumn(ColumnInfo parentPkColumn, ColumnInfo childKey, ColumnInfo junctionKey, ForeignKey fk, ColumnInfo display)
```
labkey-jeckels: To reduce duplication, can we make this constructor call the other one and pass false for `singleValue`

experiment/src/org/labkey/experiment/api/LineageForeignKey.java
```diff
@@ -411,6 +426,62 @@ public boolean isSortable()
         }
     }
 
+    private class _SingleValuedForeignKey extends _MultiValuedForeignKey
```
labkey-jeckels: Some comments here would be helpful. The class names by themselves are confusing - why is SingleValued subclassing MultiValued?

experiment/src/org/labkey/experiment/api/LineageForeignKey.java
```diff
@@ -411,6 +426,62 @@ public boolean isSortable()
...
+
+        @Override
+        protected MultiValuedLookupColumn createMultiValuedLookupColumn(ColumnInfo lookupColumn, ColumnInfo parent, ColumnInfo childKey, ColumnInfo junctionKey, ForeignKey fk)
+        {
+            return new MultiValuedLookupColumn(parent, childKey, junctionKey, fk, lookupColumn, true)
+            {
+                @Override
+                protected SQLFragment getAggregateFunction(SQLFragment sql, String sqlTypeName)
+                {
+                    if ("nvarchar".equalsIgnoreCase(sqlTypeName) || "varchar".equalsIgnoreCase(sqlTypeName))
```
labkey-jeckels: Here's an alternative idea to consider:

Instead of altering the behavior of the existing lookup to the source, what about injecting an additional column that works as a single lookup? We could even inject two columns - one for the count of matches, and another for the value when there's a single match. A DisplayColumn could depend on both the count and single match columns and handle rendering the error when there are multiple matches in a special way (like italicized text). That approach works for non-varchar columns and doesn't rely on a magic value for the prefix.

I think this would be a more robust approach. I don't have strong feelings on whether it's worth adding it now.

api/src/org/labkey/api/exp/query/SamplesSchema.java
```diff
@@ -56,6 +56,7 @@
 public class SamplesSchema extends AbstractExpSchema
 {
     public static final String SCHEMA_NAME = "samples";
+    public static final String SCHEMA_METADATA_NAME = "SampleData";
```
labkey-jeckels, line 59: I know you chose this to follow the "StudyData" precedent. But I've never been a huge fan of that name either. "AllSampleTypes" or something more unique like "*.query.xml" or "~~AllSampleTypes~~" might better signal that it's a magic value.

experiment/src/org/labkey/experiment/api/LineageForeignKey.java
labkey-susanh: As we discussed earlier, I think that continuing to use the lineage lookup column doesn't get around the inherent performance problems that come from the lineage representation, so we're working on a different approach to enabling adding lineage metadata fields into grids so they can be used for sorting and filtering the grids.  As soon as 22.3 features are wrapped up I'll continue working on that implementation plan, which is somewhat akin to what Josh is suggesting.

labkey-martyp:
https://github.com/LabKey/platform/pull/3754

PR #3085, opened 2022-02-23
Item 10056: Sample Finder v1 - Wire up new lineage filters

XingY:
#### Rationale
MSSQL does not support sub select in group concate. Lineage lookup fields are added to sample finder results grid using MVFK, which is not able to render columns that are backed by ExprColumns with sub select clauses. We will exclude those fields from sample finder when the DB used is sql server. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/124
* https://github.com/LabKey/labkey-ui-components/pull/743
* https://github.com/LabKey/sampleManagement/pull/847
* https://github.com/LabKey/biologics/pull/1165

#### Changes
* Add supportGroupConcatSubSelect to getQueryDetails response.

### DIFF
--- a/query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
+++ b/query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
@@ -159,6 +159,8 @@ public class GetQueryDetailsAction extends ReadOnlyApiAction<GetQueryDetailsActi
         if (null == tinfo)
             throw new NotFoundException("Could not find the query '" + form.getQueryName() + "' in the schema '" + form.getSchemaName() + "'!");
 
+        resp.put("supportGroupConcatSubSelect", tinfo.getSqlDialect().supportsGroupConcatSubSelect());
+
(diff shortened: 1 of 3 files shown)


labkey-matthewb (approved):

query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
```diff
@@ -159,6 +159,8 @@ public ApiResponse execute(Form form, BindException errors)
         if (null == tinfo)
             throw new NotFoundException("Could not find the query '" + form.getQueryName() + "' in the schema '" + form.getSchemaName() + "'!");
 
+        resp.put("supportGroupConcatSubSelect", tinfo.getSqlDialect().supportsGroupConcatSubSelect());
```
labkey-susanh, line 162: So the client side is going to have to know which columns correspond to these group-concat sub-select columns?  I hope we'll at least have a utility method that obfuscates this kind of icky check.

PR #3086, opened 2022-02-23
Assert if SQL script is badly named

labkey-adam:
#### Rationale
A recent SQL script was committed with bad version numbers. LabKey should flag this during development instead of silently ignoring such a script.

Note: Initially, this branch should fail to bootstrap on a full build that includes the immport module. It should then switch to success once 21.11 script changes are merged.


labkey-jeckels (approved):

labkey-matthewb:
Do we want to assert that the version is > Module.getSchemaVersion()?

labkey-adam:
I don't think we want to assert in that case. It's not uncommon that someone creates a script and tests it before bumping the module version. Seems reasonable to log a warning, though.

labkey-matthewb (approved):

PR #3087, opened 2022-02-23
Fix names and sizes of some caches

labkey-adam:
#### Rationale
Cache size "Unlimited" is rarely needed. Some cache names are confusing.

[Issue 44829: EhCache warns about unlimited size caches](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44829)

#### Changes
* Fix some names and sizes
* Just use `Container` as a cache key in a couple places

### DIFF
--- a/assay/src/org/labkey/assay/AssayController.java
+++ b/assay/src/org/labkey/assay/AssayController.java
@@ -531,7 +531,6 @@ public class AssayController extends SpringActionController
         @Override
         public void addNavTrail(NavTree root)
         {
-            setHelpTopic("");
             root.addChild("Assay List", new ActionURL(BeginAction.class, getContainer()));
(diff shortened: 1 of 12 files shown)


assay/src/org/labkey/assay/AssayController.java
```diff
@@ -531,6 +531,7 @@ protected void renderCellContents(StringBuilder html, Container c, ActionURL url
         @Override
         public void addNavTrail(NavTree root)
         {
+            setHelpTopic("");
```
labkey-jeckels: Is this intentional?

labkey-jeckels (approved):

assay/src/org/labkey/assay/AssayController.java
labkey-adam: Nope, thanks!

PR #3088, opened 2022-02-23
TableSelector.getDisplayColumnsList() fix

labkey-matthewb:
#### Rationale
TableSelector.getDisplayColumnsList() would sometimes unexpectedly replace a passed in ColumnInfo with a new version (same FieldKey different alias).  This was due to order dependency when building the "columns" map.

#### Related Pull Requests
* 

#### Changes
Explicitly prefer the passed in column to columns resolved via col.getDisplayField()

### DIFF
--- a/api/src/org/labkey/api/data/TableSelector.java
+++ b/api/src/org/labkey/api/data/TableSelector.java
@@ -16,8 +16,8 @@
 
 package org.labkey.api.data;
 
-import org.apache.commons.codec.binary.StringUtils;
 import org.apache.commons.collections4.MultiValuedMap;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
--- a/api/src/org/labkey/api/data/TableSelector.java
+++ b/api/src/org/labkey/api/data/TableSelector.java
@@ -156,8 +156,9 @@ public class TableSelector extends SqlExecutingSelector<TableSelector.TableSqlFa
         for (ColumnInfo column : arrColumns)
         {
             ColumnInfo prev = columns.put(column.getFieldKey(), column);
+            // NOTE : temporarily disable assert for merge to develop
             // this assert is stricter than necessary, but still probably good hygiene (see following check which is necessary)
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/data/TableSelector.java
```diff
@@ -16,6 +16,7 @@
 
 package org.labkey.api.data;
 
+import org.apache.commons.codec.binary.StringUtils;
```
XingY: import org.apache.commons.lang3.StringUtils  ?

api/src/org/labkey/api/data/TableSelector.java
```diff
@@ -151,20 +152,22 @@ public TableSelector(ColumnInfo column)
...
         Map<FieldKey, ColumnInfo> columns = new LinkedHashMap<>();
-        ColumnInfo existing;
 
         for (ColumnInfo column : arrColumns)
         {
-            existing = columns.get(column.getFieldKey());
-            assert null == existing || existing.getName().equals(column.getName()) : existing.getName() + " != " + column.getName();
-            columns.put(column.getFieldKey(), column);
+            ColumnInfo prev = columns.put(column.getFieldKey(), column);
+            // NOTE : temporarily disable assert for merge to develop
```
XingY, line 159: Is the assert failing in some mockup scenarios, such as TC tests? Is there a plan to re enable this check? Some additional comment on why this needs to disabled and when it can be reenabled should be helpful.

labkey-matthewb:
No known problem with the assert(), I just want to try to fix known EHR test failure, before adding this strict check.

labkey-matthewb:
BTW, I have found another order dependent glitch in the code (order of filter clauses).  Commit coming in a moment.

labkey-matthewb:
PS: the handling of allInvolvedColumn in ensureRequiredColumns()/resolveSortColumns seems convoluted.  This could probably be simplified a bit.  Maybe starting with making "ret" a LinkedHashSet and having resolveFieldKey() not return null for columns that have been previously resolved.

api/src/org/labkey/api/data/TableSelector.java
labkey-matthewb: thank you

XingY (approved):

PR #3092, opened 2022-02-25
Fb content security policy

labkey-matthewb:
#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/security/AuthenticatedResponse.java
+++ b/api/src/org/labkey/api/security/AuthenticatedResponse.java
@@ -19,6 +19,7 @@ package org.labkey.api.security;
 import org.apache.commons.collections4.MultiValuedMap;
 import org.apache.commons.collections4.multimap.AbstractSetValuedMap;
 import org.apache.commons.lang3.StringUtils;
+import org.labkey.api.util.HttpUtil;
 import org.labkey.api.util.PageFlowUtil;
 
 import javax.servlet.http.HttpServletResponse;
@@ -31,6 +32,9 @@ import java.util.Locale;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
+
+import static org.labkey.api.util.HttpUtil.ContentSecurityPolicyEnum.DefaultSrc;
+import static org.labkey.api.util.HttpUtil.Header.ContentSecurityPolicy;
 
 /**
  * User: matthewb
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/security/AuthenticatedResponse.java
```diff
@@ -39,15 +46,164 @@ public AuthenticatedResponse(HttpServletResponse response)
...
+//        // for debugging you can do this
+//        if (_safeWriter == null)
+//            _safeWriter = new SafePrintWriter(super.getWriter());
+//        return _safeWriter;
+    }
+
+    @Override
+    public void addHeader(String name, String value)
+    {
+        if (StringUtils.equalsAnyIgnoreCase("Content-Security-Policy"))
```
labkey-jeckels: Any reason you're not using a simple .equalsIgnoreCase() here? This code isn't using the name parameter at all so it'll never fire the exception

api/src/org/labkey/api/security/AuthenticatedResponse.java
```diff
@@ -39,15 +46,164 @@ public AuthenticatedResponse(HttpServletResponse response)
...
+        StringBuilder content = new StringBuilder();
+        var semi = "";
+        for (var e : _contentSecurityPolicy.asMap().entrySet())
+        {
+            content.append(semi).append(e.getKey());
+            semi = "; ";
+            for (String s : e.getValue())
+                content.append(" ").append(s);
+        }
+        super.setHeader("Content-Security-Policy", content.toString());
```
labkey-jeckels: There's enough instances of this string to extract a constant

study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -4688,6 +4692,8 @@ public ModelAndView getView(CustomizeParticipantViewForm form, boolean reshow, B
             {
                 // issue : 18595 chrome will complain that the script we are executing matches the script sent down in the request
                 getViewContext().getResponse().setHeader("X-XSS-Protection", "0");
```
labkey-jeckels: Can we delete this old header usage too?

internal/src/org/labkey/api/security/AuthFilter.java
```diff
@@ -80,7 +87,6 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
         {
             if (!"ALLOW".equals(AppProps.getInstance().getXFrameOptions()))
                 resp.setHeader("X-Frame-Options", AppProps.getInstance().getXFrameOptions());
-            resp.setHeader("X-XSS-Protection", "1; mode=block");
```
labkey-jeckels, line 83: Do we not need to migrate this one too?

labkey-jeckels (changes requested):

api/src/org/labkey/api/security/AuthenticatedResponse.java
labkey-matthewb: oops.

Still working on it (hence draft). Just thought you'd be curious.

PR #3095, opened 2022-02-25
remove X-XSS-Protection header

labkey-matthewb:
#### Rationale
X-XSS-Protection header is no longer supported by most browsers and no longer recommended.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels (approved):
I hadn't realized that this was basically unsupported these days.

PR #3098, opened 2022-02-26
44544: Sample type dataset definitions should be locked similar to assays

labkey-klum:
#### Rationale
In 21.7 we added the ability to link sample type data to studies. The dataset designer restricts what properties are editable for linked source type datasets (additional key fields, name, etc). However this was only happening for assays and not sample types.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/749

#### Changes
- generalize `DatasetDomainKindProperties` so that it uses : `sourceName`, `sourceUrl`, and `sourceType` to pass down to the client
- update the description for linked source datasets to work for both assays and sample types

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -52,7 +52,7 @@
   },
   "dependencies": {
     "@labkey/api": "1.7.2",
-    "@labkey/components": "2.134.0",
+    "@labkey/components": "2.137.4-fb-issue-44544.0",
     "@labkey/themes": "1.2.1"
   },
   "devDependencies": {
--- a/api/src/org/labkey/api/study/Dataset.java
+++ b/api/src/org/labkey/api/study/Dataset.java
@@ -95,6 +95,12 @@ public interface Dataset extends StudyEntity, StudyCachable<Dataset>
                     }
 
                     @Override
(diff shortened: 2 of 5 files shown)


core/package.json
```diff
@@ -52,7 +52,7 @@
   },
   "dependencies": {
     "@labkey/api": "1.7.2",
-    "@labkey/components": "2.134.0",
+    "@labkey/components": "2.137.4-fb-issue-44544.0",
```
cnathe, line 55: Related to the email thread we had, I assume you aren't seeing the local changes from this alpha package because I am not seeing the related package-lock.json file changes. You can cd into the /server/modules/platform/core dir and then run `npm install --legacy-peer-deps` to get that package-lock.json file updated and then commit that change as well.

cnathe (changes requested):

core/package.json
labkey-klum: done: https://github.com/LabKey/platform/pull/3098/commits/e2a2f060397f18ee7dbc827283415ebc6e976b04

strangely, I was seeing the changes locally even without this change.

cnathe (approved):

PR #3099, opened 2022-02-28
Experiment: improve container filter support

labkey-nicka:
Closed. Reissued against v22.3 with https://github.com/LabKey/platform/pull/3102.

### DIFF
--- a/api/src/org/labkey/api/exp/api/ExperimentService.java
+++ b/api/src/org/labkey/api/exp/api/ExperimentService.java
@@ -702,7 +702,13 @@ public interface ExperimentService extends ExperimentRunTypeSource
 
     List<? extends ExpProtocol> getAllExpProtocols();
 
-    List<? extends ExpProtocol> getExpProtocolsWithParameterValue(@NotNull String parameterURI, @NotNull String parameterValue, @Nullable Container c);
+    List<? extends ExpProtocol> getExpProtocolsWithParameterValue(
+        @NotNull String parameterURI,
+        @NotNull String parameterValue,
+        @Nullable Container c,
+        @Nullable User user,
+        boolean includeProjectAndShared
+    );
 
     void registerRunEditor(ExpRunEditor editor);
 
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/exp/api/ExperimentService.java
```diff
@@ -702,7 +702,13 @@ ExpRun derive(Map<ExpMaterial, String> inputMaterials, Map<ExpData, String> inpu
 
     List<? extends ExpProtocol> getAllExpProtocols();
 
-    List<? extends ExpProtocol> getExpProtocolsWithParameterValue(@NotNull String parameterURI, @NotNull String parameterValue, @Nullable Container c);
+    List<? extends ExpProtocol> getExpProtocolsWithParameterValue(
+        @NotNull String parameterURI,
+        @NotNull String parameterValue,
+        @Nullable Container c,
+        @Nullable User user,
+        boolean includeProjectAndShared
```
labkey-klum, line 710: I could also see passing in a `ContainerFilter` as well

labkey-klum (approved):

labkey-jeckels:
@labkey-nicka this may be what you said you tried without success, but I'd say that we should be passing the desired ContainerFilter into PdLookupForeignKey.create() instead of constructing one. From a quick review of the callers to the two variants of create() this looks like it should be doable but perhaps there are lurking edge cases you've already hit?

labkey-nicka:
> @labkey-nicka this may be what you said you tried without success, but I'd say that we should be passing the desired ContainerFilter into PdLookupForeignKey.create() instead of constructing one. From a quick review of the callers to the two variants of create() this looks like it should be doable but perhaps there are lurking edge cases you've already hit?

Yes, this is precisely what I was talking about. Given the target of 22.3 I'm less inclined to try again right now and will try to consolidate at a later point.

labkey-klum (approved):

PR #3100, opened 2022-02-28
44932: NPE in StudyPublishService on trial instances because the study module is not present

labkey-klum:
#### Rationale
LKSM tests were failing on trial instances because the `Study` module was not present. Additionally an error was discovered during the updating of a sample type definition that might inadvertently set up an auto link target container of : `Root`.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44932

#### Changes
- Don't add the `AutoLinkToStudyDataIterator` if the `StudyPublishService` is not registered
- Don't set the `autoLinkTargetContainer` if the value is null or blank

### DIFF
--- a/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
+++ b/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
@@ -65,6 +65,7 @@ import org.labkey.api.query.UserSchema;
 import org.labkey.api.query.ValidationException;
 import org.labkey.api.search.SearchService;
 import org.labkey.api.security.User;
+import org.labkey.api.study.publish.StudyPublishService;
 import org.labkey.api.util.Pair;
 import org.labkey.experiment.ExpDataIterators;
 import org.labkey.experiment.SampleTypeAuditProvider;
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -191,7 +192,7 @@ public DataIteratorBuilder createImportDIB(User user, Container container, DataI
             if (InventoryService.get() != null)
                 dib = LoggingDataIterator.wrap(InventoryService.get().getPersistStorageItemDataIteratorBuilder(dib, userSchema.getContainer(), userSchema.getUser(), sampleType.getMetricUnit()));
 
-            if (sampleType.getAutoLinkTargetContainer() != null)
+            if (sampleType.getAutoLinkTargetContainer() != null && StudyPublishService.get() != null)
```
labkey-susanh, line 195: I might have thought to check this within the `AutoLinkToStudyDataIterator` so all potential users would not have to do they check themselves, but probably there aren't many (any?) other users, so this works.

labkey-susanh (approved):

PR #3103, opened 2022-03-01
22.3 fb issue 44544

labkey-klum:
#### Rationale
In 21.7 we added the ability to link sample type data to studies. The dataset designer restricts what properties are editable for linked source type datasets (additional key fields, name, etc). However this was only happening for assays and not sample types.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/751
- https://github.com/LabKey/platform/pull/3098 (previous PR)

#### Changes
- generalize `DatasetDomainKindProperties` so that it uses : `sourceName`, `sourceUrl`, and `sourceType` to pass down to the client
- update the description for linked source datasets to work for both assays and sample types

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -8,7 +8,7 @@
       "name": "labkey-core",
       "version": "0.0.0",
       "dependencies": {
-        "@labkey/api": "1.7.2",
+        "@labkey/api": "1.10.0",
         "@labkey/components": "2.138.1",
         "@labkey/themes": "1.2.1"
       },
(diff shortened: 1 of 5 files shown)


core/package-lock.json
```diff
@@ -9,7 +9,7 @@
       "version": "0.0.0",
       "dependencies": {
         "@labkey/api": "1.7.2",
-        "@labkey/components": "2.134.0",
+        "@labkey/components": "2.137.4-fb-issue-44544.0",
```
cnathe, line 12: you are probably already doing this, but this should be v`2.138.1` right?

cnathe (approved):

core/package-lock.json
labkey-klum: yes, just updated package.json and package-lock.json

PR #3104, opened 2022-03-01
Fix OOR display

labkey-matthewb:
#### Rationale
OutOfRangeDisplayColumn display column broke because of a small refactor to DataColumn.getFormattedHtml(). I think this version is more inline with the intended uses of these methods.

#### Related Pull Requests
* 

#### Changes
Move the OOR concatenation to the "format" methods.  I think this will make a change to the "new" JSON response format changing displayValue to no longer have the OOR concatenated, only the formattedValue will have it.

### DIFF
--- a/api/src/org/labkey/api/data/OutOfRangeDisplayColumn.java
+++ b/api/src/org/labkey/api/data/OutOfRangeDisplayColumn.java
@@ -104,6 +104,12 @@ public class OutOfRangeDisplayColumn extends DataColumn
     }
 
     @Override
+    public String getTsvFormattedValue(RenderContext ctx)
+    {
+        return getOORPrefix(ctx) + super.getTsvFormattedValue(ctx);
+    }
+
+    @Override
     public void addQueryColumns(Set<ColumnInfo> columns)
     {
         super.addQueryColumns(columns);
--- a/api/src/org/labkey/api/data/OutOfRangeDisplayColumn.java
+++ b/api/src/org/labkey/api/data/OutOfRangeDisplayColumn.java
@@ -17,6 +17,7 @@
 package org.labkey.api.data;
 
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.QueryService;
 import org.labkey.api.util.HtmlString;
@@ -40,7 +41,6 @@ import java.util.Set;
 public class OutOfRangeDisplayColumn extends DataColumn
 {
     private ColumnInfo _oorIndicatorColumn;
-    private boolean _doneWithSuperclassConstructor = true;
 
     /**
      * Look up the OORIndicator column through QueryService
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/data/OutOfRangeDisplayColumn.java
```diff
@@ -111,9 +98,15 @@ private String getOORPrefix(RenderContext ctx)
     }
 
     @Override
-    public Object getDisplayValue(RenderContext ctx)
+    public @Nullable String getFormattedText(RenderContext ctx)
+    {
+        return getOORPrefix(ctx) + super.getFormattedText(ctx);
```
labkey-jeckels: If super.getFormattedText() is @Nullable, this seems problematic. Shouldn't we check it for null first?

api/src/org/labkey/api/data/OutOfRangeDisplayColumn.java
```diff
@@ -40,7 +41,6 @@
 public class OutOfRangeDisplayColumn extends DataColumn
 {
     private ColumnInfo _oorIndicatorColumn;
-    private boolean _doneWithSuperclassConstructor = true;
```
labkey-jeckels, line 43: Glad to see this go, but I have no idea what the scenario was that introduced it.

labkey-jeckels (approved):
Please see requested change for getFormattedText()

api/src/org/labkey/api/data/OutOfRangeDisplayColumn.java
labkey-matthewb: Thank you

labkey-matthewb: Looks like that was always a potential problem?

labkey-jeckels: Seems like it. Maybe we won't end up calling this method if getValue() returns null, but seems better to check and not return "null"

api/src/org/labkey/api/data/OutOfRangeDisplayColumn.java
```diff
@@ -103,17 +87,22 @@ private String getOORPrefix(RenderContext ctx)
...
             }
         }
-        return result.toString();
+        return "";
+    }
+
+    @Override
+    public @Nullable String getFormattedText(RenderContext ctx)
+    {
+        return getOORPrefix(ctx) + StringUtils.defaultString(super.getFormattedText(ctx));
```
labkey-jeckels: Given the JavaDoc on DisplayColumn.getFormattedText(), this should probably return null instead of an empty string.

labkey-jeckels (approved):
One more comment.

api/src/org/labkey/api/data/OutOfRangeDisplayColumn.java
labkey-matthewb: Given the comment on getFormattedText(), I think that means super.getFormattedText() may return null even if getDisplayValue() is not null!  So either we should explicitly set a format so that doesn't happen or call String.valueOf(getValue()) ourselves.

(seems kinda wrong that getFormattedText() doesn't handle the "default" case of no format.)

labkey-matthewb: Weirdly inconsistent getFormattedHtml() and getTsvFormattedValue() fall through to ConvertUtils.convert(), but getFormattedText() does not.

labkey-matthewb: MultiValuedDisplayColumn.getFormattedTexts() looks suspect.

PR #3105, opened 2022-03-01
Show buffer memory usage in Admin Console

labkey-jeckels:
#### Rationale
The JVM may be using a lot of buffer memory, causing instances to crash when the OS runs low on memory. Having some visibility will be helpful while we get full perf monitoring in place

#### Changes
* Avoid infinite mothership reporting on Throwables
* Show buffer pool usages
* Better formatting on memory info page
* Show CPU info and total OS memory

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -162,7 +162,6 @@ import org.labkey.api.settings.NetworkDriveProps;
 import org.labkey.api.settings.WriteableAppProps;
 import org.labkey.api.settings.WriteableFolderLookAndFeelProperties;
 import org.labkey.api.settings.WriteableLookAndFeelProperties;
-import org.labkey.api.study.StudyService;
 import org.labkey.api.util.*;
 import org.labkey.api.util.MemTracker.HeldReference;
 import org.labkey.api.util.SystemMaintenance.SystemMaintenanceProperties;
@@ -228,6 +227,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.sql.SQLException;
+import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -3406,13 +3410,19 @@ private MemBean(HttpServletRequest request, Set<Object> objectsToIgnore)
...
+            }
+
             // class loader:
             ClassLoadingMXBean classbean = ManagementFactory.getClassLoadingMXBean();
             if (classbean != null)
             {
-                systemProperties.add(new Pair<>("Loaded Class Count", classbean.getLoadedClassCount()));
-                systemProperties.add(new Pair<>("Unloaded Class Count", classbean.getUnloadedClassCount()));
-                systemProperties.add(new Pair<>("Total Loaded Class Count", classbean.getTotalLoadedClassCount()));
+                systemProperties.add(new Pair<>("Loaded Class Count", Formats.commaf0.format(classbean.getLoadedClassCount())));
```
labkey-matthewb: Yikes.  Formats are not thread-safe.  We should kill those constants (use getters or methods with synchronize?)

labkey-matthewb (approved):
Nice.  Some of the charts have weird scales (/0 or /-1?).  But otherwise looks good.

core/src/org/labkey/core/admin/AdminController.java
labkey-jeckels: I did some digging on this earlier. The general consensus seems to be that if you're not mutating the formatter itself, they're safe to share across threads. But I'm not sure I'm fully convinced of that argument. We could also use ThreadLocals.

labkey-matthewb: I'm not sure what the rules are but I fixed two teamcity failures recently related to this.  I'm not sure if this has become problematic with the new JVM?

labkey-jeckels:
I made the plots without real data to show look better. Not awesome, but good enough for this page.

core/src/org/labkey/core/admin/AdminController.java
labkey-jeckels: Interesting. I purged references to the shared Formats from this class but am not planning to eliminate all usages at this moment.

labkey-adam: We've seen problems with shared formats and `parse()`, but never with `format()`.

labkey-matthewb: Right.  Good point.

PR #3108, opened 2022-03-02
prefer dialect specific methods

labkey-matthewb:
#### Rationale
We tend to implement labkey SQL methods using JDBC function escape syntax by default.  However, it seems to be generally more reliable to call SQL methods in the native dialect directly.  Resolving against the dialect specific methods first lets us do this.  E.g. for SQL Server use UPPER() instead of {fn UCASE()}.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44681

#### Related Pull Requests
* 

#### Changes

also removed mysterious "::" handling.  I think this was some aborted idea related to scoping of module provided methods.

### DIFF
--- a/query/src/org/labkey/query/sql/Method.java
+++ b/query/src/org/labkey/query/sql/Method.java
@@ -1256,15 +1256,10 @@ public abstract class Method
     
     public static Method resolve(SqlDialect d, String name)
     {
+        Method m = null;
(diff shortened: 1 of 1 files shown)


labkey-martyp (approved):
Looks good. Tested as well and fixed issue.

PR #3109, opened 2022-03-02
Issue 38906: Assay data import convert string pk

labkey-martyp:
#### Rationale
Some file based assays like Simple2 have a string pk on import. This is getting an error that there is an incorrect data type. This is something we handle in other places to remap the string to the pk. Added pk mapping to assay import.

#### Changes
* Added attempted type conversion to SimpleTranslator.RemapPostConvert.fetch. If conversion fails than key not found.
* Update RemapCache to allow checking for key match when checking indices map.
* Update Assay TSV handler to use pk matching

### DIFF
--- a/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
+++ b/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
@@ -983,7 +983,7 @@ public abstract class AbstractAssayTsvDataHandler extends AbstractExperimentData
                     TableInfo lookupTable = remappableLookup.get(pd);
                     try
                     {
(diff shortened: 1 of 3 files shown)


labkey-jeckels (approved):
Looks fine to me but would be good for @labkey-susanh and @XingY to review as well. They've slogged through scenarios around how we treat numeric values - as sample names and/or RowIds.

labkey-susanh (approved):
I don't know much about the RemapCache, but this seems reasonable to me.

XingY (approved):
This looks good to me as well. 

There was a plan to repurpose includePkLookup to find sample by name, but that plan was dropped, which was a relief. The use of includePkLookup in this PR looks consistent with its original purpose to me.

PR #3110, opened 2022-03-02
Issue44305: Wiki attachments don't appear in the same order in imported folders

RosalineP:
#### Rationale
Previously, files attached to wiki pages don't appear in the same order when the folder is exported and reimported to another folder. 

#### Related Pull Requests
* N/A

#### Changes
* Add attr to wiki.xsd to keep track of file attachment order
* Sort incoming attachments based on above attr

### DIFF
--- a/wiki/schemas/wiki.xsd
+++ b/wiki/schemas/wiki.xsd
@@ -41,7 +41,7 @@
         </xsd:attribute>
         <xsd:attribute name="attachmentsOrder" type="xsd:string">
             <xsd:annotation>
-                <xsd:documentation>The order in which this page's attachments should appear</xsd:documentation>
+                <xsd:documentation>The order in which this page's attachments should appear, as a semicolon-delimited list of file names</xsd:documentation>
             </xsd:annotation>
         </xsd:attribute>
     </xsd:complexType>
--- a/wiki/schemas/wiki.xsd
+++ b/wiki/schemas/wiki.xsd
@@ -39,5 +39,10 @@
                 <xsd:documentation>Whether this page should be indexed by the search service for full text search</xsd:documentation>
             </xsd:annotation>
         </xsd:attribute>
+        <xsd:attribute name="attachmentsOrder" type="xsd:string">
+            <xsd:annotation>
+                <xsd:documentation>The order in which this page's attachments should appear</xsd:documentation>
+            </xsd:annotation>
+        </xsd:attribute>
     </xsd:complexType>
 </xsd:schema>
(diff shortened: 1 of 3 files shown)


wiki/schemas/wiki.xsd
```diff
@@ -39,5 +39,10 @@
                 <xsd:documentation>Whether this page should be indexed by the search service for full text search</xsd:documentation>
             </xsd:annotation>
         </xsd:attribute>
+        <xsd:attribute name="attachmentsOrder" type="xsd:string">
+            <xsd:annotation>
+                <xsd:documentation>The order in which this page's attachments should appear</xsd:documentation>
```
labkey-jeckels: Useful to mention that this is a semi-colon separated list of file names.

wiki/src/org/labkey/wiki/export/WikiImporterFactory.java
```diff
@@ -205,6 +208,9 @@ private Wiki importWiki(String name, String title, boolean shouldIndex, boolean
                     attachments.add(new InputStreamAttachmentFile(aIS, fileName));
                 }
             }
+            AtomicInteger inc = new AtomicInteger();
+            Map<String, Integer> attachmentOrderMap = List.of(attachmentOrder.split(";")).stream().collect(Collectors.toMap(i -> i, i -> inc.getAndIncrement()));
```
labkey-jeckels: Won't this NPE when attachmentOrder is null?

labkey-jeckels (changes requested):

wiki/schemas/wiki.xsd
RosalineP: Updated

wiki/src/org/labkey/wiki/export/WikiImporterFactory.java
RosalineP: Oops, yes, thank you. Added a nullcheck. I wasn't sure how to exercise the null case, however—the comment in WikiImporterFactory, line 131, says that its call to `importWiki` happens when there are wiki subdirectories that weren't present in the XML metadata, and while I tested with wiki subdirectories I wasn't sure how to test with wiki subdirectories that don't appear in the XML metadata.

labkey-jeckels: A couple of testing ideas:

1. Import a folder archive that has wiki attachments doesn't have the new attribute set. This could be done on a server that doesn't have your change yet, or by manually editing the XML prior to import.
2. Set a breakpoint on in this method and use the debugger to set the attachmentOrder variable to null before it uses it.

labkey-jeckels (approved):

wiki/src/org/labkey/wiki/export/WikiImporterFactory.java
RosalineP: Tested locally using method 2 👍

PR #3111, opened 2022-03-03
Item 9748: Shared freezers across containers

XingY:
#### Rationale
Location History for a freezer location is backed by audit logs. In order to show full history of a location, not limited to the current folder's history, we will need to get audit events with appropriate container filters.  

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/756
* https://github.com/LabKey/platform/pull/3111
* https://github.com/LabKey/inventory/pull/388

#### Changes
* support getAuditEvents with containerFilter

### DIFF
--- a/api/src/org/labkey/api/audit/DefaultAuditProvider.java
+++ b/api/src/org/labkey/api/audit/DefaultAuditProvider.java
@@ -19,6 +19,7 @@ package org.labkey.api.audit;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.audit.query.DefaultAuditSchema;
 import org.labkey.api.data.Container;
+import org.labkey.api.data.ContainerFilter;
 import org.labkey.api.data.SimpleFilter;
 import org.labkey.api.data.Sort;
 import org.labkey.api.data.TableInfo;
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/audit/DefaultAuditProvider.java
```diff
@@ -69,6 +70,12 @@ public boolean isViewable()
         return Collections.emptyList();
```
labkey-nicka, line 70: I would recommend telescoping the return value as well (like you did in `LogManager`).

```java
@Override
public <K extends AuditTypeEvent> List<K> getAuditEvents(Container container, User user, String eventType, @Nullable SimpleFilter filter, @Nullable Sort sort)
{
    return getAuditEvents(container, user, eventType, filter, null);
}
```

audit/src/org/labkey/audit/model/LogManager.java
```diff
@@ -201,6 +202,10 @@ public DbSchema getSchema()
     }
 
     public <K extends AuditTypeEvent> List<K> getAuditEvents(Container container, User user, String eventType, @Nullable SimpleFilter filter, @Nullable Sort sort)
+    {
+        return getAuditEvents(container, user, eventType, filter, sort, null);
+    }
+    public <K extends AuditTypeEvent> List<K> getAuditEvents(Container container, User user, String eventType, @Nullable SimpleFilter filter, @Nullable Sort sort, @Nullable ContainerFilter cf)
```
labkey-nicka, line 208: nit: Missing empty line between function definitions.

labkey-nicka (approved):

PR #3115, opened 2022-03-04
44969: Not possible to un-set ST Link-to-Study option

labkey-klum:
#### Rationale
Sample types need to be able to clear the auto copy target container from the settings panel. Previously we were only updating the option if the target container wasn't blank.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44969


RosalineP (approved):
Looks good! Pulled down and tested.

PR #3123, opened 2022-03-08
Issue 44839: Field names or labels including the special chars <, &, and > display as &lt;, &amp;, &gt; in grid view customizer

XingY:
#### Rationale
field captions returned by getQueryDetails api are html encoded. Ext tree renders row using template.apply, which does html encoding, so the result is the caption displayed is double encoded. 

#### Related Pull Requests
* 

#### Changes
* skip encoding field captions for getQueryDetails api
* update Designer panels to explicitly encode caption since they not longer encoded by the server.

### DIFF
--- a/api/src/org/labkey/api/data/JsonWriter.java
+++ b/api/src/org/labkey/api/data/JsonWriter.java
@@ -277,12 +277,12 @@ public class JsonWriter
         }
         else
         {
-            props.put("shortCaption", dc.getCaption());
+            props.put("shortCaption", dc.getCaption(null, false));
         }
 
-        props.put("caption", dc.getCaption());
+        props.put("caption", dc.getCaption(null, false));
         if (includeDomainFormat)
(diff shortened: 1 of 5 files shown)


labkey-adam (changes requested):
Actually, our standard practice now is for the server to NOT send HTML encoded content via APIs (except in very obvious cases). I don't know who is relying on these captions to be HTML encoded, but much better would be to change the API to not HTML encode (and fix up callers to encode client side if they need to).

XingY:
> Actually, our standard practice now is for the server to NOT send HTML encoded content via APIs (except in very obvious cases). I don't know who is relying on these captions to be HTML encoded, but much better would be to change the API to not HTML encode (and fix up callers to encode client side if they need to).

Thanks for the info on latest standard practice. Implementation updated.

labkey-adam (approved):
Awesome! Thanks, Xing.

PR #3124, opened 2022-03-08
Issue 44930: Increase Limit on List Names to Greater than 64 Characters

labkey-bpatel:
#### Rationale
Issue [44930](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44930): Increase Limit on List Names to Greater than 64 Characters

#### Changes
* Increase list name size to 200


labkey-klum (approved):
200 seems reasonable for now.

PR #3125, opened 2022-03-09
Issue 45008: Warnings about duplicate search.crawlcollections rows

labkey-jeckels:
#### Rationale
Some servers get a lot of warnings about attempts to insert duplicate rows into search.crawlcollections. The code is tolerant of this, so the logging is just spam.

#### Changes
* Skip logging the warning on this constraint violation

### DIFF
--- a/search/src/org/labkey/search/model/SavePaths.java
+++ b/search/src/org/labkey/search/model/SavePaths.java
@@ -15,6 +15,7 @@
  */
 package org.labkey.search.model;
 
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-adam @labkey-matthewb if you have ideas on how to avoid the constraint violation in the first place I'd be happy to change approaches

labkey-matthewb (approved):

PR #3128, opened 2022-03-09
Remove `Comparable` interface from modules

labkey-tchad:
#### Rationale
[Issue 45019: Module.compareTo breaks interface contract](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45019)
Our module `compareTo` method is not commutative when the core module is involved. With the right set/number of modules, `Arrays.sort` will complain thusly:
```
java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.util.TimSort.mergeHi(TimSort.java:903) ~[?:?]
	at java.util.TimSort.mergeAt(TimSort.java:520) ~[?:?]
	at java.util.TimSort.mergeForceCollapse(TimSort.java:461) ~[?:?]
	at java.util.TimSort.sort(TimSort.java:254) ~[?:?]
	at java.util.Arrays.sort(Arrays.java:1307) ~[?:?]
	at java.util.ArrayList.sort(ArrayList.java:1721) ~[?:?]
	at org.labkey.core.admin.AdminController$CreditsAction.getView(AdminController.java:956) ~[core-22.3-SNAPSHOT.jar:?]
```
The credits page is the only place that wants this particular ordering so it can take responsibility for creating it.

#### Changes
* Remove `Comparable` interface from modules

### DIFF
--- a/api/src/org/labkey/api/module/DefaultModule.java
+++ b/api/src/org/labkey/api/module/DefaultModule.java
@@ -153,10 +153,21 @@ public abstract class DefaultModule implements Module, ApplicationContextAware
     }
 
     @Override
-    public int compareTo(@NotNull Module m)
+    public final int compareTo(@NotNull Module m)
     {
-        //sort by name--core module will override to ensure first in sort
-        return getName().compareToIgnoreCase(m.getName());
+        // Ensure that the "Core" module is listed first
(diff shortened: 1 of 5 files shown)


labkey-adam (approved):
Change looks fine, but technically you need to reference an issue for a 22.3 PR. I'm also curious how/when this problem repros.

PR #2942, opened 2022-01-13
ExcelLoader perf

labkey-matthewb:
#### Rationale
Avoid always loading all sheet data into memory, for better performance and memory usage.  Also address https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44550

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/421
* https://github.com/LabKey/platform/pull/2942

#### Changes
*

PR #3131, opened 2022-03-11
Close open stream that was causing IOExceptions

Sigmonia:
#### Rationale
[Issue 45051: File streams are left open after import from Excel ](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45051)
Close open stream that was causing IOExceptions on Windows

#### Related PR
Seems related to https://github.com/LabKey/platform/pull/2942

#### Changes
* Add workbook.close to close method of ExcelLoader
* Use try-with-resources for ExcelLoaders

### DIFF
--- a/api/src/org/labkey/api/reader/jxl/JxlWorkbook.java
+++ b/api/src/org/labkey/api/reader/jxl/JxlWorkbook.java
@@ -381,7 +381,7 @@ public class JxlWorkbook implements Workbook
     @Override
     public void close()
     {
-        throw new UnsupportedOperationException("method not yet supported");
+        _workbook.close();
     }
 
     @Override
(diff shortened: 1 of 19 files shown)


labkey-matthewb (approved):

labkey-tchad:
Checking against DailyC, which hit this error several times: https://teamcity.labkey.org/buildConfiguration/LabKey_223Release_Community_DailySuites_DailyCSqlserver/1738092?showRootCauses=false&expandBuildChangesSection=true

api/src/org/labkey/api/reader/jxl/JxlWorkbook.java
```diff
@@ -381,7 +381,7 @@ public CellStyle getCellStyleAt(int i)
     @Override
     public void close()
     {
-        throw new UnsupportedOperationException("method not yet supported");
+        _workbook.close();
```
Sigmonia, line 384: Any chance `_workbook` can be null?
```suggestion
        if (_workbook != null)
            _workbook.close();

```

labkey-tchad: The jxl Workbook factories used to populate that won't return `null`.

PR #3133, opened 2022-03-12
45023: Unauthorized exception for readonly user - Dataset linked from flow data

labkey-klum:
#### Rationale
Datasets sourced from Flow data weren't respecting `ContextualRoles` when resolving the underlying tables:

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45023

#### Related Pull Requests
- https://github.com/LabKey/commonAssays/pull/439

#### Changes
- Teach the `ExpRunItemTable` to use `ContextualRoles` when checking permissions
-


labkey-matthewb (approved):
I added a comment, but I'll let you decide if you want to adopt it.

labkey-klum:
yes, saw that and thought it was a good suggestion, there is also this related PR for you to take a look at which uses the contextual role for the permission checking https://github.com/LabKey/commonAssays/pull/439

PR #3137, opened 2022-03-14
45034: CIML's BSIImport module failing due to S3 bug from 21.11

labkey-klum:
#### Rationale
I think this was introduced during the recent S3 work to support file watchers. We just need to make sure we are creating the directory first and then copying into the correct target temp file. The existing code was just copying the source file into a file which was the temp directory name (with no extension).

https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=45034


Sigmonia (approved):

labkey-jony:
@labkey-klum All checks except for Specimen SqlServer passed. Since the client is running on PostgreSQL, I think we can consider this good and merge, yeah?

labkey-klum:
Don't worry @labkey-jony it will get merged in due time and be in place for the formal release.

PR #3144, opened 2022-03-16
Issue 45042: Domain creation and save APIs don't properly set the description value for the domain

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45042

If you create a LabKey domain via the API (R, python, etc.), the description value for the domain that you pass to the POST body is not persisted in the LK database with the domain. Note that you can set the description for a domain (i.e. list, dataset, sample type) through the UI in LK server. This PR allows for that description to be persisted in the list, dataset, and data class cases (it was already supported for sample type creation).

#### Changes
* DataSetDomainKind/ListDomainKind/DataClassDomainKind update createDomain() to look for description value from domain object if not present in options/arguments

### DIFF
--- a/experiment/src/org/labkey/experiment/api/DataClassDomainKind.java
+++ b/experiment/src/org/labkey/experiment/api/DataClassDomainKind.java
@@ -333,6 +333,10 @@ public class DataClassDomainKind extends AbstractDomainKind<DataClassDomainKindP
         List<GWTPropertyDescriptor> properties = (List<GWTPropertyDescriptor>)domain.getFields();
         List<GWTIndex> indices = (List<GWTIndex>)domain.getIndices();
 
+        // Issue 45042: Allow for the dataClass description to be set via the create domain API calls
+        if (options.getDescription() == null && domain.getDescription() != null)
+            options.setDescription(domain.getDescription());
+
         try
         {
             ExpDataClass dataClass = ExperimentService.get().createDataClass(container, user, name, options, properties, indices, templateInfo);
--- a/list/src/org/labkey/list/model/ListDomainKind.java
+++ b/list/src/org/labkey/list/model/ListDomainKind.java
@@ -396,8 +396,10 @@ public abstract class ListDomainKind extends AbstractDomainKind<ListDomainKindPr
         ListDefinition list = ListService.get().createList(container, name, keyType, templateInfo, category);
         list.setKeyName(keyName);
 
-        String description = listProperties.getDescription() != null ? listProperties.getDescription() : domain.getDescription();
-        list.setDescription(description);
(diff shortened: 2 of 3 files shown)


experiment/src/org/labkey/experiment/api/DataClassDomainKind.java
```diff
@@ -333,6 +333,10 @@ public Domain createDomain(GWTDomain domain, DataClassDomainKindProperties optio
         List<GWTPropertyDescriptor> properties = (List<GWTPropertyDescriptor>)domain.getFields();
         List<GWTIndex> indices = (List<GWTIndex>)domain.getIndices();
 
+        // Issue 45042: Allow for the dataClass description to be set via the create domain API calls
+        if (options.getDescription() == null && domain.getDescription() != null)
```
labkey-susanh, line 337: Why the check for `options.getDescription() == null`?  Shouldn't the provided value always get passed through?

cnathe: In the UI case, yes the options are always passed through since we post those with the domain design save action. However, in the API case the user can leave out the options for this domain kind.

labkey-jeckels: I don't see how this check helps though. If the API caller didn't include a description, with this code, the domain will always end up with a null description when it's saved, won't it?

To handle the case you're describing, wouldn't we need to distinguish between undefined and null, effectively?

cnathe: The tricky part here is that both the domain object and the list/dataclass/dataset definition have a description but it is only the domain kind specific description that is shown in the UI for most (all?) domain kinds. 

In the field editor UI case, we always post (for both create and update) the full domain object and the domain kind specific options (like description, dataset category, dataset 3rd key, etc.). In those cases, the description comes through with the options map for the domain kind. 

However, in the APIs like Rlabkey and python, the description value that the user provides to the helper functions we have goes onto the domain object. So in those cases, this PR is trying to be a little more tolerant of looking in both places (i.e. the domain kind options first and then the domain object second) for that description. The other approach would be to update those API package helpers to stop putting the description with the domain object. If you think that would be a better approach, I can take a look into that.

labkey-jeckels: Thanks for the info. It seems like we should probably get rid of the description on exp.dataclass, exp.materialsource, etc and just use the domaindescriptor field. But I'm sure there's plenty of ambiguous data rows out there already.

labkey-jeckels (approved):
Ultimately would be good to consolidate but this at least seems like a step in the right direction for consistency.

PR #3145, opened 2022-03-16
Set.of() doesn't allow duplicate values

labkey-adam:
#### Rationale
Study resourceId equals container resourceId when "basic security" is selected. Switch to a real `Set` to accommodate.

### DIFF
--- a/query/src/org/labkey/query/LinkedSchema.java
+++ b/query/src/org/labkey/query/LinkedSchema.java
@@ -62,6 +62,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -581,7 +582,7 @@ public class LinkedSchema extends ExternalSchema
     {
         private static final int[] NO_GROUPS = new int[0];
 
-        private final Set<String> _allowedPolicyResourceIds;
+        private final Set<String> _allowedPolicyResourceIds = new HashSet<>();
 
         public LinkedSchemaUserWrapper(User realUser, Container sourceContainer)
         {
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):

query/src/org/labkey/query/LinkedSchema.java
```diff
@@ -590,14 +591,15 @@ public LinkedSchemaUserWrapper(User realUser, Container sourceContainer)
...
             StudyService ss = StudyService.get();
             if (null != ss)
-                study = ss.getStudy(sourceContainer);
-
-            _allowedPolicyResourceIds = null != study ? Set.of(containerResourceId, study.getPolicy().getResourceId()) : Set.of(containerResourceId);
+            {
+                Study study = ss.getStudy(sourceContainer);
+                if (null != study)
+                    _allowedPolicyResourceIds.add(study.getPolicy().getResourceId());
+            }
```
labkey-tchad, line 602: Do we want to make this immutable just to be on the safe side? Something like:
```
    Set<String> temp = new HashSet<>();
    temp.add(sourceContainer.getResourceId());

    StudyService ss = StudyService.get();
    if (null != ss)
    {
        Study study = ss.getStudy(sourceContainer);
        if (null != study)
            temp.add(study.getPolicy().getResourceId());
    }
    _allowedPolicyResourceIds = Collections.unmodifiableSet(temp);
```

labkey-tchad (approved):
Consider making Set immutable

query/src/org/labkey/query/LinkedSchema.java
labkey-adam: Possibly do this in a follow-up, but it shouldn't block merge and hold up 22.3.0. Not a real concern given that it's completely private to this class.

labkey-adam:
LinkedSchemaTest is all green: https://teamcity.labkey.org/viewLog.html?buildId=1743939&buildTypeId=LabKey_223Release_Community_DailySuites_DailyAPostgres&tab=testsInfo

ONPRC test suites have a few failures, but appear unrelated (cache timeout deadlock and socket timeout): https://teamcity.labkey.org/viewLog.html?buildId=1743935&buildTypeId=LabKey_223Release_Premium_Ehr_OnprcEhrSqlserver

PR #3151, opened 2022-03-17
Lineage lookups

labkey-matthewb:
#### Rationale
Support for filter/sort on columns in parent sample/dataclass

Under experimental feature flag for the moment.

#### Related Pull Requests
* 

#### Changes
*


api/src/org/labkey/api/data/MaterializedQueryHelper.java
```diff
@@ -394,6 +397,53 @@ public SQLFragment getFromSql(@NotNull SQLFragment selectQuery, String tableAlia
...
+                    materialized = null;
+                }
+            }
+        }
+        return materialized;
+    }
+
+
+    /* Do incremental update to existing cached data.  There is no provision for deleting rows. */
+    public void upsert()
```
labkey-susanh, line 442: Is this method needed?

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
```diff
@@ -0,0 +1,536 @@
...
+
+    static SQLFragment selectSql(SqlDialect d, SQLFragment from)
+    {
+        String cte = d.isPostgreSQL() ? pgMaterialClosureCTE : mssqlMaterialClosureCTE;
+        String select = d.isPostgreSQL() ? pgMaterialClosureSql : mssqlMaterialClosureSql;
+
+        String[] cteParts = StringUtils.splitByWholeSeparator(cte,"/*FROM*/");
+        assert cteParts.length == 2;
+
+        return new SQLFragment(cteParts[0]).append(from).append(cteParts[1]).append(select);
```
labkey-susanh, line 160: Seems like you might need some spaces around the `from` here.

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
```diff
@@ -0,0 +1,536 @@
...
+            for (var c : queryHelpers.values())
+            {
+                c.counter.incrementAndGet();
+                c.helper.uncache(null);
+            }
+        }
+    }
+
+
+//    public static void invalidateCpasType(String sourceTypeLsid)
```
labkey-susanh: Remove?

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
```diff
@@ -0,0 +1,536 @@
...
+
+
+
+    /*
+     * Code to create the lineage parent lookup column and intermediate lookups that use ClosureQueryHelper
+    */
+
+    public static MutableColumnInfo createLineageLookupColumnInfo(String columnName, FilteredTable<UserSchema> parent, ColumnInfo rowid, ExpObject source)
+    {
+        //MutableColumnInfo wrappedRowId = wrapColumn("LineageLookupTypes", _rootTable.getColumn("rowid"));
```
labkey-susanh: Remove?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -2958,211 +2958,279 @@ public void syncRunEdges(int runId)
...
-            }
+        void sync(@Nullable Map<String, Integer> cpasTypeToObjectId)
+        {
+            syncRunEdges(_runId, _runObjectId, _runLsid, _runContainer, deleteFirst, verifyEdgesNoInsert, invalidateClosureCache, cpasTypeToObjectId);
+        }
 
-            // delete all existing edges for this run
-            if (deleteFirst)
-                removeEdgesForRun(runId);
+        private void syncRunEdges(int runId, Integer runObjectId, String runLsid, Container runContainer, boolean deleteFirst, boolean verifyEdgesNoInsert, boolean invalidateClosureCache, @Nullable Map<String, Integer> cpasTypeToObjectId)
```
labkey-susanh: Seems like you could get rid of the boolean arguments that are now set in the builder.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -2958,211 +2958,279 @@ public void syncRunEdges(int runId)
...
+                SQLFragment datas = new SQLFragment()
+                        .append("SELECT d.Container, d.LSID, d.ObjectId, pa.CpasType AS pa_cpas_type FROM exp.Data d\n")
+                        .append("INNER JOIN exp.DataInput di ON d.rowId = di.dataId\n")
+                        .append("INNER JOIN exp.ProtocolApplication pa ON di.TargetApplicationId = pa.RowId\n")
+                        .append("WHERE pa.RunId = ").append(runId).append(" AND pa.CpasType IN ('").append(ExperimentRun.name()).append("','").append(ExperimentRunOutput.name()).append("')");
+
+                Collection<Map<String, Object>> fromDataLsids = new ArrayList<>();
+                Collection<Map<String, Object>> toDataLsids = new ArrayList<>();
+                new SqlSelector(getSchema(), datas).forEachMap(row -> {
+                    if (ExperimentRun.name().equals(row.get("pa_cpas_type")))
```
labkey-susanh, line 3052: I know this isn't you, but this string matching seems somewhat magical.  Is there at least a constant for this `pa_cpas_type` string?

labkey-susanh (commented):
Code seems reasonable enough, but there's something wrong in here because the Data Class columns are not added as lookups and display only the rowId values.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-matthewb: The enums ExperimentRun and ExperimentRunOutput are the constants.  "pa_cpas_type" is just a column name from the SQL immediately above.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-matthewb: yup, it was just easier to do this incrementally, but the method can be flattened now.

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
labkey-matthewb: This matches the text in the SQL string constants above.

api/src/org/labkey/api/data/MaterializedQueryHelper.java
labkey-matthewb: used by ClosureQueryHelper.incrementalRecompute().  MQH doesn't directly expose the names of names of the tables it's managing, so this seemed right.

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
labkey-susanh: On line 146, it looks like this: 
```
.append(cteParts[0]).append(" ").append(from).append(" ").append(cteParts[1])
```
So, either the extra space appends aren't needed above, or they would be needed here, or I'm not reading this correctly. Not a big deal if they're extra above, but I get bitten by these missing spaces too many times.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -670,6 +672,12 @@ protected ContainerFilter getLookupContainerFilter()
         setTitleColumn(Column.Name.toString());
 
         setDefaultVisibleColumns(defaultCols);
+
+        if (null != _ss && AppProps.getInstance().isExperimentalFeatureEnabled(ExperimentModule.EXPERIMENTAL_LINEAGE_PARENT_LOOKUP))
+        {
+            MutableColumnInfo lineageLookup = ClosureQueryHelper.createLineageLookupColumnInfo("Ancestor Lookups Placeholder", this, _rootTable.getColumn("rowid"), _ss);
```
labkey-susanh, line 678: Could just call this "Ancestors" or "Ancestor Lookups".

labkey-susanh (approved):
Definitely working better now for data classes.  I've not tested extensively, though.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
labkey-matthewb: Or "Lineage Lookups".  I figured the spec was the right place to make a decision, but that if I put a "reasonable" thing in the code it would be sticky.

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
```diff
@@ -0,0 +1,534 @@
...
+
+                SELECT CTE_.Start_, Edge.FromObjectId as End_, CTE_.Path_ || CAST(Edge.FromObjectId AS VARCHAR) || '/' as Path_, Depth_ + 1 as Depth_
+                FROM CTE_ INNER JOIN exp.Edge ON CTE_.End_ = Edge.ToObjectId
+                WHERE Depth_ < %d AND 0 = POSITION('/' || CAST(Edge.FromObjectId AS VARCHAR) || '/' IN Path_)
+                
+            )
+            """, MAX_LINEAGE_LOOKUP_DEPTH);
+
+    static String pgMaterialClosureSql = """
+            SELECT Start_, CASE WHEN COUNT(*) = 1 THEN MIN(rowId) ELSE -1 * COUNT(*) END AS rowId, targetId
```
labkey-martyp, line 85: So this gives a count of rows when not 1:1.  I imagine in some cases we'll want to provide some feedback why a value isn't showing up in a column that is not 1:1.  Any thoughts how to propagate that up in the query?  Looks like this will just be part of the join and won't match anything.

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
```diff
@@ -0,0 +1,534 @@
...
+                return null;
+            }
+        });
+        return wrappedRowId;
+    }
+
+
+    enum TableType
+    {
+        SampleType("Materials", SchemaKey.fromParts("exp","materials") )
```
labkey-martyp, line 435: Might be nice if we could alias these eventually (samples, sources, registry,...), especially if we're going to make these available in-app eventually.

labkey-martyp (commented):
This is great. A little different than what I understood from the spec, but matches more my understanding of the use case.

Seems like this could help with Sample Finder resolving lookups as well?

A couple inline comments to think about.

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
labkey-martyp: Looks like the negative row count shows up in columns that are just selecting the material or dataclass instead of a value in the material or dataclass.  Might be useful for providing user feedback in UIs, but if not providing an error in the actual response, maybe at least in a future fb provide a comment or explanation for anyone looking to use this

PR #3154, opened 2022-03-17
Some improvements to default sorting behavior.

labkey-matthewb:
#### Rationale
1) Do a better job of trying to pull-through query "ORDER BY" into view sorts
2) Try to append a unique column to all view sorts to provide better sort order stability.

#### Related Pull Requests
* 

#### Changes
Should document that 
  ```SELECT A, B, A+B AS C @hidden ... ORDER BY C```
is preferable to 
  ```SELECT A, B ... ORDER BY A+B```


labkey-adam (changes requested):
sql.Query$QueryTestCase failing with a PIVOT sort exception

labkey-adam (approved):

PR #3156, opened 2022-03-18
ForeignKey definitions should really hold a SchemaKey not a String.

labkey-matthewb:
#### Rationale
Since schemas live in a tree, we should use SchemaKey where possible instead of String to reduce ambiguity.  E.g. SchemaKey.fromParts("exp","data") instead of "exp.data"

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/data/ForeignKey.java
+++ b/api/src/org/labkey/api/data/ForeignKey.java
@@ -20,12 +20,14 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.collections.NamedObjectList;
 import org.labkey.api.query.FieldKey;
+import org.labkey.api.query.SchemaKey;
 import org.labkey.api.util.StringExpression;
 import org.labkey.data.xml.queryCustomView.FilterType;
 
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 /**
@@ -84,7 +86,14 @@ public interface ForeignKey
      * Just for introspection.
      * @return The name of the foreign user schema table.
      */
-    String getLookupSchemaName();
+    @Deprecated
+    default String getLookupSchemaName()
+    {
+        return Objects.toString(getLookupSchemaKey(),null);
+    }
+
+    /* Schema path relative to the DefaultSchema (e.g. container) */
+    SchemaKey getLookupSchemaKey();
 
     /**
      * Just for introspection.
--- a/api/src/org/labkey/api/data/ForeignKey.java
+++ b/api/src/org/labkey/api/data/ForeignKey.java
@@ -85,6 +85,8 @@ public interface ForeignKey
     /**
      * Just for introspection.
      * @return The name of the foreign user schema table.
+     *
+     * Consider using getLookupSchemaKey() instead.
      */
     @Deprecated
     default String getLookupSchemaName()
--- a/api/src/org/labkey/api/exp/property/Lookup.java
+++ b/api/src/org/labkey/api/exp/property/Lookup.java
@@ -33,8 +33,8 @@ public class Lookup
     {
     }
 
-    // Callers need to be explicit about whether this name is encoded or not
-    // I'm going to assume SchemaName is "/" encoded
+    // Callers need to be explicit about how this is encode.  This constructer assumes "SchmeaKey.toString()" encoding.
+    // Consider using ScehmaKey constructor instead.
     @Deprecated
     public Lookup(Container c, String schema, String query)
     {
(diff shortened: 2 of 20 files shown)


api/src/org/labkey/api/data/ForeignKey.java
```diff
@@ -84,7 +86,14 @@ default TableDescription getLookupTableDescription()
      * Just for introspection.
      * @return The name of the foreign user schema table.
      */
-    String getLookupSchemaName();
+    @Deprecated
```
labkey-susanh, line 91: Should probably indicate what should be used instead.

api/src/org/labkey/api/exp/property/Lookup.java
```diff
@@ -19,21 +19,34 @@
...
+    SchemaKey _schemaKey;
     String _queryName;
 
     public Lookup()
     {
     }
 
+    // Callers need to be explicit about whether this name is encoded or not
+    // I'm going to assume SchemaName is "/" encoded
+    @Deprecated
```
labkey-susanh, line 38: Also should indicate here what the preferred alternative is.

api/src/org/labkey/api/exp/property/Lookup.java
```diff
@@ -42,9 +55,15 @@ public Container getContainer()
         return _container;
     }
 
+    @Deprecated
```
labkey-susanh, line 58: Same here.

api/src/org/labkey/api/exp/property/Lookup.java
```diff
@@ -59,7 +78,7 @@ public void setContainer(Container container)
 
     public void setSchemaName(String name)
     {
-        _schemaName = name;
+        _schemaKey = null == name ? null : new SchemaKey(null, name);;
```
labkey-susanh, line 81: Seems like the method name should be `setSchemaKey` now.

api/src/org/labkey/api/query/PdLookupForeignKey.java
```diff
@@ -57,33 +57,33 @@ static public PdLookupForeignKey create(QuerySchema sourceSchema, @NotNull User
...
                 targetContainer = lookup.getContainer();
-                lookupSchemaName = lookup.getSchemaName();
+                lookupSchemaKey = lookup.getSchemaKey();
                 lookupQuery = lookup.getQueryName();
             }
         }
 
         ContainerFilter cf;
-        if ("core".equalsIgnoreCase(lookupSchemaName) && "Containers".equalsIgnoreCase(lookupQuery))
+        if ("core".equalsIgnoreCase(null==lookupSchemaKey?null:lookupSchemaKey.getName()) && "CoClassImposteriserntainers".equalsIgnoreCase(lookupQuery))
```
labkey-susanh: `CoClassImposteriserntainers`? Probably not what you mean.

api/src/org/labkey/api/query/QueryForeignKey.java
```diff
@@ -228,7 +224,7 @@ public ForeignKey build()
             boolean isLabKeyScope = null != sourceSchema && (null == sourceSchema.getDbSchema() || sourceSchema.getDbSchema().getScope().isLabKeyScope());
             if (isLabKeyScope)
             {
-                if ("core".equalsIgnoreCase(lookupSchemaName) && "containers".equalsIgnoreCase(lookupTableName) && effectiveContainer.equals(sourceSchema.getContainer()))
+                if ("core".equalsIgnoreCase(lookupSchemaName.getName()) && "containers".equalsIgnoreCase(lookupTableName) && effectiveContainer.equals(sourceSchema.getContainer()))
```
labkey-susanh: Any chance lookupSchemaName is null here?

labkey-susanh (changes requested):

api/src/org/labkey/api/exp/property/Lookup.java
labkey-matthewb: Yeah, I don't know what this thing is used for.  Seems to be some sort of API binding thing.  Didn't want to break any reflection based binding.

api/src/org/labkey/api/query/PdLookupForeignKey.java
labkey-matthewb: Maybe, maybe not...

api/src/org/labkey/api/exp/property/Lookup.java
labkey-matthewb: Also kinda weird to have mismatched get/set methods.  I'll do a usage search.

labkey-susanh: This still seems weird to me, but the effect for the user is as expected, so I won't insist.

labkey-susanh (approved):

PR #3158, opened 2022-03-18
Flag more column lists as unstable-ordered

labkey-adam:
#### Rationale
Many `TableSelector` methods require a "stable-ordered" column list, meaning they are held in a collection that has predictable iteration order. For example, `getValueMap()` populates a map, treating the first column as the key and the second column as the value. Any method that returns or operates on `ResultSet`s or `Result`s needs predictable column ordering as well, since these interfaces include methods that accept column indexes. Passing a column list as (say) a `HashSet` and using one of these methods will likely lead to unpleasant surprises, so TableSelector checks and throws if one of the designated methods is used with an unstable-ordered column collection.

The current stable-ordered check rejects `HashSet` except when its size is < 2. On the surface, the size check makes total sense... a collection of size one (or zero) definitely has predictable ordering. However, this exception contributed to [a recent issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45016) where a new scenario caused code that had previously always selected a single column to start selecting more columns. That triggered the unstable-ordered exception. Removing the size check means we'll now fail any time a HashSet is passed, which may uncover lurking bugs similar to the one we just fixed.

This PR also restricts the set implementations that are considered stable-ordered. It switches to an allow list that includes `LinkedHashSet`, `Collections.emptySet()` (used to specify all columns), and `Collections.singleton()`. Everything else is designated as unstable. Note that `Set.of()` is tempting to use for fixed column lists, but it's clearly documented as not guaranteeing any particular iteration order, so it's now rejected.

I plan to run a large number of test suites on this PR before merging.

#### Changes
* Disallow HashSet, TreeSet, and Set.of() column lists
* Remove custom module sorting on credits page - no need to treat core as special here
* Trim cache creation stack traces to pinpoint the actual cache creation line


labkey-adam:
The suites below plus the Daily, EHR Daily, Git, Module, Custom Modules, and External Modules suites are all clear

labkey-jeckels (approved):

PR #3160, opened 2022-03-22
fix inadvertently causing full reindexing

Sigmonia:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45066
If a folder is not supposed to be indexed, then we don't index it's children either. However we need to make an exception for the root container, where we don't want to index it, but we do want to index it's children. 

#### Changes
* Adjusted index path enumeration to check for "/" root folder.

### DIFF
--- a/search/src/org/labkey/search/model/DavCrawler.java
+++ b/search/src/org/labkey/search/model/DavCrawler.java
@@ -230,7 +230,8 @@ public class DavCrawler implements ShutdownListener
         if (null != start)
         {
             _log.debug("START CONTINUOUS " + start.toString());
-            _paths.updatePath(start, null, nextCrawl, true);
+            // make sure path exists, don't update if it already exists.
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
I looked at this but don't understand enough of how the crawler does its thing to give a meaningful review. @labkey-matthewb can you take a look?

labkey-jeckels:
I looked at this but don't understand enough of how the crawler does its thing to give a meaningful review. @labkey-matthewb can you take a look?

labkey-matthewb:
> 

I wrote the PR!

labkey-matthewb:
Basically, the crawler doesn’t like it when the root container “/“ says “don’t index me”.  Also, it doesn’t like it when the “start crawler” command says “look at this never before seen directory I found called “/_webdav/“.  Both of these force reindex of the children.

labkey-adam (approved):

PR #3165, opened 2022-03-23
Wiki aliases

labkey-adam:
#### Rationale
Our documentation pages have some very outdated names (cpasxml, sampleSets, workflowModules), but updating them is problematic because renaming a page would break all links that have been sent to clients via email, tickets, etc. Our documentation team has requested a new feature, wiki aliases, to allow them to rename wikis while still resolving page requests to the old name. These are patterned after container aliases. The basic spec plus doc team requests:

1. Add the option (on by default) to create a wiki alias having the old name at the time a wiki is renamed

2. Aliases are respected in only one place: if wiki-page.view is invoked with a "name" that doesn't match an existing wiki we'll then consult the list of aliases in that specific container only. If "name" is a valid alias then LabKey will redirect to the same URL, replacing "name" with the real name of the aliased wiki, and maintaining/forwarding any #anchors included with the alias name to the real page name (i.e. oldname#section would go to newname#section). Every other action (e.g., edit, delete, manage, copy, print, etc.) would require real names.

3. As a result of 2, if a wiki is later added with an existing alias name, that wiki would take precedence over the aliased one. (This matches the container alias behavior.)

4. Constraints will ensure that you can't add the same alias twice in the same container, but different containers can have different alias lists. (e.g., portals/lksm docs and main docs won't collide or cross-link to each other.)

5. The manage wiki page shows a list of all aliases associated with that wiki. The alias list can be edited here.

The back-links action will stay focused on real names (ignore aliases) so the docs pages would continue to use canonical names.

Possible future enhancements: managing via an "all aliases in this folder" view could be useful for release-time-link-checking and periodic cleanups, index aliases for full-text search (not sure that's really desired).

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45063

### DIFF
--- a/announcements/resources/schemas/dbscripts/postgresql/comm-22.000-22.001.sql
+++ b/announcements/resources/schemas/dbscripts/postgresql/comm-22.000-22.001.sql
@@ -1,0 +1,8 @@
+CREATE TABLE comm.PageAliases
+(
+    Container ENTITYID NOT NULL,
+    Alias VARCHAR(255) NOT NULL,
+    RowId INT NOT NULL,
+
+    CONSTRAINT PK_PageAliases PRIMARY KEY (Container, Alias)
+);
--- a/api/src/org/labkey/api/view/PermanentRedirectException.java
+++ b/api/src/org/labkey/api/view/PermanentRedirectException.java
@@ -5,6 +5,7 @@ import org.labkey.api.util.URLHelper;
 
 import javax.servlet.http.HttpServletResponse;
 
+/** Use when we want search engines, browsers, etc to assume that the redirecting URL is defunct and the target URL should be used going forward */
 public class PermanentRedirectException extends RedirectException
 {
     public PermanentRedirectException(@NotNull URLHelper url)
--- a/api/src/org/labkey/api/view/PermanentRedirectException.java
+++ b/api/src/org/labkey/api/view/PermanentRedirectException.java
@@ -1,0 +1,20 @@
+package org.labkey.api.view;
+
+import org.jetbrains.annotations.NotNull;
+import org.labkey.api.util.URLHelper;
+
+import javax.servlet.http.HttpServletResponse;
+
+public class PermanentRedirectException extends RedirectException
+{
+    public PermanentRedirectException(@NotNull URLHelper url)
+    {
+        super(url);
+    }
+
+    @Override
+    public int getHttpStatusCode()
+    {
+        return HttpServletResponse.SC_MOVED_PERMANENTLY;
+    }
+}
--- a/wiki/resources/web/wiki/internal/wikiEdit.js
+++ b/wiki/resources/web/wiki/internal/wikiEdit.js
@@ -172,7 +172,7 @@ function tinyMceHandleEvent(evt) {
     var bindControls = function(props) {
         // form controls
         var setDirty = function(){LABKEY.setDirty(true)};
-        $(_idSel + 'name').keypress(setDirty).change(onChangeName);
+        $(_idSel + 'name').keypress(setDirty).change(setDirty);
         $(_idSel + 'title').keypress(setDirty).change(setDirty);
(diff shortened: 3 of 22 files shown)


wiki/src/org/labkey/wiki/WikiController.java
```diff
@@ -1174,8 +1171,13 @@ public ModelAndView getView(WikiNameForm form, BindException errors)
             _wiki = WikiSelectManager.getWiki(getContainer(), name);
             boolean existing = _wiki != null;
 
-            if (null == _wiki)
+            if (!existing)
             {
+                // Redirect if name is an alias
+                String realName = WikiSelectManager.getNameForAlias(getContainer(), name);
+                if (null != realName)
+                    throw new RedirectException(getViewContext().getActionURL().clone().replaceParameter("name", realName));
```
labkey-jeckels: I can imagine some logging being useful here at some point. Maybe a debug-level message that an admin could opt into that includes the old and new names and the referrer URL?

announcements/resources/schemas/dbscripts/postgresql/comm-22.000-22.001.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE comm.PageAliases
+(
+    Container ENTITYID NOT NULL,
+    Alias VARCHAR(255) NOT NULL,
+    RowId INT NOT NULL,
+
+    CONSTRAINT PK_PageAliases PRIMARY KEY (Container, Alias)
```
labkey-jeckels, line 7: I don't see a listener for when the container goes away. Consider a FK to core.Containers too to ensure cleanup

labkey-jeckels (commented):
I know you're didn't tag anyone for review yet but I was interested so I took a quick peek.

announcements/resources/schemas/dbscripts/postgresql/comm-22.000-22.001.sql
labkey-adam: Added alias delete to existing container listener and wiki delete. I didn't add an FK since none of the comm tables has one.

wiki/src/org/labkey/wiki/WikiController.java
labkey-adam: Logging added

api/src/org/labkey/api/view/PermanentRedirectException.java
```diff
@@ -0,0 +1,20 @@
+package org.labkey.api.view;
+
+import org.jetbrains.annotations.NotNull;
+import org.labkey.api.util.URLHelper;
+
+import javax.servlet.http.HttpServletResponse;
+
+public class PermanentRedirectException extends RedirectException
```
labkey-jeckels, line 9:
```suggestion
/** Use when we want search engines, browsers, etc to assume that the redirecting URL is defunct and the target URL should be used going forward */
public class PermanentRedirectException extends RedirectException
```

wiki/resources/web/wiki/internal/wikiEdit.js
```diff
@@ -363,31 +363,9 @@ function tinyMceHandleEvent(evt) {
         window.location = _cancelUrl ? _cancelUrl : getRedirUrl();
     };
 
+    // Note: can't change an existing wiki's name
```
labkey-jeckels: Can this go away then?

labkey-jeckels (approved):

wiki/resources/web/wiki/internal/wikiEdit.js
labkey-adam: What I left is just the insert handling case. I think this is needed for dirty handling on a brand new wiki where of course the name is editable.

labkey-adam: But you're right... no need for a separate function that just sets dirty. I switched to the logic used by all the other fields.

RosalineP (approved):
👍 👍 👍

PR #3183, opened 2022-03-28
Issue 45132: Semicolon in file name truncates only on download

Sigmonia:
#### Rationale
[Issue 45132](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45132): Semicolon in file name truncates only on download
https://httpwg.org/specs/rfc6266.html#examples

#### Changes
* adjusted the header to use the `filename*` encoded value per the RFC

### DIFF
--- a/core/src/org/labkey/core/webdav/DavController.java
+++ b/core/src/org/labkey/core/webdav/DavController.java
@@ -5052,7 +5052,7 @@ public class DavController extends SpringActionController
                 if (HttpUtil.isChrome(getRequest()))
                 {
                     Path requestPath = new URLHelper(getRequest().getRequestURI()).getParsedPath();
-                    getResponse().setContentDisposition(contentDisposition + "; filename=" + requestPath.getName());
+                    getResponse().setContentDisposition(contentDisposition + "; filename*=" + PageFlowUtil.encode(requestPath.getName()));
                 }
             }
             catch (URISyntaxException x)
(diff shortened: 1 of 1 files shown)


labkey-matthewb (commented):
We might want to verify that the original chrome bug that this is a workaround for is still present.

https://bugs.chromium.org/p/chromium/issues/detail?id=1503

core/src/org/labkey/core/webdav/DavController.java
```diff
@@ -5052,7 +5052,8 @@ private WebdavStatus serveResource(WebdavResource resource, boolean content)
                 if (HttpUtil.isChrome(getRequest()))
```
labkey-martyp: Ah, my bad.  I just tested the original issue on MS Edge and it fails as well.  FF is ok.  Wonder if we can just do this format regardless of browser type?  Does FF break?

labkey-matthewb: Ideally we would remove the browser workaround and not do the filename part of this.  The browser should pull the name off the URL by default.  There's really no reason we should have to duplicate it in this header except Chrome was (is?) messed up.

labkey-matthewb: edit: though I supposed it lets us provide the file with the correct canonical casing?  In any event, it would be preferable to NOT have a browser specific code path.

Sigmonia: I'll play with it a bit more. But leaving off the `filename` parameter causes chrome to try to open it after downloading it.

labkey-matthewb: I see.  If this encoding allows us to properly encode the file name in all cases (instances?)  then maybe we could remove the isChrome check?

Sigmonia: Removed the isChrome check. Bug appears fixed/verified. I also anecdotally checked it with tar.gz files and a few other extensions that were mentioned in the comments.

Checked this PR change on Chrome, Brave, FF, and Edge (Windows)

labkey-matthewb (approved):

PR #3184, opened 2022-03-28
Render wiki comparisons in a more compact view

labkey-adam:
#### Rationale
Our documentation team and others use the wiki comparison view to review changes made to wiki content over time. The current view, however, can be difficult to use because it shows a full diff that often includes a large amount of unchanged text along with a small number of changes. Despite color highlighting of inserts and deletes, the changes can be very difficult to spot. The new approach replaces large swaths of unchanged text, if present, with a single line placeholder that states the number of unchanged lines that were omitted. This effectively highlights the changes. For example:

image

### DIFF
--- a/internal/src/org/labkey/api/util/DiffMatchPatch.java
+++ b/internal/src/org/labkey/api/util/DiffMatchPatch.java
@@ -19,6 +19,8 @@
 
 package org.labkey.api.util;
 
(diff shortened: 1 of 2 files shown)


RosalineP (approved):
Looks good, works well, didn't find nitpicks here!

PR #3192, opened 2022-03-30
Remove inline script event handlers

labkey-matthewb:
#### Rationale
Update markup generation to avoid using "on" events as much as possible.  This is a big step towards being able to run with a Content-Security-Policy that does not require "script-src; 'unsafe-inline'"

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3192
* https://github.com/LabKey/testAutomation/pull/1084
* https://github.com/LabKey/commonAssays/pull/457
* https://github.com/LabKey/OConnorLabModules/pull/89
* https://github.com/LabKey/MacCossLabModules/pull/196

#### Changes
Refactor class PageConfig to be a 'global' object that collects event handlers rather than inline them.  
Update code builders to use PageConfig
Update PageConfig to hold a generated script nonce for the current page.


api/src/org/labkey/api/jsp/JspBase.java
```diff
@@ -757,11 +771,17 @@ protected HtmlString formAction(Class<? extends Controller> actionClass, Method
     }
 
     // Provides a unique integer within the context of this request. Handy for generating element ids, etc. See UniqueID for caveats and warnings.
+    @Deprecated // makeId() is preferred
     protected int getRequestScopedUID()
     {
         return UniqueID.getRequestScopedUID(getViewContext().getRequest());
     }
 
+    protected String makeId(String prefix)
```
labkey-jeckels, line 780: Comments on this method and getScriptNonce() would be great.

internal/src/org/labkey/api/security/AuthFilter.java
```diff
@@ -211,6 +213,22 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 
         QueryService.get().setEnvironment(QueryService.Environment.USER, user);
 
+        if (AppProps.getInstance().isExperimentalFeatureEnabled("experimental-unsafe-inline"))
```
labkey-jeckels, line 216: Can we use a constant for this?

api/src/org/labkey/api/util/Button.java
```diff
@@ -189,15 +191,17 @@ private String generateOnClick(String id)
 
         if (isSubmit())
         {
-            final String submitCode = "submitForm(document.getElementById(" + quote + id + quote + ").form);return false;";
+            final String submitCode = "submitForm(document.getElementById(this.dataset[" + quote + "submitid" + quote + "]).form);return false;";
```
labkey-jeckels, line 196: I'm not seeing why this assumes "this.dataset"

api/src/org/labkey/api/util/Button.java
```diff
@@ -314,6 +328,13 @@ public ButtonBuilder submit(boolean submit)
             return this;
         }
 
+        /* ONLY WHEN RENDER MARKUP INTO JAVASSCRIPT CODE */
```
labkey-jeckels:
```suggestion
        /* ONLY WHEN RENDER MARKUP INTO JAVASCRIPT CODE */

```

api/src/org/labkey/api/util/PageFlowUtil.java
```diff
@@ -1477,72 +1472,187 @@ public static String textLink(String text, URLHelper url)
...
+            return this;
+        }
+
+        public HelpPopupBuilder script(String onClickScript)
+        {
+            this.onClickScript = onClickScript;
+            return this;
+        }
+
+        /* ONLY USE TO RENDER INTO JAVASSCRIPT CODE */
```
labkey-jeckels:
```suggestion
        /* ONLY USE TO RENDER INTO JAVASCRIPT CODE */
```

api/src/org/labkey/api/view/WebPartView.java
```diff
@@ -129,9 +131,22 @@ public ApiResponse renderToApiResponse()
...
             {
                 WebPartView.this.render(request, mr);
+                var config = HttpView.currentPageConfig();
+                var sw = new StringWriter();
+                config.endOfBodyScript(sw);
+                var script = sw.toString();
+                if (!script.isBlank())
+                {
+                    var out = mr.getWriter();
+                    out.print("<script type=\"text/javascript\" nonce=\"" + config.getScriptNonce() + "\">");
```
labkey-jeckels, line 146: HTML encode to be safe?

internal/webapp/util.js
```diff
@@ -330,3 +330,83 @@ LABKEY.internal.SortUtil = new function()
...
+                        let groupRowIds = LABKEY.notifications.grouping[groupings[i]];
+                        for (let j = 0; j < groupRowIds.length; j++) {
+                            const rowId = groupRowIds[j], info = LABKEY.notifications[rowId];
+
+                            // get the date/time display string based on the current date
+                            const monthNames = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"],
+                                    d = new Date(info.Created), today = new Date(),
+                                    dStr = d.toDateString() == today.toDateString() ? 'Today' : monthNames[d.getMonth()] + ' ' + d.getDate();
+
+                            html += "<div class='labkey-notification' id='notification-" + rowId + "' onclick='LABKEY.Notification.goToActionLink(event, " + rowId + "); return true;'>"
```
labkey-jeckels, line 389: I know this variable is called "rowId", implying an integer... but since it's untyped JS, HTML encode?

api/src/org/labkey/api/view/template/PageConfig.java
```diff
@@ -468,4 +511,148 @@ public JSONObject getPortalContext()
...
+
+
+    public String makeId(String prefix)
+    {
+        return prefix + _sid + _uid.incrementAndGet(); // we can concatenate without a separator because _sid is fixed width
+    }
+
+
+    @NotNull
+    public static String getScriptNonce(HttpServletRequest request)
```
labkey-jeckels: Can we have matching return types between the two overloaded versions of this method?

labkey-jeckels (approved):
Some minor suggestions but overall looks fine. Happy to review any followup commits if you'd like.

I think I lost a couple places where I commented that the result of PageConfig.makeId() was injected into the HTML without encoding. I know the values should be safe to pass through as-is. But since it's String->HTML...

api/src/org/labkey/api/util/Button.java
labkey-matthewb: I'll add a comment.  This is because in the markup we stash the submitId in the element using data-submitid attrubute (line 251).  This means we don't need to generate unique javascript just because the id of the target element has changed.

labkey-matthewb: PS: this code could probably find the form by looking for a parent, but I suppose this strategy allows the button to be outside the form tag it is associated with?

api/src/org/labkey/api/view/template/PageConfig.java
labkey-matthewb: They're not really overloads since one is static and used in only one place...  I could name it something like initNonceHeader().

internal/src/org/labkey/api/security/AuthFilter.java
labkey-matthewb: It's hard to import constants into the API module isn't it.  I could move from where the ExperimentalFilterFlag is defined, but that seems weird too.  Preference?

api/src/org/labkey/api/view/WebPartView.java
labkey-matthewb: This is already an HtmlString.  The fix would be to switch the surrounding code to use HtmlStringBuilder instead of writing directly to out.

internal/webapp/util.js
labkey-matthewb: I just moved that code.  But yes, that would be better.

internal/src/org/labkey/api/security/AuthFilter.java
labkey-jeckels: Since the place that acts on the setting is in the API module, declaring it there makes the most sense to me. But it's not something I'm passionate about either

api/src/org/labkey/api/util/Button.java
labkey-jeckels: I see the submitid part below, but I didn't see where `this.dataset` part was defined. I'll trust that it works though

api/src/org/labkey/api/view/WebPartView.java
labkey-jeckels: I swear I checked before and it was returning String but I'm now seeing it as HtmlString too, so never mind.

labkey-jeckels: Maybe I got confused by the static version of the method.

labkey-jeckels (approved):
I replied about where to register the experimental feature but don't feel strongly one way or the other

api/src/org/labkey/api/util/Button.java
labkey-matthewb: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

labkey-jeckels: Ah, thanks. I foolishly assumed the study `dataset`

PR #3200, opened 2022-03-31
45152: Support displaying Data Class data from subfolder

labkey-nicka:
#### Rationale
This allows for Data Class data to be viewed and interacted with from folders that are not the definition folder. This addresses [45152](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45152) so users can now delete data in a Data Class in a subfolder.

#### Changes
* `ShowDataClassAction` now supports the following:
  * Displays Data Class Properties for Data Class cross-folder.
  * Adds a "Defined In: /Folder" if viewed away from definition folder.
  * Provides modal warning when navigating away from folder.
  * Button permissions determined by target folder for action.
* Links from Data Class listing are now folder relative. This allows for data to be browsed in the current folder.


labkey-klum (approved):
looks like a nice addition

PR #3206, opened 2022-04-01
Issue 45164: Improve HTML error parsing

labkey-jeckels:
#### Rationale
It's possible to have both bad HTML and script

#### Changes
* Assume that bad HTML may have script


mbellew (commented):
Need null check before adding warning?

labkey-jeckels:
> Need null check before adding warning?

Indeed. Added

labkey-matthewb (approved):

PR #3217, opened 2022-04-03
Refactor import/export context and writer hierarchies

labkey-adam:
#### Rationale
The import/export context, writer, and importer hierarchies have some confusingly named interfaces and classes, and they're harder to implement than needed

#### Changes
- Rename a few import/export classes to better reflect their use. For example, `ImportContext` is implemented by both import and export contexts, so rename to `ImportExportContext`
- Add a default implementation for `FolderImporter.postProcess()` since only a few non-trivial overrides exist
- Now that we're no longer passing folder and study contexts to the same importer methods (except for one case: `AbstractDataStateImporter`), replace most usages of `ImportExportContext<FolderDocument.Folder>` with `FolderExportContext` or `FolderImportContext`, as appropriate
- Fix some cases where we were passing an export context to a method expecting an import context and vice versa
- In `BaseFolderWriter.write()` implementations, prefer passed in `Container` vs. `ctx.getContainer()`
- Fix up bad generics


api/src/org/labkey/api/reports/report/ScriptReportDescriptor.java
```diff
@@ -81,12 +81,11 @@ public boolean isArrayType(String prop)
     }
 
     @Override
-    public ReportDescriptorDocument getDescriptorDocument(ImportContext context)
+    public ReportDescriptorDocument getDescriptorDocument(FolderExportContext context)
     {
+        assert context != null;
```
labkey-jeckels: I gave Binal similar feedback earlier this week. What's the value in this assert? On dev machines, we'll fail with an AssertionError that gives no info. On production servers, we'll fail one line later with an NPE.

labkey-jeckels (approved):

api/src/org/labkey/api/reports/report/ScriptReportDescriptor.java
labkey-adam: We'll fail sooner with this in place 🙂

labkey-adam: Removed

PR #3224, opened 2022-04-05
Item 10180: Support Ancestor Lookup display in grids and Sample Finder

labkey-susanh:
#### Rationale
Users want to be able to include data from ancestors of samples in their sample grids and be able to sort and filter using that data.  We previously provided support for adding special lookup columns that supports the most common case of there being only one ancestor of a particular type within a sample's lineage.  Here we make modifications so the our grids display the ancestor columns well, particularly when the lineage includes multiple ancestors of a particular type (which is treated as a null value for purposes of sorting and filtering but displayed with muted text to give users an indication that their lineage is not lost). 

We also divide the sample types and data classes based on their category.  This has the advantage of using language that users are more familiar with but a slight disadvantage that if a user changes the category of a type (e.g., from "Registry" to "Media" or vice versa), the names of the query fields will change.  It is believed that the desire to do this modification is rare, but Registry and Sources were combined under one type because these seem the most likely types to get modified from one to other.  If this division of the types seems too fragile or fraught with difficulty we can use the simpler division of "Samples" and "Data" and agree to do something better for the display to users at a later date.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/799
* https://github.com/LabKey/biologics/pull/1237
* https://github.com/LabKey/sampleManagement/pull/911

#### Changes
* Remove experimental feature flag for Ancestor lookups
* Update labeling from Ancestor Lookup Placeholder to Ancestors
* Add AncestorLookupDisplayColumn
* Add conceptURI to Ancestor fields for help in rendering on the client side
* Add more table types for Ancestor lookups


labkey-jeckels (commented):
No objections here but Matt probably has more context.

@labkey-martyp you may be in the loop on this work already but this seems relevant for LKSM/EHR integration efforts.

experiment/src/org/labkey/experiment/api/AncestorLookupDisplayColumn.java
```diff
@@ -0,0 +1,76 @@
...
+        super(currentCol);
+        _lookupCol = lookupCol == null ? currentCol : lookupCol;
+    }
+
+    @Override
+    public @NotNull HtmlString getFormattedHtml(RenderContext ctx)
+    {
+        Long lookupKey = getLookupId(ctx);
+        if (lookupKey != null && lookupKey < 0)
+            return HtmlString.unsafe("<span style=\"color: gray;\"><"+ (-lookupKey) + " values></span>");
```
labkey-matthewb, line 41: CSS class?

labkey-matthewb (approved):

labkey-martyp:
> No objections here but Matt probably has more context.
> 
> @labkey-martyp you may be in the loop on this work already but this seems relevant for LKSM/EHR integration efforts.

Yes this is great. I haven't tried this PR out but saw some of the work leading up to this.  This is the exact case for EHR/SM integration to link the two data sources.  Also great to see including this update in the Sample Finder.

Sigmonia (approved):

PR #3227, opened 2022-04-06
Throw exception when lsid is not found in table selector list

ankurjuneja:
#### Rationale
EHR billing tables have extensible columns and when those table rows are updated, the existing values for extensible columns get lost when lsid is not found in table selector.

#### Related Pull Requests
* https://github.com/LabKey/tnprc_billing/pull/153


api/src/org/labkey/api/query/DefaultQueryUpdateService.java
```diff
@@ -254,6 +255,10 @@ public void updateStatistics(int currentRow)
                     }
                 }
             }
+            else
+            {
+                throw new NotFoundException("LSID value not found in table - " + table.getName());
```
labkey-jeckels: NotFoundException is useful when the client set us a request for something that doesn't exist. We don't log it as an error because it was a bogus request.

IllegalStateException is a better fit here, as this represents something that shouldn't happen. It'll get logged and be more likely to be noticed with test failures.

labkey-jeckels (changes requested):

api/src/org/labkey/api/query/DefaultQueryUpdateService.java
ankurjuneja: makes sense. changed.

api/src/org/labkey/api/query/DefaultQueryUpdateService.java
```diff
@@ -50,6 +50,7 @@
 import org.labkey.api.security.permissions.Permission;
 import org.labkey.api.security.permissions.UpdatePermission;
 import org.labkey.api.util.Pair;
+import org.labkey.api.view.NotFoundException;
```
labkey-jeckels, line 53: This can be trimmed now.

labkey-jeckels (approved):
I know you're still tracking down test failures but the fact that this change exposed those problems is a good confirmation of the approach.

PR #3229, opened 2022-04-07
Set protocol status to 'active' for Application Type 'ExperimentRun'

labkey-bpatel:
#### Rationale
During (Panorama) folder import, running into
"Protocol with LSID 'urn:lsid:labkey.com:Protocol.Folder-20:...' does not match existing protocol"
'Status differs: 'null' vs 'Active'

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/515

#### Changes
* Set protocol status to 'active' for Application Type 'ExperimentRun' in insertSimpleProtocol() to match other areas in the code (Ex. XarReader.java, line 2266 and upgrade script exp-21.016-21.017.sql)

### DIFF
--- a/experiment/resources/schemas/dbscripts/postgresql/exp-22.001-22.002.sql
+++ b/experiment/resources/schemas/dbscripts/postgresql/exp-22.001-22.002.sql
@@ -1,1 +1,1 @@
-UPDATE exp.Protocol SET Status = 'Active' WHERE ApplicationType = 'ExperimentRun';
+UPDATE exp.Protocol SET Status = 'Active' WHERE ApplicationType = 'ExperimentRun' WHERE Status IS NULL;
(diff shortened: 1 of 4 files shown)


labkey-jeckels (changes requested):
I imagine we need another upgrade script to set Status=Active for protocols created this way since the original upgrade script?

experiment/resources/schemas/dbscripts/postgresql/exp-22.001-22.002.sql
```diff
@@ -0,0 +1 @@
+UPDATE exp.Protocol SET Status = 'Active' WHERE ApplicationType = 'ExperimentRun';
```
labkey-jeckels: We don't want to stomp over any deactivated protocols, do we?

```suggestion
UPDATE exp.Protocol SET Status = 'Active' WHERE ApplicationType = 'ExperimentRun' WHERE Status IS NULL;
```

experiment/resources/schemas/dbscripts/sqlserver/exp-22.001-22.002.sql
```diff
@@ -0,0 +1 @@
+UPDATE exp.Protocol SET Status = 'Active' WHERE ApplicationType = 'ExperimentRun';
```
labkey-jeckels:
```suggestion
UPDATE exp.Protocol SET Status = 'Active' WHERE ApplicationType = 'ExperimentRun' WHERE Status IS NULL;
```

labkey-jeckels (approved):
Please see the suggestion for not blindly setting all protocols to Active

PR #3231, opened 2022-04-07
Improve AttachmentFile implementations & add WikiService.getAllContent

labkey-alan:
#### Rationale
This PR does a few things that are needed for the improved embedded attachment behavior in ELN (see changes below).

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/1239
* https://github.com/LabKey/labbook/pull/155

#### Changes
* Make the behavior for openInputStream and closeInputStream consistent across all AttachmentFile implementations
  * InputStreamAttachmentFile is the one exception, because it cannot be re-opened.
* Adds getAllContent to WikiService so modules can fetch the content for every version of a given wiki
  * This is needed so we can search previous versions for attachments and remove any attachments that do not exist in any version of a wiki.

### DIFF
--- a/api/src/org/labkey/api/wiki/WikiService.java
+++ b/api/src/org/labkey/api/wiki/WikiService.java
@@ -70,6 +70,11 @@ public interface WikiService
     String getContent(Container c, String wikiName);
 
     /**
+     * Returns the raw (unformatted) content for every version of the wiki
+     */
+    List<String> getAllContent(Container c, String wikiName);
+
+    /**
      * Update the content of a wiki
      *
      * @param wikiName The name of the wiki to update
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/wiki/WikiService.java
```diff
@@ -69,6 +69,11 @@ static void setInstance(WikiService impl)
      */
     String getContent(Container c, String wikiName);
 
+    /**
+     * Returns the raw (unformatted) content for every version of the wiki
+     */
+    List<String> getAllContent(Container c, String wikiName);
```
labkey-nicka: Mark this interface as `@Nullable` since it could return `null` in the case the wiki itself is not found.

labkey-nicka (approved):

PR #3235, opened 2022-04-08
Roundtrip wiki aliases through export/import and copy wiki

labkey-adam:
#### Rationale
Wiki aliases should be copied during folder export/import and copy wiki operations. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45207

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3165

### DIFF
--- a/wiki/src/org/labkey/wiki/WikiManager.java
+++ b/wiki/src/org/labkey/wiki/WikiManager.java
@@ -320,9 +320,8 @@ public class WikiManager implements WikiService
 
     /**
      * Attempts to add the specified aliases to the specified wiki. This is a best effort operation; failure to add an
-     * alias (e.g., an alias with that name already exists in this container) will result in an error (added to the
-     * BindException collection if not null, otherwise logged as a warning) but adding will continue and no exception
-     * will be thrown.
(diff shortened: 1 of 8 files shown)


wiki/src/org/labkey/wiki/WikiManager.java
```diff
@@ -205,6 +205,11 @@ public void insertWiki(User user, Container c, Wiki wikiInsert, WikiVersion wiki
 
             getAttachmentService().addAttachments(wikiInsert.getAttachmentParent(), files, user);
 
+            if (null != aliases)
+            {
+                WikiManager.get().addAliases(wikiInsert, aliases, null);
```
labkey-jeckels, line 210: Should we delete existing aliases if there aren't any being passed in?

labkey-jeckels (approved):

wiki/src/org/labkey/wiki/WikiManager.java
labkey-adam: No need... this is `insertWiki()`, so there are no aliases associated with the brand new wiki.

PR #3240, opened 2022-04-09
Show the "Manage Remote Connections" button only if dataintegration is present

labkey-adam:
#### Rationale
Remote connections can only be used from an ETL, so the management button should be hidden when the dataintegration module isn't available. We may eventually move the management code to the dataintegration module, but this simple approach is good enough for now.

This also updates the message at the top of the manage pages for accuracy. In particular, the previous warning seemed wrong and unnecessarily scary:

`This feature should be used with care since, depending on your configuration, any user with access to the remote site could view arbitrary data in your remote server.`

Please review the new message for accuracy.

### DIFF
--- a/query/webapp/query/browser/Browser.js
+++ b/query/webapp/query/browser/Browser.js
@@ -175,17 +175,19 @@ Ext4.define('LABKEY.query.browser.Browser', {
             });
         }
         if (LABKEY.Security.currentUser.isAdmin) {
-            tbar.push({
-                xtype: 'querybutton',
-                text: 'Manage Remote Connections',
-                tooltip: 'Manage remote connection credentials for remote LabKey server authentication.',
-                fontCls: 'fa-server',
-                stacked: true,
-                handler: function() {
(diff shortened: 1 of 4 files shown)


query/webapp/query/browser/Browser.js
```diff
@@ -175,17 +175,19 @@ Ext4.define('LABKEY.query.browser.Browser', {
...
-                handler: function() {
-                    window.location = LABKEY.ActionURL.buildURL('query', 'manageRemoteConnections');
-                },
-                scope: this
-            });
+            if ("dataintegration" in LABKEY.moduleContext) {
+                tbar.push({
+                    xtype: 'querybutton',
+                    text: 'Manage Remote Connections',
+                    tooltip: 'Manage configurations for remote connections to other LabKey Server instances.',
```
labkey-jeckels:
```suggestion
                    tooltip: 'Manage configurations for ETL connections to other LabKey Server instances.',

```

labkey-jeckels (approved):

PR #3243, opened 2022-04-11
44316: Add default selection to "Create related issue" dialog

labkey-klum:
#### Rationale
We want to introduce a new dropdown in the issue definition properties panel to allow administrators to configure the default folder that is offered when a user creates a new related issue.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44316

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/802

#### Changes
- save the default related folder in the prop store
- initialize the selection of the folder in the related issue dialog box

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0",
       "dependencies": {
         "@labkey/api": "1.10.0",
-        "@labkey/components": "2.150.0",
+        "@labkey/components": "2.150.1-fb-issue-44316.0",
         "@labkey/themes": "1.2.1"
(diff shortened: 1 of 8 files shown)


labkey-tchad:
Issues found during manual testing
- [x] Related issues should default to current issue list if no explicit default is set
- [x] Default should be ignored if user doesn't have permission to default's folder

Other functionality looks good (including shared/inherited issues domains)

cnathe (approved):

labkey-tchad:
Verified manual testing fixes 👍

PR #3244, opened 2022-04-11
Various 22.5 bugs for LKSM

XingY:
#### Rationale
Issue 45215: Performance issues with Search for Sample Manager
The getCustomSearchJson call in search action fetches each hit's custom data individually, which means at least one DB query per hit. When the number of hits is large, the DB overhead is resulting in poor performance for LKSM, who needs to getCustomSearchJson for each hit to support its advanced search function. Instead of fetching row by row, a change is made to fetch at 1,000 a batch.
Issue 45180: LKSM: Naming patterns with parents always add with square brackets
when the parent size is 1, return the object instead of the array
Issue 45007: Cannot create a new sample type with a naming pattern that references a field in the sample type
Look at draft fields in addition to DB fields

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/803
* https://github.com/LabKey/platform/pull/3244
* https://github.com/LabKey/sampleManagement/pull/916
* https://github.com/LabKey/testAutomation/pull/1072

#### Changes
* See details in Rationale section

### DIFF
--- a/api/src/org/labkey/api/data/NameGenerator.java
+++ b/api/src/org/labkey/api/data/NameGenerator.java
@@ -717,7 +717,7 @@ public class NameGenerator
 
         }
 
-        if ((isCurrentDataType || inputDataType == null) && !_domainProperties.isEmpty())
+        if ((isCurrentDataType || inputDataType == null) && _domainProperties != null && !_domainProperties.isEmpty())
         {
             Map<String, GWTPropertyDescriptor> domainFields = new CaseInsensitiveHashMap<>();
             _domainProperties.forEach(prop -> domainFields.put(prop.getName(), prop));
--- a/api/src/org/labkey/api/data/NameGenerator.java
+++ b/api/src/org/labkey/api/data/NameGenerator.java
@@ -677,10 +677,16 @@ public class NameGenerator
             else
                 dataTypes.addAll(ExperimentService.get().getDataClasses(_container, user, true));
         }
+
+        boolean isCurrentDataType = inputDataType != null && inputDataType.equals(currentDataType);
+
         if (inputDataType != null && dataTypes.isEmpty())
         {
-            _syntaxErrors.add("Invalid lineage lookup: " + fkTok.toString() + ".");
-            return null;
+            if (!isCurrentDataType)
+            {
+                _syntaxErrors.add("Invalid lineage lookup: " + fkTok.toString() + ".");
+                return null;
+            }
         }
 
         for (ExpObject dataType : dataTypes)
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -711,6 +717,28 @@ else if (dataType instanceof ExpDataClass)
 
         }
 
+        if ((isCurrentDataType || inputDataType == null) && _domainProperties != null && !_domainProperties.isEmpty())
+        {
+            Map<String, GWTPropertyDescriptor> domainFields = new CaseInsensitiveHashMap<>();
+            _domainProperties.forEach(prop -> domainFields.put(prop.getName(), prop));
+
+            GWTPropertyDescriptor col = domainFields.get(lookupField);
```
labkey-susanh, line 725: Seems like you could just do a filter on `_domainProperties` instead of constructing the `domainFields` map and then using only one value from the map.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -518,6 +518,18 @@ public ExpDataImpl getExpData(String lsid)
         return ExpDataImpl.fromDatas(new TableSelector(getTinfoData(), new SimpleFilter(FieldKey.fromParts("RowId"), rowids, IN), null).getArrayList(Data.class));
     }
 
+    @Override
+    public @NotNull List<? extends ExpData> getExpDatas(ExpDataClass dataClass, Collection<Integer> rowIds)
+    {
+        if (rowIds.size() == 0)
+            return Collections.emptyList();
+        SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("RowId"), rowIds, IN);
+        filter.addCondition(FieldKey.fromParts("classId"), dataClass.getRowId());
```
labkey-susanh, line 527: Does the `classId` help you here? The `rowId1s are the primary key so they have to be unique across data classes.  Does adding classId make it more efficient?

search/src/org/labkey/search/model/AbstractSearchService.java
```diff
@@ -663,32 +664,75 @@ public HttpView<?> getCustomSearchResult(User user, @NotNull String resourceIden
...
             return null;
         String prefix = resourceIdentifier.substring(0, i);
-        ResourceResolver res = _resolvers.get(prefix);
 
-        if (null == res)
+        if (_resolvers.containsKey(prefix))
+            return new Pair<>(prefix, resourceIdentifier.substring(i+1));
+
+        // Special case to allow customizing docs whose docId does not starting with category.
+        // For example, assay design doc id is of "containerId:assays:rowId"
```
labkey-susanh, line 678: Hmm.  Makes me wonder if those ids really should start with a category.  Not something to change here since this was just moved from elsewhere, but it makes me wonder.

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: I think the idea is to filter out stale data that happened to have the same rowId but different data class?

PR #3249, opened 2022-04-14
Switch "Body" column from hidden to simply not shown in default view

labkey-adam:
#### Rationale
The `Body` column of the `wiki.CurrentWikiVersions` and `wiki.AllWikiVersions` queries has been marked hidden to keep it out of the default view, since it doesn't format very well in a grid column. But hidden makes it too hard for most users to find, so switch to simply not showing it in the default view.


RosalineP (approved):
Nice, I like it! 
Manually tested.

PR #3138, opened 2022-03-14
Issue 45046: Site Level Project Creator role

labkey-adam:
#### Rationale
Provides a new site-level "Project Creator" role that grants the ability to create new projects and assign the Project Administrator role to the requesting user

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45046

PR #3254, opened 2022-04-17
Issue 45273: Importing groups during "create from template" can result in unauthorized exceptions

labkey-adam:
#### Rationale
Refresh import user's groups after group import to ensure new memberships for that user are recognized.

[Issue 45273: Importing groups during create from template can result in unauthorized exceptions](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45273)

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3138

### DIFF
--- a/api/src/org/labkey/api/security/User.java
+++ b/api/src/org/labkey/api/security/User.java
@@ -215,7 +215,8 @@ public class User extends UserPrincipal implements Serializable, Cloneable
 
     public void refreshGroups()
     {
-        _groups = null;
+        if (!isGuest()) // GuestUser explicitly sets groups in its constructor
+            _groups = null;
     }
 
     public boolean isFirstLogin()
--- a/api/src/org/labkey/api/security/User.java
+++ b/api/src/org/labkey/api/security/User.java
@@ -213,6 +213,10 @@ public class User extends UserPrincipal implements Serializable, Cloneable
         return _groups;
     }
 
+    public void refreshGroups()
+    {
+        _groups = null;
+    }
 
     public boolean isFirstLogin()
     {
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/security/User.java
```diff
@@ -213,6 +213,10 @@ public int[] getGroups()
         return _groups;
     }
 
+    public void refreshGroups()
+    {
+        _groups = null;
```
labkey-jeckels, line 218: Any risk of this messing up the Guest user's groups? They're set in its constructor instead of being loaded lazily in getGroups()

labkey-jeckels (approved):

api/src/org/labkey/api/security/User.java
labkey-adam: Hopefully guest isn't creating projects and importing folders, but I'll add a check in the new method

labkey-jeckels (approved):

api/src/org/labkey/api/security/User.java
```diff
@@ -213,6 +213,11 @@ public int[] getGroups()
         return _groups;
     }
 
+    public void refreshGroups()
```
labkey-matthewb, line 216: Isn't Guest a subclass?  It could override.

labkey-matthewb (approved):

api/src/org/labkey/api/security/User.java
labkey-adam: Not sure it's demonstrably better, but sure, that works too

labkey-matthewb: I'm just channeling your preference for enums to handle their own behavior differences (no switch!).  Same thing.

labkey-adam: Yep, I changed it

PR #3255, opened 2022-04-18
Issue 45275: SQL Server upgrade very slow in exp-21.013-21.014.sql

labkey-jeckels:
#### Rationale
exp-21.013-21.014.sql is populating the exp.ProtocolApplication.EntityId column by updating the rows one Table.update() call at a time. That's slow when you have more than a million rows.

#### Changes
* For SQL Server, use NEWID() via a bulk UPDATE call instead of Java upgrade code

### DIFF
--- a/experiment/resources/schemas/dbscripts/sqlserver/exp-21.013-21.014.sql
+++ b/experiment/resources/schemas/dbscripts/sqlserver/exp-21.013-21.014.sql
@@ -16,6 +16,8 @@
 
 ALTER TABLE exp.ProtocolApplication ADD EntityId ENTITYID;
 
-EXEC core.executeJavaUpgradeCode 'generateExpProtocolApplicationEntityIds';
+GO
+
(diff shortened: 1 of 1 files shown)


labkey-alan (approved):
Looks good to me, but I am wondering if we have an equivalent solution for Postgres? There is bound to be a Postgres installation out there with a similar sized database.

labkey-matthewb (approved):
Yikes, never use Table.insert/update/delete in a loop.  ParameterMapStatement is your friend.

labkey-jeckels:
@labkey-alan agreed that the perf could be problematic for large Postgres installs. I needed to get the SQL Server customer's upgrade unblocked so I merged as-is.

The script was part of 21.11 so it's large Postgres-based deployments that are on 21.10 that stand to benefit additional optimization. Postgres introduced gen_random_uuid() in version 13, but we support back to version 10 at the moment. So it wasn't a trivial update.

PR #3256, opened 2022-04-18
Item 10203: Support sample type and data class renaming

XingY:
#### Rationale
Sample type and dataclass names are not updatable once created currently. This PR adds support for allowing updating type names. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/808
* https://github.com/LabKey/platform/pull/3256
* https://github.com/LabKey/commonAssays/pull/458
* https://github.com/LabKey/sampleManagement/pull/922
* https://github.com/LabKey/biologics/pull/1255

#### Changes
* switch sample/dataclass type lsid to use container+DBsequence, instead of container+name
* update LSIDRelativizer to handle sample/dataclass lsid during folder import/export with new DB seq based lsid
* support updating sampletype and dataclass names


api/src/org/labkey/api/exp/api/ExperimentService.java
```diff
@@ -470,10 +472,14 @@ ExpMaterialProtocolInput createMaterialProtocolInput(
 
     String generateGuidLSID(Container container, Class<? extends ExpObject> clazz);
 
+    Pair<String, String> generateLsidWithDBSeq(@NotNull Container container, Class<? extends ExpObject> clazz);
```
labkey-susanh:
```suggestion
    Pair<String, String> generateLSIDWithDBSeq(@NotNull Container container, Class<? extends ExpObject> clazz);
```
Just to be consistent with the other `generateLSID` names.

experiment/src/org/labkey/experiment/LSIDRelativizer.java
```diff
@@ -74,6 +76,9 @@ protected String relativize(Lsid lsid, RelativizedLSIDs lsids)
             String prefix = lsid.getNamespacePrefix();
             String suffix = lsid.getNamespaceSuffix();
 
+            String sharedFolderSuffix = "Folder-" + ContainerManager.getSharedContainer().getRowId();
```
labkey-susanh, line 79: Is the addition of the `Folder-` prefix just for clarity (just curious; I like clarity)?

Does this change cause any problems with importing existing folder archives?

experiment/src/org/labkey/experiment/LSIDRelativizer.java
```diff
@@ -82,7 +87,20 @@ else if ("ProtocolApplication".equals(prefix))
...
+            {
+                /*
+                 * old prefix: "xxx:Sample:123.sampleTypeName"
+                 * new prefix: "xxx:Sample:Folder-123.345" (Sample:Folder-ContainerRowId.DBSeq)
+                 * relative lsid: "xxx:Sample:Folder-yyy.xarJobId.345" (need to add xarJobId because DBSeq might not be unique in target folder)
+                 */
+                String id = "";
+                int ind = suffix.indexOf(".");
+                if (ind > 0)
+                    id = suffix.substring(suffix.indexOf("."));
```
labkey-susanh:
```suggestion
                    id = suffix.substring(ind);

```

experiment/src/org/labkey/experiment/XarReader.java
```diff
@@ -1419,7 +1433,9 @@ private ExpMaterial loadMaterial(MaterialBaseType xbMaterial,
         {
             declaredType = LsidUtils.resolveLsidFromTemplate(declaredType, context, "SampleSet");
         }
-        ExpSampleTypeImpl sampleSet = checkMaterialCpasType(declaredType);
+        ExpSampleType sampleSet = checkMaterialCpasType(declaredType);
+        if (sampleSet != null)
+            declaredType = sampleSet.getLSID();
```
labkey-susanh, line 1438:
```suggestion
            declaredTypeLSID = sampleSet.getLSID();
```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -7293,30 +7313,45 @@ public ValidationException updateDataClass(@NotNull Container c, @NotNull User u
...
     {
+        ValidationException errors;
+
         // if options doesn't have a rowId value, then it is just coming from the property-editDomain action only only updating domain fields
         DataClassDomainKindProperties options = properties != null && properties.getRowId() == dataClass.getRowId() ? properties : null;
+        boolean hasNameChange = false;
         if (options != null)
         {
             validateDataClassOptions(c, u, options);
+            String newName = StringUtils.trimToNull(options.getName());
```
labkey-susanh, line 7328: This is probably caught elsewhere, but `null` seems like a bad name (as does the empty string). Perhaps that should be a check that produces a validation error?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -477,6 +477,19 @@ public Lsid getSampleTypeLsid(String sourceName, Container container)
...
+    public Pair<String, String> getSampleTypeSamplePrefixLsids(Container container)
+    {
+        Pair<String, String> lsidDbSeq = ExperimentService.get().generateLsidWithDBSeq(container, ExpSampleType.class);
+        String sampleTypeLsidStr = lsidDbSeq.first;
+        Lsid sampleTypeLsid = Lsid.parse(sampleTypeLsidStr);
+
+        String dbSeqStr = lsidDbSeq.second;
+        String samplePrefixLsid = new Lsid.LsidBuilder("Sample", "Folder-" + container.getRowId() + "." + dbSeqStr, "").toString();
+
+        return new Pair<>(sampleTypeLsid.toString(), samplePrefixLsid);
```
labkey-susanh, line 489: We get a string (`sampleTypeLsidStr`) and then convert to an `Lsid` and then return the string version of that.  Is the conversion to the `Lsid` needed?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -909,8 +924,26 @@ public Function<Map<String,Long>,Map<String,Long>> getSampleCountsFunction(@Null
     @Override
     public ValidationException updateSampleType(GWTDomain<? extends GWTPropertyDescriptor> original, GWTDomain<? extends GWTPropertyDescriptor> update, SampleTypeDomainKindProperties options, Container container, User user, boolean includeWarnings)
     {
+        ValidationException errors;
+
         ExpSampleTypeImpl st = new ExpSampleTypeImpl(getMaterialSource(update.getDomainURI()));
 
+        String newName = StringUtils.trimToNull(update.getName());
```
labkey-susanh, line 930: Same comment here about check for a null or empty name.

labkey-susanh (changes requested):

experiment/src/org/labkey/experiment/LSIDRelativizer.java
XingY: Yes it's for clarify. Now that LSID consist of db seq, which is also integer, it would help to clarify the first part is folder id.

This class is used for export so this line of change will only affect new export. Importing old archive should still work, and the manual test plan covers it.

experiment/src/org/labkey/experiment/XarReader.java
XingY: It can be an lsid, but also can be general "Material"/"Data".

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
XingY: Yes, the toString method of LSID has custom code to generate lsid.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: For update, name is optional. I've made the change to not update to null if name is null.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
XingY: Change made to not setName to null

labkey-susanh (approved):

PR #3258, opened 2022-04-18
ExperimentUpgradeCode: Use ParameterMapStatement in generateExpProtocolApplicationEntityIds

labkey-alan:
#### Rationale
The upgrade code in question is causing perf problems for large SQL Server installations, this was fixed in a previous PR by @labkey-jeckels, but the issue still needs to be addressed for Postgres. This PR updates the upgrade code to use ParameterMapStatement which should be better perf-wise.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3255

#### Changes
* ExperimentUpgradeCode: Use ParameterMapStatement in generateExpProtocolApplicationEntityIds

### DIFF
--- a/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
+++ b/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
@@ -247,7 +247,8 @@ public class ExperimentUpgradeCode implements UpgradeCode
                 }
             }
 
-            pm.executeBatch();
+            if (count > 0) pm.executeBatch();
+
             tx.commit();
         }
     }
--- a/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
+++ b/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
@@ -21,10 +21,12 @@ import org.jetbrains.annotations.NotNull;
 import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.DbScope;
 import org.labkey.api.data.DeferredUpgrade;
+import org.labkey.api.data.JdbcType;
+import org.labkey.api.data.Parameter;
+import org.labkey.api.data.ParameterMapStatement;
 import org.labkey.api.data.SQLFragment;
 import org.labkey.api.data.SqlExecutor;
 import org.labkey.api.data.SqlSelector;
-import org.labkey.api.data.Table;
 import org.labkey.api.data.TableInfo;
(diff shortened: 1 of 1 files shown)


labkey-alan:
Note: I have not yet tested this upgrade code, I will add another comment when I have.

labkey-matthewb (commented):
Nice.  Consider adding a pm.executeBatch() in the loop every few thousand rows.  I don't know what the practical JDBC limits are on total batch size, but even small batches like 1000 make a HUGE perf gain.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -222,16 +224,22 @@ public static void generateExpProtocolApplicationEntityIds(ModuleContext ctx) th
         {
             var protocolApplicationTable = ExperimentService.get().getTinfoProtocolApplication();
             List<ProtocolApplication> applications = new TableSelector(protocolApplicationTable).getArrayList(ProtocolApplication.class);
```
labkey-jeckels, line 226: It wasn't causing OutOfMemoryErrors with 1.4 million rows, but note that this pulls all the rows into memory at once. Should be OK to leave as-is unless we learn there are servers with 10+ million rows in this table.

labkey-jeckels (commented):
Will wait for the results of your testing but aside from doing batches like Matt suggested this is looking good.

labkey-jeckels:
Whatever changes happen here, they should get merged into 22.3 and not develop. Otherwise, only servers upgrading directly from 21.7 to 22.5 or later would benefit.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -222,16 +224,30 @@ public static void generateExpProtocolApplicationEntityIds(ModuleContext ctx) th
...
+
+                    if (count == 1000)
+                    {
+                        count = 0;
+                        pm.executeBatch();
+                    }
                 }
             }
 
+            pm.executeBatch();
```
labkey-matthewb:
```suggestion
            if (count > 0) pm.executeBatch();
```
I'm not sure whether calling executeBatch() with no rows is an error.

labkey-alan:
I'll cherry pick these commits into a 22.3 branch

labkey-alan:
Closing this PR in favor of one opened against 22.3:

https://github.com/LabKey/platform/pull/3261

PR #3260, opened 2022-04-19
Log errors if multiple elements have the same ID

labkey-tchad:
#### Rationale
Pages don't always define their element ids properly. We should flag when we encounter such problems in dev mode.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3192
* Example: https://github.com/LabKey/adjudication/pull/65

#### Changes
* Log event handler collisions


api/src/org/labkey/api/view/template/PageConfig.java
```diff
@@ -597,6 +604,30 @@ public void endOfBodyScript(Writer out) throws IOException
 
         // DOM CONTENT LOADED
 
+        if (AppProps.getInstance().isDevMode())
+        {
+            Map<String, Set<String>> handlers = new HashMap<>();
```
labkey-matthewb: Love the idea.  Might be able to simplify by just using a HashSet<String> handlers and if (!handlers.add(id+"#"+event)) {LOG}

labkey-matthewb (changes requested):

labkey-matthewb (approved):

PR #3261, opened 2022-04-19
Fix Issue 45275

labkey-alan:
#### Rationale
The upgrade code in question is causing perf problems for large SQL Server installations, this was fixed in a previous PR by @labkey-jeckels, but the issue still needs to be addressed for Postgres. This PR updates the upgrade code to use ParameterMapStatement which should be better perf-wise.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3255

#### Changes
* ExperimentUpgradeCode: Use ParameterMapStatement in generateExpProtocolApplicationEntityIds


labkey-alan:
I tested the upgrade code locally and it worked.

labkey-matthewb (approved):

labkey-jeckels (approved):

PR #3263, opened 2022-04-20
RunDatasetContextualRoles

labkey-matthewb:
#### Rationale
RunDatasetContextualRoles is not completely consistent with Dataset.hasPermission().  This fixes that.

#### Related Pull Requests
* 

#### Changes
*


labkey-jony (approved):
Installed test build on client's staging server and client confirmed that this is working.

PR #3272, opened 2022-04-20
Issue 45033: Missing / can't set folder descriptions (comments)

XingY:
#### Rationale
Folder dataFileUrl are saved without trailing slash. But canonicalURL for folders do end with /. Change made to query for dataFileUrl with and then without ending /. 

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1093

#### Changes
* query data using dataFileUrl with and then without ending /.
* update file browser "comment"label to "description".


XingY:
@josheckels Perhaps we should stop stripping off trailing "/" in ExpDataImpl.setDataFileURI?

labkey-jeckels (changes requested):
These trailing slashes are such a pain. I'd prefer to strip them off before doing the query instead of double-querying. 

One of the biggest headaches is that you get a different path depending on if the directory exists or not. If it exists, you get a trailing slash. If it doesn't, no trailing slash. Thus, it depends on the current state of the file system.

Additionally, we want to store them consistently in the DB. Thus, at upgrade time we'd have to figure out whether a given path in the DB is a directory or not, which of course we can only infer based on what's currently available on the file system.

XingY:
> These trailing slashes are such a pain. I'd prefer to strip them off before doing the query instead of double-querying.
> 
> One of the biggest headaches is that you get a different path depending on if the directory exists or not. If it exists, you get a trailing slash. If it doesn't, no trailing slash. Thus, it depends on the current state of the file system.
> 
> Additionally, we want to store them consistently in the DB. Thus, at upgrade time we'd have to figure out whether a given path in the DB is a directory or not, which of course we can only infer based on what's currently available on the file system.

Thanks for the details on the trailing slash. Given the inconsistence, probably leaving the stripping behavior as is makes most sense. I've updated implementation to strip off before querying.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -3874,7 +3874,14 @@ public List<ExpDataImpl> getAllExpDataByURL(File file, @Nullable Container c)
     @Override
     public List<ExpDataImpl> getAllExpDataByURL(String canonicalURL, @Nullable Container c)
     {
-        SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("DataFileUrl"), canonicalURL);
+        String dataFielUrl = canonicalURL;
```
labkey-jeckels: Typo - should be dataFileUrl, not dataFielUrl

labkey-jeckels (approved):

PR #3273, opened 2022-04-20
Export multi-tabbed grids to excel

Sigmonia:
#### Rationale
Item #10158: Multi-sheet export to Excel workbooks of multi-tab grids

#### Related Pull Requests
* [Components](https://github.com/LabKey/labkey-ui-components/pull/814)

#### Changes
* Added ExportTabsXlsxAction
* refactor ExcelWriter to simplify configuration


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -1672,6 +1672,118 @@ void _export(ExportQueryForm form, QueryView view) throws Exception
...
+            });
+
+            return queryViews;
+        }
+    }
+
+    @Marshal(Marshaller.Jackson)
+    @RequiresPermission(ReadPermission.class)
+    @Action(ActionType.Export.class)
+    public static class ExportTabsXLSXAction extends ReadOnlyApiAction<ExportTabsForm>
```
labkey-matthewb: I might nitpick about the name.  The "Tab"s bit is SM UI thing.  There are other uses that certainly have nothing to do with tabs.

labkey-matthewb (changes requested):
This seems like a pretty practical solution (e.g. more limited view specification than the full QueryForm).  I don't love the api Name.  Would love to see even a simple Junit smoke test.

api/src/org/labkey/api/data/ExcelWriter.java
```diff
@@ -508,13 +509,37 @@ public void write(OutputStream stream)
         }
     }
 
+    /**
+     * Renders the sheet then writes out the workbook to supplied stream
+     * @param response to write out the file
+     */
     public void write(HttpServletResponse response)
```
labkey-matthewb: With the refactor these write() methods are getting pretty overloaded/ambiguous.  Maybe rename with long descriptive names?  Like renderSheetAndWrite() etc?

Sigmonia: done, kept deprecated method as it is used in several other repos

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -1672,6 +1673,113 @@ void _export(ExportQueryForm form, QueryView view) throws Exception
...
+         * @param props
+         */
+        private void flattenProps(JSONObject props)
+        {
+            for(Map.Entry<String, Object> entry : props.entrySet())
+            {
+                if (entry.getValue() instanceof JSONArray)
+                {
+                    String key = entry.getKey();
+                    props.put(key, ((JSONArray) entry.getValue()).getString(0));
```
labkey-matthewb, line 1713: check length()==1?  (or at least length() > 0)

labkey-matthewb (changes requested):
Looks good.  I would still like to see one simple junit test with MockRequest and maybe two sheets (maybe just query core schema tables)

labkey-matthewb (approved):

PR #3276, opened 2022-04-21
45176: Resolving an issue always reassigns to the opener

labkey-klum:
#### Rationale
Resolved issues were always getting reassigned to the user who opened the issue, even if a different user was specified. This was a small bug fix and the problem was introduced with the merge of the API endpoint with the legacy UI code (the problem actually came from the API endpoint code)

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45176


labkey-jeckels (approved):
Might be worth adding a comment to beforeResolve() about its intended use

PR #3278, opened 2022-04-21
Fix study designer NPE

labkey-adam:
#### Rationale
Crawler threw garbage at the study designer and saw an NPE stack trace in the log.

Throw a `NotFoundException` instead. Also, get rid of unconditional logging of exceptions in `BaseRemoteService`, otherwise the crawler will balk at not found and other innocuous exceptions. `ExceptionUtil.logExceptionToMothership()` already logs the exceptions we want to see... no need to double log.


labkey-jeckels (approved):

labkey-adam:
@labkey-sweta This should fix the 22.3 `study.StudyProtocolDesignerTest` crawler failure you assigned to me. I'm assuming this is not critical to fix in 22.3 and we can just merge to develop. If I'm wrong, then we can backport.

PR #3284, opened 2022-04-22
44784: IssueAPI should handle missing priority param in validation

labkey-klum:
#### Rationale
Requests using the issue API can differ from those submitted from the UI in that the properties can be sparse. The previous validation assumed that all properties were specified in the request (even if they were null). This meant that we were missing some required checks during validation but it would be caught further down when we tried to save the issue to the DB (although this resulted in an error getting logged).

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44784

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/1094

#### Changes
- Handle validation for required fields even if they aren't specified in the request.

### DIFF
--- a/issues/src/org/labkey/issue/actions/IssueValidation.java
+++ b/issues/src/org/labkey/issue/actions/IssueValidation.java
@@ -44,11 +44,14 @@ import org.labkey.issue.model.IssueObject;
 import org.springframework.validation.Errors;
 import org.springframework.validation.MapBindingResult;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 /**
  * Created by klum on 2/3/2017.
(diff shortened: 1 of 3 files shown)


labkey-klum:
@labkey-bpatel tagging you for the CR, as I recently added you as the secondary dev feature owner for issues. 😄

issues/src/org/labkey/issue/actions/IssueValidation.java
```diff
@@ -92,9 +100,25 @@ public static void validateRequiredFields(@Nullable IssueListDef issueListDef, @
...
+                    {
+                        String msg = validator.validate(0, null);
+                        if (msg != null)
+                            requiredErrors.reject(SpringActionController.ERROR_MSG, msg);
+                    }
+                }
+            }
         }
-        if (newFields.containsKey("comment"))
-            validateRequired(user, "comment", String.valueOf(newFields.get("comment")), requiredFields, customColumnConfiguration, requiredErrors);
```
labkey-klum, line 97: Removing this code as it was never really doing anything. Comments can't be configured as required any longer

labkey-bpatel (approved):

PR #3291, opened 2022-04-27
Issue 45331: Create folder/project wizard combos to use tpl for option display

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=45331

#### Changes
* Update Ext4 combo options display tpl


labkey-adam (approved):
I was so close to having this fix myself...

PR #3295, opened 2022-04-27
Issue 45344: experiment-showMaterial.view fetches ProtocolApplications too often

labkey-jeckels:
#### Rationale
When a sample has derivations and other downstream usages, rendering its showMaterial page can take tens of seconds because we're firing off tens of thousands of redundant DB queries

#### Changes
* Don't query the exp.ProtocolApplication table based on the RowId, only to use the RowId value for downstream work
* Use a set instead of a list to track which runs have been visited


labkey-susanh (approved):
Nice.

PR #3299, opened 2022-04-28
Issue 45298: Optimize notification count retrieval

labkey-jeckels:
#### Rationale
When the notification menu is turned on, it adds a DB query to every page load. While 10-20ms extra isn't terrible, it's a lot of time to add to every single page.

#### Changes
* Cache unread notification count on a per-user basis
* Clear cache when notifications come and go

### DIFF
--- a/core/src/org/labkey/core/notification/NotificationServiceImpl.java
+++ b/core/src/org/labkey/core/notification/NotificationServiceImpl.java
@@ -25,6 +25,8 @@ import org.junit.Test;
 import org.labkey.api.action.SpringActionController;
 import org.labkey.api.admin.notification.Notification;
 import org.labkey.api.admin.notification.NotificationService;
+import org.labkey.api.cache.Cache;
+import org.labkey.api.cache.CacheManager;
 import org.labkey.api.data.CompareType;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
@@ -36,6 +38,7 @@ import org.labkey.api.data.Sort;
 import org.labkey.api.data.Table;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.TableSelector;
+import org.labkey.api.exp.api.ExpProtocol;
 import org.labkey.api.notification.NotificationMenuView;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.ValidationException;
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/notification/NotificationServiceImpl.java
```diff
@@ -82,6 +86,8 @@ private NotificationServiceImpl()
         ContainerManager.addContainerListener(this);
     }
 
+    /** Cache the number of unread notifications per user (key is a user ID) to avoid a DB query on every page load */
+    private final Cache<Integer, Long> _unreadCountCache = CacheManager.getCache(CacheManager.UNLIMITED, TimeUnit.HOURS.toMillis(1), "Unread notification counts");
```
labkey-matthewb: Consider using BlockingCache

core/src/org/labkey/core/notification/NotificationServiceImpl.java
```diff
@@ -175,21 +179,33 @@ public Notification addNotification(Container container, User user, @NotNull Not
...
         return getNotificationsByUserOrType(container, null, notifyUserId, unreadOnly);
     }
 
     @Override
-    public long getNotificationCountByUser(Container container, int notifyUserId, boolean unreadOnly)
+    public long getUnreadNotificationCountByUser(Container container, int notifyUserId)
     {
-        return createSelectorByUserOrType(container, null, notifyUserId, unreadOnly).getRowCount();
+        return _unreadCountCache.get(notifyUserId, null,
+                (k, a) -> createSelectorByUserOrType(container, null, notifyUserId, true).getRowCount());
```
labkey-matthewb: Doesn't address the fact that core.notifications doesn't have a useful index that matches the query generated by createSelectorByUserOrType()  e.g. (UserID, Type, ReadOn, Created).   I imagine that this query is the most common compared to insert/delete/update.

labkey-matthewb (approved):
This change is better than the current behavior.  I would also consider whether a change to the notifications table is warranted.

core/src/org/labkey/core/notification/NotificationServiceImpl.java
labkey-jeckels: Thanks. Updated, and I had missed that callers may pass in a Container (none currently do), so segmented the cache to handle that.

core/src/org/labkey/core/notification/NotificationServiceImpl.java
labkey-jeckels: I can split off a separate issue for adding an index. I'm hesitant to add indices without a good local test scenario to make sure it gets used

PR #3300, opened 2022-04-29
Issue 45238: NPE when importing sample type with auto-link study via pipeline job

labkey-jeckels:
#### Rationale
We want full-fidelity round tripping of sample types regardless of the import mechanism

#### Changes
* Don't assume we have a PipelineJob driving the import
* Don't assume we have an HTTP request with a URL when auto-linking samples to a target study
* Always populate xarJobId substitution, regardless of which constructor is used


labkey-jeckels:
@XingY I tagged you because your recent change in https://github.com/LabKey/platform/pull/3256 was related to one of the fixes I made here, related to the "XarJobId" substitution

XingY (approved):

PR #3303, opened 2022-04-29
Fix full index on upgrade

labkey-matthewb:
#### Rationale
When we fixed the indexer running too often, we broke the upgrade indexFull(force=true) case.

#### Related Pull Requests
*

### DIFF
--- a/search/src/org/labkey/search/SearchModule.java
+++ b/search/src/org/labkey/search/SearchModule.java
@@ -66,7 +66,7 @@ public class SearchModule extends DefaultModule
     @Override
     public Double getSchemaVersion()
     {
-        return 22.002;
+        return 22.003;
     }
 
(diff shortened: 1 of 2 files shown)


labkey-adam (approved):

search/src/org/labkey/search/SearchModule.java
```diff
@@ -66,7 +66,7 @@ public String getName()
     @Override
     public Double getSchemaVersion()
     {
-        return 22.002;
```
Sigmonia, line 69: is there a missing sql script? or is this so it kicks off the index?

labkey-adam: That's to force the reindex that needs to happen on any server that upgraded to 22.4

PR #3304, opened 2022-04-29
Issue 45374: LKSM issues when site "default display format for numbers" is set

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45374

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/826
* https://github.com/LabKey/platform/pull/3304
* https://github.com/LabKey/biologics/pull/1275
* https://github.com/LabKey/sampleManagement/pull/938

#### Changes
* add ImportAliasesDisplayColumnFactory override for getFormattedText()

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0",
       "dependencies": {
         "@labkey/api": "1.10.0",
-        "@labkey/components": "2.163.0-fb-numberFormat45374.0",
+        "@labkey/components": "2.164.0",
         "@labkey/themes": "1.2.1"
       },
       "devDependencies": {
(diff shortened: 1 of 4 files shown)


XingY (commented):
It wasn't clear to me why number format is affecting importAlias columns until I realize it's an alias column over rowId, which is an integer column. The code change should fix the format issue identified related to getQuery.api. The underlying issue is not addressed, that this column is recognized as a number field. 

The work around below should prevent other cases that this column might be treated as a number field. 

`
    private AliasedColumn createImportAliasColumn(String name, String prefix)
    {
        AliasedColumn aliasedColumn = new AliasedColumn(this, name, _rootTable.getColumn("RowId"))
        {
            @Override
            public boolean isNumericType()
            {
                return false;
            }
        };

`

cnathe:
> AliasedColumn aliasedColumn = new AliasedColumn(this, name, _rootTable.getColumn("RowId"))

Good suggestion. I reverted my previous change and applied this instead. Thanks, Xing.

XingY (approved):

PR #3305, opened 2022-04-29
Issue 45353: Cannot create a naming pattern reference (with walking the lookup) cross folder

XingY:
#### Rationale
Name expression validation is looking for lookup fields using current folder, instead of lookup container, resulting in not found error for lookup properties. 

#### Related Pull Requests
* 

#### Changes
* Use lookup container to get lookup table, not current container.


labkey-susanh (approved):

XingY:
Re-targeting 22.03
https://github.com/LabKey/platform/pull/3307

PR #3308, opened 2022-04-29
Issue 45347: Audit table data not available in linked schema

labkey-jeckels:
#### Rationale
It's effectively useless to share the auditLog schema via a linked schema as users won't have permission to see any of the data rows

#### Changes
* Special case the auditLog schema and grant CanSeeAuditLogRole in addition to the standard ReaderRole

### DIFF
--- a/query/src/org/labkey/query/LinkedSchema.java
+++ b/query/src/org/labkey/query/LinkedSchema.java
@@ -19,6 +19,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.labkey.api.audit.AbstractAuditTypeProvider;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
 import org.labkey.api.collections.CaseInsensitiveHashSet;
 import org.labkey.api.collections.CaseInsensitiveTreeSet;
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-adam @labkey-matthewb  curious what you think about this approach. It's a hack but doesn't feel terrible. We could get fancier in the future and let the linked schema definition itself declare the role(s) it should grant.

query/src/org/labkey/query/LinkedSchema.java
```diff
@@ -578,15 +580,25 @@ protected Map<String, Object> fireCustomizeParameterValues(LinkedTableInfo table
         return paramValues;
     }
 
+    private static Set<Role> getUserRoles(String sourceSchemaName)
```
labkey-matthewb: More descriptive name?  All LabKey methods are starting to sound like same to me (I know I am very guilty of this!)

Something ridiculous like this:  getContextualRoleForTargetSchema()

labkey-jeckels: That's not so ridiculous. I renamed and added a method-level comment

labkey-matthewb (changes requested):
This looks fine to me.  My meta concern is that this relies on a bunch of assumptions such as a) that only users who can grant CanSeeAuditLogRole can create LinkedSchema and b) The target container filter can't be circumvented. I think both of those are true, but I am wary of adding security exceptions if we don't assert/test/validate these assumptions.
Ideas?

labkey-jeckels:
@labkey-matthewb I like your thought process but don't have many suggestions. Both of those assumptions are correct at the moment.

InsertLinkedSchemaAction and EditLinkedSchemaAction require AdminOperationsPermission. LinkedSchemaUserWrapper enforces that only the source container's data is exposed. I don't have any good ideas for additional enforcement so I added a comment.

Note that we got push back (which I agree with) on our 21.11 change that enforced that the user who created the linked schema is still a site-level admin, as it broke linked schemas on servers where the site admin had later been demoted and/or deactivated.

labkey-matthewb (approved):

PR #3326, opened 2022-05-06
Throwing NotFound from a GWT method generates a server error.

labkey-matthewb:
#### Rationale
Of course the calling page still won't work, but stop generating  server error.

#### Related Pull Requests
* 

#### Changes
*


labkey-adam (changes requested):
This PR may have been forgotten. As we discussed, I didn't approve this because I had recently added this exception and updated the designer to correctly handle and display NotFoundExceptions, at least in my scenario. Worth understanding why your scenario wasn't handling this. My test URL was:

`http://localhost:8080/labkey/home/study-designer-designer.view?studyId=999&revision=999`

PR #3327, opened 2022-05-06
Split mothership sessions by calendar month

labkey-jeckels:
#### Rationale
Current mothership session tracking is based on how long a web server is running before it restarts. If a server restarts only to deploy a major release (like 21.11->22.3) it'll have a session that lasts four months. We want to ensure that we get at least month-level granularity so we can do more time-based reporting.

That also means we need to do a better job of separating out servers that are reporting with the same ServerGUID. Stashing the hostname and IP at the session level may help.

#### Changes
* Store hostname and IP at the session level as well as continuing to stash the most recent values at the installation level
* Split server sessions when it rolls into the next calendar month
* Chain together the sessions that are split so we can still tell their total duration if needed
* Introduce method variants to get non-null ColumnInfos and TableInfos
* Minor code cleanup

### DIFF
--- a/api/src/org/labkey/api/data/AbstractTableInfo.java
+++ b/api/src/org/labkey/api/data/AbstractTableInfo.java
@@ -532,6 +532,19 @@ abstract public class AbstractTableInfo implements TableInfo, AuditConfigurable,
         return getMutableColumn(colName, true);
     }
 
+    /** @return a BaseColumnInfo, will throw if column doesn't exist or exists and is locked */
+    @NotNull
+    public MutableColumnInfo getMutableColumnOrThrow(@NotNull String colName)
+    {
+        MutableColumnInfo result = getMutableColumn(colName, true);
+        if (result == null)
+        {
+            UserSchema schema = getUserSchema();
+            throw new IllegalArgumentException("Could not find column '" + colName + "' in " + (schema == null ? "" : (schema.getName() + ".")) + getName() + (schema == null ? "" : (" in " + schema.getContainer().getPath())));
+        }
+        return result;
+    }
+
     /**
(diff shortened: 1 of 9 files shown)


api/src/org/labkey/api/data/AbstractTableInfo.java
```diff
@@ -532,6 +532,19 @@ public MutableColumnInfo getMutableColumn(@NotNull String colName)
...
 
+    /** @return a BaseColumnInfo, will throw if column doesn't exist or exists and is locked */
+    @NotNull
+    public MutableColumnInfo getMutableColumnOrThrow(@NotNull String colName)
+    {
+        MutableColumnInfo result = getMutableColumn(colName, true);
+        if (result == null)
+        {
+            UserSchema schema = getUserSchema();
+            throw new IllegalArgumentException("Could not find column '" + colName + "' in " + (schema == null ? "" : (schema.getName() + ".")) + getName() + (schema == null ? "" : (" in " + schema.getContainer().getPath())));
```
ankurjuneja, line 543: I am curious between NotFoundException or IllegalArgumentException here.

ankurjuneja (approved):

api/src/org/labkey/api/data/AbstractTableInfo.java
labkey-jeckels: NotFoundException gets returned to the client as a 404 HTTP response code. It's appropriate for when the client sends a RowId as a GET parameter and we can't find that object. These don't get logged to mothership or written to the logs.

IllegalArgumentException is better for when it's way more likely to be a bug in the code. Here, the goal is to do better than NPE and give some context about what we failed to find. These get reported to mothership and written to the logs with a full stack trace.

PR #3336, opened 2022-05-09
Don't index files starting with '~'

Sigmonia:
#### Rationale
[Issue 45005: Exclude from search indexing any files that start with ~](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45005)

#### Changes
* Added check to indexer to ignore files starting with '~'

### DIFF
--- a/search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
+++ b/search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
@@ -1015,12 +1015,7 @@ public class LuceneSearchServiceImpl extends AbstractSearchService
     private boolean isWorkingFile(@NotNull WebdavResource r)
     {
         // MS Office opens temp/working files with '~', ignore these. Issue #45005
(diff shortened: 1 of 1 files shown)


search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
```diff
@@ -998,6 +998,10 @@ public boolean accept(WebdavResource r)
                     DocumentParser p = detectParser(r, null);
                     return p != null;
                 }
+                else if (isTempFile(r))
```
labkey-matthewb: Should probably move up with the isImage isZip checks, e.g. before we call getFileStream().

labkey-matthewb (changes requested):

search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
Sigmonia: done.

labkey-matthewb (approved):

PR #3339, opened 2022-05-10
Issue 45377: Poor performance determining required modules

labkey-jeckels:
#### Rationale
When a container has 25k+ child workbooks, calculating its required modules can be time consuming. And we do it multiple times on a single page render. Caching will provide a big speedup.

#### Changes
* Cache the required modules for a given container
* Clear the cache anytime the container tree or container properties change
* Minor assorted cleanup


api/src/org/labkey/api/data/Container.java
```diff
@@ -127,6 +133,70 @@ public class Container implements Serializable, Comparable<Container>, Securable
     private LockState _lockState = null;
     private LocalDate _expirationDate = null;
 
+    private final static Cache<GUID, Set<Module>> REQUIRED_MODULES_CACHE = new BlockingCache<>(
+            CacheManager.getCache(
+                    Constants.getMaxContainers(),
+                    CacheManager.DAY,
+                    "Required modules per containers"));
+
+    private final static CacheLoader<GUID, Set<Module>> REQUIRED_MODULE_LOADER =
```
labkey-adam: You could simply use the `Container` as the key

api/src/org/labkey/api/data/Container.java
```diff
@@ -127,6 +133,70 @@ public class Container implements Serializable, Comparable<Container>, Securable
     private LockState _lockState = null;
     private LocalDate _expirationDate = null;
 
+    private final static Cache<GUID, Set<Module>> REQUIRED_MODULES_CACHE = new BlockingCache<>(
+            CacheManager.getCache(
+                    Constants.getMaxContainers(),
+                    CacheManager.DAY,
+                    "Required modules per containers"));
```
labkey-adam: You could call `CacheManager.getBlockingCache()` and set the `CacheLoader` as part of the factory method. And then call `get(key, argument)`.

api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -2001,11 +1990,6 @@ public static MultiValuedMap<Container, Container> getContainerTree(Container ro
         MultiValuedMap<Container, Container> mm = new ArrayListValuedHashMap<>();
         mm.put(null, root);
         addChildren(root, mmIds, mm);
-        for (Container key : mm.keySet())
-        {
-            List<Container> siblings = new ArrayList<>(mm.get(key));
-            Collections.sort(siblings);
-        }
```
labkey-adam, line 2008: I guess we don't care about the sort any more?

labkey-adam (approved):
I offered a few totally optional suggestions. Let's hear it for 25K workbooks!

api/src/org/labkey/api/data/ContainerManager.java
labkey-jeckels: Unless I'm misreading the code, this never did anything. We create a copy of the list, sort it, and then throw it out.

api/src/org/labkey/api/data/Container.java
labkey-jeckels: I tried doing that originally, but there's no `get(key, argument)` version. Only `get(key)` and `get(key, argument, loader)` But if I switch to using the Container as the key, I can consolidate a bit.

api/src/org/labkey/api/data/ContainerManager.java
labkey-adam: Yep, agreed... I hadn't looked very closely at first

api/src/org/labkey/api/data/Container.java
labkey-adam: `Cache` doesn't have that variant, but `BlockingCache` does. Declare `REQUIRED_MODULES_CACHE` as a BlockingCache and voila!

PR #3347, opened 2022-05-11
Add "Clear All Objects" and "Reset" buttons to folder export page

labkey-adam:
#### Rationale
I'm tired of unchecking boxes on the export page. A simple "clear" button makes it much easier to export a single object type (or a small number of them).

image

#### Changes
* Add "Clear All Objects" and "Reset" buttons at the bottom of the folder/study export page
* Also, an unrelated refactor of to reduce `TSVGridWriter`'s reliance on holding a `Results` member variable

### DIFF
--- a/core/src/org/labkey/core/admin/exportFolder.jsp
+++ b/core/src/org/labkey/core/admin/exportFolder.jsp
@@ -92,7 +92,7 @@ Ext4.onReady(function(){
                 hideLabel: true,
                 boxLabel: parentName,
                 name: "types",
-                itemId: parentName,
+                itemId: parentName.replaceAll(',', ''),
                 inputValue: parentName,
                 checked: checked,
                 objectType: "parent"
@@ -108,7 +108,7 @@ Ext4.onReady(function(){
                         hideLabel: true,
                         boxLabel: childName,
                         name: "types",
-                        itemId: childName,
+                        itemId: childName.replaceAll(',', ''),
                         inputValue: childName,
                         checked: checked,
                         objectType: "child",
@@ -231,7 +231,7 @@ Ext4.onReady(function(){
                 text:'Reset',
                 handler: function(btn) {
                     document.getElementById('exportForm').innerHTML = '';
-                    initializeForm(initExportForm, <%=q(form.getExportType().toString())%>)
+                    initializeForm(initExportForm);
                 }
             }],
             buttonAlign:'left',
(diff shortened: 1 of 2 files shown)


core/src/org/labkey/core/admin/exportFolder.jsp
```diff
@@ -248,19 +261,24 @@ Ext4.onReady(function(){
...
+});
+
+function initializeForm(initExportForm, exportType)
+{
     LABKEY.Ajax.request({
         url: LABKEY.ActionURL.buildURL("core", "getRegisteredFolderWriters"),
         method: 'POST',
         jsonData: {
-            exportType: <%=q(form.getExportType().toString())%>
+            exportType: exportType
```
cnathe: I believe we can leave this unchanged instead of having to pass it into the new `initializeForm` function as a param

core/src/org/labkey/core/admin/exportFolder.jsp
```diff
@@ -219,6 +219,20 @@ Ext4.onReady(function(){
 
                     exportForm.getForm().submit();
                 }
+            }, {
+                text:'Clear All Objects',
+                handler: function(btn) {
+                    const leftColumnItems = exportForm.items.items[0].items.items;
```
cnathe, line 225: If only there were more `.items`...

cnathe (changes requested):
I pulled down this fb and did some local testing. It worked fine for the study export case, but for the folder export case it gave a JS error because it didn't like the comma in the "Experiments, Protocols, and Runs" value when it is used as an itemId.

Attached is a patch that should fix that part (the patch is in a zip since I can't attach a .patch file)
[exportFolder_changes.patch.zip](https://github.com/LabKey/platform/files/8673308/exportFolder_changes.patch.zip)
.

core/src/org/labkey/core/admin/exportFolder.jsp
labkey-nicka: Isn't this what component selectors are for?

labkey-adam: If I only knew what a component selector was...

core/src/org/labkey/core/admin/exportFolder.jsp
labkey-adam: Yes, agreed. I think I had the idea early on that I might allow switching between study and folder mode via some button, but dropped that idea

labkey-adam:
Thanks, Cory. That JavaScript error is an existing bug, right? I'm seeing it on labkey.org. (Happy to fix it either way.)

cnathe:
> Thanks, Cory. That JavaScript error is an existing bug, right? I'm seeing it on labkey.org. (Happy to fix it either way.)

Yes, it was unrelated to your changes. I just noticed it because of the clear all button scenario.

cnathe (approved):

PR #3348, opened 2022-05-11
Issue 44841: Support commas in names of samples and data class objects (and types)

labkey-susanh:
#### Rationale
Having commas in names of samples is, in some cases (such as pooled samples), a very natural thing to do. Because we also use the comma as a separate between multiple Ids in various places, we effectively don't support the use of commas in names. This work adds the ability for users to provide names of samples and data class objects with commas in them by quoting the values (as we do when importing other data on the server side).

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/834

#### Changes
* add utility method for unquoting a quoted identifier
* Use TabLoader to load the string of parents specified when deriving samples


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -718,7 +721,23 @@ else if (!_skipAliquot && _context.getInsertOption().mergeRows)
                         Collection<String> parentNames;
                         if (o instanceof String)
                         {
-                            parentNames = Arrays.asList(((String) o).split(","));
+                            try (TabLoader tabLoader = new TabLoader((String) o))
```
cnathe: similar to my other PR comment. I think this change from `split` to `TabLoader` could use a comment

cnathe (approved):

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -718,7 +721,26 @@ else if (!_skipAliquot && _context.getInsertOption().mergeRows)
                         Collection<String> parentNames;
                         if (o instanceof String)
                         {
-                            parentNames = Arrays.asList(((String) o).split(","));
+                            // Issue 44841: The names of the parents may include commas, so we parse the set of parent names
+                            // using TabLoader instead of just splitting on the comma.
+                            try (TabLoader tabLoader = new TabLoader((String) o))
+                            {
+                                tabLoader.setDelimiterCharacter(',');
+                                tabLoader.setUnescapeBackslashes(false);
```
cnathe: do we need this same line in the SsampleTimelineEvent.java usage or does that not apply there?

cnathe (approved):

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Good call.  Done.

PR #3352, opened 2022-05-12
Generate jars.txt for the api module

Sigmonia:
#### Rationale
[Issue 45124: Update remaining modules to generate credits jars.txt file](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45124)

#### Changes
* Update build.gradle to use ExternalDependency build utility
  * Remove reference to labkey-client-api
  * alphabetize dependencies
* Remove/ignore jars.txt from repo (now generated by build)


labkey-adam:
👍

Sigmonia:
@labkey-susanh Looks like the `version` column isn't included in the generated `jars.txt` file? do we need an update to the gradle plugin to include that column?

labkey-susanh:
The version number is in the name of the jar (almost always), so it was deliberately removed from the generated file since it's redundant (and has been a pain to update twice).  @labkey-ians Are you seeing a place where we need to show this?

api/build.gradle
```diff
@@ -159,18 +158,764 @@ dependencies {
...
-    }
-    external("net.sourceforge.jexcelapi:jxl:${jxlVersion}") {
-        // exclude this because it brings older version of log4j
-        exclude group: "log4j", module:"log4j"
-    }
-    external apache,caching,charting,gwt,jackson,javax,logging,others,r,spring
-
     // These are required for running unit tests through the server junit controller
-    external test
+//    external test
```
labkey-susanh: Should this be removed?  Did the dependencies get included below?

Sigmonia:
No it was just one of the diffs between the generated `jars.txt` and the previously checked in one.

labkey-susanh (approved):
I don't have time to do a thorough review of the jars.txt file and the dependency declarations, but the structure looks good.  Please tap someone else if there's any doubt there.  And make sure whatever suite it is that runs the CreditsTest gets run before you merge.

PR #3355, opened 2022-05-16
Rewrite limit rows implementations to correct behavior and share code

labkey-adam:
#### Rationale
When configuring offset with maxRows = ALL_ROWS, existing limit rows methods were generating invalid SQL on SQL Server and incorrect SQL on PostgreSQL. We also had dialects duplicating identical implementation.

[Issue 45482: TableSelector.getRowCount generates invalid SQL when using SQL Server](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45482)

#### Related Pull Requests
* https://github.com/LabKey/redshift/pull/31

#### Changes
* Introduce `LimitRowsSqlGenerator` to share standard implementation
* Shore up the `TableSelectorTestCase` with an additional test case (set an offset plus maxRows = ALL_ROWS) and factor out common test code
* Fix limit rows implementations and use shared code
* "master" -> "default"
* Optimize `getRowCount()` for zero rows and offset cases

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -1608,10 +1608,10 @@ public class DbScope
 
         LOG.info("Attempting to create database \"" + dbName + "\"");
 
-        String masterUrl = StringUtils.replace(url, dbName, dialect.getMasterDataBaseName());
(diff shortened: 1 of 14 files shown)


labkey-matthewb (approved):

labkey-alan (approved):
Looks good

PR #3363, opened 2022-05-17
Item 10299: Sample Finder Saving Queries

XingY:
#### Rationale
A new report type "SampleManager.SampleFinder" has been created to support saving Sample Finder filter cards. Currently, all views that a user has access to, including those that are hidden, will be displayed on Manage Views in LKS. We want to hide those hidden Sample Finder views from LKS and force user to manage them inside LKSM or LKB. A new "hideInManageViews" report property is added to hide those views from LKS Manage Views web part. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/837
* https://github.com/LabKey/platform/pull/3363
* https://github.com/LabKey/sampleManagement/pull/957
* https://github.com/LabKey/biologics/pull/1311

#### Changes
* Hide "hideInManageViews" reports from Manage Views web part


api/src/org/labkey/api/reports/report/ReportDescriptor.java
```diff
@@ -119,8 +119,9 @@ public enum Prop implements ReportProperty, SafeToRenderEnum
         json,
         serializedReportName,
         moduleReportCreatedDate, // creation date of module report, used by cds
-        showInDashboard, // used in Argos (show visible reports in the grid, show in my links if this is true)
-        showInParticipantView // boolean used in participant view for showing the saved charts for a dataset
+        showInDashboard, // show visible reports in the grid, show in my links if this is true or null
+        showInParticipantView, // boolean used in participant view for showing the saved charts for a dataset,
+        hideInManageViews // boolean used in to hide the hidden view from Managed Views. By default, hidden views are still visible in manage mode
```
cnathe: this is likely fine as is, but the other props are all "show" and this one is "hide". I assume this was done instead of having a "show in sample finder" to make this slightly more general?

cnathe:
```suggestion
        hideInManageViews // boolean used to hide the view from Managed Views. By default, hidden views are still visible in manage mode.

```

cnathe (approved):

PR #3369, opened 2022-05-20
Force reload of SMILES image on modify

XingY:
#### Rationale
When SMILES string is updated, the image content are redrawn while the filename stays the same. The stale structure image will be displayed due to browser cache in those cases. We want to add a cache-buster flag to force browser to reload the image. 

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3369
* https://github.com/LabKey/biologics/pull/1320

#### Changes
* Expose ExperimentController.DataClassAttachmentDownloadAction ExperimentUrls

### DIFF
--- a/api/src/org/labkey/api/exp/api/ExperimentUrls.java
+++ b/api/src/org/labkey/api/exp/api/ExperimentUrls.java
@@ -118,4 +118,6 @@ public interface ExperimentUrls extends UrlProvider
 
     default ActionURL getInsertMaterialQueryRowAction(Container c, TableInfo table) { return null; }
 
+    default ActionURL getDataClassAttachmentDownloadAction(Container c) { return null; }
+
 }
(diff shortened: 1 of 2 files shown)


labkey-tchad (commented):
Why do data classes need special handling for attachment columns? Wouldn't other attachment/file columns have the same problem.

XingY:
> Why do data classes need special handling for attachment columns? Wouldn't other attachment/file columns have the same problem.

True that other attachment will have same problem. And we are not fixing it for any data class attachment here either. The fix is scoped to SMIELS 2D structure image column. 

The argument is that it's an unlikely scenario that an attachment's content gets updated while the name stays the same and is attached to the same parent record (in this case, a specific exp.data). I was reluctant to make the change for SMILES 2D as well because I don't think it's a common scenario. But I guess it's still possible for user to have entered the wrong SMILES and then try to correct it so a new image is expected. 

As for general attachments, I'm inclined to wait until an actual client complains about stale thumbnails first because it's unlikely a scenario.

labkey-nicka (approved):

PR #3370, opened 2022-05-20
session cookie

labkey-matthewb:
#### Rationale
Remove old code that allows JSESSIONID for CSRF.  Add random header.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/internal/src/org/labkey/api/security/AuthFilter.java
+++ b/internal/src/org/labkey/api/security/AuthFilter.java
@@ -26,6 +26,7 @@ import org.labkey.api.security.impersonation.UnauthorizedImpersonationException;
 import org.labkey.api.settings.AppProps;
 import org.labkey.api.util.CSRFUtil;
 import org.labkey.api.util.ExceptionUtil;
+import org.labkey.api.util.GUID;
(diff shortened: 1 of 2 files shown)


labkey-adam (approved):

internal/src/org/labkey/api/security/AuthFilter.java
```diff
@@ -252,6 +255,17 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
     }
 
 
+    private void addRandomHeader(HttpServletRequest req, HttpServletResponse resp)
+    {
+        // make response size  a bit random (compressed or not)
+        StringBuilder sb = new StringBuilder(GUID.makeHash(req.getQueryString()));
+        Random r = new Random();
```
labkey-jeckels, line 262: This is probably sufficient, but should we use SecureRandom instead? If so, we may want to reuse an instance because they can be expensive to initialize (or at least used to be)

labkey-matthewb: makeHash() already has a 64 bits of SecureRandom goodness.  This is mostly just to make the length random.

PR #3372, opened 2022-05-20
Item 10201: enable sample renaming

XingY:
#### Rationale
Sample name column so far has not been editable and has been treated as an ID column. This PR enables name field edit for samples.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/841
* https://github.com/LabKey/platform/pull/3372
* https://github.com/LabKey/inventory/pull/436
* https://github.com/LabKey/labbook/pull/183
* https://github.com/LabKey/recipe/pull/66
* https://github.com/LabKey/sampleManagement/pull/962
* https://github.com/LabKey/biologics/pull/1327

#### Changes
* use DB seq instead of sample name for sample LSID
* add name and materialSourceId to sample tables, to allow Merge to use a alternative key other than LSID since LSIDs are no longer deterministic 
* new ObjectLegacyNames table to hold object's used names
* lookup samples using name rather than LSID during sample creation/import


experiment/resources/schemas/dbscripts/postgresql/exp-22.003-22.004.sql
```diff
@@ -0,0 +1,20 @@
...
+    Modified TIMESTAMP,
+    ModifiedBy INT,
+
+    CONSTRAINT PK_ObjectLegacyNames PRIMARY KEY (RowId)
+);
+
+ALTER TABLE exp.Material ADD COLUMN MaterialSourceId INT NULL;
+UPDATE exp.Material SET MaterialSourceId = (
+        SELECT distinct rowId FROM exp.materialSource WHERE materialSource.lsid = Material.cpastype
+    ) WHERE Material.cpastype <> 'Material';
```
labkey-susanh, line 18: When does `cpastype = 'Material'`? That seems somewhat incongruent.

experiment/src/org/labkey/experiment/DerivedSamplePropertyHelper.java
```diff
@@ -141,10 +150,12 @@ protected Lsid getObject(int index, @NotNull Map<DomainProperty, String> sampleP
...
             else
             {
-                lsid = _sampleType.generateSampleLSID().setObjectId(name).build();
+                lsid = _sampleType.generateNextDBSeqLSID().build();
             }
 
-            if (_lsids.containsValue(lsid) || ExperimentService.get().getExpMaterial(lsid.toString()) != null)
+            lsidName = new Pair<>(lsid, name);
+
+            if (_lsids.containsValue(lsidName) || ExperimentService.get().getExpMaterial(lsid.toString()) != null)
```
labkey-susanh: Should this be checking `getExpMeterial(name)`?  Seems like the lsid won't get duplicated since it's now using a DBSeq.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -252,4 +266,96 @@ public static void generateExpProtocolApplicationEntityIds(ModuleContext ctx) th
...
+
+        SchemaTableInfo provisionedTable = schema.getTable(domain.getStorageTableName());
+        if (provisionedTable == null)
+        {
+            LOG.error("Sample type '" + st.getName() + "' (" + st.getRowId() + ") has no provisioned table.");
+            return;
+        }
+
+        ColumnInfo nameCol = provisionedTable.getColumn("name");
+        if (nameCol == null)
```
labkey-susanh, line 320: When would this happen? (Just curious.)

experiment/src/org/labkey/experiment/LSIDRelativizer.java
```diff
@@ -102,7 +102,8 @@ else if (MATERIAL_PREFIX_PLACEHOLDER_SUFFIX.equals(lsid.getObjectId()))
             }
             else if (("Sample".equals(prefix) || "Material".equals(prefix)))
             {
-                return lsids.uniquifyRelativizedLSID("urn:lsid:" + XarContext.LSID_AUTHORITY_SUBSTITUTION + ":" + prefix + ".Folder-" + XarContext.CONTAINER_ID_SUBSTITUTION + ".${XarFileId}-" + lsids.getNextSampleId(), lsid.getObjectId(), lsid.getVersion());
+                String xarJobId = ".${XarJobId}"; // XarJobId is more concise than XarFileId
```
labkey-susanh, line 105: Changing this doesn't cause problems with parsing existing XAR files?

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -860,8 +889,10 @@ public SQLFragment getFromSQL(String alias, Set<FieldKey> selectedColumns)
         {
             for (ColumnInfo propertyColumn : provisioned.getColumns())
             {
-                // don't select lsid twice
-                if ("lsid".equalsIgnoreCase(propertyColumn.getColumnName()))
+                // don't select twice
+                if ("lsid".equalsIgnoreCase(propertyColumn.getColumnName())
+                    || "name".equalsIgnoreCase(propertyColumn.getColumnName())
+                    || "MaterialSourceId".equalsIgnoreCase(propertyColumn.getColumnName()))
```
labkey-susanh: Seems like the `Column` constants could be used here.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -1369,6 +1371,32 @@ public ExpDataClassImpl getDataClass(@NotNull Container c, @NotNull String dataC
         return getDataClass(c, null, false, dataClassName);
     }
 
+    public ExpDataClassImpl getDataClassByObjectId(Container c, Integer objectId)
+    {
+        SimpleFilter filter = SimpleFilter.createContainerFilter(c);
+        filter.addCondition(FieldKey.fromParts("ObjectId"), objectId);
+
+        DataClass dataClas = new TableSelector(getTinfoDataClass(), filter, null).getObject(DataClass.class);
```
```suggestion
        DataClass dataClass = new TableSelector(getTinfoDataClass(), filter, null).getObject(DataClass.class);
```

api/src/org/labkey/api/exp/SamplePropertyHelper.java
```diff
@@ -79,16 +80,27 @@ public void setDomainProperties(List<? extends DomainProperty> domainProperties)
 
     protected abstract ObjectType getObject(int index, @NotNull Map<DomainProperty, String> sampleProperties, @NotNull Set<ExpMaterial> parentMaterials) throws DuplicateMaterialException;
 
+    protected Pair<ObjectType, String> getObjectWithName(int index, @NotNull Map<DomainProperty, String> sampleProperties, @NotNull Set<ExpMaterial> parentMaterials) throws DuplicateMaterialException
+    {
+        return new Pair<>(getObject(index, sampleProperties, parentMaterials), null);
```
labkey-susanh, line 85: I'm not understanding why this is needed if the second argument is `null`.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -15,23 +15,32 @@
  */
 package org.labkey.experiment;
 
+import com.google.common.collect.Sets;
```
labkey-susanh, line 18: Is this the intended import?

experiment/src/org/labkey/experiment/LSIDRelativizer.java
labkey-susanh, line 105: Will this switch  to use `XarJobId` instead of `XarFieldId` cause any problems with parsing relativized LSIDs for XAR files that already exist?

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
labkey-susanh: Can the `Column` constants be used here?

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -656,6 +658,31 @@ public ExpMaterialImpl getSample(Container c, String name)
...
+        Material material = new TableSelector(ExperimentServiceImpl.get().getTinfoMaterial(), filter, null).getObject(Material.class);
+        if (material == null)
+            return null;
+        return new ExpMaterialImpl(material);
+    }
+
+    @Override
+    public ExpMaterial getEffectiveSample(Container c, String name, Date effectiveDate)
+    {
+        Integer legacyObjectId = ExperimentService.get().getObjectIdWithLegacyName(name, ExperimentServiceImpl.getNamespacePrefix(ExpMaterial.class), effectiveDate, c);
```
labkey-susanh, line 675: Not your change, but it's a bit confusing to me that "namespacePrefix" is actually the data type name.  Better naming (perhaps. dataTypeName?) would help, but is not necessary here.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -7958,6 +8021,63 @@ public List<QueryViewProvider<ExpRun>> getRunOutputsViewProviders()
...
+
+    @Override
+    public Integer getObjectIdWithLegacyName(String name, String dataType, Date effectiveDate, Container c)
+    {
+        TableInfo tableInfo = ExperimentService.get().getTinfoObjectLegacyNames();
+
+        // find the last ObjectLegacyNames record with matched name and timestamp
+        SQLFragment sql = new SQLFragment("SELECT ObjectId, Created FROM exp.ObjectLegacyNames " +
+                "WHERE Name = ? AND ObjectType = ? AND Created >= ? " +
+                "AND ObjectId IN (SELECT ObjectId FROM exp.Object WHERE Container = ?) " +
```
labkey-susanh, line 8042: Is the container filter necessary? I thought objectIds were globally unique.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -7958,6 +8021,63 @@ public List<QueryViewProvider<ExpRun>> getRunOutputsViewProviders()
...
+            Date nameEndTime = (Date) legacyNames[0].get("Created");
+            SQLFragment previousNameSql = new SQLFragment("SELECT Created FROM exp.ObjectLegacyNames " +
+                    "WHERE ObjectType = ? AND Created < ? " +
+                    "AND ObjectId IN (SELECT ObjectId FROM exp.Data WHERE Container = ?) " +
+                    "ORDER BY CREATED DESC");
+            previousNameSql.add(dataType);
+            previousNameSql.add(nameEndTime);
+            previousNameSql.add(c);
+
+            Map<String, Object>[] previouslegacyNames = new SqlSelector(tableInfo.getSchema(), previousNameSql).getMapArray();
```
```suggestion
            Map<String, Object>[] previousLegacyNames = new SqlSelector(tableInfo.getSchema(), previousNameSql).getMapArray();
```

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -344,6 +346,12 @@ protected Map<String, Object> _update(User user, Container c, Map<String, Object
         if (lsid == null)
             throw new ValidationException("lsid required to update row");
 
+        String newName = (String) row.get(Name.name());
+        String oldName = (String) oldRow.get(Name.name());
+        boolean hasNameChange = !StringUtils.isEmpty(newName) && !newName.equals(oldName);
+        if (hasNameChange && !NameExpressionOptionService.get().allowUserSpecifiedNames(c))
+            throw new ValidationException("User specified sample name not allowed");
```
labkey-susanh:
```suggestion
            throw new ValidationException("User-specified sample name not allowed");
```

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -641,6 +680,11 @@ else if (lsid != null)
                 lsidRowNumMap.put(lsid, rowNum);
                 rowNumLsid.put(rowNum, lsid);
             }
+            else if (name != null && materialSourceId != null)
+            {
+                sampleTypeId = materialSourceId;
+                nameRowNumMap.put(name, rowNum);
```
labkey-susanh, line 690: This is the reverse of the `lsidRowNumMap` (where the key is the `rowNum` and value is the `lsid`).  Is that intentional? Needed?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -540,13 +562,30 @@ public boolean next() throws BatchValidationException
                     {
                         if (_isSample)
                         {
-                            ExpMaterial sample = ExperimentService.get().getExpMaterial(lsid);
+                            ExpMaterial sample = null;
+                            if (_context.getInsertOption() == QueryUpdateService.InsertOption.MERGE && _nameCol != null)
+                            {
+                                Object nameValue = get(_nameCol);
+                                if (nameValue instanceof String)
```
labkey-susanh, line 569: When is this not a String? Just when it's null?  Would a null check be better?

experiment/resources/schemas/dbscripts/postgresql/exp-22.003-22.004.sql
XingY: Apparently a material doesn't have to be of a sample type, such as specimen. This is why I cannot add a FK on materialSourceId here.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
XingY: Probably never

experiment/src/org/labkey/experiment/LSIDRelativizer.java
XingY: It shouldn't. It should only impact new exports. This change is not needed for material but xarJobId is much more concise than guid based xarFileId so I went ahead with this switch.

api/src/org/labkey/api/exp/SamplePropertyHelper.java
XingY: This is bc there are multiple implementation of the abstract method getObject of this abstract class. But I only needed the 2nd argument for one of the implementation.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
XingY: Seems we have been using it in other places.

experiment/src/org/labkey/experiment/LSIDRelativizer.java
XingY: No, this only affects how export files are generated. Existing archives will still be translated using xarFileId, which is also fine.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: The container filter filters for ObjectIds that's associated with objects in this container only. We don't want to find a legacy name of material type in a sibling project.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: Hmm, lsidRowNumMap also has lsid as key and rowNum as value. Do you mean rowNumLsid? Yeah those maps are duplicated so allow easy lookup for different purposes.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Ah, of course. I misread that. Sorry.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
labkey-susanh: Ah yes. Thanks.

experiment/src/org/labkey/experiment/DerivedSamplePropertyHelper.java
XingY: Yep, good catch.

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: Since it's casted as String belong, I feel instanceof check is probably safer.

labkey-susanh (approved):

PR #3373, opened 2022-05-21
Issue 45519: Audit Log - Recording of folder export

labkey-jeckels:
#### Rationale
Folder exports are worth noting

#### Changes
* Create an audit log in "Project and Folder events" when someone initiates a folder export

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -15,7 +15,6 @@
  */
 package org.labkey.core.admin;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Joiner;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import org.apache.commons.beanutils.ConversionException;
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -4477,14 +4497,29 @@ public boolean handlePost(ExportFolderForm form, BindException errors) throws Ex
...
                 form.getFormat(), form.isIncludeSubfolders(), form.getExportPhiLevel(), form.isShiftDates(),
                 form.isAlternateIds(), form.isMaskClinic(), new StaticLoggerGetter(LogManager.getLogger(FolderWriterImpl.class)));
 
-            switch (form.getLocation())
+            AuditTypeEvent event = new AuditTypeEvent(ContainerAuditProvider.CONTAINER_AUDIT_EVENT, container.getId(), "Folder export initiated to " + exportOption.getDescription() + " " + (form.isIncludeSubfolders() ? "including" : "excluding") + " subfolders.");
+            if (container.getProject() != null)
+                event.setProjectId(container.getProject().getId());
+            AuditLogService.get().addEvent(getUser(), event);
+
+            switch (exportOption)
```
labkey-adam: Instead of a `switch`, why not push this code into per-enum methods?

labkey-adam (approved):
Would prefer the enum change, but not required

core/src/org/labkey/core/admin/AdminController.java
labkey-jeckels: I thought about it but chickened out as it more than doubled the work. With your nudge, I went ahead and moved it into the enum.

PR #3374, opened 2022-05-22
Biologics Registry Data Edit

labkey-martyp:
#### Rationale
Updates to facilitate editing and deleting Biologics Registry Datas in bulk and in grid, including lineage updates. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/840
* https://github.com/LabKey/platform/pull/3374
* https://github.com/LabKey/sampleManagement/pull/963
* https://github.com/LabKey/biologics/pull/1323

#### Changes
* Updates to ExpDataIterator to make exp.data lineages have the same behavior on update as samples
* Add GetDataOperationConfirmationDataAction to handle delete and update confirmation data


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -3330,30 +3329,46 @@ protected void deleteObjects(DeleteForm form)
     }
 
     @Marshal(Marshaller.Jackson)
-    @RequiresPermission(DeletePermission.class)
-    public class GetDataDeleteConfirmationDataAction extends ReadOnlyApiAction<DeleteConfirmationForm>
+    @RequiresPermission(UpdatePermission.class)
```
labkey-nicka: To be consistent with how `ExpMaterial` confirmation data is permissioned should this be `ReadPermission.class`?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -3330,30 +3329,46 @@ protected void deleteObjects(DeleteForm form)
...
+        public void validateForm(DataOperationConfirmationForm form, Errors errors)
         {
-            if (deleteForm.getDataRegionSelectionKey() == null && deleteForm.getRowIds() == null)
+            if (form.getDataRegionSelectionKey() == null && form.getRowIds() == null)
                 errors.reject(ERROR_REQUIRED, "You must provide either a set of rowIds or a dataRegionSelectionKey");
         }
 
         @Override
-        public Object execute(DeleteConfirmationForm deleteForm, BindException errors)
+        public Object execute(DataOperationConfirmationForm form, BindException errors)
```
labkey-nicka, line 3343: nit: Consider reusing the values of `form.getIds(false)` throughout this method.

```java
@Override
public Object execute(DataOperationConfirmationForm form, BindException errors)
{
    Collection<Integer> requestIds = form.getIds(false);
    List<ExpDataImpl> allData = ExperimentServiceImpl.get().getExpDatas(requestIds);

    List<Integer> notPermittedIds = new ArrayList<>();
    if (form.getDataOperation() == ExpDataImpl.DataOperations.Delete)
        notPermittedIds = ExperimentServiceImpl.get().getDataUsedAsInput(requestIds);

    return success(ExperimentServiceImpl.partitionRequestedOperationObjects(requestIds, notPermittedIds, allData));
}
```

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -995,58 +997,82 @@ private static String checkForLockedSampleRelativeChange(Set<ExpMaterial> previo
...
+            catch (BatchValidationException e)
+            {
+                throw new ExperimentException(e);
+            }
             ExpProtocolApplication outputApp = existingDerivationRun.getOutputProtocolApplication();
-            if (outputApp != null)
-                outputApp.removeMaterialInput(user, material);
-            existingSourceApp.removeMaterialInput(user, material);
+
+            if (runItem instanceof ExpMaterial)
```
labkey-nicka: nit: Consider using the instanceof call to cast the objects:

```java
if (runItem instanceof ExpMaterial material)
{
    if (outputApp != null)
        outputApp.removeMaterialInput(user, material);
    existingSourceApp.removeMaterialInput(user, material);
}
else if (runItem instanceof ExpData data)
{
    if (outputApp != null)
        outputApp.removeDataInput(user, data);
    existingSourceApp.removeDataInput(user, data);
}
```

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -995,58 +997,82 @@ private static String checkForLockedSampleRelativeChange(Set<ExpMaterial> previo
...
      */
     @NotNull
-    private static Pair<Set<ExpMaterial>, Set<ExpMaterial>> clearSampleSourceRun(User user, ExpMaterial material) throws ValidationException
+    private static Pair<Set<ExpMaterial>, Set<ExpMaterial>> clearRunItemSourceRun(User user, ExpRunItem runItem) throws ValidationException, ExperimentException
     {
-        ExpProtocolApplication existingSourceApp = material.getSourceApplication();
-        Set<ExpMaterial> previousSampleParents = Collections.emptySet();
-        Set<ExpMaterial> previousSampleChildren = Collections.emptySet();
+        ExpProtocolApplication existingSourceApp = runItem.getSourceApplication();
+        Set<ExpMaterial> previousParents = Collections.emptySet();
```
labkey-nicka: nit: This renaming seems a bit too generic. They're still `ExpMaterial`. Maybe consider `previousMaterialParents` and `previousMaterialChildren`?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -881,24 +881,26 @@ else if (o instanceof Number)
...
-                                }
                                 currentMaterialMap = new HashMap<>();
                                 currentMaterial = sample;
                                 currentMaterialMap.put(sample, sampleRole(sample));
                             }
                             else
                             {
                                 ExpData data = (ExpData) runItem;
-                                currentDataMap = Collections.singletonMap(data, dataRole(data, _user));
+                                currentDataMap = new HashMap<>();
```
labkey-nicka, line 902: In this case it will leave `currentMaterial` as `null` (which is fine). This result of this is a `null` value for the `aliquotChild` parameter to `record()` which was not previously a scenario that could be encountered. I would recommend a couple things:

1. Annotate the `record` method parameters with their supported nullable state. Something like:

```java
private static void record(
    boolean merge,
    @NotNull List<UploadSampleRunRecord> runRecords,
    @NotNull Map<ExpMaterial, String> parentMaterialMap,
    @NotNull Map<ExpMaterial, String> childMaterialMap,
    @NotNull Map<ExpData, String> parentDataMap,
    @NotNull Map<ExpData, String> childDataMap,
    @Nullable ExpMaterial aliquotParent,
    @Nullable ExpMaterial aliquotChild
)
```

2. Inside the implementation of `record()` I would add a null check before adding the aliquot child:

```java
if (record._aliquotInput != null && record._aliquotInput.equals(aliquotParent))
{
    if (aliquotChild != null)
        record._aliquotOutputs.add(aliquotChild);
    return;
}
```

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-martyp: Yeah I guess leaving it less restrictive just for confirmation data is ok. Updated

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-martyp: Updated

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-martyp: Good call. Changed

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-martyp: Yeah I was thinking those were going to be more generic when I renamed.  Added Material to name

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-martyp: That's a good catch.  Updated

labkey-nicka (approved):

PR #3376, opened 2022-05-24
Item 10374: Enable hiding of columns for session grid views in applications

labkey-susanh:
#### Rationale
We are implementing support for customizing grid views within our applications. This is the first story in this epic that adds the experimental feature flag check and, when enabled, allows users to hide columns in supported grids and revert the changes to their session views. We do not yet allow the newly update grid views to be saved. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/842
* https://github.com/LabKey/sampleManagement/pull/966

#### Changes
* Add experimental feature flag for grid customization in applications


query/src/org/labkey/query/QueryModule.java
```diff
@@ -221,6 +222,8 @@ public QuerySchema createSchema(DefaultSchema schema, Module module)
                 "The metadata is invalidated when performing actions such as creating a new List or modifying the columns on a custom view", false);
         AdminConsole.addExperimentalFeatureFlag(QueryServiceImpl.EXPERIMENTAL_SUBFOLDER_DATA_ENABLED, "Subfolder data in LabKey Products",
                 "View and interact with data from subfolders of a Product Project.", false);
+        AdminConsole.addExperimentalFeatureFlag(QueryView.EXPERIMENTAL_CUSTOMIZE_VIEWS_IN_APPS, "Customizing Grid view in LabKey Products",
+                "Customize grid views from within Biologics and Sample Manager applications.", false);
```
cnathe, line 226: minor: there are some grids in LKFM as well. I can't recall if those views would be customizable (for example a location/box page grid). I'll get to that PR next.

cnathe (approved):

query/src/org/labkey/query/QueryModule.java
labkey-susanh: Yes, sure. Given this is temporary, it doesn't seem worth updating the description, though.

PR #3380, opened 2022-05-25
ignore deadlock when updating lastIndexed

labkey-matthewb:
#### Rationale
Avoid unhandled deadlock exception

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/protein/ProteinService.java
+++ b/api/src/org/labkey/api/protein/ProteinService.java
@@ -18,6 +18,7 @@ package org.labkey.api.protein;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.action.QueryViewAction;
 import org.labkey.api.data.SimpleFilter;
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/protein/ProteinService.java
```diff
@@ -87,6 +88,8 @@ static void setInstance(ProteinService impl)
     /** Get seqId for the sequence -- there may be more than one if organism matches */
     List<Integer> getProteinSeqId(String sequence);
 
+    TableInfo getSequencesTable();
```
labkey-adam, line 91: What is this for?

labkey-adam (approved):

api/src/org/labkey/api/protein/ProteinService.java
labkey-matthewb: ??? I don't know.  That came from a merge.

PR #3387, opened 2022-05-26
Export script minor improvements

labkey-adam:
#### Rationale
Export script is a little behind the times. Also, https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=45567

#### Changes
* Escape Java strings in generated code
* Update exported Java code comments to match reality
* Adjust referrer parameter on help links
* Require CSRF token to export script
* Adjust loadLibrary.api to expand only resolved libraries


labkey-adam:
@labkey-nicka this includes the loadLibrary change we discussed. Please review that at a minimum.

api/src/org/labkey/api/action/QueryViewAction.java
```diff
@@ -107,6 +108,7 @@ else if (QueryAction.printRows.name().equals(form.getExportType()))
         }
         else if (QueryAction.exportScript.name().equals(form.getExportType()))
         {
+            CSRFUtil.validate(getViewContext());
```
labkey-nicka, line 111: Seems like a bit of a specific place to have to have this validation. Isn't this already being done? If not, should it be done for other export types? I'm coming out of left field on this one so take it for what its worth.

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -3365,9 +3366,16 @@ protected QuerySettings createQuerySettings(UserSchema schema)
...
-                result.setContainerFilterName(containerFilterType.name());
+                try
+                {
+                    ContainerFilter.Type containerFilterType = ContainerFilter.Type.valueOf(getContainerFilter());
+                    result.setContainerFilterName(containerFilterType.name());
+                }
+                catch (IllegalArgumentException e)
+                {
+                    // Remove bogus value from error message, Issue 45567
+                    throw new IllegalArgumentException("'containerFilter' parameter is not a valid org.labkey.api.data.ContainerFilter.Type enum constant");
```
labkey-nicka: nit: Consider resolving the package path of the class rather than hard coding it.

bigiron/src/org/labkey/bigiron/mysql/MySql80Dialect.java
```diff
@@ -36,4 +36,10 @@ protected Set<String> getReservedWords()
 
         return words;
     }
+
+    // MySQL 8.0 introduced EXPLAIN ANALYZE
+    protected String getExplainPrefix()
```
labkey-nicka, line 41: Ah yes, the MySql updates related to exporting scripts ... 😜

labkey-nicka (approved):

query/src/org/labkey/query/controllers/QueryController.java
labkey-adam: Okay... previous message (from Java) included the package, but I agree it's unnecessary for a user message

api/src/org/labkey/api/action/QueryViewAction.java
labkey-adam: Yes, it is specific to the export script case. Export options are GETs with no CSRF, in general, because it's not a mutating action. I'm adding the requirement here because scanners complain about our echoing of parameter values, even though we're clearly responding with plain text. My hope is a CSRF requirement will appease the scanners.

labkey-jeckels: We could also set the HTTP header to get the browser to download the file instead of rendering it directly as plain text.

api/src/org/labkey/api/query/QueryView.java
```diff
@@ -1427,6 +1428,7 @@ public void addExportScriptItems(PanelButton button)
                 if (null != url)
                 {
                     url.addParameter("scriptType", factory.getScriptType());
+                    url.addParameter(CSRFUtil.csrfName, CSRFUtil.getExpectedToken(getViewContext()));
```
labkey-jeckels: Are we OK transmitting CSRF as a GET parameter? It's more visible and leakier than POST and I've found guidance that recommends against it. https://brightsec.com/blog/csrf-token/

labkey-jeckels (changes requested):

bigiron/src/org/labkey/bigiron/mysql/MySql80Dialect.java
labkey-adam: Yeah, well...

api/src/org/labkey/api/query/QueryView.java
labkey-adam: Hoping @labkey-matthewb weighs in on that. I'd prefer a header, but I don't see how we could do that without major changes to the panel JSP. Might be a little easier to switch to POST, and of course we POST the token all the time.

api/src/org/labkey/api/action/QueryViewAction.java
labkey-adam: See other comment. I believe validate() needs to stay at this chokepoint regardless of the approach we use to convey the token, since many QueryViewAction subclasses render the export script results.

api/src/org/labkey/api/query/QueryView.java
labkey-adam: Switched to POST

labkey-jeckels: Shouldn't this parameter get dropped then?

labkey-adam: Picky picky

labkey-jeckels (approved):

PR #3390, opened 2022-05-27
Simplify wrapBuiltinColumns()

labkey-matthewb:
#### Rationale
Fix bug where column with dbsequence is present in input dataiterator

#### Related Pull Requests
* 

#### Changes
*


labkey-susanh (approved):
I'm not sure what the workbook test failures are about, but this does solve the issue we were seeing in labbook.

labkey-susanh (approved):

PR #3400, opened 2022-05-31
Support editable grid export for LabKey apps

Sigmonia:
#### Rationale
Item #10293: Export editable grids

#### Related Pull Requests
* (api-js)[https://github.com/LabKey/labkey-api-js/pull/130]
* (components)[https://github.com/LabKey/labkey-ui-components/pull/850]

#### Changes
* Added metric and audit logging to array based export methods


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -2442,6 +2451,15 @@ public void export(ConvertArraysToExcelForm form, HttpServletResponse response,
...
     }
 
+    public final static String EDITABLE_GRID_EXPORT_METRIC = "editableGridExport";
+    private void incrementEditableGridExportMetric()
+    {
+        PropertyManager.PropertyMap props = PropertyManager.getWritableProperties(ContainerManager.getRoot(), "metrics", true);
+        int count = Integer.valueOf(props.getOrDefault(EDITABLE_GRID_EXPORT_METRIC, "0")) + 1;
+        props.put(EDITABLE_GRID_EXPORT_METRIC, Integer.toString(count));
+        props.save();
+    }
```
cnathe: was there a specific reason you want with adding the `incrementEditableGridExportMetric` impl instead of going with the existing `ClientSideMetricManager`? Just curious if there was something missing from the `ClientSideMetricManager` implementation.

cnathe: you should be able to call it on the server side via something like `ClientSideMetricManager.get().increment(featureArea, metricName)`

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -2432,6 +2432,15 @@ public void export(ConvertArraysToExcelForm form, HttpServletResponse response,
                     response.setHeader("Content-disposition", "attachment; filename=\"" + filename + "\"");
                     ResponseHelper.setPrivate(response);
                     workbook.write(response.getOutputStream());
+
+                    JSONObject qInfo = rootObject.getJSONObject("queryinfo");
+                    if (qInfo != null)
+                    {
+                        QueryService.get().addAuditEvent(getUser(), getContainer(), qInfo.getString("schema"),
+                                qInfo.getString("query"), getViewContext().getActionURL(),
+                                "Exported editable grid to excel file: " + filename, null);
```
cnathe: hmm...feels a bit odd to me to have the message here be hard coded to "editable grid" since it is a generic API that doesn't know anything about the client side implementation/usage. Does it make sense to make this message generic and then possibly provide a param to the API for a more specific override?

cnathe (changes requested):

experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -609,6 +610,7 @@ public void containerDeleted(Container c, User user)
                 results.put("maxObjectObjectId", new SqlSelector(ExperimentService.get().getSchema(), "SELECT MAX(ObjectId) FROM exp.Object").getObject(Long.class));
                 results.put("maxMaterialRowId", new SqlSelector(ExperimentService.get().getSchema(), "SELECT MAX(RowId) FROM exp.Material").getObject(Long.class));
 
+                results.put("editableGridExports", PropertyManager.getProperty(PropertyManager.SHARED_USER, ContainerManager.getRoot(), "metrics", ExperimentController.EDITABLE_GRID_EXPORT_METRIC));
```
cnathe: FYI, if you take the approach mentioned below via `ClientSideMetricManager`, this can likely go away as the client side metrics will automatically get added to the metric JSON

cnathe: So can this be removed now?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -2442,6 +2453,7 @@ public void export(ConvertArraysToExcelForm form, HttpServletResponse response,
         }
     }
 
+    public final static String EDITABLE_GRID_EXPORT_METRIC = "editableGridExport";
```
cnathe: can this now be removed?

cnathe (approved):
Updates look good. But a few comments on possible cleanup to be done.

PR #3405, opened 2022-06-01
Item 10375: Save Grid Views - Save view dialog

XingY:
#### Rationale
1. As part of the epic to add view customization to the app grids, this PR adds support to allow saving custom views. Currently, saving a view with existing name always replaces the existing view. For the app grids, we are going to disallow such overrides. For LKS, we will add a confirm prompt before override/replace existing views.
2. Issue 43496: View names are not really case-insensitive

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/853
* https://github.com/LabKey/platform/pull/3405
* https://github.com/LabKey/inventory/pull/445
* https://github.com/LabKey/biologics/pull/1349
* https://github.com/LabKey/sampleManagement/pull/979

#### Changes
* Use case insensitive map for custom views
* Added replace param for SaveQueryViewsAction
* Prompt for confirm when user saves a custom view that already exists


internal/webapp/internal/ViewDesigner/Designer.js
```diff
@@ -247,6 +247,9 @@ Ext4.define('LABKEY.internal.ViewDesigner.Designer', {
                                 // Issue 13594: disallow setting inherit bit if query view has no available container filters
                                 o.inherit = containerFilterable && win.down('#inheritField').getValue();
                             }
+
+                            if (o.name && viewName && o.name.toLowerCase() !== viewName.toLowerCase())
```
cnathe: if I start on the default view and from the LKS save dialog, enter in a view name that already exists, it does not prompt to replace (looks like I only get the prompt if I start from a non-default view).

cnathe (approved):

PR #3413, opened 2022-06-02
Avoid HTML-encoding HTML on contextContent attributes

labkey-jeckels:
#### Rationale
We're HTML-encoding the "forgot password" link when a user is changing their own email address. We've got a couple of other places that are passing HTML as a string to contextContent.

#### Changes
* Use HtmlString to avoid showing raw HTML to user

### DIFF
--- a/core/src/org/labkey/core/user/changeEmail.jsp
+++ b/core/src/org/labkey/core/user/changeEmail.jsp
@@ -24,6 +24,9 @@
 <%@ page import="org.labkey.core.user.UserController.ChangeEmailAction" %>
 <%@ page import="org.labkey.core.user.UserController.DetailsAction" %>
 <%@ page import="org.labkey.core.user.UserController.UserForm" %>
+<%@ page import="org.labkey.api.util.HtmlString" %>
+<%@ page import="org.labkey.api.util.DOM" %>
+<%@ page import="org.labkey.api.util.HtmlStringBuilder" %>
 <%@ page extends="org.labkey.api.jsp.JspBase" %>
 <%
     String errors = formatMissedErrorsStr("form");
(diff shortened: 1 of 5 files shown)


core/src/org/labkey/core/view/styleGuide.jsp
```diff
@@ -276,7 +278,8 @@
         </div>
         <br>
         <h2>Context Content</h2>
-        <% String helpText = "Go to home: <a href=" + ActionURL.getBaseServerURL() + ">Server Home</a>";%>
+        <% HtmlString helpText = DOM.createHtml(DOM.createHtmlFragment("Go to home: ", DOM.A(DOM.at(DOM.Attribute.href, ActionURL.getBaseServerURL()), "Server Home"))); %>
```
labkey-adam, line 281: Could use `link()` here

core/src/org/labkey/core/user/changeEmail.jsp
```diff
@@ -60,7 +63,7 @@
     }
     else if (form.getIsPasswordPrompt())
     {
-        String resetPasswordLink = "<a href=" + h(urlFor(ResetPasswordAction.class)) + ">forgot password</a>";
+        HtmlString resetPasswordLink = DOM.createHtml(DOM.A(DOM.at(DOM.Attribute.href, urlFor(ResetPasswordAction.class)), "forgot password"));
```
labkey-adam: Could use `link()` here

labkey-adam (approved):
Optional suggestions

core/src/org/labkey/core/user/changeEmail.jsp
labkey-jeckels: Converted

core/src/org/labkey/core/view/styleGuide.jsp
labkey-jeckels: I fought with this a bit and gave up. link() needs a URLHelper, not a String URL, and I had to append the "Go to home:" prefix, so it didn't end up being any leaner. FWIW, I find it incredibly awkward to find the right way to nest Renderable vs Appendable vs builders vs HtmlString

PR #3397, opened 2022-05-30
45463: Update UI on manage study properties

labkey-klum:
#### Rationale
Simple text changes (including test automation), to address this ticket:

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45463

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/1120

PR #3417, opened 2022-06-06
update tests for link name change

labkey-klum:
#### Rationale
With the related PR to rename the link on the manage study page from `Change Study Properties` to `Study Properties`, the test was clicking on the wrong link because the folder name and manage link were now the same. Instead of fiddling with the selector, I just made the easy change to the folder name.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/3397


labkey-danield (approved):
Looks good.

PR #3418, opened 2022-06-06
Specify legacy-peer-deps flag for setup

labkey-nicka:
#### Rationale
For some unknown reason the `--legacy-peer-deps` flag has become necessary for our `npm ci` setup commands. It is possibly due to an update to the docker node version we use in TC inadvertently making this necessary for our use case (we use `npm install --legacy-peer-deps`). One interesting thing to note is as of 6/6/22 the `--legacy-peer-deps` flag is not a valid parameter to `npm ci` [according to their documentation](https://docs.npmjs.com/cli/v8/commands/npm-ci).

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3418
* https://github.com/LabKey/biologics/pull/1358
* https://github.com/LabKey/sampleManagement/pull/984
* https://github.com/LabKey/inventory/pull/447

#### Changes
* Change the `setup` command to `npm ci --legacy-peer-deps`. This is the command invoked by TeamCity.

### DIFF
--- a/assay/package.json
+++ b/assay/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "setup": "npm ci",
+    "setup": "npm ci --legacy-peer-deps",
     "build": "npm run build-dev",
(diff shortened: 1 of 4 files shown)


labkey-alan (approved):
fun

PR #3424, opened 2022-06-08
query.queryName -> queryName

labkey-matthewb:
#### Rationale
Most usages of QueryForm should support "queryName=" instead of "query.queryName=".  And at least some usages of "query.queryName" no longer work (see issue 45333).

#### Related Pull Requests
* 

#### Changes
*


labkey-matthewb:
Not sure this is right.

PR #3428, opened 2022-06-09
Don't include "Upload XAR" button if assay module is missing

labkey-tchad:
#### Rationale
Our link crawler found that the experiment begin page links to the assay XAR import page even if the assay module is not installed.
The first commit here just adds a module check before adding the XAR button. The second commit also moves the URL getter from `ExperimentUrls` to `AssayUrls`. If moving the URL getter isn't wanted, I can roll back the second commit.

```
java.lang.AssertionError: /MyStudies%20Test%20Organization/assay-chooseAssayType.view?tab=import
produced response code 404
Originating page: http://localhost:8111/labkey/MyStudies%20Test%20Organization/experiment-begin.view
  at org.junit.Assert.fail(Assert.java:89)
  at org.labkey.test.util.Crawler.crawlLink(Crawler.java:991)
  at org.labkey.test.util.Crawler.crawl(Crawler.java:823)
  at org.labkey.test.util.Crawler.crawlAllLinks(Crawler.java:786)
  at org.labkey.test.BaseWebDriverTest.checkLinks(BaseWebDriverTest.java:1458)
```

#### Changes
* Don't include XAR upload button on run group webpart if the Assay module is not present

### DIFF
--- a/api/src/org/labkey/api/assay/AssayUrls.java
+++ b/api/src/org/labkey/api/assay/AssayUrls.java
@@ -96,4 +96,5 @@ public interface AssayUrls extends UrlProvider
     ActionURL getSetDefaultValuesAssayURL(Container container, String providerName, Domain domain, ActionURL returnUrl);
     String getBatchIdFilterParam();
     ActionURL getPlateMetadataTemplateURL(Container container, AssayProvider provider);
+    ActionURL getUploadXARURL(Container container);
 }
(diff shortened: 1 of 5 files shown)


labkey-adam:
Exactly the way I would do it. In fact: https://github.com/LabKey/platform/pull/3427

Either would be fine, but my PR is approved and ready to merge

PR #3438, opened 2022-06-13
Fix flaky DataFinderTest "LABKEY.notifications[rowId] is undefined"

labkey-jeckels:
#### Rationale
Intermittent test failure in DataFinderTest looks like a race condition in NotificationViewAll.js:

https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_ImmportPostgres95/1855691?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true

#### Changes
* Check that we have a notification for the rowId in question

### DIFF
--- a/core/webapp/notification/NotificationViewAll.js
+++ b/core/webapp/notification/NotificationViewAll.js
@@ -192,7 +192,7 @@ Ext4.define('LABKEY.core.notification.NotificationViewAll', {
                                 if (notificationRec.get('ReadOn') == null)
                                 {
                                     notificationRec.set('ReadOn', new Date());
-                                    if (Ext4.isDefined(LABKEY.notifications))
+                                    if (Ext4.isDefined(LABKEY.notifications) && LABKEY.notifications[rowId])
                                         LABKEY.notifications[rowId].ReadOn = new Date();
                                 }
                             });


labkey-jeckels:
@cnathe this failure was assigned to me because I happened to be the sole commit where it failed, but it's an intermittent one predating my change. Looks like you did most of the NotificationViewAll.js implementation years ago. Does this look OK to you, or is it a sign of some other problem?

cnathe (approved):

cnathe:
> @cnathe this failure was assigned to me because I happened to be the sole commit where it failed, but it's an intermittent one predating my change. Looks like you did most of the NotificationViewAll.js implementation years ago. Does this look OK to you, or is it a sign of some other problem?

I was able to repro this JS error locally by going to the view all notifications page, clicking a single notification as "mark as read" and then clicking the "mark all as read" action for the page. This null check for the `LABKEY.notifications[rowId]` that you added fixed the error and looks correct. My only guess as to why this test failure is intermittent instead of always hit is because the DataFinderTest has a sleep a few lines before the failing check. I guess in some cases the API call must be taking just a bit longer in which case this JS error appears and is logged.

PR #3440, opened 2022-06-13
Issue 37040: Test Microsoft JDBC driver for SQL Server

### LINKED ISSUE #3439: Issue 37040: Microsoft SQL Server JDBC driver support
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=37040

- [x] Trivial local testing - 1 hr
- [x] Get driver onto Gradle classpath on TeamCity - 2 hr
- [x] Run test suites on TeamCity - 2 hr
- [x] Investigate and fix Daily failures - 5 hr
- [x] Investigate and fix ETL failures - 3 hr
- [x] Configure TeamCity to run with both jTDS and MS drivers - 3 hr @labkey-tchad 
- [x] Code review - 1 hr @labkey-matthewb 
- [x] Merge - 0.5 hr
- [x] Update docs - 3 hr
- [x] Fix `jdbcUrl` on Windows Agents - 2 hr
- [x] Update SQL Server 2014 on TeamCity agents
- [x] Investigate deadlocks

#### Hours
20.5 h

labkey-jeckels:
#### Rationale
The official Microsoft JDBC driver has some nice capabilities beyond jTDS, including higher availability.

#### Changes
* Pull in MS driver as a dependency
* Work around SQLServer timestamp equality problems by passing value as String, the only viable workaround for us
* Propagate DbScope into a few more places so we can conditionalize MS vs jTDS handling
* Convert more types to String before calling PreparedStatement.setObject()
* Work around stored procedure return value handling differences


api/src/org/labkey/api/data/dialect/StatementWrapper.java
```diff
@@ -887,7 +895,15 @@ public void setTimestamp(String parameterName, Timestamp x, Calendar cal)
     {
         try
         {
-             ((CallableStatement)_stmt).setTimestamp(parameterName, x, cal);
+            Object transformedValue = _conn.getScope().getSqlDialect().translateJdbcParameterValue(_conn.getScope(), x);
```
labkey-matthewb: I don't know if this is a problem, but this is the innerest loop of import.  It may be worth optimizing this a little.  At least cache DbScope and SqlDialect on the StatementWrapper.  I might go so far as to actually cache a java.lang.Function translateTimestamp (w/o typechecking) and java.lang.Function translateObject (w/ type checking).

labkey-matthewb: Another strategy might be to make the dialect a factory of StatementWrapper.  Then StatementWrapper could be subclassed w/ special behavior and even less indirection.

bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServer2016Dialect.java
```diff
@@ -15,17 +15,41 @@
...
 {
+    private static final FastDateFormat TIMESTAMP_FORMAT = FastDateFormat.getInstance("yyyy-MM-dd HH:mm:ss.SSS");
+
     public MicrosoftSqlServer2016Dialect(TableResolver tableResolver)
     {
         super(tableResolver);
     }
+
+    @Override
+    public Object translateJdbcParameterValue(DbScope scope, Object value)
```
labkey-matthewb: One problem is that we don't know the _target_ type.  If the target type is DATETIME2, we want even higher precision than this (e.g. important for incremental ETL).  I don't think we use DATETIME2, but I know Alan was considering it, and external databases might.  Should we format with enough digits to handle the DATETIME2 case even though it is not necessary for DATETIME?

labkey-matthewb (changes requested):

bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServer2016Dialect.java
labkey-jeckels: Switched to use Timestamp.toString() which includes the nanos

api/src/org/labkey/api/data/dialect/StatementWrapper.java
labkey-jeckels: Good thinking, and this was already supported on SqlDialect but I missed it when creating the first approach. Refactored.

labkey-matthewb (approved):

PR #3444, opened 2022-06-15
Item 10379: Save Grid Views - Manage views dialog

XingY:
#### Rationale
Renaming a customized view is not something that's supported in LKS, but we are enabling in in the apps. Instead of deleting then create a new view, uses calls 2 apis, this PR adds a new RenameQueryViewAction to allow renaming done in a single action. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/861
* https://github.com/LabKey/platform/pull/3444
* https://github.com/LabKey/sampleManagement/pull/999
* https://github.com/LabKey/biologics/pull/1374

#### Changes
* Added RenameQueryViewAction

### DIFF
--- a/query/src/org/labkey/query/controllers/QueryController.java
+++ b/query/src/org/labkey/query/controllers/QueryController.java
@@ -2428,6 +2428,102 @@ public class QueryController extends SpringActionController
         }
     }
 
+    public static class RenameQueryViewForm extends QueryForm
+    {
+        private String newName;
+
+        public String getNewName()
+        {
+            return newName;
+        }
+
+        public void setNewName(String newName)
+        {
+            this.newName = newName;
+        }
+
+    }
+
+    @RequiresPermission(ReadPermission.class)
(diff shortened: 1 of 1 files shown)


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -2428,6 +2428,102 @@ public ApiResponse execute(SimpleApiJsonForm form, BindException errors)
...
+                return new ApiSimpleResponse("success", true);
+        }
+    }
+
+    protected void renameCustomView(Container container, QueryDefinition queryDef, CustomView fromView, String newViewName, BindException errors)
+    {
+        if (newViewName != null && RESERVED_VIEW_NAMES.contains(newViewName.toLowerCase()))
+            errors.reject(ERROR_MSG, "The grid view name '" + newViewName + "' is not allowed.");
+
+        String newName = StringUtils.trimToNull(newViewName);
```
cnathe, line 2491: should we do this name trim before checking for reserved view names?

cnathe (approved):

PR #3447, opened 2022-06-15
Issue 45408: Option to export the results of Validate Queries

XingY:
#### Related Pull Requests
* 

#### Changes
* Added "Export" button to export validation result to excel

### DIFF
--- a/query/webapp/query/browser/view/Validate.js
+++ b/query/webapp/query/browser/view/Validate.js
@@ -354,7 +354,9 @@ Ext4.define('LABKEY.query.browser.view.Validate', {
 
         });
 
-        exportRows.push(...this.validationResult);
+        this.validationResult.forEach(result => {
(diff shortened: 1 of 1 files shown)


labkey-tchad:
This is currently breaking the build. Can't use ES6 `const` keyword in plain JavaScript files.
This is caused by YUICompressor in production builds.

XingY:
> This is currently breaking the build. Can't use ES6 `const` keyword in plain JavaScript files. This is caused by YUICompressor in production builds.

Thanks. Changed const to var

labkey-tchad:
The build is still broken. Not sure what else in this change isn't ES5 compatible

XingY:
> The build is still broken. Not sure what else in this change isn't ES5 compatible

Could be the array spread operation

labkey-matthewb (approved):

PR #3448, opened 2022-06-15
Issue 45634: Improve cross-folder lineage lookup support in name expressions

labkey-nicka:
#### Rationale
This addresses [Issue 45634](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45634) by passing through the target "data" container to the name generator. Currently, we do a mishmash of sometimes receiving the data container and other times receiving the container where the data type is defined. Additionally, this also updates Data Classes to be able to support lineage lookups and account for the scenario outlined in 45634.

There is also an adjacent scenario I was able to address which is lineage lookup values were failing cross-folder for non "built-in" columns specified on a data class domain. The reason for this failure is that the ExpData objects were not resolving their associated ExpDataClass due to not being provided with a user. As such, the NameGenerator would fail to resolve any properties for cross-folder data types.

#### Changes
* Pass in the "data" container in places that spawn a `NameGenerator`.
* Expose `getObjectProperties(@Nullable User user)` as part of the public interface for `ExpItem` to allow for providing user to an underlying instance retrieving properties.
* Consolidate creation of the name and aliquot name generators for Sample Types.

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
@@ -394,8 +394,28 @@ public class ExpSampleTypeImpl extends ExpIdentifiableEntityImpl<MaterialSource>
         return _object.getCategory();
     }
 
+    @NotNull
+    private NameGenerator createNameGenerator(@NotNull String expr, Container dataContainer)
+    {
+        Map<String, String> importAliasMap = null;
+        try
+        {
+            importAliasMap = getImportAliasMap();
+        }
+        catch (IOException e)
+        {
+            // do nothing
+        }
+
+        Container sampleTypeContainer = getContainer();
+        TableInfo parentTable = QueryService.get().getUserSchema(User.getSearchUser(), sampleTypeContainer, SamplesSchema.SCHEMA_NAME).getTable(getName());
+
+        Container nameGenContainer = dataContainer != null ? dataContainer : sampleTypeContainer;
+        return new NameGenerator(expr, parentTable, true, importAliasMap, nameGenContainer, getMaxSampleCounterFunction(), SAMPLE_COUNTER_SEQ_PREFIX + getRowId() + "-");
+    }
+
     @Nullable
-    public NameGenerator getNameGenerator()
+    public NameGenerator getNameGenerator(Container dataContainer)
     {
         if (_nameGen == null)
         {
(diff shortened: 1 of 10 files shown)


experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -394,8 +394,28 @@ public void setCategory(String category)
...
+        catch (IOException e)
+        {
+            // do nothing
+        }
+
+        Container sampleTypeContainer = getContainer();
+        TableInfo parentTable = QueryService.get().getUserSchema(User.getSearchUser(), sampleTypeContainer, SamplesSchema.SCHEMA_NAME).getTable(getName());
+
+        Container nameGenContainer = dataContainer != null ? dataContainer : sampleTypeContainer;
+        return new NameGenerator(expr, parentTable, true, importAliasMap, nameGenContainer, getMaxSampleCounterFunction(), SAMPLE_COUNTER_SEQ_PREFIX + getRowId() + "-");
```
XingY, line 414: The getMaxSampleCounterFunction() is still using sampleType.container always. This result in samples in child folder not included in determining the counter. getMaxSampleCounterFunction should be updated to use the correct container filter (project + child + shared). 

Issue repro:
In project folder, create an aliquot from S-1, which generates S-1-1.
In child folder, create another aliquot from S-1, generates S-1-2.
Go back to project folder, create another aliquot from S-1, expects S-1-3, but got duplicate key violation error.

XingY (changes requested):

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
XingY: Note that this is not strictly a new issue, but manifested with changes introduced in this branch. Prior to this branch, this function has been incorrect as well. But because _container is the sampleType container, the DB sequence also always fetched from the same container, making sure non-conflict seqNum is generated.

In addition to using the correct CF for getMaxSampleCounterFunction, I think another change is to always use the sampleType container for the DbSequenceManager.getPreallocatingSequence call in CounterExpressionPart, since sample names (counter) needs to be unique per sample type. See ExperimentServiceImpl.getLsidPrefixDbSeq for something similar that I had to do for SampleType DB Seqs.

labkey-nicka: > getMaxSampleCounterFunction should be updated to use the correct container filter (project + child + shared).

We could also not use a container filter at all here and rely on the `cpasType` filter to pull down what we expect. Here is the resulting query (we would not include the `cpastype` in the outer select ... just here for display purposes):

image

labkey-nicka: Addressed with 59aecd7.

XingY (approved):

PR #3458, opened 2022-06-17
Issue 45006: Provide link to lookup table/query in field designer

RosalineP:
#### Rationale
Usability improvement garnered by turning the lookup text display on a domain designer field into a direct link to the query.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/869
* https://github.com/LabKey/platform/pull/3458
* https://github.com/LabKey/biologics/pull/1386
* https://github.com/LabKey/sampleManagement/pull/1013

#### Changes
* Bump version


RosalineP:
@cnathe 
I'm a little stale on the deal with current frontend package updates—should I update, for good measure, other repos as well with this change?

cnathe:
> @cnathe I'm a little stale on the deal with current frontend package updates—should I update, for good measure, other repos as well with this change?

@RosalineP In this case since the field editor is used in the LKB and LKSM app UI, I would suggest that you bump the @labkey/components version for those two via the `biologics/package.json` and the `sampleManagement/package.json` so that TeamCity runs those suites to check for regressions.

core/package-lock.json
```diff
@@ -32,7 +32,6 @@
       "version": "2.1.2",
       "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.1.2.tgz",
       "integrity": "sha512-hoyByceqwKirw7w3Z7gnIIZC3Wx3J484Y3L/cMpXFbr7d9ZQj2mODrirNzcJa+SM3UlpWXYvKV4RlRpFXlWgXg==",
-      "dev": true,
```
cnathe: hmmm...did you run the npm install command with the legacy peer deps flag (i.e. `npm install --legacy-peer-deps`)? I would not have expected this many line changes to the package-lock.json file with just a components package bump.

cnathe (changes requested):

core/package-lock.json
RosalineP: Ah, I think I didn't! The package-lock is looking better now.

Branches also opened on LKB and LKSM!

cnathe (approved):

PR #3459, opened 2022-06-18
Refactor startup properties

### LINKED ISSUE #3487: Refactor startup properties
Migrate our startup property handling from maps with arbitrary String keys to structured handlers that document & validate valid property names. Provide a new "Available Startup Properties" admin page that lists all startup properties available on the current server.

- [x] Docker RStudio (2h) @labkey-adam 
- [x] Connectors (2h) @labkey-adam 
- [x] TrialServices (2h) @labkey-adam 
- [x] Add validation checks (1h) @labkey-adam
- [x] Code review (2h) @labkey-klum 
- [x] Merge (1h) @labkey-adam 

**Total**
10hrs

labkey-adam:
#### Rationale
Support validation and documentation of startup properties

#### Changes
Previously, any code path could retrieve a specific startup property scope any time it wanted to. Property names were all strings and the code had no idea what properties were available and which requested properties were valid.
This PR switches to a more structured model. In most cases, all valid properties for a given scope are specified as enum constants that implement a `getDescription()` method that documents each property. They must be registered and handled before module startup is complete, to ensure LabKey has the canonical list of available properties whenever it's requested. A new "Available Startup Properties" admin page lists all the valid properties and associated documentation.

A few implementation details:
* `StartupProperty`. Every startup property implements this interface, which defines `getDescription()`. In the enum case, each constant's `name()` is the default startup property name, though this can be altered by overriding `getPropertyName()`.
* `StartupPropertyEntry` is a bean that represents a single line in a properties file. It conveys the name, value, and modifier. (This class used to be called `ConfigProperty`.)
* `StartupPropertyHandler` represents a startup property scope. (Actually, multiple StartupPropertyHandlers can claim the same scope as long as they operate on a different set of StartupProperties. SiteSettings is one example of this.) StartupPropertyHandler does most of the heavy lifting. It provides the scope, a collection of StartupProperties, and the documentation properties. It also implements the `handle()` method that acts on the properties defined in its scope. All implementations extend one of the two StartupPropertyHandler flavors:
  * `StandardStartupPropertyHandler`: Most implementations extend this class. They advertise a fixed set of StartupProperties implemented as an enum. Their handle() methods receive a `Map<<T extends Enum<T> & StartupProperty>`, so the map is restricted to known properties. Many implementations enumerate the map and delegate handling to a method on the enum, though some interrogate the map directly via enum constants.
  * `LenientStartupPropertyHandler`: Extending this class supports scopes whose properties have ad hoc names, for example, an email address or a group name. This class holds a single StartupProperty, used to supply documentation. Its handle() method receives a `Collection<StartupPropertyEntry>`, since all property entries map to the same StartupProperty.
* Modules call `ModuleLoader.handleStartupProperties()` with a StartupPropertyHandler to register the handler (for documentation and validation purposes). The handler's `handle()` method is then called immediately. Most implementations save or otherwise process the startup properties at that point, although they have the option of stashing the properties for later use.


api/src/org/labkey/api/settings/StartupProperty.java
```diff
@@ -0,0 +1,22 @@
+package org.labkey.api.settings;
+
+import javax.annotation.Nullable;
+
+public interface StartupProperty
+{
+    // As a convenience for enum implementations, getStartupPropertyName() simply returns Enum.name()
```
labkey-klum: should this say `getPropertyName()`?

labkey-adam: Yes, I guess it should

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -2162,41 +2156,91 @@ public Map<String, ModuleContext> getUnknownModuleContexts()
...
+        handler.handle(map);
+    }
+
+    public <T extends StartupProperty> void handleStartupProperties(LenientStartupPropertyHandler<T> handler)
+    {
+        if (handler.performChecks())
+            startupPropertyChecks(handler);
+        StartupProperty sp = handler.getProperty();
+        handler.handle(handler.getStartupPropertyEntries().stream()
+            .peek(entry -> entry.setStartupProperty(sp))
```
labkey-klum, line 2182: It looks like we set the `StartupProperty` on the entry so we can check whether there is an unhandled startup property in the prop file. Is it possible, and should we also be checking to see if more than one instance of a `StartupProperty` is registering the same scope and enum?

labkey-klum (approved):
looks good, just a few questions added inline.

api/src/org/labkey/api/settings/StartupProperty.java
labkey-adam: Fixed in latest commit

api/src/org/labkey/api/module/ModuleLoader.java
labkey-adam: @labkey-klum I added a check for the same `StartupProperty` being used in multiple scopes; the check is in the `StartupListener`. This would indicate a coding problem (not an admin mistake), so I'm asserting in this case. You'd be hard-pressed to get into this situation accidentally, but I guess it's possible. Let me know if you had something else in mind.

PR #3462, opened 2022-06-20
Avoid complex query recursion

labkey-matthewb:
#### Rationale
Avoid stack overflow in cases of 'complex' query recursion.  Most simple cases are already handled.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/query/src/org/labkey/query/sql/Query.java
+++ b/query/src/org/labkey/query/sql/Query.java
@@ -727,13 +727,7 @@ public class Query
     // Query._depth handles most recursion, but there can be unexpected recursion caused by LinkedSchema for instance, or
     // other paths that cause a query to be compiled during resolveTable().  The thread local value makes sure this case
     // is handled as well.
-    ThreadLocal<AtomicInteger> resolveDepth = new ThreadLocal<>(){
-        @Override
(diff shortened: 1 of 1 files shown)


query/src/org/labkey/query/sql/Query.java
```diff
@@ -723,10 +724,24 @@ static void parseError(List<QueryException> errors, String message, @Nullable QN
 	}
 
 
+    // Query._depth handles most recursion, but there can be unexpected recursion caused by LinkedSchema for instance, or
+    // other paths that cause a query to be compiled during resolveTable().  The thread local value makes sure this case
+    // is handled as well.
+    ThreadLocal<AtomicInteger> resolveDepth = new ThreadLocal<>(){
```
labkey-jeckels: Shouldn't this be static, perhaps final and private too?

labkey-jeckels (changes requested):

query/src/org/labkey/query/sql/Query.java
labkey-matthewb: seems like a good idea

labkey-jeckels (approved):

PR #3466, opened 2022-06-21
retry SyncRunEdges.sync() on deadlock

labkey-matthewb:
#### Rationale
Avoid deadlocks that might cause us to not update the edges table.

#### Related Pull Requests
* 

#### Changes
*


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -3136,6 +3136,17 @@ SyncRunEdges invalidateClosureCache(boolean i)
 
         void sync(@Nullable Map<String, Integer> cpasTypeToObjectId)
         {
+            DbScope expScope = getExpSchema().getScope();
+            expScope.executeWithRetry((DbScope.RetryFn<Void>) tx -> {
+                sync_inner(cpasTypeToObjectId);
```
labkey-jeckels:
```suggestion
                syncInner(cpasTypeToObjectId);
```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -3136,6 +3136,17 @@ SyncRunEdges invalidateClosureCache(boolean i)
...
+                return null;
+            });
+        }
+
+        private void sync_inner(@Nullable Map<String, Integer> cpasTypeToObjectId)
```
labkey-jeckels:
```suggestion
        private void syncInner(@Nullable Map<String, Integer> cpasTypeToObjectId)
```

labkey-jeckels (approved):

PR #3479, opened 2022-06-23
Issue 45740: Ensure container filter context for insertion in product projects

labkey-nicka:
#### Rationale
This addresses [Issue 45740](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45740) by ensuring the correct container filter, `ContainerFilter.Types.CurrentPlusProjectAndShared`, is applied during insertion of rows in product projects. With this, the lookup validator is able to find the correct data to validate against when a field has been marked as "Ensure Value Exists in Lookup Target". This work covers Assays, Data Classes, Transform Tasks, Lists, and Sample Types.

**Opinion**
The crux of the implementation is that in order for a `TableInfo` and all of its subsequently generated bits (foreign key lookups, lookup validators, etc) to have the correct container scope it must be initialized with the expected Container Filter. This effectively means that the container filter to be used depends on _how_ the `TableInfo` is being used (read, insert/update, delete). I've attempted to encapsulate this in the notion of `schema.getTableForInsert()` where that implementation applies the container filter. This isn't ideal as it doesn't prevent someone from using just `schema.getTable()` when doing an insert. I could imagine having a `getTable` that takes something akin to a `QueryUpdateService.InsertOption` where it defaults to `READ` but allows the user to specify otherwise. That also doesn't enforce that the correct parameters are passed but does encapsulate everything in the same method(s) for hydrating a `TableInfo`. I'm open to other ideas as well.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3479
* https://github.com/LabKey/dataintegration/pull/171
* https://github.com/LabKey/biologics/pull/1396

#### Changes
* Introduce `QueryService.getProductProjectsInsertContainerFilter(Container)` which resolves the `ContainerFilter` to be used during insert/update of data in product projects.
* Call `getProductProjectsInsertContainerFilter` when spawning `TableInfo`s used for insertion.
* Miscellaneous nit improvements.

### DIFF
--- a/api/src/org/labkey/api/query/QueryService.java
+++ b/api/src/org/labkey/api/query/QueryService.java
@@ -615,24 +615,15 @@ public interface QueryService
         return col;
     }
 
-    default boolean isProductProjectsEnabled(Container container)
-    {
-        if (container == null || container.isRoot() || container.isWorkbook())
-            return false;
+    /**
+     * Resolves the ContainerFilter to be used during insert/update of data in product projects.
+     * Defaults to null if product projects are not enabled in container scope.
+     */
+    @Nullable
+    ContainerFilter getProductProjectsInsertContainerFilter(Container container, User user);
 
-        Container project;
-        if (container.isProject())
-            project = container;
-        else
-            project = container.getProject();
-
-        if (project == null)
-            return false;
-
-        // It is less than ideal to reference the folder type but for the time being
-        // this is how we recognize the appropriate container context.
-        // 1. The provided container is a LKB top-level LKB folder.
-        // 2. The provided container's top-level parent is a LKB folder.
-        return "Biologics".equals(project.getFolderType().getName());
-    }
+    /**
+     * Resolves if the product projects feature is enabled in the supplied container scope.
+     */
+    boolean isProductProjectsEnabled(Container container);
 }
--- a/api/src/org/labkey/api/query/QueryService.java
+++ b/api/src/org/labkey/api/query/QueryService.java
@@ -616,11 +616,11 @@ public interface QueryService
     }
 
     /**
-     * Resolves the ContainerFilter to be used during insert/update of data in product projects.
+     * Resolves the ContainerFilter to be used for lookups during insert/update of data in product projects.
      * Defaults to null if product projects are not enabled in container scope.
      */
     @Nullable
-    ContainerFilter getProductProjectsInsertContainerFilter(Container container, User user);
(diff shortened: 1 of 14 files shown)


experiment/src/org/labkey/experiment/api/property/LookupValidator.java
```diff
@@ -225,18 +224,20 @@ public Container getContainer()
...
+                            ValidatorContext validatorCache)
     {
         //noinspection ConstantConditions
         assert value != null : "Shouldn't be validating a null value";
 
         if (value != null)
             value = ConvertHelper.convert(value, crpField.getJavaObjectClass());
 
-        if (crpField instanceof PropertyDescriptor)
+        if (crpField instanceof PropertyDescriptor field)
```
labkey-susanh, line 239: Definitely appreciate this new syntax, Java.

labkey-susanh (approved):

api/src/org/labkey/api/query/QueryService.java
```diff
@@ -615,24 +615,15 @@ default MutableColumnInfo applyColumnTransformer(MutableColumnInfo col)
...
-
-        // It is less than ideal to reference the folder type but for the time being
-        // this is how we recognize the appropriate container context.
-        // 1. The provided container is a LKB top-level LKB folder.
-        // 2. The provided container's top-level parent is a LKB folder.
-        return "Biologics".equals(project.getFolderType().getName());
-    }
+    /**
+     * Resolves the ContainerFilter to be used during insert/update of data in product projects.
+     * Defaults to null if product projects are not enabled in container scope.
```
labkey-jeckels, line 620: Can you add to the comment and explain why we want a filter other than the current folder? I'm used to the ContainerFilter controller the WHERE clause for read operations and am not understanding how it impacts the insert. If there are multiple containers within the scope, how does it choose which one to target?

labkey-jeckels: Rereading the PR description, is this primarily to control the lookups from this table and their validations?

labkey-jeckels (changes requested):

api/src/org/labkey/api/query/QueryService.java
labkey-nicka: I've renamed this method to `getContainerFilterForLookups` and updated the associated comment.

labkey-jeckels (approved):

PR #3484, opened 2022-06-24
Issue 45376: Migrate from Universal Analytics to Google Analytics 4

labkey-jeckels:
#### Rationale
Google has deprecated "Universal" analytics, which is the only variant we support directly. GA4 is the version. Sites are also interested in being able to configure multiple styles of reporting for a transition period.

#### Changes
* Add basic support for enabling and reporting to GA4
* Allow multi-select across GA4, Universal, and custom
* Disallow selecting both Universal variants
* A few more passthroughs for the checkbox tagline
*
Screen Shot 2022-06-24 at 3 11 37 PM

### DIFF
--- a/core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
+++ b/core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
@@ -68,7 +68,7 @@ public class AnalyticsServiceImpl implements AnalyticsService
                         return "";
                     }
                 },
-        /** Use GA with URL sanitization */
+        /** Use old GA with URL sanitization */
         enabled(true)
                 {
                     @Override
@@ -77,7 +77,7 @@ public class AnalyticsServiceImpl implements AnalyticsService
                         return TRACKING_SCRIPT_TEMPLATE_ASYNC;
                     }
                 },
-        /** Use GA without replacing container paths */
+        /** Use old GA without replacing container paths */
         enabledFullURL(true)
                 {
                     @Override
@@ -86,6 +86,7 @@ public class AnalyticsServiceImpl implements AnalyticsService
                         return TRACKING_SCRIPT_TEMPLATE_ASYNC;
                     }
                 },
+        /** Use GA4 with the full URL */
         ga4FullUrl(true)
                 {
                     @Override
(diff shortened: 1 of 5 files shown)


core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
```diff
@@ -195,7 +221,7 @@ private String getSanitizedUrl(ViewContext context)
         Container container = context.getContainer();
 
         // Adding a null check for container as on rendering the error page, container can be null for a not found page
-        if (getTrackingStatus() != TrackingStatus.enabledFullURL && null != container && !container.hasPermission(UserManager.getGuestUser(), ReadPermission.class))
+        if (getTrackingStatus().contains(TrackingStatus.enabledFullURL) && null != container && !container.hasPermission(UserManager.getGuestUser(), ReadPermission.class))
```
labkey-susanh: Should this be `!contains`?

core/src/org/labkey/core/analytics/analyticsSettings.jsp
```diff
@@ -31,25 +31,49 @@
     boolean hasAdminOpsPerms = getContainer().hasPermission(getUser(), AdminOperationsPermission.class);
 %>
 <%=getTroubleshooterWarning(hasAdminOpsPerms, HtmlString.unsafe("<br>"))%>
-<p>Your LabKey Server can be configured to add JavaScript to your web site so that usage information will be sent to a web analytics system.</p>
+<p>You can configure your LabKey Server to include JavaScript that reports usage information to a web analytics system.</p>
 
 <labkey:errors/>
 <labkey:form action="<%=new ActionURL(AnalyticsController.BeginAction.class, ContainerManager.getRoot())%>" method="POST">
-    <table>
+    <table style="width: 60em;">
```
labkey-susanh, line 38: We might prefer css classes here, but I see you're just following what was done before.

labkey-susanh (approved):

core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
labkey-jeckels: Thanks. This was indeed messed up. Fixed.

core/src/org/labkey/core/analytics/analyticsSettings.jsp
labkey-jeckels: I believe I reduced the total in-line style attributes on the page, but stopped short of eradicating.

PR #3490, opened 2022-06-27
optimize RelativeDateVisitManager._updateVisitRowId()

labkey-matthewb:
#### Rationale
Performance improvement for date relative studies
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45404

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/study/src/org/labkey/study/visitmanager/RelativeDateVisitManager.java
+++ b/study/src/org/labkey/study/visitmanager/RelativeDateVisitManager.java
@@ -261,11 +261,16 @@ public class RelativeDateVisitManager extends VisitManager
 
         Study study = getStudy();
         Study visitStudy = StudyManager.getInstance().getStudyForVisits(study);
+        // NOTE: visitStudy may not equals this.getStudy() in the case of a database study where the visits may be shared
         var visitStudyVisitManager= StudyManager.getInstance().getVisitManager(visitStudy);
 
-        // Joining from ParticipantVisit to Visit using a BETWEEN is very expensive
-        // So expensive that we're going to do the join in memory instead
+        // Joining from ParticipantVisit to Visit using a BETWEEN is very expensive.
+        // So expensive that we're going to do the join in memory instead.
         // https://www.labkey.org/home/Developer/issues/issues-update.view?issueId=45404
+        //
+        // Alternatives to in-memory join a) precompute and maintain a table with this mapping
+        // b) do the JOIN on the server but use a (SELECT DISTINCT sequencenum FROM participantvisit) CTE instead of directly
+        // joining all rows in ParticipantVisit to Visit.
         StringBuilder mapDayToRowId = new StringBuilder();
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
Checking this study, it has 8,015 rows in study.Visit. There are a few nonsense Visit Days like 'Day -657433' because there are some bogus dates from the year 1111 in datasets, but all but about 20 of the rows range from -10000 to +5000 which is reasonable for a study with animals that live for a couple of decades.

Passing that many days through the SQL seems risky. I assume this study is beyond what you were anticipating with the approach?

labkey-matthewb:
> 

Not necessarily using values should be pretty scalable especially since it is just sending pairs of INT.  I think you can do very large INSERT FROM VALUES for instance.  I don't think this is similar to the large URL problem for instance.  I can google a bit to see what the SQL size limit might be.

labkey-matthewb:
I think we're many orders of magnitude from hitting a hard limit or even a practical limit (rough estimate is that this would be a 128K query).  The two alternatives are create a temp table, but don't think that would be faster than inlining the temp table OR doing a (SELECT DISTINCT ... JOIN .. BETWEEN) CTE to make the expensive JOIN smaller, then doing the big join between PV and the CTE.  The BETWEEN join will still be bad order of complexity-wse, we're just knocking down the N a bit.

The only other idea I have is to introduce another computed table sequencenum to visit.rowid.  That's what PV really started as, but there could be a more granular table (ugh).

study/src/org/labkey/study/visitmanager/RelativeDateVisitManager.java
```diff
@@ -256,22 +258,34 @@ private void _updateVisitRowId()
...
-                " WHERE ParticipantVisit.Day BETWEEN v.SequenceNumMin AND v.SequenceNumMax AND\n" +
-                "   v.Container = ?\n" +
-                " ),-1)\n";
-        sqlUpdateVisitRowId += "WHERE Container = ?";
 
         Study study = getStudy();
         Study visitStudy = StudyManager.getInstance().getStudyForVisits(study);
-        Container c = visitStudy.getContainer();
-        new SqlExecutor(schema).execute(sqlUpdateVisitRowId, c, study.getContainer());
+        var visitStudyVisitManager= StudyManager.getInstance().getVisitManager(visitStudy);
```
labkey-jeckels, line 265: Not all that important, but won't this always return the same thing as `this` since we're already in RelativeDateVisitManager?

labkey-jeckels (approved):

study/src/org/labkey/study/visitmanager/RelativeDateVisitManager.java
labkey-matthewb: Surprisingly, no.  In a dataspace study the project can hold the visits depending on how it is set up.

PR #3499, opened 2022-06-30
Issue 44797: Update data alignment with custom and standard participant views

RosalineP:
#### Rationale
When you customize the participant view in a study, our default code is showing a particularly wide first column. By default, the first column should be within a reasonable width for readability.

#### Related Pull Requests
* N/A

#### Changes
* Update styling

### DIFF
--- a/study/src/org/labkey/study/view/participantAll.jsp
+++ b/study/src/org/labkey/study/view/participantAll.jsp
@@ -239,6 +239,11 @@
     .labkey-ptid-remove, .labkey-ptid-add {
         cursor: pointer;
     }
+
+    .labkey-data-region tr td:first-child
+    {
+        width: 300px
+    }
 </style>
 
 <script type="text/javascript" nonce="<%=getScriptNonce()%>">
--- a/study/src/org/labkey/study/view/participantAll.jsp
+++ b/study/src/org/labkey/study/view/participantAll.jsp
@@ -242,7 +242,7 @@
 
     .labkey-data-region tr td:first-child
     {
-        width: 300px
+        width: 300px;
     }
 </style>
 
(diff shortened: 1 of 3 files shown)


RosalineP:
Cory is the lucky individual chosen for review, as he touched the adjacent styling most recently!

labkey-mohara (approved):
Looks great

study/src/org/labkey/study/view/participantAll.jsp
```diff
@@ -239,6 +239,11 @@
     .labkey-ptid-remove, .labkey-ptid-add {
         cursor: pointer;
     }
+
+    .labkey-data-region tr td:first-child
+    {
+        width: 300px;
```
cnathe, line 245: This code change looks good and makes sense. I pulled down the fb to test it, and it worked good for the default case. However, if I changed the dataType from ALL to DEMOGRAPHIC, this column width was no longer getting applied. I'm not sure why. Thoughts?
Screen Shot 2022-07-04 at 9 46 03 AM

cnathe (approved):

study/src/org/labkey/study/view/participantAll.jsp
RosalineP: Ah, great catch! The page was switching views between the participantAll.jsp and customizeParticipantView.jsp, the latter of which the style was not in. Update made, and the fix is now working across cases.

RosalineP: Great catch! The customized participant view was stomping over the changes. Newest commit works for all cases (except the case where a client has a pre-existing customized participant view, in which case that custom view is preserved.

PR #3502, opened 2022-07-02
Avoid SQL Server deadlock on pipeline.statusfiles

labkey-jeckels:
#### Rationale
Since switching to the Microsoft JDBC driver, we're seeing more frequent test failures with DB deadlocks on SQL Server. The problem is that the pipeline.statusfiles table is being queried and updated on different threads (background pipeline and HTTP request) and its multiple unique constraints (two plus a PK) are prone to taking locks in the wrong order across updates and reads.

TargetedMS and Flow tests were the most common to deadlock, and have passed a number of times on this FB.

#### Changes
* Update only the fields that have changed, meaning that most updates won't touch any of the indices
* Simplify common query against the table

### DIFF
--- a/pipeline/src/org/labkey/pipeline/api/PipelineStatusFileImpl.java
+++ b/pipeline/src/org/labkey/pipeline/api/PipelineStatusFileImpl.java
@@ -32,13 +32,14 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 /**
  * @author B. MacLean
  */
 public class PipelineStatusFileImpl extends Entity implements Serializable, PipelineStatusFile
 {
-    private static HashSet<String> _emailStatuses = new HashSet<>(Arrays.asList(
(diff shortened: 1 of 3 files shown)


pipeline/src/org/labkey/pipeline/api/PipelineStatusFileImpl.java
```diff
@@ -32,13 +32,14 @@
...
 /**
  * @author B. MacLean
  */
 public class PipelineStatusFileImpl extends Entity implements Serializable, PipelineStatusFile
 {
-    private static HashSet<String> _emailStatuses = new HashSet<>(Arrays.asList(
+    private static final HashSet<String> _emailStatuses = new HashSet<>(Arrays.asList(
             PipelineJob.TaskStatus.complete.toString(),
             PipelineJob.TaskStatus.error.toString(),
             PipelineJob.TaskStatus.cancelled.toString()
```
labkey-adam: Delete this? `isEmailStatus()` isn't called.

labkey-adam (approved):
NIce

PR #3503, opened 2022-07-04
Issue 45802: Remove double encoded from project/folder names in Admin Console > Files > File Directories list box

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45802

Note that we don't need to use Ext4.util.Format.htmlEncode() here in the Ext4 TreePanel root display because of the override added to ext-patch.js (see `Ext4.override(Ext4.tree.Column, {...})` which now does the encoding for us.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3469

#### Changes
* Remove `treecolumn` renderer

### DIFF
--- a/core/src/org/labkey/core/admin/view/filesSiteSettings.jsp
+++ b/core/src/org/labkey/core/admin/view/filesSiteSettings.jsp
@@ -165,9 +165,6 @@
                     header:'Project',
                     width:330,
                     dataIndex:'name',
-                    renderer: function(val){
-                        return Ext4.util.Format.htmlEncode(val);
-                    }
                 },{
                     header:'Directory',
                     width:420,


labkey-susanh (approved):

labkey-adam (approved):

labkey-adam:
@cnathe should we proactively search for other Ext4 tree components that haven't been updated to accommodate the patch override?

cnathe:
> 

@labkey-adam I did. It is a little hard to search for usages, but I found a few others and took a look to see if they needed the same change or not. Nothing else that I found at the moment.

cnathe:
@labkey-tchad can you think of a test that I could add a test case to for this and the related PR change? I'm happy to make the code change, just curious on where it would make sense.

labkey-adam:
> > 
> 
> @labkey-adam I did. It is a little hard to search for usages, but I found a few others and took a look to see if they needed the same change or not. Nothing else that I found at the moment.

Sounds good... thanks, @cnathe !

labkey-tchad:
> @labkey-tchad can you think of a test that I could add a test case to for this and the related PR change? I'm happy to make the code change, just curious on where it would make sense.

I don't think we have any existing test that visits this page. I'd suggest `FilesWebpartFileRootTest` or just a new standalone test.
Might make more sense to create a new test for this and the permissions tree encoding. They share more setup with each other than with any existing tests.

PR #3509, opened 2022-07-06
Retry after SQLServer deadlock fetching metadata

labkey-jeckels:
#### Rationale
We're seeing deadlocks on SQL Server when fetching JDBC table metadata after switching to the Microsoft driver. All of the recorded instances should be retryable with our current mechanism. Let's give it a shot!

Example failure:

https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_Ehr_BvtEhrSqlserver/1881748?buildTab=artifacts#%2FtomcatLogs.tar.gz

#### Changes
* Wrap call to get SchemaTableInfo's metadata in a retry

### DIFF
--- a/api/src/org/labkey/api/data/SchemaColumnMetaData.java
+++ b/api/src/org/labkey/api/data/SchemaColumnMetaData.java
@@ -17,7 +17,6 @@
 package org.labkey.api.data;
 
 import org.apache.commons.lang3.StringUtils;
-import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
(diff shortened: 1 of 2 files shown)


labkey-adam (changes requested):
Don't we need to retry each call separately? Or isolate the retry to each metadata query only? For example, if `loadPkColumns()` throws (as I think we see frequently) then `loadColumnsFromMetaData()` will duplicate all the columns on retry.

labkey-jeckels:
@labkey-adam I reworked to try each load method separately. I also hacked the code to simulate random failures and verified things seemed to work as long as I didn't hit three failures on a single load

labkey-adam (approved):

api/src/org/labkey/api/data/SchemaColumnMetaData.java
```diff
@@ -166,11 +168,43 @@ protected void loadFromXml(ColumnType xmlColumn, BaseColumnInfo colInfo, boolean
         colInfo.loadFromXml(xmlColumn, merge);
     }
 
+    /** Simplify the wrapping/unwrapping via a RuntimeSQLException */
+    private interface RetrySqlException
```
labkey-matthewb, line 172: Doesn't  RetryPassthroughException try to solve this same problem?

labkey-jeckels: It's a related but not the same. RetryPassthroughException doesn't get inspected to see if it's a deadlock exception.

Most codepaths should just transform SQLException->RuntimeSQLException and leave it at that. I didn't want to touch a lot of different method signatures to convert them from SQLException to RuntimeSQLException though, so I opted to unwrap it and propagate up the chain as a SQLException

PR #3510, opened 2022-07-06
Issue 45795: Add simple mechanism to persist usage metrics

labkey-jeckels:
#### Rationale
Some of our metrics, like the number of searches performed in LKSM, are counted for the current Tomcat startup. It resets to zero when Tomcat restarts. Capturing a true cumulative count will make analysis much easier going forward.

#### Changes
* Persist counts to the DB every 15 minutes and at shutdown
* Lazily load counts when tallying another instance
* Switch from ints to longs to handle an onslaught of data
* Scope based on metric
* Rename to "simple" instead of "clientSide"


labkey-jeckels:
@cnathe @labkey-adam I kept the class name unchanged even though we're thinking of using this more widely going forward, not just for client-side incremented counters. Any preferences on keeping the name or suggestions on a better one?

core/src/org/labkey/core/metrics/ClientSideMetricServiceImpl.java
```diff
@@ -1,29 +1,115 @@
...
+            public void shutdownStarted()
+            {
+                save();
+            }
+        });
     }
 
     @Override
-    public int increment(String featureArea, String metricName)
+    public long increment(String featureArea, String metricName)
```
labkey-adam: Would it be better to report these metrics by module? This method could take (module, featureArea, metricName) and prepend the module name.

core/src/org/labkey/core/metrics/ClientSideMetricServiceImpl.java
```diff
@@ -1,29 +1,115 @@
...
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 
+/**
+ * Implements simple counter-based metrics with an eye towards ease of use from browser/client code. Persists the
+ * current tally to the DB asynchronously to make the tally a cumulative one for the server.
+ */
 public class ClientSideMetricServiceImpl implements ClientSideMetricService
```
labkey-adam: I know this was originally built as a client-focused service, but I think we'll have just as many server-side usages of this over time. The current name and JavaDocs are therefore a bit misleading, IMO... perhaps generalize to `SimpleMetricsService` so it's more obviously applicable beyond client calls?

core/src/org/labkey/core/metrics/ClientSideMetricServiceImpl.java
labkey-adam: I now see that they all get stuck into "clientSideMetricCounts". Partitioning by module would obviously be a bit more work, but maybe more useful? But if we stick with a single top-level node, I'd prefer "simpleMetricCounts" or similar.

labkey-adam:
> @cnathe @labkey-adam I kept the class name unchanged even though we're thinking of using this more widely going forward, not just for client-side incremented counters. Any preferences on keeping the name or suggestions on a better one?

Sorry, saw this after my review comments. My suggestion is "SimpleMetricsService".

cnathe:
> @cnathe @labkey-adam I kept the class name unchanged even though we're thinking of using this more widely going forward, not just for client-side incremented counters. Any preferences on keeping the name or suggestions on a better one?

No need to keep the name from my point of view. As @labkey-adam suggested, going with something generic like `SimpleMetricsService` would make sense.

core/src/org/labkey/core/CoreModule.java
```diff
@@ -342,9 +344,8 @@ protected void init()
         AnalyticsProviderRegistry.setInstance(new AnalyticsProviderRegistryImpl());
         SummaryStatisticRegistry.setInstance(new SummaryStatisticRegistryImpl());
         UsageMetricsService.setInstance(new UsageMetricsServiceImpl());
-        ClientSideMetricService.setInstance(new ClientSideMetricServiceImpl());
+        SimpleMetricsService.setInstance(new SimpleMetricsServiceImpl());
```
labkey-adam: TeamCity says this is too early to register the service impl, since root container doesn't exist at this point on bootstrap

core/src/org/labkey/core/metrics/ClientSideMetricServiceImpl.java
labkey-jeckels: Renamed

core/src/org/labkey/core/metrics/ClientSideMetricServiceImpl.java
labkey-jeckels: Now reported by module. This was a fair bit of churn as you'll see in the updated diff

core/src/org/labkey/core/CoreModule.java
labkey-jeckels: Fixed

labkey-adam (approved):
Thanks for doing this, Josh

PR #3511, opened 2022-07-06
Enable cross-folder search with Shared folder support for biologics.

Sigmonia:
#### Rationale
https://github.com/LabKey/biologics/issues/1426
Add support for cross folder/level search for LKB

#### Related Pull Requests
* [Biologics](https://github.com/LabKey/biologics/pull/1428)
* [Components](https://github.com/LabKey/labkey-ui-components/pull/895)

#### Changes
* Added new SearchScopes that include the Shared folder

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -605,6 +605,7 @@ public class Container implements Serializable, Comparable<Container>, Securable
         }
         else
         {
+            //TODO should this be hasPermission(user, ReadPermission.class);  ?
             return true;
         }
     }
--- a/api/src/org/labkey/api/search/SearchScope.java
+++ b/api/src/org/labkey/api/search/SearchScope.java
@@ -17,7 +17,13 @@ package org.labkey.api.search;
 
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
+import org.labkey.api.data.ContainerType;
+import org.labkey.api.security.User;
+import org.labkey.api.security.permissions.ReadPermission;
 import org.labkey.api.util.SafeToRenderEnum;
+
+import java.util.HashMap;
+import java.util.List;
 
 /**
  * Options for how widely or narrowly to search on the server, based on the number of containers to include.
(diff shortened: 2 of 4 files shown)


api/src/org/labkey/api/data/Container.java
```diff
@@ -605,6 +605,7 @@ public boolean shouldDisplay(User user)
         }
         else
         {
+            //TODO should this be hasPermission(user, ReadPermission.class);  ?
```
cnathe, line 608: is this a TODO for this PR or for a future story?

api/src/org/labkey/api/search/SearchScope.java
```diff
@@ -17,54 +17,164 @@
...
+        }
+
+        return containers;
+    }
+
+    protected HashMap<String, Container> _getSearchableContainers(User user, Container searchRoot)
+    {
+        HashMap<String, Container> containers = new HashMap<>();
+
+        if (searchRoot.hasPermission(user, ReadPermission.class))
```
cnathe, line 155: does this need any of the "searchable" checks from line 170? or are those just for things like workbooks?

cnathe (approved):

api/src/org/labkey/api/data/Container.java
Sigmonia: this was more of a review question. While we check the read permission in most (though I don't think all) locations prior to calls to shouldDisplay, it seems like it would be more appropriate as the return value rather than a static true.

cnathe: hmm, yeah I see what you mean that it seems like it would make sense to return the boolean regarding if the user has read perm for the container at this point in the method....but it has been a static true since 2009 so I'm not sure.

api/src/org/labkey/api/search/SearchScope.java
Sigmonia: I don't think they are needed, this block is mostly cut and paste from SecurityQuery:100.
Also, I think this will only be invoked when the searchRoot is the currentContainer & shouldDisplay is true so the corresponding searchable checks should always be true.

PR #3512, opened 2022-07-06
Issue 45827: Sample Finder saved searches are not showing up on trial

XingY:
#### Rationale
The Sample Finder saved view are stored using reports. Reports are tied to ReportViewProvider, which is currently registered by study module. LKSM doesn't include study module, so those saved reports are filtered out by report-browseDataTreeAction. ReportViewProvider doesn't need to be defined in study module. This PR moves ReportViewProvider to query.

#### Related Pull Requests
* 

#### Changes
*


api/src/org/labkey/api/study/StudyService.java
```diff
@@ -235,4 +236,6 @@ static StudyService get()
     Collection<String> getParticipantIds(Study study, User user);
 
     boolean participantExists(Study study, String participantId);
+
+    ReportUtil.ReportFilter getStudyReportFilter(boolean editOnly);
```
labkey-adam, line 240: Does anyone call this?

labkey-adam (approved):
Looks good, just was curious about `getStudyReportFilter()`

api/src/org/labkey/api/study/StudyService.java
XingY: Yes, it's called in ReportViewProvider

PR #3531, opened 2022-07-13
Issue 45861: When sample types are renamed, the custom views associated with the original name are lost

XingY:
#### Rationale
[Issue 45861](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45861)

#### Related Pull Requests
* 

#### Changes
* update queryName references when sample type or dataclass is renamed


labkey-susanh (approved):
Maybe we should add a unit test/Selenium test (or an issue to create such a test).

XingY:
> Maybe we should add a unit test/Selenium test (or an issue to create such a test).

Good point. I added a task to the "Save Grid Views - Selenium Tests" backlog

PR #3532, opened 2022-07-13
Try to prevent list import DB deadlock error

labkey-jeckels:
#### Rationale
There are intermittent test failures due to a race condition between list import and list indexing that result in SQL Server DB deadlocks. Example:

https://teamcity.labkey.org/buildConfiguration/bt402/1892028?hideProblemsFromDependencies=false&expandBuildTestsSection=true&hideTestsFromDependencies=false&expandBuildChangesSection=true

#### Changes
* Recache the list's Domain before queuing for indexer, reducing the odds of the indexer needing to query the exp.DomainDescriptor and exp.PropertyDescriptor tables while other imports are happening

### DIFF
--- a/list/src/org/labkey/list/model/ListDefinitionImpl.java
+++ b/list/src/org/labkey/list/model/ListDefinitionImpl.java
@@ -412,14 +412,16 @@ public class ListDefinitionImpl implements ListDefinition
             if (ensureKey)
                 ensureKey();
 
+            Domain domain = getDomain();
+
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
FWIW, 3 of the last 5 develop builds failed with the deadlock, though there was a long series of successful builds before that. I kicked off 6 runs on this FB and they all passed.

labkey-adam (approved):

PR #3533, opened 2022-07-14
Track client API requests

labkey-adam:
#### Rationale
Metrics are helpful. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45844

#### Changes
* Detect requests from client libraries. Count them.
* Make `SimpleMetricsServiceImpl` canonicalize module name and flag unknown modules.


labkey-jeckels (approved):
Nice! Easier than I would have predicted.

labkey-adam:
> Nice! Easier than I would have predicted.

Gotta get my estimation "gold star" somehow, right?

PR #3535, opened 2022-07-14
Move synonym support into BigIron

labkey-adam:
#### Rationale
There's no reason to keep synonym support separate from the rest of SQL Server support. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=45843

#### Related Pull Requests
* https://github.com/LabKey/distributions/pull/289

#### Changes
* Copy the class and test files from Synonym into BigIron
* Remove the convoluted mechanism we use to register a SynonymTableResolver; just construct and return a static instance of this TableResolver like other dialects do.
* Synonym module will be archived


labkey-jeckels (approved):
I see that the three Synonym* classes are just transfers from the other repro and not needing a careful code review

labkey-adam:
> I see that the three Synonym* classes are just transfers from the other repro and not needing a careful code review

Correct, straight copy (with repackaging)

PR #3540, opened 2022-07-16
Fix linked schema problem with ListAuditEvent and DomainAuditEvent

labkey-jeckels:
#### Rationale
ListAuditEvent and DomainAuditEvent audit tables failed to render when passing through a linked schema because they pulled columns from the original table to use when rendering, instead of resolving them later (from the linked schema table in the repro case)

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1166

#### Changes
* Switch from using ColumnInfo from original table to resolving via FieldKey

### DIFF
--- a/api/src/org/labkey/api/exp/property/DomainAuditProvider.java
+++ b/api/src/org/labkey/api/exp/property/DomainAuditProvider.java
@@ -25,8 +25,6 @@ import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerFilter;
 import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.DataColumn;
-import org.labkey.api.data.DisplayColumn;
-import org.labkey.api.data.DisplayColumnFactory;
 import org.labkey.api.data.MutableColumnInfo;
 import org.labkey.api.data.RenderContext;
 import org.labkey.api.data.TableInfo;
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/exp/property/DomainAuditProvider.java
```diff
@@ -268,20 +264,15 @@ public void renderGridCellContents(RenderContext ctx, Writer out) throws IOExcep
...
+        public void addQueryFieldKeys(Set<FieldKey> keys)
         {
-            super.addQueryColumns(columns);
-            if (_containerId != null)
-                columns.add(_containerId);
-            if (_defaultName != null)
-                columns.add(_defaultName);
+            super.addQueryFieldKeys(keys);
+            keys.add(getContainerFieldKey());
+            keys.add(getDefaultNameFieldKey());
```
labkey-susanh, line 280: Does either of these deserve a `@NotNull` annotation?

labkey-susanh (approved):

api/src/org/labkey/api/exp/property/DomainAuditProvider.java
labkey-jeckels: Why not? They're cheap!

PR #3549, opened 2022-07-21
Add rel noopener in action button

ankurjuneja:
#### Rationale
Recent change to catch security issues with anchor tags referencing an outside server and having target="_blank" but no rel="noopener noreferrer" attribute is detecting such URLs in code base and doing great job. 

#### Changes
* Add rel="noopener noreferrer" in ActionButton when target="_blank"

### DIFF
--- a/api/src/org/labkey/api/data/ActionButton.java
+++ b/api/src/org/labkey/api/data/ActionButton.java
@@ -420,6 +420,11 @@ public class ActionButton extends DisplayElement implements Cloneable
         if (_noFollow)
             button.nofollow();
 
+        if (_target.equalsIgnoreCase("_blank"))
+        {
+            button.noOpener();
+        }
+
         if (_actionType.equals(Action.POST) || _actionType.equals(Action.GET))
         {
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
Change looks OK. Test builds likely got caught up in the Artifactory migration

api/src/org/labkey/api/data/ActionButton.java
```diff
@@ -420,7 +420,7 @@ public void render(RenderContext ctx, Writer out) throws IOException
         if (_noFollow)
             button.nofollow();
 
-        if (_target.equalsIgnoreCase("_blank"))
+        if (_target != null && _target.equalsIgnoreCase("_blank"))
```
labkey-adam, line 423: Just FYI: in this situation you can handle the null check by simply swapping the strings being compared: `"_blank".equalsIgnoreCase(_target)`. The constant string is of course never null and `equalsIgnoreCase()` (and `equals()`) will tolerate a null argument.

PR #3550, opened 2022-07-21
Experiment: APIs for external use of lineage edges

labkey-nicka:
#### Rationale
This PR exposes lineage edge APIs on the `ExperimentService` which will allow other features, beyond experiment run derivation, to persist lineage relationships between experiment objects. In support of this the `exp.edge` table has been updated to include a couple of additional columns which can help to identify edges. These columns are:

- `sourceId`: This is the experiment object that is "declaring" or "sourcing" the relationship. Most often this will likely be an object that is also in the relationship (either `toObjectId` or `fromObjectId`), however, that is not a requirement. This column is a foreign key to the `exp.object` table.
- `sourceKey`: This is a string value that can be leveraged to differentiate edges. The use case is that a caller may want to persist two edges that are the same (`toObjectId`, `fromObjectId`, and `sourceId`) but the edges are declared for different reasons. This field can be leveraged to distinctly identify theses edges based on the reason. The value is up to the caller.
- `runId`: This column is not new but the not null constraint has been dropped in order to allow for the table to persist run-less edges (edges not persisted by experiment run derivation).

The `ExperimentService` has a few new methods available on it for mutating and fetching edge relationships. Namely, `addEdges()`, `getEdges()`, and `removeEdges()`. These all model edges via the new `ExpLineageEdge`. Additionally, the get and remove methods make use of `ExpLineageEdge.Options` to allow for users to construct ad-hoc filters for fetching and removing edges. While this options currently look very similar to `ExpLineageEdge` itself (`sourceIds` being the only real difference) I've elected to have them stand on their own so that future divergence doesn't necessitate a refactor of the public APIs.

Lastly, these new APIs explicitly do not allow for run-based edges to be mutated. If a run-based edge mutation is attempted the methods will throw an exception. That said, the `getEdges()` API does allow for fetching of run-based edges.

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/1456

#### Changes
* Added `addEdges()`, `getEdges()`, and `removeEdges()` API methods to `ExperimentService` for external use.
* Added an `ExpLineageEdge` bean for serialization of rows in the `exp.edge` table.
* Experiment edge deletion expanded to cover `sourceId` column.
* Renamed `rebuildAllEdges` to `rebuildAllRunEdges` and updated implementation to only remove/refresh run-based edges.
* Removed redundant usages of `ExperimentService.get()` from within its own implementation.
* Unit tests for adding and removing edges.

### DIFF
--- a/api/src/org/labkey/api/exp/api/ExpLineageEdge.java
+++ b/api/src/org/labkey/api/exp/api/ExpLineageEdge.java
@@ -1,5 +1,7 @@
 package org.labkey.api.exp.api;
 
+import java.util.Objects;
+
 public class ExpLineageEdge
 {
     private Integer _fromObjectId;
@@ -11,6 +13,15 @@ public class ExpLineageEdge
     // Necessary for database serialization
     public ExpLineageEdge()
     {
+    }
+
+    public ExpLineageEdge(Integer fromObjectId, Integer toObjectId, Integer runId, Integer sourceId, String sourceKey)
+    {
+        _fromObjectId = fromObjectId;
+        _toObjectId = toObjectId;
+        _runId = runId;
+        _sourceId = sourceId;
+        _sourceKey = sourceKey;
     }
 
     public Integer getFromObjectId()
@@ -61,6 +72,36 @@ public class ExpLineageEdge
     public void setToObjectId(Integer toObjectId)
     {
         _toObjectId = toObjectId;
+    }
+
+    @Override
+    public String toString()
+    {
+        return String.format(
+            "fromObjectId: %d, toObjectId: %d, runId: %d, sourceId: %d, sourceKey: %s",
+            _fromObjectId, _toObjectId, _runId, _sourceId, _sourceKey
+        );
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(_fromObjectId, _toObjectId, _runId, _sourceId, _sourceKey);
+    }
+
+    @Override
+    public boolean equals(Object obj)
+    {
+        if (!(obj instanceof ExpLineageEdge edge))
+            return false;
+
+        return (
+            Objects.equals(_fromObjectId, edge.getFromObjectId()) &&
+            Objects.equals(_toObjectId, edge.getToObjectId()) &&
+            Objects.equals(_runId, edge.getRunId()) &&
+            Objects.equals(_sourceId, edge.getSourceId()) &&
+            Objects.equals(_sourceKey, edge.getSourceKey())
+        );
     }
 
     public static class Options
--- a/api/src/org/labkey/api/exp/api/ExpLineageEdge.java
+++ b/api/src/org/labkey/api/exp/api/ExpLineageEdge.java
@@ -1,0 +1,104 @@
+package org.labkey.api.exp.api;
+
+public class ExpLineageEdge
+{
+    private Integer _fromObjectId;
+    private Integer _runId;
+    private Integer _sourceId;
+    private String _sourceKey;
+    private Integer _toObjectId;
+
+    // Necessary for database serialization
+    public ExpLineageEdge()
+    {
+    }
+
+    public Integer getFromObjectId()
+    {
+        return _fromObjectId;
+    }
+
+    public void setFromObjectId(Integer fromObjectId)
+    {
+        _fromObjectId = fromObjectId;
+    }
+
+    public Integer getRunId()
+    {
+        return _runId;
+    }
+
+    public void setRunId(Integer runId)
+    {
+        _runId = runId;
+    }
+
+    public Integer getSourceId()
+    {
+        return _sourceId;
+    }
+
+    public void setSourceId(Integer sourceId)
+    {
+        _sourceId = sourceId;
+    }
+
+    public String getSourceKey()
+    {
+        return _sourceKey;
+    }
+
+    public void setSourceKey(String sourceKey)
+    {
+        _sourceKey = sourceKey;
+    }
+
+    public Integer getToObjectId()
+    {
+        return _toObjectId;
+    }
+
+    public void setToObjectId(Integer toObjectId)
+    {
+        _toObjectId = toObjectId;
+    }
+
+    public static class Options
+    {
+        public Integer fromObjectId;
+        public Integer runId;
+        public Integer sourceId;
+        public String sourceKey;
+        public Integer toObjectId;
+
+        public Options fromObjectId(Integer fromObjectId)
+        {
+            this.fromObjectId = fromObjectId;
+            return this;
+        }
+
+        public Options runId(Integer runId)
+        {
+            this.runId = runId;
+            return this;
+        }
+
+        public Options sourceId(Integer sourceId)
+        {
+            this.sourceId = sourceId;
+            return this;
+        }
+
+        public Options sourceKey(String sourceKey)
+        {
+            this.sourceKey = sourceKey;
+            return this;
+        }
+
+        public Options toObjectId(Integer toObjectId)
+        {
+            this.toObjectId = toObjectId;
+            return this;
+        }
+    }
+}
--- a/api/src/org/labkey/api/exp/api/ExpLineageEdge.java
+++ b/api/src/org/labkey/api/exp/api/ExpLineageEdge.java
@@ -1,6 +1,8 @@
 package org.labkey.api.exp.api;
 
+import java.util.HashSet;
 import java.util.Objects;
+import java.util.Set;
 
 public class ExpLineageEdge
 {
@@ -108,7 +110,7 @@ public class ExpLineageEdge
     {
         public Integer fromObjectId;
         public Integer runId;
-        public Integer sourceId;
+        public Set<Integer> sourceIds;
         public String sourceKey;
         public Integer toObjectId;
 
@@ -126,7 +128,13 @@ public class ExpLineageEdge
 
         public Options sourceId(Integer sourceId)
         {
-            this.sourceId = sourceId;
+            this.sourceIds = Set.of(sourceId);
+            return this;
+        }
+
+        public Options sourceIds(Set<Integer> sourceIds)
+        {
+            this.sourceIds = new HashSet<>(sourceIds);
             return this;
         }
 
--- a/api/src/org/labkey/api/exp/api/ExpLineageEdge.java
+++ b/api/src/org/labkey/api/exp/api/ExpLineageEdge.java
@@ -6,18 +6,18 @@ import java.util.Set;
 
 public class ExpLineageEdge
 {
-    private Integer _fromObjectId;
+    private int _fromObjectId; // not nullable
     private Integer _runId;
     private Integer _sourceId;
     private String _sourceKey;
-    private Integer _toObjectId;
+    private int _toObjectId; // not nullable
 
     // Necessary for database serialization
     public ExpLineageEdge()
     {
     }
 
-    public ExpLineageEdge(Integer fromObjectId, Integer toObjectId, Integer runId, Integer sourceId, String sourceKey)
+    public ExpLineageEdge(int fromObjectId, int toObjectId, Integer runId, Integer sourceId, String sourceKey)
     {
         _fromObjectId = fromObjectId;
         _toObjectId = toObjectId;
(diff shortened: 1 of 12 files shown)


api/src/org/labkey/api/exp/api/ExpLineageEdge.java
```diff
@@ -0,0 +1,153 @@
...
+        return (
+            Objects.equals(_fromObjectId, edge.getFromObjectId()) &&
+            Objects.equals(_toObjectId, edge.getToObjectId()) &&
+            Objects.equals(_runId, edge.getRunId()) &&
+            Objects.equals(_sourceId, edge.getSourceId()) &&
+            Objects.equals(_sourceKey, edge.getSourceKey())
+        );
+    }
+
+    public static class Options
```
labkey-susanh: At first glance `Options` seems a strange name as I'm not sure what it's used for.  I see below that it's used for specifying filtering parameters.  Would FilterParams or FilterOptions be a better name?

experiment/resources/schemas/dbscripts/postgresql/exp-22.005-22.006.sql
```diff
@@ -0,0 +1,12 @@
...
+    DROP CONSTRAINT UQ_Edge_ToFrom_RunId,
+
+    ALTER COLUMN RunId DROP NOT NULL,
+
+    ADD SourceId INT NULL,
+    ADD SourceKey VARCHAR(200) NULL,
+
+    ADD CONSTRAINT FK_Edge_SourceId_Object FOREIGN KEY (SourceId) REFERENCES exp.Object (Objectid),
+    ADD CONSTRAINT UQ_Edge_FromTo_RunId_SourceIdSourceKey UNIQUE (FromObjectId, ToObjectId, RunId, SourceId, SourceKey),
+    ADD CONSTRAINT UQ_Edge_ToFrom_RunId_SourceIdSourceKey UNIQUE (ToObjectId, FromObjectId, RunId, SourceId, SourceKey);
```
labkey-susanh: Just curious: why no underscore between `SourceId` and `SourceKey`?

experiment/src/org/labkey/experiment/api/LineageTest.java
```diff
@@ -573,6 +558,96 @@ public void testObjectInputOutput() throws Exception
...
+
+        // Does not allow run-based edge insertion
+        var exceptionThrown = false;
+        try
+        {
+            expSvc.addEdges(List.of(new ExpLineageEdge(aa.objectId, bb.objectId, -1, bb.objectId, sourceKey)));
+        }
+        catch (IllegalArgumentException e)
+        {
+            exceptionThrown = true;
```
labkey-susanh: Instead of this, you could add a `fail` call after the `addEdges` call in the try block since the test should fail if you don't throw.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8091,18 +8090,117 @@ public Integer getObjectIdWithLegacyName(String name, String dataType, Date effe
...
+    {
+        if (edges == null || edges.size() == 0)
+            return;
+
+        List<List<?>> params = new ArrayList<>();
+
+        for (var edge : edges)
+        {
+            if (edge.getRunId() != null)
+                throw new IllegalArgumentException("Failed to add lineage edge. Adding edges with a runId are not supported. Use experiment protocol inputs/outputs if run support is necessary.");
```
```suggestion
                throw new IllegalArgumentException("Failed to add lineage edge. Adding edges with a runId is not supported. Use experiment protocol inputs/outputs if run support is necessary.");
```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8091,18 +8090,117 @@ public Integer getObjectIdWithLegacyName(String name, String dataType, Date effe
...
+            filter.addCondition(FieldKey.fromParts("fromObjectId"), options.fromObjectId);
+        if (options.toObjectId != null)
+            filter.addCondition(FieldKey.fromParts("toObjectId"), options.toObjectId);
+        if (options.runId != null)
+            filter.addCondition(FieldKey.fromParts("runId"), options.runId);
+        if (options.sourceIds != null)
+        {
+            if (options.sourceIds.isEmpty())
+                filter.addWhereClause("0 = 1", new Object[]{});
+            if (options.sourceIds.size() == 1)
```
```suggestion
            else if (options.sourceIds.size() == 1)
```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8091,18 +8090,117 @@ public Integer getObjectIdWithLegacyName(String name, String dataType, Date effe
...
+        if (options.fromObjectId != null)
+            filter.addCondition(FieldKey.fromParts("fromObjectId"), options.fromObjectId);
+        if (options.toObjectId != null)
+            filter.addCondition(FieldKey.fromParts("toObjectId"), options.toObjectId);
+        if (options.runId != null)
+            filter.addCondition(FieldKey.fromParts("runId"), options.runId);
+        if (options.sourceIds != null)
+        {
+            if (options.sourceIds.isEmpty())
+                filter.addWhereClause("0 = 1", new Object[]{});
```
labkey-susanh, line 8157: What's the expected use case here?

labkey-susanh (approved):
I like it.

api/src/org/labkey/api/exp/api/ExpLineageEdge.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.Objects;
+import java.util.Set;
+
+public class ExpLineageEdge
+{
+    private Integer _fromObjectId;
+    private Integer _runId;
+    private Integer _sourceId;
+    private String _sourceKey;
+    private Integer _toObjectId;
```
labkey-jeckels: Are all of these allowed to be null? For example, should we always have a fromObjectId and toObjectId, and if so, can they be int instead of Integer?

experiment/resources/schemas/dbscripts/postgresql/exp-22.005-22.006.sql
labkey-jeckels: What do we gain from having two unique constraints over the same five columns? These can be large on disk, expensive to maintain, and cause deadlocks, so we don't want to add ones that we don't need.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-jeckels: And does using the IN clause with a single value cause problems - could we eliminate this special case?

experiment/resources/schemas/dbscripts/postgresql/exp-22.005-22.006.sql
labkey-jeckels: Hmm... I see that the original table had UQ_Edge_FromTo_RunId and UQ_Edge_ToFrom_RunId, which is the same thing but over three columns. This seems wrong to me.

I think we'd be better served by a single unique constraint, and a separate index on ToObjectId.

experiment/resources/schemas/dbscripts/postgresql/exp-22.005-22.006.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE exp.Edge
+    DROP CONSTRAINT UQ_Edge_FromTo_RunId,
+    DROP CONSTRAINT UQ_Edge_ToFrom_RunId,
+
+    ALTER COLUMN RunId DROP NOT NULL,
+
+    ADD SourceId INT NULL,
+    ADD SourceKey VARCHAR(200) NULL,
+
+    ADD CONSTRAINT FK_Edge_SourceId_Object FOREIGN KEY (SourceId) REFERENCES exp.Object (Objectid),
```
labkey-jeckels, line 10: I'm loathe to add yet another index here, but without one on SourceId, whenever a row is deleted from exp.Object the DB will likely have to do a table scan to check if there are any SourceId references to it.

labkey-jeckels (changes requested):
Let me know if you want to talk about the indices and constraints.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-nicka: The use case is that a set of sourceIds was computed where the result is the empty set. This effectively differentiates `null` (didn't specify a filter) from `emptySet()` (specified a filter but there are no values).

api/src/org/labkey/api/exp/api/ExpLineageEdge.java
labkey-nicka: Yes, `FilterOptions` is a more appropriate name. I'll change this.

experiment/resources/schemas/dbscripts/postgresql/exp-22.005-22.006.sql
labkey-nicka: I agree, there should be.

experiment/resources/schemas/dbscripts/postgresql/exp-22.005-22.006.sql
labkey-nicka: Does this suffice?

```sql
CREATE INDEX IX_Edge_SourceId ON exp.Edge(SourceId);
```

experiment/resources/schemas/dbscripts/postgresql/exp-22.005-22.006.sql
labkey-nicka: Got it. Going to switch to just declaring a single unique constraint. Additionally, adding an index on `toObjectId`.

```sql
ADD CONSTRAINT UQ_Edge_FromTo_RunId_SourceId_SourceKey UNIQUE (FromObjectId, ToObjectId, RunId, SourceId, SourceKey);
CREATE INDEX IX_Edge_ToObjectId ON exp.Edge(ToObjectId);
```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-nicka: Yes, an IN clause with a single value would suffice. I guess I'm trying to optimize before the database does by supplying the "most" correct operator (concept leak over from what I've always done on the client-side when creating filter clauses).

api/src/org/labkey/api/exp/api/ExpLineageEdge.java
labkey-nicka: Updated.

experiment/resources/schemas/dbscripts/postgresql/exp-22.005-22.006.sql
labkey-nicka: Added.

experiment/src/org/labkey/experiment/api/LineageTest.java
labkey-nicka: Much cleaner. Revised.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-nicka: Done.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-nicka: Removed the optimization.

api/src/org/labkey/api/exp/api/ExpLineageEdge.java
labkey-nicka: Done.

experiment/resources/schemas/dbscripts/postgresql/exp-22.005-22.006.sql
labkey-nicka: Added.

experiment/resources/schemas/dbscripts/postgresql/exp-22.005-22.006.sql
labkey-nicka: Added.

labkey-jeckels (approved):

PR #3558, opened 2022-07-25
Data generator for Experiment objects

labkey-susanh:
#### Rationale
We want to be able to exercise our applications that use Experiment object with data at scale. This PR provides a DataGenerator class that can be used to generate random, realistic data for sample types and data classes (other Exp types will likely come later), including the ability to add lineage relationships between objects and create aliquots.  The data generation is driven by a pipeline job that reads a `.properties` file that specifies how many sample types, samples, data classes, etc. as well as the percentage of samples that are to be derived, aliquots, etc.  There are lots of parameters, and likely to be more. Perhaps using a more structured JSON object will work better as we add even more kinds of data to be generated, but simple properties is where we start.

Much of the generation of relationships between the objects relies on the use of the genId values in the names of the entities and the ability to query for the current genId value so we can generate random names known to be valid that can be used to choose random objects.  This is not robust against all interactions, of course, and possibly the use of the IN clauses for the selection of the random samples is going to be too inefficient as we try to generate larger  data, so a different mechanism may be required in the end, but all data generation is batched to prevent overwhelming with the number of inserts or the amount of data being inserted at once. 

As currently set up, the sample types and custom data classes that are generated are added to the ones that already exist. This was easiest for testing purposes, but also allows for the possibility of choosing multiple `.properties` files when running the pipeline job to generate samples using different percentages of aliquots, number of fields, etc.

There is some code commented out (e.g., the code for creating pooled samples), and some methods that are not used, but are left for possible future needs.

**TODO** (perhaps not all in this PR)
- [x] Some accounting is not quite right when generating samples
- [x] The distribution of aliquots across generations still makes for a very wide tree with little chance of generating a deep tree
- Pooling samples - both regular samples and aliquots can be pooled

#### Related Pull Requests
* https://github.com/LabKey/biologics/pull/1462

#### Changes
* Add `DataGenerator` class that can be used to generate sample types and data classes 
* add `createSampleType` method that does not use the deprecated idCol values
* Use some non-deprecated methods for Experiment pipeline job

### DIFF
--- a/api/src/org/labkey/api/data/generator/DataGenerator.java
+++ b/api/src/org/labkey/api/data/generator/DataGenerator.java
@@ -55,6 +55,7 @@ public class DataGenerator<T extends DataGenerator.Config>
     protected Container _container;
     protected final User _user;
     protected final Logger _log;
+    protected  final PipelineJob _job;
     protected T  _config;
 
     // Keep the set of timers so we can produce a report of all times at the end
@@ -88,6 +89,16 @@ public class DataGenerator<T extends DataGenerator.Config>
         _user = user;
         _log = log;
         _config = config;
+        _job = null;
+    }
+
+    public DataGenerator(PipelineJob job, T config)
+    {
+        _container = job.getContainer();
+        _user = job.getUser();
+        _job = job;
+        _log = job.getLogger();
+        _config = config;
     }
 
     public Container getContainer()
@@ -103,6 +114,9 @@ public class DataGenerator<T extends DataGenerator.Config>
     public static void checkAlive(PipelineJob job)
             throws CancelledException
     {
+        if (job == null)
+            return;
+
         if (job.checkInterrupted())
             throw new CancelledException();
 
@@ -162,6 +176,7 @@ public class DataGenerator<T extends DataGenerator.Config>
 
     public void generateSampleTypes(String namePrefix, String namingPatternPrefix) throws ExperimentException, SQLException
     {
+        checkAlive(_job);
         int numSampleTypes = _config.getNumSampleTypes();
         int minFields = _config.getMinFields();
         int maxFields = _config.getMaxFields();
@@ -228,6 +243,7 @@ public class DataGenerator<T extends DataGenerator.Config>
     {
         TableInfo tableInfo = getSamplesSchema().getTable(sampleType.getName());
         QueryUpdateService svc = tableInfo.getUpdateService();
+        checkAlive(_job);
         int numGenerated = 0;
         int numAliquots = Math.round(numSamplesAndAliquots * _config.getPctAliquots());
 //        int numPooled = Math.round(numSamplesAndAliquots * _config.getPctPooled());
@@ -251,6 +267,7 @@ public class DataGenerator<T extends DataGenerator.Config>
         int numDerivedFromSamples = numDerived - numDerivedFromDataClass;
         if (!sampleTypeParents.isEmpty() && numDerivedFromSamples > 0)
         {
+            checkAlive(_job);
             _log.info(String.format("Generated %d samples derived from sample types", numDerivedFromSamples));
             int numPerParentType = numDerivedFromSamples / sampleTypeParents.size();
             for (String parentType : sampleTypeParents)
@@ -281,6 +298,7 @@ public class DataGenerator<T extends DataGenerator.Config>
         int numGenerated;
         do
         {
+            checkAlive(_job);
             List<Map<String, Object>> parents = getRandomSamples(sampleType, Math.min(10, Math.max(quantity, quantity/100)));
             numGenerated = generateAliquotsForParents(parents, svc, quantity - totalAliquots, 0, 1, randomInt(1, _config.getMaxGenerations()));
             totalAliquots += numGenerated;
@@ -298,7 +316,7 @@ public class DataGenerator<T extends DataGenerator.Config>
         int generatedCount = 0;
         List<Map<String, Object>> aliquots = new ArrayList<>();
         List<Map<String, Object>> rows = new ArrayList<>();
-
+        checkAlive(_job);
         for (int p = 0; p < parents.size() && generatedCount < quantity && generatedCount < MAX_BATCH_SIZE; p++)
         {
             // increase the probability we'll get some aliquots in later generations
@@ -522,6 +540,7 @@ public class DataGenerator<T extends DataGenerator.Config>
 
     private int generateDomainData(int totalRows, QueryUpdateService service, Domain domain) throws BatchValidationException, SQLException
     {
+        checkAlive(_job);
         _log.info(String.format("Generating %d rows of data ...", totalRows));
         int numImported = 0;
         int batchSize = Math.min(MAX_BATCH_SIZE, totalRows);
--- a/api/src/org/labkey/api/data/generator/DataGenerator.java
+++ b/api/src/org/labkey/api/data/generator/DataGenerator.java
@@ -188,22 +188,23 @@ public class DataGenerator<T extends DataGenerator.Config>
             _log.info(String.format("Generating %d samples for sample type '%s'.", numSamples, sampleType.getName()));
             CPUTimer timer = addTimer(String.format("%d '%s' samples", numSamples, sampleType.getName()));
             timer.start();
-            generateSamples(sampleType, numSamples, dataClassParents, parentTypes);
+            int numGenerated = generateSamples(sampleType, numSamples, dataClassParents, parentTypes);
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,756 @@
...
+            _nameData.put(sampleTypeName, new NamingPatternData(prefixWithIndex, sampleType.getCurrentGenId()));
+            _sampleTypes.add(sampleType);
+            numFields = Math.min(numFields + fieldIncrement, maxFields);
+        }
+        timer.stop();
+
+        _log.info(String.format("Generating %d sample types took %s", numSampleTypes, timer.getDuration() + "."));
+    }
+
+    public ExpSampleType generateSampleType(String sampleTypeName, @Nullable String namingPattern, int numFields) throws ExperimentException, SQLException
```
labkey-nicka: Consider making this method `protected` or `private` given that it seems like `generateSampleTypes(String namePrefix, String namingPatternPrefix)` is what is intended for external use.

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,756 @@
...
+        public static final String MAX_POOL_SIZE = "maxPoolSize";
+        public static final String MIN_SAMPLES = "minSamples";
+        public static final String MAX_SAMPLES = "maxSamples";
+        public static final String MAX_ALIQUOTS_PER_SAMPLE = "maxAliquotsPerSample";
+        public static final String MAX_CHILDREN_PER_PARENT = "maxChildrenPerParent";
+        public static final String MAX_GENERATIONS = "maxGenerations";
+        public static final String MIN_NUM_FIELDS = "minFields";
+        public static final String MAX_NUM_FIELDS = "maxFields";
+
+        int _numSampleTypes = 0;
```
labkey-nicka, line 640: nit: Value assignments can be removed from these properties as their default value is specified in the constructor.

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,756 @@
...
+        }
+        else
+        {
+            parentInput = "MaterialInputs";
+            parentObject = SampleTypeService.get().getSampleType(_container, _user, parentQueryName);
+        }
+        _log.info(String.format("Generating %d '%s' samples derived from '%s/%s' ...", quantity, sampleType.getName(), parentInput, parentQueryName));
+        CPUTimer timer = addTimer(String.format("%d '%s/%s' derived samples", quantity, parentInput, parentQueryName));
+        timer.start();
+        NamingPatternData namingData = _nameData.get(parentQueryName);
```
labkey-nicka: nit: This `namingData` appears to be unused.

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,756 @@
...
+
+    public void generateSampleTypes(String namePrefix, String namingPatternPrefix) throws ExperimentException, SQLException
+    {
+        int numSampleTypes = _config.getNumSampleTypes();
+        int minFields = _config.getMinFields();
+        int maxFields = _config.getMaxFields();
+
+        int fieldIncrement = numSampleTypes <= 1 ? 0 : (maxFields - minFields)/(numSampleTypes-1);
+        SampleTypeService service = SampleTypeService.get();
+        CPUTimer timer = new CPUTimer(String.format("%d sample types", numSampleTypes));
```
labkey-nicka: Consider using `addTimer()`.

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,756 @@
...
+//            else
+//            {
+//                _log.info(String.format("Generated %d pooled sample and then ran out of parents.", i+1));
+//            }
+//        }
+//        return service.updateRows(_user, _container, rows, oldKeys, null, null);
+    }
+
+
+    private List<String> getRandomNames(String namePrefix, long startIndex, long endIndex, int quantity)
```
labkey-nicka: Consider having this generate a list of unique names.

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,756 @@
...
+                    rowNum++;
+                }
+                p++;
+            }
+            ListofMapsDataIterator rowsDI = new ListofMapsDataIterator(rows.get(0).keySet(), rows);
+            numImported += service.importRows(_user, _container, rowsDI, errors, null, null);
+            if (errors.hasErrors())
+                throw errors;
+
+            _log.info("... " + numImported);
```
labkey-nicka: nit: Not sure this interstitial logging is very useful.

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,756 @@
...
+        fieldPrefixes.add(new FieldPrefix("date", "DateField"));
+    }
+
+    public DataGenerator(Container container, User user, T config, Logger log)
+    {
+        _container = container;
+        _user = user;
+        _log = log;
+        _config = config;
+        _samplesSchema = QueryService.get().getUserSchema(_user, _container, SchemaKey.fromParts(SamplesSchema.SCHEMA_NAME));
```
labkey-nicka: nit: Given that this generator might extend to additional schema types in the future I don't think we should hold onto instances of the schemas here. We're already holding onto the `_user` and `_container` so we should be able to resolve these dynamically whenever we need and allow the underlying services to handle caching.

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -0,0 +1,756 @@
...
+        return getRowsByRandomNames(tableInfo, _nameData.get(sampleType.getName()), sampleType.getCurrentGenId(), quantity, Set.of("Name", "RowId"));
+    }
+
+    private List<Map<String, Object>> getRandomDataClassObjects(ExpDataClass dataClass, int quantity)
+    {
+        TableInfo tableInfo = _dataClassSchema.getTable(dataClass.getName());
+        return getRowsByRandomNames(tableInfo, _nameData.get(dataClass.getName()), dataClass.getCurrentGenId(), quantity, Set.of("Name", "RowId"));
+    }
+
+    protected List<Map<String, Object>> getRowsByRandomNames(TableInfo tableInfo, NamingPatternData namingData, long endIndex, int quantity, Set<String> columns)
```
labkey-nicka, line 371: Nifty.

labkey-nicka (approved):
This is a great start!

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: I was thinking either could be useful, but certainly the plural one is what's used now.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: I have imagined a case where you instantiate the Config with the (currently non-existent) default constructor and then use the setters to update to non-default values.  Not used that way currently, of course, except in my imagination...

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: Thanks.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: Removed.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: It is when you're wondering if anything is happening or you've gone into some infinite loop. I'll leave it for now.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: Good thought. Switched to using a `Set` for uniqueness, but leaving it so it generates at most `quantity` names.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: Just trying to avoid boilerplate.  Added getter methods instead that will resolve dynamically.

labkey-nicka (approved):

PR #3570, opened 2022-08-01
Custom fields on Workflow Jobs - Job type field definition

cnathe:
#### Rationale
We want to support workflow jobs to replace the work of the issue tracker by supporting teams’ needs to use Workflow as a tool for requesting work or assays to be performed on samples. One thing that has been called out as a primary blocker is the ability to standardize fields captured around a specific request.

This PR adds the domain kind allowSampleSubjectProperties option so that we can disable/override it in the new workflow job template domain kind.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/922
* https://github.com/LabKey/sampleManagement/pull/1137
* https://github.com/LabKey/biologics/pull/1487
* https://github.com/LabKey/inventory/pull/507
* https://github.com/LabKey/platform/pull/3570

#### Changes
* add domain kind allowSampleSubjectProperties option

### DIFF
--- a/api/gwtsrc/org/labkey/api/gwt/client/model/GWTDomain.java
+++ b/api/gwtsrc/org/labkey/api/gwt/client/model/GWTDomain.java
@@ -44,6 +44,7 @@ public class GWTDomain<FieldType extends GWTPropertyDescriptor> implements IsSer
     private boolean allowAttachmentProperties;
     private boolean allowFlagProperties;
     private boolean allowTextChoiceProperties;
+    private boolean allowSampleSubjectProperties;
     private boolean allowTimepointProperties;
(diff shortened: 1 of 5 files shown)


labkey-susanh (approved):
I assume you are intentionally choosing not to update the package.json in other modules in platform at this time.

cnathe:
> I assume you are intentionally choosing not to update the package.json in other modules in platform at this time.

The changes to the field editor don't apply to those other platform modules, so I was going to just bump core at this time. Given that we have seen some issues with start-link lately in a few of the modules, I plan to make a pass for all of the LK package.json files to bump the components and build packages soon.

PR #3571, opened 2022-08-01
Issue 45522: GridPanel search does not find extended characters in MSSQL

labkey-martyp:
#### Rationale
MSSQL doesn't recognize extended character set by default like PG does. Using JDBC parameters will set the necessary metadata. Need to create the like value expression string in query to use JDBC parameter.

#### Changes
* Update QClause like value expression to use JDBC parameter.


labkey-jeckels (changes requested):
Needs adjustment so that we keep doing the LIKE-related escaping but not the other escaping that happens earlier in this method.

labkey-martyp:
> Needs adjustment so that we keep doing the LIKE-related escaping but not the other escaping that happens earlier in this method.

Good call. Removed that and did a little cleanup.

labkey-jeckels (approved):

PR #3580, opened 2022-08-03
Issue 46021: Duplicate sample names resulting in 22.7 upgrade failure

XingY:
#### Rationale
Prior to 22.7, sample name uniqueness in a sample type relies on unique lsid, or unique (container, cpastype, name). When created from UI or api, LSIDs for samples of the same type follows a fixed format and would guarantee all sample names are unique. However, when lsid are constructed manually, such as importing from xar, LSID can no longer guarantee name uniqueness. In a subfolder setup, unique key (container, cpastype, name) also fails to guarantee uniqueness of samples.

We are adding a unique key constraint on expsampletype.name to help support sample renaming feature. Prior to that, samples with duplicate names needs to be cleaned up (renamed).

#### Related Pull Requests
* 

#### Changes
* Add a uniquifySampleNames step to rename duplicate samples prior to adding the unique constraint

### DIFF
--- a/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
+++ b/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
@@ -393,7 +393,10 @@ public class ExperimentUpgradeCode implements UpgradeCode
                 for (String lsid : lsids)
                 {
                     if (lsid.startsWith(lsidPrefix))
+                    {
                         sampleLSIDToKeep = lsid;
+                        break;
+                    }
                 }
 
                 // rename all but one samples by adding suffix _2, _3, etc
--- a/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
+++ b/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
@@ -19,6 +19,7 @@ import com.google.common.collect.Sets;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
+import org.labkey.api.collections.CaseInsensitiveHashSet;
 import org.labkey.api.data.ColumnInfo;
 import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.DbSchema;
@@ -30,6 +31,7 @@ import org.labkey.api.data.ParameterMapStatement;
 import org.labkey.api.data.PropertyStorageSpec;
 import org.labkey.api.data.SQLFragment;
 import org.labkey.api.data.SchemaTableInfo;
+import org.labkey.api.data.SimpleFilter;
 import org.labkey.api.data.SqlExecutor;
 import org.labkey.api.data.SqlSelector;
 import org.labkey.api.data.TableChange;
@@ -46,6 +48,7 @@ import org.labkey.api.pipeline.PipeRoot;
 import org.labkey.api.pipeline.PipelineJob;
 import org.labkey.api.pipeline.PipelineJobService;
 import org.labkey.api.pipeline.PipelineService;
+import org.labkey.api.query.FieldKey;
 import org.labkey.api.util.FileUtil;
 import org.labkey.api.util.GUID;
 import org.labkey.api.util.URLHelper;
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -332,6 +340,94 @@ private static void setSampleName(ExpSampleTypeImpl st)
...
+                .append("SELECT LSID, name\n")
+                .append("FROM ").append(ExperimentService.get().getTinfoMaterial(), "m").append("\n")
+                .append("WHERE m.name IN (").append("\n")
+                .append("  SELECT name FROM ").append(ExperimentService.get().getTinfoMaterial(), "mi").append("\n")
+                .append("  WHERE mi.CpasType = ? ")
+                .add(targetCpas)
+                .append("  GROUP BY (name) HAVING COUNT(*) > 1").append("\n")
+                .append(") \n")
+                .append("AND m.CpasType = ?")
+                .add(targetCpas);
```
labkey-adam, line 364: This grouping on name is case-sensitive on PostgreSQL but case-insensitive on SQL Server. Sets below are always case-insensitive. Any issue there?

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -332,6 +340,94 @@ private static void setSampleName(ExpSampleTypeImpl st)
...
+            {
+                List<String> lsids = sampleLsids.get(sampleName);
+                LOG.warn(lsids.size() + " samples with the name '" + sampleName + "' are found for sample type " + st.getName() + ". ");
+
+                // prefer to keep the sample whose LSID conforms to the sample type's defined prefix
+                String sampleLSIDToKeep = lsids.get(0);
+                for (String lsid : lsids)
+                {
+                    if (lsid.startsWith(lsidPrefix))
+                        sampleLSIDToKeep = lsid;
```
labkey-adam, line 397: Add a `break;` here? Any reason to keep looping once we find a match?

labkey-adam (approved):
A couple questions to consider but no need for re-review

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
XingY: Good question.

This would result in slight difference on how/when samples are renamed for Postgresql vs sql server. Sample names that differ only in casing would not be renamed for Postgresql, but would be for sql server. But both behavior work work to allow addition of unique index (case sensitive).

labkey-adam: Are we planning to add a unique index? I wonder if we'd want it to be case-insensitive instead. (Sounds like maybe a Hannah decision.) I'm not looking to hold up this fix, but just noting that we'll likely need to re-run "uniquify" (or a case-insensitive version of it) before adding the constraint.

XingY: Yes, a unique index is added to the provisioned table after this uniquify. It's not a trivial decision to make to go case insensitive for sample (and data?) names. Definitely not a 22.7 thing. 

`
        Set<PropertyStorageSpec.Index> newIndices =  Collections.unmodifiableSet(Sets.newLinkedHashSet(Arrays.asList(new PropertyStorageSpec.Index(true, "name"))));
       
 StorageProvisioner.get().addOrDropTableIndices(domain, newIndices, true, TableChange.IndexSizeMode.Normal);
`

labkey-adam: Okay, got it. Makes sense that the provisioner is responsible for this. This seems good for now.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -332,6 +340,97 @@ private static void setSampleName(ExpSampleTypeImpl st)
...
+                .append("  GROUP BY (name) HAVING COUNT(*) > 1").append("\n")
+                .append(") \n")
+                .append("AND m.CpasType = ?")
+                .add(targetCpas);
+
+        TableInfo table = ExperimentService.get().getTinfoMaterial();
+        @NotNull Map<String, Object>[] results = new SqlSelector(ExperimentService.get().getSchema(), sql).getMapArray();
+        if (results.length > 0)
+        {
+            LOG.warn("Samples with duplicate names found for sample type " + st.getName());
```
labkey-jeckels:
```suggestion
            LOG.warn(results.length + " duplicate name(s) found for sample type " + st.getName());
```

labkey-jeckels (approved):
Looks good to me

PR #3581, opened 2022-08-03
Issue 45375: Add index to speed common queries against core.notifications

labkey-jeckels:
#### Rationale
There are a handful of reasonably frequent queries against core.notifications that are taking >20ms to complete. An index should help.

#### Changes
* Add index on UserId column, which is a common WHERE clause filter on all the queries

### DIFF
--- a/core/module.properties
+++ b/core/module.properties
@@ -1,6 +1,6 @@
 Name: Core
 ModuleClass: org.labkey.core.CoreModule
-SchemaVersion: 22.002
+SchemaVersion: 22.003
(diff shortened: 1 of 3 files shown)


labkey-adam:
I've introduced a core script with the same versions in my Signing PR. I can bump mine to the next increment, but there will likely be timing problems given that you're merging to 22.8 and I'm merging to develop. Is it important that your change targets 22.8?

labkey-adam (approved):

PR #3584, opened 2022-08-05
Log Query Analyzer error messages

labkey-klum:
#### Rationale
To help investigate occasional automated test failures a few changes were made to the query analysis tool:
- Instead of displaying a generic error dialog, we will try to show the standard LabKey error dialog with details about the execption.
- For the server side `AnalyzeQueriesAction`, log any exceptions to the error log for the `QueryController` logger.

[Related Issue
](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45908)


labkey-danield (commented):
To me this looks like it should do the job, but I will defer to Cory to approve.

cnathe (approved):

PR #3586, opened 2022-08-05
45752: post-upgrade unexpected query validation

labkey-klum:
#### Rationale
This change restores the previous behavior where CDISC and REDCap imports skipped query validation by default. These are the only two places where `PipelineService.runGenerateFolderArchiveAndImportJob` is used but I added one with a signature to allow an `ImportOptions` object to be passed in as well.

[Related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45752)


labkey-adam:
I think you need a merge from develop. Thanks to me.

labkey-adam (approved):

PR #3589, opened 2022-08-08
Prevent deletion of objects referenced in active Notebooks

labkey-susanh:
#### Rationale
We want to prevent deletion of samples, data class objects, and notebooks that are referenced in active notebooks so the notebooks remain accurate.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/927
* https://github.com/LabKey/biologics/pull/1499
* https://github.com/LabKey/sampleManagement/pull/1153
* https://github.com/LabKey/labbook/pull/249

#### Changes
* Add methods to `ExperimentService` to implement a registry of object referencers
* Add `ObjectReferencer` interface for modules to implement for checking for references to entities given ids, schema name, and query name.
* Update Grid view for data class objects to check for parentage before deleting data class objects
* Move and generalize `confirmDelete.js` to make usable for notebook delete confirmation as well.


internal/webapp/dataregion/confirmDelete.js
```diff
@@ -76,6 +68,7 @@ LABKEY.experiment.confirmDelete = function(dataRegionName, schemaName, queryName
                 Ext4.Msg.show({
                     title: numCanDelete > 0 ? "Permanently delete " + numCanDelete + " " + canDeleteNoun : "No " + nounPlural + " can be deleted",
                     msg: text,
+                    width: 450, // added to keep the text from being swallowed vertically
```
labkey-klum, line 83: I've seen this a couple of times, makes me wonder if there is a bug with Ext 4 and the latest versions of Chrome. I'll sync with Cory to see if there is something we can add into `ext-patches.js` to fix the sizing problem.

assay/src/org/labkey/assay/AssayController.java
```diff
@@ -1721,4 +1723,58 @@ public ApiResponse execute(Object object, BindException errors)
...
+
+    @Marshal(Marshaller.Jackson)
+    @RequiresPermission(ReadPermission.class)
+    public static class GetAssayRunDeletionConfirmationDataAction extends ReadOnlyApiAction<OperationConfirmationForm>
+    {
+
+        @Override
+        public Object execute(OperationConfirmationForm form, BindException errors) throws Exception
+        {
+            Collection<Integer> permittedIds = form.getIds(false);
```
labkey-klum, line 1735: Do we need to clear the `DataRegionSelection` if there are no dependencies found, or does it not matter since those rows won't be on the page later? I'm assuming we aren't clearing these here because we want to maintain selection if we reshow the page.

labkey-klum (approved):

internal/webapp/dataregion/confirmDelete.js
labkey-susanh: Yeah, I got this hint from Cory (never would have changed the width to fix the height...)

assay/src/org/labkey/assay/AssayController.java
labkey-susanh: We don't clear because this is just checking for dependencies, and the user might decide not to proceed.

labkey-alan (approved):

PR #3591, opened 2022-08-08
45831: File watcher UI is available without the Premium module

labkey-klum:
#### Rationale
When the premium module is not deployed we would previously show this message to the user

image

A subsequent change removed this check, although since no file pipeline trigger types were registered the user couldn't create a valid trigger configuration. This change restores the original upsell message.

[Related issue](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=45831)

### DIFF
--- a/pipeline/src/org/labkey/pipeline/createPipelineTrigger.jsp
+++ b/pipeline/src/org/labkey/pipeline/createPipelineTrigger.jsp
@@ -28,14 +28,13 @@
 <%@ page import="com.fasterxml.jackson.databind.ObjectMapper" %>
 <%@ page import="java.util.HashMap" %>
 <%@ page import="org.labkey.api.util.UniqueID" %>
+<%@ page import="org.labkey.api.pipeline.trigger.PipelineTriggerRegistry" %>
 <%@ page extends="org.labkey.api.jsp.JspBase" %>
 <%@ taglib prefix="labkey" uri="http://www.labkey.org/taglib" %>
 <%!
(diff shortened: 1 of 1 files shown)


pipeline/src/org/labkey/pipeline/createPipelineTrigger.jsp
```diff
@@ -28,14 +28,13 @@
...
 <%@ page import="org.labkey.api.util.UniqueID" %>
+<%@ page import="org.labkey.api.pipeline.trigger.PipelineTriggerRegistry" %>
 <%@ page extends="org.labkey.api.jsp.JspBase" %>
 <%@ taglib prefix="labkey" uri="http://www.labkey.org/taglib" %>
 <%!
     @Override
     public void addClientDependencies(ClientDependencies dependencies)
     {
-        dependencies.add("internal/jQuery");
-//        dependencies.add("http://localhost:3001/createPipelineTrigger.js");
```
cnathe, line 38: we've been leaving in these commented out dependencies so that it is easier to switch to hot module reload mode when doing dev on this React entry point. So the jQuery dep can go, but I'd leave in this line 38.

cnathe (approved):

pipeline/src/org/labkey/pipeline/createPipelineTrigger.jsp
labkey-klum: 👍

PR #3595, opened 2022-08-10
Issue 46060: Excel exports are leaving permanent 'temporary poifiles'

labkey-adam:
#### Rationale
We aren't closing `Workbook` instances and we aren't calling `SXSSFWorkbook.dispose()`. We're getting what we deserve.

#### Changes
* Refactor `ExcelWriter` for the jillionth time. Now, `renderWorkbook()` is the key method that all callers invoke. Privatize, protectorize, or eliminate all other render methods. Code paths that need special handling (multiple sheets, special workbook settings) can subclass and override `renderSheets()`. This enables us to create the Workbook in a try-with-resources to ensure resources are closed in all scenarios.

### DIFF
--- a/api/src/org/labkey/api/data/ExcelWriter.java
+++ b/api/src/org/labkey/api/data/ExcelWriter.java
@@ -532,18 +532,19 @@ public class ExcelWriter implements ExportWriter, AutoCloseable
     }
 
     /**
-     * By default, this renders a single sheet and writes the workbook. Subclasses can override to write multiple sheets, etc.
+     * By default, this renders a single sheet and writes the workbook. Subclasses can override to write multiple sheets
+     * or otherwise customize the workbook.
      */
     protected void renderSheets(Workbook workbook)
     {
-        renderNewSheet();
+        renderNewSheet(workbook);
     }
 
-    public void renderNewSheet()
+    protected void renderNewSheet(Workbook workbook)
     {
         _currentRow = 0;
         _currentSheet++;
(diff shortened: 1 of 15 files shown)


api/src/org/labkey/api/data/ExcelWriter.java
```diff
@@ -379,208 +356,218 @@ public List<String> getHeaders()
...
-     * @param response to write to
-     * @param filenamePrefix string to prepend to a time stamp as filename
-     */
-    public void writeWorkbook(HttpServletResponse response, String filenamePrefix)
+    protected void renderSheet(Workbook workbook, int sheetNumber)
     {
-        try (ServletOutputStream outputStream = getOutputStream(response, filenamePrefix, _docType))
+        Sheet sheet;
+        //TODO: Pass render context all the way through Excel writers...
+        RenderContext ctx = new RenderContext(HttpView.currentContext());
```
labkey-jeckels, line 489: With the old method signature taking an HttpServletResponse, it implied that this couldn't be call outside of an HTTP request. This is less clear now. Probably OK.

labkey-jeckels (approved):

api/src/org/labkey/api/data/ExcelWriter.java
labkey-adam: I don't love pulling the current context out of thin air, but this isn't really new. We had `renderSheetAndWrite(OutputStream)` that essentially became `renderWorkbook(OutputStream)`.

PR #3598, opened 2022-08-11
Reference updated UI component for customize grid dialog that has testing hooks.

labkey-danield:
#### Rationale
Updated UI component used in the Customize Grid Dialog. Testing hooks added to make it easier to navigate expanded tree options.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1188

#### Changes
* Updated component reference.
* Updated calls to test helpers for customizing grid views in LKS.


core/package.json
```diff
@@ -52,7 +52,7 @@
   },
   "dependencies": {
     "@labkey/api": "1.15.0",
-    "@labkey/components": "2.203.0",
+    "@labkey/components": "2.203.0-fb-addViewTestsToGridPanelTest.0",
```
cnathe: @labkey-danield let me know when you are ready to merge this and I'll do the publish process to get this updated to the final release version before the actual merge happens

study/test/src/org/labkey/test/tests/study/StudyPublishTest.java
```diff
@@ -1046,7 +1047,7 @@ private String createOneOfFilterString(String[] values)
         return filterStr;
     }
 
-    private void createCustomView(String name, String dataset, String[] ptids, boolean shared)
+    private void createCustomView(String name, String dataset, String[] ptids, @Nullable Boolean shared)
```
labkey-tchad, line 1050: I don't think `shared` should be nullable in `createCustomView`. For an existing view, it makes sense to leave the setting unchanged but not so much for a new view.

labkey-tchad (approved):
Just a nit

PR #3599, opened 2022-08-11
Test no-question-mark mode

labkey-adam:
#### Rationale
We'd like to rid ourselves of the superfluous question mark. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=1713

#### Changes
* Move to setters and constructors that take ActionURL, not String. Migrate sketchy String url concatenation to ActionURL.addParameter().

### DIFF
--- a/core/src/org/labkey/core/CoreModule.java
+++ b/core/src/org/labkey/core/CoreModule.java
@@ -988,7 +988,7 @@ public class CoreModule extends SpringModule implements SearchService.DocumentPr
                 false);
 
         // Temporary hack to force no-question-mark mode for a round of testing. TODO: Remove
-        ExperimentalFeatureService.get().setFeatureEnabled(AppProps.EXPERIMENTAL_NO_QUESTION_MARK_URL, true, null);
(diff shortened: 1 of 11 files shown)


core/src/org/labkey/core/CoreModule.java
```diff
@@ -987,6 +987,9 @@ public void moduleStartupComplete(ServletContext servletContext)
                 "Don't append '?' to URLs unless there are query parameters.",
                 false);
 
+        // Temporary hack to force no-question-mark mode for a round of testing. TODO: Remove
+        //ExperimentalFeatureService.get().setFeatureEnabled(AppProps.EXPERIMENTAL_NO_QUESTION_MARK_URL, true, null);
```
labkey-jeckels, line 991: Is this where you want to leave this PR?

labkey-jeckels (approved):

core/src/org/labkey/core/CoreModule.java
labkey-adam: Yes... I'd like to commit the changes in batches to get coverage both with and without the flag on, and to ensure the branches don't get too far out-of-date. After merging these PRs I expect to open a new set for the next round... uncommenting this line again for that.

PR #3600, opened 2022-08-11
Issue 46056: Folder disappears when import is canceled after transform script warning

XingY:
#### Rationale
[issue link](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46056)

We want to modify the behavior so canceling out of transform warning would only delete the working directory if starting from new upload.

#### Related Pull Requests
* 

#### Changes
* Only delete temp folder when canceling out of transform warning
* Clear session file queue for protocol when canceling out of transform warning

### DIFF
--- a/api/src/org/labkey/api/assay/actions/AssayRunsAction.java
+++ b/api/src/org/labkey/api/assay/actions/AssayRunsAction.java
@@ -20,6 +20,7 @@ import org.apache.commons.io.FileUtils;
 import org.labkey.api.assay.AssayProvider;
 import org.labkey.api.assay.AssayRunsView;
 import org.labkey.api.assay.AssayUrls;
+import org.labkey.api.assay.PipelineDataCollector;
 import org.labkey.api.data.DataRegionSelection;
 import org.labkey.api.exp.api.ExpProtocol;
 import org.labkey.api.qc.TsvDataExchangeHandler;
@@ -32,6 +33,8 @@ import org.springframework.validation.BindException;
 import org.springframework.web.servlet.ModelAndView;
 
 import java.io.File;
+
+import static org.labkey.api.assay.AssayFileWriter.TEMP_DIR_NAME;
 
 /**
  * User: brittp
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/assay/actions/AssayRunsAction.java
```diff
@@ -88,7 +91,13 @@ public ModelAndView getView(AssayRunsForm summaryForm, BindException errors) thr
         {
             File tempDir = TsvDataExchangeHandler.removeWorkingDirectory(summaryForm, getUser());
             if(null != tempDir && tempDir.exists())
-                FileUtils.deleteDirectory(tempDir);
+            {
+                if (tempDir.getParentFile().isDirectory() && tempDir.getParentFile().getName().equals(TEMP_DIR_NAME)) // tempDir is at .uploadTemp/tempGUID for temp import files
+                    FileUtils.deleteDirectory(tempDir);
+                else // don't delete files on the server's file system that are not temp files, but do remove them from file queue in session.
+                    PipelineDataCollector.clearFileQueue(context.getRequest().getSession(true), context.getContainer(), _protocol);
```
labkey-martyp, line 98: Just curious, in the case we do delete the file does it automatically get cleared from the session?

labkey-jeckels: I don't think it will be in the session - PipelineDataCollector uses session to remember which file(s) as user selects from the file browser but the other alternatives like TextAreaDataCollector (for copy/pasted TSVs) or FileUploadDataCollector don't.

api/src/org/labkey/api/assay/actions/AssayRunsAction.java
```diff
@@ -88,7 +91,13 @@ public ModelAndView getView(AssayRunsForm summaryForm, BindException errors) thr
         {
             File tempDir = TsvDataExchangeHandler.removeWorkingDirectory(summaryForm, getUser());
             if(null != tempDir && tempDir.exists())
-                FileUtils.deleteDirectory(tempDir);
+            {
+                if (tempDir.getParentFile().isDirectory() && tempDir.getParentFile().getName().equals(TEMP_DIR_NAME)) // tempDir is at .uploadTemp/tempGUID for temp import files
```
labkey-jeckels, line 95: It would be nice if we could defer to whatever AssayDataCollector instance was in use for the upload, but I imagine that's not being propagated into this action. I think this if/else should be sufficient to cover the four implementations we have at the moment, but worth testing a re-import of an existing run that's reusing the original file if you haven't already.

labkey-jeckels (approved):
I recommend testing cancelling an import for a copy/pasted TSV and a re-import of an existing run (reusing the original file) but I think both will behave reasonably with this code.

PR #3604, opened 2022-08-12
Always capture thread name for connection leak logging

labkey-jeckels:
#### Rationale
While we don't always want to always pay the perf overhead of capturing creation stack traces for ResultSets, knowing the thread responsible can sometimes be enough to track down code that fails to close it

#### Changes
* Capture and log the creating thread name after a leak
* Misc lint cleanup

### DIFF
--- a/api/src/org/labkey/api/data/ResultSetImpl.java
+++ b/api/src/org/labkey/api/data/ResultSetImpl.java
@@ -42,6 +42,7 @@ public class ResultSetImpl extends LoggingResultSetWrapper implements TableResul
 
     private final @Nullable DbScope _scope;
     private final @Nullable Connection _connection;
+    private final String _creatingThreadName;
     private int _maxRows;
     private boolean _countComplete;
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/data/ResultSetImpl.java
```diff
@@ -50,7 +51,7 @@ public class ResultSetImpl extends LoggingResultSetWrapper implements TableResul
     protected int _size;
 
     // for resource tracking
-    private StackTraceElement[] _debugCreated = null;
+    private StackTraceElement[] _debugCreated;
```
labkey-adam: `final`?

labkey-adam (approved):

api/src/org/labkey/api/data/ResultSetImpl.java
labkey-jeckels: Now so very final

PR #3607, opened 2022-08-15
Issue 46019: CustomViewUtil update to account for filter value of type Integer

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46019

#### Changes
* CustomViewUtil update to account for filter value of type Integer when getting filterInfo from json filters


labkey-jeckels (approved):
Consider `Objects.toString()` for this pattern.

cnathe:
> Consider `Objects.toString()` for this pattern.

Thanks for the suggestion. Switched to `Objects.toString(value, "")`.

PR #3610, opened 2022-08-16
Update JSP for assay run deletion confirmation with message about referenced runs

labkey-susanh:
#### Rationale
The Previous PR incorporate a javascript confirmation alert for deletion. of assay runs, in keeping with the behavior we have for samples and data class objects, but this alert then missed notifying users about runs linked to studies and possible permission restrictions there.  Here, we revert the change to use the javascript alert and pass through data to the JSP page to communicate to users when assay runs cannot be deleted because of object references. 

#### Related Pull Requests
#3589

#### Changes
* Revert change to delete button action in ExperimentRunListView
* Update object referencer interface with a method for providing a description of the references

### DIFF
--- a/api/src/org/labkey/api/exp/ExperimentRunListView.java
+++ b/api/src/org/labkey/api/exp/ExperimentRunListView.java
@@ -85,10 +85,6 @@ public class ExperimentRunListView extends QueryView
         AssayService svc = AssayService.get();
         if (svc != null)
             addClientDependencies(svc.getClientDependenciesForImportButtons());
-        addClientDependencies(Set.of(
-                ClientDependency.fromPath("Ext4"),
-                ClientDependency.fromPath("dataregion/confirmDelete.js")
-        ));
     }
 
     public static QuerySettings getRunListQuerySettings(UserSchema schema, ViewContext model, String tableName, boolean allowCustomizations)
--- a/api/src/org/labkey/api/exp/ExperimentRunListView.java
+++ b/api/src/org/labkey/api/exp/ExperimentRunListView.java
@@ -49,7 +49,6 @@ import org.labkey.api.view.DataView;
 import org.labkey.api.view.HttpView;
 import org.labkey.api.view.NavTree;
 import org.labkey.api.view.ViewContext;
-import org.labkey.api.view.template.ClientDependencies;
 import org.labkey.api.view.template.ClientDependency;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/exp/ExperimentRunListView.java
```diff
@@ -176,19 +175,12 @@ protected void populateButtonBar(DataView view, ButtonBar bar)
 
         if (showDeleteButton())
         {
-            ActionURL url = PageFlowUtil.urlProvider(ExperimentUrls.class).getDeleteRunsURL(context.getContainer());
+            ActionURL url = PageFlowUtil.urlProvider(ExperimentUrls.class).getDeleteSelectedExpRunsURL(context.getContainer(), getReturnURL());
```
cnathe, line 174: the previous PR added client dependencies at ~ line 88. should those be removed as well?

```
        addClientDependencies(Set.of(
                ClientDependency.fromPath("Ext4"),
                ClientDependency.fromPath("dataregion/confirmDelete.js")
        ));
```

cnathe: Also, is the `getDeleteRunsURL` from `ExperimentUrls.java` still needed (added from previous PR)?

experiment/src/org/labkey/experiment/ConfirmDelete.jsp
```diff
@@ -45,9 +45,30 @@
         cancelUrl = successUrl;
 %>
 
+<% if (bean.getReferencedItems().size() > 0) { %>
```
cnathe: looks like `bean.getReferencedItems().size()` is called six times. maybe just stash that in a var?

cnathe (approved):

api/src/org/labkey/api/exp/ExperimentRunListView.java
labkey-susanh: Removed. Thanks.

experiment/src/org/labkey/experiment/ConfirmDelete.jsp
labkey-susanh: Sure. did that for other size values we check within this page too.

labkey-susanh:
Assay tests are no passing on this branch: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_ModuleSuites_LuminexPostgres/1949705

PR #3613, opened 2022-08-16
Fix URL handling in DataRegion, ExperimentMembershipDisplayColumn, etc.

labkey-adam:
#### Rationale
The MS2Test has been failing when the no-question-mark-urls experimental feature is turned on. I tracked this down to Customize Grid -> Apply View redirecting to an invalid URL (e.g., /labkey/MS2VerifyProject/ms2folder/ms2-showList.view?%2Flabkey%2FMS2VerifyProject%2Fms2folder%2Fms2-showList.view). I believe this is caused by faulty logic in the `DataRegion.js` `_getParameters()` function, which checks for the presence of a '?' to extract the query string, but then treats the whole URL as a query string if no '?' is found. This results in the code thinking the entire URL is a parameter with a blank value. Hilarity ensues.

This is a very simple fix (move a single curly brace), but best reviewed with whitespace changes ignored.

Note: Before merging, I will verify that the usual suite of platform tests as well as MS2Test pass. I'll also comment out the hard-coded setting of the experimental feature, since we're not ready for that on develop quite yet.

See [Issue 1713: Suppress trailing ? in URL strings if no parameters](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=1713)

#### Changes
* Fix DataRegion._getParameters() logic when ? is not present on current URL
* Adjust function `setSearchString()` - this was an optional change, but we're moving toward not appending '?' unless we have parameters
* Delete function `getSearchString()` - this was also an optional change, but the function appears deprecated, unused, and broken in the no-question-mark case
* Fix URL handling in ExperimentMembershipDisplayColumn, AttachmentServiceImpl, etc.


cnathe (approved):
The DataRegion.js change looks right to me for the handling of the "?" in the qString to determine parameters.

PR #3620, opened 2022-08-18
Enabling product tier configuration with product groups and product features 

labkey-susanh:
#### Rationale
The LKSM Starter product does not include the Workflow and Assay functionality,  but there are also cases where we want to turn on one or the other of these products and, in general, want a mechanism for the server to communicate a set of features that are enabled so we can check for feature availability when deciding to present functionality in applications.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/936

#### Changes
* Add ProductConfiguration class and Admin console page for choosing a product from a group
* Add ProductFeature enum
* Add product groups registry in AdminConsole

### DIFF
--- a/api/src/org/labkey/api/settings/AdminConsole.java
+++ b/api/src/org/labkey/api/settings/AdminConsole.java
@@ -25,6 +25,7 @@ import org.labkey.api.view.ActionURL;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -62,6 +63,7 @@ public class AdminConsole
 
     private static final Map<SettingsLinkType, Collection<AdminLink>> _links = new HashMap<>();
     private static final Set<ExperimentalFeatureFlag> _experimentalFlags = new ConcurrentSkipListSet<>();
+    private static final Set<ProductGroup> _productGroups = new ConcurrentSkipListSet<>();
 
     static
     {
@@ -174,6 +176,7 @@ public class AdminConsole
             return _title;
         }
 
+        @Override
         public String getDescription()
         {
             return _description;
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/settings/AdminConsole.java
```diff
@@ -204,4 +207,58 @@ public String getPropertyName()
...
+    {
+        _productGroups.add(group);
+    }
+
+    public static Set<ProductGroup> getProductGroups()
+    {
+        return Collections.unmodifiableSet(_productGroups);
+    }
+
+    public static Set<String> getProductFeatureList()
```
labkey-adam: `getProductFeatureSet()`?

api/src/org/labkey/api/settings/ProductFeature.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * A listing of functionality that is available in the non-starter tiers of our applications
+ */
+public enum ProductFeature
+{
+    Assay("assay"),
+    ELN("labbook"),
+    FreezerManagement("inventory"),
+    SampleManagement("sampleManager"),
```
labkey-adam: sampleManager -> sampleManagement?

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -8741,6 +8744,89 @@ public void addNavTrail(NavTree root)
...
+        public void setProductKey(String productKey)
+        {
+            this.productKey = productKey;
+        }
+
+    }
+
+    @AdminConsoleAction
+    @RequiresPermission(AdminOperationsPermission.class)
+    public class ProductConfigurationAction extends FormViewAction<Object>
```
labkey-adam: Could be a `SimpleViewAction`, I think. Is `getSuccessURL()` ever called?

labkey-adam (approved):
A few suggestions, but no need for re-review

PR #3622, opened 2022-08-19
Convert PipelineJob.serializeJob from static to an instance method

bbimber:
@labkey-jeckels This is related to: https://www.labkey.org/ONPRC/Support%20Tickets/issues-update.view?issueId=46134. I am targeted 22.3 since this is where i wrote it, but in the ticket you mentioned develop and i could live with that. I dont know if other repos use serialize? Maybe panorama or another pipeline-heavy module?


labkey-jeckels:
@bbimber I'd prefer to not do this against 22.3 since we're not actively maintaining that branch anymore. I've created a FB and PR against develop with the change. We'll see if TeamCity finds any other modules that need updating, but I don't have any other code locally that's impacted. https://github.com/LabKey/platform/pull/3636

PR #3630, opened 2022-08-23
45614 : Add ${OriginalSourcePath} replacement into pipeline job

labkey-klum:
#### Rationale
Currently, if a custom script pipeline job is started via the filebrowser import button the `${OriginalSourcePath}` token is not available but if run from a file watcher that token is available. In the former case, while the input file is available, it is relative from the temp working directory and would need to be normalized if they needed the full path. 

From a convenience and consistency standpoint we want to support the token in both scenarios. [Docs](https://www.labkey.org/Documentation/wiki-page.view?name=rPipeline&_docid=wiki%3A32d7097b-ed56-1034-b734-fe851e088836) seem to suggest that the replacement is also available.

[Related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45614)


labkey-jeckels (approved):
Is this the best behavior when there are multiple input files? Or is that never the case in this use case?

labkey-klum:
@labkey-jeckels, agreed that it's probably not the best way to handle multiple input files, but as you mentioned it's not likely for this particular scenario. One thing I did consider for the multiple input case is to add replacements like : ${OriginalSourcePath[0]}, ${OriginalSourcePath[1]} etc.

PR #3633, opened 2022-08-24
Fix UploadModuleTest.testUpdateModule failure

labkey-jeckels:
#### Rationale
Holding a reference to a module becomes a leak after that module is deleted.

#### Changes
* Use string references instead and resolve to module objects when needed


labkey-adam:
FYI: I kicked off re-runs of the failed suites in this branch

labkey-adam (approved):

labkey-jeckels:
@labkey-adam Thanks. I got a pass on the suite that includes this test, and had requeued the BVT-A/B on Postgres. All were green so I've merged.

PR #3636, opened 2022-08-25
Convert PipelineJob.serializeJob from static to an instance method

labkey-jeckels:
#### Rationale
There's no reason for PipelineJob.serializeJob() to be static, and we'd like to provide a convenient mechanism for subclasses to be able to customize their behavior.

#### Changes
* Switch from static to instance method


bbimber:
@labkey-jeckels, thanks, seems good to me

labkey-jeckels:
@bbimber build and test results look good, so no needs for edits in other repos. I'll merge to develop after I get a PR approval from someone

labkey-susanh (approved):

PR #3640, opened 2022-08-25
Fix credit error checking in production mode

labkey-adam:
#### Rationale
A change I made yesterday broke special handling we had in place for `dependencies.txt` vs. `jars.txt` comparisons in production mode. Instead of restoring the special handling, I made production mode work the same as dev mode... removing additional special handling in `DefaultModule`. This special handling was likely needed before the introduction of `dependencies.txt`, but isn't now.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3634

#### Changes
* With the exception of Tomcat jars (which aren't worth looking for in production mode), handle `dependencies.txt` vs. `jars.txt` validation the same in production and dev modes.
* Update Tomcat jar names to the latest.
* Catch an additional error condition that was previously missed: a jars.txt file with no jars listed.

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -1037,6 +1037,9 @@ public class AdminController extends SpringActionController
 
         private @NotNull String getErrors(@Nullable String wikiSource, String creditsFilename, Collection<String> foundFilenames, String fileType, String foundWhere, @Nullable String wikiSourceSearchPattern)
         {
+            if (foundFilenames.isEmpty() && null != wikiSource && "jars.txt".equals(creditsFilename))
+                return WIKI_LINE_SEP + "**WARNING: jars.txt file exists when no external jars are present in " + _component + "**";
+
             Set<String> documentedFilenames = new CaseInsensitiveTreeSet();
 
             if (null != wikiSource && null != wikiSourceSearchPattern)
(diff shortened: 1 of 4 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -1037,6 +1037,9 @@ private static class CreditsView extends WebPartView
 
         private @NotNull String getErrors(@Nullable String wikiSource, String creditsFilename, Collection<String> foundFilenames, String fileType, String foundWhere, @Nullable String wikiSourceSearchPattern)
         {
+            if (foundFilenames.isEmpty() && null != wikiSource && "jars.txt".equals(creditsFilename))
+                return WIKI_LINE_SEP + "**WARNING: jars.txt file exists when no external jars are present in " + _component + "**";
```
labkey-susanh, line 1041: Will this trigger a TC test failure so we can fix it?

labkey-susanh (approved):

core/src/org/labkey/core/admin/AdminController.java
labkey-adam: Yes, just like the warnings below. Any warnings on this page result in a failure.

labkey-adam: So basic, it's in the `BasicTest`

PR #3645, opened 2022-08-27
Upgrade several dependencies

labkey-adam:
#### Rationale
Accommodate javax.servlet-api upgrade

#### Related Pull Requests
* https://github.com/LabKey/server/pull/310

#### Changes
* New coordinates for new version
* Implement a few new methods in servlet wrapper classes

### DIFF
--- a/api/src/org/labkey/api/module/SpringModule.java
+++ b/api/src/org/labkey/api/module/SpringModule.java
@@ -499,6 +499,12 @@ public abstract class SpringModule extends DefaultModule
         }
 
         @Override
+        public ServletRegistration.Dynamic addJspFile(String servletName, String jspFile)
+        {
+            return null;
+        }
+
+        @Override
         public <T extends Servlet> T createServlet(Class<T> aClass)
         {
             return null;
@@ -602,6 +608,48 @@ public abstract class SpringModule extends DefaultModule
 
         @Override
         public void declareRoles(String... strings)
+        {
+
+        }
+
+        @Override
+        public String getVirtualServerName()
+        {
+            return null;
+        }
+
+        @Override
+        public int getSessionTimeout()
+        {
+            return 0;
+        }
+
+        @Override
+        public void setSessionTimeout(int sessionTimeout)
+        {
+
+        }
+
+        @Override
+        public String getRequestCharacterEncoding()
+        {
+            return null;
+        }
+
+        @Override
+        public void setRequestCharacterEncoding(String encoding)
+        {
+
+        }
+
+        @Override
+        public String getResponseCharacterEncoding()
+        {
+            return null;
+        }
+
+        @Override
+        public void setResponseCharacterEncoding(String encoding)
         {
 
         }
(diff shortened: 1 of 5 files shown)


core/src/org/labkey/core/CoreModule.java
```diff
@@ -293,11 +293,9 @@ public class CoreModule extends SpringModule implements SearchService.DocumentPr
 
     static
     {
-        // Accept most of the standard Quartz properties, but set a system property to skip Quartz's update check.
-        // These properties need to be set here (previously set in startBackgroundThreads()), so that if any other
-        // module touches Quartz in its setup, it initializes with these setting.
+        // Accept most of the standard Quartz properties, but set the misfire threshold to five minutes. Set it early so
+        // if any other module touches Quartz in its setup, it initializes with this setting.
```
labkey-susanh, line 299: Dunno much about our use of Quartz, so I'll defer to, uh, Josh on this one.

labkey-adam: I don't know anything about the "misfire threshold" change but somebody thought it was important (I think Binal added this, so likely an EHR thing).

BTW: The quartz upgrade check that I disabled a while back has been removed, so that flag is not available in the newest version.

api/src/org/labkey/api/module/SpringModule.java
```diff
@@ -606,6 +612,48 @@ public void declareRoles(String... strings)
 
         }
 
+        @Override
+        public String getVirtualServerName()
+        {
+            return null;
```
labkey-jeckels, line 618: Does anyone know why some methods pass through to _wrapper while others no-op and return null? If so, adding a comment would be lovely.

core/src/org/labkey/core/CoreModule.java
```diff
@@ -293,11 +293,9 @@ public class CoreModule extends SpringModule implements SearchService.DocumentPr
 
     static
     {
-        // Accept most of the standard Quartz properties, but set a system property to skip Quartz's update check.
-        // These properties need to be set here (previously set in startBackgroundThreads()), so that if any other
-        // module touches Quartz in its setup, it initializes with these setting.
+        // Accept most of the standard Quartz properties, but set the misfire threshold to five minutes. Set it early so
```
labkey-jeckels:
```suggestion
        // Accept most of the standard Quartz properties, but set the misfire threshold to five minutes. This prevents
        // Quartz from dropping scheduled work if a lot of items fire at the same time, like a lot of ETLs triggering at 2AM.
        // This can overwhelm the thread pool running them so they don't complete in the default 1 minute window. Set it early so
```

core/src/org/labkey/core/CoreModule.java
labkey-jeckels: Suggested a comment update to clarify the misfire threshold setting.

labkey-jeckels (approved):

api/src/org/labkey/api/module/SpringModule.java
labkey-adam: @labkey-jeckels I don't know why there's very little delegation here. It looks like this wrapper is only used to load spring XML files, so I suspect its just enough impl to support that process.

PR #3646, opened 2022-08-29
Turn on no-question-mark experimental feature for all servers

labkey-adam:
#### Rationale
Just following the plan: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=1713

Note: This will be merged after the 22.9 snapshot branch is created.

### DIFF
--- a/core/module.properties
+++ b/core/module.properties
@@ -1,6 +1,6 @@
 Name: Core
 ModuleClass: org.labkey.core.CoreModule
-SchemaVersion: 22.004
+SchemaVersion: 22.005
 Label: Administration and Essential Services
 Description: The Core module provides central services such as login, \
     security, administration, folder management, user management, \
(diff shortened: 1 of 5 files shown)


labkey-jeckels (approved):

labkey-tchad (commented):
What's the difference between `executeJavaInitializationCode` and `executeJavaUpgradeCode`? Other scripts I checked use the same method regardless of postgres/SQLServer.

labkey-adam:
> What's the difference between `executeJavaInitializationCode` and `executeJavaUpgradeCode`? Other scripts I checked use the same method regardless of postgres/SQLServer.

They are functionally identical, but the intent is for `executeJavaInitializationCode` calls to remain after they are consolidated into bootstrap scripts, whereas `executeJavaUpgradeCode` calls (and code) get removed. Although I expect to remove these particular scripts well before consolidation time (because the experimental feature will soon go away), I should have used executeJavaUpgradeCode in both cases... thanks for noticing!

PR #3653, opened 2022-08-31
Issue 46229: Use time-based ordering to jitter points in Panorama QC plots

labkey-jeckels:
#### Rationale
In Panorama QC folder plots, you can group data based on acquisition date. However, this randomly jitters the points within the day.  Like we do in the reproducibility plots, it would be better to order them based on their acquisition time. This would make it clearer when a change happens mid-day. 

#### Changes
* When using "sequential" positioning, bucket based on the grouping (series) so that points are evenly distributed within a series

### DIFF
--- a/internal/webapp/vis/src/internal/D3Renderer.js
+++ b/internal/webapp/vis/src/internal/D3Renderer.js
@@ -1987,21 +1987,25 @@ LABKEY.vis.internal.D3Renderer = function(plot) {
                     jitterIndex[x] = true;
                 } else if (geom.position === position.sequential) {
 
-                    if (!jitterIndex[x]) {
+                    // We calculate based on both the x-axis position and (optionally) within the series as configured via groupBy
+                    const xGrouped = x + (geom.groupBy ? ('-' + row[geom.groupBy]) : '');
+
+                    if (!jitterIndex[xGrouped]) {
                         // Count how many points we have to distribute across the grouping, and reset the current
                         // indices for the groupings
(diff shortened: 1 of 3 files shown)


internal/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -1987,21 +1987,25 @@ LABKEY.vis.internal.D3Renderer = function(plot) {
...
                         // Count how many points we have to distribute across the grouping, and reset the current
                         // indices for the groupings
                         jitters = {};
                         for (var i = 0; i < data.length; i++) {
-                            var x2 = geom.xAes.getValue(data[i]);
-                            var count = jitterIndex[x2] || 0;
-                            jitterIndex[x2] = count + 1;
+                            const x2 = geom.xAes.getValue(data[i]);
+                            const x2Grouped = x2 + (geom.groupBy ? ('-' + data[i][geom.groupBy]) : '');
+                            const count = jitterIndex[x2Grouped] || 0;
```
cnathe, line 2000: even though I assume IntelliJ suggested this, we should still be using "var" here instead of let/const.

cnathe (approved):

internal/webapp/vis/src/internal/D3Renderer.js
cnathe: disregard. We did some testing with the `gradlew compressClientLibs` task and `const` is fine to include

PR #3654, opened 2022-08-31

labkey-bpatel:
#### Rationale
Issue [43374](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43374): Some characters in file names cause a problem when creating exp.data rows in FileImporter

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1226

#### Changes
* Refactor encode special characters ("+" and "%") in container paths - push this encoding into FilesWebPart._getRootPath()
* Compensate for this encoding in Browser.js, Upload.js, Webdav.js for the actions to work as expected in the Files webpart


filecontent/webapp/File/system/Webdav.js
```diff
@@ -215,9 +225,11 @@
 
         getXMLProxyCfg: function ()
         {
+            var contextBaseURL = this.getContextBaseURL().replaceAll("%25", "%").replaceAll("%2B", "+");
```
labkey-jeckels: There are a half dozen or so lines with this same decoding. Can you create a shared code chunk in case we discover another problematic character in the future?

labkey-jeckels: And add comments that point back and forth between the Java and JS code that needs to be kept in sync, too.

labkey-jeckels (changes requested):

filecontent/webapp/File/system/Webdav.js
labkey-bpatel: done!

api/src/org/labkey/api/files/view/FilesWebPart.java
```diff
@@ -409,6 +410,8 @@ private static String _getRootPath(Container c, @Nullable String relativePath, b
         if (!rootPath.endsWith("/"))
             rootPath += "/";
 
+        // Issue 43374: Some characters in file names cause a problem when creating exp.data rows in FileImporter.
+        // This fix requires decoding these characters so that they don't get encoded twice - see getURLWithCharsReplaced() in Webdav.js and its usages.
```
labkey-jeckels: I find this comment confusing, since the line it's attached to is encoding, not decoding.
```suggestion
        // Need to centrally encode characters that are special in URIs but not in most file systems. See getURLWithCharsReplaced() in Webdav.js for special client-side handling of these characters.

```

labkey-jeckels (approved):

PR #3667, opened 2022-09-09
Projects, Enhance UX to disallow cross-project actions earlier

XingY:
#### Rationale
The StatementUtil is not wired up to handle sample or data that exists in other folders. For performance concern, we will not allow cross folder merge. But instead, will throw error if a merge contains data from a different folder. This PR adds a step to the data iterator to check for cross folder data.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/957
* https://github.com/LabKey/platform/pull/3667
* https://github.com/LabKey/inventory/pull/540
* https://github.com/LabKey/sampleManagement/pull/1206
* https://github.com/LabKey/biologics/pull/1578

#### Changes
* added CrossFolderRecordDataIterator to check for cross folder data presence during import with merge
* added GetCrossFolderDataSelectionAction to check for cross folder selection for a data region

### DIFF
--- a/api/src/org/labkey/api/dataiterator/CrossFolderRecordDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/CrossFolderRecordDataIterator.java
@@ -1,0 +1,123 @@
+package org.labkey.api.dataiterator;
+
+import org.jetbrains.annotations.Nullable;
+import org.labkey.api.collections.CaseInsensitiveHashMap;
+import org.labkey.api.data.ColumnInfo;
+import org.labkey.api.data.Container;
+import org.labkey.api.data.TableInfo;
+import org.labkey.api.query.BatchValidationException;
+import org.labkey.api.query.QueryService;
+import org.labkey.api.query.QueryUpdateService;
+import org.labkey.api.query.ValidationException;
+import org.labkey.api.security.User;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Supplier;
+
+// throws error if key values found in related folders
+public class CrossFolderRecordDataIterator extends WrapperDataIterator
+{
+    private final DataIteratorContext _context;
+    final QueryUpdateService qus;
+    final User user;
+    final Container c;
+
+    final CachingDataIterator _unwrapped;
+    final ArrayList<ColumnInfo> pkColumns = new ArrayList<>();
+    final ArrayList<Supplier<Object>> pkSuppliers = new ArrayList<>();
+
+    final Map<String,Object> _extraKeyValueMap;
+
+    int lastPrefetchRowNumber = -1;
+
+    protected CrossFolderRecordDataIterator(DataIterator in, DataIteratorContext context, TableInfo target, @Nullable Set<String> keys, @Nullable Map<String,Object> extraKeyValueMap)
+    {
+        super(in);
+        _context = context;
+        qus = target.getUpdateService();
+        user = target.getUserSchema().getUser();
+        c = target.getUserSchema().getContainer();
+        _extraKeyValueMap = extraKeyValueMap;
+
+        _unwrapped = (CachingDataIterator)in;
+
+        var map = DataIteratorUtil.createColumnNameMap(in);
+        Collection<String> keyNames = null==keys ? target.getPkColumnNames() : keys;
+        for (String name : keyNames)
+        {
+            Integer index = map.get(name);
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/dataiterator/CrossFolderRecordDataIterator.java
```diff
@@ -0,0 +1,123 @@
...
+    final User user;
+    final Container c;
+
+    final CachingDataIterator _unwrapped;
+    final ArrayList<ColumnInfo> pkColumns = new ArrayList<>();
+    final ArrayList<Supplier<Object>> pkSuppliers = new ArrayList<>();
+
+    final Map<String,Object> _extraKeyValueMap;
+
+    int lastPrefetchRowNumber = -1;
```
labkey-susanh, line 35: Is there a reason for mixing the syntax on these class member names? (Some use the `_` prefix and some do not.)

api/src/org/labkey/api/dataiterator/CrossFolderRecordDataIterator.java
```diff
@@ -0,0 +1,123 @@
...
+    public static DataIteratorBuilder createBuilder(DataIteratorBuilder dib, TableInfo target, @Nullable Set<String> keys, @Nullable Map<String,Object> extraKeyValueMap)
+    {
+        return context ->
+        {
+            DataIterator di = dib.getDataIterator(context);
+            if (null == di)
+                return null;
+
+            assert !di.supportsGetExistingRecord();
+            if (di.supportsGetExistingRecord())
```
labkey-susanh: Are we meaning to assure that we've not queried for existing records twice?  A comment about why we do this assertion/if statement here might be useful.

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -1171,5 +1162,30 @@ protected AttachmentParentFactory getAttachmentParentFactory()
...
+            Set<String> dataNames = new HashSet<>();
+            for (Map.Entry<Integer, Map<String, Object>> keyMap : keys.entrySet())
+            {
+                Object oName = keyMap.getValue().get("Name");
+                Object oClassId = keyMap.getValue().get("ClassId");
+
+                if (oName != null && oClassId != null)
+                {
+                    dataNames.add((String) oName);
+                    dataClassId = (Integer) (_converter.convert(Integer.class, oClassId));
```
labkey-susanh: Is `dataClassId` going to change within this loop? If not, we could save a little time by not reassigning if the dataClassId is not null.  If so, then that's a problem when we use this id.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8209,6 +8209,33 @@ public int removeEdges(ExpLineageEdge.FilterOptions options)
...
+
+        TableInfo tableInfo = isSample ? ExperimentService.get().getTinfoMaterial() : ExperimentService.get().getTinfoData();
+        SQLFragment currentFolderCountSql = new SQLFragment()
+                .append(" SELECT COUNT(*) FROM ")
+                .append(tableInfo, "t")
+                .append("\nWHERE Container = ? ")
+                .add(container.getId())
+                .append("\nAND RowId ");
+        dialect.appendInClauseSql(currentFolderCountSql, rowIds);
+        int currentFolderSelectionCount = new SqlSelector(expSchema, currentFolderCountSql).getArrayList(Integer.class).get(0);
```
labkey-susanh, line 8225: Can you do `.getObject(Integer.class)` instead?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -657,6 +657,30 @@ public List<Map<String, Object>> getRows(User user, Container container, List<Ma
...
 
+    @Override
+    public boolean hasExistingRowsInOtherContainers(Container container, Map<Integer, Map<String, Object>> keys)
+    {
+        Integer sampleTypeId = null;
+        Set<String> sampleNames = new HashSet<>();
+        for (Map.Entry<Integer, Map<String, Object>> keyMap : keys.entrySet())
+        {
+            String name = getMaterialName(keyMap.getValue());
+            Integer materialSourceId = getMaterialSourceId(keyMap.getValue());
```
labkey-susanh: Same question here about whether we can save a little processing by not reassigning if the `materialSourceId` is not null.

labkey-susanh (approved):

PR #3668, opened 2022-09-09
46138: NPE in datasets FK

labkey-klum:
#### Rationale
Needed to clean up some assumptions in order to fix the related issue. The main problem is that we can't always assume the source has a sequence number field. Of course if the source isn't a dataset we can only join to demographics datasets in the study.

[Related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46138)

### DIFF
--- a/study/src/org/labkey/study/query/DatasetAutoJoinTable.java
+++ b/study/src/org/labkey/study/query/DatasetAutoJoinTable.java
@@ -68,6 +68,8 @@ public class DatasetAutoJoinTable extends VirtualTable
     // The "_Key" FieldKey that has possibly been remapped.
     private FieldKey _keyFieldKey;
 
+    private ColumnInfo _sequenceNumColumn;
+
     public DatasetAutoJoinTable(StudyQuerySchema schema, ContainerFilter cf, DatasetDefinition source,
                                 @Nullable ColumnInfo participantIdColumn,
                                 @Nullable FieldKey sequenceNumFieldKey,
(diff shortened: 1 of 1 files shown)


study/src/org/labkey/study/query/DatasetAutoJoinTable.java
```diff
@@ -99,18 +101,22 @@ public DatasetAutoJoinTable(StudyQuerySchema schema, ContainerFilter cf, Dataset
...
 
-            // SequenceNum is always available
-            Objects.requireNonNull(parent.getColumn(sequenceNumFieldKey.getName()));
-            var colSequenceNum = new AliasedColumn(parent, "SequenceNum", parent.getColumn(sequenceNumFieldKey.getName()));
-            colSequenceNum.setHidden(true);
-            addColumn(colSequenceNum);
-
-            // The extra key property is not always available.
-            if (_keyPropertyName != null)
+            // Issue: 46138 sequence number isn't always available
```
labkey-matthewb: Maybe add 
    if (!isDemographicData()) {...}
around this whole block?

labkey-matthewb (changes requested):

study/src/org/labkey/study/query/DatasetAutoJoinTable.java
labkey-matthewb: !source.isDemographicData()

labkey-matthewb (approved):

PR #3669, opened 2022-09-10
Updates for import and export of bioloigcs registry and media data

labkey-susanh:
#### Rationale
To enable easier updating of our demo data for LKB, we need to add some updates to the sample type and data class folder writer and importer. To keep some of the special logic within the Biologics module, we introduce a ColumnExporter interface and registry, but the actual import and export of the media and registry data classes stays here.  For media, we exclude these runs and protocols from the experiment XAR exports as they are handled by a new folder writer and importer in the recipe module.

#### Related Pull Requests
* https://github.com/LabKey/recipe/pull/71
* https://github.com/LabKey/biologics/pull/1580

#### Changes
* Introduce `ColumnExporter` interface and registry
* Updates to SampleTypeAndDataClassFolderWriter and Importer to account for special handling of media and registry
* exclude runs and protocols created for media from XAR exports

### DIFF
--- a/experiment/src/org/labkey/experiment/LSIDRelativizer.java
+++ b/experiment/src/org/labkey/experiment/LSIDRelativizer.java
@@ -105,17 +105,22 @@ public enum LSIDRelativizer implements SafeToRenderEnum
                     id = suffix.substring(ind);
                 return lsids.uniquifyRelativizedLSID("urn:lsid:" + XarContext.LSID_AUTHORITY_SUBSTITUTION + ":" + prefix + ".Folder-" + containerSubstitution+ ".${XarJobId}" + id, lsid.getObjectId(), lsid.getVersion());
             }
-            else if (("Sample".equals(prefix) || "Material".equals(prefix) || ("Data".equals(prefix) && !lsid.getObjectId().contains("%2F"))))
+            else if ("Sample".equals(prefix) || "Material".equals(prefix))
             {
                 String xarJobId = ".${XarJobId}"; // XarJobId is more concise than XarFileId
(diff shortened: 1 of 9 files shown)


experiment/src/org/labkey/experiment/LSIDRelativizer.java
```diff
@@ -112,7 +111,7 @@ else if ("Data".equals(prefix))
                 // UNDONE: Maybe there is a better way to detect when we should use ${AutoFileLSID}?
                 return AutoFileLSIDReplacer.AUTO_FILE_LSID_SUBSTITUTION;
             }
-            else if (suffix != null && SUFFIX_PATTERN.matcher(suffix).matches())
+            else if (suffix != null && (SUFFIX_PATTERN.matcher(suffix).matches() || XAR_IMPORT_SUFFIX_PATTERN.matcher(suffix).matches()))
```
Sigmonia, line 114: any concern/need for these pattern matchers should be case insensitive? It looks like the values are internally generated so i don't think so, but just want to confirm.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -7278,6 +7280,19 @@ public void registerObjectReferencer(ObjectReferencer referencer)
...
+    @Override
+    public void registerColumnExporter(ColumnExporter exporter)
+    {
+        _columnExporters.add(exporter);
+    }
+
+    @Override
+    public List<ColumnExporter> getColumnExporters()
+    {
+        return _columnExporters;
```
Sigmonia, line 7292: Should this be unmodifiable or a copy?

Sigmonia (approved):

experiment/src/org/labkey/experiment/LSIDRelativizer.java
labkey-susanh: I thought of that, but I think we're fine leaving as is.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Doesn't seem necessary.

PR #3670, opened 2022-09-10
Issue 46314: Add unique users metric

labkey-jeckels:
#### Rationale
We'd like better insights into how many users are logging in regularly.

#### Changes
* Add new metric for the unique number of users in the current month

### DIFF
--- a/core/src/org/labkey/core/CoreModule.java
+++ b/core/src/org/labkey/core/CoreModule.java
@@ -275,7 +275,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import static org.labkey.api.settings.StashedStartupProperties.homeProjectFolderType;
--- a/core/src/org/labkey/core/CoreModule.java
+++ b/core/src/org/labkey/core/CoreModule.java
@@ -264,15 +264,18 @@ import java.sql.Connection;
 import java.sql.SQLException;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/CoreModule.java
```diff
@@ -1032,6 +1034,9 @@ public void moduleStartupComplete(ServletContext servletContext)
             results.put("totalLogins", new TableSelector(userAuditTable, new SimpleFilter(FieldKey.fromParts("comment"), UserManager.UserAuditEvent.LOGGED_IN, CompareType.CONTAINS), null).getRowCount());
             results.put("userLimits", new LimitActiveUsersSettings().getMetricsMap());
             results.put("systemUserCount", UserManager.getSystemUserCount());
+            Calendar cal = new GregorianCalendar();
+            cal.set(cal.get(Calendar.YEAR), cal.get(Calendar.MONTH), 1, 0, 0, 0);
+            results.put("uniqueUserCountThisMonth", UserManager.getUniqueUsersCount(cal.getTime()));
```
labkey-adam, line 1039: This will be a count for current calendar month, right? Would a rolling "last 30 days" (or some other fixed time period) be easier to analyze for long-term trend purposes?

labkey-adam (approved):

core/src/org/labkey/core/CoreModule.java
labkey-jeckels: I asked the services team, who requested this metric, that same question. This was their preference.

We have a few other login-related metrics that currently look at the last 30 days. I'm trying to capture these behaviors in the notes for each metric and think that we should ultimately standardize after seeing how this new metric works out in practice.

PR #3671, opened 2022-09-12
Issue 46122: Pipeline actions don't return to the starting URL, instead redirect to folder homepage

labkey-jeckels:
#### Rationale
We used to get users back to the folder tab they started via .lastFilter. That not longer remembers the tab name, and users have requested that this particular workflow continue to work the same.

#### Changes
* Pull the returnUrl through the full sequence and send the user back there

### DIFF
--- a/api/src/org/labkey/api/pipeline/file/FileAnalysisTaskPipeline.java
+++ b/api/src/org/labkey/api/pipeline/file/FileAnalysisTaskPipeline.java
@@ -75,7 +75,7 @@ public interface FileAnalysisTaskPipeline extends TaskPipeline<FileAnalysisTaskP
     FilePathFilter getInitialFileTypeFilter();
 
     @NotNull
-    URLHelper getAnalyzeURL(Container c, String path, ReturnURLString parsedReturnUrl);
+    URLHelper getAnalyzeURL(Container c, String path, @Nullable ReturnURLString parsedReturnUrl);
 
     @NotNull
     Map<FileType, List<FileType>> getTypeHierarchy();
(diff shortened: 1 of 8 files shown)


filecontent/webapp/File/panel/Browser.js
```diff
@@ -275,8 +275,13 @@ Ext4.define('File.panel.Browser', {
          * @private
          */
         _getActions : function(cb, containerPath, scope) {
+            var params = {path : scope.getFullFolderOffset() };
+            var returnUrl = LABKEY.ActionURL.getParameter('returnUrl');
```
labkey-nicka: There is another function for this as well. `LABKEY.ActionURL.getReturnUrl()`. Same feedback for usage in `analyze.jsp`.

api/src/org/labkey/api/pipeline/file/FileAnalysisTaskPipeline.java
```diff
@@ -74,7 +75,7 @@ interface FilePathFilter extends FileFilter, DirectoryStream.Filter<Path>
     FilePathFilter getInitialFileTypeFilter();
 
     @NotNull
-    URLHelper getAnalyzeURL(Container c, String path);
+    URLHelper getAnalyzeURL(Container c, String path, ReturnURLString parsedReturnUrl);
```
labkey-nicka: Denote `@Nullable` on `parsedReturnUrl`.

labkey-nicka (approved):
Looks good. I presume it's known why `.lastFilter` no longer supports this and/or is not still supported.

labkey-jeckels:
@labkey-nicka here's the change that stopped returning to the original folder tab https://github.com/LabKey/platform/pull/3426

PR #3684, opened 2022-09-16
Issue 45368: Attachment fields containing field names with spaces doesn't show attachment

labkey-bpatel:
#### Rationale
Issue [45368](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45368): Attachment fields containing field names with spaces doesn't show attachment 

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/1240

#### Changes
* Get column alias when the name (for attachment column with spaces) is not present in the statement parameter

### DIFF
--- a/api/src/org/labkey/api/dataiterator/StatementDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/StatementDataIterator.java
@@ -199,8 +199,6 @@ public class StatementDataIterator extends AbstractDataIterator
                     to = stmt.getParameter(col.getPropertyURI());
                 if (to == null)
                     to = stmt.getParameter(col.getName());
-                if (to == null)
-                    to = stmt.getParameter(col.getAlias());//Issue 45368: Attachment fields containing field names with spaces doesn't show attachment
                 if (null != to)
                 {
                     FieldKey mvName = col.getMvColumnName();
--- a/api/src/org/labkey/api/dataiterator/StatementDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/StatementDataIterator.java
@@ -199,6 +199,8 @@ public class StatementDataIterator extends AbstractDataIterator
                     to = stmt.getParameter(col.getPropertyURI());
                 if (to == null)
                     to = stmt.getParameter(col.getName());
+                if (to == null)
+                    to = stmt.getParameter(col.getAlias());
                 if (null != to)
                 {
                     FieldKey mvName = col.getMvColumnName();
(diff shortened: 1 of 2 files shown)


labkey-jeckels (commented):
@labkey-matthewb this seems OK to me but would appreciate your quick glance too

api/src/org/labkey/api/dataiterator/StatementDataIterator.java
```diff
@@ -199,6 +199,8 @@ void init()
                     to = stmt.getParameter(col.getPropertyURI());
                 if (to == null)
                     to = stmt.getParameter(col.getName());
+                if (to == null)
```
labkey-matthewb: Before making this change, I'm curious why this column wasn't correctly matched in DataIteratorUtil.matchColumns().  What is different about this case?

labkey-matthewb (changes requested):

api/src/org/labkey/api/dataiterator/StatementDataIterator.java
labkey-bpatel: It matches in DataIteratoUtil.matchColumns(). 
However, the stmt map here doesn't contain the field with spaces (ex. 'attach field 1'), it contains the field with underscores instead (ex 'attach_field_1').

It isn't clear exactly where the column name with underscores get set - putting a breakpoint in StatementDataIterator ctor (line 105), did not give any clues - it is called from TableInsertDataIterator (line 147), but it doesn't pass in the value for ParameterMapStatement argument. Maybe i am missing something obvious. Your insight is appreciated.

labkey-matthewb: I see.  I have a suspicion about this.  The SQL and ParameterMapStatement are generated in StatementUtils.createStatement().  That code uses UpdateableTableInfo.remapSchemaColumns() to handle the case where a column name in the user schema is different than the column name in the underlying database schema.  That could be the failure point.

labkey-bpatel: Thanks for the insight.

Looks like StatementUtils.createStatement() calls createParameter() (~line 743) to create, well, a Parameter, during which it ends up calling BaseColumnInfo.jdbcRsNameFromName(), which replaces spaces with underscores.  It now makes sense _why_ Statement parameter contains field with underscores instead of the actual col with spaces.

ListTable.remapSchemaColumns() returns an empty map, btw.

labkey-bpatel: I've reverted this change

PR #3685, opened 2022-09-16
Differentiate metrics from servers sharing a cloned DB

labkey-jeckels:
#### Rationale
Our metric tracking gets confused when we have multiple servers using DBs with the same ServerInstallationGUID. That's common for staging and production servers since they frequently take a DB backup and restore it. Less commonly, one production server forks from another.

Reviewing actual metric data, we reliably pull the server host name and can use it to distinguish these instances from each other.

#### Changes
* Identify servers based on GUID and host name
* Stop caring if an install used the Windows installer, which is long dead
* Track more per-instance properties, like organization name and server IP, solely the session level as they can change over time


labkey-jeckels:
@labkey-martyp happy to walk through this with you, as the nominal Mothership feature owner.

labkey-martyp (approved):
Thanks for the walkthru. Looks good.

PR #3686, opened 2022-09-16
Fix url passing to query webpart ajax call

ankurjuneja:
#### Rationale
Labkey Querywebpart api does not filter pound symbol which is often used in the urls and sends it as a parameter value back to the server when updating the grid on container filters.

#### Related Pull Requests
* https://github.com/LabKey/MacCossLabModules/pull/249

#### Changes
* strip off everything after a pound for a query parameter

### DIFF
--- a/api/webapp/clientapi/dom/DataRegion.js
+++ b/api/webapp/clientapi/dom/DataRegion.js
@@ -3139,10 +3139,11 @@ if (!LABKEY.DataRegions) {
         if (LABKEY.Utils.isString(qString) && qString.length > 0) {
 
             var qmIdx = qString.indexOf('?');
+            var qStringWithoutPound;
             if (qmIdx > -1) {
                 qString = qString.substring(qmIdx + 1);
                 var poundIdx = qString.indexOf('#');
-                var qStringWithoutPound = poundIdx > -1 ? qString.substr(0, poundIdx) : qString;
+                qStringWithoutPound = poundIdx > -1 ? qString.substr(0, poundIdx) : qString;
             }
 
             if (qStringWithoutPound.length > 1) {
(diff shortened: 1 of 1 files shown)


api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -3152,7 +3152,9 @@ if (!LABKEY.DataRegions) {
                 });
 
                 $.each(pairs, function(i, pair) {
-                    p = pair.split('=', 2);
+                    var poundIdx = pair.indexOf('#');
```
labkey-jeckels: I think we should strip this earlier - right after we do the ? handling. There's no need to check each parameter separately.

labkey-jeckels (changes requested):

api/webapp/clientapi/dom/DataRegion.js
ankurjuneja: done

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -3141,10 +3141,12 @@ if (!LABKEY.DataRegions) {
             var qmIdx = qString.indexOf('?');
             if (qmIdx > -1) {
                 qString = qString.substring(qmIdx + 1);
+                var poundIdx = qString.indexOf('#');
+                var qStringWithoutPound = poundIdx > -1 ? qString.substr(0, poundIdx) : qString;
             }
 
-            if (qString.length > 1) {
-                var pairs = qString.split('&'), p, key,
+            if (qStringWithoutPound.length > 1) {
```
labkey-jeckels: I'm surprised this works - `qStringWithoutPound` is defined inside the scope of the if() body. But then again JS scoping is weird. I'd simplify ever so slightly and just reassign `qString`. We don't have a `qStringAfterQuestion`, for example.

labkey-jeckels (approved):

PR #3689, opened 2022-09-16
Add metrics for study timepoint types

labkey-adam:
#### Rationale
Study timepoint metrics could be useful: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46331

Reload columns are no longer used: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46316

Unrelated: fix the MySQL driver lookup name (not currently used, but might as well not be wrong) and remove a few HttpClient 4.x references.


labkey-jeckels (approved):
Did you mean to include the MySQL driver and HttpClient changes here? Fine if so, but maybe add to the PR description

labkey-adam:
> Did you mean to include the MySQL driver and HttpClient changes here? Fine if so, but maybe add to the PR description

Yes, since these were trivial changes just waiting for a PR. I'll update the comment.

PR #3690, opened 2022-09-16
Issue 46264: Improve handling of schema XML vs DB schemas reconciliation

labkey-jeckels:
#### Rationale
Mismatches between the expected and actual DB schema causes all sorts of headaches and bugs.

#### Changes
* Proactively check for problems shortly after startup and report to admins.


labkey-adam (approved):
Also tested locally

PR #3691, opened 2022-09-17
Issue 46346: Perf issues rendering HTML wikis

labkey-jeckels:
#### Rationale
We've seen servers get bogged down with wiki rendering.

#### Changes
* Don't double-render the page, once for HTML and once for ClientDependencies. Do it once for both and use our caching mechanism.
* Reduce usage of wildly inefficient DOM NodeList implementation method
* Don't cache as many wikis - they're potentially huge


labkey-adam (approved):
Code looks good, but why 22.9? Are we actually planning to release a 22.9.1?

labkey-jeckels:
I'm not anticipating a 22.9.1 release but there's another fix I think we may push to labkey.org via a snapshot build. https://github.com/LabKey/premium/pull/338

PR #3693, opened 2022-09-20
Projects, Cross-Projects Configuration / Setup UX

labkey-nicka:
#### Rationale
Factor our folder renaming into `ContainerManager` to make consistent logic available to other server-side services.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/961
* https://github.com/LabKey/platform/pull/3693
* https://github.com/LabKey/biologics/pull/1610
* https://github.com/LabKey/sampleManagement/pull/1223
* https://github.com/LabKey/inventory/pull/547

#### Changes
* Factor out folder renaming logic to `ContainerManager`.

### DIFF
--- a/api/src/org/labkey/api/data/ContainerManager.java
+++ b/api/src/org/labkey/api/data/ContainerManager.java
@@ -198,6 +198,11 @@ public class ContainerManager
         return getRoot();
     }
 
+    private static DbScope.Transaction ensureTransaction()
+    {
+        return CORE.getSchema().getScope().ensureTransaction(DATABASE_QUERY_LOCK);
+    }
+
 
     private static int getNewChildSortOrder(Container parent)
     {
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -1579,6 +1579,54 @@ public static boolean move(Container c, final Container newParent, User user) th
...
+            if (c.getParent().isRoot())
+                throw new IllegalArgumentException("The server already has a project with this name.");
+            else
+                throw new IllegalArgumentException("The " + (c.getParent().isProject() ? "project " : "folder ") + c.getParent().getPath() + " already has a folder with this name.");
+        }
+
+        try (DbScope.Transaction tx = CORE.getSchema().getScope().ensureTransaction(DATABASE_QUERY_LOCK))
+        {
+            // Rename
+            rename(c, user, folderName);
```
labkey-klum: Does it make sense to make the old rename function private and force users to use the new variant? Or if the two functions can be combined even further?

labkey-klum (approved):

api/src/org/labkey/api/data/ContainerManager.java
labkey-nicka: I've consolidated a bit further and inverted the call structure between these two functions.

PR #3695, opened 2022-09-20
Issue 46354: Study import overwriting system properties

labkey-martyp:
#### Rationale
When swapping between system properties and non-system properties for dataset columns in a study import, the study properties get overwritten to local property descriptor values.

#### Changes
In study import check if changing to or from system property and handle that as a property descriptor swap between the system property and a local property descriptor (to or from).  Let OntologyManager handle finding the system property or creating the local property descriptor and any property descriptor cleanup after swap.

### DIFF
--- a/api/src/org/labkey/api/exp/property/DomainProperty.java
+++ b/api/src/org/labkey/api/exp/property/DomainProperty.java
@@ -74,6 +74,7 @@ public interface DomainProperty extends ImportAliasable, MutableColumnConceptPro
     void setFormat(String s);
     void setRequired(boolean b);
     void setHidden(boolean hidden);
+    void setSwap(boolean swap);
     void setShownInInsertView(boolean shown);
     void setShownInUpdateView(boolean shown);
     void setShownInDetailsView(boolean shown);
--- a/api/src/org/labkey/api/exp/property/DomainProperty.java
+++ b/api/src/org/labkey/api/exp/property/DomainProperty.java
@@ -74,7 +74,6 @@ public interface DomainProperty extends ImportAliasable, MutableColumnConceptPro
     void setFormat(String s);
     void setRequired(boolean b);
     void setHidden(boolean hidden);
-    void setSwap(boolean swap);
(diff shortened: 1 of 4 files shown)


labkey-jeckels (commented):
Is this a scenario we need to support, or can we detect it and reject the change? That would be simpler and still prevent us from getting into a bad state. I'd like to avoid needing to carry the "swap" state around on the property descriptor.

Do we separately need to improve the scenario where you bootstrap a server, import an EHR study that uses system properties, and then delete it? Does that end up deleting the PropertyDescriptor records for the system properties, as I believe we discussed last week?

labkey-martyp:
> Is this a scenario we need to support, or can we detect it and reject the change? That would be simpler and still prevent us from getting into a bad state. I'd like to avoid needing to carry the "swap" state around on the property descriptor.
> 
> Do we separately need to improve the scenario where you bootstrap a server, import an EHR study that uses system properties, and then delete it? Does that end up deleting the PropertyDescriptor records for the system properties, as I believe we discussed last week?

No I cannot find any cases where the system property descriptor is being deleted when deleting a project, domain or column.  The code for this appears to all be container scoped and the system property descriptors are created in the Shared project. The issue really is from changing to/from system properties on study import.

The swap boolean isn't great, it's just there because there's a matching scenario with a different expected result in the FDA mobile app.  

The intent here was to allow the conversion to/from system properties without losing data.  Not supporting this use case will require deleting the entire dataset, at least when going from a system property.  But we have a similar error when changing the PK of a dataset, requiring the user to truncate first.  So the swap code in this PR isn't really necessary if we don't want to support it.  I can just throw an error.  That is easier to maintain.

labkey-martyp:
@labkey-jeckels Ok I made the changes we discussed, to get rid of the _swap variable on the DomainProperty and instead just check if changing to/from a system property.  I verified it fixes the issue and I tested across most tests on TC with no new failures.

labkey-jeckels (approved):

PR #3697, opened 2022-09-20
Use ancestor name in sample type naming pattern

XingY:
#### Rationale
We support using parent input or parent lookup (fields other than name) in name expression. Direct parents are provided during new sample creation so it's conveniently available to use during name generation. In some scenarios, it's also desirable to use name or properties from the parent of the parent. This PR allows querying for up to 5 generations of ancestor lookup values in name generation.  

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1245
* https://github.com/LabKey/sampleManagement/pull/1226

#### Changes
* Support ancestor lookup for up to 5 generations in sample name generation
* Use ExpLineageOptions.LineageExpType enum instead of string for lineage exp type
* Added ExpLineageTree to format ExpLineage result into a tree structure


api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -825,10 +875,15 @@ private void initialize(@Nullable Map<String, String> importAliases)
...
+                if (token instanceof FieldKey fkTok)
                 {
-                    FieldKey fkTok = (FieldKey)token;
-                    List<String> fieldParts = fkTok.getParts();
+                    int previousErrorCount = _syntaxErrors.size();
+                    List<String> fieldParts = processFieldParts(fkTok, partAncestorOptions, user, dataClassLSIDs, sampleTypeLSIDs);
+                    if (!_syntaxErrors.isEmpty() && _syntaxErrors.size() > previousErrorCount) // if ancestor lookup syntax error, continue
+                        continue;
+                    List<Pair<ExpLineageOptions.LineageExpType, String>> ancestorPaths = null;
+                    if (partAncestorOptions.get(fkTok.encode()) != null)
```
labkey-susanh: Perhaps stash the encoded value in a variable and use `partAncestorOptions.containsKey()`? Or is there a chance the value will be null even when the key is there?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1732,6 +1942,13 @@ protected StringExpressionFactory.StringPart parsePart(String expression)
                 return new NameGenerator.CounterExpressionPart(namePrefixExpression, startInd, numberFormat, _container, _getNonConflictCountFn, _counterSeqPrefix);
             }
 
+            // if contains ancestor expression, substitute ..[MaterialInputs/type1] with ..[MaterialInputs::type1] before parsing, to avoid splitting at /.
```
labkey-susanh, line 1947: This is unfortunate that we have to get around our own syntax in this way.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -2469,6 +2699,15 @@ public void testNameExpressionPreviews()
             verifyPreview("S-${Inputs/SampleTypeBeingCreated/name}", "S-parentname");
             verifyPreview("S-${parentAlias}", "S-Parent101", Collections.singletonMap("parentAlias", "MaterialInputs/SampleTypeA"), null);
             verifyPreview("S-${parentAlias/name}", "S-parentname", Collections.singletonMap("parentAlias", "MaterialInputs/SampleTypeA"), null);
+
+            // ancestor lookup
+            verifyPreview("S-${MaterialInputs/..[MaterialInputs]/name}", "S-parentname");
```
labkey-susanh: Should it be "ancestorName" instead of "parentName"?

api/src/org/labkey/api/exp/api/ExpLineageOptions.java
```diff
@@ -15,16 +15,46 @@
...
+
+        LineageExpType(String type)
+        {
+            this._type = type;
+        }
+
+        public static @Nullable LineageExpType fromValue(String value)
+        {
+            for (LineageExpType type : LineageExpType.values()) {
+                if (type._type.equalsIgnoreCase(value)) {
```
labkey-susanh: I'm curious what the use case is that requires the lower case comparison and thus the deviation from the usual `fromValue` for the enum?

api/src/org/labkey/api/exp/api/ExpLineageOptions.java
```diff
@@ -15,16 +15,46 @@
...
  */
 public class ExpLineageOptions extends ResolveLsidsForm
 {
+    public enum LineageExpType
+    {
+        ALL("ALL"),
+        Data("Data"),
+        Material("Material"),
+        ExperimentRun("ExperimentRun"),
+        Object("Object");
```
labkey-susanh: Since the "type" values are the same as the enum names, is it actually required?

labkey-susanh (approved):

api/src/org/labkey/api/data/NameGenerator.java
XingY: good point. updated

api/src/org/labkey/api/exp/api/ExpLineageOptions.java
XingY: I thought about relaxing api param casing but on second thought it's probably not needed/good idea. Updated to remove ignoreCase

labkey-susanh: Seems like the fromValue could just be removed altogether then, in favor of `valueOf` (which is what I meant to say originally).  And then this enum could just be a simple enum without the _type member.

XingY: The problem with valueOf is, it throws IllegalArgumentException when the param doesn't match a enum value. expType is not validated in the api so it can be anything.

PR #3702, opened 2022-09-21
Assay design metric for counts of number of sample lookups

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46345

Counts both lookups to the "samples" schema and "exp.Materials" table.

- batchDomain -> number of assay design sample lookup fields in the Batch domain
- runDomain -> number of assay design sample lookup fields in the Run domain
- resultDomain -> number of assay design sample lookup fields in the Results domain
- resultDomainWithMultiple -> number of assay design Results domains with more than one sample lookup field


cnathe:
Screen Shot 2022-09-21 at 2 19 05 PM

labkey-susanh (approved):

PR #3707, opened 2022-09-22
Attempt to fix OutOfMemoryError

labkey-jeckels:
#### Rationale
https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyDPostgres/2014795?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildProblemsSection=true&expandBuildTestsSection=true

#### Changes
* Fix code that merges old and new data


mothership/src/org/labkey/mothership/MothershipManager.java
```diff
@@ -426,6 +426,7 @@ else if (StringUtils.isEmpty(currentValue))
         ObjectMapper mapper = new ObjectMapper();
         try
         {
+            log.info("Merging JSON. Old is " + currentValue + " characters, new is " + newValue.length());
```
labkey-tchad: Should this be logged at debug level or are you just planning on removing it before merging?
```suggestion
            log.debug("Merging JSON. Old is " + currentValue + " characters, new is " + newValue.length());

```

labkey-tchad (approved):

mothership/src/org/labkey/mothership/MothershipManager.java
labkey-adam: Also, did you mean `currentValue.length()`?

labkey-adam (changes requested):
Update logging

mothership/src/org/labkey/mothership/MothershipManager.java
labkey-jeckels: Yes

labkey-adam (approved):

PR #3710, opened 2022-09-23
Add new org.json JSON library plus a few usages

labkey-adam:
#### Rationale
We're migrating to a new version of `org.json.JSONObject` / `org.json.JSONArray`. See https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46305.

#### Changes
* Add dependency for current org.json.JSONObject and org.json.JSONArray library
* Fix `ApiJsonWriter`, `ApiSimpleResponse`, `SimplaeApiJsonForm`, `AuthenticationProvider`, and other classes to support new JSON library
* Register Jackson seriailzers / deserializers for new JSON library
* Indicate that new JSONObject and JSONArray are safe-to-render in JSPs

### DIFF
--- a/api/src/org/labkey/api/jsp/LabKeyJspWriter.java
+++ b/api/src/org/labkey/api/jsp/LabKeyJspWriter.java
@@ -46,7 +46,7 @@ public class LabKeyJspWriter extends JspWriterWrapper
     public void print(Object obj) throws IOException
     {
         // These are the only objects we consider safe-to-render
-        if (null == obj || obj instanceof SafeToRender || obj instanceof Number || obj instanceof Boolean | obj instanceof JSONObject || obj instanceof JSONArray)
+        if (null == obj || obj instanceof SafeToRender || obj instanceof Number || obj instanceof Boolean || obj instanceof JSONObject || obj instanceof JSONArray)
         {
             super.print(obj);
         }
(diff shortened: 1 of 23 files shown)


api/src/org/labkey/api/jsp/LabKeyJspWriter.java
```diff
@@ -44,7 +46,7 @@ public void print(String s) throws IOException
     public void print(Object obj) throws IOException
     {
         // These are the only objects we consider safe-to-render
-        if (null == obj || obj instanceof SafeToRender || obj instanceof Number || obj instanceof Boolean)
+        if (null == obj || obj instanceof SafeToRender || obj instanceof Number || obj instanceof Boolean | obj instanceof JSONObject || obj instanceof JSONArray)
```
labkey-jeckels:
```suggestion
        if (null == obj || obj instanceof SafeToRender || obj instanceof Number || obj instanceof Boolean || obj instanceof JSONObject || obj instanceof JSONArray)
```

labkey-jeckels (approved):

api/src/org/labkey/api/jsp/LabKeyJspWriter.java
labkey-adam: Thanks... fixed

PR #3712, opened 2022-09-24
Improve Functionality of `TestMothershipReportAction`

labkey-tchad:
#### Rationale
`TestMothershipReportAction` uses the server's real `ServerSessionGUID`, which can cause reports to be associated with the wrong server. Various other test helpers need to be updated to target a particular host.

#### Related Pull Requests
* #3707 

#### Changes
* Use random session GUID when generating test `MothershipReport` for `TestMothershipReportAction`
* Enable server debug logging during mothership tests
* Update `MothershipReportTest.testJsonMetrics` to check server and session info
* `MothershipHelper`
  * Add methods to query server installation info
  * Add methods for mocking and submitting mothership reports (verified but unused)
* `ShowInstallationDetailPage`
  * Don't assume that 'localhost' has `serverInstallationId = 1`
  * Update methods for fetching server IP, distribution, and host name


mothership/test/src/org/labkey/test/pages/mothership/ShowInstallationDetailPage.java
```diff
@@ -32,20 +47,62 @@ public ShowInstallationDetailPage(WebDriver driver)
...
+
+    public static ShowInstallationDetailPage beginAt(WebDriverWrapper driver, String hostName) throws IOException, CommandException
+    {
+        Integer serverInstallationId = new MothershipHelper(driver).getServerInstallationId(hostName);
+        return beginAt(driver, serverInstallationId);
+    }
+
+    private static Integer getServerInstallationId(WebDriverWrapper driver, String hostName) throws IOException, CommandException
+    {
+        SelectRowsCommand selectRows = new SelectRowsCommand(MOTHERSHIP_CONTROLLER, SERVER_INSTALLATION_QUERY);
```
labkey-jeckels: While they happen to share a name, the schema and controller aren't necessarily the same thing. Separate constants with the same value seems slightly preferable

mothership/test/src/org/labkey/test/pages/mothership/ShowInstallationDetailPage.java
```diff
@@ -32,20 +47,62 @@ public ShowInstallationDetailPage(WebDriver driver)
...
+        return getSessionProperty("ServerIP");
+    }
+
     public String getDistributionName()
     {
-        return getInstallationValue("Distribution");
+        return getSessionProperty("Distribution");
+    }
+
+    private String getSessionProperty(String ServerIP)
```
labkey-jeckels:
```suggestion
    private String getSessionProperty(String serverIP)
```

mothership/test/src/org/labkey/test/pages/mothership/ShowInstallationDetailPage.java
```diff
@@ -32,20 +47,62 @@ public ShowInstallationDetailPage(WebDriver driver)
...
+    {
+        if (getSessionsGrid().getDataRowCount() == 0)
+        {
+            throw new IllegalStateException("No session info for server: " + getServerHostName());
+        }
+        return getSessionsGrid().getDataAsText(0, ServerIP);
```
labkey-jeckels:
```suggestion
        return getSessionsGrid().getDataAsText(0, serverIP);
```

mothership/test/src/org/labkey/test/util/mothership/MothershipHelper.java
```diff
@@ -26,46 +38,117 @@
...
+    }
+
+    public Map<String, Object> getLatestServerInfo() throws IOException, CommandException
+    {
+        return getServerInstallationInfo(null);
+    }
+
+    private Map<String, Object> getServerInstallationInfo(String hostName) throws IOException, CommandException
+    {
+        SelectRowsCommand selectRows = new SelectRowsCommand(MOTHERSHIP_CONTROLLER, SERVER_INSTALLATION_QUERY);
```
labkey-jeckels: Schema name instead of controller name

labkey-jeckels (approved):

PR #3721, opened 2022-09-27
Make Parameter dumb again

labkey-matthewb:
#### Rationale
One Parameter constructor was substituting getJdbcRsName() instead of using the passed in column name. This breaks code since the caller expects to use the column name.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45368

The getJdbcRsName() code was a "fix" for https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=26694.  Hopefully we have regression tests for that.

#### Related Pull Requests
* 

#### Changes
* removed getJdbcRsName() code
* removed Parameter members that it didn't use itself (scale, parameter) and changed caller to keep track of these.

### DIFF
--- a/api/src/org/labkey/api/data/Parameter.java
+++ b/api/src/org/labkey/api/data/Parameter.java
@@ -17,16 +17,15 @@
 package org.labkey.api.data;
 
 import com.google.common.primitives.Ints;
-import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
-import org.json.old.JSONObject;
 import org.labkey.api.attachments.AttachmentFile;
 import org.labkey.api.exp.Lsid;
 import org.labkey.api.query.QueryService;
 import org.labkey.api.util.ResultSetUtil;
 import org.labkey.api.util.UnexpectedException;
+import org.labkey.api.util.logging.LogHelper;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/data/Parameter.java
```diff
@@ -261,13 +232,11 @@ public void setValue(@Nullable Object in)
                 return;
             }
 
-            if (value instanceof AttachmentFile)
+            if (value instanceof final AttachmentFile attachmentFile)
```
labkey-jeckels, line 235: I didn't know this was a supported syntax!

labkey-jeckels (approved):

labkey-bpatel:
@labkey-matthewb can we please merge to release22.7-SNAPSHOT

api/src/org/labkey/api/data/Parameter.java
labkey-bpatel: yeah ditto, had to do a double take on this line
pretty cool

labkey-matthewb:
> @labkey-matthewb can we please merge to release22.7-SNAPSHOT

I'd rather not?  Seems like a bit much for a stable release.  But I'll defer to triage/Josh.

labkey-bpatel:
> > @labkey-matthewb can we please merge to release22.7-SNAPSHOT
> 
> I'd rather not? Seems like a bit much for a stable release. But I'll defer to triage/Josh.

@labkey-matthewb i didn't see Hannah's comment on the [ticket](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45368), it is targeted for 22.11, so nevermind my comment

PR #3722, opened 2022-09-27
Sample Finder, Assay results as filter

XingY:
#### Rationale
In order to support finding samples based on their assay data properties, a new compare type COLUMN_IN is added to allow constructing "sampleId IN (select sampleId from assayresult where...)" using filter type. Another new compare type COLUMN_NOT_IN is also added to support finding samples that don't have assay result "sampleId NOT IN (select sampleId from assayresult)"

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/969
* https://github.com/LabKey/platform/pull/3722
* https://github.com/LabKey/sampleManagement/pull/1248
* https://github.com/LabKey/biologics/pull/1628

#### Changes
*


query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -264,6 +271,125 @@ public WhereClause createFilterClause(@NotNull FieldKey fieldKey, Object value)
...
     };
 
+    private static SQLFragment getColumnInSql(@NotNull FieldKey fieldKey, Set<String> params, User user, Container container, Map<FieldKey, ? extends ColumnInfo> columnMap, boolean negate)
+    {
+        ColumnInfo col = columnMap.get(fieldKey);
+        SQLFragment colFrag = new SQLFragment(col == null ? fieldKey.toString() : col.getAlias());
+        String[] values = params.toArray(new String[0]);
+
+        if (values.length < 3)
+            throw new NotFoundException("Incorrect number of arguments provided");
```
Sigmonia: nit: `NotFoundException` seems to apply more to a resource reference (e.g., 404 error) vs a missing/malformed parameter or argument

```suggestion
            throw new IllegalArgumentException("Incorrect number of arguments provided");
```

query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -264,6 +271,125 @@ public WhereClause createFilterClause(@NotNull FieldKey fieldKey, Object value)
...
+            throw new NotFoundException("QueryName not specified");
+
+        if (user == null || container == null)
+            throw new NotFoundException("Invalid context");
+
+        QuerySchema querySchema = DefaultSchema.get(user, container, schemaName);
+        if (!(querySchema instanceof UserSchema userSchema))
+            throw new NotFoundException("Could not find the specified schema in the folder '" + container.getPath() + "'");
+
+        SqlDialect dialect = querySchema.getDbSchema().getSqlDialect();
```
labkey-matthewb: This looks good to me.  My only comment would be to consider not breaking up the "value" into parts.  Just using SQL to do this makes the feature more general.  E.g. instead of ~columnin=columnA,schemaB,tableC,D=E  you could simply do ~columnin=SELECT columnA From schmeaB.tableC WHERE D=E

labkey-matthewb (changes requested):

query/src/org/labkey/query/QueryServiceImpl.java
XingY: I needed the schemaName to get the QuerySchema/UserSchema and the selectColumn to select from the the fromSql. 

Are you suggesting to parse "SELECT xx FROM schema.table .." and extract those information from the parsed string?

query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -264,6 +271,125 @@ public WhereClause createFilterClause(@NotNull FieldKey fieldKey, Object value)
...
+        if (!StringUtils.isEmpty(whereSql))
+        {
+            selectIdsSql = selectIdsSql.append(" WHERE (")
+                    .append(whereSql)
+                    .append(")");
+        }
+
+        QueryService qs = QueryService.get();
+        QueryDefinition qd;
+        qd = qs.createQueryDef(user, container, userSchema, GUID.makeGUID().replace("-", ""));
```
Sigmonia: Need the separate initialization?
```suggestion
        QueryDefinition qd = qs.createQueryDef(user, container, userSchema, GUID.makeGUID().replace("-", ""));
```

query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -264,6 +271,125 @@ public WhereClause createFilterClause(@NotNull FieldKey fieldKey, Object value)
...
+        //TODO qd.setContainerFilter();
```
Sigmonia, line 282: TODO still needed? or addressed after this story?

query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -264,6 +271,125 @@ public WhereClause createFilterClause(@NotNull FieldKey fieldKey, Object value)
...
+        qd.setSql(selectIdsSql.getSQL());
+        ArrayList<QueryException> qerrors = new ArrayList<>();
+        TableInfo t = qd.getTable(userSchema, qerrors, true, true);
```
Sigmonia, line 285: do we need to check the `qerrors`? or switch to inline?

query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -264,6 +271,125 @@ public WhereClause createFilterClause(@NotNull FieldKey fieldKey, Object value)
...
+        if (t == null)
+            throw new NotFoundException("Unable to find the specified table");
+
+        SQLFragment fromSql = t.getFromSQL("_");
+
+        SQLFragment sql = new SQLFragment();
+        sql.append("(").append(colFrag)
+                .append(")")
```
Sigmonia: nit: extend the chain?
```suggestion
        SQLFragment sql = new SQLFragment()
                .append("(")
                .append(colFrag)
                .append(")")
```

query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -264,6 +271,125 @@ public WhereClause createFilterClause(@NotNull FieldKey fieldKey, Object value)
...
+    {
+        @Override
+        public SimpleFilter.FilterClause createFilterClause(@NotNull FieldKey fieldKey, Object value, User user, Container container)
+        {
+            return new SimpleFilter.SQLClause(new SQLFragment(), fieldKey)
+            {
+                @Override
+                public SQLFragment toSQLFragment(Map<FieldKey, ? extends ColumnInfo> columnMap, SqlDialect dialect)
+                {
+                    Set<String> params = parseParams(value, getValueSeparator());
```
Sigmonia: Style/nit: call parameter validation here? though I guess having it right before usage in the sql generation method also makes sense--a little torn.

query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -264,6 +271,125 @@ public WhereClause createFilterClause(@NotNull FieldKey fieldKey, Object value)
...
```
Sigmonia: same as above

query/src/org/labkey/query/QueryServiceImpl.java
labkey-matthewb: You can use the UserSchema for the primary table (e.g.  tinfo.getUserSchema() in buildFilter()).  The schema and table in the SELECT will be resolved relative to that.  

For the column, I guess you'd use t.getColumns().get(0).getValueSql("_").

labkey-matthewb: It is a bit annoying that the schema is not passed around explicitly.  But note that even in addUrlFilters it is usually available indirectly via selectColumns.get(0).getParentTable().getUserSchema().

Sigmonia (approved):

query/src/org/labkey/query/QueryServiceImpl.java
XingY: Yes, this will be addressed as part of cross folder assay story

query/src/org/labkey/query/QueryServiceImpl.java
XingY: Thanks for the suggestions. This works great.

query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -264,6 +266,85 @@ public WhereClause createFilterClause(@NotNull FieldKey fieldKey, Object value)
         }
     };
 
+    private static SQLFragment getColumnInSql(@NotNull FieldKey fieldKey, Object value, User user, Container container, Map<FieldKey, ? extends ColumnInfo> columnMap, @NotNull List<ColumnInfo> selectColumns, boolean negate)
+    {
+        if (user == null || container == null)
+            throw new NotFoundException("Invalid context");
+
+        ColumnInfo col = columnMap.get(fieldKey);
```
labkey-matthewb, line 274: I can't remember the difference in usage between columnMap and selectColumns and if/when they can be null/empty.  I wonder if we can assert that col is not null and then use "col" instead of "selectColumns.get(0)"?

labkey-matthewb (approved):

PR #3728, opened 2022-09-29
Issue 46412: When importing assay data, collect column names from results domain

labkey-susanh:
#### Rationale
[Issue 46412](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46412) Because we were inferring the set of columns using the first row of data provided from an import from the grid, this caused us to lose some data when the first row entered in a grid did not contain all of the fields for an assay run. This PR changes the initialization of the TSVWriter by passing in all results domain column names so we can pick up all fields from each record.

#### Changes
* Update TSVWriter initialization during assay result import


cnathe (approved):
Yes, that seems much better to use the full domain prop set instead of just those included in the data row.

PR #3731, opened 2022-09-30
Issue 46371: LKS grid column header menus don't work when scrolled down so they 'lock'

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46371

DataRegion.js code for header locking is making a clone of the header row. This results in two DOM elements for each header cell menu item with the same ID. In the attachEventHandler Util.js function, we are using document.getElementById() to find the DOM element for the event handling. This is then only finding the first element and not attaching to the cloned / floating header row elements.

Ideally we'd switch away from this "header row cloning" behavior and use the CSS `position: sticky`. I mocked this up locally and it worked well. We'd have to figure out a few things like what to do with the floating pagination component, styling for the floating header bottom border, etc. 

#### Changes
* update attachEventHandler to switch to document.querySelectorAll() instead of document.getElementById()

### DIFF
--- a/api/webapp/clientapi/dom/Utils.js
+++ b/api/webapp/clientapi/dom/Utils.js
@@ -866,10 +866,11 @@ LABKEY.Utils = new function(impl, $) {
             return;
         const fn = function()
         {
-            if (typeof el === "string")
-                el = document.getElementById(el);
-            if (el)
-                el['on' + eventName] = handler;
+            if (typeof el === "string") {
+                const list = document.querySelectorAll('#' + el);
+                for (let i in list)
+                    list[i]['on' + eventName] = handler;
+            }
         };
         (immediate || document.readyState!=="loading") ? fn() : document.addEventListener('load', fn);
     };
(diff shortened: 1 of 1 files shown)


api/webapp/clientapi/dom/Utils.js
```diff
@@ -859,17 +859,37 @@ LABKEY.Utils = new function(impl, $) {
         (immediate || document.readyState!=="loading") ? fn() : document.addEventListener('load', fn);
     };
 
+    impl.isValidQuerySelector = function(selector) {
```
labkey-nicka: nit: I don't think we should expose another method on the `LABKEY.Utils` interface if we don't have to. You can move this above the returned object and just use it inside of `attachEventHandler`.

labkey-nicka (approved):

api/webapp/clientapi/dom/Utils.js
cnathe: Done. Thanks.

PR #3734, opened 2022-09-30
Issue 46282: View vs download options for file and attachment fields

labkey-jeckels:
#### Rationale
We want consistent handling for viewing and downloading files across data types. We are standardizing on showing the file in the browser when supported (like PDFs and images) and opening in a new tab/window. Admins can also opt to always download on a per-field basis.

#### Changes
* Support an "attachment" format option for file and attachment fields, forcing the link to work as a download
* Make it easier for individual download actions to indicate they want to download vs view
* Make lists, datasets, sample types, data classes, and assay file/attachments behave consistently
* Remove unused code in MimeMap and use a singleton instance in more places

### DIFF
--- a/api/schemas/tableInfo.xsd
+++ b/api/schemas/tableInfo.xsd
@@ -726,6 +726,8 @@
                         is the string to display when false, and "null" is the string to display when null. For date and
                         datetime columns the values Date and DateTime can be used to display as
                         either date or date and time with the currently configured display formats for this folder.
+                        For file and attachment columns, use "inline" or "attachment" to request that the download
+                        action render within the browser or save as a file, respectively.
                         Supported for SQL metadata, dataset import/export and list import/export.
                     </xs:documentation>
 				</xs:annotation>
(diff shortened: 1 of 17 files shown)


api/schemas/tableInfo.xsd
```diff
@@ -726,6 +726,8 @@
                         is the string to display when false, and "null" is the string to display when null. For date and
                         datetime columns the values Date and DateTime can be used to display as
                         either date or date and time with the currently configured display formats for this folder.
+                        For file and attachment columns, use "inline" or "attachment" to request that the download
+                        action render within the browser or save as a file, respectively.
```
labkey-martyp: Is there a consistent default behavior we could note here?

api/src/org/labkey/api/util/MimeMap.java
```diff
@@ -194,30 +178,15 @@ public String getContentType(String extn)
...
-        int length = fileName.length();
-
-        int newEnd = fileName.lastIndexOf('#');
-        if (newEnd == -1) newEnd = length;
-        // Instead of creating a new string.
-        //         if (i != -1) {
-        //             fileName = fileName.substring(0, i);
-        //         }
-        int i = fileName.lastIndexOf('.', newEnd);
+        int i = fileName.lastIndexOf('.');
```
labkey-martyp: Can use FileUtil.getExtension instead of doing this right?

If there's some concern about fragment Ids or other complexities then maybe an external library like org.apache.commons.io.FilenameUtils.getExtension.

labkey-martyp (approved):
Looks good, just a couple comments

PR #3735, opened 2022-10-03
Proposed updates to the Announcements API

vagisha:
#### Rationale
We would like to convert the Panorama Public requests message board to a secure message board so that we can communicate with the submitters over the message board while keeping each message private to the associated submitter.
Messages are posted automatically via the Panorama Public module when a new request is submitted / updated / deleted, and when the data is copied to Panorama Public.  If the submitter has any questions, or a Panorama Public admin needs to contact  the submitter, they should post a response to the message thread instead of sending emails. 

#### Changes
- Added a method to insert an announcement with status and member list
- Added a method to return the latest post for a message thread
- Use the title of the latest post as the email subject


labkey-jeckels:
LK repo copy of the FB created to get TeamCity to build and test

labkey-jeckels (approved):

PR #3740, opened 2022-10-04
Issue 45372: QClause should not process columns by isHidden bit

labkey-nicka:
#### Rationale
This addresses [Issue 45372](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45372) by removing the check against `column.isHidden()` when collecting the set of columns to query against for a `QClause` query.

#### Changes
* Undo check against `column.isHidden()`.

### DIFF
--- a/api/src/org/labkey/api/data/CompareType.java
+++ b/api/src/org/labkey/api/data/CompareType.java
@@ -820,9 +820,13 @@ public abstract class CompareType
                         _queryColumns.add(column);
                     }
 
-                    // skip uninteresting things like 'LSID' and 'SourceProtocolApplication'
-                    if (column.isHidden())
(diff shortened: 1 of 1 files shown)


labkey-susanh (approved):
Seems like we might now get some false positives if we're now including system-defined fields.  Perhaps the risk of that is low enough to be tolerated, though?

labkey-susanh:
Would we want to exclude any columns based on a ConceptURI possibly?

labkey-nicka:
> Seems like we might now get some false positives if we're now including system-defined fields. Perhaps the risk of that is low enough to be tolerated, though?

There is further filtering of columns that happens in this function:

image

I think we see if this can be tolerated.

PR #3750, opened 2022-10-06
Convert IpynbReport to use nbconvert service

labkey-matthewb:
#### Rationale
Converting to a micro-service architecture simplifies cloud deployment

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/reports/report/python/IpynbReport.java
+++ b/api/src/org/labkey/api/reports/report/python/IpynbReport.java
@@ -362,6 +362,10 @@ public class IpynbReport extends DockerScriptReport
             FileUtils.write(new File(working, CONFIG_FILE), reportConfig.toString(), StringUtilsLabKey.DEFAULT_CHARSET);
 
             URL service = getServiceAddress(context.getContainer());
+            // For testing, just return if the remoteURL host is "noop.test"
+            if ("noop.test".equals(service.getHost()))
+                return 0;
+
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/reports/report/python/IpynbReport.java
```diff
@@ -166,15 +193,20 @@ public HttpView<?> renderReport(ViewContext context) throws Exception
...
         LOG.trace("BEFORE: " + workingDirectory.getPath() + "\n\t" +
                 StringUtils.join(beforeExecute.stream().map(f ->
                         f.getPath().replace(workingDirectory.toString(), "") + " : " + f.length()).toArray(), "\n\t"));
 
-        ExecuteStrategy ex = new DockerRunTarStdinStdout();
+        ExecuteStrategy ex;
+        if (null != getServiceAddress(context.getContainer()))
+            ex = new WebServiceExecuteStrategy();
+        else
+            ex = new DockerRunTarStdinStdout();
```
labkey-klum: fine to leave for now, but if we only plan to support a webservice based endpoint, we could consider removing the docker based code.

labkey-klum (approved):
looks like you need to merge develop into your branch to fix the test failures, also fine to just merge since that is the only failure : `Constants$TestCase`

labkey-tchad (approved):

PR #3751, opened 2022-10-06
Issue 46473: Don't retain a stale reference to Container object

labkey-jeckels:
#### Rationale
When cached objects hold a reference to Containers, they may end up referencing a stale copy

#### Changes
* Switch from storing the Container itself to a Supplier<Container>

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -1727,7 +1727,8 @@ public class Container implements Serializable, Comparable<Container>, Securable
     }
 
     /** Convert a container into a reference that can be used to get the latest version of the container object. See issue 46473 */
-    public static Supplier<Container> supplierOf(Container c)
+    @NotNull
+    public static Supplier<Container> supplierOf(@Nullable Container c)
     {
         final GUID id = c == null ? null : c.getEntityId();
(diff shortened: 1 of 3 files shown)


labkey-jeckels:
@labkey-matthewb what do you think of this approach? I tried to get something we can easily swap into the dozens or hundreds of places that have the same underlying problem.

api/src/org/labkey/api/data/Container.java
```diff
@@ -1724,4 +1725,11 @@ public void setExpirationDate(LocalDate expirationDate)
     {
         _expirationDate = expirationDate;
     }
+
+    /** Convert a container into a reference that can be used to get the latest version of the container object. See issue 46473 */
+    public static Supplier<Container> supplierOf(Container c)
+    {
+        final GUID id = c == null ? null : c.getEntityId();
+        return () -> id == null ? null : ContainerManager.getForId(id);
```
labkey-matthewb, line 1734: I sometimes wonder about having thread/request local caches for stuff like this...

labkey-matthewb (approved):
I think this is a very useful pattern to consider adopting for anything cacheable.  It will result in a lot more contention on the container cache, of course.

PR #3282, opened 2022-04-22
45265: Lists: Default to not including project/shared lists

labkey-nicka:
#### Rationale
This PR addresses [45625](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45265) by switching the default behavior for fetching list definitions to only include definitions from the provided container. List query behavior continues to resolve cross-folder and the `ListQuerySchema` has been updated to explicitly resolve list definitions cross-folder.

Additionally, I've included a couple fixes URL composition for list actions. Namely, when an action resolves a list by `listId` then the scope of list definitions retrieved is for the provided container. If an action resolves a list by `name` then the scope is the provided container plus the container's project and the /Shared folder. This aligns with the underlying data structures since listIds are only unique to a container where as resolving a list by name resolves cross-folder.

#### Related Pull Requests
* #3001 
* #3264

#### Changes
* `ListService` and `ListManager` updated to default to `false` for any implicitly usages of `includeProjectAndShared`.
* `ListQuerySchema` updated to explicitly resolve list definitions cross-folder. Lone exception is `getTableNames` where it defaults to only the current container scope. The latter addresses [45198](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45198).
* Add `urlImport` interface to `ListDefinition` for resolving the import bulk data URL for a list in LKS.
* Resolve import and details URLs for list definitions by "name" so they resolve cross-folder.
*  (Miscellaneous) Fix logic in permissions check for picklists.

PR #3752, opened 2022-10-06
Lists: cross-folder support improvements

labkey-nicka:
#### Rationale
With #3001 (and all subsequent related PRs) we introduced support for cross-folder data in Lists. This was done out of necessity to support cross-folder data in LKB where Lists are utilized as the backing structure in a number of areas. As such, it works for query purposes (CRUD operations), but did not cover updates the LKS user experience to support cross-folder list data. This PR makes several notable updates to improve the experience.

#### Related Pull Requests
* #3001

#### Changes
* Prevent list name duplication within the scope of a list. Mimics behavior already in place for Sample Types (which also can have data which spans multiple folders).
* Default the `list-begin.view` to using the `CurrentPlusProjectAndShared` container filter.
* Relabel the `Container` column as `Folder` for the list management table.
* Add the `Folder` field to the default set of fields for the list management table.
* Resolve to the current folder for the details URL for the `Name` column on the list management table. Formerly, this resolved to the folder where the list was defined.
* Update the set of available container filters for lists to reflect those that best fit how data can be stored in lists.
* Support the product project container filter paradigm for lookups in query insert/update views. This means lookups (a.k.a. dropdowns) in those views will now include data from other folders. This behavior is consistent with what we do in our applications. Additionally, this is only enabled when the user is working from within a "Product Projects" enabled folder.

### DIFF
--- a/experiment/resources/schemas/exp.xml
+++ b/experiment/resources/schemas/exp.xml
@@ -216,7 +216,7 @@
           </column>
           <column columnName="Container" />
           <column columnName="Name" >
-              <url>/list/grid.view?listId=${ListId}</url>
+              <url>/list/grid.view?name=${Name}</url>
           </column>
           <column columnName="DomainId" />
           <column columnName="KeyName" />
--- a/experiment/resources/schemas/exp.xml
+++ b/experiment/resources/schemas/exp.xml
@@ -216,6 +216,10 @@
           </column>
           <column columnName="Container" />
           <column columnName="Name" >
+              <!--
+                Resolve lists by "name" instead of "listId" as "name" will resolve cross-folder
+                where as "listId"s are only unique within a folder.
+              -->
               <url>/list/grid.view?name=${Name}</url>
           </column>
           <column columnName="DomainId" />
--- a/list/src/org/labkey/list/model/ListDomainKind.java
+++ b/list/src/org/labkey/list/model/ListDomainKind.java
@@ -265,7 +265,7 @@ public abstract class ListDomainKind extends AbstractDomainKind<ListDomainKindPr
         String type = getType(keyType, category);
         StringBuilder typeURI = getBaseURI(name, type, c);
 
-        // 13131: uniqueify the lsid for situations where a preexisting list was renamed
+        // Issue 13131: uniqueify the lsid for situations where a preexisting list was renamed
         int i = 1;
         String sTypeURI = typeURI.toString();
         String uniqueURI = sTypeURI;
@@ -362,13 +362,13 @@ public abstract class ListDomainKind extends AbstractDomainKind<ListDomainKindPr
         String keyName = listProperties.getKeyName();
 
         if (StringUtils.isEmpty(name))
-            throw new ApiUsageException("List name must not be null");
+            throw new ApiUsageException("List name is required");
         if (name.length() > MAX_NAME_LENGTH)
             throw new ApiUsageException("List name cannot be longer than " + MAX_NAME_LENGTH + " characters");
-        if (ListService.get().getList(container, name, false) != null)
+        if (ListService.get().getList(container, name, true) != null)
             throw new ApiUsageException("The name '" + name + "' is already in use.");
         if (StringUtils.isEmpty(keyName))
-            throw new ApiUsageException("List keyName must not be null");
+            throw new ApiUsageException("List keyName is required");
 
         KeyType keyType = getDefaultKeyType();
 
@@ -382,7 +382,8 @@ public abstract class ListDomainKind extends AbstractDomainKind<ListDomainKindPr
         }
 
         ListDefinition.Category category;
-        try {
+        try
+        {
             category = listProperties.getCategory() != null ? ListDefinition.Category.valueOf(listProperties.getCategory()) : null;
         }
         catch (IllegalArgumentException e)
@@ -700,7 +701,7 @@ public abstract class ListDomainKind extends AbstractDomainKind<ListDomainKindPr
     @Override
     public boolean matchesTemplateXML(String templateName, DomainTemplateType template, List<GWTPropertyDescriptor> properties)
     {
-        if(!(template instanceof ListTemplateType))
+        if (!(template instanceof ListTemplateType))
             return false;
 
         ListOptionsType options = ((ListTemplateType) template).getOptions();
@@ -723,5 +724,4 @@ public abstract class ListDomainKind extends AbstractDomainKind<ListDomainKindPr
     {
         var props = getBaseProperties(d);
     }
-
 }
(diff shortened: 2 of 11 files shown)


experiment/resources/schemas/exp.xml
```diff
@@ -216,7 +216,7 @@
           </column>
           <column columnName="Container" />
           <column columnName="Name" >
-              <url>/list/grid.view?listId=${ListId}</url>
+              <url>/list/grid.view?name=${Name}</url>
```
labkey-susanh, line 223: Why this change? Just for readability? (With our recent troubles regarding using names of things as references instead of ids and subsequently wanting to rename, I'm on the lookout for the usages of names instead of ids.)

list/src/org/labkey/list/model/ListDomainKind.java
```diff
@@ -362,13 +362,13 @@ public Domain createDomain(GWTDomain domain, ListDomainKindProperties listProper
         String keyName = listProperties.getKeyName();
 
         if (StringUtils.isEmpty(name))
-            throw new ApiUsageException("List name must not be null");
+            throw new ApiUsageException("List name is required");
         if (name.length() > MAX_NAME_LENGTH)
             throw new ApiUsageException("List name cannot be longer than " + MAX_NAME_LENGTH + " characters");
-        if (ListService.get().getList(container, name, false) != null)
+        if (ListService.get().getList(container, name, true) != null)
```
labkey-susanh, line 368: I believe changing from `false` to `true` here means there could be duplicate names produced (for a regular list vs. a picklist). Is that intentional?

list/src/org/labkey/list/model/ListManager.java
```diff
@@ -207,10 +207,9 @@ private Collection<ListDef> getAllScopedLists(@NotNull Container container, bool
      */
     String getListTableName(TableInfo ti)
     {
-        if (ti instanceof ListTable)
-            return ((ListTable)ti).getRealTable().getSelectName();
-        else
-            return ti.getSelectName();  // if db is being upgraded from <= 13.1, lists are still SchemaTableInfo instances
+        if (ti instanceof ListTable lti)
+            return lti.getRealTable().getSelectName();
+        return ti.getSelectName();  // if db is being upgraded from <= 13.1, lists are still SchemaTableInfo instances
```
labkey-susanh, line 212: This comment makes me thing we can get rid of this if statement...

list/src/org/labkey/list/model/ListManagerTable.java
```diff
@@ -49,18 +50,28 @@ public ListManagerTable(ListManagerSchema userSchema, TableInfo table, Container
         super(table, userSchema, cf);
 
         addWrapColumn(_rootTable.getColumn(FieldKey.fromParts("ListID")));
-        addWrapColumn(_rootTable.getColumn(FieldKey.fromParts("Name")));
+        {
+            var column = addWrapColumn(_rootTable.getColumn(FieldKey.fromParts("Name")));
+
+            // Lists can contain data that spans multiple folders.
+            // Override the container context for the details URL to always be the current folder.
```
labkey-susanh: Does this mean that links from the Name column for items added in a different folder will present as Not Found, or is this preventing that from happening?

experiment/resources/schemas/exp.xml
labkey-nicka: When referencing lists they resolve cross-folder by name. If an ID is specified then it only resolves within the requesting folder. This is specifically called out in the specification.

> These links will need to be updated to use the "name" URL parameter rather than the "listId" parameter as "name" is resolvable cross-folder where as "listId" is only ever resolved to the current folder.

list/src/org/labkey/list/model/ListDomainKind.java
labkey-nicka: This allows us to enforce _not_ having duplicate names across folders within scope.

list/src/org/labkey/list/model/ListManager.java
labkey-nicka: Agreed, although not planning on changing the logic with this work.

list/src/org/labkey/list/model/ListManagerTable.java
labkey-nicka: Hmm, maybe the comment needs to be updated. Here is what this is addressing from the specification:

> Update links in on the listing of lists page (i.e. list-begin.view) to go to the list and show data in the current container. Currently, it navigates to the folder where the list is defined.

If a list is visible from the list management page then clicking through to it within that folder should always resolve.

experiment/resources/schemas/exp.xml
labkey-nicka: We could consider revising the URL resolution to allow `listId` to resolve cross-folder, however, there are other side-effects of doing so that we'll need to consider (I can't recall them at the moment).

labkey-susanh: Got it. Perhaps worth leave a comment here to this effect?

list/src/org/labkey/list/model/ListManagerTable.java
labkey-susanh: Ah, I also missed that this was the name of the list, not the name of an item in the list, but a comment update would be helpful.

experiment/resources/schemas/exp.xml
labkey-nicka: After review I now remember why `listId` resolution cross-folder is problematic. The `listId` of a list is only unique within a folder. This means that if you allow the `listId` to resolve cross-folder you end up fetching whatever list has that `listId` in a folder scope above. Obviously, this is not what we want. See https://github.com/LabKey/platform/pull/3282.

labkey-susanh: So we'll have name uniqueness but not id uniqueness. Well, that I definitely did not expect.  So, yeah, comments in the code would be good.

labkey-nicka: We won't truly have name uniqueness either since we'll continue to be backwards compatible with instances that may have had overlapping names within a folder scope.

labkey-nicka: Comment added.

list/src/org/labkey/list/model/ListManagerTable.java
labkey-nicka: Revised the comment.

list/src/org/labkey/list/model/ListDomainKind.java
labkey-susanh: My bad. I was looking at `getLists(Container container, @Nullable User user, boolean checkVisibility)`, not `getList(Container container, String name, boolean includeProjectAndShared)`. Makes a difference. 🙂

labkey-susanh (approved):

labkey-klum (approved):

PR #3754, opened 2022-10-10
EHR Sample Manager: All sample types metadata overlay

labkey-martyp:
#### Rationale
Similar to studyData.query.xml, add the ability to apply metadata to all sample types in a container. For EHR Sample Manager module, this is applying a java customizer to all sample types.

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/422

#### Changes
* Add AllSampleTypes metadata overlay for sample types

### DIFF
--- a/api/src/org/labkey/api/exp/query/SamplesSchema.java
+++ b/api/src/org/labkey/api/exp/query/SamplesSchema.java
@@ -56,6 +56,7 @@ import java.util.Set;
 public class SamplesSchema extends AbstractExpSchema
 {
     public static final String SCHEMA_NAME = "samples";
+    // AllSampleTypes.query.xml metadata file is applied to all Sample Types in a container
     public static final String SCHEMA_METADATA_NAME = "AllSampleTypes";
     public static final SchemaKey SCHEMA_SAMPLES = SchemaKey.fromParts(SamplesSchema.SCHEMA_NAME);
     public static final String SCHEMA_DESCR = "Contains data about the samples used in experiment runs.";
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/exp/query/SamplesSchema.java
```diff
@@ -56,6 +56,7 @@
 public class SamplesSchema extends AbstractExpSchema
 {
     public static final String SCHEMA_NAME = "samples";
+    public static final String SCHEMA_METADATA_NAME = "AllSampleTypes";
```
labkey-jeckels, line 60: Please add a comment describing its purpose. And let's be sure to update the docs with this behavior too.

labkey-jeckels (approved):

labkey-bpatel (approved):

api/src/org/labkey/api/exp/query/SamplesSchema.java
labkey-martyp: Added comment. @labkey-mohara do we have somewhere we could document this behavior. The behavior is described in the PR description above.  This page has a blurb about StudyData.query.xml which is the same behavior but for datasets. 

https://www.labkey.org/Documentation/wiki-page.view?name=columnMetadata

labkey-mohara: Docs updated: https://www.labkey.org/Documentation/wiki-page.view?name=columnMetadata#overrides

PR #3755, opened 2022-10-10
Issue 44868: Study publish to include custom default views for any datasets selected

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44868

#### Changes
* CreateChildStudyPipelineJob update to add dataset default view entityIds to the list of views to export/import

### DIFF
--- a/study/src/org/labkey/study/model/ChildStudyDefinition.java
+++ b/study/src/org/labkey/study/model/ChildStudyDefinition.java
@@ -246,6 +246,16 @@ public class ChildStudyDefinition
         _views = views;
     }
 
+    public void addView(String id)
+    {
+        List<String> viewsList = new ArrayList<>();
+        if (_views != null)
+            viewsList.addAll(Arrays.stream(_views).toList());
+        viewsList.add(id);
+        _views = new String[viewsList.size()];
+        _views = viewsList.toArray(_views);
+    }
+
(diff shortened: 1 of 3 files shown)


study/src/org/labkey/study/model/ChildStudyDefinition.java
```diff
@@ -246,6 +246,16 @@ public void setViews(String[] views){
...
     }
 
+    public void addView(String id)
+    {
+        List<String> viewsList = new ArrayList<>();
+        if (_views != null)
+            viewsList.addAll(Arrays.stream(_views).toList());
+        viewsList.add(id);
+        _views = new String[viewsList.size()];
+        _views = viewsList.toArray(_views);
```
labkey-susanh, line 256: So unfortunate that this adding to an array is this involved. One wonders if _views would be better off as a List, but that's probably a change for another time.

labkey-susanh (approved):

study/src/org/labkey/study/model/ChildStudyDefinition.java
cnathe: I agree. And I had the same thought about changing _views to be a List<String> instead of String[], but there are several other member variables that are String[] types so I thought it would be better to leave it as is for now.

PR #3765, opened 2022-10-13
46233: Study custom tab location not saved correctly

labkey-klum:
#### Rationale
Moving tabs on some portal pages (like study) don't always work. See the tracking issue for details, but the root cause is that the list of tabs returned from `Portal.getTabPages` can return more tabs than are actually show. At render time a tab page is asked whether or not it should be visible and a recent study change registered the specimen tab but would hide it if the specimen module was not present. 

The change is to not include these tab pages when handling any reordering of the tabs.

[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46233)


labkey-klum:
assigning to @labkey-alan for review since it looks like he worked on this code ten years ago 😺

labkey-alan (approved):
Looks good to me

PR #3766, opened 2022-10-15
Startup properties for SSO configuration logos

labkey-adam:
#### Rationale
Cloud team has requested a way to set SSO configuration logos via startup properties. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=45972

### DIFF
--- a/api/src/org/labkey/api/security/AuthenticationConfiguration.java
+++ b/api/src/org/labkey/api/security/AuthenticationConfiguration.java
@@ -1,32 +1,42 @@
 package org.labkey.api.security;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.attachments.Attachment;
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/security/AuthenticationConfiguration.java
```diff
@@ -93,31 +110,111 @@
...
+                        }
+
+                        try (InputStream is = supplier.get())
+                        {
+                            saveLogo(user, type, is);
+                        }
+                    }
+                    catch (Exception e)
+                    {
+                        LOG.error("Failed to save authentication logo " + value, e);
```
labkey-susanh, line 171: Since the delete is outside of the try-catch, failing to save could mean the logo gets deleted with nothing replacing it. Is that what we want?

labkey-adam: @labkey-susanh `AttachmentService` requires delete before save, so the only other option I see is to move all the operations into a single transaction (I don't think moving delete into the try-catch will change the behavior). Failure to save would most likely be a startup property that doesn't point to a file, an admin error. I think it's fine that save failure results in a delete... same as leaving the value blank... especially since the failure will be noisily logged.

labkey-susanh (approved):

PR #3773, opened 2022-10-20
Less cryptic error message for unresolved sample Ids during assay import

labkey-susanh:
#### Rationale
When we do not find a sample ID during assay import, the error message can be a little less tech-like than the generic "Failed to convert" message. We could possibly use the less tech-y version of the message for other data as well, but being uncertain about the range of "failures to convert", I was aiming to keep this change small

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/1304
* https://github.com/LabKey/labkey-ui-components/pull/988
* https://github.com/LabKey/biologics/pull/1667

#### Changes
* Update the error messaging when failed to find sample ID.

### DIFF
--- a/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
+++ b/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
@@ -444,7 +444,7 @@ public abstract class AbstractAssayTsvDataHandler extends AbstractExperimentData
             final ContainerFilter cf = QueryService.get().getContainerFilterForLookups(container, user);
             final TableInfo dataTable = provider.createProtocolSchema(user, container, protocol, null).createDataTable(cf);
 
-            Map<ExpMaterial, String> inputMaterials = checkData(container, user, dataTable, dataDomain, rawData, settings, resolver);
+            Map<ExpMaterial, String> inputMaterials = checkData(container, user, dataTable, dataDomain, rawData, settings, resolver, cf);
 
             List<Map<String, Object>> fileData = convertPropertyNamesToURIs(rawData, dataDomain);
 
@@ -708,7 +708,8 @@ public abstract class AbstractAssayTsvDataHandler extends AbstractExperimentData
                                                Domain dataDomain,
                                                List<Map<String, Object>> rawData,
                                                DataLoaderSettings settings,
-                                               ParticipantVisitResolver resolver)
+                                               ParticipantVisitResolver resolver,
+                                               ContainerFilter containerFilter)
             throws ValidationException, ExperimentException
     {
         final ExperimentService exp = ExperimentService.get();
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
```diff
@@ -988,7 +989,10 @@ else if (o instanceof MvFieldWrapper mvWrapper)
                         Object remapped = cache.remap(lookupTable, (String)o, true);
                         if (remapped == null)
                         {
-                            errors.add(new PropertyValidationError("Failed to convert '" + pd.getName() + "': Could not translate value: " + o, pd.getName()));
+                            if (pd.getConceptURI() != null && SAMPLE_CONCEPT_URI.equals(pd.getConceptURI()))
+                                errors.add(new PropertyValidationError(  o + " not found in the current context.", pd.getName()));
```
XingY, line 994: This special treatment for SAMPLE_CONCEPT_URI seems a bit random. Can you instead leave this be and customize the msg in on the client side in resolveErrorMessage?

labkey-susanh: Could do, but this seems a better error message in general, whether importing from LKS or the applications, so I chose to make the change server side.

XingY (approved):

api/src/org/labkey/api/exp/PropertyColumn.java
```diff
@@ -63,9 +63,9 @@ public PropertyColumn(@NotNull PropertyDescriptor pd, @NotNull TableInfo tinfoPa
     /**
      * @param container container in which the query is going to be executed. Used optionally as a join condition, and
      * to construct the appropriate TableInfo for lookups.
-     * @param joinOnContainer when creating the join as a LookupColumn, whether or not to also match based on container 
+     * @param joinOnContainer when creating the join as a LookupColumn, whether or not to also match based on container
      */
-    public PropertyColumn(@NotNull final PropertyDescriptor pd, @NotNull final ColumnInfo lsidColumn, final Container container, User user, boolean joinOnContainer)
+    public PropertyColumn(@NotNull final PropertyDescriptor pd, @NotNull final ColumnInfo lsidColumn, final Container container, User user, boolean joinOnContainer, ContainerFilter containerFilter)
```
labkey-nicka: nit: Mark `@Nullable ContainerFilter cf` in the signatures updated with this change.

labkey-nicka (approved):

XingY (approved):

api/src/org/labkey/api/exp/PropertyColumn.java
labkey-susanh: Done.

PR #3777, opened 2022-10-21
Add script engine metrics

labkey-matthewb:
#### Rationale
Measure usage of "alternate" R script engines (docker, rserve)

#### Related Pull Requests
* 

#### Changes
*


labkey-matthewb:
Curious about opinion on using array vs map {"1":{}, "2":{}}

labkey-jeckels (changes requested):
I haven't done any work to support array-based values in metrics so far, so I recommend sticking with maps/objects.

Additionally, I recommend using a meaningful key instead of 1, 2, etc. I know it's a little harder here because users are allowed to have multiple configs for some. I suggest using the factory class name as the key, and adding "2", "3", etc as a suffix if there are multiple for a given type.

labkey-jeckels (approved):
Looks good, thanks.

PR #3782, opened 2022-10-24
Fix various FlowTests broken by change to use Supplier<Container>

labkey-jeckels:
#### Rationale
My change to use a Supplier<Container> instead of holding a Container reference in IndentifiableBase broke Jackson serialization. Example test failures:

https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyCPostgres/2068558?hideProblemsFromDependencies=false&expandBuildTestsSection=true&hideTestsFromDependencies=false&expandBuildChangesSection=true

#### Changes
* Use a Jackson-annotated class instead of a lambda

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -1738,7 +1738,8 @@ public class Container implements Serializable, Comparable<Container>, Securable
     {
         private final GUID _entityId;
 
-        public ContainerSupplier(GUID entityId)
+        @JsonCreator
+        public ContainerSupplier(@JsonProperty("entityId") GUID entityId)
         {
             _entityId = entityId;
         }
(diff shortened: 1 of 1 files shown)


labkey-matthewb (approved):

api/src/org/labkey/api/data/Container.java
```diff
@@ -1731,6 +1731,23 @@ public void setExpirationDate(LocalDate expirationDate)
...
+        private final GUID _entityId;
+
+        @JsonCreator
+        public ContainerSupplier(@JsonProperty("entityId") GUID entityId)
+        {
+            _entityId = entityId;
+        }
+
+        @Override
+        public Container get()
```
labkey-adam: Annotate as `@Nullable`?

labkey-adam (approved):

PR #3789, opened 2022-10-26
Label NameExpression column as "Naming Pattern"

labkey-susanh:
#### Rationale
We've been using "Naming Pattern" in LKSM since the beginning and it seems to resonate better than "Name Expression", so this PR updates the label on this field to "Naming Pattern", obviating the need for an override in LKSM and providing one more bit of consistency.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/1326

#### Changes
* Update field label in Data Classes and Sample Types.

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpSampleTypeTableImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpSampleTypeTableImpl.java
@@ -67,7 +67,6 @@ public class ExpSampleTypeTableImpl extends ExpTableImpl<ExpSampleTypeTable.Colu
             case LSID:
             case MaterialLSIDPrefix:
             case Name:
-            case AliquotNameExpression:
             case LabelColor:
             case MetricUnit:
             case AutoLinkTargetContainer:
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/api/ExpSampleTypeTableImpl.java
```diff
@@ -58,14 +58,15 @@ public MutableColumnInfo createColumn(String alias, Column column)
...
                 var columnInfo = wrapColumn(alias, _rootTable.getColumn("Container"));
                 columnInfo.setURL(new DetailsURL(new ActionURL(ExperimentController.ShowSampleTypeAction.class, getContainer())));
                 return columnInfo;
+            }
             case Description:
             case LSID:
             case MaterialLSIDPrefix:
             case Name:
-            case NameExpression:
             case AliquotNameExpression:
```
XingY: Does aliquot name expression needs label?

labkey-susanh: Indeed. Thank you.

XingY (approved):

PR #3801, opened 2022-10-31
Issue 46598: Use a specified directory structure for uploaded temp files

labkey-jeckels:
#### Rationale
We're leaking temp files. Some are from HTTP-based uploads and we can do a better job cleaning them up.

#### Changes
* Write HTTP uploads into a ./httpUploads subdirectory
* Clean out that subdirectory on server startup
* Delete files that failed virus scans


labkey-matthewb:
Consider hooking into ShutdownListener as well.

labkey-matthewb (approved):

PR #3816, opened 2022-11-02
Remove "no question marks in urls" experimental feature, helper methods, etc.

labkey-adam:
#### Rationale
We turned on the "no question marks in urls" experimental feature by default for 22.11; now we're removing the experimental feature entirely. See https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=1713


labkey-tchad (approved):
🥳

PR #3827, opened 2022-11-08
Projects in Sample Manager

XingY:
#### Rationale
Now that "Projects" are supported for other products and not just Biologics, the logic to check that a folder isProductProjectsEnabled needs to be updated to check for "Project" feature for the site's configured product tier that's active based on the folder's selected product type. "Project" feature is checked not just for queries, and with its tie to Container's folder type, it makes more sense to check projects in Container than as a QueryService. This PR also handles showing "Project" column in entity grids and bypassing them during import. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1016
* https://github.com/LabKey/platform/pull/3827
* https://github.com/LabKey/inventory/pull/601
* https://github.com/LabKey/recipe/pull/75
* https://github.com/LabKey/sampleManagement/pull/1371
* https://github.com/LabKey/biologics/pull/1716

#### Changes
* Added new "Projects" ProductFeature
* Moved isProductProjectsEnabled from QueryService to Container
* Updated isProductProjectsEnabled to check for "Projects" ProductFeature for the current folder type's product type
* Added "Folder" column to the default view for samples, data and assay runs, if the current folder is a product that supports "Projects" and that sub projects have been created in the current folder context
* Add util to report "Projects" metrics 
* Handle relabled container column during sample and data import


api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -577,6 +577,45 @@ public static Map<String, Integer> getFolderTypeNameContainerCounts(Container ro
         return nameCounts;
     }
 
+    @NotNull
+    public static Map<String, Integer> getProductProjectsCounts(@NotNull FolderType folderType)
```
labkey-susanh: Maybe `getProductProjectsMetrics` so as not to confuse with `getProductProjectsCount`?

api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -577,6 +577,45 @@ public static Map<String, Integer> getFolderTypeNameContainerCounts(Container ro
         return nameCounts;
     }
 
+    @NotNull
+    public static Map<String, Integer> getProductProjectsCounts(@NotNull FolderType folderType)
+    {
+        Container root = getRoot();
+        Map<String, Integer> metrics = new TreeMap<>();
+        List<Integer> counts = new ArrayList<>();
+        for (Container c : getAllChildren(root))
```
labkey-susanh: Might be more efficient to use `getAllChildrenWithModule` here. We create lots of containers that will never participate here (e.g., workbooks, and containers for ELNs).

api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -577,6 +577,45 @@ public static Map<String, Integer> getFolderTypeNameContainerCounts(Container ro
...
 
+    @NotNull
+    public static Map<String, Integer> getProductProjectsCounts(@NotNull FolderType folderType)
+    {
+        Container root = getRoot();
+        Map<String, Integer> metrics = new TreeMap<>();
+        List<Integer> counts = new ArrayList<>();
+        for (Container c : getAllChildren(root))
+        {
+            if (!c.getFolderType().getName().equals(folderType.getName()))
```
labkey-susanh, line 588: Should we also check that this container is a project? If users have created their subfolders as Biologics folders, say (even though that's not needed), we'll add metrics for those folders as well.

api/src/org/labkey/api/settings/AdminConsole.java
```diff
@@ -219,9 +221,16 @@ public static Set<ProductGroup> getProductGroups()
...
     public static Set<String> getProductFeatureSet()
+    {
+        return getProductFeatureSet(null);
+    }
+
+    public static Set<String> getProductFeatureSet(@Nullable String groupKey)
     {
         Set<String> productFeatures = new HashSet<>();
         AdminConsole.getProductGroups().forEach(group -> {
+            if (!StringUtils.isEmpty(groupKey) && !groupKey.equals(group.getKey()))
```
labkey-susanh: Maybe we want to add an `AdminConsole.getProductGroup(groupKey)` method?

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/ExpDataClassDataTestCase.jsp
```diff
@@ -269,7 +268,7 @@ private void testInsertIntoSubfolder(ExpDataClassImpl dataClass, TableInfo table
     }
 
     assertEquals(1, ret.size());
-    assertEquals(sub.getId(), ret.get(0).get("container"));
+    assertEquals(sub.getId(), ret.get(0).get("folder"));
```
XingY, line 271: @labkey-matthewb Can you please confirm the change made here (remove 'container' from row, and fetch 'folder' from ret) as well as SimpleTranslator (treat Folder column as already added columns to skip) makes sense? I don't think we support inserting data into a container by looking at the row data's container value, but rather it should be based on containerPath context. But maybe there is some scenarios that I don't know of.

api/src/org/labkey/api/data/ContainerManager.java
XingY: Good point, changed to get 1 generation from root only

PR #3831, opened 2022-11-08
RserveScriptEngine: local and shared-nothing

labkey-matthewb:
#### Rationale
Rserve currently assumes a remote machine with file shares for working files and pipeline.  This PR enables two new modes.
* Rserve running locally (no file mapping required)
* Rserve running remotely (or in a container) with no file sharing (working files must be copied)

#### Related Pull Requests
* 

#### Changes
*


api/src/org/labkey/api/reports/report/r/RserveScriptEngine.java
```diff
@@ -30,20 +32,65 @@
...
 public class RserveScriptEngine extends RScriptEngine
 {
     private static final Logger LOG = LogManager.getLogger(RserveScriptEngine.class);
 
-    private String localHostIP = "127.0.0.1";
-    private String localHostName = "localhost";
-    private static final int INITIAL_R_SESSONS = 5;
+    public enum ModusOperandi
+    {
+        Local(false, false),         // similar usage to classic "R.exe" reports: local execution with access to entire file system (no file mapping needed)
```
labkey-tchad, line 58: Should we even allow this `Local` mode? Even if RServe was running locally, wouldn't one want it running as a separate user than the user running tomcat; probably without permission to tomcat's temp files and LabKey file roots?

labkey-matthewb: This is a good question.  I was wondering if we could supercede (delete, deprecate) the other R report mechanisms (docker and local).   The answer I think is yes we could.  But you are correct that in a perfect world we would not have a “local, non-sandboxed” solution.  Maybe it is useful for small installations.  So this is a question about future direction of this feature area.  Is our priority to reduce complexity and get sites migrated to Rserve, or is our priority to sandboxing and migrating away from non-sandboxed solutions? @labkey-klum @labkey-hannah @labkey-jeckels

labkey-matthewb: I think we merge for now and then address this question.

api/src/org/labkey/api/pipeline/file/PathMapper.java
```diff
@@ -44,4 +47,23 @@
...
+    }
+
+    static URI pathToUri(String path) throws URISyntaxException
+    {
+        if (path.startsWith("file:"))
+            return new URI(path);
+        return FileUtil.getAbsoluteCaseSensitiveFile(new File(path)).toURI();
+    }
+
+    static String UriToPath(URI uri)
```
labkey-jeckels:
```suggestion
    static String uriToPath(URI uri)

```

api/src/org/labkey/api/reports/report/r/RserveScriptEngine.java
```diff
@@ -120,6 +219,57 @@ public static Object eval(ViewContext context, String function, String reportSes
...
 
+
+    protected void copyWorkingDirectoryToRemote(RConnection rconn) throws IOException
+    {
+        if (!mo.requiresCopyFiles())
+            return;
+
+        LOG.debug("Copy files in working directory to remote server");
+        File wd = getWorkingDir(getContext());
+        // recursive???
```
labkey-jeckels: I think the current code will fail if there are any directories in the working directory. The JavaDocs say the FileInputStream constructor will throw a FileNotFoundException. Maybe we should fail with a more helpful message about not supporting recursion?

labkey-jeckels (approved):

api/src/org/labkey/api/reports/report/r/RserveScriptEngine.java
labkey-jeckels: I agree that we should merge this now.

I vote for prioritizing migrating all SaaS instances to containerized Rserve. I'm not especially worried about sandboxing for on-premise solutions (though I'm not against it), but like the idea of reducing complexity and consolidating on Rserve everywhere. I don't have a good sense of how many existing R scripts expect to have access to the full file system though.

PR #3837, opened 2022-11-10
MicrosoftSqlServerVersion to capture supported versions and their properties

labkey-adam:
#### Rationale
An `enum` that captures all supported versions of Microsoft SQL Server and their properties (reported version number, tested, deprecated, allowed as primary data source, dialect factory to use, etc.) eases maintenance and testing of the dialect selection rules. See `PostgreSqlVersion` as the model.

#### Changes
* Add `MicrosoftSqlServerVersion` enum to capture supported versions and properties
* Replace version-specific `if` statements with generic logic that delegates to the enum properties
* Add junit test to verify expected dialects are returned for every version int
* Add admin warning for deprecated SQL Server version, consistent with PostgreSQL behavior

### DIFF
--- a/bigiron/src/org/labkey/bigiron/mssql/BaseMicrosoftSqlServerDialect.java
+++ b/bigiron/src/org/labkey/bigiron/mssql/BaseMicrosoftSqlServerDialect.java
@@ -89,6 +89,8 @@ abstract class BaseMicrosoftSqlServerDialect extends SqlDialect
     private volatile String _versionYear = null;
     private volatile Edition _edition = null;
 
+    private HtmlString _adminWarning = null;
+
     @SuppressWarnings("unused")
     enum Edition
     {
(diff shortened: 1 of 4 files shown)


labkey-adam:
Old dependencies sqlserver suite (running 2014) is all green as well: https://teamcity.labkey.org/viewLog.html?buildId=2109787&buildTypeId=LabKey_2211Release_Internal_InternalSuites_OldDependenciesSqlserver&tab=buildResultsDiv

bigiron/src/org/labkey/bigiron/mssql/BaseMicrosoftSqlServerDialect.java
```diff
@@ -1687,12 +1689,19 @@ public void afterCoreUpgrade(ModuleContext context)
...
+
     @Override
     public void addAdminWarningMessages(Warnings warnings)
     {
         ClrAssemblyManager.addAdminWarningMessages(warnings);
 
-        if (WarningService.get().showAllWarnings() || "2008R2".equals(_versionYear) || "2012".equals(_versionYear))
+        if (null != _adminWarning)
+            warnings.add(_adminWarning);
+        else if (WarningService.get().showAllWarnings())
```
labkey-jeckels, line 1704: In a future refactor, we should probably add a `showAllWarnings` parameter to `addAdminWarningMessages()` instead of expecting every implementation to separately check `showAllWarnings()`

labkey-jeckels (approved):

bigiron/src/org/labkey/bigiron/mssql/BaseMicrosoftSqlServerDialect.java
labkey-adam: Yep, that would be handy

PR #3842, opened 2022-11-11
Consolidate SVG filtering

labkey-adam:
#### Rationale
Filtering of SVG source is important to avoid Batik exceptions. We want that filtering to be applied consistently. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=46634

#### Changes
* Introduce simple `SvgSource` class that performs standard filtering
* Update all SVG conversion methods to take SvgSource parameters and update callers
* Move `renderSvgToPdf()` to `DocumentConversionService.svgPdf()` so it can be with its friends
* Move `DocumentConversionServiceImpl` to the Core module


core/src/org/labkey/core/reports/DocumentConversionServiceImpl.java
```diff
@@ -1,114 +1,144 @@
-/*
```
labkey-jeckels, line 1: Can you update the .gitattributes reference to this file? I'm hoping it will improve the diff

labkey-jeckels (approved):

core/src/org/labkey/core/reports/DocumentConversionServiceImpl.java
labkey-adam: Updated, but didn't seem to help. I could try reverting and re-committing...

labkey-jeckels: Probably not worth it. Not a huge file, and sometimes the diffs look better in IntelliJ than in GitHub

labkey-adam: Yes, it looks reasonable in IntelliJ

PR #3847, opened 2022-11-14
Simplify URLHelper/ActionURL JSON serialization

labkey-adam:
#### Rationale
`SignalDataRawTest` starting failing because the new `JSONObject` is more aggressive about serializing all fields of a Java object. The old JSONObject it would serialize `ActionURL` as a simple relative URL string (ActionURL.toString()) but the new JSONObject was throwing `org.json.JSONException: JavaBean object contains recursively defined member variable of key "originalPropertyValue"`. Implementing `JSONString` allows us to dictate the serialization; in this case, we simply quote toString().

Also, add a simple Junit test for ActionURL and expand the existing Jackson test to exercise Jackson + new JSONObject serialization.

### DIFF
--- a/api/src/org/labkey/api/data/JsonTest.java
+++ b/api/src/org/labkey/api/data/JsonTest.java
@@ -138,7 +138,7 @@ public class JsonTest extends Assert
         // NOTE: In both cases, the date value is deserialized as a string because JSON sucks
         JSONObject jsonOrgRoundTrip =  new JSONObject(jacksonToString);
         JSONObject jacksonRoundTrip = mapper.readValue(jacksonToString, JSONObject.class);
-        assertTrue(jsonOrgRoundTrip.similar(jacksonRoundTrip)); // Note: JSONObject.equals() doesn't work as expected
+        assertTrue("Not equivalent: " + jsonOrgRoundTrip + " vs. " + jacksonRoundTrip, jsonOrgRoundTrip.similar(jacksonRoundTrip)); // Note: JSONObject.equals() doesn't work as expected
     }
 
     @Test
--- a/api/src/org/labkey/api/data/JsonTest.java
+++ b/api/src/org/labkey/api/data/JsonTest.java
@@ -18,19 +18,23 @@ package org.labkey.api.data;
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.PropertyAccessor;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.json.old.JSONArray;
-import org.json.old.JSONObject;
+import org.json.JSONArray;
+import org.json.JSONObject;
 import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.labkey.api.util.JsonUtil;
+import org.labkey.api.view.ActionURL;
+import org.labkey.api.view.HttpView;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Calendar;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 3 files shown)


labkey-adam:
All SignalData tests pass with these changes: https://teamcity.labkey.org/viewLog.html?buildId=2115778&buildTypeId=LabKey_Trunk_LabkeyPremiumTrunk_GitModules_SignalDataPostgres95

api/src/org/labkey/api/data/JsonTest.java
```diff
@@ -86,19 +89,45 @@ public void jacksonTest()
...
 
-        final Date d = new GregorianCalendar(2011, 11, 3).getTime();
+        final Date d = new GregorianCalendar(2011, Calendar.DECEMBER, 3).getTime();
 
         JSONObject obj = new JSONObject();
         obj.put("str", "hello");
         obj.put("arr", new JSONArray(Arrays.asList("one", null, 3, new JSONObject(Collections.singletonMap("four", 4)))));
         obj.put("nul", (Object)null);
-        obj.put("d", d);
+//        obj.put("d", d);  //TODO: new JSONObject serializes date-times as ISO
```
labkey-susanh, line 130: I assume this is not to be done with this PR.

api/src/org/labkey/api/data/JsonTest.java
```diff
@@ -109,7 +138,16 @@ public void jsonOrgViaJackson() throws IOException
         // NOTE: In both cases, the date value is deserialized as a string because JSON sucks
         JSONObject jsonOrgRoundTrip =  new JSONObject(jacksonToString);
         JSONObject jacksonRoundTrip = mapper.readValue(jacksonToString, JSONObject.class);
-        assertEquals(jsonOrgRoundTrip, jacksonRoundTrip);
+        assertTrue(jsonOrgRoundTrip.similar(jacksonRoundTrip)); // Note: JSONObject.equals() doesn't work as expected
+    }
+
+    @Test
+    public void jsonOrgTest()
```
labkey-susanh, line 145: `oldJsonOrgTest`?

api/src/org/labkey/api/data/JsonTest.java
```diff
@@ -109,7 +138,16 @@ public void jsonOrgViaJackson() throws IOException
         // NOTE: In both cases, the date value is deserialized as a string because JSON sucks
         JSONObject jsonOrgRoundTrip =  new JSONObject(jacksonToString);
         JSONObject jacksonRoundTrip = mapper.readValue(jacksonToString, JSONObject.class);
-        assertEquals(jsonOrgRoundTrip, jacksonRoundTrip);
+        assertTrue(jsonOrgRoundTrip.similar(jacksonRoundTrip)); // Note: JSONObject.equals() doesn't work as expected
```
labkey-susanh: Is there a useful message that can be added to this test so it will be easier to see what's wrong if it fails?

labkey-susanh (approved):

api/src/org/labkey/api/data/JsonTest.java
labkey-adam: This is actually (starting to) test the new JSONObject (org.json.JSONObject). I threw in the comparison against old for now, but that will disappear when that library disappears.

api/src/org/labkey/api/data/JsonTest.java
labkey-adam: Correct... need to talk with AWOL Matt about this. Best to get the test fixed now, IMO.

api/src/org/labkey/api/data/JsonTest.java
labkey-adam: Sure

PR #3854, opened 2022-11-16
Refactor WarningProvider to pass in showAllWarnings flag

labkey-adam:
#### Rationale
There's no need for every WarningProvider implementation to check the `showAllWarnings()` flag. Instead, we can pass it in as a parameter. This is more convenient and a good reminder.

Also, standardize dialect warning messages.

### DIFF
--- a/api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
+++ b/api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
@@ -1884,7 +1884,8 @@ public abstract class PostgreSql91Dialect extends SqlDialect
         if (null != _adminWarning)
             warnings.add(_adminWarning);
         else if (showAllWarnings) // PostgreSqlDialectFactory.getStandardWarningMessage() is not accessible from here, so hard-code a sample warning
-            warnings.add(HtmlString.of("LabKey Server has not been tested against version 22.7. PostgreSQL 15.x is the recommended version."));
+            warnings.add(HtmlString.of("LabKey Server has not been tested against this version. PostgreSQL 15.x is the recommended version."));
+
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
```diff
@@ -1879,10 +1879,12 @@ public void setAdminWarning(HtmlString warning)
     }
 
     @Override
-    public void addAdminWarningMessages(Warnings warnings)
+    public void addAdminWarningMessages(Warnings warnings, boolean showAllWarnings)
     {
         if (null != _adminWarning)
             warnings.add(_adminWarning);
+        else if (showAllWarnings) // PostgreSqlDialectFactory.getStandardWarningMessage() is not accessible from here, so hard-code a sample warning
+            warnings.add(HtmlString.of("LabKey Server has not been tested against version 22.7. PostgreSQL 15.x is the recommended version."));
```
labkey-jeckels: 22.7? How old fashioned.
```suggestion
            warnings.add(HtmlString.of("LabKey Server has not been tested against this version. PostgreSQL 15.x is the recommended version."));

```

labkey-jeckels (approved):

api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
labkey-adam: That was PG 22.7, which is still a very long way off. But okay.

labkey-jeckels: Ah, right. I misread the message.

PR #3857, opened 2022-11-17
Remove support for PostgreSQL 10.x

labkey-adam:
#### Rationale
PostgreSQL 10.x reached end of life on November 10, 2022: https://www.postgresql.org/support/versioning/. Following our normal practice, we've announced that we'll remove support for 10.x in 22.12.

### DIFF
--- a/core/src/org/labkey/core/dialect/PostgreSqlDialectFactory.java
+++ b/core/src/org/labkey/core/dialect/PostgreSqlDialectFactory.java
@@ -22,7 +22,6 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.junit.Assert;
 import org.junit.Test;
-import org.labkey.api.collections.CsvSet;
 import org.labkey.api.data.DbScope;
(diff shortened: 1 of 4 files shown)


labkey-jeckels:
I can't quite remember our policy - does this mean we refuse to connect to 9.x and earlier, including for external data sources?

labkey-adam:
Yes, we're already refusing to connect to everything < PG 10.x. This means we'll stop connecting to 10.x as well. No special provision for external data sources in PostgreSQL land. I think the main reason we special case SQL Server is for licensing reasons, which don't apply to PG.

labkey-jeckels (approved):

labkey-jeckels:
@labkey-adam I can imagine scenarios where a third-party Postgres-based app isn't compatible with a new version of Postgres. But no need to handle that until it actually happens, and I like incentivizing people to be on supported versions.

labkey-adam:
Agreed... as you know, we keep all the "old" dialects around, making it easy to resurrect them if we ever have to.

PR #3858, opened 2022-11-18
Issue 46705: LKB: Sample Timeline is very slow or not loading at all

XingY:
#### Rationale
Workflow and Assay events are currently not included in the sample's timeline when viewed from child projects. This is due to the check of those features looking at the current folder's feature set, instead of its parent project. 

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/3858
- https://github.com/LabKey/sampleManagement/pull/1402

#### Changes
* check self + parent project for enabled features

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -1638,7 +1638,13 @@ public class Container implements Serializable, Comparable<Container>, Securable
         return false;
     }
 
-    public boolean isProductProjectsEnabled()
+    /**
+     * Check a feature is enabled at either its parent project, or itself
+     * @param feature
+     * @param atProjectOnly Only check Home Project for feature
+     * @return
+     */
+    public boolean isFeatureEnabled(ProductFeature feature, boolean atProjectOnly)
     {
         if (isWorkbook())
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/Container.java
```diff
@@ -1647,7 +1653,24 @@ public boolean isProductProjectsEnabled()
...
-        return project.getFolderType().isProductFeatureEnabled(ProductFeature.Projects);
+        boolean enabledAtProject = project.getFolderType().isProductFeatureEnabled(feature);
+        if (atProjectOnly || enabledAtProject)
+            return enabledAtProject;
+
+
+        return getFolderType().isProductFeatureEnabled(feature);
+    }
+
+    // don't use for ProductFeature.Projects, which should be atProjectOnly
```
labkey-nicka: Maybe this method should check if the `ProductFeature` is `ProductFeature.Projects` and return `isProductProjectsEnabled()` in that case?

XingY: good idea. done.

labkey-nicka (approved):

PR #3863, opened 2022-11-19
Add startup properties for full-text search settings

labkey-adam:
#### Rationale
Startup properties for full-text search settings are needed to ensure accurate search results after upgrades and to help clean up <tomcat>/temp. See:

* https://docs.google.com/document/d/1btU0_n16ZCxC1Y2F3SW3Jx1LasY4NY6mSPATKdtx32E/edit
* https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=45271
* https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=46598

#### Changes
* Support string substitution of system properties in full-text search index path. Add help popup to search admin page showing all properties and their escaped values.
* Add startup properties for all search settings/actions on the indexer admin page
* Log the reason we're deleting the index or initiating a full reindex
* Simplify some `SimpleRedirectAction`s to take a form
* `FileUtils.makeLegalName()`: add junit test, fix ending char bug, optimize, consolidate to a single list of illegal chars


api/src/org/labkey/api/admin/AdminBean.java
```diff
@@ -148,6 +156,34 @@ public static List<RecentUser> getRecentUsers()
...
+            )
+           .append(HtmlString.unsafe("</table>\n"));
+
+        return builder.getHtmlString();
+    }
+
+    private static HtmlString formatValue(String key, String value)
+    {
+        // Format GUID properties with monospace font
+        return key.endsWith("Guid") ? HtmlStringBuilder.of(HtmlString.unsafe("<span style=\"font-family:monospace\">"))
```
labkey-jeckels:
```suggestion
        return key.toLowerCase().endsWith("guid") ? HtmlStringBuilder.of(HtmlString.unsafe("<span style=\"font-family:monospace\">"))

```

api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -1070,25 +1071,28 @@ public static String makeLegalName(String name)
...
-            if(idx == name.length() - 1 && (ch == ' ' || ch == '.'))
-                ch = '_';
-
             ret[idx] = ch;
         }
+
+        //can't end with space (windows)
+        //can't end with period (windows)
+        int lastIndex = ret.length - 1;
+        char ch = ret[lastIndex];
```
labkey-jeckels, line 1097: This will blow up if we're passed an empty string. That's not a great filename so maybe it'll never happen, but probably worth bounds checking.

core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
```diff
@@ -80,17 +80,14 @@
...
-
-            AdminBean.getPropertyMap().entrySet().stream()
-                .map(e -> "<tr valign=top class=\"labkey-row\"><td>" + h(e.getKey()) + "</td><td>" + h(e.getValue()) + "</td></tr>\n")
-                .collect(Collectors.joining()) +
-
-            "</table>";
+        HtmlString shortNameHelp = HtmlStringBuilder.of("The header short name supports string substitution of specific server properties. For example, the value:")
+            .append(HtmlString.unsafe("<br><br><code>&nbsp;&nbsp;LabKey ${releaseVersion}</code><br><br>"))
+            .append("will currently result in this header text:")
+            .append(HtmlString.unsafe("<br><br><code>&nbsp;&nbsp;LabKey " + AdminBean.releaseVersion + "</code><br><br>"))
```
labkey-jeckels: Low risk, but shouldn't we HTML encode the value of AdminBean.releaseVersion

search/src/org/labkey/search/view/indexerAdmin.jsp
```diff
@@ -45,12 +51,27 @@ if (null == ss)
 }
 else
 {
+    HtmlString indexPathHelp = HtmlStringBuilder.of("The index path setting supports string substitution of specific server properties. For example, the value:")
+        .append(HtmlString.unsafe("<br><br><code>&nbsp;&nbsp;.\\temp\\${serverGuid}\\labkey_full_text_index</code><br><br>"))
+        .append("will currently result in this path:")
+        .append(HtmlString.unsafe("<br><br><code>&nbsp;&nbsp;.\\temp\\" + AdminBean.serverGuid + "\\labkey_full_text_index</code><br><br>"))
```
labkey-jeckels: Similar - should always be safe, but HTML encoding still seems like best practice

labkey-jeckels (approved):
A few minor nits but overall looks good.

PR #3866, opened 2022-11-22
Remove support for deprecated startup properties

labkey-adam:
#### Rationale
We deprecated the following startup properties (in favor of their alternatives) back in June, 2022:

- SiteRootSettings.siteRootFile -> SiteSettings.siteFileRoot
- LookAndFeelSettings.homeProjectInitWebparts -> SiteSettings.homeProjectWebparts
- SiteSettings.defaultDomain -> Authentication.DefaultDomain
- SiteSettings.experimentalFeature.disableGuestAccount -> ExperimentalFeature.disableGuestAccount

It's now time to stop accepting the deprecated properties. We'll also now throw a startup exception if unrecognized startup properties are provided.

We will wait to merge until the cloud deployment team has scrubbed all uses of the deprecated properties from deployment scripts, etc.

### DIFF
--- a/api/src/org/labkey/api/module/ModuleLoader.java
+++ b/api/src/org/labkey/api/module/ModuleLoader.java
@@ -2197,7 +2197,7 @@ public class ModuleLoader implements Filter, MemTrackerListener
                 .map(entry -> entry.getScope() + "." + entry.getName() + ": " + entry.getValue()).toList();
 
             if (!unknown.isEmpty())
-                throw new NotFoundException("Unknown startup propert" + (unknown.size() == 1 ? "y: " : "ies: ") + unknown);
+                throw new IllegalArgumentException("Unknown startup propert" + (unknown.size() == 1 ? "y: " : "ies: ") + unknown);
 
             // Failing this check indicates a coding issue, so execute it only when assertions are on
             assert checkPropertyScopeMapping();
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -2190,14 +2191,13 @@ public String getName()
...
                 .filter(entry -> null == entry.getStartupProperty())
-                .forEach(entry -> {
-                    // Suppress ERROR logging on this special snowflake. TODO: Remove this hack once Server Provisioner, Accounterer, etc. are updated. See Issue 45867 and Issue 45842
-                    Level logLevel = "SiteSettings".equals(entry.getScope()) && "experimentalFeature.disableGuestAccount".equals(entry.getName()) ? Level.WARN : Level.ERROR;
-                    _log.log(logLevel, "Unknown startup property: " + entry.getScope() + "." + entry.getName() + ": " + entry.getValue());
-                });
+                .map(entry -> entry.getScope() + "." + entry.getName() + ": " + entry.getValue()).toList();
+
+            if (!unknown.isEmpty())
+                throw new NotFoundException("Unknown startup propert" + (unknown.size() == 1 ? "y: " : "ies: ") + unknown);
```
labkey-jeckels: This will work fine in practice, but I always try to use NotFoundException when I need to send a 404 back and IllegalArgumentException or IllegalStateException for non-HTTP-status-code errors

labkey-jeckels (approved):

api/src/org/labkey/api/module/ModuleLoader.java
labkey-adam: Okay

PR #3871, opened 2022-11-23
Updates for handling of product projects

labkey-susanh:
#### Rationale
During testing of the incorporation of products into LKSM, a few bugs were founds and adjustments requested.  Here we implement part of the request to show the Project field by default in grids and detail views when there are product projects available.

#### Related Pull Requests
* https://github.com/LabKey/inventory/pull/619
* https://github.com/LabKey/labkey-ui-components/pull/1033
* https://github.com/LabKey/sampleManagement/pull/1411
* https://github.com/LabKey/biologics/pull/1747

#### Changes
* `getProductProjectsCount `-> `hasProductProjects` and adjust to not count notebook as product projects
* Update product configuration page to use a single radio group for less confusion
* Include the container column in picklist default views

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -1653,17 +1653,18 @@ public class Container implements Serializable, Comparable<Container>, Securable
     /**
      * Returns the subfolder count of the project container, if product projects feature is enabled in project
      */
-    public int getProductProjectsCount()
+    public boolean hasProductProjects()
     {
         if (!isProductProjectsEnabled())
-            return 0;
+            return false;
 
         Container project = getProject();
 
         if (project == null)
-            return 0;
+            return false;
 
-        return project.getChildren().size();
+        // need to exclude the notebook folders in particular here
+        return project.getChildren().stream().anyMatch(c -> c.getContainerType().isInFolderNav());
     }
 
     public boolean isDataspace()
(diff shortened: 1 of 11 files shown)


api/src/org/labkey/api/data/Container.java
```diff
@@ -1678,17 +1678,18 @@ public boolean isProductProjectsEnabled()
...
 
         Container project = getProject();
 
         if (project == null)
-            return 0;
+            return false;
 
-        return project.getChildren().size();
+        // need to exclude the notebook folders in particular here
+        return project.getChildren().stream().anyMatch(c -> c.getContainerType().isInFolderNav());
```
XingY, line 1692: Good point about notebook folders, should ContainerManager.getProductProjectsMetrics be updated to exclude notebook folders as well?

XingY: Also, should we exclude "Support" containers from here?

labkey-nicka (approved):

api/src/org/labkey/api/data/Container.java
labkey-susanh: Good thought about the metrics.   I'm going to verify whether current production deployments use the home project or not and would have the project-looking "Support" project there or not.

labkey-susanh: Solution has been to delete the /Home/support folder altogether as part of the SampleManagerFolderType configuration.

PR #3874, opened 2022-11-24
Issue 46796: NAb copy to study can be very slow on SQL Server

labkey-jeckels:
#### Rationale
A slow NAb query is causing test failures on SQL Server.

#### Changes
* Use the PropertyId instead of making the query find it multiple times


labkey-jeckels:
Closing in favor of the 22.11 PR at https://github.com/LabKey/platform/pull/3877

PR #3876, opened 2022-11-28
Issue 46050: Import alias failing silently on ingredients through file import with merge

labkey-martyp:
#### Rationale
Aliases are not currently supported for merge and replace operations. This PR flips that boolean to enable aliases for those operations and adds some junit tests. I've run nearly all our TC tests and I'm not seeing any issues related to this change. Manual testing has all been as expected.

#### Changes
* Flip alias boolean on merge and replace operations
* Add junit tests


api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -970,6 +970,8 @@ protected static Map<String, String> createMVMapping(Domain domain)
     @TestWhen(TestWhen.When.BVT)
     public static class TestCase extends Assert
     {
+        private boolean _useAlias = false;
```
labkey-jeckels, line 973: I believe that JUnit will create a single instead of the test class and run all tests on it. That means that once one of the tests runs that sets `_useAlias = true` it will stay true. I recommend adding a line to set it back to false in `resetList()` which should run `@Before` each individual test.

labkey-jeckels (approved):

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
labkey-martyp: It's actually instantiating a new class for each test. Seems a little counter to what you would expect though, given there's a @BeforeClass annotation

PR #3877, opened 2022-11-28
Issue 46796: NAb copy to study can be very slow on SQL Server

labkey-jeckels:
#### Rationale
A slow NAb query is causing test failures on SQL Server.

#### Changes
* Use the PropertyId instead of making the query find it multiple times


labkey-tchad:
No luck: https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_DailySuites_DailyDSqlserver/2139609

PR #3881, opened 2022-11-29
Allow saveRows to use an array of commands where each can supply containerPath

bbimber:
@labkey-jeckels Would it be possible to make a clone of this branch on the LK repo if you're willing to entertain this?

The rationale is that saveRows is a client API that lets the user submit an array of commands as a transaction. This PR augments each command object so it can supply containerPath, which would override the top-level containerPath. I added an integration test to test the basic usage. The broad use-case is to let the client submit a transacted list of commands that span containers.

### DIFF
--- a/query/src/org/labkey/query/controllers/QueryController.java
+++ b/query/src/org/labkey/query/controllers/QueryController.java
@@ -4119,7 +4119,7 @@ public class QueryController extends SpringActionController
         protected Container getContainerForCommand(JSONObject json)
         {
             Container container;
-            String containerPath = json.optString(PROP_CONTAINER_PATH);
+            String containerPath = StringUtils.trimToNull(json.optString(PROP_CONTAINER_PATH));
             if (containerPath == null)
             {
                 container = getContainer();
--- a/query/src/org/labkey/query/controllers/QueryController.java
+++ b/query/src/org/labkey/query/controllers/QueryController.java
@@ -4116,7 +4116,7 @@ public class QueryController extends SpringActionController
             String containerPath = json.optString(PROP_CONTAINER_PATH);
             if (containerPath == null)
             {
-                return getContainer();
+                container = getContainer();
             }
             else
             {
@@ -4127,6 +4127,7 @@ public class QueryController extends SpringActionController
                 }
             }
 
+            // Issue 21850: Verify that the user has at least some sort of basic access to the container. We'll check for more downstream
             if (!container.hasPermission(getUser(), ReadPermission.class) &&
                     !container.hasPermission(getUser(), DeletePermission.class) &&
                     !container.hasPermission(getUser(), InsertPermission.class) &&
--- a/query/src/org/labkey/query/controllers/QueryController.java
+++ b/query/src/org/labkey/query/controllers/QueryController.java
@@ -36,6 +36,10 @@ import org.jetbrains.annotations.Nullable;
 import org.json.old.JSONArray;
 import org.json.old.JSONException;
 import org.json.old.JSONObject;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
 import org.labkey.api.action.*;
 import org.labkey.api.admin.AdminUrls;
 import org.labkey.api.attachments.SpringAttachmentFile;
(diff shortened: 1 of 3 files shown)


labkey-nicka:
I like this.

bbimber:
@labkey-nicka: when I submit a PR to the LK fork TeamCity doesnt pick up on it. Would you be able to make a clone of 22.11_fb_saverows on the Labkey/platform repo, so TeamCity recognizes this and runs the tests?

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -7603,4 +7643,108 @@ controller.new NewQueryAction(),
...
+
+            Container project2 = ContainerManager.getForPath(PROJECT_NAME2);
+            if (project2 != null)
+            {
+                ContainerManager.deleteAll(project2, TestContext.get().getUser());
+            }
+        }
+
+        @Test
+        public void testCrossFolderSaveRows() throws Exception
```
labkey-jeckels: I'd like to have another test case that validates the command fails and gets rolled back when the user doesn't have permission to the target container in a second command

labkey-jeckels (changes requested):

labkey-jeckels:
@bbimber FB copy created: https://github.com/LabKey/platform/tree/22.11_fb_saverows

I can merge updates with additional test coverage if you add them

query/src/org/labkey/query/controllers/QueryController.java
bbimber: ok

bbimber:
@labkey-jeckels test just added

query/src/org/labkey/query/controllers/QueryController.java
labkey-jeckels: Thanks for the test addition. Initial review looks good. I've pulled it into the LK repo to let TeamCity validate it.

bbimber:
@labkey-jeckels, thoughts on interpreting the results of these? The EHR tests all have a pre-existing WNPRC_EHR Test failure. 
 BVT-A has a bunch of DomainDesigner failures:

https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtASqlserver/2143836?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildTestsSection=true&expandBuildChangesSection=true

I dont see any reason these should be related to this change, but the last release22.11-SNAPSHOT BVT-A did pass: https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtAPostgres?mode=builds

bbimber:
@labkey-jeckels, actually i see the issue. can you please merge my last commit into the labkey side of the fork? i think this will fix the issue. prior to this change, when no containerPath was provided, it was interpreting this as empty string, not null, and this ended up converting to the root container. Obviously not desired and now fixed.

labkey-jeckels:
Update merged. The Postgres BVT-A passed and the SQLServer failure seems unrelated and a problem with FirefoxDriver. I queued a new build to confirm.

bbimber:
@labkey-jeckels: looks like the BVTs passed now

labkey-jeckels (approved):

bbimber:
Thanks @labkey-jeckels

bbimber:
@labkey-jeckels: I found a subtle problem with this. As-is, the JS trigger script layer initializes itself and it assumes the effective container is the ViewContext's container. In effect:

- core/resources/scripts/labkey/init.js is called by the ScriptTrigger._try
- This calls PageFlowUtil.jsInitObject()
- @labkey-adam apparently already noted the problem (see line 2260). Even if the TableInfo is using a different container, this code always defaults back to the view container.

I've been poking around this code for a bit and dont see an obvious way to fix that. We could add new properties in ScriptTrigger ~ line 250. However, labkey.js calls init.js using require(), and JS modules called via require() dont seem to have access to the scriptContext the java code injects. I'm seeing if we could pass context into the modules from require(), but no luck yet.

Are you aware of a way to force context into the JS scripts that might help here? Thanks for any ideas. I'm going to keep looking into this as well.

bbimber:
See solutions in #3902

PR #3897, opened 2022-12-05
46857: Missing Data When Updating DataClasses

labkey-klum:
#### Rationale
This is an initial pass at trying to address this [issue](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=46857).

A number of problems were discovered in LKB when trying import data into the registry data structures (data classes) using the `update` option. All of these are related to trigger scripts in the app and how they are setup. AFAIK, at least for this issue, the trigger scripts in LKB are fine, all of the issues are in the platform code especially in the utility class `TriggerDataBuilderHelper`.

The data loss in the ticket was caused by line 120 of `TriggerDataBuilderHelper`, the parameter to `outputAllColumns` in the `CoerceDataIterator` would null out the remaining columns not specified in the incoming data. The other primary issue is that the class was only implemented to handle the insert case so modifying it to handle at least merge should address this and other related issues. 

There were some changes made to the `ExistingRecordDataIterator` and how it is set up in the DI since in the update case we are using it to fetch the current row.

I haven't done this yet but one thing Matt and I discussed was adding some validation to prevent update triggers from modifying the PK of the table. While none of the LKB trigger scripts attempt to do this, we probably either need this or should introduce a new, different contract API into the `Trigger` class.

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
@@ -818,15 +818,21 @@ public class ExpDataClassDataTableImpl extends ExpRunItemTableImpl<ExpDataClassD
 
             SimpleTranslator step0 = new SimpleTranslator(input, context);
             step0.setDebugName("step0");
-            step0.selectAll(Sets.newCaseInsensitiveHashSet("lsid", "dataClass", "genId"));
+
+            // if we are merging, to support existing row support we expect either the lsid to be present in the incoming data
+            Set<String> selectSet = context.getInsertOption().mergeRows
+                    ? Sets.newCaseInsensitiveHashSet("dataClass", "genId")
+                    : Sets.newCaseInsensitiveHashSet("lsid", "dataClass", "genId");
+            step0.selectAll(selectSet);
 
             TableInfo expData = svc.getTinfoData();
             ColumnInfo lsidCol = expData.getColumn("lsid");
 
             // TODO: validate dataFileUrl column, it will be saved later
 
-            // Generate LSID before inserting
-            step0.addColumn(lsidCol, (Supplier<String>) () -> svc.generateGuidLSID(c, ExpData.class));
+            // Generate LSID before inserting, but not when merging
+            if (!context.getInsertOption().mergeRows)
+                step0.addColumn(lsidCol, (Supplier<String>) () -> svc.generateGuidLSID(c, ExpData.class));
 
             // auto gen a sequence number for genId - reserve BATCH_SIZE numbers at a time so we don't select the next sequence value for every row
             ColumnInfo genIdCol = _dataClass.getTinfo().getColumn(FieldKey.fromParts("genId"));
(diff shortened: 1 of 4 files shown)


experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -818,15 +818,21 @@ public DataIterator getDataIterator(DataIteratorContext context)
...
             TableInfo expData = svc.getTinfoData();
             ColumnInfo lsidCol = expData.getColumn("lsid");
 
             // TODO: validate dataFileUrl column, it will be saved later
 
-            // Generate LSID before inserting
-            step0.addColumn(lsidCol, (Supplier<String>) () -> svc.generateGuidLSID(c, ExpData.class));
+            // Generate LSID before inserting, but not when merging
+            if (!context.getInsertOption().mergeRows)
+                step0.addColumn(lsidCol, (Supplier<String>) () -> svc.generateGuidLSID(c, ExpData.class));
```
XingY, line 835: So when merging, no LSID will be generated? What happens to new rows?

labkey-klum: I'm expecting the LSID to be included in the uploaded data, I'm open to suggestions and note that this is only needed by the `ExistingRecordDataIterator` since it relies on `QueryUpdateService.GetRow` to fetch existing records. See `ExpDataClassDataTableImpl.getRow`.

XingY: No... LSID will definitely not be included in the uploaded data, at least not for any LKSM use cases...

XingY: So for samples, there is an alternative set of unique keys: materialSourceId + name. For data: classid + name. Can those be used for existing record instead?

labkey-klum: @labkey-matthewb would know best how difficult it would be to have that class resolve the row by alternate keys, but yes I agree that this existing code won't work for LKSM.

labkey-matthewb: ExistingRecordDataIterator will use any set of columns as the key.  This is specified in the constructor.  I don't know if 
TriggerDataBuilderHelper currently has enough information to choose an alternate key.  The current usage is in an Experiment specific code path which knows to use the correct columns.

XingY: I've issued a separate PR targeting 22.11 that use hard coded alternative keys for samples and data in TriggerDataBuilderHelper: 
https://github.com/LabKey/platform/pull/3901

One thing to keep in mind is that merge is not an update. A datafile used for merge can include both new and existing records at the same time. So it's not correct to use "Update" triggers for merge actions. We are working on a feature to explicitly distinguish import vs merge vs update. Before that, merge should continue to trigger insert instead of update events.

PR #3901, opened 2022-12-06
46857: Missing Data When Updating DataClasses

XingY:
#### Rationale
DataClass data and samples uses ExpDataIterator to handle insert/merge into their materialized table as well as a provisioned table. The automatically generated LSID is the default key for those tables. When import with merge, TriggerDataBuilderHelper needs to outputAllColumns in the CoerceDataIterator, or else any columns not provided in the import file will get blanked out. For data types whose key are included in the import file (such as lists or dataset), the key can be used to query for existing data. For dataclass and samples, an alternative set of keys needs be to used since LSIDs are an auto generated during import and cannot be used to query existing data.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3897

#### Changes
* hard code TriggerDataBuilderHelper to use alternative keys for samples and dataclass during import with merge

### DIFF
--- a/api/src/org/labkey/api/dataiterator/TriggerDataBuilderHelper.java
+++ b/api/src/org/labkey/api/dataiterator/TriggerDataBuilderHelper.java
@@ -19,12 +19,16 @@ import org.labkey.api.data.AbstractTableInfo;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.triggers.Trigger;
+import org.labkey.api.exp.query.ExpDataClassDataTable;
+import org.labkey.api.exp.query.ExpMaterialTable;
 import org.labkey.api.query.BatchValidationException;
 import org.labkey.api.query.ValidationException;
 import org.labkey.api.security.User;
 
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * User: matthewb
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/dataiterator/TriggerDataBuilderHelper.java
```diff
@@ -117,9 +121,28 @@ public DataIterator getDataIterator(DataIteratorContext context)
             if (!_target.hasTriggers(_c))
                 return pre;
             pre = LoggingDataIterator.wrap(pre);
-            DataIterator coerce = new CoerceDataIterator(pre, context, _target);
+
+            boolean includeAllColumns = !(context.getInsertOption().mergeRows && (_target instanceof ExpMaterialTable || _target instanceof ExpDataClassDataTable));
```
labkey-nicka: I'm really not fond of type checking for Experiment tables here. These are generic data iterators that should not be aware of specific implementations. Can we go a bit deeper here and somehow expose this via the tableInfo interface? These addition of table-specific columns makes me think that could possibly be a better route for abstraction.

labkey-nicka (approved):

api/src/org/labkey/api/dataiterator/TriggerDataBuilderHelper.java
labkey-nicka: Thanks for that additional bit of refactoring. Hopefully, we can elevate the notion of merge keys further in the near future so this helper doesn't have to check for `ExpTable` at all. I could imagine datasets, which also support multiple other keys, could/should leverage this same mechanism (and/or this all should be converted to use whatever mechanism datasets use if it generically available) @labkey-matthewb.

labkey-klum (approved):

PR #3902, opened 2022-12-06
Add mechanism for TableInfo to directly provide the container

bbimber:
This is related to #3881. This description was revised as the PR evolved to reflect the final approach. The background is that when interacting with a table, JS triggers can be called. Currently, that JS code initializes and calls HttpView.currentView().getViewContext() in order to identify the effective container and user.  That's usually right but not ideal since technically the TableInfo could be from a different container than the current HttpRequest. One example that exposed this the new behavior from SaveRows to allow different commands to specify a distinct containerPath. 

Currently, the meat of the JS/trigger code happens in labkey.js and init.js. The core idea of this PR is:

1) In ScriptTrigger.java, take the current ContainerUser (which is what will match the UserSchema/TableInfo), and add a context property for the effective ContainerUser. 
2) Read these properties in RhinoService.getRuntimeScope() line ~827. 
3) Rather than use the old init.js, which called PageFlowUtil.jsInitObject() to load context, the new method adds a custom ModuleScript. Any JS trigger code can then call require('serverContext') to load that map with the values like container, user, etc. . This replaces the container-less call to PageFlowUtil.jsInitObject(). 

Note: this does allow JS scripts to get compiled and initialized without having ContainerUser set, since some background tasks seem to do this, but if those scripts attempt to call require('serverContext'), which also means require('LABKEY'), they will fail.


bbimber:
Note: I added a test case here to exercise and demonstrate this: https://github.com/LabKey/testAutomation/pull/1325. It might help with code review.

Another way to exercise this is to open an EHR data entry screen, since this performs lots of requests to the server via saveRows. Whenever you type into one of the fields, without actually hitting save, it'll begin to make requests. This is a good way to rapidly see the containerPath inside the JS triggers.

bbimber:
@labkey-jeckels The last commit is basically cleanup over the original proposal and it goes a bit further to completely replace the original no-arg PageFlowUtil.jsInitObject() with a cleaner require('serverContext') module. In addition to fixing the problem I encountered with cross-folder saveRows, I'd argue this is an improvement to how triggers function.

Is there a chance someone could copy this branch to the labkey side of the fork (including https://github.com/LabKey/testAutomation/pull/1325) so TeamCity will run tests?

labkey-jeckels:
@bbimber this is on my radar but I'm traveling this week and have limited bandwidth. I'll mirror the FB, hopefully later today.

bbimber:
@labkey-jeckels: no problem whatsoever. I fully understand these are not the highest priority and have no problem waiting. 

Separately: would you consider giving me (and is it even possible) kind of limited permissions to the LK repos so I could at least make feature branches that kick off TeamCity? Or perhaps it would be possible for branches created from the BimberLab or BimberLabInternal orgs to kick off TC, like occurs for the other modules? This would prevent you from needing to do unnecessary work and would let me ensure all tests pass before I bug LK staff.

labkey-jeckels:
LK copies of the FBs created. I'll review the change when I have a bit more time, and am very open to ways to streamline the process. I know @labkey-tchad has previously mentioned hooks as a way to trigger some of this automatically, but can also investigate permission updates next week.

bbimber:
thanks @labkey-jeckels. like i said above this contains some debugging code that I left to aid code review (and which will also error on Team City to ensure some of my assumptions are right), but I expect we will ultimately remove some of it. I also left in place an alternate method to poke in containerId for discussion, but I expect to remove it. There should be TODO and comments over these kind of things. 

Just to reiterate: while this problem was identified from cross-folder imports, this particular PR is really about ensuring that the UserSchema/TableInfo provides the container/user directly to the trigger script. In the current code, the container/user is pulled from the ViewContext and completely divorced from that TableInfo. These are virtually always the exact same. My cross folder import exposed one case where it isnt. It's entirely possible there are others, and most would be really easy to miss.

api/src/org/labkey/api/data/triggers/ScriptTrigger.java
```diff
@@ -47,6 +49,9 @@
  */
 /* package */ class ScriptTrigger implements Trigger
 {
+    // TODO: this string is also set in RhinoService, which is in internal. Can I share this somehow??
+    public static final String SCRIPT_CONTAINERUSER_KEY = "~~EffectiveContainerUser~~";
```
labkey-jeckels, line 52: Internal inherits from API, so RhinoService should be able to see this, except that the class is package-protected, not public. Moving the constant to the Trigger interface should let you reference it in both uses.

labkey-jeckels (changes requested):
@bbimber I'm catching up on this PR. Have you seen the test failures in the BVT-A suites? It's erroring when trying to populate the cache, which compiles the script. The current caches aren't scoped to a particular user or request so we can't have them retain a reference to a ContainerUser.

I haven't yet pulled this down to try locally but I'd expect we'd see similar failures on dev machines. Are you seeing them?

Maybe we can get away without calling getRuntimeScope() in RhineScriptEngine.compile(), since Context.compileReader() is overloaded and takes a scope argument, but then calls the variant that doesn't need the scope.

That issue aside, the ServerContextModuleScript version seems slightly preferable to me.

api/src/org/labkey/api/data/triggers/ScriptTrigger.java
bbimber: Thanks @labkey-jeckels. I agree ServerContextModuleScript is a little better.

Did you see my last commit (which was several days ago: https://github.com/LabKey/platform/pull/3902/commits/c16b05951abd9b5d1a8e715ae7d393bbfb8933bf)? it was addressing that test failure. I did two things, and I'd like to see a full TC run. I restored the default behavior in which script that do not explicitly have container set will fall back to HttpContext.getCurrentView(). I also added logging when this happens to see if it actually occurs outside that one test case. If the latter is true, and i think it is, I might be able to make passing container more explicit so future code that might enter this pathway doesnt run afoul. Or I'd at least like to make require('scriptContext') throw an exception. 

Anyway - I'll address these comments and clean things up to make what I think is a final proposal. Would you be able to merge https://github.com/LabKey/platform/pull/3902/commits/c16b05951abd9b5d1a8e715ae7d393bbfb8933bf into the LabKey side of the fork so I can see the TC results?

labkey-jeckels: Yes, I did and https://github.com/LabKey/platform/commit/c16b05951abd9b5d1a8e715ae7d393bbfb8933bf is already merged into the LK side. You can see it in the TeamCity change log from the most recent runs:

https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtAPostgres/2163355?buildTab=changes&showFiles=true&showGraph=false&expandRevisionsSection=true

What I missed is that it's noticing the problem in the post-test check for ERROR messages. That means it got all the way through and then noticed the logging. Happily, that means that there aren't likely to be other callers that will hit this. If we adjust the call to Context.compileReader() and perhaps explicitly set the content in the RhinoService.TestCase we should sidestep the issue.

I can merge an updated PR in the LK repo and kick off a larger test suite to give us more confidence.

bbimber: @labkey-jeckels, would you mind merging my last commit into the LK fork? Can you also please merge the TestAutomation PR: https://github.com/LabKey/testAutomation/pull/1325? The RhinoService test will fail here without it, since that PR adds a new trigger in simpletest module

labkey-jeckels: FBs in both repos are now current. I'll wait for the default set of tests to complete and can queue additional suites if they're looking good.

bbimber:
@labkey-jeckels looks like the BVTs passed: https://teamcity.labkey.org/project/LabKey_2211Release?mode=builds. Thoughts on next steps?

labkey-jeckels:
@bbimber Yes, BVT-A and -B look good. The EHR BVTs didn't run because of a build failure. I merged in the latest release22.11-SNAPSHOT changes into the FB for the platform repo that should clear it up. If the EHR BVTs pass on the upcoming run I can then fire off Dailies and full EHR suites to give us better coverage.

bbimber:
sounds good

bbimber:
@labkey-jeckels: thoughts on the FileWatcherPipelineTriggerType failure? Ian has some changes related to Filewatchermovepipelinejob.java here, but it's not clear to me if this is related:

https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Premium_Ehr_BvtEhrSqlserver/2171864?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildTestsSection=true&expandBuildChangesSection=true

bbimber:
@labkey-jeckels: i'm happy to investigate that failure, which is happening on this branch and not develop (though maybe it's possible it's some change not in this branch?); however, I dont have read access to premium module so I cant tell much about what that stack is trying to do.

bbimber:
> @labkey-jeckels: i'm happy to investigate that failure, which is happening on this branch and not develop (though maybe it's possible it's some change not in this branch?); however, I dont have read access to premium module so I cant tell much about what that stack is trying to do.

Also, I should have the dailies fixed in a sec: https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_DailySuites_DailyBPostgres/2172973?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true. These are just tests that need updating.

bbimber:
The dailies should be fixed with this TestAutomation PR: https://github.com/LabKey/testAutomation/pull/1325/commits/d4b32eb57a41ed53dac94a27c112140aa0148722

labkey-jeckels:
@bbimber I merged your testAutomation change into the LK FB. I added some additional logging for the FileWatcher test failure. I don't see how your changes would make it fail, but something is clearly happening in this branch

bbimber:
@labkey-jeckels. If this fails again, could I get read access to premium so I can have a look and run that locally? Is this TeamCity run going to go against the latest version of premium, in that that has related changes?

bbimber:
@labkey-jeckels: something does seem to be happening: https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Premium_ModulesSuites_PremiumSqlserver2014/2174589?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true.

is read access to premium possible?

bbimber:
@labkey-jeckels, I think this stack is telling us the problem. I'm happy to look into this and propose solutions, but I cant read the premium module repo. I could propose a dirty fix but I'd like to understand what it's trying to do to figure out the right solution.


```
ERROR PipelineJob              2022-12-15T18:59:15,469            JobThread-3.6 : (from pipeline job log file D:\teamcity\work\53ecd7367fa0a825\build\deploy\files\__FileWatcherTestProject\@files\fileWatcherMoveJob_2022-12-15_18-59-15.log: Error processing file watcher move)
java.util.EmptyStackException: null
	at java.util.Stack.peek(Stack.java:101) ~[?:?]
	at org.labkey.api.view.HttpView.currentView(HttpView.java:316) ~[api-22.11-SNAPSHOT.jar:?]
	at org.labkey.api.script.RhinoEngine.getRuntimeScope(RhinoService.java:846) ~[internal-22.11-SNAPSHOT.jar:?]
	at org.labkey.api.script.RhinoScriptEngine.eval(RhinoScriptEngine.java:207) ~[internal-22.11-SNAPSHOT.jar:?]
	at org.labkey.api.script.RhinoEngine.eval(RhinoService.java:866) ~[internal-22.11-SNAPSHOT.jar:?]
	at org.labkey.api.script.RhinoScriptEngine.eval(RhinoScriptEngine.java:256) ~[internal-22.11-SNAPSHOT.jar:?]
	at org.labkey.api.script.RhinoEngine.eval(RhinoService.java:874) ~[internal-22.11-SNAPSHOT.jar:?]
	at javax.script.AbstractScriptEngine.eval(AbstractScriptEngine.java:231) ~[java.scripting:?]
	at org.labkey.premium.trigger.FileWatcherMovePipelineJob.doWork(FileWatcherMovePipelineJob.java:268) ~[premium-22.11-SNAPSHOT.jar:?]
	at org.labkey.premium.trigger.FileWatcherMoveTask.run(FileWatcherMoveTask.java:29) ~[premium-22.11-SNAPSHOT.jar:?]
	at org.labkey.api.pipeline.PipelineJob.runActiveTask(PipelineJob.java:844) ~[api-22.11-SNAPSHOT.jar:?]
	at org.labkey.api.pipeline.PipelineJob.run(PipelineJob.java:1084) ~[api-22.11-SNAPSHOT.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539) ~[?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) ~[?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]
```

bbimber:
@labkey-jeckels, read access would still help, but I checked in what will probably fix it here (https://github.com/LabKey/platform/pull/3902/commits/93747cb2d5aa0920bfa3625cb03210d50349b312). This will fail tests b/c the error logging in RhinoService is still there, but I dont think the actual test will fail anymore.

The core change is that the code no longer assumes an HttpView will be set. These scripts wont have container set; however, those scripts probably never needed it. if a script calls require('serverContext'), or if it calls require('LABKEY') then it will die, but I'd bet these usages dont use that.

Could you please push my changes to the LK fork?

labkey-jeckels:
I've merged the latest change and queued another round of tests.

Thanks for finding that helpful logging output for the root cause. The file watcher test isn't trying to do anything fancy with containers -  it's just that it's running in a background thread (as you figured out) so there's no HttpView, HTTP request, ViewContext, etc.

bbimber:
@labkey-jeckels Due to cases like what filewatcher is apparently doing, I was hoping to find a more explicit way to poke-in ContainerUser (like overloading the method with a separate argument), but I just couldnt find a reasonable way to do that in the code with all the cases where one script references and loads another script. I think having the code unable to load the "LABKEY" library without having container, as happens now, is a reasonable compromise. 

Anyway, my changes will still cause test failures, but I think the only reason is the log.error() call I left in, which file watcher is probably still going to hit. We can remove that logging if everything passes otherwise.

bbimber:
@labkey-jeckels Looks like dailies just passed: https://teamcity.labkey.org/project/LabKey_2211Release?branch=triggerContainerPath&mode=builds. 

I checked in a change to drop the unnecessary log.error call, and it seems to me like it  should be good to go. Thoughts?

bbimber:
@labkey-jeckels is there anything further you would like from the PR? From what I can see in team city, tests seem to pass. My commit on Friday dropped some of the debugging logging, but did not change functionality otherwise.

labkey-jeckels:
@bbimber I just merged your logging change in the LK FB. I don't have any other requests at the moment, and should have time later today or tomorrow to do a final review.

labkey-jeckels:
Nick, since you know more about JS `require()` than me and looked at the earlier PR that necessitates this follow, can you take a look here too?

The main problem that Ben discovered is that the trigger scripts were pulling the Container from the ViewContext, which is no longer guaranteed to be a match with each individual command's containerPath.

internal/src/org/labkey/api/script/RhinoScriptEngine.java
```diff
@@ -291,9 +291,9 @@ public Object invokeMethod(Object thiz, String name, Object... args)
                 thiz = cx.toObject(thiz, topLevel);
             }
 
-            Scriptable engineScope = getRuntimeScope(context);
+            // TODO: review this
```
labkey-nicka: This looks good to me. Can remove this TODO.

internal/src/org/labkey/api/script/RhinoScriptEngine.java
```diff
@@ -408,8 +408,7 @@ public CompiledScript compile(java.io.Reader script) throws ScriptException {
                 filename = "<Unknown Source>";
             }
 
-            Scriptable scope = getRuntimeScope(context);
-            Script scr = cx.compileReader(scope, preProcessScriptSource(script), filename, 1, null);
+            Script scr = cx.compileReader(preProcessScriptSource(script), filename, 1, null);
```
labkey-nicka, line 410: Interesting we were using the deprecated variant here which just throws the `scope` on the floor.

labkey-nicka (approved):
Overall, nice approach. I like the consolidation into `serverContext`. I'm not clear on how the initial comments on this PR now relate to the current implementation so you might consider revising that to match especially since there is reference to this PR in the code.

labkey-jeckels:
@bbimber if you can update the PR description based on the final approach I'll get it merged. Something is going wrong with the TeamCity build/test reporting to the PR. It may be easiest to go with a PR from the LK copy of the FB. The downside is that it'll show me as the author instead of you.

https://github.com/LabKey/platform/pull/3942

labkey-tchad:
> @bbimber if you can update the PR description based on the final approach I'll get it merged. Something is going wrong with the TeamCity build/test reporting to the PR. It may be easiest to go with a PR from the LK copy of the FB. The downside is that it'll show me as the author instead of you.
> 
> #3942

@labkey-jeckels Once you merged 22.11-SNAPSHOT into the other branch it was no longer sharing commits with this branch so this PR stopped seeing status reported from TeamCity.
@bbimber If you force-push the changes from Josh's branch, this PR will see the existing TeamCity results. Assuming you have a `labkey` remote pointing at our fork, it would be something like this:
```
git fetch labkey
git checkout 22.11_fb_triggerContainerPath
git reset --hard 7da312698ebfccb96c2fe1a67e3dad9a9b4aea25
git push --force
```

bbimber:
@labkey-jeckels and @labkey-nicka OK, top-level comment has been edited. I'm fine moving to the other PR if you prefer, but I cant edit that one directly.

internal/src/org/labkey/api/script/RhinoScriptEngine.java
bbimber: done

internal/src/org/labkey/api/script/RhinoScriptEngine.java
bbimber: no changes actually needed here, right?

labkey-nicka: No. Thanks.

labkey-jeckels:
Alternate version merged with https://github.com/LabKey/platform/pull/3942

PR #3908, opened 2022-12-07
Issue 46465: Fix behavior of grid actions when using a filtered set of selections

labkey-susanh:
#### Rationale
When a user selects items on a grid and then filters the grid so only a subset of the items is shown, the expectation is that any actions performed will use the selections shown in the filtered grid. For many actions, however, we pick up the original set of selections for performing actions. This PR modifies the behavior in these cases to use a snapshot selection in order to get at the filtered items.

#### Related Pull Requests
- https://github.com/LabKey/sampleManagement/pull/1444
- https://github.com/LabKey/biologics/pull/1778
- https://github.com/LabKey/labkey-ui-components/pull/1046
- https://github.com/LabKey/inventory/pull/637

#### Changes
* When exporting, we want to obey filters as well as selection key if present
* Allow for use of snapshot selections when getting operation confirmation data

### DIFF
--- a/api/src/org/labkey/api/data/RenderContext.java
+++ b/api/src/org/labkey/api/data/RenderContext.java
@@ -479,7 +479,11 @@ public class RenderContext implements Map<String, Object>, Serializable
             container = _viewContext.getContainer();
         }
         if (_currentRegion != null && _showRows == ShowRows.SELECTED || _showRows == ShowRows.UNSELECTED)
+        {
             buildSelectedFilter(filter, tinfo, _showRows == ShowRows.UNSELECTED);
+            // Issue 46465: If you've selected items then filtered the grid, we want to respect the filter
+            filter.addUrlFilters(url, name, displayColumns, user, container);
+        }
         else
             filter.addUrlFilters(url, name, displayColumns, user, container);
 
--- a/api/src/org/labkey/api/data/RenderContext.java
+++ b/api/src/org/labkey/api/data/RenderContext.java
@@ -481,11 +481,9 @@ public class RenderContext implements Map<String, Object>, Serializable
         if (_currentRegion != null && _showRows == ShowRows.SELECTED || _showRows == ShowRows.UNSELECTED)
         {
             buildSelectedFilter(filter, tinfo, _showRows == ShowRows.UNSELECTED);
-            // Issue 46465: If you've selected items then filtered the grid, we want to respect the filter
-            filter.addUrlFilters(url, name, displayColumns, user, container);
         }
-        else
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/data/RenderContext.java
```diff
@@ -479,7 +479,11 @@ public SimpleFilter buildFilter(TableInfo tinfo, List<ColumnInfo> displayColumns
             container = _viewContext.getContainer();
         }
         if (_currentRegion != null && _showRows == ShowRows.SELECTED || _showRows == ShowRows.UNSELECTED)
+        {
             buildSelectedFilter(filter, tinfo, _showRows == ShowRows.UNSELECTED);
+            // Issue 46465: If you've selected items then filtered the grid, we want to respect the filter
+            filter.addUrlFilters(url, name, displayColumns, user, container);
```
cnathe: this line is the same as the else statement. could this instead just be moved outside of the if/else?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -7536,12 +7579,26 @@ public void setPicklistName(String picklistName)
...
         public Set<Integer> getIds(boolean clear)
         {
             if (_rowIds != null)
                 return _rowIds;
-            Set<Integer> selectedIds = DataRegionSelection.getSelectedIntegers(getViewContext(), getDataRegionSelectionKey(), clear);
+            Set<Integer> selectedIds;
+            if (_useSnapshotSelection)
+                selectedIds = new HashSet<>(DataRegionSelection.getSnapshotSelectedIntegers(getViewContext(), getDataRegionSelectionKey()));
+            else
+                selectedIds = DataRegionSelection.getSelectedIntegers(getViewContext(), getDataRegionSelectionKey(), clear);
```
cnathe, line 7574: looks like this type of code block is repeated 3 times in this PR. any way to refactor it to be shared or add a static util helper somewhere in this class?

cnathe (changes requested):

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: Created a new form that does this override for two of the three cases. The third case, because it has to handle picklists, still has its own override.

api/src/org/labkey/api/data/RenderContext.java
labkey-susanh: Sure. Simpler.

cnathe (approved):

PR #3911, opened 2022-12-08
Issue 46830: Decouple SampleNameGenCounter from the ContextListener logger

XingY:
#### Rationale
DbSequence.Preallocate allow reserving a batch of ids from a sequence to avoid excess DB actions for certain tasks. Those Preallocated sequence gets added to ContextListener and will be cleaned up during server shut down. The withCounter naming syntax uses DBSequence to get the count of a unique string value. Preallocates are used and put in a local cache to allow efficient name generation that uses withCounter. For Preallocates used by withCounter, we should properly clean them up after the name generation task is done, instead of relying on ContextListener to clean them up.

#### Related Pull Requests
* 

#### Changes
* move counterSequence cache from CounterExpressionPart to State
* explicitly clean up counterSequences on State.close
* use a single shutdownlistener for all Preallocates

### DIFF
--- a/api/src/org/labkey/api/data/NameGenerator.java
+++ b/api/src/org/labkey/api/data/NameGenerator.java
@@ -1327,6 +1327,7 @@ public class NameGenerator
         private int _rowNumber = 0;
         private final Map<Tuple3<String, Object, FieldKey>, Object> _lookupCache;
         private final Map<String, ArrayList<Object>> _ancestorCache;
+        private final Map<String, Map<String, DbSequence>> _prefixCounterSequences;
 
         private State(boolean incrementSampleCounts)
         {
@@ -1340,12 +1341,30 @@ public class NameGenerator
             _user = User.getSearchUser();
             _lookupCache = new HashMap<>();
             _ancestorCache = new HashMap<>();
+            _prefixCounterSequences = new HashMap<>();
         }
 
         @Override
         public void close()
         {
             _rowNumber = -1;
+
+            for (Map.Entry<String, Map<String, DbSequence>> prefixCounterSequence : _prefixCounterSequences.entrySet())
+            {
+                String counterSeqPrefix = prefixCounterSequence.getKey();
+                Map<String, DbSequence> counterSequences = prefixCounterSequence.getValue();
+                for (Map.Entry<String, DbSequence> counterSequence: counterSequences.entrySet())
+                {
+                    DbSequence seq = counterSequence.getValue();
+                    if (seq instanceof DbSequence.Preallocate)
+                    {
+                        ((DbSequence.Preallocate) seq).shutdownPre();
+                        ((DbSequence.Preallocate) seq).shutdownStarted();
+                        ((DbSequence.Preallocate) seq).done();
+                        DbSequenceManager.invalidatePreallocatingSequence(_container, counterSeqPrefix + counterSequence.getKey(), 0);
+                    }
+                }
+            }
         }
 
         private String nextName(Map<String, Object> rowMap,
@@ -1432,7 +1451,11 @@ public class NameGenerator
             // allow using alternative expression for evaluation.
             // for example, use AliquotNameExpression instead of NameExpression if sample is aliquot
             FieldKeyStringExpression expression = altExpression != null ? altExpression : _parsedNameExpression;
-            String name = expression.eval(ctx);
+            String name = null;
+            if (expression instanceof NameGenerationExpression)
+                name = ((NameGenerationExpression) expression).eval(ctx, _prefixCounterSequences);
+            else
+                name = expression.eval(ctx);
             if (name == null || name.length() == 0)
                 throw new IllegalArgumentException("The data provided are not sufficient to create a name using the naming pattern '" + expression.getSource() + "'.  Check the pattern syntax and data values.");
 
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1340,12 +1341,30 @@ private State(boolean incrementSampleCounts)
...
+            for (Map.Entry<String, Map<String, DbSequence>> prefixCounterSequence : _prefixCounterSequences.entrySet())
+            {
+                String counterSeqPrefix = prefixCounterSequence.getKey();
+                Map<String, DbSequence> counterSequences = prefixCounterSequence.getValue();
+                for (Map.Entry<String, DbSequence> counterSequence: counterSequences.entrySet())
+                {
+                    DbSequence seq = counterSequence.getValue();
+                    if (seq instanceof DbSequence.Preallocate)
+                    {
+                        ((DbSequence.Preallocate) seq).shutdownPre();
```
labkey-matthewb: This really belongs in DbSequence.  Maybe seq.sync().  I don't think we really need to call done().

Maybe like this?

        public synchronized void sync()
        {
            if (null != _lastReservedValue && _currentValue != _lastReservedValue)
            {
                DbSequenceManager.setSequenceValue(this, _currentValue);
                _lastReservedValue = _currentValue;
            }
        }

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1340,12 +1341,30 @@ private State(boolean incrementSampleCounts)
...
             _rowNumber = -1;
+
+            for (Map.Entry<String, Map<String, DbSequence>> prefixCounterSequence : _prefixCounterSequences.entrySet())
+            {
+                String counterSeqPrefix = prefixCounterSequence.getKey();
+                Map<String, DbSequence> counterSequences = prefixCounterSequence.getValue();
+                for (Map.Entry<String, DbSequence> counterSequence: counterSequences.entrySet())
+                {
+                    DbSequence seq = counterSequence.getValue();
+                    if (seq instanceof DbSequence.Preallocate)
```
labkey-matthewb: Instead of using instanceof implement base method (see below) which noops .

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -2094,23 +2171,34 @@ public String getValue(Map map)
...
 
                     if (_getNonConflictCountFn != null)
                         existingCount = _getNonConflictCountFn.apply(prefix);
+                    if (counterSequences == null) // no cache
+                        counterSeq = DbSequenceManager.get(_container, _counterSeqPrefix + prefix);
+                    else
+                    {
+                        // use PreallocatingSequence to handle generating multiple aliquots from the same sample
+                        // PreallocatingSequences opened by CounterExpressionPart are cleaned up by State.close()
+                        counterSeq = DbSequenceManager.getPreallocatingSequence(_container, _counterSeqPrefix + prefix);
```
labkey-matthewb, line 2179: Sorry, I'm not that familiar with NameGenerator.  This is a permanent counter and can't be implemented with an in memory counter?  Are we generating permanent counters on the fly based on user entered name expressions?

labkey-matthewb (changes requested):

api/src/org/labkey/api/data/NameGenerator.java
XingY: I need to call done() to remove this sequence from ContextListener.shutdownListeners. If not, user will get a log entry about this sequence when server shuts down.

api/src/org/labkey/api/data/NameGenerator.java
XingY: Those are permanent counters based on name expression. It cannot be in memory counter because we need to how how many times a substring occurred not just in this run, but in all previous actions.

api/src/org/labkey/api/data/NameGenerator.java
labkey-matthewb: We could also change the shutdownStarted() to just call sync() so that it noops.  What is causing logging???  

Is having a lot of ContextListeners causing a problem?  It's odd that we have one per sequence, instead of one static listener for DbSequence.

XingY: Each opened Preallocate gets added to ContextListener._shutdownListeners. I believe the intention is do call the sync so that preallocated but unused sequence gets reset. Each shutdownListener would log a "Calling xx shutdownPre()" entry in the log. If there are lots of Preallocate on the server, the shut down process takes a long time to clean up those preallocated, and the large number of log entries pollutes the sever log file. 

There are really 2 problems.
1.  Preallocate are almost never synced and have to rely on _shutdownListeners to clean up / sync.
2. One _shutdownListener per Preallocate, which causes excessive logging. 

For the 1st problem, this PR try to address ones caused by withCounter usage. Because withCount is used for Aliquot naming pattern, the number of those sequences can be huge. There are other usages of Preallocate in the app that lingers until server gets shut down, but hopefully there won't be as many of those. 
For the 2nd problem, I agree a single listener for all sequences sounds a good way to go.

labkey-matthewb: I'll look at turning this into one listener.  That should fix a lot of sins.

labkey-matthewb: OK, I got rid of all the listeners.  That should help.  I think we can just change NameGenerator to use sync() and not worry about the listeners now.

XingY: Great! This looks much better now. I've adjusted NameGenerator accordingly.

labkey-matthewb (approved):

labkey-susanh (approved):

PR #3914, opened 2022-12-12
App logout page redirect back to app when container is an app project subfolder

cnathe:
#### Rationale
Since Project subfolders are of folder type Collaboration the project-start redirect doesn't work as it does with the main app folder. This PR fixes that by adding a special case check for this in the success.html to create the URL directly to the app instead of via project-start.

#### Changes
* add check in success.html for app Project subfolder for "Return to Application" button href

### DIFF
--- a/core/resources/views/success.html
+++ b/core/resources/views/success.html
@@ -16,6 +16,12 @@
     }
 
     document.getElementById('success-header').innerHTML = header;
+
+    // if logged out of a product and container is not the project, redirect back to app instead of project-start
+    if (product && LABKEY.container.parentPath !== '/') {
+        startHref = LABKEY.ActionURL.buildURL(product.replaceAll(' ', '').toLowerCase(), 'app');
+    }
+
     document.getElementById('success-content').innerHTML = content
         + (LABKEY.user.isGuest ? '<a class="labkey-button primary" href="' + startHref + '">Return to Application</a>' : '');
 </script>


core/resources/views/success.html
```diff
@@ -16,6 +16,12 @@
     }
 
     document.getElementById('success-header').innerHTML = header;
+
+    // if logged out of a product and container is not the project, redirect back to app instead of project-start
+    if (product && LABKEY.container.parentPath !== '/') {
+        startHref = LABKEY.ActionURL.buildURL(product.replaceAll(' ', '').toLowerCase(), 'app');
```
labkey-susanh, line 22: This works, but might be a little fragile. Server side, I would have gone with finding the project and using its start URL, but I don't think that's readily available within JS.

labkey-susanh (approved):

core/resources/views/success.html
labkey-susanh: Instead of `LABKEY.container.parentPath !== '/'`, it looks like you could use `LABKEY.container.type === 'project'`.

cnathe: Yes, I initially went down the route of calling the API to get the project info via the LABKEY.Security.getContainers, but at the point of this page the user is Guest so that didn't work. Also, the LABKEY.container.type is undefined for guest users as the LABKEY.container object has limited info in this case....so I went with the given implementation because of those restrictions.

labkey-susanh: Sounds reasonable.  I didn't realize guest users would see the container type either.

PR #3915, opened 2022-12-12
Handle update and clarify merge - app UI

XingY:
#### Rationale
Merge has been misrepresented as Update in the product UI for ease of understanding, which causes confusion when users are expecting an update action. This PR modifies the app UI to allow user to Update data from file under an experimental flag. The server side work to support Update from file will be done in a follow up story. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1053
* https://github.com/LabKey/platform/pull/3915
* https://github.com/LabKey/inventory/pull/642
* https://github.com/LabKey/sampleManagement/pull/1457
* https://github.com/LabKey/biologics/pull/1791

#### Changes
* added InsertOption.UPDATE
* added new experimental flag ALLOW_IMPORT_WITH_UPDATE
* added supportMerge to TableInfo to allow showing merge toggle on UI

### DIFF
--- a/study/src/org/labkey/study/query/DatasetTableImpl.java
+++ b/study/src/org/labkey/study/query/DatasetTableImpl.java
@@ -468,6 +468,8 @@ public class DatasetTableImpl extends BaseStudyTable implements DatasetTable
         addColumn(datasetId);
 
         addFolderColumn();
+
+        setSupportMerge(true);
     }
 
 
--- a/study/src/org/labkey/study/query/DatasetTableImpl.java
+++ b/study/src/org/labkey/study/query/DatasetTableImpl.java
@@ -469,7 +469,7 @@ public class DatasetTableImpl extends BaseStudyTable implements DatasetTable
 
         addFolderColumn();
 
-        setSupportMerge(true);
+        setSupportMerge(getDataset().getKeyManagementType() == Dataset.KeyManagementType.None);
     }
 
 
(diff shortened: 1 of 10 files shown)


study/src/org/labkey/study/query/DatasetTableImpl.java
```diff
@@ -468,6 +468,8 @@ public SQLFragment getValueSql(String tableAliasName)
         addColumn(datasetId);
 
         addFolderColumn();
+
+        setSupportMerge(true);
```
labkey-susanh: I'm a little unclear on datasets here. Will they always support merge? It seems like there is some case, somewhat akin to the autoincement keys for lists, where merge cannot be supported.

XingY: dataset uses participant + other columns as key. participant ids are not auto increments. The merge option is enabled for all dataset on UI looking from code.

labkey-susanh: The spec says in the non-requirements functionality that for "Data sets that have integer managed third-key fields, we do not currently support merge".  Not sure where this manifests in the code, though.

XingY: @labkey-ians Ian, I think you implemented merge for dataset. Do you know if certain datasets are excluded from merge function?

XingY: changes made to disable merge for dataset that uses third key (guid or rowid auto int)

labkey-susanh (approved):

PR #3918, opened 2022-12-13
Push inactive users to the bottom of the "impersonate user" drop-down

labkey-adam:
#### Rationale
Show that I can code in ExtJS. And https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46449

### DIFF
--- a/internal/webapp/Impersonate.js
+++ b/internal/webapp/Impersonate.js
@@ -64,7 +64,7 @@ Ext4.define('LABKEY.Security.ImpersonateUser', {
                     '<tpl if="active">',
                         '<div class="x4-boundlist-item">{email:htmlEncode} ({displayName:htmlEncode})</div>',
                     '<tpl else>',
-                        '<div class="x4-boundlist-item x4-item-disabled" style="color: #999999;">{email:htmlEncode} ({displayName:htmlEncode}) (inactive)</div>',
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
Looks OK, but see my comment in the issue about the formatting.

internal/webapp/Impersonate.js
```diff
@@ -64,7 +66,10 @@ Ext4.define('LABKEY.Security.ImpersonateUser', {
                     '<tpl if="active">',
                         '<div class="x4-boundlist-item">{email:htmlEncode} ({displayName:htmlEncode})</div>',
                     '<tpl else>',
-                        '<div class="x4-boundlist-item x4-item-disabled" style="color: #999999;">{email:htmlEncode} ({displayName:htmlEncode}) (inactive)</div>',
+                        '<div ',
+                        '<tpl>data-qtip="Inactive users can\'t be impersonated"</tpl>',
```
cnathe: does this need to be wrapped in `<tpl>`. also, I don't see a closing `</div>` for the line above.

cnathe:
```suggestion
                        'data-qtip="Inactive users can\'t be impersonated"',

```

cnathe (approved):

PR #3919, opened 2022-12-13
Don't attempt to include modules that would be excluded

labkey-tchad:
#### Rationale
We are starting to use startup properties on TeamCity to run test servers with different module sets; generating a list of modules from a specified distribution. Some distributions include modules that aren't tested on TeamCity but we might want to test on that distribution otherwise (e.g. `AssayPortal` in the `panoramaweb` distribution). This change allows us to ignore these missing modules if they are explicitly excluded via the `ModuleLoader.exclude` startup property (otherwise we get an NPE when `ModuleLoader` tries to get the name of the non-existent module).
```
java.lang.NullPointerException: Cannot invoke "org.labkey.api.module.Module.getName()" because "m" is null
	at org.labkey.api.module.ModuleLoader.loadModules(ModuleLoader.java:951) ~[api-23.1-SNAPSHOT.jar:?]
	at org.labkey.api.module.ModuleLoader.doInitWithSourceModule(ModuleLoader.java:291) ~[api-23.1-SNAPSHOT.jar:?]
	at org.labkey.api.module.ModuleLoader.doInit(ModuleLoader.java:483) ~[api-23.1-SNAPSHOT.jar:?]
	at org.labkey.api.module.ModuleLoader.init(ModuleLoader.java:244) ~[api-23.1-SNAPSHOT.jar:?]
```

This change also makes `ModuleLoader.loadModules` log an error/warning in this scenario, instead of throwing the NPE.

#### Related Pull Requests
* https://github.com/LabKey/gradlePlugin/pull/159

#### Changes
* Don't attempt to include modules that would be excluded
* Log a warning/error if 'ModuleLoader.include' references missing modules
* Make `ModuleLoaderStartupProperties.getList()` return a copy of the module list

### DIFF
--- a/api/src/org/labkey/api/module/ModuleLoader.java
+++ b/api/src/org/labkey/api/module/ModuleLoader.java
@@ -18,7 +18,6 @@ package org.labkey.api.module;
 import org.apache.commons.beanutils.BeanUtils;
 import org.apache.commons.collections4.MultiValuedMap;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -938,8 +937,7 @@ public class ModuleLoader implements Filter, MemTrackerListener
 
         // filter by startup properties if they were specified
         LinkedList<String> includeList = ModuleLoaderStartupProperties.include.getList();
-        Set<String> excludeList = Collections.newSetFromMap(new CaseInsensitiveHashMap<>());
-        excludeList.addAll(ModuleLoaderStartupProperties.exclude.getList());
+        Set<String> excludeSet = Sets.newCaseInsensitiveHashSet(ModuleLoaderStartupProperties.exclude.getList());
 
         List<String> missingModules = new ArrayList<>();
         CaseInsensitiveTreeMap<Module> includedModules = moduleNameToModule;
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -937,22 +938,42 @@ else if (!moduleNamePattern.matcher(module.getName()).matches())
...
+                if (!excludeList.contains(moduleName)) // Don't look up excluded modules or include their dependencies
+                {
+                    Module m = moduleNameToModule.get(moduleName);
+                    if (m == null)
+                    {
+                        missingModules.add(moduleName);
+                    }
+                    else
+                    {
+                        // add module to includedModules, add dependencies to includeList (of course it's too soon to call getResolvedModuleDependencies) */
```
```suggestion
                        // add module to includedModules, add dependencies to includeList (of course it's too soon to call getResolvedModuleDependencies)

```

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -937,22 +938,42 @@ else if (!moduleNamePattern.matcher(module.getName()).matches())
 
         // filter by startup properties if they were specified
         LinkedList<String> includeList = ModuleLoaderStartupProperties.include.getList();
-        LinkedList<String> excludeList = ModuleLoaderStartupProperties.exclude.getList();
+        Set<String> excludeList = Collections.newSetFromMap(new CaseInsensitiveHashMap<>());
```
```suggestion
        Set<String> excludeSet = Sets.newCaseInsensitiveHashSet();

```

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -937,22 +938,42 @@ else if (!moduleNamePattern.matcher(module.getName()).matches())
...
+                        if (null == includedModules.put(m.getName(), m))
+                            includeList.addAll(m.getModuleDependenciesAsSet());
+                    }
+                }
             }
         }
 
+        if (!missingModules.isEmpty())
+        {
+            _log.log(AppProps.getInstance().isDevMode() ? Level.WARN : Level.ERROR,
```
labkey-adam: I recently dialed down our handling of unknown startup properties, from throwing to INFO log. We'll see what happens on our cloud deployments; ERROR might be too strict.

labkey-adam (approved):
A few simple requests, but no need for re-review

PR #3922, opened 2022-12-13
SQL Server: improve support for unicode characters

labkey-nicka:
#### Rationale
This adds support for unicode characters for SQL Server by prefixing string literals `N'string'`.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3890

#### Changes
* Override `quoteStringLiteral()` with new implementation `MicrosoftSqlServerStringHandler` to prefix string literals.
* Utilize `dialect.getStringHandler().quoteStringLiteral()` to process "sourceKey" filter parameter in experiment lineage.
* Remove use of `SQLFragment` in "sourceKey" processing.

### DIFF
--- a/bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServerStringHandler.java
+++ b/bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServerStringHandler.java
@@ -1,0 +1,13 @@
+package org.labkey.bigiron.mssql;
+
+import org.labkey.api.data.dialect.StandardDialectStringHandler;
+
+public class MicrosoftSqlServerStringHandler extends StandardDialectStringHandler
+{
+    @Override
+    protected String stringValue(String value)
+    {
+        // Prefix string literals with N to force Unicode
+        return "N" + quoteStringLiteral(value);
+    }
+}
--- a/bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServerStringHandler.java
+++ b/bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServerStringHandler.java
@@ -5,9 +5,9 @@ import org.labkey.api.data.dialect.StandardDialectStringHandler;
 public class MicrosoftSqlServerStringHandler extends StandardDialectStringHandler
 {
     @Override
-    protected String stringValue(String value)
+    public String quoteStringLiteral(String value)
     {
         // Prefix string literals with N to force Unicode
-        return "N" + quoteStringLiteral(value);
+        return "N" + super.quoteStringLiteral(value);
     }
 }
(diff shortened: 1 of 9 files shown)


bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServerStringHandler.java
```diff
@@ -0,0 +1,13 @@
+package org.labkey.bigiron.mssql;
+
+import org.labkey.api.data.dialect.StandardDialectStringHandler;
+
+public class MicrosoftSqlServerStringHandler extends StandardDialectStringHandler
+{
+    @Override
+    protected String stringValue(String value)
```
labkey-matthewb: Why not override quoteStringLiteral() and use super.quoteStringLiteral()?  This seems round about.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -2564,14 +2565,14 @@ private Pair<String,String> getRunGraphCommonTableExpressions(SQLFragment ret, S
             {
                 select = select.replace("$LSIDS$", lsidsFrag.getRawSQL());
                 if (options.getSourceKeySQL() != null)
-                    select = select.replace("$SOURCEKEY$", "'" + options.getSourceKeySQL().getParams().get(0) + "'");
+                    select = select.replace("$SOURCEKEY$", dialect.getStringHandler().substituteParameters(options.getSourceKeySQL()));
```
labkey-matthewb: This should just be dialect.getStringHandler().quoteStringLiteral(optinos.getSourceKey())
This is very circuitous...
If we need to generalize this in the future (e.g. to support IN or other expressions ) then this whole thing will need to be reworked.  We use substituteParameters() for debug logging and the like, I don't think we use it for production SQL.

labkey-matthewb (changes requested):

bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServerStringHandler.java
labkey-nicka: Agreed. Much better. Fixed.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-nicka: > This should just be dialect.getStringHandler().quoteStringLiteral(optinos.getSourceKey()) This is very circuitous... 

Agreed. Fixed.

> If we need to generalize this in the future (e.g. to support IN or other expressions ) then this whole thing will need to be reworked. We use substituteParameters() for debug logging and the like, I don't think we use it for production SQL.

It is not apparent at all from seeing this method that it is intended for debug logging and the like. If it is not intended for production then we should at minimum make a note and at most make it throw an exception in production environments.

labkey-matthewb (approved):

PR #3924, opened 2022-12-14
Fix link reports

labkey-adam:
#### Rationale
Viewing a link report throws an exception instead of redirecting. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46942


labkey-jeckels (approved):
Approved, from IntelliJ!

PR #3927, opened 2022-12-14
Issue 46939: Naming Patterns for Mixture Batches Not Working in SubProjects

XingY:
#### Rationale
NameGenerator uses parentTable tableInfo for getting lookup column's LookupTableInfo. In a multi projects setup, we allow looking up to self + home.  This PR would allow name generator to also use lookup values from self + home.

#### Related Pull Requests
* 

#### Changes
* use CF for the samples and dataclass TableInfo that's used by NameGenerator

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
@@ -866,7 +866,15 @@ public class ExpDataClassDataTableImpl extends ExpRunItemTableImpl<ExpDataClassD
 //              TODO check if this covers all the functionality, in particular how is alternateKeyCandidates used?
                 di = LoggingDataIterator.wrap(new CoerceDataIterator(di, context, ExpDataClassDataTableImpl.this, false));
 
-                di = LoggingDataIterator.wrap(new NameExpressionDataIterator(di, context, ExpDataClassDataTableImpl.this, getContainer(), _dataClass.getMaxDataCounterFunction(), DATA_COUNTER_SEQ_PREFIX + _dataClass.getRowId() + "-")
+                TableInfo dataClassTInfo = ExpDataClassDataTableImpl.this;
+                if (c.hasProductProjects() && !c.isProject())
+                {
+                    User user = getUserSchema().getUser();
+                    ContainerFilter cf = new ContainerFilter.CurrentPlusProjectAndShared(c, user); // use lookup CF
+                    dataClassTInfo = QueryService.get().getUserSchema(user, c, "exp.data").getTable(getName(), cf);
+                }
+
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -866,7 +866,15 @@ public DataIterator getDataIterator(DataIteratorContext context)
 //              TODO check if this covers all the functionality, in particular how is alternateKeyCandidates used?
                 di = LoggingDataIterator.wrap(new CoerceDataIterator(di, context, ExpDataClassDataTableImpl.this, false));
 
-                di = LoggingDataIterator.wrap(new NameExpressionDataIterator(di, context, ExpDataClassDataTableImpl.this, getContainer(), _dataClass.getMaxDataCounterFunction(), DATA_COUNTER_SEQ_PREFIX + _dataClass.getRowId() + "-")
+                TableInfo dataClassTInfo = ExpDataClassDataTableImpl.this;
+                if (c.hasProductProjects() && !c.isProject())
```
labkey-nicka, line 910: I would put a comment referring to this issue and speak to why this is being done. Same for the other snippet for samples.

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -408,9 +409,16 @@ private NameGenerator createNameGenerator(@NotNull String expr, Container dataCo
         }
 
         Container sampleTypeContainer = getContainer();
-        TableInfo parentTable = QueryService.get().getUserSchema(User.getSearchUser(), sampleTypeContainer, SamplesSchema.SCHEMA_NAME).getTable(getName());
-
         Container nameGenContainer = dataContainer != null ? dataContainer : sampleTypeContainer;
+
+        User user = User.getSearchUser();
```
labkey-nicka, line 414: I see that `User.getSearchUser()` is used throughout this class to refer to the user, however, I'm a bit concerned this isn't actually scoped correctly (i.e. will not return the same results as the current user for the generated container filter). Perhaps `getTinfo().getUserSchema().getUser()` with proper  defensive checks?

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -408,9 +409,16 @@ private NameGenerator createNameGenerator(@NotNull String expr, Container dataCo
...
 
         Container sampleTypeContainer = getContainer();
-        TableInfo parentTable = QueryService.get().getUserSchema(User.getSearchUser(), sampleTypeContainer, SamplesSchema.SCHEMA_NAME).getTable(getName());
-
         Container nameGenContainer = dataContainer != null ? dataContainer : sampleTypeContainer;
+
+        User user = User.getSearchUser();
+        ContainerFilter cf = null;
+
+        if (dataContainer != null && dataContainer.hasProductProjects())
```
labkey-nicka, line 417: Just curious, why does this check differ from the seemingly equivalent check in Data Classes?

labkey-nicka (approved):

PR #3930, opened 2022-12-15
Simplify HttpUtil and ImageUtil

labkey-adam:
#### Rationale
Upgrading HttpClient 4.x -> 5.2. Also taking the opportunity to simplify and eliminate unused code in HttpUtil and ImageUtil.

#### Changes
* Upgrade to HttpClient 5.2
* Rename `HttpUtil.get()` to `getText()` and return contents as a String. This is what the one caller wants and it allows us to close resources properly.
* Switch to new fancy ResponseHandler


labkey-adam:
@labkey-matthewb Let me know if you think I've removed something useful

labkey-matthewb (approved):

PR #3940, opened 2022-12-20
Issue 46857: Don't use ExistingRecordDataIterator for trigger helper for biologics eval data loading

XingY:
#### Rationale
The eval data loading uses Folder Import, which uses merge to load those sample and dataclass data. This is problematic in 2 ways: folder import is not officially supported for Biologics and merge is not supported for some registries. The changes made in TriggerDataBuilderHelper to use ExistingRecordDataIterator is not working for the eval data due to the state of data and is causing test failure: https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_BiologicsPostgres95/2180057?buildTab=overview
I believe the eval data needs to be updated to add those missed columns. For now, we can treat eval data loading as a NEW load that writes into a clean folder, and allow the continued use of CoerceDataIterator.outputAllColumns for this type of import to avoid the test failure.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3940
* https://github.com/LabKey/biologics/pull/1803
* https://github.com/LabKey/trialServices/pull/35

* https://github.com/LabKey/platform/pull/3901

#### Changes
* added isNewFolderImport importOption to allow using CoerceDataIterator.outputAllColumns for TriggerDataBuilderHelper for eval data loading
* Skipping adding ExistingRecordIterators in TriggerDataBuilderHelper for registries that don't support merge

### DIFF
--- a/api/src/org/labkey/api/admin/FolderImportContext.java
+++ b/api/src/org/labkey/api/admin/FolderImportContext.java
@@ -25,7 +25,6 @@ import org.labkey.api.module.FolderTypeManager;
 import org.labkey.api.reports.report.ReportDescriptor;
 import org.labkey.api.reports.report.view.ReportUtil;
 import org.labkey.api.security.User;
-import org.labkey.api.util.GUID;
 import org.labkey.api.util.XmlBeansUtil;
 import org.labkey.api.util.XmlValidationException;
 import org.labkey.api.writer.VirtualFile;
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/admin/FolderImportContext.java
```diff
@@ -56,6 +55,10 @@ public class FolderImportContext extends AbstractFolderContext
 
     private static final String FOLDER_IMPORT_DB_SEQUENCE_PREFIX = "FolderImportJobCounter-";
 
+    private boolean _isNewFolderImport; // if we know the target folder is empty, can skip certain merge logic
```
labkey-susanh, line 58: Maybe this should be called "mergeData" instead?  For Biologics in particular, the folder that is being imported to is not entirely "new" since the creation of the Biologics folder results in the creation of of certain lists and a few default Protein Sequences.

labkey-susanh (approved):
Just one comment on the naming here, but no need for re-review.  It's a bit of a hack, but I don't have a better suggestion at this point.

api/src/org/labkey/api/admin/FolderImportContext.java
XingY: "Merge" is what I originally had. With the changes, the folder import is still doing a Merge, just part of the trigger handling checks for _isNewFolderImport. I chose not to use merge because InsertOption.merge means a different thing.

PR #3948, opened 2022-12-23
Restore mock ViewContext to fix background triggers needing user and URL

labkey-jeckels:
#### Rationale
Additional tests last night turned up a problem with the PR I merged that correctly tracks the container for a given TableInfo trigger event even if it's a different container from the originating ViewContext. Ideally we'd not rely on the ViewContext at all and may be able to revisit, but need to get these scenarios working in the meantime. They only rely on the User and URL, so they aren't varying from operation to operation at this time.

https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Premium_Ehr_DailyEhrPostgres/2185329?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildTestsSection=true&expandBuildChangesSection=true

https://github.com/LabKey/platform/pull/3942

#### Changes
* Push a mock view context when we don't have one already

### DIFF
--- a/api/src/org/labkey/api/data/triggers/ScriptTrigger.java
+++ b/api/src/org/labkey/api/data/triggers/ScriptTrigger.java
@@ -235,6 +235,13 @@ public class ScriptTrigger implements Trigger
 
     private <T> T _try(Container c, User user, Map<String, Object> extraContext, ScriptFn<T> fn)
     {
+        ViewContext.StackResetter viewContextResetter = null;
+        if (!HttpView.hasCurrentView())
+        {
+            // Push a view context if we don't already have one available. It may be pulled to get the user
+            // in ext4Bridge.js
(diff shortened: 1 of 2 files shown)


labkey-jeckels:
@bbimber FYI

bbimber:
thanks @labkey-jeckels. yes, I optimistically removed that MockViewContext, but I guess this needs it.  This is a background pipeline job that caused this, right? In any event, a MockViewContext has been injected by LK code for a long time, so it's probably not that much of an issue. It might be possible to side-step whatever code is trying to grab the URL, which I'd be happy to look at next week, but I'm not sure if's that big a deal to have this.

labkey-klum (approved):

labkey-jeckels:
@bbimber I was able to handle the URL part without the mock ViewContext but passing the User around was going to be a lot harder.

PR #3949, opened 2022-12-23
Handle update and clarify merge - samples and dataclasses

XingY:
#### Rationale
Merge has been misrepresented as Update in the product UI for ease of understanding, which causes confusion when users are expecting an update action. This PR includes the api/server side work to support Update from file for dataclasses and samples. 

Note that this PR only add limited supports for samples and dataclass updates. Dataset and lists update support will be completed in a follow up pr. Advanced features such as file watcher, storage data import, indexer will also be in a follow up story.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1065
* https://github.com/LabKey/platform/pull/3949
* https://github.com/LabKey/labbook/pull/355
* https://github.com/LabKey/sampleManagement/pull/1493
* https://github.com/LabKey/biologics/pull/1817

#### Changes
* modified ExistingDataRecordIterator to be able to check if all provided are existing data for UPDATE action, as well as checking for Cross folder items for UPDATE and MERGE
* added SampleUpdateAliquotedFromDataIterator to fetch AliquotedFromLSID for samples during UPDATE so that AliquotedFrom don't need to be provided by users
* added ImportWithUpdateDerivationDataIterator to handle input/outputs during UPDATE
* don't empty out columns that are absent during UPDATE
* don't check for required fields for UPDATE
* use Update trigger type for UPDATE


api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -175,7 +175,7 @@ public static DataIteratorBuilder createBuilder(DataIteratorBuilder dib, TableIn
             if (di.supportsGetExistingRecord())
                 return di;
             QueryUpdateService.InsertOption option = context.getInsertOption();
-            if (option.mergeRows)
+            if (option.mergeRows || option.updateOnly)
```
labkey-matthewb: I think it's good to encapsulate the sematics of checks like this into the enum itself.  E.g. option.mayUpdateRows()

api/src/org/labkey/api/dataiterator/NoNewRecordValidationDataIterator.java
```diff
@@ -0,0 +1,51 @@
...
+    {
+        super(in, context, target, keys, extraKeyValueMap, batchSize);
+    }
+
+    @Override
+    public void validate(Map<Integer, Map<String, Object>> keysMap)
+    {
+        try
+        {
+            qus.verifyExistingRows(user, c, new ArrayList<>(keysMap.values()));
```
labkey-matthewb: This seems very expensive.  Like the code we removed from detailed audit logging.  Why not use ExistingRecordDataIterator and getExistingRecord()?

api/src/org/labkey/api/data/StatementUtils.java
```diff
@@ -847,8 +847,17 @@ else if (column.getName().equalsIgnoreCase(updatable.getObjectURIColumnName()) &
                 sqlfUpdate.append(values.get(i));
                 updateCount++;
             }
-            sqlfUpdate.append(sqlfWherePK);
-            sqlfUpdate.append(";\n");
+
+            if (Operation.update == _operation && updateCount == 0)
+            {
+                sqlfUpdate.append(new SQLFragment(keys.values().iterator().next().getSelectName()));
```
labkey-matthewb, line 871: Maybe we should throw here?  The caller could be given the chance to skip this iterator or rethrow a meaninful error.

labkey-matthewb, line 871: I don't know the vocabulary stuff very well, but will this allow an update to add a vocabulary property?

api/src/org/labkey/api/dataiterator/RecordValidationDataIterator.java
```diff
@@ -0,0 +1,107 @@
...
+            for (int p=0 ; p<pkColumns.size() ; p++)
+                keyMap.put(pkColumns.get(p).getColumnName(), pkSuppliers.get(p).get());
+            if (_extraKeyValueMap != null)
+                keyMap.putAll(_extraKeyValueMap);
+            keysMap.put(lastPrefetchRowNumber, keyMap);
+        }
+        while (--rowsToFetch > 0 && _delegate.next());
+
+        validate(keysMap);
+        if (qus.hasExistingRowsInOtherContainers(c, keysMap))
```
labkey-matthewb: Again this is very expensive.  Are we doing all this work just to report "there is a record in the another container" instead of  "record is not found".  It seems like we should only  select existing records in the current container, and when/if we find a missing record _THEN_ choose whether we want to do the work to distinguish "not found" from "found in another container"?  Maybe I'm missing something.

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -61,6 +72,13 @@ public abstract class ExistingRecordDataIterator extends WrapperDataIterator
         this.existingColIndex = in.getColumnCount()+1;
         this.useMark = useMark;
 
+        UserSchema userSchema = target.getUserSchema();
+        user = userSchema != null ? userSchema.getUser() : null;
+        c = userSchema != null ? userSchema.getContainer() : null;
+        boolean isExpDataOrMaterial = target instanceof ExpMaterialTable || target instanceof ExpDataClassDataTable;
+        _checkCrossFolderData = isExpDataOrMaterial && (option.updateOnly || option.mergeRows);
+        _verifyExisting = option.updateOnly;
```
labkey-matthewb, line 83: This really breaks encapsulation (ExistingRecordDataIterator knowing about the implementation requirements of particualar tables), also instanceof here is fragile.  These flags should be either passed into createBuilder() or maybe stashed as properties in the DataIteratorContext.

labkey-matthewb:
I may be wrong, but the only place I see that we can make use of ImportOption.UPDATE is via AbstractQueryImportAction (and subclasses).  Is there a plan for API support?  We can do that as part of another story, but it would be odd to support in the UI but not in the API.

labkey-matthewb:
It would be great to have a junit test for InsertOption.UPDATE generically, and the SM special cases.

api/src/org/labkey/api/dataiterator/TriggerDataBuilderHelper.java
```diff
@@ -132,13 +142,13 @@ public DataIterator getDataIterator(DataIteratorContext context)
...
 
-            boolean includeAllColumns = (!context.getInsertOption().mergeRows || mergeKeys == null) || isNewFolderImport;
+            boolean includeAllColumns = ((!context.getInsertOption().mergeRows && !context.getInsertOption().updateOnly) || mergeKeys == null) || isNewFolderImport;
             DataIterator coerce = new CoerceDataIterator(pre, context, _target, includeAllColumns);
             coerce = LoggingDataIterator.wrap(coerce);
 
             if (includeAllColumns)
                 return LoggingDataIterator.wrap(new BeforeIterator(new CachingDataIterator(coerce), context));
-            else if (!_target.supportMerge())
+            else if (context.getInsertOption().mergeRows && !_target.supportMerge())
```
labkey-matthewb: This really needs to be updated to supportsInsertOption(InsertOption option).  See https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42788

api/src/org/labkey/api/query/QueryService.java
```diff
@@ -440,6 +440,15 @@ public String getVerbPastTense()
...
         }
+
+        public static AuditAction getImportAuditAction(QueryUpdateService.InsertOption insertOption)
+        {
+            if (insertOption == QueryUpdateService.InsertOption.UPDATE)
+                return UPDATE;
+            if (insertOption == QueryUpdateService.InsertOption.MERGE)
+                return MERGE;
+            return INSERT;
+        }
```
labkey-matthewb: I think it's better to either a) add getAuditAction() to the enum (preferred) or b) enumerate all the values and throw in the default case (see See pattern in AbstractQueryImportAction._execute()) .  Otherwise, we things can get missed.

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -30,26 +30,7 @@
...
-import org.labkey.api.data.SimpleFilter;
-import org.labkey.api.data.Sort;
-import org.labkey.api.data.SqlSelector;
-import org.labkey.api.data.Table;
-import org.labkey.api.data.TableInfo;
-import org.labkey.api.data.TableSelector;
-import org.labkey.api.data.UnionContainerFilter;
-import org.labkey.api.data.UpdateableTableInfo;
+import org.labkey.api.data.*;
 import org.labkey.api.data.dialect.SqlDialect;
```
labkey-matthewb: comment?

labkey-matthewb (changes requested):

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
XingY: I initially had it passed in to createBuilder, but the case that would be broken is TriggerDataBuilderHelper, as it doesn't know the params to pass in. I can look into having them on DataIteratorContext

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -1299,5 +1319,39 @@ public boolean hasExistingRowsInOtherContainers(Container container, Map<Integer
...
+            SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("ClassId"), dataClassId);
+            filter.addCondition(FieldKey.fromParts("Name"), dataNames, CompareType.IN);
+            filter.addCondition(FieldKey.fromParts("Container"), container);
+
+            List<String> foundNames = new TableSelector(ExperimentService.get().getTinfoData(), Collections.singleton("name"), filter, null).getArrayList(String.class);
+            if (dataNames.size() > foundNames.size())
+            {
+                foundNames.forEach(dataNames::remove);
+                String absentName = (new ArrayList<>(dataNames)).get(0);
+                throw new QueryUpdateServiceException("Row not found for " + absentName);
```
labkey-susanh: Will there always be only one? If not, any reason not to report on all missing names?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -681,14 +684,43 @@ public boolean hasExistingRowsInOtherContainers(Container container, Map<Integer
...
+            if (name != null)
+                sampleNames.add(name);
+
+            if (sampleTypeId == null)
+                sampleTypeId = getMaterialSourceId(keyMap);
+        }
+
+        SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("MaterialSourceId"), sampleTypeId);
+        filter.addCondition(FieldKey.fromParts("Name"), sampleNames, CompareType.IN);
+        filter.addCondition(FieldKey.fromParts("Container"), container);
```
labkey-susanh: I'm a little unclear on this, but if the Name exists but from a different container (e.g., it's in the parent container), are we meant to allow that update?  It seems like that is being handled elsewhere, which is probably as it should be since the error message is more specific.  Maybe it's worth a comment here?

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
XingY: done

api/src/org/labkey/api/dataiterator/NoNewRecordValidationDataIterator.java
XingY: I've got rid of this iterator and switched to use ExistingRecordDataIterator

api/src/org/labkey/api/data/StatementUtils.java
XingY: done

api/src/org/labkey/api/dataiterator/RecordValidationDataIterator.java
XingY: I've got rid of this iterator in favor of using ExistingRecordDataIterator.

We don't want to update/merge a record that exist in a different folder. Apart from user expectation, using update statement on tables without checking their folder context can be dangerous. Because samples and data consist of 2 tables, exp.material/data and the provisioned. While exp.material/data has a container column, their provisioned table do not. An update on cross folder item would skip exp table, but would update the provisioned.

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
XingY: changes made to stashe in the DataIteratorContext.

api/src/org/labkey/api/dataiterator/TriggerDataBuilderHelper.java
XingY: done

api/src/org/labkey/api/query/QueryService.java
XingY: done

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
XingY: This function is now removed as part of the refactoring to use ExistingRecordDataIterator for checking cross folder / existence of data.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: This function is now removed as part of the refactoring to use ExistingRecordDataIterator for checking cross folder / existence of data.

api/src/org/labkey/api/data/StatementUtils.java
XingY: Good reminder. I've added vocabulary properties handling as part of the follow up story.

XingY:
> It would be great to have a junit test for InsertOption.UPDATE generically, and the SM special cases.

Junit tests added for samples, dataclass and list:
https://github.com/LabKey/platform/pull/3949/commits/8ca7b14fa949073260b0c4d34a310d6fc5fea5b0

For LKSM use cases, I did add junit test that uses detailed logging. The other piece that I'm not able to simulate is the "Projects" feature (the check for cross folder data), since it's based on site configuration. I'll defer that to integration test.

XingY:
> I may be wrong, but the only place I see that we can make use of ImportOption.UPDATE is via AbstractQueryImportAction (and subclasses). Is there a plan for API support? We can do that as part of another story, but it would be odd to support in the UI but not in the API.

By API, do you mean the updateRows api? Or some import api? If the former, there is a follow up story to use data iterators to do bulk updates.

experiment/src/org/labkey/experiment/api/ExpSampleTypeTestCase.jsp
```diff
@@ -1087,4 +1087,122 @@ public void testExpMaterialPermissions() throws Exception
...
+        qus.loadRows(user, c, new ListofMapsDataIterator(rowsToUpdate.get(0).keySet(), rowsToUpdate), context, null);
+    }
+    catch (Exception e)
+    {
+        hasError = true;
+        assertTrue(e.getMessage().contains("Sample does not exist: S-1-absent."));
+    }
+    assertTrue(hasError);
+
+    // AliquotedFrom is supplied but doesn't match the current aliquot status / parents should get ignored
```
labkey-susanh, line 1191: Ignored? Wouldn't this be an error? Or is AliquotedFrom always ignored now for update.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -807,10 +856,19 @@ public List<Map<String, Object>> getRows(User user, Container container, List<Ma
             Map<String, Object> materialMap = getMaterialMap(getMaterialRowId(k), getMaterialLsid(k), user, container, addInputs);
             if (materialMap != null)
                 result.add(materialMap);
+            else if (failOnAbsent)
+                throw new QueryUpdateServiceException("TODO");
```
labkey-susanh: TODO.

study/src/org/labkey/study/query/DatasetTableImpl.java
```diff
@@ -469,7 +469,8 @@ public SQLFragment getValueSql(String tableAliasName)
 
         addFolderColumn();
 
-        setSupportMerge(getDataset().getKeyManagementType() == Dataset.KeyManagementType.None);
+        if(getDataset().getKeyManagementType() == Dataset.KeyManagementType.None)
+            removeDisallowedInsertOption(QueryUpdateService.InsertOption.MERGE);
```
labkey-susanh: The double-negative aspect of this `removeDisallowed` gives me pause. I might have chosen to setDisallowed or setAllowed, but that's mainly a style thing.

labkey-susanh (commented):
This seems fine to me, but I'll defer to Matt for final review.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: actually this is no longer needed. removed.

study/src/org/labkey/study/query/DatasetTableImpl.java
XingY: renamed

experiment/src/org/labkey/experiment/api/ExpSampleTypeTestCase.jsp
XingY: I'm torn between error out and ignore as well... I've made the UI update to let user know that AliquotedFrom cannot be updated during file preview step. We can iterate on this behavior in the future.

api/src/org/labkey/api/data/AbstractTableInfo.java
```diff
@@ -1966,15 +1966,20 @@ public boolean hasDeleteURLOverride()
...
+    }
+
+    public boolean setAllowedInsertOption(QueryUpdateService.InsertOption option)
+    {
+        return _disallowedInsertOptions.remove(option);
     }
 
     @Override
-    public boolean supportMerge()
+    public boolean supportInsertOption(QueryUpdateService.InsertOption option)
```
labkey-matthewb: I think SimpleTable supports MERGE by default (this is sometimes a problem).
We could leave it this way by default, and try to track down the cases where we actually _want_ MERGE to work, but there might be a lot of usages.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -1162,6 +1182,41 @@ public void addRowError(ValidationException vex)
             assertNull(rows.get(3).get("i"));
         }
 
+        @Test
```
labkey-matthewb, line 1185: Maybe we could add this to  _insertRowsUsingDIB?  Might cause a bunch of errors initially, but would force consistency.

    if (!_queryTable.supportsInsertOption(context.getInsertOption()))
        throw new IllegalStateException();

api/src/org/labkey/api/data/TableInfo.java
```diff
@@ -758,8 +758,8 @@ default List<Sort.SortField> getSortFields()
         return List.of();
     }
 
-    default boolean supportMerge()
+    default boolean supportInsertOption(QueryUpdateService.InsertOption option)
     {
-        return false;
+        return true;
```
labkey-matthewb, line 763: Hmm, should probably return false if getUpdateService() is null, but we don't have a cheap way to know that....  Maybe switch default in the future...

labkey-matthewb (approved):

PR #3950, opened 2022-12-23
Prevent schema upgrade problems when modules disappear and reappear

labkey-adam:
#### Rationale
We believe that some (maybe all?) recently highlighted schema upgrade problems were due to modules that were installed, removed, and then re-added to a deployment years later. The server checks if the core module is too old, but doesn't check any other module versions. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46922

#### Changes
* Detect "unknown modules" and encourage their deletion to head off these problems well in advance. This warning appears only when the server is running in production mode.
* Refuse to upgrade if any module has a schema version that's too old. Do the check and bail before core upgrade to prevent partial upgrades.


labkey-jeckels (approved):
I agree that this is will be too noisy on many dev machines. Skipping the warning about modules disappearing outside of production mode sounds like a good approach to me.

labkey-adam:
I'm good with that. Since most of us won't be exercising this code, I think we'd be wise to do a bit of manual testing on the PR as is, and then add the production mode check before merge. Let me know if you're up for testing or I can find someone else. It can also wait until next week.

labkey-jeckels:
@labkey-adam that strategy sounds good and @labkey-tchad volunteered to test in standup today.

labkey-tchad (approved):
Verified warning for unknown modules and that reintroducing a module after it's past the upgrade window causes a startup error.

PR #3953, opened 2022-12-26
Don't use searchuser for sample name generation

XingY:
#### Rationale
This PR address the code review feedbacks for https://github.com/LabKey/platform/pull/3927

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3927
* https://github.com/LabKey/platform/pull/3953
* https://github.com/LabKey/recipe/pull/76

#### Changes
* user context user instead of search user for name generator

### DIFF
--- a/api/src/org/labkey/api/exp/api/ExpSampleType.java
+++ b/api/src/org/labkey/api/exp/api/ExpSampleType.java
@@ -189,6 +189,12 @@ public interface ExpSampleType extends ExpObject
                             @Nullable Set<ExpMaterial> parentSamples,
                             @Nullable Container container) throws ExperimentException;
 
+    String createSampleName(@NotNull Map<String, Object> rowMap,
+                            @Nullable Set<ExpData> parentDatas,
+                            @Nullable Set<ExpMaterial> parentSamples,
+                            @Nullable Container container,
+                            @Nullable User user
+    ) throws ExperimentException;
 
     void setIdCol1(String s);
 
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/exp/api/ExpSampleType.java
```diff
@@ -189,6 +189,12 @@ String createSampleName(@NotNull Map<String, Object> rowMap,
                             @Nullable Set<ExpMaterial> parentSamples,
                             @Nullable Container container) throws ExperimentException;
 
+    String createSampleName(@NotNull Map<String, Object> rowMap,
```
labkey-nicka, line 192: Rather than introducing another signature I'd recommend adding `@Nullable User user` to the above signature.

api/src/org/labkey/api/exp/api/ExpSampleType.java
```diff
@@ -189,6 +189,12 @@ String createSampleName(@NotNull Map<String, Object> rowMap,
                             @Nullable Set<ExpMaterial> parentSamples,
                             @Nullable Container container) throws ExperimentException;
 
+    String createSampleName(@NotNull Map<String, Object> rowMap,
+                            @Nullable Set<ExpData> parentDatas,
+                            @Nullable Set<ExpMaterial> parentSamples,
+                            @Nullable Container container,
+                            @Nullable User user
+    ) throws ExperimentException;
```
labkey-nicka, line 197: Should `void createSampleNames(...)` also be updated to take a `User`? I don't have any usages of that on my local instance so I'm unsure if it is still used at all.

labkey-nicka (approved):

PR #3969, opened 2022-12-30
Issue 46949: Sample timeline detail links aren't shown for subfolder samples when parent is in project level

XingY:
#### Rationale
Issue [46949](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46949): Sample timeline detail links aren't shown for subfolder samples when parent is in project level 
Sample timeline needs to be able to find renamed entity with its legacy name at a specific time point. The utils for finding those entities currently don't work in cross folder scenarios. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1070
* https://github.com/LabKey/platform/pull/3969
* https://github.com/LabKey/sampleManagement/pull/1505
* https://github.com/LabKey/inventory/pull/668
* https://github.com/LabKey/biologics/pull/1829

#### Changes
* support containerFilter for getEffectiveSampleType, getEffectiveSample, getEffectiveDataClass and getEffectiveData utils


experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -675,10 +682,23 @@ public ExpMaterialImpl getSample(Container c, String name)
...
+            if (containerIds != null && containerIds.size() > 1)
+                filter = new SimpleFilter(FieldKey.fromParts("Container"), containerIds, CompareType.IN);
+        }
+
+        return filter;
+    }
+
+    private ExpMaterialImpl getSampleByObjectId(Integer objectId)
+    {
+        SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("ObjectId"), objectId);
```
labkey-susanh, line 688: I suppose the container filter is not really needed here since the objectId is globally unique.

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -675,10 +682,23 @@ public ExpMaterialImpl getSample(Container c, String name)
         return new ExpMaterialImpl(material);
     }
 
-    private ExpMaterialImpl getSampleByObjectId(Container c, Integer objectId)
+    private SimpleFilter getContainerFilterClause(Container c, @Nullable ContainerFilter cf)
```
labkey-susanh: This seems like it might belong with SimpleFilter, and I'm surprised there isn't such a think already.

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -675,10 +682,23 @@ public ExpMaterialImpl getSample(Container c, String name)
         return new ExpMaterialImpl(material);
     }
 
-    private ExpMaterialImpl getSampleByObjectId(Container c, Integer objectId)
+    private SimpleFilter getContainerFilterClause(Container c, @Nullable ContainerFilter cf)
     {
         SimpleFilter filter = SimpleFilter.createContainerFilter(c);
-        filter.addCondition(FieldKey.fromParts("ObjectId"), objectId);
+
+        if (cf != null)
```
labkey-nicka: Rather than composing the filter here I'd recommend using `ContainerFilter.createFilterClause()`. The main reason is that all container filters may not operate in the same way so we don't want to presume an IN clause, etc. The usage would be something like this:

```java
public ExpMaterialImpl getSample(Container c, String name, @Nullable ContainerFilter cf)
{
    SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("CpasType"), getLSID());
    filter.addCondition(FieldKey.fromParts("Name"), name);
    if (cf != null)
        filter.addCondition(cf.createFilterClause(ExperimentServiceImpl.get().getExpSchema(), FieldKey.fromParts("Container")));

    Material material = new TableSelector(ExperimentServiceImpl.get().getTinfoMaterial(), filter, null).getObject(Material.class);
    if (material == null)
        return null;
    return new ExpMaterialImpl(material);
}
```

And then you can just delete this entire `SimpleFilter getContainerFilterClause(Container c, @Nullable ContainerFilter cf)` method.

api/src/org/labkey/api/data/Container.java
```diff
@@ -1692,6 +1693,19 @@ public boolean hasProductProjects()
         return project.getChildren().stream().anyMatch(c -> c.getContainerType().isInFolderNav());
     }
 
+    public ContainerFilter getProjectDataContainerFilter(User user)
```
labkey-nicka: Please rectify this with `BiologicsManager.getContainerFilter()`. I would also recommend using the `create()` pattern.

```java
@Nullable
public static ContainerFilter getContainerFilter(@NotNull Container c, User user)
{
    if (!c.isProductProjectsEnabled() || c.isRoot() || ContainerManager.getSharedContainer().equals(c))
        return null;

    if (c.isProject())
        return ContainerFilter.Type.CurrentAndSubfoldersPlusShared.create(c, user);
    return ContainerFilter.Type.CurrentPlusProjectAndShared.create(c, user);
}
```

labkey-nicka (changes requested):

api/src/org/labkey/api/data/Container.java
labkey-nicka: Also, consider renaming to `getProductProjectsDataContainerFilter` or `getProductProjectsContainerFilter`.

XingY: done

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
XingY: done

labkey-nicka (approved):

PR #3971, opened 2023-01-02
 FR 44763 Access mean and CV of any metric cross last N runs

ankurjuneja:
#### Rationale
This PR provides support for the calculations of trailing means and trailing cv for the last N runs of Panorama quality control.

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/651

#### Changes
* Trailing mean and Trailing CV calculations
* changes in plotting api to support new plots


api/src/org/labkey/api/visualization/Stats.java
```diff
@@ -156,4 +156,52 @@ public static Double[] getMovingRanges(Double[] values, boolean forcePositiveRes
         }
         return mR;
     }
+
+    public static Double[] getTrailingMeans(Double[] values, int N)
```
labkey-jeckels, line 167: A little doc here would be helpful. Particularly, capturing that the length of the result will be values.length - N will help people understand the intended usage

api/src/org/labkey/api/visualization/Stats.java
```diff
@@ -156,4 +156,52 @@ public static Double[] getMovingRanges(Double[] values, boolean forcePositiveRes
...
+        int numOfTrailingValues = values.length - N + 1;
+        Double[] trailingCVs = new Double[numOfTrailingValues];
+        int start = 0;
+        int end = N;
+        for (int i = 0; i < numOfTrailingValues; i++)
+        {
+            Double[] vals = getValuesFromRange(values, start, end);
+            double sd = getStdDev(vals, false);
+            double mean = getMean(vals);
+            trailingCVs[i] = sd/mean * 100;
```
labkey-jeckels: Mean could be 0 - should check and use null as the CV

labkey-jeckels (approved):

api/src/org/labkey/api/visualization/Stats.java
labkey-jeckels: @ankurjuneja divide by zero is still possible here. Please correct.

labkey-jeckels (approved):
See concern about divide by zero, but otherwise looks good

PR #3976, opened 2023-01-04
JSONObject migration in search module

labkey-adam:
#### Rationale
Use the latest

### DIFF
--- a/search/src/org/labkey/search/view/search.jsp
+++ b/search/src/org/labkey/search/view/search.jsp
@@ -17,8 +17,8 @@
 %>
 <%@ page import="org.apache.commons.lang3.StringUtils" %>
 <%@ page import="org.jetbrains.annotations.NotNull" %>
-<%@ page import="org.json.old.JSONArray" %>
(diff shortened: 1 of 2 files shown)


search/src/org/labkey/search/view/search.jsp
```diff
@@ -530,8 +530,8 @@
                 %>
                 <a href="<%=h(previousURL)%>">&lt; Previous</a>
                 <%
-                    if (pageNo < pageCount)
-                    {
+                        if (pageNo < pageCount)
+                        {
```
labkey-susanh, line 534: This spacing change seems very suspect, but I guess you know what you're doing.

labkey-susanh (approved):

labkey-nicka (approved):

search/src/org/labkey/search/view/search.jsp
labkey-adam: This inner `if` wasn't indented nor aligned with its closing brace, which triggered my OCD

PR #3984, opened 2023-01-05
Skip expensive work to improve PIVOT query metadata retrieval

labkey-adam:
#### Rationale
Determining the column list of a PIVOT query can be very expensive, so avoid it, if possible. Back-ported from 23.1; see cherry-picked commits for more details.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -27,8 +27,6 @@ List apache = [
         "org.apache.commons:commons-lang3:${commonsLang3Version}",
         "commons-pool:commons-pool:${commonsPoolVersion}",
         "commons-validator:commons-validator:${commonsValidatorVersion}",
-        "org.apache.httpcomponents:httpclient:${httpclientVersion}",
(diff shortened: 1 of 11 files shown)


api/build.gradle
```diff
@@ -27,8 +27,6 @@ List apache = [
         "org.apache.commons:commons-lang3:${commonsLang3Version}",
         "commons-pool:commons-pool:${commonsPoolVersion}",
         "commons-validator:commons-validator:${commonsValidatorVersion}",
-        "org.apache.httpcomponents:httpclient:${httpclientVersion}",
```
labkey-jeckels, line 30: Is this part of the change intentional?

labkey-adam: Not needed, but it seems to be harmless so I left it in. This just means we can't reference the old HttpClient from JSPs in 22.11. I checked, and we have no JSPs that reference any version of HttpClient.

labkey-jeckels (approved):

PR #3987, opened 2023-01-05
move study design domains

labkey-klum:
#### Rationale
The original intent of the study design tables were to scope those five domains at the project level. In reality this only partly happened, while the domain URI is scoped to the project, the domains are created in the study container, so if the study is not at the project level the domain and property descriptors are scoped to the study folder, see `StudyDesignManager.ensureStudyDesignDomains` which gets called at study creation time.

The referenced bug can occur if the study which owns the domains is deleted and those descriptors and properties get cleaned up by the container listener. If there is a sibling study or in another subfolder, you'll see the NPE. The fix is simple, just ensure the domains in the correct folder, the more complicated part is moving the domains in existing studies.

The upgrade script will move both `DomainDescriptors` and `PropertyDescriptors` if the domain container isn't at the project level. It's unlikely that there will be `PropertyDescriptors` associated with those domains as I don't recall allowing the user to mutate those domains, but since we don't shut it off in the domain kind we still should check.

I don't recall having to do this type of cleanup before, so if either of you can make sure I'm not missing anything or orphaning objects.

[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46986)


labkey-adam:
Probably need to merge from develop to get this building

study/src/org/labkey/study/designer/StudyDesignManager.java
```diff
@@ -703,6 +706,72 @@ public void ensureStudyDesignDomains(Container c, User user)
...
+            log.info("Moving the domain " + domain.getTypeURI() + " to the project level.");
+            moveStudyDomain(domain, studyContainer.getProject());
+        }
+    }
+
+    private void moveStudyDomain(Domain domain, Container container)
+    {
+        try (DbScope.Transaction transaction = OntologyManager.getExpSchema().getScope().ensureTransaction())
+        {
+            OntologyManager.clearCaches();
```
labkey-adam: You might want to clearCaches() at the end or in a commit task to avoid race conditions. Not really a concern for a normal upgrade scenario (which is single-threaded), but it could be a problem if upgrade code is invoked via the UI.

labkey-adam (approved):
Suggest moving `clearCaches()`, but no need for re-review

study/src/org/labkey/study/designer/StudyDesignManager.java
```diff
@@ -703,6 +706,72 @@ public void ensureStudyDesignDomains(Container c, User user)
...
+            {
+                SQLFragment sql = new SQLFragment("UPDATE ").append(OntologyManager.getTinfoPropertyDescriptor(), "")
+                        .append(" SET Container = ? WHERE PropertyID ")
+                        .add(container.getId())
+                        .appendInClause(propertyDescriptors, OntologyManager.getSqlDialect());
+
+                new SqlExecutor(OntologyManager.getExpSchema()).execute(sql);
+            }
+
+            // update any property descriptors associated with this domain
```
cnathe:
```suggestion
            // update the domain descriptor associated with this domainId

```

cnathe (approved):
I'm sure this issue was my doing (back in 2013!). Thanks for cleaning things up.

PR #3993, opened 2023-01-06
Issue 46753: Slow call to query-updateRows call for data class tables

labkey-jeckels:
#### Rationale
Things go faster if you do them once instead of per row.

#### Changes
* Avoid recreating TableInfos when working on a batch of rows
* CachingSupplier to make it easier to be lazy about calculating values
* Improve type safety around ExpDataClass and ExpDataClassImpl


labkey-susanh (commented):
Caching. Nice. I'm going to defer to Xing for approval since I'm not sure how it might interact with the other work going on with data iterators.

labkey-matthewb (approved):

XingY (approved):

PR #3996, opened 2023-01-08
Handle update and clarify merge - remove experimental flag

XingY:
#### Rationale
This PR removes isImportWithUpdateEnabled experimental flag to support the new update action for apps. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1079
* https://github.com/LabKey/testAutomation/pull/1351
* https://github.com/LabKey/platform/pull/3996
* https://github.com/LabKey/dataintegration/pull/199
* https://github.com/LabKey/inventory/pull/684
* https://github.com/LabKey/sampleManagement/pull/1528
* https://github.com/LabKey/biologics/pull/1850

#### Changes
* Remove "allowImportUsingUpdateOnly" experimental flag
* Add assert(supportsInsertOption(context.getInsertOption())) to various locations
* Allow bypassing assert(supportsInsertOption(context.getInsertOption())) for study reload and certain folder import tasks
* Enable SearchIndexIterator using lsid from existing record for updates
* Support vocabulary domain properties update/merge
* Added EntityInsertPanelForUpdate to LabKeyUIComponentsPage
* Allow "Update" action for samples pipeline trigger
* Issue 46649: Request to support "auditBehavior" as a Custom Parameter for "Import Samples From Data File" file watcher task
* Issue 46639: "SampleId" column header not recognized when loading samples from pipeline trigger


api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -467,6 +476,16 @@ protected Map<String, Object> insertRow(User user, Container container, Map<Stri
         return outputRows;
     }
 
+    // not yet supported
```
labkey-susanh, line 479: Is this comment still relevant? The call to `_insertUpdateRowsUsingDIB` suggests that it is supported.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -1807,9 +1810,22 @@ public boolean next() throws BatchValidationException
...
-                if (null != lsid)
-                    _lsids.add(lsid);
+                if (_useExistingRecord)
+                {
+                    Map<String, Object> map = getExistingRecord();
+                    if (map != null && map.containsKey("lsid"))
+                    {
+                        String lsid = (String) map.get("lsid");
+                        if (null != lsid)
+                            _lsids.add(lsid);
```
labkey-susanh, line 1820: Should we throw an error if we don't have an lsid to add at this point or if we don't get a map for he existing record? I'm not sure when that would happen, but it seems like a problem of some sort.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -233,6 +246,28 @@ public DataIteratorBuilder createImportDIB(User user, Container container, DataI
...
     {
         assert _sampleType != null : "SampleType required for insert/update, but not required for read/delete";
+
+        // Issue 44256: We want to support "Name", "SampleId" and "Sample Id" for easier import
+        // Issue 46639: "SampleId" column header not recognized when loading samples from pipeline trigger
+        try
+        {
+            if (rows instanceof DataLoader) // junit test uses ListofMapsDataIterator
+            {
+                ColumnDescriptor[] columnDescriptors = ((DataLoader) rows).getColumns(SAMPLE_ALT_IMPORT_NAME_COLS);
```
labkey-susanh, line 256: Do we not need to support the importAlias parameters we were getting from getRenamedColumns?

experiment/src/org/labkey/experiment/pipeline/SampleReloadTask.java
```diff
@@ -70,13 +77,29 @@ public RecordedActionSet run()
...
-        // parse any options
         if (params.containsKey(MERGE_OPTION))
-            _mergeData = Boolean.parseBoolean(params.get(MERGE_OPTION));
+        {
+            String insertOption = params.get(MERGE_OPTION);
+            log.info("data insert option set as : " + insertOption);
+
+            if ("true".equalsIgnoreCase(insertOption))
+                _insertOption = QueryUpdateService.InsertOption.MERGE;
+            else if ("update".equalsIgnoreCase(insertOption))
```
labkey-susanh: This is a little strange that the option can be either "true" or "update". Why this mixture?

XingY: I agree this is strange. The correct way is to modify SampleReloadTask in expermentContext.xml to have a renamed property name and a new list of import options and also upgrade any existing file watcher. But I decided to go with the easy way to avoid an upgrade script after finding out that list and dataset file watcher don't need similar work because they don't even support merge. Also I wasn't sure if it's a real scenario that user would only "update" but not inserting any new record using file watcher.

I might still have the java upgrade code if folks insist doing it the "right" way.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: the handling of getRenamedColumns has not changed. this change simply always add the 2 columns additionally.

study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -7208,6 +7208,12 @@ public ModelAndView getView(IdForm form, BindException errors) throws Exception
             return getDefaultImportView(form, errors);
         }
 
+        @Override
+        protected boolean skipInsertOptionValidation()
```
XingY, line 7212: This is a workaround to allow SchemaTableInfo study.participants to bypass insert validation.

I don't know Study.ImportAlternateIdMappingAction enough to understand why SchemaTableInfo StudySchema.getInstance().getTableInfoParticipant() instead of ParticipantTable (QuerySchema table) is used for updating Participants. Does anyone know? @LabKey/triage

labkey-adam: I don't know for sure, but it was written 10 years ago and, at that time, study code would often update the database tables directly. Not our current preference, of course.

experiment/src/org/labkey/experiment/pipeline/SampleReloadTask.java
labkey-susanh: It seems not unreasonable that a file watcher would be intended to do an update, but I'm unclear if an upgrade script would be able to make all the changes needed for file watchers clients have created.  Is all the configuration for these in the database and not in external XML files or some such thing?  Another option, if the upgrade script would cause problems with backward compatibility for exiting file watchers, could be a second boolean property, though that's less than ideal since we'd have to check that the two booleans are not in conflict with each other.

I'm definitely not a fan of keeping it this way.  My vote is either we do the upgrade script to avoid this kind of hybrid property or we leave "update" unsupported for file watchers at the moment (possibly opening an issue so we have better chance of remembering).

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
labkey-susanh: Ah, I think I see. This logic was moved here instead of the action so the columns get added for any action, right?

experiment/src/org/labkey/experiment/pipeline/SampleReloadTask.java
XingY: Changes pushed to use java based upgrade script

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: yep

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
XingY: This is not yet supported in this branch. If it would cause much confusion I can roll back this change.

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: I think there is argument for either way...If the update itself succeed but indexing is skipped it's probably fine?

labkey-susanh: Yes, I think that's true.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
labkey-susanh: Up to you, I think. If it's going to linger a long time to confuse someone other than you, I would probably remove it, but if it's just getting picked up in the next story, no need.

experiment/src/org/labkey/experiment/pipeline/SampleReloadTask.java
labkey-susanh: Thanks. Is it worth having constants for those options? (Or are there already constants that could be used?)

labkey-susanh (approved):

PR #3999, opened 2023-01-09
Issue 47052: Improve handling for casing module name mismatches in co…

labkey-adam:
…re.modules

#### Rationale
We want the `core.Modules.Name` column to be case-insensitive on all databases. Remove duplicates and switch to a case-insensitive unique index.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3950

### DIFF
--- a/api/src/org/labkey/api/module/ModuleLoader.java
+++ b/api/src/org/labkey/api/module/ModuleLoader.java
@@ -1656,7 +1656,7 @@ public class ModuleLoader implements Filter, MemTrackerListener
             scope.getSqlDialect().dropSchema(_core.getSchema(), schema);
         }
 
-        Table.delete(getTableInfoModules(), context.getName());
+        removeModuleContext(context);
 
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -1681,6 +1681,16 @@ public void removeModule(ModuleContext context, boolean deleteFiles)
         clearUnknownModuleCount();
     }
 
+    // Simple variant of removeModule(); just remove the row in the table and the context from the map.
+    public void removeUnknownModuleContext(ModuleContext context)
```
labkey-jeckels: I guess we could make removeModule() call this but maybe not important

labkey-jeckels (approved):

api/src/org/labkey/api/module/ModuleLoader.java
labkey-adam: Done

PR #4000, opened 2023-01-09
Issue 46969: ActionURL.getExtraPath() inconsistency

labkey-matthewb:
#### Rationale
Force consistent behavior for getExtraPath() also while we're here:

new method getParsedExtraPath()
fix ActionURL.getParsedPath() to match getPath()

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/view/ActionURL.java
+++ b/api/src/org/labkey/api/view/ActionURL.java
@@ -475,21 +475,37 @@ public class ActionURL extends URLHelper implements Cloneable
     }
 
 
+    /*
+    * This is the portion path that is not used to resolve the controller-action.
+    * THis is usually the container path or container id.
+     */
     public ActionURL setExtraPath(String extraPath)
     {
-        if (null == extraPath) extraPath = "";
-        setParsedPath(Path.parse(extraPath));
+        if (StringUtils.isEmpty(extraPath))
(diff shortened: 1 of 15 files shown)


api/src/org/labkey/api/webdav/SimpleDocumentResource.java
```diff
@@ -68,15 +69,18 @@ public SimpleDocumentResource(Path path, String documentId, String containerId,
         _modifiedBy = modifiedBy;
         _modified = toTime("modified", modified, properties);
         // Make sure the "execute URL" container is supplied as a GUID, not a path, since paths are not stable. Also,
```
labkey-adam:
```suggestion
        // Make sure the "execute URL" container is supplied as a GUID, not a path, since paths are not stable.
```

api/src/org/labkey/api/view/ActionURL.java
```diff
@@ -475,22 +475,38 @@ public ActionURL setContainer(Container c)
     }
 
 
+    /*
+    * This is the portion path that is not used to resolve the controller-action.
+    * THis is usually the container path or container id.
```
labkey-adam:
```suggestion
    * This is usually the container path or container id.
```

labkey-adam (approved):
A couple comment typos could be fixed

PR #4003, opened 2023-01-10
Let the warnings pass through when loading Query Definition

labkey-bpatel:
#### Rationale
Fix for WNPRC test failure "Error on line 4: Query or table not found: ehr_billing_public.aliases"
image

These failures are unveiled as a result of recent change that added a check for presence of errors; however, in the above case, there are warnings generated while loading Query Definition for the linked table (ehr_billing_public.aliases) and are not actual errors.

#### Changes
* Add isError() check to ensure there are indeed any errors, and let the warnings pass through

### DIFF
--- a/query/src/org/labkey/query/QueryDefinitionImpl.java
+++ b/query/src/org/labkey/query/QueryDefinitionImpl.java
@@ -565,7 +565,8 @@ public abstract class QueryDefinitionImpl implements QueryDefinition
                     table = createTable(schema, errors, includeMetadata, null, skipSuggestedColumns);
                 }
 
-                if (null == table || (null != errors && !errors.isEmpty() && errors.stream().anyMatch(error -> error instanceof QueryParseException && ((QueryParseException)error).isError())))
+                if (null == table || (null != errors && !errors.isEmpty() && !errors.stream().allMatch(error -> error instanceof QueryParseException && ((QueryParseException)error).isWarning())))
+
                     return null;
 
                 log.debug("Caching table " + schema.getName() + "." + table.getName());
--- a/query/src/org/labkey/query/QueryDefinitionImpl.java
+++ b/query/src/org/labkey/query/QueryDefinitionImpl.java
@@ -565,7 +565,7 @@ public abstract class QueryDefinitionImpl implements QueryDefinition
                     table = createTable(schema, errors, includeMetadata, null, skipSuggestedColumns);
                 }
 
-                if (null == table || (null != errors && !errors.isEmpty()))
+                if (null == table || (null != errors && !errors.isEmpty() && errors.stream().anyMatch(error -> error instanceof QueryParseException && ((QueryParseException)error).isError())))
                     return null;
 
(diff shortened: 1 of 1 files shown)


query/src/org/labkey/query/QueryDefinitionImpl.java
```diff
@@ -565,7 +565,7 @@ public TableInfo getTable(@NotNull UserSchema schema, @Nullable List<QueryExcept
                     table = createTable(schema, errors, includeMetadata, null, skipSuggestedColumns);
                 }
 
-                if (null == table || (null != errors && !errors.isEmpty()))
+                if (null == table || (null != errors && !errors.isEmpty() && errors.stream().anyMatch(error -> error instanceof QueryParseException && ((QueryParseException)error).isError())))
```
labkey-jeckels: This is subtly different from the similar check here: https://github.com/LabKey/platform/blob/6fb3151a624541e6b2f7133b6527e2415bf191b5/query/src/org/labkey/query/LinkedSchema.java#L345

The other code checks to see if it's explicitly tagged as a warning. This checks if it's an error. If there's some other variant of QueryException in the list, it will continue through.

I haven't tested but I think this will be a more consistent approach:

```suggestion
                if (null == table || (null != errors && !errors.isEmpty() && !errors.stream().allMatch(error -> error instanceof QueryParseException && ((QueryParseException)error).isWarning())))

```

labkey-jeckels (changes requested):

query/src/org/labkey/query/QueryDefinitionImpl.java
labkey-bpatel: Thanks, Josh! That is exactly what i had initially, and it works, then changed it to check for error instead.

labkey-bpatel: Thanks, Josh! That is exactly what i had initially, and it works, then changed it to check for error instead.

labkey-bpatel: Thanks, Josh! That is exactly what i had initially, and it works, then changed it to check for error instead.

labkey-jeckels (approved):
The change looks fine to me but let's check with @labkey-klum and @labkey-tchad to confirm this will get merged into develop. I can't quite remember what happens automatically vs via special request.

labkey-tchad:
> The change looks fine to me but let's check with @labkey-klum and @labkey-tchad to confirm this will get merged into develop. I can't quite remember what happens automatically vs via special request.

23.1-SNAPSHOT will get merged forward tomorrow and then again next Wednesday when we release 23.1.0

labkey-jeckels:
Thanks @labkey-tchad. I'll go ahead and merge now so it gets included tomorrow.

labkey-jeckels:
@labkey-bpatel the tests are failing in 22.11 so we should backport this. @labkey-tchad can we safely cherry-pick merge this into 22.11?

PR #4004, opened 2023-01-10
Issue 47060: NAb calculation of sample group PercentNeutralizationInitialDilution incorrect when plate layout uses ReverseDilutionDirection

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47060

#### Changes
- use the PercentNeutralization value for the DilutionData row where the Dilution matches the MinDilution instead of just DilutionOrder = 1


cnathe:
Note: client requested that we backport this fix to 22.11, so I will be closing this PR and created a new one targeting that version.

labkey-klum (approved):
interesting, I didn't know how many people were aware of that setting

PR #4006, opened 2023-01-10
Move non-code resources from internal to core and ms2

labkey-adam:
#### Rationale
First step in eliminating the `internal` module

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/566


labkey-susanh (approved):
I'm not going to claim to have read or even browsed all these files, but with this change, can we update the `internal/build.gradle` file to remove the resources configuration there?  Have we moved all of those resource out now?

labkey-matthewb (approved):
It builds.  Check it in.

labkey-adam:
> I'm not going to claim to have read or even browsed all these files, but with this change, can we update the `internal/build.gradle` file to remove the resources configuration there? Have we moved all of those resource out now?

I can and will, but my hope is to eliminate this entire module shortly

PR #4016, opened 2023-01-12
Here is another idea for taming robot indexers

labkey-matthewb:
#### Rationale
use rel=canonical to help ignore containerFilter variations

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/action/PermissionCheckableAction.java
+++ b/api/src/org/labkey/api/action/PermissionCheckableAction.java
@@ -174,9 +174,12 @@ public abstract class PermissionCheckableAction implements Controller, Permissio
         }
 
         boolean requiresLogin = actionClass.isAnnotationPresent(RequiresLogin.class);
-        if (requiresLogin && user.isGuest())
+        if (requiresLogin)
         {
-            throw new UnauthorizedException();
+            if (user.isGuest())
+                throw new UnauthorizedException();
+            if (this instanceof BaseViewAction<?> viewaction)
+                viewaction.getPageConfig().setNoIndex();
         }
 
         // User must have ALL permissions in this set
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/action/PermissionCheckableAction.java
```diff
@@ -174,9 +174,12 @@ private void _checkActionPermissions(Set<Role> contextualRoles) throws Unauthori
         }
 
         boolean requiresLogin = actionClass.isAnnotationPresent(RequiresLogin.class);
-        if (requiresLogin && user.isGuest())
+        if (requiresLogin)
         {
-            throw new UnauthorizedException();
+            if (user.isGuest())
+                throw new UnauthorizedException();
+            if (this instanceof BaseViewAction<?> viewaction)
```
labkey-jeckels, line 181: Do we expect this to be common? Won't crawlers always be guests?

Note there are a number of unit test failures when there's not a full HTTP request in progress: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_BvtBPostgres/2225932?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true

api/src/org/labkey/api/view/template/PageConfig.java
```diff
@@ -410,17 +411,27 @@ private String getCanonicalLink(URLHelper current)
...
+        var params = current.getParameters();
+        URLHelper u = current.clone().deleteParameters();
+
+        for (var pair : params)
         {
-            if (null != current.getParameter(p))
-                u = (null==u ? current.clone() : u).deleteParameter(p);
+            if (ignoreParameters.contains(pair.getKey()))
+                continue;
+            if (pair.getKey().endsWith(CONTAINER_FILTER_NAME))
```
labkey-jeckels, line 428:
```suggestion
            // Strip container filters from the URL to prevent crawlers from over-indexing a URL with different parameter values
            if (pair.getKey().endsWith(CONTAINER_FILTER_NAME))

```

labkey-jeckels (changes requested):

api/src/org/labkey/api/action/PermissionCheckableAction.java
labkey-matthewb: I was thinking about how we might use annotations as hints (or explicitly).  I agree this is not functional...

PR #4017, opened 2023-01-12
More reliable approach for de-duplicating module names

labkey-adam:
#### Rationale
Previous approach didn't handle multiple unknown modules whose names differed by case only

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3999

#### Changes
* Deduplicate by updating module names with "_1", "_2", etc. suffixes
* Use the pruned module list when checking for modules that are too old for upgrades

### DIFF
--- a/core/src/org/labkey/core/CoreUpgradeCode.java
+++ b/core/src/org/labkey/core/CoreUpgradeCode.java
@@ -15,11 +15,16 @@
  */
 package org.labkey.core;
 
+import org.apache.commons.collections4.MultiValuedMap;
+import org.apache.commons.lang3.mutable.MutableInt;
 import org.apache.logging.log4j.Logger;
+import org.labkey.api.collections.CaseInsensitiveArrayListValuedMap;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.CoreSchema;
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/CoreUpgradeCode.java
```diff
@@ -76,28 +82,37 @@ public void setDefaultExcludedProjects(ModuleContext context)
...
+        // For each canonical name, de-duplicate every context after the first one
+        multiMap.asMap().values().stream()
+            .filter(contexts -> contexts.size() > 1)
+            .forEach(contexts -> {
+                MutableInt counter = new MutableInt(1);
+                contexts.stream()
+                    .skip(1)
+                    .forEach(ctx -> {
+                        String oldName = ctx.getName();
+                        String newName = oldName + "_" + counter.incrementAndGet();
```
labkey-jeckels, line 112: I suppose it's theoretically possible that someone already has modules like `Demo`, `demo`, and `demo_2` in their list. But if so, they can deal with it themselves.

labkey-jeckels (approved):
Tested with my local DB including duplicated unknown module names.

core/src/org/labkey/core/CoreUpgradeCode.java
labkey-adam: Agreed

PR #4023, opened 2023-01-16
Issue 47043: Increase property name length to account for long folder names as part of property name

labkey-susanh:
#### Rationale
We currently allow folder names to have 255 characters. When a folder is deleted, we make an entry in the prop.properties table with the property name `<folder name>-TABDELETED-FOLDER-Biologics`, which then exceeds the 255 field length for prop.properties.name. In order to be able to delete these folders with really long names, we need more room for a property name derived from it

#### Changes
* Increate prop.properties.name length.

### DIFF
--- a/core/module.properties
+++ b/core/module.properties
@@ -1,6 +1,6 @@
 Name: Core
 ModuleClass: org.labkey.core.CoreModule
-SchemaVersion: 23.001
+SchemaVersion: 23.002
 Label: Administration and Essential Services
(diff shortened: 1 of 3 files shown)


labkey-jeckels (approved):
Note that SQL Server limits PK sizes to 900 bytes, and the PK for prop.properties is (set, name). We're safe here (2 bytes per NVARCHAR character), and perhaps why you chose 400 as the limit.

PR #4027, opened 2023-01-17
Issue 46778: File browser dropdown menu shows 'blank' options (that work) for buttons that are not shown

XingY:
#### Rationale
Issue [46778](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46778): File browser dropdown menu shows 'blank' options (that work) for buttons that are not shown

#### Related Pull Requests
* 

#### Changes
* hide action from context menu if it's hidden from toolbar

### DIFF
--- a/filecontent/webapp/File/panel/Browser.js
+++ b/filecontent/webapp/File/panel/Browser.js
@@ -848,6 +848,7 @@ Ext4.define('File.panel.Browser', {
 
         if (tbarConfig) {
             var actionConfig;
+            var hiddenActions = [];
 
             // Iterate across tbarConfig as button ordering is determined by array order
             for (i=0; i < tbarConfig.length; i++) {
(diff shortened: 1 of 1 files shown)


filecontent/webapp/File/panel/Browser.js
```diff
@@ -3507,8 +3512,22 @@ Ext4.define('File.panel.Browser', {
...
             }
 
+            //Issue 46778: File browser dropdown menu shows 'blank' options (that work) for buttons that are not shown
+            var visibleItems = [];
+            for (var i=0; i < items.length; i++) {
+                var item = items[i];
+                if (!item)
+                    continue;
+
+                if (item === '-')
```
labkey-nicka, line 3522: There are other types of spacing elements that can be used. You may want to consider a `Utils.isString(item)` check instead.

labkey-nicka (approved):

PR #4038, opened 2023-01-19
Hooks to support new Panorama reporting - row spanning, pivot column headings

labkey-jeckels:
#### Rationale
Our grid rendering doesn't quite give the necessary hooks for DisplayColumns to let cells span multiple rows, or to customize the column headings for pivoted data.

#### Changes
* RowSpanner to let a DisplayColumn merge cells
* Make a few variables protected to give access to subclasses


api/src/org/labkey/api/data/DisplayColumn.java
```diff
@@ -94,6 +94,38 @@ public abstract class DisplayColumn extends RenderColumn
     protected Set<ClientDependency> _clientDependencies = new LinkedHashSet<>();
     private final List<ColumnAnalyticsProvider> _analyticsProviders = new ArrayList<>();
 
+    /** Handles spanning multiple rows in a grid. A separate interface to allow for easier mixing and matching with DisplayColumn implementations. */
+    public interface RowSpanner
+    {
+        /** @return total rows the current row should span. 1 is standard */
+        int getRowSpan(RenderContext ctx);
+        void addQueryColumns(Set<FieldKey> fieldKeys);
```
cnathe, line 102: I can look for the implementation of this in the Panorama case, but curious where the QueryColumns added here will be used. Is it in the getRowSpan() to get some data from the given context that will decide if/when to span rows?

cnathe (approved):

api/src/org/labkey/api/data/DisplayColumn.java
labkey-jeckels: Yes, there's a `COUNT` expression in the query that tells me how many rows will be part of the grouping. This means that I don't have to scroll forward in the ResultSet to peek if the following rows are part of the same group, and then rewind.

PR #4039, opened 2023-01-19
Issue 46951: clear domain cache early in process of updating the domain

labkey-susanh:
#### Rationale
When in the process of updating a domain, if the domain cache is cleared too late it can happen that the tableInfo and the domain get out of sync, resulting in an incomplete/incorrect domain for fields that have been added.  Here we add a sometimes-redundant call to clear the domain cache early in the process of updating the domain so this mismatch doesn't happen.

#### Changes
* Clear the domain cache held by `OntologyManager` earlier

### DIFF
--- a/api/src/org/labkey/api/exp/property/DomainUtil.java
+++ b/api/src/org/labkey/api/exp/property/DomainUtil.java
@@ -725,6 +725,10 @@ public class DomainUtil
         {
             if (validationException.getErrors().isEmpty())
             {
+                // Issue 46951: be sure the domain is invalidated early in the process of updating so we don't get
+                // a mismatched tableInfo and domain
+                OntologyManager.clearCaches();
+
                 // Reorder the properties based on what we got from GWT
                 Map<String, DomainProperty> dps = new HashMap<>();
                 for (DomainProperty dp : d.getProperties())


api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -725,6 +725,10 @@ public static ValidationException updateDomainDescriptor(GWTDomain<? extends GWT
         {
             if (validationException.getErrors().isEmpty())
             {
+                // Issue 46951: be sure the domain is invalidated early in the process of updating so we don't get
```
labkey-matthewb, line 728: I think we can be more surgical than this.   However, I think we have a basic problem that the DbSchema cache of TableInfo is not a DatabaseCache (not transaction aware).  I think we need to change that first for schemas with provisioned tables.  Then we can fix the order of invalidating objects here.

labkey-matthewb (changes requested):

api/src/org/labkey/api/exp/property/DomainUtil.java
labkey-susanh: This sounds like more surgery than I have training for so am going to need some more pointers, unless you or @labkey-adam is interested in taking this on 😄. Do you mean converting the `DbSchemaCache._cache` to be a DatabaseCache? Does the SchemaTableInfoCache also need to get updated?

labkey-susanh:
Closing in favor of the more surgical approach here https://github.com/LabKey/platform/pull/4045

PR #4041, opened 2023-01-19
Add support for round tripping of study custom security settings

labkey-klum:
#### Rationale
Folder export/import did not support the specific study security settings which maps groups as well as per dataset permissions. Some groups would work around this by using the import/export policy file that exists on the page, but it would be better if these settings could be handled automatically in the export/import workflow.

image

#### Changes
- Introduce an option under study export options to serialize the custom security policy settings
- Refactor the `StudyPermissionExporter` to handle export and import. This class did most of the work for the existing UI based capabilities

image

### DIFF
--- a/study/src/org/labkey/study/importer/StudySecurityPolicyImporter.java
+++ b/study/src/org/labkey/study/importer/StudySecurityPolicyImporter.java
@@ -1,0 +1,73 @@
+package org.labkey.study.importer;
+
+import org.apache.xmlbeans.XmlObject;
+import org.labkey.api.admin.ImportException;
+import org.labkey.api.query.ValidationException;
+import org.labkey.api.util.XmlBeansUtil;
+import org.labkey.api.util.XmlValidationException;
+import org.labkey.api.writer.VirtualFile;
+import org.labkey.study.security.permissions.StudyPermissionExporter;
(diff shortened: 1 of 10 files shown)


study/src/org/labkey/study/importer/StudySecurityPolicyImporter.java
```diff
@@ -0,0 +1,73 @@
...
+        if (!ctx.isDataTypeSelected(getDataType()))
+            return;
+
+        if (isValidForImportArchive(ctx, root))
+        {
+            ctx.getLogger().info("Loading " + getDescription());
+
+            StudyDocument.Study.StudySecurity studySecurity = ctx.getXml().getStudySecurity();
+            String policyFileName = studySecurity.getFile();
+            ctx.getLogger().info("Loading security policy file from " + policyFileName);
```
labkey-jeckels: Move this into the if() statement below?

labkey-adam (approved):
As we discussed, maybe default this option to unchecked, as we do with other permissions options

labkey-jeckels (approved):

study/src/org/labkey/study/importer/StudySecurityPolicyImporter.java
labkey-klum: 👍

PR #4044, opened 2023-01-19
Handle update and clarify merge - use DIB for updateRows

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1088
- https://github.com/LabKey/biologics/pull/1875
- https://github.com/LabKey/commonAssays/pull/568
- https://github.com/LabKey/inventory/pull/699
- https://github.com/LabKey/labbook/pull/391
- https://github.com/LabKey/platform/pull/4044
- https://github.com/LabKey/premium/pull/361
- https://github.com/LabKey/recipe/pull/79
- https://github.com/LabKey/sampleManagement/pull/1556
- https://github.com/LabKey/signing/pull/12
- https://github.com/LabKey/adjudication/pull/71
- https://github.com/LabKey/customModules/pull/110
- https://github.com/LabKey/OConnorLabModules/pull/100
- https://github.com/LabKey/testAutomation/pull/1364

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/dataiterator/DropColumnsDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/DropColumnsDataIterator.java
@@ -1,0 +1,57 @@
+package org.labkey.api.dataiterator;
+
+import org.labkey.api.data.ColumnInfo;
+
+import java.util.Set;
+import java.util.function.Supplier;
+
+public class DropColumnsDataIterator extends WrapperDataIterator
+{
+    int[] indexMap;
+    int columnCount = 0;
+
+    public DropColumnsDataIterator(DataIterator di, Set<String> drop)
+    {
+        super(di);
+        int inputColumnCount = di.getColumnCount();
+        indexMap = new int[inputColumnCount+1];
+        for (int inIndex = 0; inIndex <= inputColumnCount; inIndex++)
+        {
+            String name = di.getColumnInfo(inIndex).getName();
+            if (!drop.contains(name))
+            {
+                indexMap[++columnCount] = inIndex;
+            }
+        }
+    }
+
+    @Override
+    public int getColumnCount()
+    {
+        return columnCount;
+    }
+
+    @Override
+    public Object get(int i)
+    {
+        return super.get(indexMap[i]);
+    }
+
+    @Override
+    public ColumnInfo getColumnInfo(int i)
+    {
+        return super.getColumnInfo(indexMap[i]);
+    }
+
+    @Override
+    public Object getConstantValue(int i)
+    {
+        return super.getConstantValue(indexMap[i]);
+    }
+
+    @Override
+    public Supplier<Object> getSupplier(int i)
+    {
+        return super.getSupplier(indexMap[i]);
+    }
+}
--- a/api/src/org/labkey/api/dataiterator/DropColumnsDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/DropColumnsDataIterator.java
@@ -5,6 +5,10 @@ import org.labkey.api.data.ColumnInfo;
 import java.util.Set;
 import java.util.function.Supplier;
 
+/*
+ * See SimpleTranslator.selectAll(@NotNull Set<String> skipColumns) for similar functionality, but SampleTranslator
+ * copies data, this is straight pass through.
+ */
 public class DropColumnsDataIterator extends WrapperDataIterator
 {
     int[] indexMap;
--- a/api/src/org/labkey/api/query/CacheClearingQueryUpdateService.java
+++ b/api/src/org/labkey/api/query/CacheClearingQueryUpdateService.java
@@ -100,9 +100,9 @@ public abstract class CacheClearingQueryUpdateService implements QueryUpdateServ
     }
 
     @Override
-    public List<Map<String, Object>> updateRows(User user, Container container, List<Map<String, Object>> rows, List<Map<String, Object>> oldKeys, @Nullable Map<Enum, Object> configParameters, @Nullable Map<String, Object> extraScriptContext) throws InvalidKeyException, BatchValidationException, QueryUpdateServiceException, SQLException
+    public List<Map<String, Object>> updateRows(User user, Container container, List<Map<String, Object>> rows, List<Map<String, Object>> oldKeys, BatchValidationException errors, @Nullable Map<Enum, Object> configParameters, @Nullable Map<String, Object> extraScriptContext) throws InvalidKeyException, BatchValidationException, QueryUpdateServiceException, SQLException
     {
-        List<Map<String, Object>> result = _service.updateRows(user, container, rows, oldKeys, configParameters, extraScriptContext);
+        List<Map<String, Object>> result = _service.updateRows(user, container, rows, oldKeys, errors, configParameters, extraScriptContext);
         clearCache();
         return result;
     }
--- a/api/src/org/labkey/api/query/SimpleQueryUpdateService.java
+++ b/api/src/org/labkey/api/query/SimpleQueryUpdateService.java
@@ -85,6 +85,7 @@ public class SimpleQueryUpdateService extends DefaultQueryUpdateService
         boolean useDib = oldKeys == null && ExperimentalFeatureService.get().isFeatureEnabled(USE_BATCH_UPDATE_ROWS);
         useDib = useDib && !(configParameters != null && Boolean.TRUE == configParameters.get(QueryUpdateService.ConfigParameters.SkipBatchUpdateRows));
         useDib = useDib && !getQueryTable().hasTriggers(container);
+        useDib = useDib && hasUniformKeys(rows);
 
         if (useDib)
         {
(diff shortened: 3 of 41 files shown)


api/src/org/labkey/api/dataiterator/DropColumnsDataIterator.java
```diff
@@ -0,0 +1,57 @@
+package org.labkey.api.dataiterator;
+
+import org.labkey.api.data.ColumnInfo;
+
+import java.util.Set;
+import java.util.function.Supplier;
+
+public class DropColumnsDataIterator extends WrapperDataIterator
```
labkey-susanh, line 12: Adding some javadocs here to explain when/how to use this would be good.

api/src/org/labkey/api/query/CacheClearingQueryUpdateService.java
```diff
@@ -100,9 +100,9 @@ public int mergeRows(User user, Container container, DataIteratorBuilder rows, B
     }
 
     @Override
-    public List<Map<String, Object>> updateRows(User user, Container container, List<Map<String, Object>> rows, List<Map<String, Object>> oldKeys, @Nullable Map<Enum, Object> configParameters, @Nullable Map<String, Object> extraScriptContext) throws InvalidKeyException, BatchValidationException, QueryUpdateServiceException, SQLException
+    public List<Map<String, Object>> updateRows(User user, Container container, List<Map<String, Object>> rows, List<Map<String, Object>> oldKeys, BatchValidationException errors, @Nullable Map<Enum, Object> configParameters, @Nullable Map<String, Object> extraScriptContext) throws InvalidKeyException, BatchValidationException, QueryUpdateServiceException, SQLException
     {
-        List<Map<String, Object>> result = _service.updateRows(user, container, rows, oldKeys, configParameters, extraScriptContext);
+        List<Map<String, Object>> result = _service.updateRows(user, container, rows, oldKeys, errors, configParameters, extraScriptContext);
```
labkey-susanh, line 105: Does this need to check for errors before proceeding?

api/src/org/labkey/api/query/SimpleQueryUpdateService.java
```diff
@@ -74,6 +77,23 @@ public List<Map<String, Object>> insertRows(User user, Container container, List
...
     }
 
+    @Override
+    public List<Map<String, Object>> updateRows(User user, Container container, List<Map<String, Object>> rows, List<Map<String, Object>> oldKeys,
+                                                BatchValidationException errors, @Nullable Map<Enum, Object> configParameters, Map<String, Object> extraScriptContext)
+            throws InvalidKeyException, BatchValidationException, QueryUpdateServiceException, SQLException
+    {
+        boolean useDib = oldKeys == null && ExperimentalFeatureService.get().isFeatureEnabled(USE_BATCH_UPDATE_ROWS);
+        useDib = useDib && !(configParameters != null && Boolean.TRUE == configParameters.get(QueryUpdateService.ConfigParameters.SkipBatchUpdateRows));
+        useDib = useDib && !getQueryTable().hasTriggers(container);        if (useDib)
```
labkey-susanh: Formatting here is weird.

api/src/org/labkey/api/query/SimpleQueryUpdateService.java
```diff
@@ -74,6 +77,23 @@ public List<Map<String, Object>> insertRows(User user, Container container, List
...
+        {
+            List<Map<String, Object>> result = super._updateRowsUsingDIB(user, container, rows, getDataIteratorContext(errors, InsertOption.UPDATE, configParameters), extraScriptContext);
+            afterInsertUpdate(result==null?0:result.size(), errors);
```
labkey-susanh:
```suggestion
            afterInsertUpdate(result==null ? 0 : result.size(), errors);
```

api/src/org/labkey/api/query/UserSchemaAction.java
```diff
@@ -275,9 +273,12 @@ else if (table.getPkColumnNames().size() > 1)
 
                         // 18292 - updateRows expects a null list in the case of an "empty" or null map.
                         List<Map<String, Object>> oldKeys = (oldValues == null || oldValues.isEmpty()) ? null : Collections.singletonList(oldValues);
-                        ret = qus.updateRows(form.getUser(), form.getContainer(), rows, oldKeys, null, null);
+                        ret = qus.updateRows(form.getUser(), form.getContainer(), rows, oldKeys, batchErrors, null, null);
                     }
                 }
+                if (batchErrors.hasErrors())
+                    batchErrors.addToErrors(errors);
```
labkey-susanh, line 280: Seems a little strange to add the `batchErrors` to the `BindException`, but I guess that's the only way out of this method for those errors.

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -876,6 +890,20 @@ public DataIterator getDataIterator(DataIteratorContext context)
             if (null == input)
                 return null;           // Can happen if context has errors
 
+            var drop = new CaseInsensitiveHashSet();
+            for (int i = 1; i <= input.getColumnCount(); i++)
+            {
+                String name = input.getColumnInfo(i).getName();
+                if (isReservedHeader(name))
+                    drop.add(name);
```
labkey-susanh, line 908: Am I reading this right that this causes us to essentially ignore all fields from exp.data except those in `ALLOWED_IMPORT_HEADERS`? It's certainly true that the other fields should not be set by the user, but I might have expected an error to be thrown if a user tried to supply those values.

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -876,6 +890,20 @@ public DataIterator getDataIterator(DataIteratorContext context)
...
+                else if ("ClassId".equalsIgnoreCase(name))
```
labkey-susanh:
```suggestion
                else if (Column.ClassId.name().equalsIgnoreCase(name))
```

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -336,22 +339,45 @@ public List<Map<String, Object>> insertRows(User user, Container container, List
...
+        {
+            Map<Enum, Object> finalConfigParameters = configParameters == null ? new HashMap<>() : configParameters;
+            finalConfigParameters.put(Options.UseLsidForUpdate, true);
+
+            DbScope scope = getSchema().getDbSchema().getScope();
+            results = scope.executeWithRetry(transaction ->
+                    super._updateRowsUsingDIB(user, container, rows, getDataIteratorContext(errors, InsertOption.UPDATE, finalConfigParameters), extraScriptContext));
+
+            if (InventoryService.get() != null && results != null && results.size() > 0 && !errors.hasErrors())
+                InventoryService.get().recomputeSampleTypeRollup(_sampleType, container, false);
```
labkey-susanh, line 363: Does this not need to happen if not using the data iterator?  

Would it make sense to have this be part of the `onSamplesChanged()` method to centralize where it gets called?

XingY: When updating not using data iterator, InventoryItemTable.complete trigger would call recompute to fix up rollup changes that's due to volume or unit changes. When using DIB, we have wired up InventoryItemTable.complete trigger to defer batch recompute.

api/src/org/labkey/api/query/CacheClearingQueryUpdateService.java
XingY: I don't think so. This is consistent with other methods.

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
XingY: If user customized grid to show hidden fields, then export the grid, edit, upload, I expect the import should just update the updatable columns and ignore the others. We also have junit tests that uses the result of api calls to feed to updateRows, which contains those other columns.

labkey-susanh: My main question here is if we're being more permissive than in other places.  Do we do a similar thing for sample types, for example? What about other domains such as lists or data sets?

Also, I wonder if we can use the "isUserEditable" setting for columns somehow (perhaps not because I know this is used for things other than signaling if a column is editable; for example, it is used to help a column get into the details view, even if not really editable. So unfortunate.)

XingY: The logic to drop most columns for data table is copied from samples, see SampleTypeUpdateServiceImpl isReservedHeader. For the 2 types, we needed to drop columns because we add constant or other generated columns that might collide with with non-editable columns. For lists or data sets, because no such constant or auto columns are added during import, it's fine not to drop columns that are provided. For Created/Updated/By/Container columns, there is separate logic that's applied to all tables to drop incoming columns. 

I think switching to check if a column isUserEditable might work. The exception is with Name column (which we don't want to drop in any event), which behaves differently based on the folder's allow user specified names config. So I think I'll just keep using a static list to exclude/include for now.

labkey-susanh: Thanks for the explination.

labkey-susanh (approved):

PR #4051, opened 2023-01-20
ColumnLogging.remapFieldKeys()

labkey-matthewb:
#### Rationale
ColumnLogging participating in remapFieldKeys() makes for much more consistent behavior
Also Junit tests

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/data/ColumnLogging.java
+++ b/api/src/org/labkey/api/data/ColumnLogging.java
@@ -117,7 +117,7 @@ public class ColumnLogging implements Comparable<ColumnLogging>
     @Override
     public int compareTo(@NotNull ColumnLogging o)
     {
-        int ret = this._originalSchemaName.compareTo(o._originalSchemaName);
+        int ret = this._originalSchemaName.compareToIgnoreCase(o._originalSchemaName);
         if (0 == ret)
             ret = this.getOriginalTableName().compareToIgnoreCase(o.getOriginalTableName());
         if (0 == ret)
--- a/api/src/org/labkey/api/data/ColumnLogging.java
+++ b/api/src/org/labkey/api/data/ColumnLogging.java
@@ -20,6 +20,8 @@ import org.jetbrains.annotations.NotNull;
 import org.labkey.api.query.FieldKey;
 
 import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/data/ColumnLogging.java
```diff
@@ -82,15 +114,12 @@ public String getOriginalTableName()
...
-    public void setOriginalTableName(String originalTableName)
-    {
-        _originalTableName = originalTableName;
-    }
-
     @Override
     public int compareTo(@NotNull ColumnLogging o)
     {
-        int ret = this.getOriginalTableName().compareToIgnoreCase(o.getOriginalTableName());
+        int ret = this._originalSchemaName.compareTo(o._originalSchemaName);
```
labkey-jeckels: compareToIgnoreCase() here too?

query/src/org/labkey/query/sql/QuerySelectView.java
```diff
@@ -425,6 +355,86 @@ else if (!shouldLogNameLoggings.isEmpty())
     }
 
 
+    @NotNull
+    private static Set<ColumnInfo> getRequiredDataLoggingColumns(TableInfo table, @NotNull QueryLogging queryLogging, Set<ColumnInfo> allInvolvedColumns, List<ColumnInfo> allColumns, Map<FieldKey, ColumnInfo> columnMap)
+    {
+        // Check allInvolved columns for which need to be logged
+        // Logged columns may also require data logging (e.g. a patientId)
+        // If a data logging column cannot be found, we will only disallow this query (throw an exception)
+        //      if the logged column that requires data logging is in allColumns (which is selected columns plus ones needed for sort/filter)
```
labkey-jeckels, line 364: I'm not clear on what happens in this case - the description seems to be missing the expected behavior.

labkey-jeckels (approved):

api/src/org/labkey/api/data/ColumnLogging.java
labkey-matthewb: probably better yes.

query/src/org/labkey/query/sql/QuerySelectView.java
labkey-matthewb: Yes, this is problematic.  None of this code is new.

PR #4052, opened 2023-01-21
Issue 46814: Creating Project from Template folder doesn't copy Portal page layout

labkey-jeckels:
#### Rationale
For a long time, we've been creating duplicate copies of a new folder's main portal page.

#### Changes
* Create the portal page definition with the correct name, `DefaultDashboard`, instead of relying on our migration code to copy it
* Get rid of duplicate copies that already exist in the DB
* Use our constant


labkey-adam (dismissed):

labkey-adam (changes requested):
Looks like the script fails on bootstrap because the core scripts run before the props schema is created. Need to either make the updates conditional, move them to props scripts (probably more appropriate), or use deferred upgrade code.

labkey-jeckels:
> Looks like the script fails on bootstrap because the core scripts run before the props schema is created. Need to either make the updates conditional, move them to props scripts (probably more appropriate), or use deferred upgrade code.

I noticed the same thing and moved from core->prop

labkey-adam (approved):

PR #4055, opened 2023-01-24
Fix folder types that don't use the default portal page as their primary tab

labkey-jeckels:
#### Rationale
My fix to avoid double-creating portal page definitions didn't handle folder types that opt out of the default portal, `DefaultDashboard`

https://github.com/LabKey/platform/pull/4052

#### Changes
* Use `getDefaultTab()` when initializing the portal's web parts to pick up whatever the FolderType subclass wants to use

### DIFF
--- a/api/src/org/labkey/api/module/DefaultFolderType.java
+++ b/api/src/org/labkey/api/module/DefaultFolderType.java
@@ -134,7 +134,7 @@ public class DefaultFolderType implements FolderType
             for (Portal.WebPart part : required)
                 part.setPermanent(true);
 
-        List<Portal.WebPart> all = new ArrayList<>();
+        ArrayList<Portal.WebPart> all = new ArrayList<>();
         List<Portal.WebPart> existingParts = Portal.getEditableParts(c);
 
         if (existingParts.isEmpty())
@@ -191,7 +191,7 @@ public class DefaultFolderType implements FolderType
 
         active.addAll(requiredActive);
         c.setActiveModules(active, user);
-        Portal.saveParts(c, DEFAULT_DASHBOARD, all);
+        Portal.saveParts(c, getDefaultTab().getName(), all);
 
         // A few things left to do; ordering is important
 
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/module/DefaultFolderType.java
```diff
@@ -134,7 +134,7 @@ public void configureContainer(Container c, User user)
             for (Portal.WebPart part : required)
                 part.setPermanent(true);
 
-        List<Portal.WebPart> all = new ArrayList<>();
+        ArrayList<Portal.WebPart> all = new ArrayList<>();
```
labkey-nicka: nit: Just curious, what does this accomplish?

api/src/org/labkey/api/module/DefaultFolderType.java
```diff
@@ -191,7 +191,7 @@ public void configureContainer(Container c, User user)
 
         active.addAll(requiredActive);
         c.setActiveModules(active, user);
-        Portal.saveParts(c, DEFAULT_DASHBOARD, all);
+        Portal.saveParts(c, getDefaultTab().getName(), all);
```
labkey-nicka, line 212: Is there a chance `getDefaultTab()` returns `null` (doesn't appear so based on current implementation of `getDefaultTabs()`)? I see other checks in this file being defensive against `null`. Otherwise, could annotate `getDefaultTab()` as `@NotNull`. You likely already considered this.

labkey-nicka (approved):

api/src/org/labkey/api/module/DefaultFolderType.java
labkey-jeckels: This was a bad merge. In my PR yesterday, I changed it from ArrayList->List. That somehow got reverted when I started my new FB. I'll fix.

api/src/org/labkey/api/module/DefaultFolderType.java
labkey-jeckels: `DefaultFolderType.getDefaultTab()` can't return null. I'll annotate as such and can remove at least one null check. However, another implementation of `FolderType`, `CustomFolderType` is nullable. Thus, not all checks can go.

PR #4057, opened 2023-01-24
Add debug logging to pipeline codepath

bbimber:
Add debug logging to pipeline codepath to investigate dropped pipeline jobs.

This is related to: 
https://www.labkey.org/ONPRC/Support%20Tickets/issues-details.view?issueId=44318
https://www.labkey.org/ONPRC/Support%20Tickets/issues-details.view?issueId=47138

### DIFF
--- a/pipeline/src/org/labkey/pipeline/api/AbstractPipelineQueue.java
+++ b/pipeline/src/org/labkey/pipeline/api/AbstractPipelineQueue.java
@@ -80,7 +80,7 @@ public abstract class AbstractPipelineQueue implements PipelineQueue
         }
         else
         {
-            LOG.debug("Unable to set status to waiting for job: " + job.toString());
+            LOG.debug("Unable to set status to waiting for job: " + job.toString() + ", GUID: " + job.getJobGUID());
         }
     }
 
--- a/pipeline/src/org/labkey/pipeline/api/AbstractPipelineQueue.java
+++ b/pipeline/src/org/labkey/pipeline/api/AbstractPipelineQueue.java
@@ -69,7 +69,7 @@ public abstract class AbstractPipelineQueue implements PipelineQueue
                 PipelineStatusManager.setStatusFile(job, user, PipelineJob.TaskStatus.waiting, null, true);
             }
 
-            job.getLogger().debug("Resetting Job ID: " + job.getJobGUID())
+            job.getLogger().debug("Resetting Job ID: " + job.getJobGUID());
             PipelineStatusManager.resetJobId(job.getContainer(), job.getLogFilePath(), job.getJobGUID());
         }
 
--- a/pipeline/src/org/labkey/pipeline/api/AbstractPipelineQueue.java
+++ b/pipeline/src/org/labkey/pipeline/api/AbstractPipelineQueue.java
@@ -69,6 +69,7 @@ public abstract class AbstractPipelineQueue implements PipelineQueue
                 PipelineStatusManager.setStatusFile(job, user, PipelineJob.TaskStatus.waiting, null, true);
             }
 
+            job.getLogger().debug("Resetting Job ID: " + job.getJobGUID())
             PipelineStatusManager.resetJobId(job.getContainer(), job.getLogFilePath(), job.getJobGUID());
         }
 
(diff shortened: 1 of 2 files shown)


pipeline/src/org/labkey/pipeline/mule/EPipelineQueueImpl.java
```diff
@@ -307,9 +307,14 @@ public static void dispatchJob(PipelineJob job) throws UMOException
...
         try
         {
+            _log.debug("Dispatching job: " + job);
             client = new MuleClient();
             client.dispatch(EPipelineQueueImpl.PIPELINE_QUEUE_NAME, job, null);
         }
+        catch (Exception e)
+        {
+            _log.debug("Error dispatching job: " + job, e);
+        }
```
labkey-jeckels:
```suggestion
        catch (UMOException | RuntimeException e)
        {
            _log.debug("Error dispatching job: " + job, e);
            throw e;
        }
```

I haven't tested this (or even compiled it, actually) but we need to rethrow to avoid changing the execution path.

labkey-jeckels (commented):
One requested change before I mirror this on the LK repo.

These look like good logging spots to get better insight into what's happening.

pipeline/src/org/labkey/pipeline/api/AbstractPipelineQueue.java
```diff
@@ -77,6 +78,10 @@ public void addJob(final PipelineJob job) throws PipelineValidationException
             // Delay until the transaction has been committed so other threads can find the job in the database
             PipelineSchema.getInstance().getSchema().getScope().addCommitTask(() -> enqueue(job), DbScope.CommitTaskOption.POSTCOMMIT);
         }
+        else
+        {
+            LOG.debug("Unable to set status to waiting for job: " + job.toString());
```
labkey-jeckels: It's probably fine as-is for our current purposes, but would it be helpful to consistently log the GUID instead of or in addition to the toString()?

pipeline/src/org/labkey/pipeline/mule/EPipelineQueueImpl.java
bbimber: good point

bbimber: @labkey-jeckels  Why limit to just two exception classes if we're just going to rethrow it anyway?

pipeline/src/org/labkey/pipeline/api/AbstractPipelineQueue.java
bbimber: so like my last commmit?

pipeline/src/org/labkey/pipeline/mule/EPipelineQueueImpl.java
```diff
@@ -307,9 +307,15 @@ public static void dispatchJob(PipelineJob job) throws UMOException
...
         try
         {
+            _log.debug("Dispatching job: " + job);
             client = new MuleClient();
             client.dispatch(EPipelineQueueImpl.PIPELINE_QUEUE_NAME, job, null);
         }
+        catch (Exception e)
+        {
+            _log.debug("Error dispatching job: " + job, e);
+            throw e;
```
labkey-jeckels, line 317: I didn't think this would compile because you can't throw `Exception` without declaring that you throw that base class, hence my suggestion of targeting those two specific subclasses. But I'm wrong!

labkey-jeckels:
Mirrored into LK repo for build/test. I'll check the results tomorrow and merge assuming no snags. https://github.com/LabKey/platform/tree/22.11_fb_pipeplinelog

pipeline/src/org/labkey/pipeline/mule/EPipelineQueueImpl.java
bbimber: yeah, i see why you'd say that, actually. apparently that is OK though

bbimber:
@labkey-jeckels : seems like BVTs passed: https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_BvtBPostgres?branch=pipeplinelog&buildTypeTab=overview&mode=builds

labkey-jeckels (approved):

bbimber:
thanks @labkey-jeckels

PR #4059, opened 2023-01-25
Logging to track down bad URL for editing user domain

labkey-jeckels:
#### Rationale
A production server is generating URLs for editing the user domain that point to the root container, when `/Shared` is the right path. I think we're likely resolving the wrong `DomainKind` but am not sure which one is winning over `UsersDomainKind`

#### Changes
* Log the domain kind


labkey-adam:
@labkey-jeckels I don't know which server this is, but was this addressed by my 22.11 fix? https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46668

labkey-adam:
Or maybe the server is on 22.11 already?

labkey-jeckels:
Already fixed in 22.11.

PR #4065, opened 2023-01-27
Update data generator to support folders, assay designs, and storage hierarchy

labkey-susanh:
#### Rationale
Here we beef up our capabilities for generating data at scale to include the generation of folders, assay designs, and storage hierarchies. This PR does not include the generation of data within any of these entities, only the definition of the entities themselves.

#### Related Pull Requests
- https://github.com/LabKey/inventory/pull/707
- https://github.com/LabKey/biologics/pull/1891

#### Changes
* add `DataGeneratorRegistry` and `DataGenerationDriver` interface
* add `generateFolders` method to `DataGenerator`
* add `AssayDesignGenerator`
* In `createContainerList`, don't add the current container more than once.

### DIFF
--- a/api/src/org/labkey/api/data/generator/DataGenerator.java
+++ b/api/src/org/labkey/api/data/generator/DataGenerator.java
@@ -48,6 +48,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.stream.Collectors;
 
 public class DataGenerator<T extends DataGenerator.Config>
 {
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -174,14 +178,46 @@ public void setCustomDataClasses(List<ExpDataClass> customDataClasses)
...
+        checkAlive(_job);
+        int numFolders = _config.getNumFolders();
+        if (numFolders <= 0)
+        {
+            _log.info(String.format("No folders generated because %s=%d", Config.NUM_FOLDERS, numFolders));
+            return;
+        }
+        CPUTimer timer = addTimer(String.format("%d sub-folders", numFolders));
+        timer.start();
+        Set<String> currentChildren = ContainerManager.getChildren(getContainer()).stream().map(Container::getName).collect(Collectors.toSet());
```
labkey-nicka: Should this be a case-insensitive Set?

api/src/org/labkey/api/data/generator/DataGeneratorRegistry.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.Map;
+
+public class DataGeneratorRegistry
+{
+    public enum DataType {
+        AssayDesigns,
+        Storage,
+    };
+
+    private static Map<DataType, DataGenerator.DataGenerationDriver> _dataGeneratorMap = new HashMap<>();
```
labkey-nicka: nit: Can be `final`.

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -637,6 +682,7 @@ public static long randomLong(long startIndex, long endIndex)
         public static final String MIN_NUM_FIELDS = "minFields";
         public static final String MAX_NUM_FIELDS = "maxFields";
 
+        int _numFolders = 0;
```
labkey-nicka: Why do these have default values provided here (as opposed to the default values in the constructor) where as you've elected to remove the default values in the `BiologicsDataGenerator`?

labkey-nicka (approved):

api/src/org/labkey/api/data/generator/DataGeneratorRegistry.java
labkey-susanh: Thanks. Done.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: 'cause I forgot to update this class. Done.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: Yes, I guess we are case-insensitive when checking for folder name overlaps.  Done.

PR #4066, opened 2023-01-27
Separate user menu retrieval from product menu retrieval

labkey-susanh:
#### Rationale
The product menu no longer includes a User section, so there's no need to retrieve that data for rendering the product menu, and there's no need to retrieve all of the product menu data for showing the user menu. Here we add a separate action for getting just the User section, and remove it from the general product menu retrieval.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1093 

#### Changes
* Add separate action and method for getting user menu

### DIFF
--- a/core/api-src/org/labkey/api/products/ProductMenuProvider.java
+++ b/core/api-src/org/labkey/api/products/ProductMenuProvider.java
@@ -18,7 +18,6 @@ package org.labkey.api.products;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.data.Container;
-import org.labkey.api.util.CPUTimer;
 import org.labkey.api.util.HelpTopic;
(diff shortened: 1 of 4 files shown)


core/api-src/org/labkey/api/products/ProductMenuProvider.java
```diff
@@ -18,6 +18,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.data.Container;
+import org.labkey.api.util.CPUTimer;
```
cnathe: is this import needed/used?

cnathe (approved):

core/api-src/org/labkey/api/products/ProductMenuProvider.java
labkey-susanh: Oops. from a different change list. Removed.

PR #4072, opened 2023-01-30
Data container when updating using DIB

XingY:
#### Rationale
When updating using DIB, update should fail when data container is different from current container. 

#### Related Pull Requests
* 

#### Changes
* make sure validateExistingData is always true for update actions
* fix container validation for samples when updating using lsid


api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -1258,20 +1258,11 @@ public void UPDATE() throws Exception
             // test existing row value is not updated/erased
             assertEquals(2, rows.get(2).get("i"));
 
-            // update should skip the new record row silently and update the rest rows successfully
+            // update should fail if a new record is provided and VerifyExistingData = true
```
labkey-susanh: Is `VerifyExistingData` still used?

labkey-susanh (approved):

PR #4073, opened 2023-01-30
Transaction cache optimization

labkey-adam:
#### Rationale
`TransactionCache` holds a shared cache and a private cache specific to the current thread's transaction. Previously, it would read through to the shared cache until _any_ write operation occurred within the transaction, at which point it would switch exclusively to the private cache for the remainder of the transaction; this often meant many queries to populate the initially empty private cache. This new implementation reads through or delegates to the private cache on an entry-by-entry basis; a write operation on a specific entry blocks read-through for only that key. `clear()` is the only operation that causes a hard switchover to the private cache. (As always, avoid using `clear()`!)

#### Changes
* Optimize TransactionCache to use per-entry read-through and delegation
* Restore previous `DbSchemaType` cache key behavior and add comments
* Switch `UserCache` from a `DatabaseCache` subclass that sets up shared and private caches as `BlockingCaches` to a `BlockingCache` that wraps a `DatabaseCache`. This is cleaner, standard practice and avoid the need for `TransactionCache` to jump through hoops to get ahold of the needed `CacheLoader`.

### DIFF
--- a/api/src/org/labkey/api/cache/TransactionCache.java
+++ b/api/src/org/labkey/api/cache/TransactionCache.java
@@ -25,7 +25,7 @@ import java.util.Set;
 /**
  * A read-through, transaction-specific cache. Reads through to the shared cache for each entry until a write operation
  * (put or remove) occurs on that entry, at which point only the private cache is consulted for this entry for the
- * remainder of the transaction. This should provide for good performance while avoiding pollution of the shared cache
+ * remainder of the transaction. This should provide good performance while avoiding pollution of the shared cache
  * during the transaction and in the case of a rollback.
  * User: adam
  * Date: Nov 9, 2009
@@ -40,7 +40,6 @@ public class TransactionCache<K, V> implements Cache<K, V>
 
     /** Cache shared by other threads */
     private final Cache<K, V> _sharedCache;
-    private final @Nullable CacheLoader<K, V> _privateCacheLoader;
 
     /** Our own private, transaction-specific cache, which may contain database changes that have not yet been committed */
     private final Cache<K, V> _privateCache;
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/cache/TransactionCache.java
```diff
@@ -17,26 +17,34 @@
...
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.labkey.api.util.Filter;
 
 import java.util.Set;
 
 /**
- * A read-through, transaction-specific cache. Reads through to the shared cache until any write occurs, at which point
- * it switches to using a private cache for the remainder of the transaction. This avoids polluting the shared cache
+ * A read-through, transaction-specific cache. Reads through to the shared cache for each entry until a write operation
```
labkey-jeckels, line 26: We don't want to be caching mutable objects at all. But we do, and this approach pollutes the shared cache if an object's already been loaded and the transaction never commits.

labkey-adam:
This branch is nearly all green on BVTs and Dailies

api/src/org/labkey/api/cache/TransactionCache.java
labkey-adam: Previous approach had that risk as well since we'd `get()` from the shared cache before any write occurred. And often we transact a single object mutation (though not always). And of course non-transaction-aware caches always have this problem.

api/src/org/labkey/api/cache/TransactionCache.java
```diff
@@ -47,81 +56,74 @@ public TransactionCache(Cache<K, V> sharedCache, Cache<K, V> privateCache)
...
-            // in the private cache
-            v = _sharedCache.get(key);
-
-            if (null == v)
-            {
-                v = loader.load(key, arg);
-                _hasWritten = true;
-                _privateCache.put(key, v);
-            }
+            v = loader.load(key, arg);
```
labkey-matthewb, line 81: Maybe a comment here about how this behaves wrt BlockingCache (e.g. it doesn't because _privateCache is effectively thread local)

labkey-matthewb (approved):
I like it.

api/src/org/labkey/api/cache/TransactionCache.java
labkey-adam: Done

PR #4074, opened 2023-01-31
Simplify portal defaults, get rid of 10-year-old migration code

labkey-jeckels:
#### Rationale
We've keep migration code for portal layouts for 10+ years. With my recent improvements to how we initialize portal pages, we can finally get rid of it.

We can also stop retaining site-wide state in `_activePortalPage` which is no longer needed to highlight the correct tab.

#### Related Pull Requests
* https://github.com/LabKey/immport/pull/85

#### Changes
* No need to copy across portal page names - we now create them correctly in the first place
* Convert side-wide member variable, `_activePortalPage` to a local variable


labkey-adam:
So sad to see this beautiful code going away

labkey-adam (approved):

labkey-adam (approved):

PR #4080, opened 2023-01-31
Issue 47194: A deleted user causes havoc with timelines and jobs

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47194

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1092
- https://github.com/LabKey/labkey-ui-premium/pull/30

#### Changes
- add null check in AbstractTimelineEvent getUserObject()

### DIFF
--- a/api/src/org/labkey/api/audit/AbstractTimelineEvent.java
+++ b/api/src/org/labkey/api/audit/AbstractTimelineEvent.java
@@ -3,6 +3,7 @@ package org.labkey.api.audit;
 import org.json.JSONObject;
 import org.labkey.api.data.Container;
 import org.labkey.api.security.User;
+import org.labkey.api.security.UserManager;
 import org.labkey.api.settings.LookAndFeelProperties;
 
 import java.text.SimpleDateFormat;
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/audit/AbstractTimelineEvent.java
```diff
@@ -28,6 +28,7 @@
 
     public static Object getUserObject(User user, User currentUser)
     {
+        if (user == null) return null;
```
labkey-susanh, line 37: Looks like the callers of this method are just putting the value in a map. I assume we handle the null value in the map well?

labkey-susanh (approved):

api/src/org/labkey/api/audit/AbstractTimelineEvent.java
cnathe: As discussed. I did have to make client side changes to handle this "null" as well so that we display "<unknown user>". I also just pushed some changes that will at least display the userId in some cases that we can at least return that value from the server side to the client.

labkey-susanh: The comment above this line that returns null doesn't really make sense to me.

api/src/org/labkey/api/audit/AbstractTimelineEvent.java
```diff
@@ -26,9 +27,27 @@
...
-        return createEntityObject(user.getUserId(), user.getDisplayName(currentUser), null, "user");
+        // Issue 47194: if the user has been deleted, still show the userId in the timeline event details
+        if (user == null) return null;
+
+        return getUserObject(user.getUserId(), user, currentUser);
+    }
+
+    private static Object getUserObject(int userId, User user, User currentUser)
+    {
+        User _user = user != null ? user : UserManager.getUser(userId);
```
labkey-susanh: Syntax wars here. I think usually in Java the `_` prefix is for member variables for a class. Not a big deal, unless it comes into use more often, in which case I will likely grumble more.

labkey-susanh (approved):

api/src/org/labkey/api/audit/AbstractTimelineEvent.java
cnathe: good point, removed comment from here but left it at line 45

api/src/org/labkey/api/audit/AbstractTimelineEvent.java
cnathe: good point. changed to `User eventUser`

PR #4083, opened 2023-02-01
Move getReportInfos.api to reports controller since it does not rely on study module

cnathe:
#### Rationale
This was discovered because of a 22.11 > Trial LIMS Integration Test > BiologicsReportTest.testOpenReportInBiologics test failure. The study module does not exist on this instance so the query to get the report infos for a given grid never happens. This query actually has nothing to do with the study module, so the action is being moved to the ReportsController in the query module instead.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4083
* https://github.com/LabKey/labkey-ui-components/pull/1096
* https://github.com/LabKey/biologics/pull/1906
* https://github.com/LabKey/inventory/pull/715
* https://github.com/LabKey/sampleManagement/pull/1585

#### Changes
* move GetReportInfosAction from study module to query module ReportsController

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0",
       "dependencies": {
         "@labkey/api": "1.18.0",
-        "@labkey/components": "2.280.0",
+        "@labkey/components": "2.288.0-fb-getReportInfosAction.0",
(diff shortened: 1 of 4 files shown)


labkey-susanh (approved):
Yay for this. Makes much more sense.

PR #4086, opened 2023-02-02
Add data generator for adding samples to storage

labkey-susanh:
#### Rationale
For the next round of data generation, we want to put samples into storage, so we need some utility methods to help find the samples that are not already in storage.  We also modify a few data generation method signatures in preparation for generating data within folders.

#### Related Pull Requests
* https://github.com/LabKey/inventory/pull/716
* https://github.com/LabKey/biologics/pull/1908

#### Changes
* Add new data generation type: `SamplesInStorage`
* Add label colors when generating sample types
* Add metric units when generating sample types
* Update some method signatures to take container as an argument, in preparation for generating samples in folders
* Add new utility methods for getting the count of samples in a container (with a container filter) and the sample of a particular sample type using a container filter.

### DIFF
--- a/api/src/org/labkey/api/data/generator/DataGenerator.java
+++ b/api/src/org/labkey/api/data/generator/DataGenerator.java
@@ -53,6 +53,29 @@ import java.util.stream.Collectors;
 
 public class DataGenerator<T extends DataGenerator.Config>
 {
+    private static final List<String> UNITS = List.of("g", "mg", "uL", "mL", "unit");
+    private static final List<String> LABEL_COLORS = new ArrayList<>();
+    static {
+        LABEL_COLORS.add("#800000"); // maroon,
+        LABEL_COLORS.add("#B22222"); // firebrick
+        LABEL_COLORS.add("#DC143C"); // crimson
+        LABEL_COLORS.add("#FA8072"); // salmon
+        LABEL_COLORS.add("#FFD700"); // gold
+        LABEL_COLORS.add("#808000"); // olive
+        LABEL_COLORS.add("#9ACD32"); // yellow green
+        LABEL_COLORS.add("#6B8E23"); // olive drab
+        LABEL_COLORS.add("#008000"); // olive
+        LABEL_COLORS.add("#556B2F"); // dark olive green
+        LABEL_COLORS.add("#008080"); // teal
+        LABEL_COLORS.add("#00FFFF"); // aqua
+        LABEL_COLORS.add("#4682B4"); // steel blue
+        LABEL_COLORS.add("#1E90FF"); // dodger blue
+        LABEL_COLORS.add("#0000CD"); // medium blue
+        LABEL_COLORS.add("#8A2BE2"); // blue violet
+        LABEL_COLORS.add("#4B0082"); // indigo
+        LABEL_COLORS.add("#8B008B"); // dark magenta
+        LABEL_COLORS.add("#D2691E"); // chocolate
+    }
     protected static final int MAX_BATCH_SIZE = 10000;
     protected Container _container;
     protected final User _user;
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -254,7 +277,8 @@ private ExpSampleType generateSampleType(String sampleTypeName, @Nullable String
         SampleTypeService service = SampleTypeService.get();
         _log.info(String.format("Creating Sample Type '%s' with %d fields", sampleTypeName, numFields));
         return service.createSampleType(_container, _user, sampleTypeName,
-                "Generated sample type", props, List.of(), namingPattern);
+                "Generated sample type", props, List.of(), -1, -1, -1, -1, namingPattern, null, null, null,
+                LABEL_COLORS.get(randomInt(0, LABEL_COLORS.size())), UNITS.get(randomInt(0, UNITS.size())), null, null, null);
```
labkey-nicka: nit: Could consider having another random access method for a `List<T>` (similar to what `DataGenerator` already has for `T[] array`):

```java
public static <T> T randomIndex(List<T> list)
{
    return list.get(randomInt(0, list.size()));
}
```

api/src/org/labkey/api/exp/api/ExpSampleType.java
```diff
@@ -48,9 +48,15 @@ public interface ExpSampleType extends ExpObject
     String getMaterialLSIDPrefix();
 
 
-    /** pass in a container to request a sample */
+    /** pass in a container to request a set of samples*/
```
labkey-nicka: nit: Reads a little weird to me. Maybe something like `/** All samples in a given container */` or `/** Returns all samples in a given container */`.

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -53,6 +53,29 @@
...
+        LABEL_COLORS.add("#FFD700"); // gold
+        LABEL_COLORS.add("#808000"); // olive
+        LABEL_COLORS.add("#9ACD32"); // yellow green
+        LABEL_COLORS.add("#6B8E23"); // olive drab
+        LABEL_COLORS.add("#008000"); // olive
+        LABEL_COLORS.add("#556B2F"); // dark olive green
+        LABEL_COLORS.add("#008080"); // teal
+        LABEL_COLORS.add("#00FFFF"); // aqua
+        LABEL_COLORS.add("#4682B4"); // steel blue
+        LABEL_COLORS.add("#1E90FF"); // dodger blue
```
labkey-nicka, line 72: ⚾ ⚾ ⚾

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -657,13 +657,33 @@ public void setMaterialLSIDPrefix(String s)
...
 
+    @Override
+    public long getSamplesCount(Container c, @Nullable ContainerFilter cf)
+    {
+        SimpleFilter filter = SimpleFilter.createContainerFilter(c);
+        filter.addCondition(FieldKey.fromParts("CpasType"), getLSID());
+        if (cf != null)
+            filter.addCondition(cf.createFilterClause(ExperimentServiceImpl.get().getExpSchema(), FieldKey.fromParts("Container")));
+
+        return new TableSelector(ExperimentServiceImpl.get().getTinfoMaterial(), filter, null).getRowCount();
```
labkey-nicka: By default `TableSelector` will include `ALL_COLUMNS` in the query. When querying for row counts it should suffice to only include a PK column which in this case is `RowId` I believe.

labkey-nicka (approved):

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: Good thought. Done.

api/src/org/labkey/api/exp/api/ExpSampleType.java
labkey-susanh: Yeah. it's weird. Updated the wording here and elsewhere.

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
labkey-susanh: Good catch. Thanks.

PR #4087, opened 2023-02-02
Issue 47213: TempTableInClauseGenerator filling up memory with cache keys

labkey-martyp:
#### Rationale
TempTableInClauseGenerator caches references to temp tables using a concatenation of the values for the key.  On very large operations this can take up a lot of heap space. This PR uses a hash of the concatenated string for the cache key which greatly reduces the size.

#### Changes
* Hash In Clause Temp Table cache key
* Move HashHelpers out of internal


api/src/org/labkey/api/data/TempTableInClauseGenerator.java
```diff
@@ -180,11 +181,11 @@ else if (!(i instanceof Integer))
...
     private String getCacheKey(@NotNull JdbcType jdbcType, @NotNull Collection<?> params)
     {
-        StringBuilder key = new StringBuilder(jdbcType.name());
+        StringBuilder paramBuilder = new StringBuilder();
         for (Object param : params)
-            key.append("_").append(param);
+            paramBuilder.append(param);
 
-        return key.toString();
+        return jdbcType.name() + "_" + params.size() + "_" + HashHelpers.hash(paramBuilder.toString());
```
labkey-jeckels, line 189: A comment here that describes the motivation for hashing would be nice.

labkey-jeckels (approved):
@labkey-klum heads up that this will likely cause a trivial merge conflict when it gets to develop due to Adam's move of all Internal module Java files into API.

api/src/org/labkey/api/data/TempTableInClauseGenerator.java
labkey-martyp: Comment added

PR #4088, opened 2023-02-02
Add public getter for average and calls to CPUTimer

labkey-danield:
#### Rationale
CPUTimer had public getters for all the values it collected except average and the number of calls made. If you wanted those you had to get the string output from the timer then parse the string, etc...

#### Related Pull Requests
* None.

#### Changes
* Added getAverage()
* Added getCalls()

### DIFF
--- a/api/src/org/labkey/api/util/CPUTimer.java
+++ b/api/src/org/labkey/api/util/CPUTimer.java
@@ -131,7 +131,7 @@ public class CPUTimer
 
     public float getAverage()
     {
-        return (_calls == 0 ? 0 : (float)getTotalMilliseconds() / _calls);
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/util/CPUTimer.java
```diff
@@ -129,6 +129,16 @@ public String getDuration()
     }
     
 
+    public float getAverage()
+    {
+        return (_calls == 0 ? 0 : (float)getTotalMilliseconds() / _calls);
```
labkey-susanh:
```suggestion
        return _calls == 0 ? 0 : (float)getTotalMilliseconds() / _calls;
```
Not necessary, of course.

labkey-susanh (approved):

PR #4097, opened 2023-02-06
Issue 46449: Push inactive users to the bottom of the impersonate user drop-down list

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46449
Changes from the related PR made improvements to the display of inactive users in the impersonate users modal dropdown input. This PR increases the width of the hover tooltip that is displayed when you hover over an inactive users in the dropdown so that the text is not truncated to a single line.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/3918

#### Changes
- increase width of hover tooltip for inactive users in impersonate users modal dropdown


labkey-adam (approved):
Thanks!

PR #4101, opened 2023-02-08
Fix cached-miss handling in TransactionCache

labkey-adam:
#### Rationale
`TranactionCache` has never been able to distinguish between the shared cache holding a cached miss (which should be treated as null) vs. a missing entry (which should result in a load into the private cache). This can lead to significant performance issues if many nulls are being cached, as can be seen in https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47234.


labkey-matthewb (approved):

labkey-jeckels (approved):

labkey-nicka (approved):
Verified that this addresses the performance problem we are specifically seeing in [Issue 47234](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47234).

PR #4104, opened 2023-02-08
Issue 47245: Samples with a recomputeRollup value of "true" remain in the system

XingY:
#### Rationale
Certain inventory insert/update events resulted in samples that are not aliquot parents being marked as need recalc. For those events, we indiscriminately marked a sample as need recalc, if itself is not an aliquot. In reality, only root samples with aliquots ever needs rollup recalc. On action complete, the code to recal only handled parent samples, resulting in orphaned true recomputeRollup samples in system. Those orphaned records slows down every sample insert/update actions, signaling the need to recalc, but never got cleaned up. 

This PR aim to prevent those orphaned records and clean up existing ones.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/4104
- https://github.com/LabKey/inventory/pull/725

#### Changes
- upgrade script to clean up orphaned recomputeRollup=true samples
- move RecomputeAliquotRollup here from InventoryController, with additional handling of clean up orphaned records

### DIFF
--- a/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
+++ b/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
@@ -21,6 +21,8 @@ import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.labkey.api.collections.CaseInsensitiveHashSet;
 import org.labkey.api.data.ColumnInfo;
+import org.labkey.api.data.Container;
+import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.DbSchema;
 import org.labkey.api.data.DbScope;
 import org.labkey.api.data.PropertyStorageSpec;
@@ -33,17 +35,22 @@ import org.labkey.api.data.TableChange;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.TableSelector;
 import org.labkey.api.data.UpgradeCode;
+import org.labkey.api.exp.api.ExpSampleType;
 import org.labkey.api.exp.api.ExperimentService;
+import org.labkey.api.exp.api.SampleTypeService;
 import org.labkey.api.exp.api.StorageProvisioner;
 import org.labkey.api.exp.property.Domain;
 import org.labkey.api.exp.property.PropertyService;
+import org.labkey.api.inventory.InventoryService;
 import org.labkey.api.module.ModuleContext;
 import org.labkey.api.query.FieldKey;
 import org.labkey.experiment.api.ExpSampleTypeImpl;
 import org.labkey.experiment.api.ExperimentServiceImpl;
 import org.labkey.experiment.api.MaterialSource;
 import org.labkey.api.exp.api.SampleTypeDomainKind;
+import org.labkey.experiment.api.SampleTypeServiceImpl;
 
+import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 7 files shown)


experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -235,4 +242,88 @@ private static void fillSampleName(ExpSampleTypeImpl st, Domain domain, DbScope
...
+                containerSampleTypes.get(container).add(cpastype);
+            }
+            return containerSampleTypes;
+        }
+
+        return Collections.emptyMap();
+    }
+
+    /**
+     * Called from exp-22.003-20.004.sql
```
labkey-nicka: This comment needs to be updated.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1282,4 +1285,73 @@ public String getOperationNotPermittedMessage(Collection<? extends ExpMaterial>
         }
         return message + " " + operation.getDescription() + ".";
     }
+
+    private Collection<Integer> _getNonAliquotParentsWithRecalcSql(String sampleTypeLsid, Container container) throws SQLException
+    {
+        SQLFragment sql = new SQLFragment(
+                """
+                        SELECT parent.rowId
+                        FROM exp.material AS parent
```
labkey-nicka, line 1294: I'd recommend referencing the `getTinfoMaterial()` here for consistency. Same for `public long getRecomputeRollupRowCount(ExpSampleType sampleType, Container container)`.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1282,4 +1285,73 @@ public String getOperationNotPermittedMessage(Collection<? extends ExpMaterial>
...
+            }
+        }
+        return parentIds.size();
+    }
+
+    // volumne/unit update events might have marked sample that is not an aliquot parent as need re-calc, which should be cleaned up
+    @Override
+    public int resetRecomputeFlagForNonParents(ExpSampleType sampleType, Container container) throws IllegalStateException, SQLException
+    {
+        Collection<Integer> parentIds = _getNonAliquotParentsWithRecalcSql(sampleType.getLSID(), container);
```
labkey-nicka: Rather than pulling these into memory can these two queries be done together where the `Rowid IN(...)` is replaced with a joined result?

labkey-nicka (approved):

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -235,4 +242,88 @@ private static void fillSampleName(ExpSampleTypeImpl st, Domain domain, DbScope
...
         LOG.info("Sample type '" + st.getName() + "' (" + st.getRowId() + ") updated 'name' column, count=" + count);
     }
+
+    private static Map<String, List<String>> findContainersWithAliquotRecomputes()
+    {
+        SQLFragment sql = new SQLFragment()
+                .append("SELECT distinct cpastype, container\n")
+                .append("FROM ").append(ExperimentService.get().getTinfoMaterial(), "m").append("\n")
+                .append("WHERE m.recomputerollup = ?")
+                .add(true);
```
labkey-susanh, line 252: Do you want to use `getBooleanTRUE()` here?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1282,4 +1285,73 @@ public String getOperationNotPermittedMessage(Collection<? extends ExpMaterial>
...
+                        SELECT parent.rowId
+                        FROM exp.material AS parent
+                        WHERE parent.container = ?
+                            AND parent.RecomputeRollup = ?
+                            AND parent.lsid NOT IN (SELECT distinct rootMaterialLsid FROM exp.material WHERE rootMaterialLsid IS NOT NULL)
+                            AND parent.cpastype = ?"""
+        );
+
+        sql.add(container.getId());
+        sql.add(true);
```
labkey-susanh, line 1302: Also possibly use `getBooleanTRUE()` here.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1282,4 +1285,73 @@ public String getOperationNotPermittedMessage(Collection<? extends ExpMaterial>
         }
         return message + " " + operation.getDescription() + ".";
     }
+
+    private Collection<Integer> _getNonAliquotParentsWithRecalcSql(String sampleTypeLsid, Container container) throws SQLException
```
labkey-susanh: Is the `_` prefix necessary for private methods?  That's not generally done in Java, I think.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1282,4 +1285,73 @@ public String getOperationNotPermittedMessage(Collection<? extends ExpMaterial>
...
+        if (!parentIds.isEmpty())
+        {
+            TableInfo tableInfo = ExperimentService.get().getTinfoMaterial();
+            DbScope scope = tableInfo.getSchema().getScope();
+
+            try (Connection c = scope.getConnection())
+            {
+                SQLFragment sql = new SQLFragment("UPDATE ").append(tableInfo, "").
+                        append(" SET RecomputeRollup = ? WHERE RowId ");
+                sql.add(false);
```
labkey-susanh, line 1326: `SqlDialect.getBooleanFALSE()`

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-adam: `sql.add(true)` is safe, succinct, and generally preferable to string concatenation.

Note that `sql.add(container)` should work as well... we know how to serialize Container parameters.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
XingY: It definitely can. I wanted to provide more logging for upgrade and more detailed information on how many records are out of sync. But I don't know if execute(sql) would always return the number of rows impacted?

PR #4106, opened 2023-02-09
Issue 47276: Bootstrap server does not contain Project web part on home page

labkey-jeckels:
#### Rationale
The portal page legacy code cleanup impacted the layout for bootstrapped servers too, due to the sequence in which the portal is initialized vs when folder types are registered.

Also fixes [Issue 47299: Custom header menus disappeared on upgrade to 23.2](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47299)

#### Changes
* Move adding the project web part a little later in the startup sequence so that folder types are available
* Don't delete menubar items from `portal.default` pages

### DIFF
--- a/api/src/org/labkey/api/module/ModuleLoader.java
+++ b/api/src/org/labkey/api/module/ModuleLoader.java
@@ -593,7 +593,7 @@ public class ModuleLoader implements Filter, MemTrackerListener
         synchronized (_modulesLock)
         {
             for (ModuleContext context : getAllModuleContexts())
-                _moduleContextMap.put(context.getName(), context);
+                _moduleContextMap.putIfAbsent(context.getName(), context); // Don't replace the "Core" context otherwise we'll overwrite its _newInstall and _originalVersion properties
 
             // Refresh our list of modules as some may have been filtered out based on dependencies or DB platform
             modules = getModules();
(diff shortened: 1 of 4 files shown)


labkey-jeckels:
@labkey-adam if you have a better idea for how to sequence the steps here, or for tracking whether it's a new install or not, let me know. The `ModuleContext` passed to `startupAfterSpringConfig()` says `false` when you call `isNewInstall()` and `getOriginalSchemaVersion()` returns `23.003`

labkey-adam:
> @labkey-adam if you have a better idea for how to sequence the steps here, or for tracking whether it's a new install or not, let me know. The `ModuleContext` passed to `startupAfterSpringConfig()` says `false` when you call `isNewInstall()` and `getOriginalSchemaVersion()` returns `23.003`

@labkey-jeckels Module contexts should remember, for the duration of the server session, their original schema version and whether they're a new install. All of them were, except for the core context since we load all the module contexts after core upgrade creates the core.Modules table and inserts a row for the core module, which fools the subsequent load into thinking it's not new. I added a simple fix for this and removed `_bootstrapping`. Please test.

Note: There's also `ModuleLoader.isNewInstall()` for cases when a ModuleContext isn't available.

labkey-adam (approved):

labkey-adam (approved):

PR #4112, opened 2023-02-13
Column logging refactor

labkey-matthewb:
#### Rationale
The query compiler now participates in generating ColumnLogging objects. 

#### Related Pull Requests
* 

#### Changes
*


api/src/org/labkey/api/collections/ArrayListMap.java
```diff
@@ -237,30 +239,27 @@ public boolean containsValue(Object value)
...
-            if (i < _row.size())
+            K key = findEntry.getKey();
+            int i = findEntry.getValue();
+            Object v = i>=_row.size() ? DOES_NOT_CONTAINKEY : _row.get(i);
+            if (v != DOES_NOT_CONTAINKEY)
             {
-                if (_row.get(i) != DOES_NOT_CONTAINKEY)
-                    r.add(new Pair<>(e.getKey(), (V)_row.get(i)));
+                while (entryList.size() <= i)
+                    entryList.add(null);
```
labkey-jeckels, line 258: I'm not understanding why we add nulls to the List and then filter them out when converting to a LinkedHashSet. Couldn't we just add the non-null values to the LinkedHashSet directly?

api/src/org/labkey/api/collections/ArrayListMap.java
```diff
@@ -302,6 +301,37 @@ public Collection<V> values()
...
+
+            {
+                while (_i < _row.size() && _row.get(_i) == DOES_NOT_CONTAINKEY)
+                    _i++;
+            }
+
+            @Override
+            public boolean hasNext()
+            {
+                return _i < _row.size();
```
labkey-jeckels, line 322: Is this correct? Seems like the remaining entries might all be `DOES_NOT_CONTAINKEY`

api/src/org/labkey/api/data/ColumnLogging.java
```diff
@@ -17,56 +17,122 @@
...
+
+    // QuerySelectView.getRequiredDataLoggingColumns() previously used ColumnLogging.equals() to detect
+    // that a column in "allInvolvedColumns" is the same as a column used in a CustomView.
+    //
+    // I don't like this complexity, maybe we can update getRequiredDataLoggingColumns() to not rely on isFromSameColumn().
+    // For now, we can improve how we identify that two ColumnLogging columns represent the same logical underlying column.
+    // E.g. SELECT A.x, B.x FROM Table as A, Table as B     A.x and B.x are not the same even though they come from the same schema/table/column
+    //      SELECT A.createdBy.UserId, B.createdBy.UserId   A.UserId and B.UserId are not the same even though they come from the same schema/table/column
+//    @NotNull protected final String _uniqueColumnKey;
+    @Nullable
```
labkey-jeckels: Can these commented variables be deleted? As it stands now, the `@Nullable` applies to `_shouldLogName` which doesn't make sense for a `boolean` to be nullable

api/src/org/labkey/api/data/ColumnLogging.java
```diff
@@ -17,56 +17,122 @@
...
 
 
-    // we don't usually want to change the original table, but if we do here you go
+    // we don't usually want to change the original table, but if we do here you go.  See ListTable
     public ColumnLogging reparent(TableInfo table)
     {
-        return new ColumnLogging(table.getSchema().getName(), table.getName(), _originalColumnFieldKey, _shouldLogName, _dataLoggingColumns, _loggingComment, _selectQueryAuditProvider);
+        String uniqueColumnKey = makeUniqueKey(table, _originalColumnFieldKey);
+        return new ColumnLogging(table.getUserSchema().getSchemaName(), table.getName(), _originalColumnFieldKey,
+//                uniqueColumnKey, _columnScopeKey,
```
labkey-jeckels: Many places that create ColumnLogging objects have these constructor args commented out. Are they useful to retain?

api/src/org/labkey/api/data/ColumnLogging.java
```diff
@@ -117,16 +202,43 @@ public String getOriginalTableName()
     @Override
     public int compareTo(@NotNull ColumnLogging o)
     {
-        int ret = this._originalSchemaName.compareToIgnoreCase(o._originalSchemaName);
-        if (0 == ret)
-            ret = this.getOriginalTableName().compareToIgnoreCase(o.getOriginalTableName());
-        if (0 == ret)
-            ret = this.getOriginalColumnFieldKey().compareTo(o.getOriginalColumnFieldKey());
-        return ret;
+        throw new UnsupportedOperationException();
```
labkey-jeckels: Should we just not implement `Comparable`?

query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -2659,10 +2660,108 @@ public SQLFragment getSelectSQL(TableInfo table, @Nullable Collection<ColumnInfo
                                     int maxRows, long offset, boolean forceSort, @NotNull QueryLogging queryLogging)
     {
         var query = new Query(table.getUserSchema());
-        var selectView = QuerySelectView.create(query, table, selectColumns, filter, sort, maxRows, offset, forceSort, queryLogging);
+        var selectView = QuerySelectView.create(query, table, selectColumns, filter, sort, maxRows, offset, forceSort, queryLogging, false);
         return selectView.getSql();
     }
 
+    @Override
+    public SelectBuilder getSelectBuilder(TableInfo table)
```
labkey-jeckels, line 2674: Seems like a factory method - rename to `createSelectBuilder`?

query/src/org/labkey/query/sql/AbstractQueryRelation.java
```diff
@@ -0,0 +1,425 @@
...
+ * User: matthewb
+ * Date: Feb 17, 2009
+ * Time: 2:18:05 PM
+ *
+ * A query relation represent a portion of a query with column list.  All columns exported by a relation
+ * must have a one part name.  
+ */
+public abstract class AbstractQueryRelation implements QueryRelation
+{
+    private static final Logger _log = LogManager.getLogger(QuerySelect.class);
```
labkey-jeckels:
```suggestion
    private static final Logger _log = LogManager.getLogger(AbstractQueryRelation.class);
```

query/src/org/labkey/query/sql/AbstractQueryRelation.java
```diff
@@ -0,0 +1,425 @@
...
+            // columns that implement ColumnResolvingRelation, should override this method
+            // that's kinda the point.
+            assert !(getTable() instanceof ColumnResolvingRelation);
+            return null;
+        }
+
+        /** See QNode.gatherInvolvedSelectColumns() and interface QueryRelation.ColumnResolvingRelation*/
+        abstract public Collection<AbstractQueryRelation.RelationColumn> gatherInvolvedSelectColumns(Collection<AbstractQueryRelation.RelationColumn> collect);
+
+        PHI getPHI()
```
labkey-jeckels, line 275: Should this be an abstract method to ensure that all subclasses think about what they should return?

query/src/org/labkey/query/sql/QAggregate.java
```diff
@@ -38,209 +41,213 @@ public class QAggregate extends QExpr
...
+        REGR_COUNT(false, false)
             {
                 @Override
                 boolean dialectSupports(@Nullable SqlDialect d)
                 {
                     return null != d && d.isPostgreSQL();
                 }
             },
-        REGR_INTERCEPT
+        REGR_INTERCEPT(false, false)            // TODO: Should this be true?
```
labkey-jeckels: `false` seems OK to me. It's possible that the intercept happens to be a value in the input data, but won't normally be

query/src/org/labkey/query/sql/QQuery.java
```diff
@@ -156,4 +161,26 @@ public void addFieldRefs(Object referant)
          * Instead, we just addRef() all "upward" references in QuerySelect.getField()
          */
     }
+
+    @Override
+    public Collection<AbstractQueryRelation.RelationColumn> gatherInvolvedSelectColumns(Collection<AbstractQueryRelation.RelationColumn> collect)
+    {
+        var relation = getQuerySelect();
+        AbstractQueryRelation.RelationColumn col = relation.getFirstColumn();
```
labkey-jeckels, line 169: Is there always just a single column in this case? Is this enforced somewhere?

query/src/org/labkey/query/sql/Query.java
```diff
@@ -105,14 +100,21 @@
...
     private boolean _parsingWith = false;
     private boolean _allowDuplicateColumns = true;
 
+
+/*    // Caches for helping track columns through the query, see also RelationColumn.getUniqueName()
+    final IdentityHashMap<QueryTable, Boolean> _qtables = new IdentityHashMap<>();
+    public Map<String,FieldKey> _mapQueryUniqueNamesToSelectAlias = null;
+    final public Map<String,QueryRelation> _mapUniqueNamesToQueryRelation = new HashMap<>();
+*/
+    /* Whenever a RelationColumn with a new unique name is created, add it should be added to this map */
```
labkey-jeckels, line 115:
```suggestion
    /* Whenever a RelationColumn with a new unique name is created, it should be added to this map */
```

query/src/org/labkey/query/sql/QueryLookupWrapper.java
```diff
@@ -384,36 +453,61 @@ public FieldKey getFieldKey()
...
+        ColumnLogging getColumnLogging()
+        {
+            computeColumnLoggings();
+            return _columnLogging;
+        }
     }
 
 
-    class PassThroughColumn extends _WrapperColumn
+    class PassThroughColumn extends QLWColumn
```
labkey-jeckels, line 493: I'm having a hard time understanding the difference between `PassThroughColumn` and `QueryLookupColumn`. Maybe a comment on both?

query/src/org/labkey/query/sql/QueryPivot.java
```diff
@@ -993,6 +1050,11 @@ public Sort getDefaultSort()
             return new Sort();
         }
 
+        @Override
+        public void afterConstruct()
```
labkey-jeckels, line 1054: Useless override

query/src/org/labkey/query/sql/QuerySelect.java
```diff
@@ -1329,6 +1336,12 @@ public QueryTableInfo getTableInfo()
             // Hold a separate reference so we can null it out if the container filter changes
             private SQLFragment _sqlAllColumns = null;
 
+            @Override
+            protected void afterInitializeColumns()
```
labkey-jeckels, line 1340: Useless override

query/src/org/labkey/query/sql/QuerySelect.java
```diff
@@ -2389,6 +2445,18 @@ void copyColumnAttributesTo(@NotNull BaseColumnInfo to)
                 to.setKeyField(false);
         }
 
+        @Override
+        ColumnLogging getColumnLogging()
```
labkey-jeckels: Useless override

query/src/org/labkey/query/sql/QuerySelectView.java
```diff
@@ -356,31 +381,35 @@ private SQLFragment getSelectSQL(TableInfo table, @Nullable Collection<ColumnInf
...
-                if (null != columnLogging && columnLogging.shouldLogName())
+                shouldLogNameLoggings.add(columnLogging);
+                if (null == selectQueryAuditProvider)
+                    selectQueryAuditProvider = columnLogging.getSelectQueryAuditProvider();
+                if (null == columnLoggingComment)
+                    columnLoggingComment = columnLogging.getLoggingComment();
+                if (null != columnLogging.getException())
+                {
+                    UnauthorizedException uae = columnLogging.getException();
+                    // Use UnexpectedException to be consistent I guess??
```
labkey-jeckels, line 405: Use `UnexpectedException` or `UnauthorizedException`?

query/src/org/labkey/query/sql/QueryUnion.java
```diff
@@ -184,23 +255,99 @@ public void declareFields()
...
+        initColumnCalled = true;
+
+        // attach all sourceColumns to the corresponding UnionColumn
+        for (UnionColumn col : _unionColumns)
+            for (UnionTerm term : _termList)
+                col.addSourceColumn(term.columns.get( col._ordinal));
+    }
+
+
+    /* Hopefully we can evantually avoid doing all this work when column logging is not enabled, so do this as late as possible.
```
labkey-jeckels:
```suggestion
    /* Hopefully we can eventually avoid doing all this work when column logging is not enabled, so do this as late as possible.

```

query/src/org/labkey/query/sql/QueryUnion.java
```diff
@@ -184,23 +255,99 @@ public void declareFields()
...
+            }
+        }
+
+
+        // generate a ColumnLogging object for each UnionColumn
+        // step 1) For each source column create a ColumnLogging object that would
+        //  make sense in the context of the UNION query.
+        // step 2) for each output column derive a ColumnLogging from all its source columns
+
+        QueryTableInfo fakeTableInfo = new QueryTableInfo(this, "UNION")
```
labkey-jeckels, line 312: We also have one of these in QueryLookupWrapper. Worth sharing?

query/src/org/labkey/query/sql/QueryUnion.java
```diff
@@ -184,23 +255,99 @@ public void declareFields()
...
+                shouldLog |= source.columnLogging.shouldLogName();
+                sqap = null != sqap ? sqap : source.columnLogging.getSelectQueryAuditProvider();
+                ex = null != ex ? ex : source.columnLogging.getException();
+                if (null == ex)
+                    dataLoggingColumns.addAll(source.columnLogging.getDataLoggingColumns());
+            }
+            if (null != ex)
+                col._columnLogging = ColumnLogging.error(shouldLog, sqap, ex.getMessage());
+            else
+                col._columnLogging = new ColumnLogging(this.getSchema().getSchemaName(), "UNION", col.getFieldKey(), shouldLog, dataLoggingColumns, "TODO", sqap);
```
labkey-jeckels: `TODO`?

query/src/org/labkey/query/sql/QueryUnion.java
```diff
@@ -331,22 +482,53 @@ public SQLFragment getFromSQL(String alias)
...
+
+            ColumnLogging columnLogging = unionColumn.getColumnLogging();
+            if (null == columnLogging.getException())
+                columnLogging = new ColumnLogging(getSchema().getName(), getParentTable().getName(), getFieldKey(),
+                        columnLogging.shouldLogName(), columnLogging.getDataLoggingColumns(), columnLogging.getLoggingComment(), columnLogging.getSelectQueryAuditProvider());
+            setColumnLogging(columnLogging);
+        }
+
+        @Override
+        public ColumnLogging getColumnLogging()
```
labkey-jeckels: Useless override

api/src/org/labkey/api/collections/ArrayListMap.java
labkey-matthewb: We want the order of entrySet() and iterator() and values() to match.  The order we're trying to be consistent with is array index order.
This loop is iterating the keys, but they are not in any order, so we stash them in an array so we can match the keys to the values.  This does raise the idea of making the FindMap be a LinkedHashMap, then we could force the FindMap to iterate in a defined order.

api/src/org/labkey/api/data/ColumnLogging.java
labkey-matthewb: Thanks, that is now drek.  I tried to imitate the old behavior of identifying two columns as  being the "same" by comparing ColumnLogging objects, until I decided it was hopeless.  I'll track these down.

api/src/org/labkey/api/data/ColumnLogging.java
labkey-matthewb: I wanted to make sure that no one is using .equals() or .compareTo() since the old usage of these methods were all cases of bad logic.

query/src/org/labkey/query/sql/QuerySelectView.java
labkey-matthewb: We need  a separate design/PR to figure out the right error to throw.  I'm being consistent with the existing code in that problems with being able to do required audit logging throws Unauthorized.  I don't particularly like that choice.

query/src/org/labkey/query/sql/QueryUnion.java
labkey-matthewb: I thought about it.  Then I thought, it might be even better to get rid of the need for the fake table.  But this train ran out of track.

api/src/org/labkey/api/data/ColumnLogging.java
```diff
@@ -17,56 +17,122 @@
...
     public ColumnLogging reparent(TableInfo table)
     {
-        return new ColumnLogging(table.getSchema().getName(), table.getName(), _originalColumnFieldKey, _shouldLogName, _dataLoggingColumns, _loggingComment, _selectQueryAuditProvider);
+        String uniqueColumnKey = makeUniqueKey(table, _originalColumnFieldKey);
+        return new ColumnLogging(table.getUserSchema().getSchemaName(), table.getName(), _originalColumnFieldKey,
+//                uniqueColumnKey, _columnScopeKey,
+                _shouldLogName, _dataLoggingColumns, _loggingComment, _selectQueryAuditProvider);
     }
 
+    public ColumnLogging addScope(String scope)
```
labkey-adam: Rename since `scope` is unused?

api/src/org/labkey/api/data/ColumnLogging.java
```diff
@@ -117,16 +202,43 @@ public String getOriginalTableName()
...
+
 
     public SelectQueryAuditProvider getSelectQueryAuditProvider()
     {
         return _selectQueryAuditProvider;
     }
+
+
+    /* We need a way for Query to report an error in generating a ColumnLogging object.  It does not know ahead of
+     * time if column logging is enabled.  I don't particularly like this pattern.  I would be better to maybe
```
labkey-adam:
```suggestion
     * time if column logging is enabled.  I don't particularly like this pattern.  It would be better to maybe
```

api/src/org/labkey/api/data/ColumnLogging.java
```diff
@@ -117,16 +202,43 @@ public String getOriginalTableName()
...
+     * psas in a QueryLogging objet into query compile?
```
labkey-adam:
```suggestion
     * pass in a QueryLogging object into query compile?
```

labkey-adam (changes requested):

api/src/org/labkey/api/collections/ArrayListMap.java
labkey-matthewb: On next() and construct we skip ahead.

api/src/org/labkey/api/collections/ArrayListMap.java
labkey-matthewb: added a comment to consider using LinkedHashMap

api/src/org/labkey/api/collections/ArrayListMap.java
labkey-jeckels: Yes, I see that next() will skip blank entries. But what if only blank entries remain? hasNext() will return true even though it shouldn't, won't it?

labkey-matthewb: I'm not seeing it.  In this case next() will iterate to the end of the list and we'll be sitting at _i==_row.size() and return false.
When next() returns it has to be true that either a) we are on an entry or b) off the end.

query/src/org/labkey/query/sql/QueryUnion.java
labkey-matthewb: Thanks.  Also see 47399

api/src/org/labkey/api/collections/ArrayListMap.java
labkey-matthewb: Added assert

labkey-jeckels: Ah, my mistake. I wasn't catching that next() was advancing *after* it had pulled the element it will return. You're right, this should be fine.

api/src/org/labkey/api/data/ColumnLogging.java
labkey-jeckels: Makes sense for `equals()` Seems like the best way to make sure code doesn't try to compare us is to not implement Comparable and in the first place.

query/src/org/labkey/query/sql/QAggregate.java
labkey-jeckels:
```suggestion
        REGR_INTERCEPT(false, false)

```

query/src/org/labkey/query/sql/QueryLookupWrapper.java
labkey-jeckels: This a comment on the difference would still be nice

labkey-jeckels (approved):
@labkey-matthewb there are a couple of comments that you didn't respond to related to comments, etc. If you saw them and don't think it's worth adjusting, that's fine, but wanted to check they weren't missed.

api/src/org/labkey/api/data/ColumnLogging.java
labkey-matthewb: I see, missed that.

query/src/org/labkey/query/sql/QAggregate.java
labkey-matthewb: I think these should probably be (true, true), using the same logic as for SUM/MIN/MAX?

labkey-adam: Or (true, false). I guess you're arguing that the intercept could be an actual value in the single row case? I have a hard time seeing how a single floating point number, unconnected to any PTID, could ever be interpreted as PHI. This is very different to me than MIN/MAX which can operate on strings. But I don't think it matters much either way, so you could go with (true, true) to be conservative.

labkey-matthewb: I just did a test, and it requires two rows to return non-null value.  So I’m good leaving it as well.

labkey-jeckels: I don't log a comment, but I also thought about this and decided it was safe. Sure, it's possible that the calculated intercept happens to be exactly the same as `y` value where `x==0` but that seems like having a values that happens to be equal to the standard deviation - the same number, but not PHI

labkey-adam (approved):

PR #4114, opened 2023-02-14
Update merge participants page to also look for datasets by label

labkey-martyp:
#### Rationale
mergeParticipants.jsp queries study.Datasets for datasets to check for Id updates and then perform the update using the dataset names.  When checking for the Id, selectRows returns the dataset title/label, if defined, so need to identify the table by the label if not found by name.

#### Changes
* Get dataset label
* Identify dataset by title/label if not found by name

### DIFF
--- a/study/src/org/labkey/study/view/mergeParticipants.jsp
+++ b/study/src/org/labkey/study/view/mergeParticipants.jsp
@@ -199,7 +199,7 @@
         LABKEY.Query.selectRows({
             schemaName : 'study',
             queryName : 'Datasets',
-            columns: 'Name, DataSetId',
+            columns: 'Name, Label, DataSetId',
             sort: 'Name',
             success : function(details){
                 var rows = details.rows;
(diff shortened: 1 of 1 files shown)


study/src/org/labkey/study/view/mergeParticipants.jsp
```diff
@@ -310,6 +312,14 @@
                 success: function (data)
                 {
                     var table = this.tables[data.queryName];
+                    if (!table) {
+                        table = Object.values(this.tables).find(v => v.label === data.queryName);
+                    }
+                    if (!table) {
+                        console.error("Dataset: " + tableName + " not found.");
```
labkey-jeckels, line 319: Should we show this to the user (in an alert or writing to a `div`) or is it unlikely to happen?

labkey-jeckels (approved):

study/src/org/labkey/study/view/mergeParticipants.jsp
labkey-martyp: This would be a dataset whose name or title did not match what is listed in study.datasets which I think is very unlikely

PR #4116, opened 2023-02-14
Don't allow NULL values for pipeline.StatusFiles.FilePath

labkey-jeckels:
#### Rationale
We've never wanted to have null values in pipeline.StatusFile's FilePath column, but we've never explicitly prevented it. A recent support ticket had errors related to a null value. This won't prevent the root cause there, but it may make it easier to track down.

#### Changes
* Fail before attempting to update the value to null with other info about the row
* Make the column NOT NULL in the DB

### DIFF
--- a/pipeline/resources/schemas/dbscripts/sqlserver/pipeline-23.001-23.002.sql
+++ b/pipeline/resources/schemas/dbscripts/sqlserver/pipeline-23.001-23.002.sql
@@ -1,0 +1,8 @@
+DELETE FROM pipeline.StatusFiles WHERE FilePath IS NULL;
+
+-- Make pipeline.StatusFiles.FilePath NOT NULL, dropping and recreating unique constraint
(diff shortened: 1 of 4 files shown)


pipeline/resources/schemas/dbscripts/sqlserver/pipeline-23.001-23.002.sql
```diff
@@ -0,0 +1,8 @@
+DELETE FROM pipeline.StatusFiles WHERE FilePath IS NULL;
```
labkey-susanh, line 1: Is a GO needed here? (I never know...)

labkey-susanh (approved):

pipeline/resources/schemas/dbscripts/sqlserver/pipeline-23.001-23.002.sql
labkey-jeckels: Not this time. It can be required for DDL statements but not this particular sequence.

PR #4121, opened 2023-02-15
Add “ExpirationDate” field to exp.material

XingY:
#### Rationale
Support "expirationdate" for samples

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4121
* https://github.com/LabKey/sampleManagement/pull/1607
* https://github.com/LabKey/biologics/pull/1932
* https://github.com/LabKey/inventory/pull/737

#### Changes
* add exp.MaterialExpDate column
* fix getDatabaseCustomViews to exclude module reports
* refactor updateCustomViewFieldKeyChange out of CustomViewQueryChangeListener so it can be used for column rename as well


experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -427,6 +427,16 @@ public StringExpression getURL(ColumnInfo parent)
...
                 return wrapColumn(alias, _rootTable.getColumn("AliquotUnit"));
+            case MaterialExpDate:
+            {
+                var ret = wrapColumn(alias, _rootTable.getColumn("MaterialExpDate"));
+                ret.setLabel("Expiration Date");
+                ret.setHidden(false); // TODO: true until system fields can be disabled
+                ret.setShownInDetailsView(true);
+                ret.setShownInInsertView(true);
+                ret.setShownInUpdateView(true);
+                return ret;
```
cnathe, line 438: seems like an odd combination of setting these props in the exp.xml and here or in both. I'm not sure which is preferred but it seems like some of these "setShown..." are the defaults anyway. I agree with the call to setHidden being here once we allow for the enable/disable part.

cnathe (approved):

experiment/resources/schemas/exp.xml
```diff
@@ -323,6 +323,12 @@
           <shownInInsertView>false</shownInInsertView>
           <shownInUpdateView>false</shownInUpdateView>
       </column>
+      <column columnName="MaterialExpDate">
```
cnathe, line 326: do we want to add any `<description>` to this field?

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
XingY: I think they are for different tables. The exp.xml describes exp.material, but ExpMaterialTableImpl describes samples.sampletype.

cnathe: Oh yeah, good point. thanks

PR #4125, opened 2023-02-16
Update CPUTimer With More Functionality

labkey-danield:
Changed some variable values so tests can be run against db dump locally.
Check for an error banner while waiting for a grid to load. Replace reference to test sample type with Blood sample type. Updated CPUTimer to store time from first and last calls. Added comments to CPUTimer when getters return time in nanoTime and in milliseconds. Added a stop and save methods to CPUTimer that allow the timer to be stopped without saving data, then can call save to save data.

#### Rationale
Updated the CPUTimer to include the duration of the first and last start/stop calls.
Added a stopWithoutSave method that will stop the time but wont immediately save the information. This will allow the perf test to validate that an action being measured completed successfully. If successful the test can call the new "save" method to record the times. If not successful the test can discard the times.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/1613

#### Changes
* Added stopWithoutSave method.
* Added ability to record and get the duration of the first and last start/stop calls.
* Added clarifying comments identifying which getters return nanoTime duration and which ones return milliseconds.


api/src/org/labkey/api/util/CPUTimer.java
```diff
@@ -74,22 +75,59 @@ public boolean started()
         return _start > 0;
     }
 
+    /**
+     * Will stop the timer and save all accumulated data (duration, max, min etc...).
+     *
+     * @return Always returns true.
```
labkey-susanh, line 81: Is the return value necessary for in these methods that always return true?

api/src/org/labkey/api/util/CPUTimer.java
```diff
@@ -74,22 +75,59 @@ public boolean started()
...
+     */
+    public boolean save()
+    {
+        if (_stop > _start)
+        {
+            long elapsed = (_stop - _start);
+            _cumulative += elapsed;
+            _min = Math.min(_min, elapsed);
+            _max = Math.max(_max, elapsed);
+            _first = _first == 0 ? elapsed : _first;
```
labkey-susanh: This would be slightly more efficient if written as an if (`if _first == 0) _first = elapsed`), instead of always reassigning, though one hopes that compiler optimizations might do the right thing here.

labkey-susanh (approved):

api/src/org/labkey/api/util/CPUTimer.java
labkey-jeckels: This pattern can make them convenient to use in `assert` statements.

labkey-jeckels (approved):

api/src/org/labkey/api/util/CPUTimer.java
labkey-danield: I was following the existing pattern. Sounds like I will keep it for now.

api/src/org/labkey/api/util/CPUTimer.java
labkey-danield: Agreed, but I've updated it anyway.

PR #4129, opened 2023-02-19
Less restrictive projects

labkey-nicka:
#### Rationale
This incorporates updates to make product projects less restrictive in terms of how referencing data across projects. There are two new experimental flags that introduce two new behaviors:

1. EXPERIMENTAL_PRODUCT_ALL_FOLDER_LOOKUPS - Allow for lookup fields in product projects to query across all folders within the top-level folder.
2. EXPERIMENTAL_PRODUCT_PROJECT_DATA_LISTING_SCOPED - List only project-specific data within product projects.

#### Related Pull Requests
- https://github.com/LabKey/labkey-api-js/pull/145
- https://github.com/LabKey/labkey-ui-components/pull/1114
- https://github.com/LabKey/labkey-ui-premium/pull/47
- https://github.com/LabKey/biologics/pull/1942
- https://github.com/LabKey/sampleManagement/pull/1617
- https://github.com/LabKey/inventory/pull/742
- https://github.com/LabKey/recipe/pull/81
- https://github.com/LabKey/platform/pull/4129

#### Changes
- Add new `AllInProjectPlusShared` container filter type. Current project, Shared project, and all folders in the current (LKS) project.
- Add new `getDefaultLookupContainerFilter` to `UserSchema` to support resolving container filter for lookups against a schema.
- Add `getLookupContainerFilter` to `AbstractContainerFilterable` for resolving container filter for lookups on columns.
- Override container filters in `PdLookupForeignKey` when experimental all folder lookups are enabled.
- Remove duplicate logic to declare default container filter (generally "current"). Resolve from `UserSchema`.
- Declare experimental flags and utility methods in Query module.

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -1700,8 +1700,8 @@ public class Container implements Serializable, Comparable<Container>, Securable
         if (isProject())
             return new ContainerFilter.CurrentAndSubfoldersPlusShared(this, user);
 
-        if (getProject() != null && !QueryService.get().isProductProjectsAllFolderScopeEnabled())
-            return new ContainerFilter.CurrentPlusProjectAndShared(this, user);
+        if (getProject() != null && QueryService.get().isProductProjectsAllFolderScopeEnabled())
+            return new ContainerFilter.AllInProjectPlusShared(this, user);
 
         return ContainerFilter.current(this);
     }
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -1697,11 +1697,18 @@ public class Container implements Serializable, Comparable<Container>, Securable
         if (!isProductProjectsEnabled())
             return ContainerFilter.current(this);
 
+        if (QueryService.get().isProductProjectsAllFolderScopeEnabled())
+        {
+            if (isProject() || getProject() != null)
+                return new ContainerFilter.AllInProjectPlusShared(this, user);
+            return ContainerFilter.current(this);
+        }
+
         if (isProject())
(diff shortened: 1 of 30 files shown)


api/src/org/labkey/api/data/Container.java
```diff
@@ -1696,9 +1697,17 @@ public ContainerFilter getProductProjectsDataContainerFilter(User user)
         if (!isProductProjectsEnabled())
             return ContainerFilter.current(this);
 
+        if (QueryService.get().isProductProjectsAllFolderScopeEnabled())
+        {
+            if (isProject() || getProject() != null)
```
labkey-susanh, line 1702: Doesn't this mean that subfolders also get the `AllInProductPlusShared` filter.  I guess that's intentional for LKS and we change it on the client side for the listing pages.

api/src/org/labkey/api/data/ContainerFilter.java
```diff
@@ -1196,6 +1204,32 @@ public Type getType()
...
+    {
+        public AllInProjectPlusShared(Container c, User user)
+        {
+            super(c, user);
+        }
+
+        @Override
+        public Collection<GUID> generateIds(Container currentContainer, Class<? extends Permission> perm, Set<Role> roles)
+        {
+            Collection<GUID> containers = super.generateIds(currentContainer, perm, roles);
```
labkey-susanh, line 1217: `generateIds` seems an unfortunate name, but not something to change here.

labkey-susanh (approved):

api/src/org/labkey/api/data/Container.java
labkey-nicka: Yes, it does mean that and yes this is one place where we differentiate between container filter and container filter for listing data.

PR #4130, opened 2023-02-20
Move stored amount fields from inventory.item table to exp.materials table

labkey-susanh:
#### Rationale
We want to make it easier for users to track the amounts of samples, whether they are in storage or not. Currently, users define their own domain fields for tracking amounts, which doesn't allow us to use the application functionality for editing these amounts or compute aliquot amounts. When we begin supporting inventory item, it will also be more common that the samples would not be tracked in storage, but we still want to allow users to provide amounts and units in a way the application can easily interact with them. By moving the amount and units fields from the inventory.item table into the exp.materials table, these values can be tracked before, during and after the samples are in storage.

#### Related Pull Requests
- https://github.com/LabKey/inventory/pull/743

#### Changes
- Move data for inventory.items.volume and inventory.items.volumeUnits columns to exp.materials table
- Remove initialVolume, volume, and volumeUnits columns from inventory.items table
- Rename "Volume" and "VolumeUnits" to "StoredAmount" and "Units"
- Move classes and methods associated with measurements and amounts into api module in a new measurement package
- Move aliquot rollup methods into sample data iterators


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -277,6 +278,96 @@ protected ValidatorIterator getValidatorIterator(DataIterator validateInput, Dat
...
+                if (existingUnits == null)
+                    existingUnits = _sampleType.getMetricUnit();
+                Double newAmount = _storedAmountCol == null ? null : (Double) get(_storedAmountCol);
+                String newUnits = _unitsCol == null ? null : (String) get(_unitsCol);
+                if (newUnits == null)
+                    newUnits = _sampleType.getMetricUnit();
+                if (_aliquotedFromCol != null)
+                    aliquotParentLsids.add((String) get(_aliquotedFromCol));
+
+                if (!aliquotParentLsids.isEmpty() && (!Objects.equals(existingAmount, newAmount) || !Objects.equals(existingUnits, newUnits)))
```
XingY: This logic seems problematic. 

Existing record map might not always be present, for example, when import with merge from LKS, without detailed audit log.

For merge/update, if existing record is present, compare volume/unit. If existing record is not present (either merge a new record, or detailed audit is not on), then recalc needs to be set as long as the record is aliquot. This would include parents that are false positive for recalc, but better than missing recalc.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -277,6 +278,96 @@ protected ValidatorIterator getValidatorIterator(DataIterator validateInput, Dat
...
+                    aliquotParentLsids.add((String) existingMap.get(SampleUpdateAliquotedFromDataIterator.ALIQUOTED_FROM_LSID_COLUMN_NAME));
+                }
+                if (existingUnits == null)
+                    existingUnits = _sampleType.getMetricUnit();
+                Double newAmount = _storedAmountCol == null ? null : (Double) get(_storedAmountCol);
+                String newUnits = _unitsCol == null ? null : (String) get(_unitsCol);
+                if (newUnits == null)
+                    newUnits = _sampleType.getMetricUnit();
+                if (_aliquotedFromCol != null)
+                    aliquotParentLsids.add((String) get(_aliquotedFromCol));
```
XingY: For insert/merge, `get(_aliquotedFromCol))` returns parent name, not parent lsid.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -277,6 +278,96 @@ protected ValidatorIterator getValidatorIterator(DataIterator validateInput, Dat
...
+        {
+            return _storedAmountCol != null || _unitsCol != null;
+        }
+
+        @Override
+        public boolean next() throws BatchValidationException
+        {
+            boolean hasNext = super.next();
+            // skip processing if there are errors upstream
+            if (_context.getErrors().hasErrors() || !hasAmountData())
```
XingY: aliquot count change also triggers recalc, but I believe that's already handled in ExperimentSerivceImpl.saveExpMaterialAliquotOutputs so checking amount here is sufficient.

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: For update, get(aliquotedfromlsidcol) should return parent lsid.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2096,12 +2187,16 @@ else if (isMergeOrUpdate)
             // Wire up derived parent/child data and materials
             DataIteratorBuilder step6 = LoggingDataIterator.wrap(new ExpDataIterators.DerivationDataIteratorBuilder(step5, _container, _user, isSample, _dataTypeObject, false));
 
-            // Hack: add the alias and lsid values back into the input so we can process them in the chained data iterator
             DataIteratorBuilder step7 = step6;
+            if (isSample)
```
XingY: I believe this step can be skipped for INSERT. Whenever an new aliquot is created, it always triggers recal (handled by ExperimentServiceImpl.saveExpMaterialAliquotOutputs)

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -354,21 +354,25 @@ public List<Map<String, Object>> updateRows(User user, Container container, List
...
                     super._updateRowsUsingDIB(user, container, rows, getDataIteratorContext(errors, InsertOption.UPDATE, finalConfigParameters), extraScriptContext));
-
-            if (InventoryService.get() != null && results != null && results.size() > 0 && !errors.hasErrors())
-                InventoryService.get().recomputeSampleTypeRollup(_sampleType, container, false);
         }
         else
         {
             results = super.updateRows(user, container, rows, oldKeys, errors, configParameters, extraScriptContext);
 
+            if (!_sampleType.isMedia() && rows != null && (rows.get(0).containsKey("StoredAmount") || rows.get(0).containsKey("Amount") || rows.get(0).containsKey("Units")))
```
XingY: The row by row updateRows might have different set of fields in each row. The recal logic should be checked in each _update, where we have accurate new vs old information for each row.

XingY (changes requested):

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Logic updated to account for missing existing record.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Capturing either name or lsid values for aliquot parents now.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Thanks. Added condition on import type.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
labkey-susanh: Ugh. Well, I guess this is already slow. Moved this check into `_update`.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -277,6 +279,108 @@ protected ValidatorIterator getValidatorIterator(DataIterator validateInput, Dat
...
+                    if (aliquotedFrom != null)
+                        aliquotParentLsids.add(aliquotedFrom);
+                }
+                Measurement existingMeasurement = new Measurement(existingAmount, existingUnits, _sampleType.getMetricUnit());
+
+                Double newAmount = _storedAmountCol == null ? null : (Double) get(_storedAmountCol);
+                String newUnits = _unitsCol == null ? null : (String) get(_unitsCol);
+                Measurement newMeasurement = new Measurement(newAmount, newUnits, _sampleType.getMetricUnit());
+
+                if (_aliquotedFromCol != null && get(_aliquotedFromCol) != null)
```
XingY: aliquotedFromLsid should be preferred if updateOnly. The AliquotedFrom provided for updateOnly is ignored.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -277,6 +279,108 @@ protected ValidatorIterator getValidatorIterator(DataIterator validateInput, Dat
...
+
+        @Override
+        public boolean next() throws BatchValidationException
+        {
+            boolean hasNext = super.next();
+            // skip processing if there are errors upstream
+            if (_context.getErrors().hasErrors() || !hasAmountData())
+                return hasNext;
+
+            if (hasNext)
```
XingY, line 339: Should this be handled in a smaller batch size? Is there any danger with a too large set of updatedSampleLsids in memory?

api/src/org/labkey/api/data/measurement/AmountDataColumn.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @Override
+    public @NotNull HtmlString getFormattedHtml(RenderContext ctx)
+    {
+        Object convertedValue = convertToDisplayUnits(ctx);
+        return HtmlString.unsafe(convertedValue == null ? null : convertedValue.toString());
+    }
+
+    @Override
+    public Object getDisplayValue(RenderContext ctx)
```
XingY, line 110: When using display column, sort and filter wouldn't work correctly. Is there a plan to handle that?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -197,11 +200,11 @@ public int importRows(User user, Container container, DataIteratorBuilder rows,
         int ret = _importRowsUsingDIB(user, container, rows, null, getDataIteratorContext(errors, InsertOption.INSERT, configParameters), extraScriptContext);
         if (ret > 0 && !errors.hasErrors())
         {
-            if (InventoryService.get() != null)
+            if (!_sampleType.isMedia())
```
XingY, line 203: The only case that can bypass recalc for media sample type is updateRows. For insert/delete, the addition or removal of aliquots triggers aliquot created count recalc, see ExperimentServiceImpl.saveExpMaterialAliquotOutputs

experiment/resources/schemas/exp.xml
```diff
@@ -296,6 +296,27 @@
       <column columnName="SampleState">
           <description>Represents the status of the sample</description>
       </column>
+      <column columnName="StoredAmount">
+          <columnTitle>Amount</columnTitle>
+          <description>The amount of this sample currently on hand</description>
+          <displayColumnFactory>
+              <className>org.labkey.api.data.measurement.AmountDataColumn$Factory</className>
```
XingY, line 303: Do we have to use a display column? This would make sort/filtering not reliable. Since we are already doing so much manipulation of data during insert/update, can we just convert the amount and unit during insert/update, so the stored db value would match the expected display value?

XingY (changes requested):
Can we not use display column? Also, media sample type needs to recalc for aliquot created count.

experiment/resources/schemas/exp.xml
XingY: I guess the scenario that's messy is when sample type's unit gets changed. But then we can just update all amounts and units then?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -277,6 +279,108 @@ protected ValidatorIterator getValidatorIterator(DataIterator validateInput, Dat
...
+            {
+                Map<String, Object> existingMap = getExistingRecord();
+                Double existingAmount = null;
+                String existingUnits = null;
+                Set<String> aliquotParentLsids = new HashSet<>();
+                Set<String> aliquotParentNames = new HashSet<>();
+                if (existingMap != null)
+                {
+                    existingAmount = (Double) existingMap.get("StoredAmount");
```
XingY, line 348: It's likely that my story to support disabling system fields will get merged first. When you merge in those changes, you might need to mark storedamount and units as not disableble for now, and then have a follow up story to support allow disabling them.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
labkey-susanh: Ah, shoot. Yes. Gotta do this for media since we support aliquots for media now.

labkey-susanh: Ok, deferring this until we do the follow-on story for media where the amount fields are properly supported.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: I don't know that memory would be an issue, but the giant IN clause might be. I'll see about batching this up, but might defer it to a separate branch, depending on what else I need to fix.

api/src/org/labkey/api/data/measurement/AmountDataColumn.java
labkey-susanh: We have this problem already (a value of 100 uL will sort larger than 2 mL), so we aren't addressing it immediately.

experiment/resources/schemas/exp.xml
labkey-susanh: We chose to store the data as entered because of better data integrity, so I'll leave this as is for now. We can discuss further with @labkey-hannah when she's back.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Thanks for the reminder.

api/src/org/labkey/api/data/measurement/UnitsDataColumn.java
```diff
@@ -0,0 +1,127 @@
...
+        String displayUnitStr;
+        Object units = ctx.get(_unitsField);
+        if (units == null)
+            displayUnitStr = defaultUnitsStr;
+        else
+        {
+            try
+            {
+                // if units is compatible with default unit, use default units otherwise, use unit
+                Measurement.Unit storedUnit = Measurement.Unit.valueOf(units.toString());
```
XingY: units.toString().trim()?

experiment/resources/schemas/exp.xml
XingY: Is there any way for users to see the data as stored? When the fields were with inventory.items, users can go to items table to see the actual value, even though on samples grids they uses display column. How would users get this information with the changes to override display column at this level?

XingY: Also, not all sample types have type level units. LKS doesn't. Would the indiscriminate use of display column complicate certain scenarios unnecessarily?

XingY (approved):

api/src/org/labkey/api/data/measurement/UnitsDataColumn.java
labkey-susanh: Done. Thanks.

experiment/resources/schemas/exp.xml
labkey-susanh: I hope the display columns account for the absence of the display units well, but it is difficult (at least when testing) to not have the raw values available.  Perhaps it's worth adding separate columns (e.g., "RawAmount" and "RawUnits") that wrap the amount and units columns, respectively, and don't have a display column factory.  Brings in more columns that might conflict with existing domain columns, of course.

PR #4132, opened 2023-02-20
Delete Projects In App

RosalineP:
#### Rationale
It is desired that we surface a way to delete projects within the apps. We both provide an endpoint supplying a summary of to-be-deleted data for a user to review prior to project deletion, and add UI to support project deletion.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4132
* https://github.com/LabKey/labbook/pull/429
* https://github.com/LabKey/sampleManagement/pull/1619
* https://github.com/LabKey/labkey-api-js/pull/147
* https://github.com/LabKey/labkey-ui-components/pull/1122
* https://github.com/LabKey/labkey-ui-premium/pull/53
* https://github.com/LabKey/inventory/pull/750
* https://github.com/LabKey/biologics/pull/1958

#### Changes
* Supply optional ability to append a comment to project deletion audit log message
* Create getDetailedSummary()
* Create Summary obj
* Create GetModuleSummaryAction
* Implement getDetailedSummary() for:
    * CoreModule — Subfolder count
    * ExperimentModule — Number of data classes, individual data class row counts, number of sample types, number of sample rows, number of sample types, number of run groups, batch count, raw material count
    * ListModule — Number of picklists
* Fix bug in Data Class table where incorrect Data Counts were displayed due to containerFilter issue


core/src/org/labkey/core/CoreModule.java
```diff
@@ -1139,6 +1140,15 @@ public void startBackgroundThreads()
         TempTableTracker.init();
     }
 
+    @Override
+    public List<Summary> getDetailedSummary(Container c)
```
RosalineP, line 1144: Note that this will count the mildly hidden child containers that are created for notebook concepts. I suppose we want to filter these out—but how are they differentiated from regular containers?

RosalineP:
While setting up the frontend, I noticed some strings that need to be updated. I'll fix these up shortly.
```
Pick Lists -> Picklists
NucSequence -> Nucleotide sequences
ProtSequence -> Protein sequences
Ingredientss -> Ingredients (bad plural-ing)
Mixturess -> Mixtures (bad plural-ing)
```

api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -1726,7 +1726,7 @@ public static boolean isDeleting(final Container c)
     }
 
     // Delete a container from the database
-    public static boolean delete(final Container c, User user)
+    public static boolean delete(final Container c, User user, String comment)
```
labkey-nicka: Annotate `@Nullable String comment`. Same for `deleteAll()`.

api/src/org/labkey/api/module/MockModule.java
```diff
@@ -201,6 +202,12 @@ public Collection<String> getSummary(Container c)
         return Collections.emptyList();
     }
 
+    @Override
+    public ArrayList<Summary> getDetailedSummary(Container c)
+    {
+        return new ArrayList<>();
```
labkey-nicka: Consider `Collections.emptyList()`

api/src/org/labkey/api/module/Module.java
```diff
@@ -205,6 +205,12 @@ default void startBackgroundThreads()
      */
     @NotNull Collection<String> getSummary(Container c);
 
+    /**
+     * @param c container in which the items would be stored
+     * @return description of the objects that this module has stored in the container. More detailed than above
```
RosalineP: Considering editing this comment. 'Used for deletion summary in products' or such, with a matching 'Used for deletion summary in server' for getSummary().

api/src/org/labkey/api/module/Summary.java
```diff
@@ -0,0 +1,32 @@
+package org.labkey.api.module;
+
+import org.json.JSONObject;
+
+public class Summary
```
labkey-nicka, line 10: I think this is a good start on the abstraction, however, it feels almost _too abstract_. It's difficult to know what the purpose of this class is by looking at it. I would recommend adding a class description comment describing it's purpose.

api/src/org/labkey/api/module/Module.java
labkey-nicka: Good question. Our abstraction here is not good, however, the answer for now is to filter these containers for `container.isInFolderNav()`.

experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -704,6 +695,71 @@ public Collection<String> getSummary(Container c)
...
+        if (assayTypeCount > 0)
+            summaries.add(new Summary(assayTypeCount, "Assay Type"));
+
+        // Run count
+        int runGroupCount = ExperimentService.get().getExperiments(c, user, false, true).size();
+        if (runGroupCount > 0)
+            summaries.add(new Summary(runGroupCount, "Assay run"));
+
+        // Number of Data Classes
+        List<? extends ExpDataClass> dataClasses = ExperimentService.get().getDataClasses(c, user, true);
```
labkey-nicka: This should call `getDataClasses(c, user, false)` as we don't want to include definitions from other containers.

experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -704,6 +695,71 @@ public Collection<String> getSummary(Container c)
...
+        TableInfo dataClassesTable = ExpSchema.TableType.DataClasses.createTable(expSchema, null, ContainerFilter.Type.CurrentPlusProjectAndShared.create(c, user));
+        Map<String, Object> dataClassResults = new TableSelector(dataClassesTable, dataClassesTable.getColumns("Name,DataCount"), null, null).getValueMap();
+        for (String k : dataClassResults.keySet())
+        {
+            int count = ((Long) dataClassResults.get(k)).intValue();
+            if (count != 0)
+                summaries.add(new Summary(count, k));
+        }
+
+        // Batches
```
labkey-nicka: Batches and Raw Materials should not be explicitly processed by the Experiment module. I would recommend either:

1. Move the processing of these Sample Types to the Biologics module.
2. Do what we do for Data Classes and process the sample row counts with their associated sample type (as opposed to just grouping the counts under "Samples").

I think (2) is a better approach as it deals with all data classes and sample types uniformly and other modules don't have to care.

experiment/src/org/labkey/experiment/api/ExpDataClassTableImpl.java
```diff
@@ -144,7 +144,9 @@ public MutableColumnInfo createColumn(String alias, Column column)
             case DataCount:
             {
                 SQLFragment sql = new SQLFragment("(SELECT COUNT(*) FROM ").append(ExperimentServiceImpl.get().getTinfoData(), "d")
-                    .append(" WHERE d.classId = ").append(ExprColumn.STR_TABLE_ALIAS + ".rowid)");
+                    .append(" WHERE d.classId = ").append(ExprColumn.STR_TABLE_ALIAS + ".rowid")
```
labkey-nicka, line 147: Hmm, I'm surprised this wasn't already scoped to a container or by a container filter. That said, maybe instead of altering what this column does we should make this query ourselves independently for this feature and scope the container filter appropriately since I'm not sure these will ever align.

api/src/org/labkey/api/module/MockModule.java
RosalineP: The `List <Object>` type of `Collections.emptyList()` seems to conflict with the desired `ArrayList<Summary>` type?

api/src/org/labkey/api/module/Summary.java
RosalineP: Wrote one up! Hopefully that is more clear.

api/src/org/labkey/api/module/MockModule.java
labkey-nicka: Ah, yes. I don't think the specific implementation of List matters here. I would update the interface to be `List<Summary>` rather than `ArrayList<Summary>` and then all will be right in the world.

experiment/src/org/labkey/experiment/ExperimentModule.java
RosalineP: Went and implemented (2)!

experiment/src/org/labkey/experiment/api/ExpDataClassTableImpl.java
RosalineP: (Talked about over gChat, this fix will remain in place)

api/src/org/labkey/api/module/MockModule.java
RosalineP: Ahh, yes, aligned MockModule and DefaultModule!

api/src/org/labkey/api/module/Summary.java
```diff
@@ -0,0 +1,63 @@
...
+ * pluralization. Some given noun names are replaced by more reader-friendly versions.
+ */
+public class Summary
+{
+    private final int count;
+    private final String nounSingular;
+    private final String nounPlural;
+
+    private String replaceNounTitles(String nounSingular) {
+        String[] search = new String[] {
```
labkey-nicka, line 17: While I understand the need for this...it is really unfortunate. Having this class contain a private mapping of specific data structures defined in other modules makes for poor separation of concerns. This is probably fine to leave for now but we really need a way to pluralize things on the server. @labkey-susanh I know you did some work for the menu sections. Is there anything we could leverage that would know these mappings already?

experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -704,6 +695,69 @@ public Collection<String> getSummary(Container c)
...
+        UserSchema userSchema = QueryService.get().getUserSchema(user, c, SchemaKey.fromParts(ExpSchema.SCHEMA_NAME));
+        ExpSampleTypeTable sampleTypeTable = ExperimentService.get().createSampleTypeTable(ExpSchema.TableType.SampleSets.toString(), userSchema, ContainerFilter.Type.CurrentPlusProjectAndShared.create(c, user));
+        sampleTypeTable.populate();
+        Map<String, Object> tsSamplesResults = new TableSelector(sampleTypeTable, sampleTypeTable.getColumns("Name,SampleCount"), null, null).getValueMap();
+        for (String k : tsSamplesResults.keySet())
+        {
+            int count = ((Long) tsSamplesResults.get(k)).intValue();
+            if (count != 0)
+            {
+                Summary s = k.equals("MixtureBatches")
```
labkey-nicka, line 751: This is another spot where we're leaking abstraction.

RosalineP: Agreed—unfortunately this one fits particularly badly with the need for an accurate pluralization combined with need for name replacement.

api/src/org/labkey/api/module/Summary.java
labkey-susanh: You could possibly use the ProductRegistry to get the menu sections and then filter based on the key matching the name of the data class or sample type and pick off the label from the menu section.  For pluralizing you could use `StringUtilsLabKey.pluarlize`. It returns a string containing the number and the singular or singular + 's' noun. (Wouldn't handle MolecularSpecies well, of course).

RosalineP: Yeah, I did see that `pluralize` helper fn but it didn't have the ability to handle things like 'Data Class' -> 'Data Classes', and having an inaccurate plural seemed bad enough to justify the handling here—open to further thoughts on this though of course!
I'll take a look at that ProductRegistry idea, thanks!

labkey-nicka (approved):

PR #4133, opened 2023-02-20
Issue 47330: quotation marks in JMX-exposed objects breaks datadog's …

labkey-jeckels:
…jmxfetch

#### Rationale
Developers may include problematic characters in cache names and other objects that are exposed via JMX

#### Changes
* Replace potentially problematic characters

### DIFF
--- a/api/src/org/labkey/api/mbean/LabKeyManagement.java
+++ b/api/src/org/labkey/api/mbean/LabKeyManagement.java
@@ -16,6 +16,10 @@
 package org.labkey.api.mbean;
 
 import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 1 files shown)


labkey-adam:
This looks fine, but what's the downside of unconditionally calling `ObjectName.quote()`? It seems like that would make the strings legal while preserving their contents.

labkey-adam (approved):

labkey-jeckels:
> This looks fine, but what's the downside of unconditionally calling `ObjectName.quote()`? It seems like that would make the strings legal while preserving their contents.

That would be fine if I had a convenient way to test it, as I'm more confident that the underscores will work than the quoted values, since I can't find much in the way of what will be acceptable.

PR #4136, opened 2023-02-21
Add new moderator review option

vagisha:
#### Rationale
Add a new moderator review option that allows Authors to respond to a thread  but not start a new thread. This will be useful when we convert the Panorama Public support board to a secure message board.  On this message board submitters should be able to respond, without moderator review, to posts that are automatically added when they submit their data to Panorama Public.   They should not be able to start new threads, however. 

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3735


announcements/src/org/labkey/announcements/customize.jsp
```diff
@@ -90,7 +90,8 @@
         </td>
     </tr>
     <%
-        boolean moderatorReview = "InitialPost".equals(settings.getModeratorReview()) || "All".equals(settings.getModeratorReview());
+        boolean moderatorReview = "InitialPost".equals(settings.getModeratorReview()) || "All".equals(settings.getModeratorReview())
```
labkey-jeckels: This seems like a good time to switch these checks to use the ModeratorReview enum values instead of hard-coded strings

announcements/src/org/labkey/announcements/customize.jsp
```diff
@@ -104,6 +105,10 @@
                     <td valign="top"><input type="radio" name="moderatorReview" value="InitialPost"<%=checked("InitialPost".equals(settings.getModeratorReview()))%>></td>
                     <td><b>Initial Post</b> - Authors must have their initial message(s) approved by a moderator. Subsequent messages are approved automatically.</td>
                 </tr>
+                <tr>
+                    <td valign="top"><input type="radio" name="moderatorReview" value="NewThread"<%=checked("NewThread".equals(settings.getModeratorReview()))%>></td>
```
labkey-jeckels: Same thing here and the other, preexisting checks. Something like:

```suggestion
                    <td valign="top"><input type="radio" name="moderatorReview" value="<%= h(ModeratorReview.NewThread.toString())"<%=checked(ModeratorReview.NewThread.toString().equals(settings.getModeratorReview()))%>></td>

```

labkey-jeckels (changes requested):
One set of requests for JSP improvements. If you can make the updates in your FB I will mirror them into LK's repo to let TeamCity run the tests.

labkey-jeckels:
I've mirrored the feature branches and queued the updated test:

https://teamcity.labkey.org/buildConfiguration/LabKey_2211Release_Community_DailySuites_DailyCPostgres?branch=vs_secure-message-new-option&buildTypeTab=overview&mode=builds

labkey-jeckels (commented):
@vagisha can you merge the latest changes from the LabKey release22.11-SNAPSHOT branch? There's a commit that's not yet in your branch that's causing a cross-repo build break

labkey-jeckels (approved):

PR #4139, opened 2023-02-22
Allow passing DataIteratorContext to AbstractQueryImport.importData

labkey-martyp:
#### Rationale
Provide a version of AbstractQueryImportAction.importData that takes a DataIteratorContext parameter instead of a handful of the parameters that are saved in the context. This is needed in EHR to pass in more options for the import.

A next iteration on this would probably be to factor out usages of the other importData functions to use a DataIteratorContext, but that is out of scope for this budget/feature.

#### Related Pull Requests
* https://github.com/LabKey/ehrModules/pull/549

#### Changes
* Factor the current implementation to take a DataIteratorContext parameter
* Move the current creation of DataIteratorContext to another importData version

### DIFF
--- a/api/src/org/labkey/api/dataiterator/DataIteratorContext.java
+++ b/api/src/org/labkey/api/dataiterator/DataIteratorContext.java
@@ -50,7 +50,7 @@ public class DataIteratorContext
     boolean _allowImportLookupByAlternateKey = false;
     private final Set<String> _passThroughBuiltInColumnNames = new CaseInsensitiveHashSet();
     private final Set<String> _dontUpdateColumnNames = new CaseInsensitiveHashSet();
-    private final Set<String> _alternateKeys = new CaseInsensitiveHashSet();
+    private Set<String> _alternateKeys = new CaseInsensitiveHashSet();
     private String _dataSource;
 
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/dataiterator/DataIteratorContext.java
```diff
@@ -180,6 +180,11 @@ public Set<String> getAlternateKeys()
         return _alternateKeys;
     }
 
+    public void setAlternateKeys(Set<String> alternateKeys)
```
labkey-jeckels: This isn't necessary - for better or worse we already have other places that mutate the set returned from `getAlternateKeys()`

https://github.com/LabKey/dataintegration/blob/aa361f61460d802bd5255a22173623c634737b36/src/org/labkey/di/pipeline/TransformTask.java#L149

labkey-jeckels (approved):
One idea for simplifying the change.

api/src/org/labkey/api/dataiterator/DataIteratorContext.java
labkey-martyp: Ah got it. Backed out that change.

PR #4143, opened 2023-02-22
Support Enable/Disable “System Default Fields”

XingY:
#### Rationale
Add support to enable/disable non domain table columns for tables that consist of a materialized table and a domain table.  

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1120
- https://github.com/LabKey/platform/pull/4143
- https://github.com/LabKey/sampleManagement/pull/1630
- https://github.com/LabKey/biologics/pull/1953

#### Changes
* add exp.DomainDescriptor.SystemFieldConfig column
* modify domain related beans to support disabledSystemFields
* handle reading/setting disabledSystemFields from systemFieldConfig
* hide samples and dataclass columns based on disabledSystemFields


api/src/org/labkey/api/exp/property/DomainKind.java
```diff
@@ -312,6 +314,32 @@ public boolean matchesTemplateXML(String templateName, DomainTemplateType templa
...
+        disallowedFields = new CaseInsensitiveHashSet(disallowedFields);
+
+        List<String> allowedDisabledFields = new ArrayList<>();
+        for (String field : disabledSystemFields)
+        {
+            if (!disallowedFields.contains(field))
+                allowedDisabledFields.add(field);
+        }
+
+        return allowedDisabledFields;
```
labkey-susanh: Could do this with something like `disabledSystemFields.stream().filter(f -> !disallowedFields.contains(f))`;

I guess this is being defensive, in case the config got saved somehow with some fields that can't be disabled, but I hope we're being defensive when we save so we toss out the ones that shouldn't be disabled.

api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -618,6 +621,8 @@ public static ValidationException updateDomainDescriptor(GWTDomain<? extends GWT
         if (updateDomainName)
             d.setName(update.getName());
 
+        d.setDisabledSystemFields(update.getDisabledSystemFields());
```
labkey-susanh: Should this be checking for fields that might not be allowed to be disabled?

experiment/src/org/labkey/experiment/api/ExpTableImpl.java
```diff
@@ -85,6 +88,25 @@ protected ExpTableImpl(String name, TableInfo rootTable, UserSchema schema, Cont
...
+    public Set<String> getDisabledSystemFields()
+    {
+        if (_disabledSystemFields == null)
+        {
+            Domain domain = getDomain();
+            if (domain != null)
+            {
+                _disabledSystemFields = new CaseInsensitiveHashSet();
+
+                List<String> disabledFields = domain.getDisabledSystemFields();
```
labkey-susanh, line 102: Would it be easier if getDisabledSystemFields always returned a non-null object?

experiment/src/org/labkey/experiment/api/ExpTableImpl.java
```diff
@@ -85,6 +88,25 @@ protected ExpTableImpl(String name, TableInfo rootTable, UserSchema schema, Cont
...
+                if (disabledFields != null && !disabledFields.isEmpty())
```
labkey-susanh: The `!disabledFields.isEmpty()` can be removed

experiment/src/org/labkey/experiment/api/property/DomainImpl.java
```diff
@@ -1391,6 +1395,66 @@ public boolean isProvisioned()
...
+            configJson = new JSONObject();
+        }
+        else
+            configJson = new JSONObject(systemFieldConfigStr);
+
+        JSONArray existingDisabled = configJson.optJSONArray(DISABLED_SYSTEM_FIELDS_KEY);
+        if (emptyFields && existingDisabled == null)
+            return;
+
+        JSONArray fields = new JSONArray(disabledSystemFields);
```
labkey-susanh, line 1452: Does this work ok if `disabledSystemFields` is null?

query/src/org/labkey/query/controllers/GetQueryDetailsAction.java
```diff
@@ -291,6 +291,9 @@ public ApiResponse execute(Form form, BindException errors)
 
             if (!tinfo.getAltKeysForUpdate().isEmpty())
                 resp.put("altUpdateKeys", tinfo.getAltKeysForUpdate());
+
+            if (tinfo.getDisabledSystemFields() != null && !tinfo.getDisabledSystemFields().isEmpty())
```
labkey-susanh: Could save constructing the disabledSystemFields twice by stashing in a variable before the if statement.

labkey-susanh (approved):

PR #4145, opened 2023-02-22
Issue 45875: Filters on multiple grids on a page are removed when customizing one of the grids

labkey-bpatel:
#### Rationale
Issue [45875](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45875): Filters on multiple grids on a page are removed when customizing one of the grids

#### Changes
* Only handle/skip filtering on the current grid that is being customized when multiple grids are present on the page

### DIFF
--- a/api/webapp/clientapi/dom/DataRegion.js
+++ b/api/webapp/clientapi/dom/DataRegion.js
@@ -3163,7 +3163,8 @@ if (!LABKEY.DataRegions) {
 
                                     // Special prefix that should remove all filters, but no other parameters for the current grid
                                     if (skipPrefix.indexOf(ALL_FILTERS_SKIP_PREFIX) === (skipPrefix.length - 2)) {
-                                        if (key.indexOf(region.name) >= 0 && key.indexOf('~') > 0) {
+                                        if (key.indexOf(region.name + '.') == 0 && key.indexOf('~') > 0) {
+
                                             stop = true;
                                             return false;
                                         }
--- a/api/webapp/clientapi/dom/DataRegion.js
+++ b/api/webapp/clientapi/dom/DataRegion.js
@@ -3163,7 +3163,7 @@ if (!LABKEY.DataRegions) {
 
                                     // Special prefix that should remove all filters, but no other parameters for the current grid
                                     if (skipPrefix.indexOf(ALL_FILTERS_SKIP_PREFIX) === (skipPrefix.length - 2)) {
-                                        if (key.indexOf(region.name + '.') == 0 && key.indexOf('~') > 0) {
+                                        if (key.indexOf(region.name + '.') === 0 && key.indexOf('~') > 0) {
 
                                             stop = true;
                                             return false;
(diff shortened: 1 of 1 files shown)


api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -3161,9 +3161,9 @@ if (!LABKEY.DataRegions) {
                             $.each(skipPrefixSet, function (j, skipPrefix) {
                                 if (LABKEY.Utils.isString(skipPrefix)) {
 
-                                    // Special prefix that should remove all filters, but no other parameters
+                                    // Special prefix that should remove all filters, but no other parameters for the current grid
                                     if (skipPrefix.indexOf(ALL_FILTERS_SKIP_PREFIX) === (skipPrefix.length - 2)) {
-                                        if (key.indexOf('~') > 0) {
+                                        if (key.indexOf(region.name) >= 0 && key.indexOf('~') > 0) {
```
labkey-jeckels: This looks like a good approach, but if we had `query` and `query2` on the same page, it would still get confused, or `mylist` and `list`

```suggestion
                                        if (key.indexOf(region.name + '.') == 0 && key.indexOf('~') > 0) {

```

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -3176,11 +3176,12 @@ if (!LABKEY.DataRegions) {
                                             }
                                         }
                                         else if (key.toLowerCase().indexOf(skipPrefix.toLowerCase()) === 0) {
-                                            // only skip filters, parameters, and sorts
-                                            if (key === skipPrefix ||
+                                            // only skip filters, parameters, and sorts for the current grid
+                                            if (key.indexOf(region.name) >= 0 &&
```
labkey-jeckels:
```suggestion
                                            if (key.indexOf(region.name + '.') == 0 &&

```

labkey-jeckels (approved):
Can you test the suggested changes here? If they work for you, go ahead and merge

api/webapp/clientapi/dom/DataRegion.js
labkey-bpatel: I have multiple lists and queries on the page with titles 'query' or 'list' and it does not get confused since region.name is unique within a given page. For example, for the same list displayed twice on the page with title 'My List', region.name is list4 and list5, and for Query wp, it doesn't use title either, region.name is qwp7 or qwp8, for example.

labkey-jeckels: Yes, but I bet that if you manually set [dataRegionName](https://www.labkey.org/download/clientapi_docs/javascript-api/symbols/LABKEY.QueryWebPart.html#constructor) on two different QueryWebParts on the same page to `list` and `mylist`, for example, they'd get confused.

labkey-bpatel: Got it! Thanks for clarifying

PR #4146, opened 2023-02-22
Fix occasional "Expected a QueryDefinition object" error with temp queries

labkey-jeckels:
#### Rationale
I occasionally see "Expected a QueryDefinition object" errors when loading pages with multiple grids backed by QueryWebParts that pass SQL instead of referencing a saved query. There's a race condition with setting up the session-backed storage of the temp queries

#### Changes
* Synchronize around the critical section for creating the session query storage
* Migrate more calls to use SchemaKey instead of String for schema names


labkey-adam (approved):
Yay schema name -> SchemaKey

PR #4151, opened 2023-02-24
Issue 47356: Rewrite returnURL in the event of a list name change

labkey-nicka:
#### Rationale
This addresses [Issue 47356](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47356) by rewriting the returnURL parameters in the event of a name change to a list.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1395

#### Changes
* Rewrite the returnURL parameters in the event of a list name change.
* Defer processing of a returnURL until after we've resolved whether an explicit returnURL was provided on the URL.

### DIFF
--- a/core/src/client/ListDesigner/ListDesigner.tsx
+++ b/core/src/client/ListDesigner/ListDesigner.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import React, { ReactNode } from 'react';
-import { ActionURL, getServerContext, PermissionTypes, Security } from '@labkey/api';
+import { ActionURL, PermissionTypes, Security } from '@labkey/api';
 import {
     Alert,
     LoadingSpinner,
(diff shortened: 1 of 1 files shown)


XingY (commented):
Instead of rewriting return url, should be fix be to fix the list detail url so it uses listid instead of name as param? That's how we handle sample type and data class designers.

labkey-nicka:
> Instead of rewriting return url, should be fix be to fix the list detail url so it uses listid instead of name as param? That's how we handle sample type and data class designers.

I'm not sure which view "list detail" is referring to, however, most list views already support finding lists by "name" or by "listId". The reason this distinction exists for lists is that listId's are unique to a folder. So you could have a list in the top-level folder with ID = 1 and a list in a subfolder with ID = 1. If you were coming from a cross-folder context we would still need to inspect the returnURL to determine which folder it is and then resolve the listId, otherwise, we might return the user to the wrong list. Name, however, is supported cross-folder hence the approach I've taken.

An overall architecture change we could consider is enumerating listIds as unique across the system (like we do for almost all other data structures) as opposed to unique within a container.

XingY:
> > Instead of rewriting return url, should be fix be to fix the list detail url so it uses listid instead of name as param? That's how we handle sample type and data class designers.
> 
> I'm not sure which view "list detail" is referring to, however, most list views already support finding lists by "name" or by "listId". The reason this distinction exists for lists is that listId's are unique to a folder. So you could have a list in the top-level folder with ID = 1 and a list in a subfolder with ID = 1. If you were coming from a cross-folder context we would still need to inspect the returnURL to determine which folder it is and then resolve the listId, otherwise, we might return the user to the wrong list. Name, however, is supported cross-folder hence the approach I've taken.
> 
> An overall architecture change we could consider is enumerating listIds as unique across the system (like we do for almost all other data structures) as opposed to unique within a container.

I see... I didn't realize listId is not unique globaly.

XingY (approved):

PR #4110, opened 2023-02-11
Issue 47195: Enterprise pipeline temp directories are not always cleaned up

Sigmonia:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=47195

#### Related Pull Requests
* https://github.com/LabKey/MacCossLabModules/pull/311

#### Changes
* Removed cleanup call from done method
* Marked enterprise pipeline jobs submitted
* Adjusted LocalDirectory constructors

PR #4152, opened 2023-02-27
Fix ETL failure running with enterprise pipeline introduced by fix 47195

labkey-jeckels:
#### Rationale
The fix for Issue 47195: Enterprise pipeline temp directories are not always cleaned up, https://github.com/LabKey/platform/pull/4110, introduced a failure when running ETLs with the enterprise pipeline configured

#### Changes
* If we don't have a task, we're done and know we haven't been cancelled


labkey-jeckels:
@bbimber FYI

Sigmonia (approved):

PR #4154, opened 2023-02-28
Issue 47404: Sample Manager: Property field containing '${}' should be disallowed

XingY:
#### Rationale
When sample property field contains '${' substring, the column caption is not treated as a ConstantStringExpression, resulting in exception:
https://www.labkey.org/_mothership/mothership-showStackTraceDetail.view?exceptionStackTraceId=32066

#### Related Pull Requests
* 

#### Changes
* better handling of DisplayColumn.getCaption to avoid NPE
* disallow domain field containing ${} pattern

### DIFF
--- a/api/src/org/labkey/api/util/StringExpressionFactory.java
+++ b/api/src/org/labkey/api/util/StringExpressionFactory.java
@@ -53,6 +53,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static org.labkey.api.data.AbstractTableInfo.LINK_DISABLER;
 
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/util/StringExpressionFactory.java
```diff
@@ -76,6 +78,8 @@
 
     public static final StringExpression EMPTY_STRING = new ConstantStringExpression("");
 
+    public static final String SUBSTITUTION_EXP_PREFIX = ".*\\$\\{.+}.*";
```
labkey-susanh, line 81: Do you not need to escape the ending brace? Also, a nit, but this pattern is not just the prefix, is it?

labkey-susanh (approved):

api/src/org/labkey/api/util/StringExpressionFactory.java
XingY: Ending brace doesn't need encoding. This pattern detects substring, not just prefix.

PR #4159, opened 2023-03-01
Issue 47202: Options to reduce payload of getContainers.api response

labkey-jeckels:
#### Rationale
Editing a lookup's target container on a server with tens of thousands of containers can crash or bog down the browser as it gets a huge response of mostly unneeded data

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/4159
- https://github.com/LabKey/labkey-api-js/pull/149
- https://github.com/LabKey/labkey-ui-components/pull/1129
- https://github.com/LabKey/sampleManagement/pull/1646
- https://github.com/LabKey/inventory/pull/756
- https://github.com/LabKey/biologics/pull/1973

#### Changes
* New optional parameters that filter out workbooks and unneeded properties when simply building the container hierarchy
* update @labkey/api and @labkey/components package version in core/package.json

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -1367,13 +1367,13 @@ public class Container implements Serializable, Comparable<Container>, Securable
         return toJSON(user, true, true);
     }
 
-    public Map<String, Object> toJSON(User user, boolean includePermissions, boolean includeAssortedProps)
+    public Map<String, Object> toJSON(User user, boolean includePermissions, boolean includeStandardProps)
     {
         Map<String, Object> containerProps = new HashMap<>();
         Container parent = getParent();
         containerProps.put("name", getName());
         containerProps.put("path", getPath());
-        if (includeAssortedProps)
+        if (includeStandardProps)
         {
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/data/Container.java
```diff
@@ -1364,44 +1364,59 @@ private Container findDescendant(Container parent, String name)
 
     public Map<String, Object> toJSON(User user)
     {
-        return toJSON(user, true);
+        return toJSON(user, true, true);
     }
 
-    public Map<String, Object> toJSON(User user, boolean includePermissions)
+    public Map<String, Object> toJSON(User user, boolean includePermissions, boolean includeAssortedProps)
```
cnathe: minor: to match the form prop, should this be called `includeStandardProperties` instead of `includeAssortedProps`?

cnathe (approved):

api/src/org/labkey/api/data/Container.java
labkey-jeckels: Yes. I pushed an update to make it consistent.

PR #4162, opened 2023-03-01
47355: Query metadata not persisting to XML correctly

labkey-klum:
#### Rationale
The UI metadata editor doesn't always translate `ColumnInfo` properties to `MetadataColumnJSON` properly, in this case not preserving the name of the wrapped column. The reason we see the error only after the user saves some seemingly unrelated XML metadata via the source editor is because the UI metadata editor only applies the last metadata when there are multiples.

The process that is followed is:
- Table columns are converted to their `MetadataColumnJSON` form (of course all applicable metadata has been applied to the table).
- The most recent XML metadata is parsed and any columns in that metadata are converted to their `MetadataColumnJSON` form and replaces those fields taken from step 1.
- The list of fields are returned to the client for rendering.

The related bug involved a table that had `.query.xml` furnished which created the wrapped MVFK columns. Due to the aforementioned bug, we were relying on the `.query.xml` metadata to serialize those fields correctly and once this was no longer parsed, it couldn't rely on `ColumnInfo` serialization to work.

[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47355)


labkey-matthewb (approved):

labkey-matthewb (commented):
instanceof seems to always cause future bugs.  A long term solution might be to add getWrappedColumnName() to ColumnInfo or ColumnRenderProperties

labkey-klum:
> 



> 

I like that idea better, since there are implementations of a wrapped column where we don't hold onto the column being wrapped like : `WrappedColumnInfo.wrapAsCopy`

labkey-klum:
> 

The general fix does result in more fields annotated with the `wrappedColumnName` this is probably accurate but I'll probably push that change into 23.4 to get a little more testing time.

labkey-matthewb:
> > 
> 
> The general fix does result in more fields annotated with the `wrappedColumnName` this is probably accurate but I'll probably push that change into 23.4 to get a little more testing time.

That seems like a good option.

PR #4165, opened 2023-03-01
Avoid pulling in display columns for "requiredColumns"

labkey-matthewb:
#### Rationale
E.g. Don't Pull in Container/DisplayName when Container is not explicitly added to the DataRegion.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/data/DataRegion.java
+++ b/api/src/org/labkey/api/data/DataRegion.java
@@ -3051,21 +3051,25 @@ public class DataRegion extends DisplayElement
             {
                 var dr = new DataRegion();
                 dr.setColumns(List.of(table.getColumn("Title"), table.getColumn("Container")));
-                Results rs = dr.getResults(new RenderContext(context));
-                assertEquals(4, rs.getFieldMap().size());
-                assertTrue(rs.getFieldMap().containsKey(FieldKey.fromParts("Title")));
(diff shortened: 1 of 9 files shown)


api/src/org/labkey/api/data/TableSelector.java
```diff
@@ -342,12 +343,29 @@ public Results getResultsAsync(final boolean cache, final boolean scrollable, Ht
         return asyncRequest.waitForResult(() -> getResults(cache, scrollable));
     }
 
+    /**
+     * Setting this options asks the TableSelector add additional display columns to the generated SQL, as well
```
labkey-susanh:
```suggestion
     * Setting this options asks the TableSelector to add additional display columns to the generated SQL, as well

```

api/src/org/labkey/api/data/DataRegion.java
```diff
@@ -3030,4 +3034,39 @@ protected boolean useTableWrap()
     {
         return true;
     }
+
+
+
+    public static class TestCase extends Assert
```
labkey-susanh, line 3040: Did this test class get registered as a junit test?

labkey-susanh (approved):

api/src/org/labkey/api/data/DataRegion.java
labkey-matthewb: Thanks, don’t know where that went…

PR #4170, opened 2023-03-03
fix for MS2 nested dataregion

labkey-matthewb:
#### Rationale
Undo change to RenderContext.addDisplayColumns().  Restore behavior that MS2 nested data regions depends on.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/data/DataRegion.java
+++ b/api/src/org/labkey/api/data/DataRegion.java
@@ -348,8 +348,15 @@ public class DataRegion extends DisplayElement
 
     public void setDisplayColumns(List<DisplayColumn> displayColumns)
     {
-        clearColumns();
-        displayColumns.forEach(this::addDisplayColumn);
+        /** NOTE - the cleaner thing to do here would be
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/data/DataRegion.java
```diff
@@ -348,8 +348,15 @@ public List<String> getDisplayColumnNames()
 
     public void setDisplayColumns(List<DisplayColumn> displayColumns)
     {
-        clearColumns();
-        displayColumns.forEach(this::addDisplayColumn);
+        /** NOTE - the cleaner thing to do here would be
+         *         clearColumns();
+         *         displayColumns.forEach(this::addDisplayColumn);
+         * however, this breaks MS2 which seems to funny things with nested RenderContexts
```
labkey-susanh:
```suggestion
         * however, this breaks MS2 which seems to do funny things with nested RenderContexts
```

labkey-susanh (approved):

PR #4175, opened 2023-03-04
Issue 47440: LKSM: withCounter jumping

XingY:
#### Rationale
WithCounter name expression uses DBSequence to generate a next sequence number for a particular prefix string in a specific container. It's reported that users are seeing generated count using withCounter name expression would jump. This is mostly due to DBSequence CUD intentionally uses its own transaction to simulate database sequence behavior, and partly due to the cached preallocates remembering the last used count.

This PR introduce an experimental flag, under which, DB sequence used for withCounter can be reclaimed in the event of transaction rollback. ReclaimablePreallocate preallocate sequences and can be rolled back when no competing transaction is also using the sequence. ReclaimableDBSequence will be used if strict no gap is desired, which will be used if the withCoutner syntax has a 3rd parameter to opt in for this behavior

#### Related Pull Requests
* 

#### Changes
* Added "Use strict incremental withCounter expression" experimental flag
* Introduce ReclaimablePreallocate and ReclaimableDBSequence and related util to allow rolling back of DB sequence updates
* Switch WithCounter syntax to use Reclaimable when experimental flag is enabled
* Explicitly sync withCounter sequences at the end of name generation task and remove them from cache
* Note that Create/Read of ReclaimablePreallocate continues to use its own transaction because those are mostly harmless


labkey-matthewb:
There are a lot of issues to consider with this which makes me hesitant to consider this for 23.3 without some testing under load/contention.

- This will effectively serialize insert into tables that use these counters
- That may be OK, but we would probably want to grab locks on involved counters at the beginning of the main transaction to reduce deadlocks
-  We locate sequences with one filter/index (container, name, id) and update using another (container, rowid).  I wonder if that increases the chance of deadlock scenarios, potentially with unrelated sequences?
- Deadlocks on unrelated sequences might be more of a problem on SQL Server (just guessing).  We might want to consider aiming to load the DbSequence objects using the name index outside the transaction, and only using the rowid index in the transaction.
-  Unrelated, why do we use container in the update statement?

labkey-jeckels:
I also worry about deadlocks, especially on SQLServer. Do we have a stress test that covers this scenario? Or could be adapted to cover it? @labkey-martyp @labkey-danield https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46808

XingY:
> There are a lot of issues to consider with this which makes me hesitant to consider this for 23.3 without some testing under load/contention.
> 
> * This will effectively serialize insert into tables that use these counters
> * That may be OK, but we would probably want to grab locks on involved counters at the beginning of the main transaction to reduce deadlocks
> * We locate sequences with one filter/index (container, name, id) and update using another (container, rowid).  I wonder if that increases the chance of deadlock scenarios, potentially with unrelated sequences?
> * Deadlocks on unrelated sequences might be more of a problem on SQL Server (just guessing).  We might want to consider aiming to load the DbSequence objects using the name index outside the transaction, and only using the rowid index in the transaction.
> * Unrelated, why do we use container in the update statement?

Thanks Matt. 

I agree this is too much for 23.3. The other option is to have this change go in 23.3 under an experimental flag, which would be used by a few hosted postgres users only. We can then test and evaluate this change in develop. 

- Not sure what it would take to grab lock at the beginning of the main transaction... The update/insert sqls has XLOCK and ROWLOCK applied to them. Also, all mutating actions are synchronized. 
- Do you mean to use (container, name, id) for UPDATE when useCurrentTransaction == true?  
- I agree container seems not needed in Update since PK is rowId

XingY:
> I also worry about deadlocks, especially on SQLServer. Do we have a stress test that covers this scenario? Or could be adapted to cover it? @labkey-martyp @labkey-danield https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46808

Good pointer. We plan to make available of this feature in 23.3 under an experimental flag only. We should definitely add scenarios in stress test in develop.

labkey-martyp:
> I also worry about deadlocks, especially on SQLServer. Do we have a stress test that covers this scenario? Or could be adapted to cover it? @labkey-martyp @labkey-danield https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46808

ExperimentStressTest would be a good one to use. Right now it's failing due to a different sql server deadlock issue. I should be able to get back to that this week.

XingY:
> > I also worry about deadlocks, especially on SQLServer. Do we have a stress test that covers this scenario? Or could be adapted to cover it? @labkey-martyp @labkey-danield https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46808
> 
> ExperimentStressTest would be a good one to use. Right now it's failing due to a different sql server deadlock issue. I should be able to get back to that this week.

Thanks Marty. FYI, I'm separately working on deferring aliquot rollup compute. Now that inventory tables no longer triggers recompute, I might as well reenable this test in my other branch. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47246

labkey-martyp:
> > > I also worry about deadlocks, especially on SQLServer. Do we have a stress test that covers this scenario? Or could be adapted to cover it? @labkey-martyp @labkey-danield https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46808
> > 
> > 
> > ExperimentStressTest would be a good one to use. Right now it's failing due to a different sql server deadlock issue. I should be able to get back to that this week.
> 
> Thanks Marty. FYI, I'm separately working on deferring aliquot rollup compute. Now that inventory tables no longer triggers recompute, I might as well reenable this test in my other branch. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47246

Oh great, I didn't realize that work was being done. Yeah that RecomputeRollup column is a pretty hot column during import, glad to hear you're deferring the rollup.

api/src/org/labkey/api/data/DbSequenceManager.java
```diff
@@ -148,7 +178,18 @@ private static int ensure(Container c, String name, int id)
         getRowIdSql.add(name);
         getRowIdSql.add(id);
 
-        return executeAndMaybeReturnInteger(tinfo, getRowIdSql);
+        Integer rowId = executeAndMaybeReturnInteger(tinfo, getRowIdSql);
+
```
labkey-matthewb, line 179: check null==rowId

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1316,6 +1335,22 @@ public String generateName(@NotNull State state, @NotNull Map<String, Object> ro
         return state.nextName(rowMap, parentDatas, parentSamples, extraPropsFns, altExpression);
     }
 
+    private static void cleanSeq(DbSequence seq, Container container)
+    {
+        if (seq instanceof DbSequence.ReclaimablePreallocate)
```
labkey-matthewb: Instance of should be pushed into the method on the individual classes or to DbSequenceManager.  The user of DbSequencerManager shouldn't be need to help with the implementation differences.

labkey-matthewb:
I'm having a hard time convincing myself this is correct.  I believe that in order to be correct, the code depends on only one transaction can be using a particular reclaimable db sequence at a time.  This _might_ be true on postgres due to the lock collected in getRowId(withUpdateLock=true).  However, the use of a map by getReclaimablePreallocateSequence() makes it possible (certainly with sqlserver), that the same object may be returned to two callers.  When the sequence is removed from the map by invalidateReclaimablePreallocateSequence(), this does not do anything about instances of this object that may still be held by other threads.  

The very simplest change I can think of to make this work would be to use a regular DbSequence (w/o preallocation) with the useCurrentTransaction() idea.  To do both transaction and preallocation and useCurrentTransaction() might be possible, but will require some tricky synchronization between the in-memory sequence objects and the sql transactions.

XingY:
> I'm having a hard time convincing myself this is correct. I believe that in order to be correct, the code depends on only one transaction can be using a particular reclaimable db sequence at a time. This _might_ be true on postgres due to the lock collected in getRowId(withUpdateLock=true). However, the use of a map by getReclaimablePreallocateSequence() makes it possible (certainly with sqlserver), that the same object may be returned to two callers. When the sequence is removed from the map by invalidateReclaimablePreallocateSequence(), this does not do anything about instances of this object that may still be held by other threads.
> 
> The very simplest change I can think of to make this work would be to use a regular DbSequence (w/o preallocation) with the useCurrentTransaction() idea. To do both transaction and preallocation and useCurrentTransaction() might be possible, but will require some tricky synchronization between the in-memory sequence objects and the sql transactions.

The ReclaimablePreallocateSequence is a half way solution that allows rollback of sequences without race condition. The newly introduced 3rd param to withCounter force ReclaimablePreallocateSequence to only reserve 1 sequence, effectively turning it into a non preallocate. For some clients / scenarios, the continuity are more important than performance, so the 3rd param to the name expression will allow them to opt in for correctness over performance. But the default syntax would allow better performance and correctness in a non race condition scenario: if a single tread has to rollback, but no competing thread is using the same sequence, then no gap will be produced. 

I worry the use of DBSeq without preallocate would be too slow. It's used by aliquot generation which is known to be produced at large quantities. So I feel a compromised solution might be better.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -496,6 +510,11 @@ static Pair<List<String>, List<String>> validateWithCounterSyntax(String nameExp
                 if (format.charAt(0) != '\'' || format.charAt(format.length()-1) != '\'')
                     errorMessages.add(String.format("Format string starting at position %d for 'withCounter' substitution pattern should be enclosed in single quotes.", commaIndex + 1));
             }
+            if (!StringUtils.isEmpty(noGapBool))
+            {
+                if (!("true".equals(noGapBool) || "'true'".equals(noGapBool) || "false".equals(noGapBool) || "'false'".equals(noGapBool)))
```
labkey-matthewb: I don't like the pattern of using "bare" boolean values because it makes future changes or generalization of the syntax difficult.  Maybe the use of named options would be better.

labkey-matthewb (changes requested):
I think this all looks good except I would not use getReclaimablePreallocateSequence() in NameGenerator for 23.3.   I don't think the in-memory object will stay synced with the database in all scenarios.

XingY:
> I think this all looks good except I would not use getReclaimablePreallocateSequence() in NameGenerator for 23.3. I don't think the in-memory object will stay synced with the database in all scenarios.

Do you think that we might be able to get ReclaimablePreallocateSequence 100% correct in develop? If so, then sure we can use the non preallocates in 23.3 and safely switch later. But if not, I worry if we make DBSequence without preallocate the default in 23.3, then we won't able to default to the faster ReclaimablePreallocate later, because we don't want clients to see behavior regression.

labkey-matthewb:
> > I think this all looks good except I would not use getReclaimablePreallocateSequence() in NameGenerator for 23.3. I don't think the in-memory object will stay synced with the database in all scenarios.
> 
> Do you think that we might be able to get ReclaimablePreallocateSequence 100% correct in develop? If so, then sure we can use the non preallocates in 23.3 and safely switch later. But if not, I worry if we make DBSequence without preallocate the default in 23.3, then we won't able to default to the faster ReclaimablePreallocate later, because we don't want clients to see behavior regression.

That is a really good question…. I _think_ we can.  DbSequence works because the database is “in charge” and it always asks the database what the next value is.  Preallocate is a bit of a hybrid, but you can think of it as being “in charge”, because sync() is telling the database what the next value is.  Transacted Preallocate is tricky because we want to be able to sync() (tell the database) and also handle rollback (probably ask the database).  I think we can make this work by using DbScope.addCommitTask().

XingY:
> > > I think this all looks good except I would not use getReclaimablePreallocateSequence() in NameGenerator for 23.3. I don't think the in-memory object will stay synced with the database in all scenarios.
> > 
> > 
> > Do you think that we might be able to get ReclaimablePreallocateSequence 100% correct in develop? If so, then sure we can use the non preallocates in 23.3 and safely switch later. But if not, I worry if we make DBSequence without preallocate the default in 23.3, then we won't able to default to the faster ReclaimablePreallocate later, because we don't want clients to see behavior regression.
> 
> That is a really good question…. I _think_ we can. DbSequence works because the database is “in charge” and it always asks the database what the next value is. Preallocate is a bit of a hybrid, but you can think of it as being “in charge”, because sync() is telling the database what the next value is. Transacted Preallocate is tricky because we want to be able to sync() (tell the database) and also handle rollback (probably ask the database). I think we can make this work by using DbScope.addCommitTask().

Sounds good. changes made to not use preallocates in 23.3

labkey-matthewb (approved):

PR #4177, opened 2023-03-06
Issue 47434: APIKey improvements

labkey-jeckels:
#### Rationale
Users reported headaches with the pipe character in the key when passing values through environment variables. We can omit the prefix altogether, and make a few small incremental improvements.

#### Changes
* Drop the `apikey|` and `session|` prefixes from API keys
* Switch to SHA256 for hashing
* Update comments based on longer-term plans


api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -599,14 +599,15 @@ else if ("true".equalsIgnoreCase(request.getHeader("LabKey-Disallow-Global-Roles
     {
         String apiKey;
 
-        // Prefer Basic auth
+        // Accept API keys via basic auth if the username is "apikey". Supported for backwards compatibility.
```
labkey-adam:
```suggestion
        // Accept API keys via basic auth if the username is "apikey". Supported for backwards compatibility and clients that don't support custom headers.

```

labkey-adam (approved):

labkey-adam:
@labkey-jeckels we'll need to adjust `CasTicketManager` which extends `SessionKeyManager` (and is currently not compiling). Let me know if you want me to look at fixing it.

PR #4179, opened 2023-03-07
Issue 47433

XingY:
#### Rationale
[Issue 47246](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47246): Can recomputing of inventory item rollup be deferred
Issue 47433: Don't log ExperimentException to mothership when unable to set genId to a given value

#### Related Pull Requests
* 

#### Changes
* use a background thread to recompute sample rollup as a POSTCOMIMT task
* add aliquot generation to stress test and re-enable stress test

### DIFF
--- a/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
+++ b/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
@@ -7430,39 +7430,47 @@ public class ExperimentController extends SpringActionController
         }
 
         @Override
-        public Object execute(GenIdForm form, BindException errors) throws Exception
+        public Object execute(GenIdForm form, BindException errors)
         {
             ApiSimpleResponse resp = new ApiSimpleResponse();
             resp.put("success", true);
 
-            if (SampleTypeDomainKind.NAME.equalsIgnoreCase(form.getKindName()))
+            try
             {
-                if (!getContainer().hasPermission(getUser(), DesignSampleTypePermission.class))
-                    throw new UnauthorizedException("Insufficient permissions.");
-
-
-                ExpSampleType sampleType = SampleTypeService.get().getSampleType(form.getRowId());
-                if (sampleType != null)
-                    sampleType.ensureMinGenId(form.getGenId());
-                else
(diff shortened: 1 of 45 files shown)


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -7430,39 +7430,47 @@ public void validateForm(GenIdForm form, Errors errors)
...
 
-            if (SampleTypeDomainKind.NAME.equalsIgnoreCase(form.getKindName()))
+            try
             {
-                if (!getContainer().hasPermission(getUser(), DesignSampleTypePermission.class))
-                    throw new UnauthorizedException("Insufficient permissions.");
+                if (SampleTypeDomainKind.NAME.equalsIgnoreCase(form.getKindName()))
+                {
+                    if (!getContainer().hasPermission(getUser(), DesignSampleTypePermission.class))
+                        throw new UnauthorizedException("Insufficient permissions.");
```
labkey-nicka, line 7443: nit: I don't think this message needs to be provided. Can just `throw new UnauthorizedException()` and it will get our standard handling for messaging.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -7430,39 +7430,47 @@ public void validateForm(GenIdForm form, Errors errors)
...
-                if (sampleType != null)
-                    sampleType.ensureMinGenId(form.getGenId());
-                else
+                    ExpSampleType sampleType = SampleTypeService.get().getSampleType(form.getRowId());
+                    if (sampleType != null)
+                        sampleType.ensureMinGenId(form.getGenId());
+                    else
+                    {
+                        resp.put("success", false);
+                        resp.put("error", "Sample type does not exist.");
```
labkey-nicka, line 7452: nit: We have the `errors` collection. Could that be used instead so this action doesn't have one-off error handling? I realize this was here before just something to consider.

labkey-nicka (approved):

PR #4183, opened 2023-03-08
ExpMaterialTableImpl optimization to skip the INNER JOIN to the provisioned table when no columns from that domain are selected

cnathe:
#### Rationale
While triaging the app grid loading times for a sample type with 500k rows, I noticed that most of the time was being spent on two queries: one to get the total row count of the sample type and the other to get the selected keys for the query / user session. When profiling both of these queries, I found that they were both doing an INNER JOIN from the exp.materials table to the provisioned table for the given sample type. In both of these cases the INNER JOIN wasn't needed as no columns were being selected from that provisioned table. This PR uses the `selectedColumns` param of `getFromSQL` method to determine if that join is needed or can be skipped.

#### Changes
* add a `checkSelectedColumnsFromRootOnly` method to ExpMaterialTableImpl that when true can be used to skip the INNER JOIN for the generated SQL

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
@@ -1097,7 +1097,7 @@ public class ExpMaterialTableImpl extends ExpRunItemTableImpl<ExpMaterialTable.C
         return sql;
     }
 
-    // if all selectedColumns are in the root table, no need to join to the provisioned table (i.e. example is the COUNT(*) case)
+    // if all selectedColumns are in the root table, no need to join to the provisioned table (e.g., in the COUNT(*) case)
     private boolean checkSelectedColumnsFromRootOnly(Set<FieldKey> selectedColumns, Set<String> rootColNames)
     {
         boolean selectedRootOnly = true;
--- a/experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
@@ -1026,10 +1026,12 @@ public class ExpMaterialTableImpl extends ExpRunItemTableImpl<ExpMaterialTable.C
     {
         TableInfo provisioned = null == _ss ? null : _ss.getTinfo();
 
-        selectedColumns = computeInnerSelectedColumns(selectedColumns);
-
         // all columns from exp.material except lsid
         Set<String> dataCols = new CaseInsensitiveHashSet(_rootTable.getColumnNameSet());
+
+        boolean skipProvisionedJoin = checkSelectedColumnsFromRootOnly(selectedColumns, dataCols);
+
+        selectedColumns = computeInnerSelectedColumns(selectedColumns);
 
         SQLFragment sql = new SQLFragment();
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -1026,11 +1026,13 @@ public SQLFragment getFromSQL(String alias, Set<FieldKey> selectedColumns)
...
         TableInfo provisioned = null == _ss ? null : _ss.getTinfo();
 
-        selectedColumns = computeInnerSelectedColumns(selectedColumns);
-
         // all columns from exp.material except lsid
         Set<String> dataCols = new CaseInsensitiveHashSet(_rootTable.getColumnNameSet());
 
+        boolean skipProvisionedJoin = checkSelectedColumnsFromRootOnly(selectedColumns, dataCols);
+
+        selectedColumns = computeInnerSelectedColumns(selectedColumns);
```
labkey-susanh, line 1034: Can this be skipped if `skipProvisionedJoin` is true?

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -1095,6 +1097,32 @@ public SQLFragment getFromSQL(String alias, Set<FieldKey> selectedColumns)
         return sql;
     }
 
+    // if all selectedColumns are in the root table, no need to join to the provisioned table (i.e. example is the COUNT(*) case)
```
```suggestion
    // if all selectedColumns are in the root table, no need to join to the provisioned table (e.g., in the COUNT(*) case)

```

labkey-susanh (approved):

labkey-matthewb:
Can you drop in an example of the before and after generated SQL?  For say the count(*) query I’m curious if it makes sense to drop the materialized table instead of the provisioned table.  It seems simpler/faster to count all the rows from a table than a filtered subset of rows from a very large table.  Of course, it may require even more refactoring to accomplish that.

cnathe:
> Can you drop in an example of the before and after generated SQL? For say the count(*) query I’m curious if it makes sense to drop the materialized table instead of the provisioned table. It seems simpler/faster to count all the rows from a table than a filtered subset of rows from a very large table. Of course, it may require even more refactoring to accomplish that.

I agree that it would likely be quicker to count the materialized table rows rather then the filtered set of rows from exp.materials. Attached are the before and after SQL from the changes in this PR. From my local testing on the sample type with 500k rows, the count(*) query was taking approx 15 sec before and now takes <1 sec.

[getSelected_after.txt](https://github.com/LabKey/platform/files/10922901/getSelected_after.txt)
[getSelected_before.txt](https://github.com/LabKey/platform/files/10922902/getSelected_before.txt)
[totalRowsCount_after.txt](https://github.com/LabKey/platform/files/10922903/totalRowsCount_after.txt)
[totalRowsCount_before.txt](https://github.com/LabKey/platform/files/10922904/totalRowsCount_before.txt)

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
cnathe: done

labkey-matthewb (approved):
This is a great step.

PR #4184, opened 2023-03-08
Fix setRecomputeFlagForSampleNames and QueryWebPart test

labkey-susanh:
#### Rationale
- `setRecomputeFlagForSampleNames` was looking for LSIDs not names.
- `columns` parameter for `LABKEY.QueryWebPart` should be a comma-separated string not an array

#### Related Pull Requests
* #4130 

#### Changes
* Fixes for the above.

### DIFF
--- a/experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
@@ -1402,7 +1402,7 @@ public class SampleTypeServiceImpl extends AbstractAuditHandler implements Sampl
     {
         DbScope scope = ExperimentService.get().getSchema().getScope();
         TableInfo materialTable = ExperimentService.get().getTinfoMaterial();
-        String updateSqlStr = "UPDATE " + materialTable.getSelectName() + " SET RecomputeRollup = ? WHERE LSID AND RecomputeRollup = ?";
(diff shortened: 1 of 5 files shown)


XingY (approved):

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1416,7 +1416,7 @@ public void setRecomputeFlagForSampleNames(ExpSampleType sampleType, Set<String>
     {
         DbScope scope = ExperimentService.get().getSchema().getScope();
         TableInfo materialTable = ExperimentService.get().getTinfoMaterial();
-        String updateSqlStr = "UPDATE " + materialTable.getSelectName() + " SET RecomputeRollup = ? WHERE CpasType = ? AND LSID ";
+        String updateSqlStr = "UPDATE " + materialTable.getSelectName() + " SET RecomputeRollup = ? WHERE CpasType = ? AND Name ";
```
XingY: Should we add a where condition that RecomputeRollup = false, to narrow down the set of rows to update? It might help performance and reduce the number of rows that gets locked by this transaction

labkey-tchad (approved):

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Added, though I'm unclear how much effect it will have since I'm not sure how often it will get set to true twice before being recomputed, but every little bit helps.

PR #4185, opened 2023-03-08
TableSelector getAggregates optimization to only use PK columns when requesting COUNT(*) aggregate

cnathe:
#### Rationale
In the related PR, we made an optimization so that when generating the select SQL for a sample type, we skip the INNER JOIN to the materialized table when no columns from that table were being selected. This PR make the same type of change for Data Class tables. 

In addition, it makes a change to the TableSelector getAggregates() method so that if just the COUNT(*) aggregate is being requested, which happens for every DataRegion in LKS, we only use the PK columns for the given table. This allows for that aggregate query to make use of the optimizations made for the sample type and data class tables as mentioned above.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4183

#### Changes
* Apply the checkSelectedColumnsFromRootOnly() check to Data Classes as well as Sample Types
* TableSelector getAggregates() method update to only use PK columns when getting just the COUNT(*) aggregate


api/src/org/labkey/api/data/TableSelector.java
```diff
@@ -465,7 +465,12 @@ public boolean exists()
     // TODO: Convert to return Map<FieldKey, List<Aggregate.Result>>
     public Map<String, List<Result>> getAggregates(final List<Aggregate> aggregates)
     {
-        final AggregateSqlFactory sqlFactory = new AggregateSqlFactory(_filter, aggregates, _columns);
+        // If we are only asking for the COUNT(*) aggregate, then we only need the PK columns.
+        // This can make a big performance difference for Sample Type and Data Class tables as they can then skip
+        // the join between the exp schema base table and the materialized table for the given table.
+        Collection<ColumnInfo> aggColumns = aggregates.size() == 1 && aggregates.get(0).isCountStar() ? _table.getPkColumns() : _columns;
```
labkey-matthewb: Need to check that getPkColumns is non-empty().

labkey-matthewb (changes requested):

api/src/org/labkey/api/data/TableSelector.java
labkey-matthewb: It's also possible to do add an expr column like "new ExprColumn(table, "~one~", new SqlFragment("1"))

experiment/src/org/labkey/experiment/api/ExpRunItemTableImpl.java
```diff
@@ -281,4 +281,29 @@ public void setDefaultVisibleColumns(@Nullable Iterable<FieldKey> list)
         super.setDefaultVisibleColumns(allowed);
     }
 
+    // if all selectedColumns are in the root table, no need to join to the provisioned table (e.g., in the COUNT(*) case)
+    protected boolean checkSelectedColumnsFromRootOnly(Set<FieldKey> selectedColumns, Set<String> rootColNames)
```
labkey-susanh: I might change the name of this method to more clearly indicate what the true or false value it returns indicates. Something like `containsOnlyRootColumns`.

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -682,17 +684,20 @@ public SQLFragment getFromSQL(String alias)
...
+            }
         }
         sql.append(" FROM ");
         sql.append(_rootTable, "d");
-        sql.append(" INNER JOIN ").append(provisioned, "p").append(" ON d.lsid = p.lsid) ");
+        if (!skipProvisionedJoin)
+            sql.append(" INNER JOIN ").append(provisioned, "p").append(" ON d.lsid = p.lsid");
         String subAlias = alias + "_dc_sub";
-        sql.append(subAlias);
+        sql.append(") ").append(subAlias);
```
labkey-susanh, line 706: Was this closing paren always missing? I'm not seeing where the open paren got added.

experiment/src/org/labkey/experiment/api/ExpRunItemTableImpl.java
cnathe: Good call. Changed.

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
cnathe: Nope, the closing paren was always there. It was split out from end of the previous line 693.

api/src/org/labkey/api/data/TableSelector.java
cnathe: Done. Thanks for the suggestion!

labkey-susanh (approved):

labkey-matthewb:
Note for future possible changes:  We might also want to detect case when only provisioned columns are present.  It would be great to make the common lookup to sample type usages also avoid this join.  (with special case handling for LSID which can come from either table).

labkey-matthewb (approved):

PR #4186, opened 2023-03-08
QueryService.getSelectBuilder()

labkey-matthewb:
#### Rationale
There are a lot of getSelectSql() overloads in QueryService.  The SelectBuilder allows us to make a more sane API, but maybe a bit more verbose.  We can also roll up some of the selector() and select() methods.  Lastly, this builder will propagate QueryLogging between getSelectSql() and the Selector for execution helping with the usage pattern there.

#### Related Pull Requests
* 

#### Changes
*


api/src/org/labkey/api/dataiterator/QueryDataIteratorBuilder.java
```diff
@@ -39,6 +39,8 @@
 import java.util.List;
 import java.util.Map;
 
+import static org.labkey.api.data.DataRegion.MessagePart.filter;
```
labkey-nicka, line 42: Unnecessary import.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1810,7 +1810,10 @@ else if (value.size() == 0)
                             Collection<FieldKey> fields = Collections.singleton(relativeFieldKey);
                             Map<FieldKey, ColumnInfo> cols = QueryService.get().getColumns(lookupTable, fields);
 
-                            try (Results results = QueryService.get().select(lookupTable, cols.values(), filter, null))
+                            var select = QueryService.get().getSelectBuilder(lookupTable)
```
labkey-nicka, line 1869: I'm definitely in favor of dropping the need to pass in `null` for sort almost everywhere.

api/src/org/labkey/api/query/QueryService.java
```diff
@@ -229,38 +232,41 @@ UserSchema createLinkedSchema(User user, Container container, String name, Strin
      * @param schema The query schema context used to parse the sql query in.
      * @param sql The LabKey query string.
      * @return a TableSelector
+     *
+     * superseded by getSelectBuilder()
```
labkey-nicka: nit: Consider `superseded by {@link QueryService#getSelectBuilder}` here an in similar comments.

labkey-nicka (commented):
I like it!

labkey-adam (changes requested):
Is there a compelling reason to hang `getSelectBuilder()` off QueryService? Wouldn't it be more convenient to allow creating a new `SelectBuilder` via `SelectBuilder.get(TableInfo)`?

api/src/org/labkey/api/query/QueryService.java
labkey-matthewb: Done

labkey-matthewb:
> Is there a compelling reason to hang `getSelectBuilder()` off QueryService? Wouldn't it be more convenient to allow creating a new `SelectBuilder` via `SelectBuilder.get(TableInfo)`?

That pattern would be nice.  However, getSelectBuilder(schema,sql) makes use of internal query.main classes.  So, adding these to a service interface makes sense, in that case at least.

labkey-adam (approved):

PR #4188, opened 2023-03-09
Issue 47246: improve aliquot rollup recompute

### LINKED ISSUE #4190: Issue 47246: Can recomputing of inventory item rollup be deferred?
[Issue 47246](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47246)

- [x] support recompute in background thread
- [x] skip intermediate step to set recompute flag
- [x] get rid of recompute column
- [x] code review
- [x] manual testing @labkey-danield 
- [x] tc review and merge

XingY:
#### Rationale
The calculation of aliquot rollup is a tedious process. In the dataiterators, we mark a parent/root as need recal by updating the flag to true. Then at the end of the data iterator, we do the calculation and flip the flag back. The intermediate updates increases the chance of race condition. This PR explores an alternative approach that no longer uses the needrecalc flag DB column. 

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4188
* https://github.com/LabKey/biologics/pull/2000
* https://github.com/LabKey/sampleManagement/pull/1675
* https://github.com/LabKey/inventory/pull/773

#### Changes
* drop exp.material.RecomputeRollup column
* repurpose AliquotRollupDataIterator to determine parents to recompute and hold on to the value in memory, rather than setting the flag value
* no longer temporarily update recomputeflag to true, instead, handle update of rollup values at the end of dataiterator
* if the set of samples to recompute is large, defer to an after commit background job
* if the set of samples to recompute < 20, do it in the same transaction so that app users will get refreshed grid data immediately


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -122,6 +120,8 @@
 import static org.labkey.api.data.CompareType.IN;
 import static org.labkey.api.exp.api.ExperimentService.ALIASCOLUMNALIAS;
 import static org.labkey.api.exp.query.ExpMaterialTable.Column.RootMaterialLSID;
+import static org.labkey.experiment.api.SampleTypeUpdateServiceDI.PARENT_RECOMPUTE_LSID_COL;
+import static org.labkey.experiment.api.SampleTypeUpdateServiceDI.PARENT_RECOMPUTE_NANE_COL;
```
labkey-susanh:
```suggestion
import static org.labkey.experiment.api.SampleTypeUpdateServiceDI.PARENT_RECOMPUTE_NAME_COL;
```

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -312,76 +310,87 @@ public static class AliquotRollupDataIterator extends WrapperDataIterator
...
-            if (_context.getErrors().hasErrors() || !hasAliquotData() || _context.getConfigParameterBoolean(SampleTypeService.ConfigParameters.DeferAliquotRuns))
-                return hasNext;
-
-            if (hasNext)
+            if (i == _parentLsidToRecomputeCol || i == _parentNameToRecomputeCol)
             {
+                if (_isInsert)
+                {
+                    if (i == _parentNameToRecomputeCol && _aliquotedFromCol != null)
+                        return get(_aliquotedFromCol); // recompute parent when new aliquot is created
```
labkey-susanh, line 342: Is this meant to be checking the parent or the root?

OK, I see below that it really is parent.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -312,76 +310,87 @@ public static class AliquotRollupDataIterator extends WrapperDataIterator
...
+                    Double newAmount = _storedAmountCol == null ? null : (Double) get(_storedAmountCol);
+                    String newUnits = _unitsCol == null ? null : (String) get(_unitsCol);
 
-                // check for aliquotedFromLsidCol first since, for updateOnly cases, we prefer data in this
-                // column over aliquotedFrom; we ignore any user-supplied value in aliquotedFrom
-                if (_aliquotedFromLsidCol != null && get(_aliquotedFromLsidCol) != null)
-                    aliquotParentLsids.add((String) get(_aliquotedFromLsidCol));
-                else if (_aliquotedFromCol != null && get(_aliquotedFromCol) != null && amountMayHaveChanged)
-                    updatedSampleNames.add(String.valueOf(get(_aliquotedFromCol)));
+                    boolean amountChanged = !(Objects.equals(existingAmount, newAmount) && Objects.equals(existingUnits, newUnits));
```
labkey-susanh, line 369: This doesn't account for changes that represent the same amount (3 L and 3 mL).  It's perhaps not a  likely kind of change and a tradeoff between efficiency of recomputing more than necessary vs. doing more work for equality checking.  I'm guessing the former is usually more expensive than the latter, but there's also a question of when that time is better spent, of course.

We could add a non-normalized equality check in the Measurement class in order to encapsulate this logic better.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -312,76 +310,87 @@ public static class AliquotRollupDataIterator extends WrapperDataIterator
...
+                    if (!amountChanged && (_storedAmountCol == null || _unitsCol == null))
+                    {
+                        if (_storedAmountCol == null && !existingUnits.equals(newUnits))
```
labkey-susanh: Seems like `existingUnits` could be null here.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -312,76 +310,87 @@ public static class AliquotRollupDataIterator extends WrapperDataIterator
...
+                            amountChanged = true;
+                        if (_unitsCol == null && !existingAmount.equals(newAmount))
```
labkey-susanh: Same thing about null check here.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -6944,14 +6941,14 @@ else if (rec._action.getActionSequence() == SIMPLE_PROTOCOL_OUTPUT_STEP_SEQUENCE
 
         private void saveExpMaterialAliquotOutputs(List<ProtocolAppRecord> protAppRecords) throws ValidationException
         {
-            Set<String> parentLsids = new HashSet<>();
+//            Set<String> parentLsids = new HashSet<>();
```
labkey-susanh: If this gets restored, let's please rename this to `rootLsids`. The use of `parents` is confusing.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -197,25 +188,104 @@ protected DataIteratorBuilder preTriggerDataIterator(DataIteratorBuilder in, Dat
...
+    {
+        if (outputRows == null || outputRows.isEmpty())
+            return null;
+
+        Set<String> parentLsids = new HashSet<>();
+        Set<String> parentNames = new HashSet<>();
+        if (outputRows.size() == 1 && outputRows.get(0).containsKey(PARENT_RECOMPUTE_LSID_SET))
+        {
+            parentLsids.addAll((Collection<? extends String>) outputRows.get(0).get(PARENT_RECOMPUTE_LSID_SET));
+            parentNames.addAll((Collection<? extends String>) outputRows.get(0).get(PARENT_RECOMPUTE_NAME_SET));
```
labkey-susanh: This `get` might return null. Perhaps move the `containsKey` checks before each of these `addAll` calls.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -298,19 +359,7 @@ public int mergeRows(User user, Container container, DataIteratorBuilder rows, B
...
-                    SampleTypeService.get().recomputeSampleTypeRollup(_sampleType, container, false);
-                }
-                catch (SQLException e)
-                {
-                    throw new RuntimeException(e);
-                }
-            }
-
-            onSamplesChanged();
+            onSamplesChanged(null); // mergeRows not really used, skip wiring recalc
```
labkey-susanh: Not used because it's not possible, even via api calls, or just not common?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -969,28 +993,80 @@ protected Map<String, Object> getRow(User user, Container container, Map<String,
...
+
+    }
+
     private void fireSamplesChanged()
     {
         if (_sampleType != null)
             _sampleType.onSamplesChanged(getUser(), null);
     }
 
+    private void fireSamplesRecalc(boolean needRecalc, Container recalcContainer)
```
labkey-susanh: Is this used?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -1104,6 +1182,9 @@ public DataIterator getDataIterator(DataIteratorContext context)
             final int batchSize = context.getInsertOption().batch ? BATCH_SIZE : 1;
             addGenId.addSequenceColumn(genIdCol, sampleType.getContainer(), ExpSampleTypeImpl.SEQUENCE_PREFIX, sampleType.getRowId(), batchSize, sampleType.getMinGenId());
             addGenId.addUniqueIdDbSequenceColumns(ContainerManager.getRoot(), materialTable);
+            // only add when AliquotedFrom column is not null
```
labkey-susanh, line 1209: Is this currently true or meant to be true?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -197,25 +188,104 @@ protected DataIteratorBuilder preTriggerDataIterator(DataIteratorBuilder in, Dat
...
+        Map<Enum, Object> finalConfigParameters = configParameters == null ? new HashMap<>() : configParameters;
+        finalConfigParameters.put(ExperimentService.QueryOptions.GetSampleRecomputeCol, true);
+        int ret = _importRowsUsingDIB(user, container, rows, outputRows, getDataIteratorContext(errors, InsertOption.INSERT, finalConfigParameters), extraScriptContext);
         if (ret > 0 && !errors.hasErrors())
         {
-            if (!_sampleType.isMedia())
+            onSamplesChanged(outputRows);
+            audit(QueryService.AuditAction.INSERT);
+        }
+        // TODO handle outputRows
```
labkey-susanh: Handle how?

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: The check of !Objects.equals(existingUnits, newUnits) does account for (3 L and 3 mL).

The old code also checks for (3L vs 3000mL) and skip equal changes. But I purposefully removed that logic, because:
- It's incorrect in certain cases. In the case of absent new unit, the old code assumes sample type unit. This is wrong. An absent unit could mean not updating existing unit. But the existing unit might be different from sample type unit.
- It's a rare use case. I figure it's better to recalculate in those rare case, than to compare for all rows just to catch those rare cases.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: It's used by UPSERT, which I don't think it's possible to do with our query api

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Yes, right. I meant the 3L vs 3000mL case. Good point about the different reasons for the absent unit during updates. So many cases to consider!

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: Fixed

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -240,20 +240,26 @@ public FieldKeyStringExpression getParsedNameExpression()
 
     private final String _currentDataTypeName; // used for name expression validation/preview at creation time, before the SampleType or DataClass is created
 
-    public NameGenerator(@NotNull String nameExpression, @Nullable TableInfo parentTable, boolean allowSideEffects, @Nullable Map<String, String> importAliases, @Nullable Container container, Function<String, Long> getNonConflictCountFn, String counterSeqPrefix, boolean validateSyntax, @Nullable List<? extends GWTPropertyDescriptor> domainProperties, String currentDataTypeName)
+    public NameGenerator(@NotNull String nameExpression, @Nullable TableInfo parentTable, boolean allowSideEffects, @Nullable Map<String, String> importAliases, @Nullable Container container, Function<String, Long> getNonConflictCountFn, String counterSeqPrefix, boolean validateSyntax, @Nullable List<? extends GWTPropertyDescriptor> domainProperties, String currentDataTypeName, boolean allBulkRemapCache)
```
labkey-susanh, line 248: The number of arguments for these constructors is a bit ridiculous, but we can save that refactor for another day.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -2183,6 +2188,9 @@ public String getValue(Map map, @Nullable Map<String, DbSequence> counterSequenc
 
                     long currentSeqMax = counterSeq.current();
 
+                    if (_getNonConflictCountFn != null)
+                        existingCount = _getNonConflictCountFn.apply(prefix);
+
```
labkey-susanh, line 2248: Are these NameGenerator changes a part of the aliquot rollup changes or just incidental?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -295,10 +296,8 @@ public AliquotRollupDataIteratorBuilder(@NotNull DataIteratorBuilder in, ExpSamp
         public DataIterator getDataIterator(DataIteratorContext context)
         {
             DataIterator pre = _in.getDataIterator(context);
-
-            if (!context.getInsertOption().allowUpdate)
-                return pre; // recompute for new insert are already handled in ExperimentalServiceImpl.saveExpMaterialAliquotOutputs
-
+//            if (!context.getInsertOption().allowUpdate)
+//                return pre; // recompute for new insert are already handled in ExperimentalServiceImpl.saveExpMaterialAliquotOutputs
```
labkey-susanh: Remove?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -312,76 +311,92 @@ public static class AliquotRollupDataIterator extends WrapperDataIterator
...
+                    {
+                        if (_storedAmountCol == null && !Objects.equals(existingUnits, newUnits))
+                            amountChanged = true;
+                        if (_unitsCol == null && !Objects.equals(existingAmount, newAmount))
+                            amountChanged = true;
+                    }
 
-                if (!aliquotParentLsids.isEmpty() && amountMayHaveChanged)
-                    updatedSampleLsids.addAll(aliquotParentLsids);
+                    //TODO for merge will existing map be empty for new record?
```
labkey-susanh: Still TODO?

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -396,7 +396,7 @@ public void setCategory(String category)
     }
 
     @NotNull
-    private NameGenerator createNameGenerator(@NotNull String expr, @Nullable Container dataContainer, @Nullable User user)
+    private NameGenerator createNameGenerator(@NotNull String expr, @Nullable Container dataContainer, @Nullable User user, boolean skipMaxSampleCounterFunction)
```
labkey-susanh: A comment about when this should be true or false would be helpful here.

labkey-susanh (approved):

api/src/org/labkey/api/data/NameGenerator.java
XingY: It's part of the effort to get rid of the deadlocks from ExperimentStressTest. In the end I'm unable to get around the deadlock on ExperimentRun so I won't be enabling stress test in this branch. 

I could roll back those special handling added to reduce deadlock for stress test, or maybe will just leave them in. They do help with deadlocks when run locally in sql server:
- skip _getNonConflictCountFn to avoid the read/write lock on exp.material
- skip aliquot roll up to to avoid deadlock on exp.material
- don't use bulk remap cache

PR #4191, opened 2023-03-10
Issue 47311: Rewrite MultiValuedLookupColumn FieldKey

labkey-nicka:
#### Rationale
This addresses [Issue 47311](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47311) by rewriting the FieldKey for `MultiValuedLookupColumn` to be composed from the parent plus the display field name.

#### Changes
* Recompose the FieldKey for `MultiValuedLookupColumn` in it's constructor.
* Simplify various casting scenarios that use `instanceof`.
* Cleanup dead comments / commented out code.

### DIFF
--- a/api/src/org/labkey/api/data/MultiValuedDisplayColumn.java
+++ b/api/src/org/labkey/api/data/MultiValuedDisplayColumn.java
@@ -55,16 +55,15 @@ public class MultiValuedDisplayColumn extends DisplayColumnDecorator implements 
         super(dc);
 
         _boundCol = dc.getColumnInfo();
-        BaseColumnInfo lookupCol = null;
-        if (_boundCol.getFk() instanceof MultiValuedForeignKey)
+        ColumnInfo lookupCol = null;
+        if (_boundCol.getFk() instanceof MultiValuedForeignKey mvfk)
         {
             // Retrieve the value column so it can be used when rendering json or tsv values.
-            MultiValuedForeignKey mvfk = (MultiValuedForeignKey)_boundCol.getFk();
             ColumnInfo childKey = mvfk.createJunctionLookupColumn(_boundCol);
             if (childKey != null && childKey.getFk() != null)
             {
                 ForeignKey childKeyFk = childKey.getFk();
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/data/MultiValuedDisplayColumn.java
```diff
@@ -55,24 +55,23 @@ public MultiValuedDisplayColumn(DisplayColumn dc, boolean boundColumnIsNotMultiV
...
+                lookupCol = childKeyFk.createLookupColumn(childKey, childKeyFk.getLookupColumnName());
                 if (lookupCol == null)
                 {
                     LOG.warn("Failed to create lookup column from '" + childKey.getName() + "' to '" + childKeyFk.getLookupSchemaName() + "." + childKeyFk.getLookupTableName() + "." + childKeyFk.getLookupColumnName() + "'");
                 }
                 else
                 {
                     // Remove the intermediate junction table from the FieldKey
-                    lookupCol.setFieldKey(new FieldKey(_boundCol.getFieldKey(), lookupCol.getFieldKey().getName()));
+                    ((BaseColumnInfo) lookupCol).setFieldKey(new FieldKey(_boundCol.getFieldKey(), lookupCol.getFieldKey().getName()));
```
labkey-matthewb: It would be more general/correct to use MutableColumnInfo for most of these BaseColumnInfo casts.

labkey-matthewb:
It does look like those ContainerFilter comments might still be relevant.  If anything, it seems like the Application project stuff would be a good candidate to show that we need to think about this.

labkey-matthewb (approved):
See comment about using MutableColumnInfo

api/src/org/labkey/api/data/MultiValuedDisplayColumn.java
labkey-nicka: Done. Thanks.

labkey-nicka:
> It does look like those ContainerFilter comments might still be relevant. If anything, it seems like the Application project stuff would be a good candidate to show that we need to think about this.

Thanks. I've added them back.

PR #4193, opened 2023-03-13
Update media sample types to store amount and units data in exp.materials fields

labkey-susanh:
#### Rationale
Since we now have the StoredAmount and Units fields in the exp.materials tables, we will use these to store the actual amount and units for mixture batches and the separate amount and units for raw materials (instead of a single text field for these two values catenated together). We need an upgrade script to move the data around and then need to bind batches from the exp.materials fields instead of the properties attached to the output step of the mixture batch run.

#### Related Pull Requests
- https://github.com/LabKey/recipe/pull/82
- https://github.com/LabKey/biologics/pull/2002
- https://github.com/LabKey/sampleManagement/pull/1677
- https://github.com/LabKey/labkey-ui-components/pull/1139

#### Changes
* Add ability to parse strings into amounts and units
* Add aliquot and amount fields to media sample types


api/src/org/labkey/api/data/measurement/Measurement.java
```diff
@@ -6,8 +6,13 @@
 import org.junit.Test;
 import org.labkey.api.data.ConversionExceptionWithMessage;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 public class Measurement
 {
+    private static final String NUMBER_REGEX = "[+\\-]?\\d+(?:\\.\\d+)?(?:[Ee][+\\-]\\d+)?";
```
XingY, line 14: Since we accept E, i wonder if we should check for overflow issues

XingY (approved):

labkey-nicka (approved):

api/src/org/labkey/api/data/measurement/Measurement.java
labkey-susanh: I think `Double.valueOf((String) amountObj)` (used in the `convertToAmount` method handles this as best we can. That is (from the JavaDocs)

>  ... s is regarded as representing an exact decimal value in the usual "computerized scientific notation" or as an exact hexadecimal value; this exact numerical value is then conceptually converted to an "infinitely precise" binary value that is then rounded to type double by the usual round-to-nearest rule of IEEE 754 floating-point arithmetic, which includes preserving the sign of a zero value. Note that the round-to-nearest rule also implies overflow and underflow behaviour; if the exact value of s is large enough in magnitude (greater than or equal to (MAX_VALUE + ulp(MAX_VALUE)/2), rounding to double will result in an infinity and if the exact value of s is small enough in magnitude (less than or equal to MIN_VALUE/2), rounding to float will result in a zero. Finally, after rounding a Double object representing this double value is returned.

Happy to discuss if you think there's more we should do here.

PR #4197, opened 2023-03-13
Issue 47495: respect lookup column provided by established foreign key

labkey-nicka:
#### Rationale
This addresses [Issue 47495](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47495) by updating the Query XML metadata generation to respect the lookup column name on an established `ForeignKey` if provided. Formerly, the metadata generator would configure all lookup metadata to only be to a primary key which is not always the case.

#### Changes
* Respect the `ForeignKey.getLookupColumnName()` if available when generating Query XML metadata.
* Fix up some usages of `instanceof` casts
* Revert FK equivalency check to a String to String comparison. Greatly reduces the amount of XML generated for most columns with an FK where the FK is not modified by the user. Reverted from change in #3156.

### DIFF
--- a/api/src/org/labkey/api/data/ColumnInfo.java
+++ b/api/src/org/labkey/api/data/ColumnInfo.java
@@ -56,7 +56,6 @@ public interface ColumnInfo extends ColumnRenderProperties
                 return new AttachmentDisplayColumn(colInfo);
             }
 
-
             DataColumn dataColumn = new DataColumn(colInfo);
             if (colInfo.getPropertyType() == PropertyType.MULTI_LINE)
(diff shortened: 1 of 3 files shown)


labkey-klum (approved):

labkey-matthewb:
One thing to be aware of.  A lot of metadata performance is due to loading lookup tables for stupid things.  I don’t know that this is the case here.  However, it may be worth checking that getLookupColumn() does not indirectly cause us to instantiate the remote table.  There are a few cases where we remember whether a value is actually set or not which is handy to avoid computing values, that are not set and could be left blank in the generated/copied metadata.

PR #4198, opened 2023-03-14
Issue 43559: Domain designer error on invalid lookup

labkey-martyp:
#### Rationale
This throws an error and doesn't allow a domain design to be saved if it has an invalid lookup. The use case would be when a lookup target table is altered after a domain is created. Next time the domain design is opened the field and form give an error and won't allow the domain to be saved until corrected.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43559&_docid=issue%3A43559

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1141

#### Changes
- Add lookupIsValid flag to property descriptor calculated for API query
- Calculate valid lookup based on container and table name within schema
- Update @labkey/components package with UI handling of lookupIsValid


labkey-jeckels (changes requested):
This is causing BVT failures.

```
java.lang.NullPointerException: Cannot invoke "org.labkey.api.query.UserSchema.getTableNames()" because the return value of "org.labkey.api.query.QueryService.getUserSchema(org.labkey.api.security.User, org.labkey.api.data.Container, String)" is null
  at org.labkey.api.exp.property.DomainUtil.isValidPdLookup(DomainUtil.java:184)
  at org.labkey.api.exp.property.DomainUtil.getDomainDescriptor(DomainUtil.java:264)
  at org.labkey.api.exp.property.DomainUtil.getDomainDescriptor(DomainUtil.java:199)
  at org.labkey.api.exp.property.AbstractDomainKind.getDomainNamePreviews(AbstractDomainKind.java:321)
```

api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -172,6 +173,22 @@ else if (property.getLookup() != null)
...
 
+    private static boolean isValidPdLookup(User user, Container c, GWTPropertyDescriptor p)
+    {
+        Container lookupContainer = p.getLookupContainer() == null ? c : ContainerService.get().getForPath(p.getLookupContainer());
+
+        if (null == lookupContainer)
+            return false;
+
+        // Only checking names here to avoid creating tableinfo each time. Avoids potential infinite loop.
+        Set<String> tableNames = QueryService.get().getUserSchema(user, lookupContainer, p.getLookupSchema()).getTableNames();
```
labkey-jeckels: Can we use a `DefaultSchema` to help cache schemas in case multiple lookups for the same domain reference the same schemas? @labkey-matthewb

api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -172,6 +173,22 @@ else if (property.getLookup() != null)
...
+
+        if (tableNames.stream().anyMatch(p.getLookupQuery()::equalsIgnoreCase))
```
labkey-jeckels: Could also use a `CaseInsensitiveHashSet`

api/src/org/labkey/api/exp/property/DomainUtil.java
labkey-martyp: > This is causing BVT failures.
> 
> ```
> java.lang.NullPointerException: Cannot invoke "org.labkey.api.query.UserSchema.getTableNames()" because the return value of "org.labkey.api.query.QueryService.getUserSchema(org.labkey.api.security.User, org.labkey.api.data.Container, String)" is null
>   at org.labkey.api.exp.property.DomainUtil.isValidPdLookup(DomainUtil.java:184)
>   at org.labkey.api.exp.property.DomainUtil.getDomainDescriptor(DomainUtil.java:264)
>   at org.labkey.api.exp.property.DomainUtil.getDomainDescriptor(DomainUtil.java:199)
>   at org.labkey.api.exp.property.AbstractDomainKind.getDomainNamePreviews(AbstractDomainKind.java:321)
> ```

This is fixed. Templates sneaking in invalid schemas.

labkey-martyp: @josheckels QueryService.getUserSchema does use a DefaultSchema. Is that what you're referring to?

labkey-jeckels: Right - it creates a DefaultSchema behind the scenes, but it uses a new one for each time that you call `QueryService.get().getUserSchema()`. That means that it can't reuse a `UserSchema`. However, if you create and reuse your own `DefaultSchema` (one per container), it would be able to cache and reuse the schemas, which can sometimes be expensive to construct. It might be easier to talk if I'm not yet making sense.

core/package.json
```diff
@@ -55,7 +55,7 @@
   },
   "dependencies": {
     "@labkey/api": "1.18.8",
-    "@labkey/components": "2.307.2",
+    "@labkey/components": "2.308.0-fb-domain-invalid-lookup.0",
```
cnathe: since the field editor component is used in both LKB and LKSM apps, I'd suggest also opening an fb for those modules and bumping this package version there as well.

api/src/org/labkey/api/exp/property/DomainUtil.java
labkey-matthewb: Yes, reusing DefaultSchema is a good idea in general.  In this case, it's not immediately obvious who would be hanging on to the DefaultSchema.  Would the caller of getDomainDescriptor() pass it in?

labkey-jeckels: I suggested to Marty that he make `getDomainDescriptor()` manage it. That would at least reuse the `DefaultSchema` for all of the properties in a single domain.

labkey-jeckels (approved):
Thanks for the DefaultSchema changes. They look good so I'm approving, though the TeamCity build is failing for unrelated reasons.

PR #4202, opened 2023-03-15
Treat admin reactivation as recent user activity

labkey-adam:
#### Rationale
When the disable inactive accounts compliance feature is enabled, reactivated user accounts are deactivated again unless the user manages to login before the next maintenance run. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47471

Also, we weren't re-running the schema validation check after schema XML files changed, leading to stale warnings.

#### Related Pull Requests
* https://github.com/LabKey/compliance/pull/183

### DIFF
--- a/api/src/org/labkey/api/security/UserManager.java
+++ b/api/src/org/labkey/api/security/UserManager.java
@@ -942,9 +942,16 @@ public class UserManager
             Table.update(currentUser, CoreSchema.getInstance().getTableInfoPrincipals(),
                     Collections.singletonMap("Active", active), userId);
 
-            // update the modified bit on the users table
             Map<String, Object> map = new HashMap<>();
-            map.put("Modified", new Timestamp(System.currentTimeMillis()));
+
+            // Treat re-activation as an activity for the purpose of inactivity tracking, Issue 47471
+            if (active)
+            {
+                Timestamp now = new Timestamp(System.currentTimeMillis());
+                map.put("LastActivity", now);
+            }
+
+            // Call update unconditionally to ensure Modified & ModifiedBy are always updated
             Table.update(currentUser, CoreSchema.getInstance().getTableInfoUsers(), map, userId);
 
             removeRecentUser(userToAdjust);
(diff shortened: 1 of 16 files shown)


api/src/org/labkey/api/security/UserManager.java
```diff
@@ -942,9 +942,16 @@ public static void setUserActive(User currentUser, User userToAdjust, boolean ac
...
 
-            // update the modified bit on the users table
             Map<String, Object> map = new HashMap<>();
-            map.put("Modified", new Timestamp(System.currentTimeMillis()));
+
+            // Treat re-activation as an activity for the purpose of inactivity tracking, Issue 47471
+            if (active)
+            {
+                Timestamp now = new Timestamp(System.currentTimeMillis());
+                map.put("LastActivity", now);
```
XingY, line 951: Instead of adding LastActivity, is it possible to query user history to find the last activation timestamp?

core/src/org/labkey/core/view/template/bootstrap/CoreWarningProvider.java
```diff
@@ -71,12 +71,13 @@ public CoreWarningProvider()
...
 
-    public void startSchemaCheck()
+    public void startSchemaCheck(int delaySeconds)
     {
         // Issue 46264 - proactively check all DB schemas against the schema XML
-        // Launch this in the background, but delay by 10 seconds to reduce impact on other startup tasks
-        JobRunner.getDefault().execute(TimeUnit.SECONDS.toMillis(10), () ->
+        JobRunner.getDefault().execute(TimeUnit.SECONDS.toMillis(delaySeconds), () ->
         {
+            _dbSchemaWarnings.clear();
```
XingY, line 79: Is this thread safe? I guess so because it's inside run()?

XingY (approved):

core/src/org/labkey/core/view/template/bootstrap/CoreWarningProvider.java
labkey-adam: Yes, `_dbSchemaWarnings` is a `ConcurrentHashMap`, but more importantly this is always executed by a single-thread job runner, so they won't overlap.

api/src/org/labkey/api/security/UserManager.java
labkey-adam: This would probably work in practice, but parsing the audit log comments seems risky since they change and some clients truncate the audit tables periodically, and we're planning to introduce an auto-truncate feature this year.

PR #4204, opened 2023-03-16
ELN: Global Attachments

labkey-alan:
#### Rationale
This PR adds a few methods to the AttachmentService that are needed by labbook in order to better support our "global attachments" optimization. See the related labbook PR for more information.

#### Related Pull Requests
* https://github.com/LabKey/labbook/pull/441

#### Changes
* AttachmentService: Add methods
  * add deleteAttachments(AttachmentParent parent, Collection<String> names, @Nullable User auditUser)
  * add download(HttpServletResponse response, AttachmentParent parent, String filename, @Nullable String alias, boolean inlineIfPossible)
  * add getAttachments(AttachmentParent parent, Collection<String> names)
BaseDownloadAction
  * Add getAlias to InlineDownloader interface

### DIFF
--- a/api/src/org/labkey/api/attachments/AttachmentService.java
+++ b/api/src/org/labkey/api/attachments/AttachmentService.java
@@ -63,6 +63,8 @@ public interface AttachmentService
         return ServiceRegistry.get().getService(AttachmentService.class);
     }
 
+    void download(HttpServletResponse response, AttachmentParent parent, String filename, @Nullable String alias, boolean inlineIfPossible) throws ServletException, IOException;
+
     void download(HttpServletResponse response, AttachmentParent parent, String name, boolean inlineIfPossible) throws ServletException, IOException;
 
     HttpView getHistoryView(ViewContext context, AttachmentParent parent);
--- a/api/src/org/labkey/api/attachments/AttachmentService.java
+++ b/api/src/org/labkey/api/attachments/AttachmentService.java
@@ -42,6 +42,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
 import java.util.List;
+import java.util.Map;
 
 /**
  * User: adam
@@ -75,6 +76,13 @@ public interface AttachmentService
     void deleteAttachments(Collection<AttachmentParent> parents);
 
     /**
+     * Deletes the attachments with the given names from the given AttachmentParent.
+     * @param parent: The AttachmentParent to delete files from
+     * @param names: The file names to delete from the AttachmentParent
+     */
+    void deleteAttachments(AttachmentParent parent, Collection<String> names, @Nullable User auditUser);
+
+    /**
      * @param auditUser set to null to skip audit
      */
     void deleteAttachment(AttachmentParent parent, String name, @Nullable User auditUser);
@@ -84,6 +92,12 @@ public interface AttachmentService
     void copyAttachment(AttachmentParent parent, Attachment a, String newName, User auditUser) throws IOException;
 
     void moveAttachments(Container newContainer, List<AttachmentParent> parents, User auditUser) throws IOException;
+
+    /**
+     * You should not use this method. It only exists so an upgrade script that manually touches the attachment table
+     * can reset the caches of what it touched.
+     */
+    void clearCaches(Collection<String> containerIds);
 
     @NotNull
     List<AttachmentFile> getAttachmentFiles(AttachmentParent parent, Collection<Attachment> attachments) throws IOException;
@@ -100,6 +114,15 @@ public interface AttachmentService
 
     @Nullable
     Attachment getAttachment(AttachmentParent parent, String name);
+
+    /**
+     * Gets attachments from the given parent that match the given names. This method intentionally skips hitting the
+     * AttachmentCache in order to prevent very large AttachmentParents that would explode the cache with tens of
+     * thousands of items when we are only interested in a few.
+     * @param parent: The AttachmentParent to fetch attachments from
+     * @param names: The list of attachment names to fetch
+     */
+    Map<String, Attachment> getAttachments(AttachmentParent parent, Collection<String> names);
 
     void writeDocument(DocumentWriter writer, AttachmentParent parent, String name, boolean asAttachment) throws ServletException, IOException;
 
--- a/api/src/org/labkey/api/attachments/AttachmentService.java
+++ b/api/src/org/labkey/api/attachments/AttachmentService.java
@@ -93,12 +93,6 @@ public interface AttachmentService
 
     void moveAttachments(Container newContainer, List<AttachmentParent> parents, User auditUser) throws IOException;
 
-    /**
-     * You should not use this method. It only exists so an upgrade script that manually touches the attachment table
-     * can reset the caches of what it touched.
-     */
-    void clearCaches(Collection<String> containerIds);
-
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
```diff
@@ -874,6 +903,37 @@ private WebPartView getResultSetView(SQLFragment sql, String title, @Nullable Ac
...
+        {
+            for (Attachment attachment : getAttachments(parent))
+                if (names.contains(attachment.getName()))
+                    attachments.put(attachment.getName(), attachment);
+        }
+        else
+        {
+            SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("Parent"), parent.getEntityId());
+            filter.addCondition(FieldKey.fromParts("DocumentName"), names, CompareType.IN);
+            // Note: we are intentionally skipping the AttachmentCache here. If we hit the cache here while getting
```
labkey-alan, line 917: I believe this is the right thing to do, but I am not 100% sure. Maybe we should just use the normal code path for retrieving attachments, but that would mean that the first user to look at any notebook would immediately load every attachment in the global attachments parent into the cache, which may be a mistake. Right now that would mean for our largest customer a minimum of 4864 attachments, but the actual number may be much, much higher.

labkey-alan: If we really are concerned about not caching the Notebook related attachment queries we could put a cache manager in the labbook module that caches the result of this method, this way we're still caching, but at a layer above, and we could even make the cache keys specific to the "sourceId", instead of specific to the parent here. This way we could cache by each notebook or entry.

api/src/org/labkey/api/attachments/AttachmentService.java
```diff
@@ -85,6 +93,12 @@ static AttachmentService get()
 
     void moveAttachments(Container newContainer, List<AttachmentParent> parents, User auditUser) throws IOException;
 
+    /**
+     * You should not use this method. It only exists so an upgrade script that manually touches the attachment table
+     * can reset the caches of what it touched.
+     */
+    void clearCaches(Collection<String> containerIds);
```
labkey-jeckels: Instead of needing to add this method with the comment saying to not use it, we could call the existing `CacheManager.clearAllKnownCaches()` method. I'd say we're better served by having the upgrade take a second or two longer as it repopulates instead of introducing a method we don't want anyone to call.

core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
```diff
@@ -127,6 +129,7 @@ public class AttachmentServiceImpl implements AttachmentService, ContainerManage
 {
     private static final String UPLOAD_LOG = ".upload.log";
     private static final Map<String, AttachmentType> ATTACHMENT_TYPE_MAP = new HashMap<>();
+    private static final Set<String> ATTACHMENT_COLUMNS = new CsvSet("Parent, Container, DocumentName, DocumentSize, DocumentType, Created, CreatedBy, LastIndexed");
```
labkey-jeckels: Nit: I've never liked `CsvSet` much, as it's always seemed silly to me to have to parse a hard-coded string. I prefer `Set.of("Parent", "Container", ...)` but it also doesn't matter that much

core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
```diff
@@ -409,6 +412,21 @@ public void deleteAttachmentIndex(AttachmentParent parent)
...
+        SearchService ss = SearchService.get();
+        if (ss != null)
+            ss.deleteResource(makeDocId(parent, name));
+
+        new SqlExecutor(coreTables().getSchema()).execute(sqlDelete(parent, name));
+        if (parent instanceof AttachmentDirectory)
+            ((AttachmentDirectory)parent).deleteAttachment(auditUser, name);
+
+        if (null != auditUser)
+            addAuditEvent(auditUser, parent, name, "The attachment " + name + " was deleted");
```
labkey-jeckels, line 431: Given the comment about double-deleting above, should we only log the audit event if the `execute()` call actually deleted a row (and returned a non-zero number of rows that were affected)?

core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
labkey-jeckels: I think it depends on the usage pattern. If this method is called once when we render a method, caching isn't that important. If we end up running the query many times, caching is probably useful. If a cache will speed things up, we might want one that uses the parent ID and name as the cache key instead of grabbing everything that belongs to a parent.

labkey-jeckels (approved):
Some comments but looks OK to me. Let me know if you want to discuss any of my feedback.

core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
labkey-alan: This was just copied from somewhere in attachment land, which is probably very old, I'll update it.

api/src/org/labkey/api/attachments/AttachmentService.java
labkey-alan: Yeah I like your idea, since it is only ever used once during the upgrade, so I can call that method once at the end of the upgrade script.

core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
labkey-alan: We weren't doing that before, and this code is just old code moved to a shared spot.

core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
labkey-alan: This method is only called when we fully "hydrate" a notebook, which only happens when a user is on the notebook edit page, or the submit/sign pages.

labkey-jeckels: OK. If we just run the query once for those operations, it doesn't seem worth optimizing with a cache.

labkey-adam (approved):

core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
```diff
@@ -409,6 +412,21 @@ public void deleteAttachmentIndex(AttachmentParent parent)
...
 
+    private void _deleteAttachment(AttachmentParent parent, String name, @Nullable User auditUser)
+    {
+        checkSecurityPolicy(auditUser, parent);   // Only check policy if there are attachments (a client may delete attachment and policy, but attempt to delete again)
+        SearchService ss = SearchService.get();
+        if (ss != null)
+            ss.deleteResource(makeDocId(parent, name));
+
+        new SqlExecutor(coreTables().getSchema()).execute(sqlDelete(parent, name));
+        if (parent instanceof AttachmentDirectory)
```
labkey-nicka, line 427: nit: Consider using the `instanceof` to cast a new variable instance.

```java
if (parent instanceof AttachmentDirectory directory)
    directory.deleteAttachment(auditUser, name);
```

labkey-nicka (approved):

labkey-klum (approved):

PR #4216, opened 2023-03-22
Implement TableResultSet.getWrapped() as better alternative to type cast

labkey-matthewb:
#### Rationale
Adding getWrapped() is a small step to break implementation expectations about ResultSet implemention returned by QueryService methods (or SelectBuilder).

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/query/QueryService.java
+++ b/api/src/org/labkey/api/query/QueryService.java
@@ -650,7 +650,7 @@ public interface QueryService
 
         SQLFragment buildSqlFragment();
         SqlSelector buildSqlSelector(Map<String, Object> parameters);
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/query/QueryService.java
```diff
@@ -650,7 +650,7 @@ default SelectBuilder columns(ColumnInfo... cols)
 
         SQLFragment buildSqlFragment();
         SqlSelector buildSqlSelector(Map<String, Object> parameters);
```
labkey-adam, line 652: Shouldn't `parameters` be `@Nullable` here as well?

labkey-adam (changes requested):

labkey-adam (approved):

PR #4217, opened 2023-03-22
Issue 47229: Support all characters within aliases

labkey-nicka:
#### Rationale
This addresses [Issue 47229](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47229) by refactoring the handling of alias value parsing to respect quoted values. The `AliasInputHelper` is responsible for all code paths (API, Excel Import, XAR import, etc) for parsing Alias input. Here is a brief set of examples and how they would parse after these changes:

```
Alias
bob              // 1 "bob"
bob,mike         // 2 "bob" and "mike"
"bob,mike"       // 1 "bob,mike"
bob;mike         // 1 "bob;mike"
"bob;mike"       // 1 "bob;mike"
[bob,mike]       // 2 "[bob" and "mike]"
"bob","mike"     // 2 "bob" and "mike"
"bo,b","m,ike"   // 2 "bo,b" and "m,ike"
[bob,            // 1 "[bob"
,bob]            // 1 "bob]"
[]               // 1 "[]"
```

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4217
* https://github.com/LabKey/labkey-ui-components/pull/1145
* https://github.com/LabKey/biologics/pull/2017
* https://github.com/LabKey/sampleManagement/pull/1697
* https://github.com/LabKey/inventory/pull/782

#### Changes
* Refactor `AliasInputHelper.splitAliases()` to account for quoted characters. Remove special handling for `[`, `]`, and `,` as these are now supported characters.
* Remove recursive aspect of parsing alias values instead opting for a pattern where only primitives (strings and integers) are supported as values within collection-like types.
* Add unit tests.

### DIFF
--- a/experiment/src/org/labkey/experiment/api/AliasInsertHelper.java
+++ b/experiment/src/org/labkey/experiment/api/AliasInsertHelper.java
@@ -15,8 +15,8 @@
  */
 package org.labkey.experiment.api;
 
-import org.apache.commons.lang3.StringUtils;
 import org.json.JSONArray;
+import org.junit.Test;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.SQLFragment;
@@ -28,13 +28,14 @@ import org.labkey.api.exp.api.ExperimentService;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.security.User;
 
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.stream.Collectors;
+
+import static org.junit.Assert.assertEquals;
 
 public class AliasInsertHelper
 {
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/api/AliasInsertHelper.java
```diff
@@ -96,46 +132,22 @@ else if (value instanceof Integer i)
...
-                if (o == null)
-                    continue;
-                String s = StringUtils.trim(String.valueOf(o));
-                if (s.length() == 0)
-                    continue;
-
-                parts.add(s);
-            }
+        Set<String> aliasNames = new HashSet<>();
+        String[] tokens = aliases.split(",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)", -1);
```
labkey-susanh, line 140: A comment for this regex would surely be helpful.

labkey-susanh (approved):
This seems solid, but you might consider using TabLoader, just so we have more things going through that same code path and can be more clear on uniformity in parsing support.

experiment/src/org/labkey/experiment/api/AliasInsertHelper.java
labkey-nicka: Added.

PR #4220, opened 2023-03-23
Move ownership of compliance schema

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=47547

Management of the "compliance" schema was moved to the Signing module, but nobody told the `core.Modules` table.

To prevent future problems related to schema moves, log an error if we detect a schema that's claimed by multiple rows in the Modules table and refuse to delete a schema that's claimed by a known module.

#### Related Pull Requests
* https://github.com/LabKey/signing/pull/1

### DIFF
--- a/api/src/org/labkey/api/module/ModuleLoader.java
+++ b/api/src/org/labkey/api/module/ModuleLoader.java
@@ -30,6 +30,7 @@ import org.labkey.api.collections.CaseInsensitiveHashSet;
 import org.labkey.api.collections.CaseInsensitiveHashSetValuedMap;
 import org.labkey.api.collections.CaseInsensitiveTreeMap;
 import org.labkey.api.collections.CaseInsensitiveTreeSet;
+import org.labkey.api.collections.LabKeyCollectors;
 import org.labkey.api.collections.Sets;
 import org.labkey.api.data.ConnectionWrapper;
 import org.labkey.api.data.Container;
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -685,6 +688,27 @@ public void addStaticWarnings(@NotNull Warnings warnings, boolean showAllWarning
         _log.info("LabKey Server startup is complete; " + execution.getLogMessage());
     }
 
+    // Check for multiple modules claiming the same schema. Inspired by Issue 47547.
+    private void warnAboutDuplicateSchemas(Collection<ModuleContext> values)
+    {
+        Map<String, String> schemaToModule = CaseInsensitiveHashMap.of();
+        for (ModuleContext ctx : values)
+        {
+            String schemas = ctx.getSchemas();
```
labkey-jeckels, line 697: @labkey-adam will this return the schemas owned, according to the `core.modules` table or the ones that the Module object is currently claiming?

labkey-jeckels (approved):

api/src/org/labkey/api/module/ModuleLoader.java
labkey-adam: @labkey-jeckels these come from the database

PR #4221, opened 2023-03-24
Make the demographics date field not required

labkey-klum:
#### Rationale
For date based and continuous studies we no longer want the date field to be required for demographics datasets. Additionally we will remove the behavior that automatically inserted the start date into demographics (and participant alias datasets) if a value was not provided.

[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47344)

### DIFF
--- a/study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
+++ b/study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
@@ -277,27 +277,6 @@ public class DatasetDataIteratorBuilder implements DataIteratorBuilder
          */
 
         //
-        // date
-        //
-
-        if (!timetype.isVisitBased()
-                && null == indexVisitDate
-                && (_datasetDefinition.isDemographicData() || _datasetDefinition.isParticipantAliasDataset())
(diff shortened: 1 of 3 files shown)


labkey-jeckels:
This is great!

labkey-jeckels (commented):
Looks good to me. I'm kind of surprised it's this small of a change to enable. Having Matt's review would be great too so leaving this as just a comment.

study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
```diff
@@ -276,27 +276,6 @@ else if (match.getPropertyType() == PropertyType.FILE_LINK)
...
-                @Override
-                public Object call()
-                {
-                    return start;
-                }
-            });
-            it.indexVisitDateOutput = indexVisitDate;
-        }
-
         //
```
labkey-matthewb, line 279: I don't know why this fallback behavior would be important.  Was it just to satisfy the 'non-null' date?

labkey-matthewb (approved):

PR #4240, opened 2023-03-28
Issue 46744: Row selectors in QueryWebPart with study schema

RosalineP:
#### Rationale
Currently, it is not possible to use LabKey.QueryWebPart with a study dataset and specify that you do not want selectors to appear. This fix resolves that issue while not appearing to affect default dataset row selection behavior.

#### Related Pull Requests
* N/A

#### Changes
* Remove code block setting selector override


RosalineP:
Pending TC results.

PR #4241, opened 2023-03-28
Issue 46744: Row selectors in QueryWebPart with study schema

RosalineP:
#### Rationale
Currently, it is not possible to use LabKey.QueryWebPart with a study dataset and specify that you do not want selectors to appear. This fix resolves that issue while not appearing to affect default dataset row selection behavior.

#### Related Pull Requests
* N/A

#### Changes
* Remove code block setting selector override

### DIFF
--- a/study/src/org/labkey/study/query/DatasetQueryView.java
+++ b/study/src/org/labkey/study/query/DatasetQueryView.java
@@ -36,6 +36,7 @@ import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.DataRegion;
 import org.labkey.api.data.DisplayColumn;
 import org.labkey.api.data.MenuButton;
+import org.labkey.api.data.PanelButton;
 import org.labkey.api.data.RenderContext;
 import org.labkey.api.data.SimpleDisplayColumn;
 import org.labkey.api.data.SimpleFilter;
@@ -411,8 +412,20 @@ public class DatasetQueryView extends StudyQueryView
 
         populateChartsReports(bar);
 
-        List<String> recordSelectorColumns = view.getDataRegion().getRecordSelectorValueColumns();
-        bar.add(createExportButton(recordSelectorColumns));
+        if (showExportButtons())
+        {
+            List<String> recordSelectorColumns = view.getDataRegion().getRecordSelectorValueColumns();
+
+            PanelButton exportButton = createExportButton(recordSelectorColumns);
+            if (exportButton.hasSubPanels())
+            {
+                if ((recordSelectorColumns != null && !recordSelectorColumns.isEmpty()) || (getTable() != null && !getTable().getPkColumns().isEmpty()))
+                {
+                    bar.setAlwaysShowRecordSelectors(true);
+                }
+                bar.add(exportButton);
+            }
+        }
 
         User user = getUser();
         var table = getTable();
--- a/study/src/org/labkey/study/query/DatasetQueryView.java
+++ b/study/src/org/labkey/study/query/DatasetQueryView.java
@@ -414,12 +414,6 @@ public class DatasetQueryView extends StudyQueryView
         List<String> recordSelectorColumns = view.getDataRegion().getRecordSelectorValueColumns();
         bar.add(createExportButton(recordSelectorColumns));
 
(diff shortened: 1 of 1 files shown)


study/src/org/labkey/study/query/DatasetQueryView.java
```diff
@@ -414,12 +414,6 @@ protected void populateButtonBar(DataView view, ButtonBar bar)
         List<String> recordSelectorColumns = view.getDataRegion().getRecordSelectorValueColumns();
         bar.add(createExportButton(recordSelectorColumns));
 
-        // Duplicates logic from super.populateButtonBar(), but we need selectors
```
RosalineP, line 417: Note that doing `(((recordSelectorColumns != null && !recordSelectorColumns.isEmpty()) || (getTable() != null && !getTable().getPkColumns().isEmpty()) && showRecordSelectors()))` is an alternative. Currently don't see an effect to getting rid of this code block, though.

RosalineP:
Looks like some of the TC builds got removed partway through green runs—will merge up and re-trigger soon

labkey-nicka (commented):
As discussed in person I think we should try to align the behavior of `DatasetQueryView` as much with `QueryView` as possible to ensure respecting view settings (i.e. QueryWebPart settings). There are a couple options as I see it:

1. Replicate the behavior from `QueryView` where the selection boxes are forced to display regardless of view settings when the export panels are available: 
```java
if (showExportButtons())
{
    List<String> recordSelectorColumns = view.getDataRegion().getRecordSelectorValueColumns();

    PanelButton exportButton = createExportButton(recordSelectorColumns);
    if (exportButton.hasSubPanels())
    {
        if ((recordSelectorColumns != null && !recordSelectorColumns.isEmpty()) || (getTable() != null && !getTable().getPkColumns().isEmpty()))
        {
            bar.setAlwaysShowRecordSelectors(true);
        }
        bar.add(exportButton);
    }
}
```
2. Fully remove the `recordSelectorColumns` lines like you have done but also remove the corollary logic in `QueryView` for respecting this when part of export. It seems odd to me that exporting would dictate the presence of selection checkboxes over/above a user supplied specific configuration asking not to display checkboxes. For this we'd want to verify the scenario in [Issue 24530](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=24530) to ensure that the checkboxes display as desired for audit log grids.

An additional discrepancy to note is that `DatasetQueryView` is not respecting `showExportButton()` either. Either way we go here I think we should probably fix that as well.

study/src/org/labkey/study/query/DatasetQueryView.java
labkey-nicka: As discussed in person I think we should try to align the behavior of `DatasetQueryView` as much with `QueryView` as possible to ensure respecting view settings (i.e. QueryWebPart settings). There are a couple options as I see it:

1. Replicate the behavior from `QueryView` where the selection boxes are forced to display regardless of view settings when the export panels are available: 
```java
if (showExportButtons())
{
    List<String> recordSelectorColumns = view.getDataRegion().getRecordSelectorValueColumns();

    PanelButton exportButton = createExportButton(recordSelectorColumns);
    if (exportButton.hasSubPanels())
    {
        if ((recordSelectorColumns != null && !recordSelectorColumns.isEmpty()) || (getTable() != null && !getTable().getPkColumns().isEmpty()))
        {
            bar.setAlwaysShowRecordSelectors(true);
        }
        bar.add(exportButton);
    }
}
```
2. Fully remove the `recordSelectorColumns` lines like you have done but also remove the corollary logic in `QueryView` for respecting this when part of export. It seems odd to me that exporting would dictate the presence of selection checkboxes over/above a user supplied specific configuration asking not to display checkboxes. For this we'd want to verify the scenario in [Issue 24530](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=24530) to ensure that the checkboxes display as desired for audit log grids.

An additional discrepancy to note is that `DatasetQueryView` is not respecting `showExportButton()` either. Either way we go here I think we should probably fix that as well.

RosalineP: (After discussing in-person we opted for option 1)

labkey-nicka (approved):

PR #4255, opened 2023-03-30
Enable lineage relationship between custom data classes in apps

XingY:
#### Rationale
Adding support to allow data class to have import aliases, just like sample types. 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1159
- https://github.com/LabKey/platform/pull/4255
- https://github.com/LabKey/sampleManagement/pull/1723
- https://github.com/LabKey/biologics/pull/2054

#### Changes
* schema change to add DataParentImportAliasMap to exp.dataclass table
* save/update dataclass domain with additional parent alias param
* update dataclass DIB to take parent alias as extra fn param

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
@@ -124,6 +124,7 @@ import org.labkey.experiment.ExpDataIterators.PersistDataIteratorBuilder;
 import org.labkey.experiment.controllers.exp.ExperimentController;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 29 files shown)


experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -997,10 +1006,18 @@ else if (Column.ClassId.name().equalsIgnoreCase(name))
                 di = LoggingDataIterator.wrap(new NameExpressionDataIterator(di, context, dataClassTInfo, getContainer(), _dataClass.getMaxDataCounterFunction(), DATA_COUNTER_SEQ_PREFIX + _dataClass.getRowId() + "-")
                         .setAllowUserSpecifiedNames(NameExpressionOptionService.get().allowUserSpecifiedNames(getContainer()))
                         .addExtraPropsFn(() -> {
-                            if (c != null)
```
cnathe, line 1000: is this `c != null` check still needed inside the try/catch?

cnathe (approved):

cnathe (changes requested):
Actually, just thought of this. You should bump the package.json @labkey/component versions in the platform modules so that we can verify the Data Class Designer changes there.

cnathe (approved):

PR #4256, opened 2023-03-30
Move VARCHAR cast to ExprColumn

labkey-matthewb:
#### Rationale
We can't make both array_agg() and json_object() work at the same time with "auto" casting here.  This gets us back to where we were.  We may need to add more smarts about generating SQL for overloaded methods in postgres.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/query/src/org/labkey/query/sql/Method.java
+++ b/query/src/org/labkey/query/sql/Method.java
@@ -1445,6 +1445,8 @@ public abstract class Method
             return f.getParams().get(0) instanceof String;
         if (f.getParams().size() > 0)
             return false;
+        if (s.endsWith("::VARCHAR"))
+            s = s.substring(0, s.length()-"::VARCHAR".length());
         // am I 'normal' SQL String with no embedded single-quotes?
         if (s.length() >= 2 && s.startsWith("'"))
             return s.length()-1 == s.indexOf('\'',1);
@@ -1454,7 +1456,7 @@ public abstract class Method
         return false;
     }
 
-
+    /** see {@link #isSimpleString(SQLFragment)} */
     public static String toSimpleString(SQLFragment f)
     {
         if (!isSimpleString(f))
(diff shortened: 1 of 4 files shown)


query/src/org/labkey/query/sql/Method.java
```diff
@@ -1454,15 +1456,18 @@ public static boolean isSimpleString(SQLFragment f)
         return false;
     }
 
-
+    /** see {@link #isSimpleString(SQLFragment)} */
```
labkey-nicka, line 1463: Consider adding unit tests for both `isSimpleString` and `toSimpleString`.

query/src/org/labkey/query/sql/Method.java
```diff
@@ -1454,15 +1456,18 @@ public static boolean isSimpleString(SQLFragment f)
...
     {
         if (!isSimpleString(f))
             throw new IllegalArgumentException(f.toDebugString());
         String s = f.getSQL();
         if ("?".equals(s) && f.getParams().size()==1)
             return (String)f.getParams().get(0);
-        assert(s.endsWith("'"));
+        assert(s.startsWith("'") || s.startsWith("N'"));
+        assert(s.endsWith("'") || s.endsWith("'::VARCHAR"));
+        if (s.endsWith("::VARCHAR"))
```
labkey-nicka, line 1473: Does case-sensitivity matter?

labkey-nicka (approved):

query/src/org/labkey/query/sql/Method.java
labkey-matthewb: No, but we are always generating these strings, so the case shouldn't vary.

query/src/org/labkey/query/sql/Method.java
labkey-matthewb: YES!

labkey-susanh (approved):

PR #4266, opened 2023-04-03
 Update use of deprecated methods

labkey-matthewb:
#### Rationale
append(Object) -> appendValue(), etc.

#### Related Pull Requests
* 

#### Changes
*


api/src/org/labkey/api/data/dialect/SqlDialect.java
```diff
@@ -581,10 +586,34 @@ public String getXorOperator()
...
+     * @param sql
+     * @param distinct
+     * @param sorted
+     * @param delimiter Simple Java string to use as a delimiter (not SQL!)
+     * @return SQLFragment holding dialect-specific GROUP_CONCAT expression
+     */
+    public final SQLFragment getGroupConcat(SQLFragment sql, boolean distinct, boolean sorted, @NotNull String delimiter)
+    {
+        SQLFragment delimiterFrag = new SQLFragment().appendStringLiteral(delimiter, this);
+        return getGroupConcat(sql, distinct, sorted, new SQLFragment(delimiterFrag));
```
labkey-adam, line 604: Why is `delimiterFrag` (a `SQLFragment`) wrapped in a `SQLFragment`?

query/src/org/labkey/query/sql/QuerySelectView.java
```diff
@@ -273,7 +273,7 @@ private SQLFragment getSelectSQL(TableInfo table, @Nullable Collection<ColumnInf
                 selectFrag.append(strComma);
                 selectFrag.append(column.getValueSql(tableAlias));
                 selectFrag.append(" AS ");
-                selectFrag.append(dialect.makeLegalIdentifier(column.getAlias()));
+                selectFrag.appendIdentifier(dialect.makeLegalIdentifier(column.getAlias()));
```
labkey-adam, line 276: Should we add `appendIdentifier(String, SqlDialect)`?

labkey-adam (approved):
A couple suggestions/questions, but no need for re-review

query/src/org/labkey/query/sql/QuerySelectView.java
labkey-matthewb: That would be nice.  But it's another huge cross repository change because currently callers are responsible for propelty calling makeLegalIdentifer() or getSelectName() (quoted) or other methods.

PR #4270, opened 2023-04-04
Issue 47289: Export List Archive if the user is an Admin of the folders of the selected Lists, else throw Permission error

labkey-bpatel:
#### Rationale
Issue [47289](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47289): Export List Archive if the user is an Admin of the folders of the selected Lists, else throw Permission error

Currently, List export is only supported for lists defined in the current folder. This PR changes that behavior to allow exporting of the Lists if the user is an Admin of the Folder of the selected Lists, otherwise throws Permission error.

#### Changes
* Update ExportListArchiveAction to export Lists if the user is an admin of the folders of the selected Lists


list/src/org/labkey/list/controllers/ListController.java
```diff
@@ -942,13 +943,13 @@ public void export(ListDefinitionForm form, HttpServletResponse response, BindEx
...
+            // Issue 47289: Export List Archive if the user is an Admin of the folders of the selected Lists, else throw Permission error
             for (Pair<Integer, Container> pair : getListIdContainerPairs(selection, c, errorMessages))
             {
-                if (pair.second != c)
+                if (!pair.second.hasPermission(getUser(), AdminPermission.class))
                 {
-                    errorMessages.add(String.format("Cannot export lists defined in %s from %s. List export is only supported for lists defined in the current folder.", pair.second.getPath(), c.getName()));
-                    break;
+                    String listName = Objects.requireNonNull(ListManager.get().getList(pair.second, pair.first)).getName();
+                    throw new UnauthorizedException(String.format("You do not have the permission to export List '%s' from Folder '%s'.", listName, pair.second.getName()));
```
labkey-jeckels:
```suggestion
                    throw new UnauthorizedException(String.format("You do not have the permission to export List '%s' from Folder '%s'.", listName, pair.second.getPath()));

```

list/src/org/labkey/list/controllers/ListController.java
```diff
@@ -942,13 +943,13 @@ public void export(ListDefinitionForm form, HttpServletResponse response, BindEx
             Set<String> selection = DataRegionSelection.getSelected(form.getViewContext(), false);
             List<Integer> listIDs = new ArrayList<>();
 
-            // List export is only supported for lists defined in the current folder
+            // Issue 47289: Export List Archive if the user is an Admin of the folders of the selected Lists, else throw Permission error
             for (Pair<Integer, Container> pair : getListIdContainerPairs(selection, c, errorMessages))
             {
-                if (pair.second != c)
+                if (!pair.second.hasPermission(getUser(), AdminPermission.class))
```
labkey-jeckels: This action requires DesignListPermission in the current folder. Seems like we should check for that instead of AdminPermission.

labkey-jeckels (changes requested):
Two small suggestions and a question:

What do we end up exporting for lists defined in another container? The definition and all rows? Rows from the current container? Rows from the container in which it's defined? I remember talking about this original behavior with @labkey-nicka so tagging him here.

labkey-bpatel:
> Two small suggestions and a question:
> 
> What do we end up exporting for lists defined in another container? The definition and all rows? Rows from the current container? Rows from the container in which it's defined? I remember talking about this original behavior with @labkey-nicka so tagging him here.

Great point, it looks like I need to visit that part, currently, it only exports the list in the current container even though we gather selected ListIds from all the containers that user has DesignListPermission to.

list/src/org/labkey/list/controllers/ListController.java
labkey-bpatel: done

list/src/org/labkey/list/controllers/ListController.java
labkey-bpatel: done

labkey-bpatel:
@labkey-jeckels @labkey-nicka 

I am still revising, but please review the changes for 'Export List Archive'.

I still need to handle the case where Lists has the same name in different containers. Strangely, List Ids are not unique in different containers. So exporting of Lists with the same Name and the same Id in different containers doesn't currently work - working on fixing this.

Also, unrelated to my changes - the link to the List doesn't work as expected if the List is in a different container, so looking into this as well.

labkey-nicka:
> I am still revising, but please review the changes for 'Export List Archive'.
> 
> I still need to handle the case where Lists has the same name in different containers. Strangely, List Ids are not unique in different containers. So exporting of Lists with the same Name and the same Id in different containers doesn't currently work - working on fixing this.
> 
> Also, unrelated to my changes - the link to the List doesn't work as expected if the List is in a different container, so looking into this as well.

[Issue 45265](https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=45265) was opened to address a previous discrepancy in what is included in list export. As for the linking I'm happy to discuss this, but as you pointed out the IDs are only unique within a folder, so the linking is built to account for this. The difference being if the URL is `listId` parameter based or `name` based.

labkey-jeckels:
@labkey-bpatel your changes look like they're heading in a good direction. I'll hold a full review until you've pushed more updates based on our conversation this afternoon

labkey-bpatel:
> > I am still revising, but please review the changes for 'Export List Archive'.
> > I still need to handle the case where Lists has the same name in different containers. Strangely, List Ids are not unique in different containers. So exporting of Lists with the same Name and the same Id in different containers doesn't currently work - working on fixing this.
> > Also, unrelated to my changes - the link to the List doesn't work as expected if the List is in a different container, so looking into this as well.
> 
> [Issue 45265](https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=45265) was opened to address a previous discrepancy in what is included in list export. As for the linking I'm happy to discuss this, but as you pointed out the IDs are only unique within a folder, so the linking is built to account for this. The difference being if the URL is `listId` parameter based or `name` based.

@labkey-nicka As per our chat today, I am leaving the List URLs as-is in this PR, and have created this ticket: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47663

labkey-bpatel:
> @labkey-bpatel your changes look like they're heading in a good direction. I'll hold a full review until you've pushed more updates based on our conversation this afternoon

@labkey-jeckels you had asked about the same List Ids in the export settings.xml:
```
  <list name="List 1" id="1"/>
  <list name="List 2" id="1"/>
```
When we import that exported List in the Folder that already has a List with id=1, then it creates List 1 with a new id=2 and List 2 with a new id=3.

I've requested review again.

api/src/org/labkey/api/admin/FolderExportContext.java
```diff
@@ -170,4 +174,14 @@ public void setListIds(Integer[] listIds)
     {
         _listIds = new HashSet<>(Arrays.asList(listIds));
     }
+
+    public List<Pair<Integer, Container>> getLists()
```
labkey-jeckels, line 187: Can we migrate the uses of `setListIds()` and `getListIds()`? Or at least comment when you should use them instead of the new `getLists()` and `setLists()`?

labkey-jeckels (approved):
One more suggestion but no additional review needed unless you want it.

api/src/org/labkey/api/admin/FolderExportContext.java
labkey-bpatel: Comments added.
I tried creating a new ListExportContext to migrate List specific getters and setters, however, it started to get a bit more involved than a simple move, so resorted to adding comments.

PR #4271, opened 2023-04-04
Migrate QueryException and ReportsController

labkey-adam:
#### Rationale
Move to the new library


api/src/org/labkey/api/reports/report/view/ReportUtil.java
```diff
@@ -668,7 +666,7 @@ public static String makeExceptionString(Exception e, String formatString)
     /**
      * Generic form that can be used for serializing report information via json
      */
-    public static class JsonReportForm extends ReturnUrlForm implements CustomApiForm
+    public static class JsonReportForm extends ReturnUrlForm implements NewCustomApiForm
```
labkey-matthewb, line 669: I realize these are old, but what awful class names: CustomApiForm and NewCustomApiForm

labkey-matthewb (approved):

api/src/org/labkey/api/reports/report/view/ReportUtil.java
labkey-adam: Yeah, once we've eliminated `CustomApiForm` (the old one) I plan to rename `NewCustomApiForm` to something better

PR #4274, opened 2023-04-05
Initial ExperimentController.MoveSamplesAction

cnathe:
#### Rationale
This story will add the API action to do the sample move for the simple case, i.e. those samples that don’t have any parent lineage. The story will also put in place some of the initial validation checks on what the targetContainer can be and what samples can be moved. 

#### Changes
* add initial version of ExperimentController.MoveSamplesAction
  * validation of the targetContainer API param and the sample rowIds API param
  * only allows move from parent project to child, from child to parent project, or from child to sibling
  * updates container value in exp.material, exp.object, and exp.materialaliasmap tables for given sample RowIds
* configure platform/experiment module for jest integration tests
* add jest integration tests for success and error cases of MoveSamplesAction


experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1619,4 +1624,76 @@ private Map<Integer, List<Pair<Double, String>>> getSampleAliquotAmounts(Collect
...
+
+        return materialRowsUpdated;
+    }
+
+    private int materialRowContainerUpdate(List<Integer> sampleIds, Container targetContainer, User user)
+    {
+        TableInfo materialTable = getTinfoMaterial();
+        SQLFragment materialUpdate = new SQLFragment("UPDATE ").append(materialTable)
+                .append(" SET container = ?").add(targetContainer.getEntityId())
+                .append(", modified = ?").add(new Date())
```
labkey-susanh: These queries should probably use the new `addValue` methods instead of `add`.

cnathe: done

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
```diff
@@ -0,0 +1,316 @@
...
+    editorUserOptions = { requestContext: await server.createRequestContext(editorUser) };
+
+    const subEditorUser = await server.createUser('test_subeditor@expctrltest.com', 'pwSuperA1!');
+    await server.addUserToRole(subEditorUser.username, SecurityRole.Reader, PROJECT_NAME);
+    await server.addUserToRole(subEditorUser.username, SecurityRole.Editor, subfolder1.path);
+    await server.addUserToRole(subEditorUser.username, SecurityRole.Editor, subfolder2.path);
+    subEditorUserOptions = { requestContext: await server.createRequestContext(subEditorUser) };
+
+    const authorUser = await server.createUser('test_author@expctrltest.com', 'pwSuperA1!');
+    await server.addUserToRole(authorUser.username, SecurityRole.Author, PROJECT_NAME);
```
labkey-susanh, line 43: The Author role does not seem like it is used all that often and at least it's documentation says it is limited to support on the message boards. Is it useful as a testing role here?

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
```diff
@@ -0,0 +1,316 @@
...
+                rowIds: [sampleRowId]
+            }, { ...topFolderOptions, ...editorUserOptions }).expect(400);
+
+            // Assert
+            const { exception, success } = response.body;
+            expect(success).toBe(false);
+            expect(exception).toEqual('All samples must be from the current container for the move operation.');
+
+            const sampleExistsInTop = await sampleExists(sampleRowId, topFolderOptions);
+            expect(sampleExistsInTop).toBeFalsy();
```
labkey-susanh: We are actually expecting "false" here, right? Any reason to have the looser check here and below?

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
```diff
@@ -0,0 +1,316 @@
...
+            const { samplesMoved, success } = response.body;
+            expect(success).toBe(true);
+            expect(samplesMoved).toEqual(1);
+
+            const sampleExistsInSub1 = await sampleExists(sampleRowId, subfolder1Options);
+            expect(sampleExistsInSub1).toBeFalsy();
+            const sampleExistsInSub2 = await sampleExists(sampleRowId, subfolder2Options);
+            expect(sampleExistsInSub2).toBeTruthy();
+        });
+    });
```
labkey-susanh, line 350: Is it worth having a happy path test for moving more than one sample?

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
```diff
@@ -0,0 +1,316 @@
...
+});
```
labkey-susanh, line 351: Does the container get deleted automatically after the test is done? If not, it might be useful to add a test that deletes containers that have had samples moved from them. (Locally, when I hadn't updated the object table but had updated the exp.materials table and deleted the source container a SQL exception was thrown.)

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -7635,4 +7635,132 @@ public ModelAndView getView(Object o, BindException errors) throws SQLException
...
+        {
+            if (current.isRoot() || target.isRoot())
+                return false;
+
+            if (current.equals(target))
+                return false;
+
+            boolean moveFromProjectToChild = current.isProject() && target.getParent().equals(current);
+            boolean moveFromChildToProject = !current.isProject() && current.getParent().equals(target);
+            boolean moveFromChildToSibling = !current.isProject() && current.getParent().equals(target.getParent());
```
labkey-susanh: This will technically allow moving from a sub-subfolder to a subfolder (and other various scenarios involving sub-subfolders).  Seems fine though as even in those cases (which can't be achieved through the app UI) it achieves the functionality we want.

labkey-susanh (approved):

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
cnathe: I was using it here as a way to give InsertPermissions without UpdatePermissions.

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
cnathe: Yes, expecting `true` or `false` explicitly. Changed to match that.

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
cnathe: yes, added

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
cnathe: oh good idea! I'll add that.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
cnathe: Yeah, I thought of that too but talked myself out of adding the checks re: sub-subfolders. It shouldn't actually be that hard and will keep the supported scenarios smaller for now, so I'll add it. Thanks.

PR #4276, opened 2023-04-06
Update AbstractForeignKey to use the target container, rather than source, for ContainerFilter

bbimber:
@labkey-jeckels: this is related to https://github.com/LabKey/ehrModules/pull/574

I believe this is probably longstanding behavior exposed when creating a cross-container FK in LabKey. If one does not specify the target container on creation, this code will default back to the container of the source schema, rather than the target schema.

Within this PR the actual change of consequence is really just one method in AbstractForeignKey. I renamed the getSchema() method to getLookupSchema() to make it more clear that it is different from the existing getSourceSchema() method. I dont know if this will break the build in other modules I dont have locally. There's one change in LaboratoryModule associated with the method name change.

### DIFF
--- a/api/src/org/labkey/api/data/AbstractForeignKey.java
+++ b/api/src/org/labkey/api/data/AbstractForeignKey.java
@@ -116,19 +116,26 @@ public abstract class AbstractForeignKey implements ForeignKey, Cloneable
             return new ContainerFilter.SimpleContainerFilterWithUser(getLookupUser(), c);
         }
         ContainerFilter cf = null;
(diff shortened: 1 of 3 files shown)


bbimber:
@labkey-martyp or @labkey-jeckels: any comments on this issue? any chance someone could mirror this branch to the LK fork? I'd really like to see how tests behave. thanks in advance.

bbimber:
closing in favor ok the LK fork version: https://github.com/LabKey/platform/pull/4307

PR #4279, opened 2023-04-06
Generalize assay consistency check

labkey-klum:
#### Rationale
A prior commit added the ability to run diagnostic queries against assay results to determine if any discrepancies had entered between thaw list, assay results and linked to study data. This PR moves some of that code into the `platform` module so that is can potentially be used in other scenarios.

[original PR](https://github.com/LabKey/nihs/pull/39)

#### Related Pull Requests
https://github.com/LabKey/nihs/pull/48


api/src/org/labkey/api/assay/AbstractAssayProvider.java
```diff
@@ -16,6 +16,7 @@
 
 package org.labkey.api.assay;
 
+import org.apache.logging.log4j.Logger;
```
labkey-jeckels: Are these and other import changes intentional and needed?

assay/src/org/labkey/assay/AssayManager.java
```diff
@@ -908,4 +912,64 @@ List<AssayProvider> getRegisteredAssayProviders()
...
+                if (!errors.isEmpty())
+                {
+                    errors.forEach(ve -> log.error(ve.getMessage()));
+                    return null;
+                }
+
+                return checker.getValidationSql(container, user, protocol, assayDataTable);
+            }
+            else
+                log.error("Assay data table not found for protocol : " + protocol.getName());
```
labkey-jeckels: Logging the container path is often helpful in scenarios like this

study/src/org/labkey/study/query/DatasetQueryView.java
```diff
@@ -686,19 +698,47 @@ public void setVisitRowId(int visitRowId)
...
+        private Set<String> _divergedRows = Collections.emptySet();
+        private ColumnInfo _lsid;
+
+        public DatasetDataRegion(Container container, User user, DatasetDefinition dataset)
+        {
+            if (dataset.isPublishedData() && ExperimentalFeatureService.get().isFeatureEnabled(EXPERIMENTAL_LINKED_DATASET_CHECK))
+            {
+                Dataset.PublishSource publishSource = dataset.getPublishSource();
+                // currently limit this to assay linked records
+                if (publishSource.getSourceType().equals("assay"))
```
labkey-jeckels: Worth a null check, or will this always be non-null if `isPublishedData()` is true?

study/src/org/labkey/study/query/DatasetQueryView.java
```diff
@@ -686,19 +698,47 @@ public void setVisitRowId(int visitRowId)
...
+                    if (expObject instanceof ExpProtocol protocol)
+                    {
+                        LinkedResultsChecker helper = new LinkedResultsChecker(user, dataset);
+                        Collection<String> results = AssayService.get().checkResults(container, user, protocol, helper, String.class);
+                        if (!results.isEmpty())
+                        {
+                            _divergedRows = new HashSet<>(results);
+                            _lsid = dataset.getTableInfo(user).getColumn(FieldKey.fromParts("Lsid"));
+
+                            addMessageSupplier(x -> List.of(new DataRegion.Message("Highlighted rows may have different subject or " +
```
labkey-jeckels, line 733: This says "may have different". Are there cases where we'd report a mismatch even though the data's in sync? Maybe when the data was absent on the assay side and filled in by a user when linking to the dataset?

study/src/org/labkey/study/query/DatasetQueryView.java
```diff
@@ -686,19 +698,47 @@ public void setVisitRowId(int visitRowId)
...
+                {
+                    ExpObject expObject = publishSource.resolvePublishSource(dataset.getPublishSourceId());
+                    if (expObject instanceof ExpProtocol protocol)
+                    {
+                        LinkedResultsChecker helper = new LinkedResultsChecker(user, dataset);
+                        Collection<String> results = AssayService.get().checkResults(container, user, protocol, helper, String.class);
+                        if (!results.isEmpty())
+                        {
+                            _divergedRows = new HashSet<>(results);
+                            _lsid = dataset.getTableInfo(user).getColumn(FieldKey.fromParts("Lsid"));
```
labkey-jeckels, line 731: I would probably have used the RowId from the assay data instead of the much longer LSID.

study/src/org/labkey/study/query/DatasetQueryView.java
```diff
@@ -750,6 +790,87 @@ protected void addHeaderMessage(StringBuilder headerMessage, RenderContext ctx)
...
+            }
+
+            @Override
+            public SQLFragment getValidationSql(Container container, User user, ExpProtocol protocol, TableInfo dataTable)
+            {
+                var sqs = StudyQuerySchema.createSchema(_dataset.getStudy(), user, null);
+                TableInfo datasetTable = sqs.getDatasetTable(_dataset, ContainerFilter.EVERYTHING);
+
+                String studyVisit = _dataset.getStudy().getTimepointType().isVisitBased() ? "SequenceNum" : "Date";
+                if (datasetTable instanceof FilteredTable<?> filteredTable)
```
labkey-jeckels, line 851: It's probably out of scope for this work, but I'd love to see this be injected as an ExprColumn on AssayDatasetTable instead of being a side query. That would let the user filter for good/bad rows, and not pull in a potentially huge number of mismatched row identifiers into memory.

labkey-jeckels (approved):
Assorted ideas/suggestions but I'm OK as-is since it's an experimental feature.

study/src/org/labkey/study/query/DatasetQueryView.java
labkey-klum: correct, or if the user changes or corrects the value during the link to study workflow

PR #4280, opened 2023-04-07
Issue 47144: Handle large files on Panorama Public via Symlinks

labkey-martyp:
#### Rationale
Add file deleted and file replaced to File Listeners so Panorama Public can update symlinks. Add ExperimentService.getExpDatasAtAndUnderPath for export directory cleanup.

#### Related Pull Requests
* https://github.com/LabKey/MacCossLabModules/pull/344

#### Changes
* Add file deleted and file replaced events and handlers
* Add ExperimentService.getExpDatasAtAndUnderPath
* When adding files to archive don't use canonical path for symlinks as there is different behavior depending on the OS


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -7757,6 +7757,30 @@ public List<ExpDataImpl> getExpDatasUnderPath(@NotNull File path, @Nullable Cont
...
+        List<ExpDataImpl> childDatas = ExpDataImpl.fromDatas(new TableSelector(getTinfoData(), filter, null).getArrayList(Data.class));
+
+        // Include exp.data at the path itself
+        if (prefix.endsWith("/"))
+            prefix = prefix.substring(0, prefix.length() - 1);
+
+        filter = new SimpleFilter();
+        filter.addCondition(FieldKey.fromParts("datafileurl"), prefix);
+
+        childDatas.addAll(ExpDataImpl.fromDatas(new TableSelector(getTinfoData(), filter, null).getArrayList(Data.class)));
```
labkey-jeckels: I don't think we need this separate query, since "/myParent/myFile" both starts with and is equal to "/myParent/myFile".

It would be nice to consolidate this with `getExpDatasUnderPath()` by adding a boolean argument for whether it should return exact matches or not, which would control if it adds the `CompareType.NEQ` filter

api/src/org/labkey/api/files/FileContentService.java
```diff
@@ -288,6 +288,21 @@ default int fireFileMoveEvent(@NotNull Path src, @NotNull Path dest, @Nullable U
         return 0;
     }
 
+    /** Notifies all registered FileListeners that a file or directory has been replaced */
+    default void fireFileReplacedEvent(@NotNull File replaced, @Nullable User user, @Nullable Container container){}
```
labkey-jeckels: I know that we have File/Path pairs for other methods here, but I think we should be consolidating on Path variants going forward. Consider dropping the new File ones to help steer people towards the desired approach

labkey-jeckels (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-martyp: The reason for both queries is the path parameter is a directory so it will be /myParent/myFile/ which works great for getting the child datas. The last separator ensures datas starting with /myParent/myFile2 are not in result. Then the last file separator is removed to get the datas with datafileurl set to the current path /myParent/myFile and adds those to the list.

I combined it with getExpDatasUnderPath and added the boolean.

api/src/org/labkey/api/files/FileContentService.java
labkey-martyp: Ok done.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-jeckels: Ah, right. I wasn't thinking about siblings that might be a variant of the name. Two queries is likely fine here, but for future reference `OrClause` can come in handy in cases like this.

PR #4283, opened 2023-04-09
Test to validate that actions and forms are declared correctly

labkey-adam:
#### Rationale
I came across one registered action that can't be instantiated and a few actions that don't declare their command classes correctly. Let's keep ourselves honest with a validation test.

Stop registering `PlateBasedUploadWizardAction` as an action class... it's abstact!

Specify a valid command class method name in `ExportSVGAction`

### DIFF
--- a/api/src/org/labkey/api/module/actionAndFormTest.jsp
+++ b/api/src/org/labkey/api/module/actionAndFormTest.jsp
@@ -37,7 +37,7 @@
                         }
                         catch (Exception e)
                         {
-                            errorMessages.add(e.getMessage() + " while attempting to construct " + ad.getActionClass());
+                            errorMessages.add(e.getMessage() + " while attempting to construct " + ad.getActionClass() + " from controller " + controller);
                             return null;
                         }
                     })
--- a/api/src/org/labkey/api/module/actionAndFormTest.jsp
+++ b/api/src/org/labkey/api/module/actionAndFormTest.jsp
@@ -61,20 +61,20 @@
 
         // Identify all form getter methods that return an old JSONObject (or an array or list of them). This is useful
         // for the JSONObject migration process.
-        for (Class<?> formClass : formClasses)
-        {
-            for (Method method : formClass.getDeclaredMethods())
-            {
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/module/actionAndFormTest.jsp
```diff
@@ -0,0 +1,91 @@
...
+<%@ page import="java.util.LinkedList" %>
+<%@ page import="java.util.List" %>
+<%@ page import="java.util.Objects" %>
+<%@ page import="java.util.Set" %>
+<%@ page import="java.util.TreeSet" %>
+<%@ page import="static org.junit.Assert.*" %>
+<%@ page extends="org.labkey.api.jsp.JspTest.DRT" %>
+
+<%!
+    private final Logger _log = LogHelper.getLogger(getClass(), "");
```
labkey-jeckels:
```suggestion
    private final Logger _log = LogHelper.getLogger(getClass(), "Testing of action and form classes");
```

api/src/org/labkey/api/module/actionAndFormTest.jsp
```diff
@@ -0,0 +1,91 @@
...
+                            errorMessages.add(e.getMessage());
+                            return Object.class;
+                        }
+                    })
+                    .filter(aClass -> Object.class != aClass)
+                    .forEach(formClasses::add);
+            }
+        }
+
+        // Identify all form getter methods that return an old JSONObject (or an array or list of them). This is useful
```
labkey-jeckels, line 71: Did you want to keep this around?

labkey-jeckels (approved):

api/src/org/labkey/api/module/actionAndFormTest.jsp
labkey-adam: Yes, at least until the `JSONObject` migration is complete. This highlights the forms that need to be migrated and tested in concert with migrating `BaseApiAction.populateForm()`, etc. to use the new JSONObject.

PR #4289, opened 2023-04-11
Sample status validation on ExperimentController.MoveSamplesAction

cnathe:
#### Rationale
ExperimentController.MoveSamplesAction to check for invalid sample status
- don't allow move of Locked sample statuses
- don't allow move if the sample status would not be valid in the new container's scope

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/4289
- https://github.com/LabKey/sampleManagement/pull/1759

#### Changes
- ExperimentController.MoveSamplesAction updates to add validation based on the selected sample's status
- Replace usages of DataStateManager.getInstance().getStateForRowId() with version that uses getAllProjectStates()

### DIFF
--- a/experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
+++ b/experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
@@ -59,11 +59,20 @@ afterAll(async () => {
     return server.teardown();
 });
 
-async function createSample(sampleName: string, folderOptions: RequestOptions) {
+async function createSample(sampleName: string, folderOptions: RequestOptions, sampleState?: number) {
     const materialResponse = await server.post('query', 'insertRows', {
         schemaName: 'samples',
         queryName: SAMPLE_TYPE_NAME,
-        rows: [{ name: sampleName }]
+        rows: [{ name: sampleName, sampleState }]
+    }, { ...folderOptions, ...editorUserOptions }).expect(successfulResponse);
+    return caseInsensitive(materialResponse.body.rows[0], 'rowId');
+}
+
+async function createStatus(label: string, folderOptions: RequestOptions, stateType = 'Available') {
+    const materialResponse = await server.post('query', 'insertRows', {
+        schemaName: 'core',
+        queryName: 'DataStates',
+        rows: [{ label, stateType, publicdata: true }]
     }, { ...folderOptions, ...editorUserOptions }).expect(successfulResponse);
     return caseInsensitive(materialResponse.body.rows[0], 'rowId');
 }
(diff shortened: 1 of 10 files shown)


experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
```diff
@@ -294,7 +415,8 @@ describe('ExperimentController', () => {
 
         it('success, move sample from subfolder to sibling', async () => {
             // Arrange
-            const sampleRowId = await createSample('sub1-movetosub2-1', subfolder1Options);
+            const statusRowId = await createStatus('parent-Available-3', topFolderOptions);
+            const sampleRowId = await createSample('sub1-movetosub2-1', subfolder1Options, statusRowId);
```
labkey-susanh: Are we still testing the case that a sample has no status at all?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -7803,6 +7802,31 @@ private void validateSampleIds(MoveSamplesForm form, Errors errors)
...
+                if (sampleStatus == null) continue;
+
+                // prevent move for locked samples
+                if (!material.isOperationPermitted(SampleTypeService.SampleOperations.Move))
+                {
+                    invalidStatusSamples.add(material);
+                }
+                // prevent moving samples if data QC state doesn't exist in target container scope (i.e. home project),
+                // only applies when moving from child to parent or child to sibling
+                else if (!getContainer().isProject() && sampleStatus.getContainer().equals(getContainer()))
```
labkey-susanh, line 7820: If status is defined in the /Shared container, I think we would also allow that. We do return the statuses from /Shared in getAllStates, so probably we should check for that.

labkey-susanh (changes requested):

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
cnathe: This should already cover the /Shared container defined statuses case (though I should update the comment). This check is for the not allowed case (i.e. if the current container is not a LK project and the sample's status is defined in that container).

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
cnathe: I can add that.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: Ah yes, you're right. My bad.

labkey-susanh (approved):

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
cnathe: looks like we do already have the " successfully move sample without a status" case in the ispec.ts file.

PR #4291, opened 2023-04-11
Issue 47509: Resolve sample names that are integers during assay data import

labkey-susanh:
#### Rationale
[Issue 47509](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47509)

Sometimes sample names are composed of only integers. These are difficult to distinguish from rowIds when resolving the lookup to a sample during assay import.  In related issues for sample import using integer-name samples, we updated the logic for resolving samples to first look for the sample assuming the integer value is its name instead of rowId.  We adopt similar logic here when importing via a file since it is rare that users will provide rowId values for the samples being referenced. If they do, they will surely also have the name of the sample available to use during assay data import in cases where there might be ambiguities based on numeric naming patterns.

#### Related Pull Requests
- https://github.com/LabKey/sampleManagement/pull/1762

#### Changes
* Update `checkData` method in `AbstractAssayTsvDataHandler` to go through `exp.findExpMaterial` even for integer values
* Update `AbstractAssayTsvDataHandler.importFile` to make sure the sampleId data being imported matches the resolved MaterialInputs from `checkData`.

### DIFF
--- a/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
+++ b/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
@@ -1051,7 +1051,7 @@ public abstract class AbstractAssayTsvDataHandler extends AbstractExperimentData
                         materialInputs.putIfAbsent(material, pd.getName());
                         rowInputLSIDs.add(material.getLSID());
                     }
-                    else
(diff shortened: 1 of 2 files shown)


cnathe (approved):

cnathe:
Closing as we need a slightly different approach to handle the sample rowId scenario that comes from the App editable grid case.

PR #4292, opened 2023-04-11
Clone permissions to an existing user

labkey-adam:
#### Rationale
Today, admins can "clone permissions" from an existing user when adding new user accounts. This PR adds the ability to clone permissions to an _existing_ user, which means the target user's current group memberships and role assignments are deleted and replaced with those of the source user. A key use case: a user leaves a particular organizational role and a new user takes on that role; old user's permissions are then cloned to the new user.

https://docs.google.com/document/d/1Px2H4Q2JSgYtXLBX9pMLQUn07SmoRSXl_sVz78sU5nA/edit


api/src/org/labkey/api/audit/provider/GroupAuditProvider.java
```diff
@@ -96,13 +96,13 @@ public String getEventName()
     @Override
     public String getLabel()
     {
-        return "Group events";
+        return "Group and role events";
```
labkey-tchad, line 99: @labkey-sweta This will require an update to `AuditLogTest` and `TroubleshooterRoleTest`

labkey-adam: Good point... thanks, @labkey-tchad

labkey-sweta: Updated the test with new group event label 
https://github.com/LabKey/testAutomation/pull/1488

core/webapp/clonePermissions.js
```diff
@@ -0,0 +1,44 @@
+function showUserAccess()
```
labkey-klum, line 1: Not required for this code review but in general it's a good habit to not bleed functions into the global namespace (at least this is what Nick tells me). The prescribed pattern is to wrap it in a function, there are a bunch of examples of this but take a look at `NLPUtil.js`.

labkey-klum (approved):

core/webapp/clonePermissions.js
labkey-adam: Yeah, I've heard that too 😄

PR #4296, opened 2023-04-12
Migrate study & survey to new JSONObject library

labkey-adam:
#### Rationale
The continuing saga. Migrate all remaining old `JSONObject` use in study & survey. Migrate all but one use of `CustomApiForm`.


pipeline/src/org/labkey/pipeline/PipelineController.java
```diff
@@ -1588,31 +1574,31 @@ public Object execute(PipelineTriggerForm form, BindException errors) throws Exc
...
         @Override
-        public void bindProperties(Map<String, Object> props)
+        public void bindJson(JSONObject json)
         {
-            MutablePropertyValues params = new MutablePropertyValues(props);
+            MutablePropertyValues params = new BaseApiAction.JsonPropertyValues(json);
             BaseViewAction.defaultBindParameters(this, "form", params);
 
-            Object assayProvider = props.get("assay provider");
+            Object assayProvider = json.get("assay provider");
```
labkey-jeckels: Should these be `opt` methods? Since we're null checking or instanceof checking the values, won't that fail if they're not present?

labkey-jeckels (approved):

pipeline/src/org/labkey/pipeline/PipelineController.java
labkey-adam: Yes. I try to switch to `opt()`, `optJSONArray()`, etc. in these cases. Thanks for catching these.

labkey-adam: Fixed

PR #4300, opened 2023-04-13
Move and create audit logs for samples that are moving

labkey-susanh:
#### Rationale
When we move samples from one container to another, we want to move their audit logs along with them and add new audit entries to log the move, when the audit behavior indicates that we should. We also want to move the associated inventory items and their audit logs when the inventory module is present.

#### Related Pull Requests

- https://github.com/LabKey/inventory/pull/822
- https://github.com/LabKey/sampleManagement/pull/1772

#### Changes
* Update MoveSamples API action to return counts for various data objects that are updated as a result of the call
* Add summary and detailed audit log entries to document the move, if audit behavior type indicates we should
* Move the sample timeline audit logs for the samples that are moved.


experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
```diff
@@ -72,12 +73,72 @@ async function sampleExists(sampleRowId: number, folderOptions: RequestOptions)
...
+
+    async function verifySampleTypeAuditLogs(sourceFolderOptions: RequestOptions, targetFolderOptions: RequestOptions, sampleIds: number[], userComment: string = null): Promise<number>
+    {
+        const sampleTypeEventsInSource = await getSampleTypeAuditLogs(SAMPLE_TYPE_NAME, sourceFolderOptions, 2);
+        const samplesPhrase = sampleIds.length == 1 ? "1 sample" : sampleIds.length + " samples";
+        const targetPath = targetFolderOptions.containerPath.charAt(0) === '/' ? targetFolderOptions.containerPath : '/' + targetFolderOptions.containerPath;
+        const sourcePath = sourceFolderOptions.containerPath.charAt(0) === '/' ? sourceFolderOptions.containerPath : '/' + sourceFolderOptions.containerPath;
+        expect(caseInsensitive(sampleTypeEventsInSource[0], 'Comment')).toEqual("Moved " + samplesPhrase + " to " + targetPath);
+        const transactionId = caseInsensitive(sampleTypeEventsInSource[0], 'transactionId');
+        expect(transactionId).toBeTruthy();
```
cnathe, line 123: I was thrown off by the `toBeTruthy` check here. Is this a way to check that the transactionId is defined and > 0?

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
```diff
@@ -72,12 +73,72 @@ async function sampleExists(sampleRowId: number, folderOptions: RequestOptions)
...
+    async function verifyDetailedAuditLogs(sourceFolderOptions: RequestOptions, targetFolderOptions: RequestOptions, sampleIds: number[], transactionId: number, userComment: string = null) {
+        for (const sampleRowId of sampleIds) {
+            const sampleEventsInSource = await getSampleTimelineAuditLogs(sampleRowId, sourceFolderOptions);
+            expect(sampleEventsInSource).toHaveLength(0);
+        }
+
+        for (const sampleRowId of sampleIds) {
+            const sampleEventsInTarget = await getSampleTimelineAuditLogs(sampleRowId, targetFolderOptions);
+            expect(sampleEventsInTarget).toHaveLength(2);
+            console.log(sampleEventsInTarget);
```
cnathe: keep or remove console.log?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1624,49 +1634,86 @@ private Map<Integer, List<Pair<Double, String>>> getSampleAliquotAmounts(Collect
...
+                }
+                String samplesPhrase = StringUtilsLabKey.pluralize(samples.size(), "sample");
+                addSampleTypeAuditEvent(user, sourceContainer, sampleType, transaction.getAuditId(),
+                        "Moved " + samplesPhrase + " to " + targetContainer.getPath(), userComment, "moved samples out");
+                addSampleTypeAuditEvent(user, targetContainer, sampleType, transaction.getAuditId(),
+                        "Moved " + samplesPhrase  + " from " + sourceContainer.getPath(), userComment, "moved samples in");
+
+                // move the events associated with the samples that have moved
+                SampleTimelineAuditProvider auditProvider = new SampleTimelineAuditProvider();
+                updateCounts.put("sampleAuditEvents", auditProvider.moveEvents(targetContainer, sampleIds));
```
cnathe: this is within the for loop for the sample types right? won't this then keep replacing the `sampleAuditEvents` count instead of getting the sum?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1624,49 +1634,86 @@ private Map<Integer, List<Pair<Double, String>>> getSampleAliquotAmounts(Collect
...
+
+                // create summary audit entries for the source and target containers
```
cnathe: should this comment say "create detailed ..." instead of "create summary ..."?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1624,49 +1634,86 @@ private Map<Integer, List<Pair<Double, String>>> getSampleAliquotAmounts(Collect
...
+                auditBehavior = ((ExpSampleTypeImpl) sampleType).getTinfo().getAuditBehavior(auditBehavior);
```
cnathe: I was wondering why we needed to split and loop over the sample types in this PR, but I see now. It is because of the possibility that each sample type could have a different default audit behavior set, right?

cnathe (approved):

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
labkey-susanh: Yes. This just says "We set a value for the transactionId" and below I check that the other log(s) have the same id.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Ah shoot. I fixed the inventory ones but not the sample ones. Done.

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
labkey-susanh: Ah yes. Remove now.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Yes, and because the sample type audit logs are per sample type.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: It should move to above the summary events. Done.

PR #4303, opened 2023-04-14
Migrate ExperimentJSONConverter and AssayJSONConverter

labkey-adam:
#### Rationale
You guessed it, more JSONObject migration. In this case, ExperimentJSONConverter, AssayJSONConverter, and associated users of these classes.


pipeline/src/org/labkey/pipeline/PipelineUpgradeCode.java
```diff
@@ -47,26 +48,25 @@ public static void updateSampleReloadTaskMergeDataOptions(ModuleContext ctx) thr
...
             {
                 Object customConfiguration = trigger.getCustomConfiguration();
                 if (customConfiguration != null && !customConfiguration.toString().equals("") && customConfiguration.toString().contains(oldKey))
                 {
-                    JSONObject json = null;
                     try
                     {
-                        ObjectMapper mapper = new ObjectMapper();
-                        json = mapper.readValue(customConfiguration.toString(), JSONObject.class);
+                        // Note: This is overkill -- new JSONObject(customConfiguration.toString()) should work
```
labkey-jeckels, line 58: Is it worth retaining the overkill?

labkey-jeckels (approved):

pipeline/src/org/labkey/pipeline/PipelineUpgradeCode.java
labkey-adam: I think so, given that there's no automated testing of this code path and the method will be deleted before too long. I did some nominal testing of my refactor, but I'm trying to minimize the changes to minimize risk.

labkey-adam: My primary goal with the comment was to prevent propagating this pattern elsewhere.

PR #4304, opened 2023-04-14
Improve performance/caching of scriptContext for trigger scripts

bbimber:
I noticed an ETL was taking much longer than previously. When I profiled it in Jprofiler, I saw that this is probably due to constantly recompiling serverContext in RhinoService, which was added in #3942.. 

I believe this PR addresses that, by calculating serverContext once, and caching this instead of just caching ContainerUser. When I tried this locally, it restored perf back to close to where it was before. 

@labkey-nicka: I think you reviewed the original feature - thoughts on this?

### DIFF
--- a/api/src/org/labkey/api/data/triggers/ScriptTrigger.java
+++ b/api/src/org/labkey/api/data/triggers/ScriptTrigger.java
@@ -17,6 +17,7 @@ package org.labkey.api.data.triggers;
 
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.json.JSONObject;
 import org.labkey.api.data.Container;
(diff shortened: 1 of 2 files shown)


bbimber:
@labkey-nicka: my last commit is something I should have thought of the first time. rather than request the entire jsInitObject and serialize to a string, it makes a whole lot more sense to lazily grab it. I still think the second change here makes sense too. So in summary:

1) Cache the compiled ModuleScript
2) Within that script, lazily request jsInitObject, rather than request it upfront and store that big JSON string

Because I dont have commit permissions here, if you could mirror this branch to the labkey side of the fork, this would let tests run against it. Thanks.

bbimber:
closing in favor of the new branch on the labkey side of the fork: https://github.com/LabKey/platform/pull/4306

PR #4306, opened 2023-04-14
Improve performance for trigger scripts and serverContext

bbimber:
I originally opened this PR using a branch on bbimber/platform, but now I can make branches here, so I'm closing that PR in favor of this.

Background:

I noticed an ETL was taking much longer than previously. When I profiled it in Jprofiler, I saw that this is probably due to constantly recompiling serverContext in RhinoService, which was added in https://github.com/LabKey/platform/pull/3942..

I believe this PR addresses that, by calculating serverContext once, and caching this instead of just caching ContainerUser. When I tried this locally, it restored perf back to close to where it was before.

I also took it a step further. Rather than request the entire jsInitObject and serialize to a string every single time, it makes a whole lot more sense to lazily call jsInitObject only when code requests serverContext. So in summary:

- Cache the compiled ModuleScript
- Within that script, lazily request jsInitObject, rather than request it upfront and store that big JSON string

@labkey-nicka: I think you reviewed the original feature - thoughts on this?

### DIFF
--- a/api/src/org/labkey/api/data/triggers/ScriptTrigger.java
+++ b/api/src/org/labkey/api/data/triggers/ScriptTrigger.java
@@ -17,7 +17,6 @@ package org.labkey.api.data.triggers;
 
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
-import org.json.JSONObject;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.DbScope;
 import org.labkey.api.data.PHI;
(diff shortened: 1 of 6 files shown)


labkey-jeckels:
See more context in https://www.labkey.org/ONPRC/Support%20Tickets/issues-details.view?issueId=47705

api/src/org/labkey/api/data/triggers/ScriptTrigger.java
```diff
@@ -287,6 +279,42 @@ private <T> T _try(Container c, User user, Map<String, Object> extraContext, Scr
...
+            catch (URISyntaxException e)
+            {
+                throw new IllegalStateException("A URI exception should never occur since this defines the URI");
+            }
+        }
+    }
+
+    public static Script getServerContext(Container c, User u)
+    {
+        String jsCode = "org.labkey.api.util.PageFlowUtil.jsInitObject(org.labkey.api.writer.ContainerUser.create(org.labkey.api.data.ContainerManager.getForId('" + c.getId() + "'), org.labkey.api.security.UserManager.getUser(" + u.getUserId() + ")), null, null, false).toMap()";
```
labkey-nicka: nit: This seems like something that might be missed in a future refactor of these classes/methods. Ideally, these would be composed of Java references utilizing the package path (I have no idea how to do that -- something, something class loader). @labkey-jeckels may have an idea.

labkey-nicka (approved):

api/src/org/labkey/api/data/triggers/ScriptTrigger.java
labkey-jeckels: @labkey-nicka Just doing a class.getName() is enough. I pushed an update. It won't follow refactors of the method names, but it's a stronger reference than just having the string.

PR #4307, opened 2023-04-14
Update AbstractForeignKey to use the target container, rather than source, for ContainerFilter

bbimber:
@labkey-jeckels: this is related to https://github.com/LabKey/ehrModules/pull/574

I believe this is probably longstanding behavior exposed when creating a cross-container FK in LabKey. If one does not specify the target container on creation, this code will default back to the container of the source schema, rather than the target schema, in getLookupContainerFilter().

The only real change here is that in getLookupContainerFilter(), the code preferentially takes the container of the target user schema, not the source user schema. To make this clearer in the code, I renamed the getSchema() method to getLookupSchema(). There is also the existing getSourceSchema() method. This name change resulted in touching a few other bits of code, but those changes should have no meaningful change on behavor.

There's one change in LaboratoryModule associated with the method name change. I hope tests will catch other build breaks, but as of writing this I dont see any.

### DIFF
--- a/api/src/org/labkey/api/data/AbstractForeignKey.java
+++ b/api/src/org/labkey/api/data/AbstractForeignKey.java
@@ -15,19 +15,34 @@
  */
 package org.labkey.api.data;
 
+import org.apache.commons.collections.map.CaseInsensitiveMap;
 import org.apache.logging.log4j.LogManager;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.labkey.api.collections.CaseInsensitiveMapWrapper;
 import org.labkey.api.collections.NamedObjectList;
+import org.labkey.api.exp.list.ListDefinition;
+import org.labkey.api.exp.list.ListService;
+import org.labkey.api.exp.property.Domain;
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/data/AbstractForeignKey.java
```diff
@@ -101,8 +101,19 @@ protected AbstractForeignKey(QuerySchema sourceSchema, @Nullable ContainerFilter
...
+
+        return null;
+    }
+
+    protected QuerySchema getSourceSchema()
+    {
+        return _sourceSchema;
+    }
+
+    protected @Nullable QuerySchema getLookupSchema()
```
labkey-jeckels, line 132:
```suggestion
    /** @return the schema (in the target container) if this lookup explicitly targets something other than its parent schema and container */
    protected @Nullable QuerySchema getLookupSchema()

```

labkey-jeckels:
This change and the related EHR PR look fine to me. I've queued some extra test runs to confirm and plan to merge tomorrow unless there are surprises.

bbimber:
Thanks Josh.  I have an idea on a simple integration test to exercise this that I'll plan to write when I'm back. I'm all for merging this but if you prefer to keep it coordinated as one PR that's fine too.

bbimber:
@labkey-jeckels: test case now added. the tests passed prior to the addition of the test, I believe.

labkey-jeckels (approved):

PR #4309, opened 2023-04-14
Fix class cast exception in SurveyManager

labkey-adam:
#### Rationale
`SurveyTest` is failing due to a class cast exception.

This test pointed out that `SurveyManager` was incorrectly casting the "lookup" entry in the map returned by `JsonWriter.getMetaData()` to a new `JSONObject`. Fix by switching JsonWriter to use the new library. I could have gone with a more targeted fix, but this is the direction we want to move.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4296


labkey-jeckels (approved):

labkey-adam:
Pending SurveyTest run: https://teamcity.labkey.org/viewQueued.html?itemId=2395257

labkey-adam:
Daily C continues to crank away, but the SurveyTest passed so I'm merging...

PR #4321, opened 2023-04-18
TableInfo.getSQLName()

labkey-matthewb:
#### Rationale
SQLFragment version of TableInfo.getSelectName()

#### Related Pull Requests
* 

#### Changes
*


wiki/src/org/labkey/wiki/WikiManager.java
```diff
@@ -336,13 +336,13 @@ public void addAliases(Wiki wiki, @NotNull Collection<String> aliases, @Nullable
...
             SQLFragment sql = new SQLFragment("INSERT INTO ")
-                .append(CommSchema.getInstance().getTableInfoPageAliases().getSelectName())
+                .append(CommSchema.getInstance().getTableInfoPageAliases())
                 .append(" (Container, Alias, PageRowId) SELECT ?, ?, ?\n")
                 .add(wiki.getContainerId())
                 .add(alias)
                 .add(wiki.getRowId())
                 .append("WHERE NOT EXISTS (SELECT * FROM ")
-                .append(CommSchema.getInstance().getTableInfoPageAliases().getSelectName())
+                .append(CommSchema.getInstance().getTableInfoPageAliases().getFromSQL("__pa"))
```
labkey-jeckels: I'm not clear on why this is using `getFromSQL()` instead of just appending the `TableInfo` with an alias.

labkey-jeckels (approved):
Nice cleanup. One question but overall seems ready to go

wiki/src/org/labkey/wiki/WikiManager.java
labkey-matthewb: I'm not clear on that either.

PR #4312, opened 2023-04-15
Performance Evaluation for Data at Scale - Improve Insert/UpdateRows calls

XingY:
#### Rationale
- selectRows is called immediately after insert/update to select the full results as response. Often times, the full results in response is not needed. 

- Detail audit log is enabled for sample CUD actions for generating timelines. ExistingRecordDataIterator is used to query the previous sample detail so that the comparison against the updates can be generated. When merge/update samples, ExistingRecordDataIterator queries for the full sample details, including exp.material, joined by samples.sampleprovisioned, joined by calculated inventory storage tables. Then, the sample's lineage is queried for generating audit log for parent events. Often times during an update, only a small subset of columns are included in the rows data. For columns that aren't updated, there is no need to query for them in ExistingRecordDataIterator.

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/151
* https://github.com/LabKey/labkey-ui-components/pull/1175
* https://github.com/LabKey/labkey-ui-premium/pull/80
* https://github.com/LabKey/platform/pull/4312
* https://github.com/LabKey/inventory/pull/826
* https://github.com/LabKey/biologics/pull/2096
* https://github.com/LabKey/sampleManagement/pull/1777

#### Changes
- allow insertRows and updateRows to skip reselect rows
- determine queryTable for existing sample based on update columns
- only query for columns and lineage are needed for sample existing record

PR #4322, opened 2023-04-18
Performance Evaluation for Data at Scale - sample detailed audit logging

XingY:
#### Rationale
Detail audit log is enabled for sample CUD actions for generating timelines. ExistingRecordDataIterator is used to query the previous sample detail so that the comparison against the updates can be generated. When merge/update samples, ExistingRecordDataIterator queries for the full sample details, including exp.material, joined by samples.sampleprovisioned, joined by calculated inventory storage tables. Then, the sample's lineage is queried for generating audit log for parent events. Often times during an update, only a small subset of columns are included in the rows data. For columns that aren't updated, there is no need to query for them in ExistingRecordDataIterator. 

#### Related Pull Requests
* 

#### Changes
* determine queryTable for existing sample based on update columns
* only query for columns that are needed
* only query for lineage when needed.


XingY:
ported to https://github.com/LabKey/platform/pull/4312

PR #4325, opened 2023-04-18
Move files associated with sample domain fields when the samples change projects

labkey-susanh:
#### Rationale
In support of moving samples from one project to another, we need to move any files associated with File type fields in the sample domains to the directory for the destination project.  We do this as a post-commit hook on the sample move transaction so that rolling back is not necessary if the transaction fails.  We also record the update of the file path in the detailed audit log for the samples that move.

#### Changes
* Move files attached to each sample
* Add detailed audit logs for the file path updates.


api/src/org/labkey/api/exp/api/SampleTypeService.java
```diff
@@ -28,6 +28,9 @@
 import org.labkey.api.gwt.client.model.GWTIndex;
 import org.labkey.api.gwt.client.model.GWTPropertyDescriptor;
 import org.labkey.api.qc.DataState;
+import org.labkey.api.query.BatchValidationException;
+import org.labkey.api.query.InvalidKeyException;
+import org.labkey.api.query.QueryUpdateServiceException;
```
cnathe: remove unused imports

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -157,7 +161,7 @@ public static SampleTypeServiceImpl get()
         return (SampleTypeServiceImpl) SampleTypeService.get();
     }
 
-    private static final Logger LOG = LogManager.getLogger(SampleTypeServiceImpl.class);
+    private static final Logger LOG = LogHelper.getLogger(SampleTypeServiceImpl.class, "Info about sample typeoperations");
```
cnathe:
```suggestion
    private static final Logger LOG = LogHelper.getLogger(SampleTypeServiceImpl.class, "Info about sample type operations");

```

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1751,4 +1786,135 @@ private int materialAliasMapRowContainerUpdate(List<Integer> sampleIds, Containe
...
+        Map<Integer, List<FileFieldRenameData>> sampleFileRenames = new HashMap<>();
+        List<? extends DomainProperty> fileDomainProps = sampleType.getDomain()
+                .getProperties().stream()
+                .filter(prop -> PropertyType.FILE_LINK.getTypeUri().equals(prop.getRangeURI())).toList();
+        if (fileDomainProps.isEmpty())
+            return sampleFileRenames;
+
+        FileContentService fileService = FileContentService.get();
+        if (fileService == null)
+            return sampleFileRenames;
```
cnathe: minor but could move this up to the top of the method. if we don't have a FileContentService then we don't need to do the work to get the sample type domain and check it for file fields.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1751,4 +1786,135 @@ private int materialAliasMapRowContainerUpdate(List<Integer> sampleIds, Containe
...
+        return sampleFileRenames;
+    }
+
+    private File getTargetFile(String absoluteFilePath, @NotNull Container sourceContainer, @NotNull Container targetContainer, @Nullable User user)
+    {
+        if (absoluteFilePath == null)
+            return null;
+
+        FileContentService fileService = FileContentService.get();
+        if (fileService == null)
```
cnathe, line 1847: redundant check here?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1751,4 +1786,135 @@ private int materialAliasMapRowContainerUpdate(List<Integer> sampleIds, Containe
...
+
+        File file = new File(absoluteFilePath);
+        if (!file.exists())
+        {
+            LOG.warn("File '" + absoluteFilePath + "' not found and cannot be moved");
+            return null;
+        }
+
+        File sourceFileRoot = fileService.getFileRoot(sourceContainer);
+        if (sourceFileRoot == null)
```
cnathe, line 1861: I'm curious if this check and the one below for the targetFileRoot can be moved up outside of the `getTargetFile` method. Does this check need to be made for every file being moved or could it just be done once up in `updateSampleFilePaths`?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1751,4 +1786,135 @@ private int materialAliasMapRowContainerUpdate(List<Integer> sampleIds, Containe
         aliasMapUpdate.append(")");
         return new SqlExecutor(aliasMapTable.getSchema()).execute(aliasMapUpdate);
     }
+
+    // return the map of file renames
+    private Map<Integer, List<FileFieldRenameData>> updateSampleFilePaths(TableInfo sampleTypeTable, ExpSampleType sampleType, List<ExpMaterial> samples, Container targetContainer, User user)
```
cnathe: IntelliJ is telling me that the `sampleTypeTable` param is unused. Can it be removed?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1751,4 +1786,135 @@ private int materialAliasMapRowContainerUpdate(List<Integer> sampleIds, Containe
...
+                    List<FileFieldRenameData> fieldRenameData = sampleFileRenames.get(sample.getRowId());
+                    fieldRenameData.add(renameData);
+                }
+            }
+        }
+
+        return sampleFileRenames;
+    }
+
+    private File getTargetFile(String absoluteFilePath, @NotNull Container sourceContainer, @NotNull Container targetContainer, @Nullable User user)
```
cnathe: can the `user` param be removed? looks like it is unused

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1751,4 +1786,135 @@ private int materialAliasMapRowContainerUpdate(List<Integer> sampleIds, Containe
...
+        File targetFile = new File(targetPath);
+        while (targetFile.exists())
+        {
+            name = baseName + "-" + duplicateCount;
+            if (extension != null)
+                name += "." + extension;
+
+            targetFile = new File(targetFile.getParentFile(), name);
+            duplicateCount++;
+        }
```
cnathe: I assume there is somewhere else in the code that does this sort of de-duplication of file names. Does this make sense to factor out into FileUtil or somewhere shared?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1751,4 +1786,135 @@ private int materialAliasMapRowContainerUpdate(List<Integer> sampleIds, Containe
...
+        return targetFile;
+    }
+
+    private boolean moveFile(FileFieldRenameData renameData, User user, Container targetContainer)
```
cnathe: more unused params here: user and targetContainer

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1751,4 +1786,135 @@ private int materialAliasMapRowContainerUpdate(List<Integer> sampleIds, Containe
...
+                        renameData.sourceFile.getAbsolutePath(),
+                        renameData.targetFile.getAbsolutePath(),
+                        renameData.sampleTypeName,
+                        renameData.sampleName,
+                        renameData.fieldName));
+                return false;
+            }
+        }
+
+        return true;
```
cnathe, line 1905: I found this set of if/else statements a little hard to follow. Thoughts on flipping the logic so that it first checks if the `renameData.targetFile.getParentFile().exists()` and if not creates the dirs, then just calling `renameData.sourceFile.renameTo(renameData.targetFile)` once?

cnathe (changes requested):

api/src/org/labkey/api/exp/api/SampleTypeService.java
labkey-susanh: Done.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Done.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Some of the redundancy I leave in just so IntelliJ doesn't complain.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Moved it out of this method.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Removed

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Removed

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: I assume there is too, but I couldn't find it...  Finally found it in `AssayFileWriter`. This is what is used in `AbstractQueryImportAction` as well.  Could arguably be moved somewhere more general, but I'll leave it for now.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Done. Thanks.  The error checking evolved..

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Done

cnathe (approved):

PR #4329, opened 2023-04-19
Support for restricted issue lists

labkey-klum:
#### Rationale
To support a client request for restricted issue lists we will introduce changes to enable the configuration on a per issue list basis. Restricted issue lists will have the ability to limit access to individual issues at the point at which a user tries to access an issue to:
- view details
- update
- resolve, close, reopen

The feature will not filter access from the issue list view nor will it prevent access of any tables in the query schema. 

Access checking, and storage of admin UI settings is delegated to a new interface `RestrictedIssueProvider` which will need to be registered by the module providing the implementation.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1177
- https://github.com/LabKey/onprcEHRModules/pull/675

#### Changes
- introduce `RestrictedIssueProvider` to delegate permission checking as well as settings storage.
- previously issue APIs were split between the module api jar and core api, the change consolidates all into core api
- introduce a module property to control display of the admin UI in the issue list designer


issues/src/org/labkey/issue/model/IssueManager.java
```diff
@@ -195,7 +198,40 @@ private static IssueObject _getIssue(@Nullable Container c, int issueId)
...
+
+            // need to check all related issues
+            for (Integer relatedIssue : issue.getRelatedIssues())
+            {
+                IssueObject related = _getIssue(c, user, relatedIssue);
+                if (related != null)
+                    relatedIssues.add(related);
+            }
+
+            if (!provider.hasPermission(user, issue, relatedIssues, errors) && !errors.isEmpty())
```
cnathe: the `!errors.isEmpty()` being in the if statement here was confusing me. I assume that the errors list is being populated during the call to `provider.hasPermission()` right? If so, wouldn't the returned boolean from that be enough to enter the if statement. Is there a scenario where that would return `false` but `errors` would be empty?

cnathe (approved):

issues/src/org/labkey/issue/model/IssuesListDefServiceImpl.java
```diff
@@ -308,4 +335,22 @@ public void deleteIssueDefsForDomain(User user, Domain domain)
     {
         IssueManager.deleteIssueDefsForDomain(user, domain);
```
labkey-adam, line 336: Should invoke some kind of provider cleanup here or in `IssueManager.deleteIssueDefsForDomain(user, domain)` to delete up the props

labkey-adam (changes requested):
Also see Cory's comments here and the other PR

issues/src/org/labkey/issue/model/IssueManager.java
labkey-klum: That's true, I'll modify this so only the boolean return value is enough to generate the failure and will use the returned error message or a default error message if none is returned.

labkey-adam (approved):

PR #4334, opened 2023-04-20
Improve support for Oracle 19c

labkey-adam:
#### Rationale
Stop warning about modern Oracle versions, add a test data source feature, fix a few issues discovered in testing, and do some spring clean-up. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47662

#### Related Pull Requests
* https://github.com/LabKey/server/pull/460
* https://github.com/LabKey/server/pull/470

#### General Changes
* Introduce a simple data source testing feature that, when invoked on a data source, enumerates all schemas and tables, querying the row count and the first 100 rows of each table. Specific schemas and tables can be excluded to avoid poor performance and other problems. Though not a comprehensive test, this ensures we attempt to query every table and column in the data source, highlighting problems with certain data types, etc.
* Limit default ORDER BY in the no-PK case to five columns
* Stop attempting to resolve column names against property names in `SimpleUserSchema`. This resulted in dropping legitimate columns when property names collided with subsequent column names.

#### Oracle-Specific Changes
* Rewrite paging using `OFFSET` and `FETCH` keywords introduced in Oracle 12c. This required some refactoring of our shared `limitRows()` methods. 
* Implement `isSortableDataType()` on Oracle, correctly tagging `CLOB`, `BLOB`, and custom data types as unsortable
* Translate weird Oracle TIMESTAMP instances to Timestamp
* Implement "show estimated execution plan" on Oracle
* Remove warnings claiming modern versions of Oracle are not tested
* Introduce `OracleVersion` and dialect classes for all major versions
* Note: I attempted to switch the JDBC driver from `ojdbc11` to `ojdbc11-production` to support `XMLType` and other features, but had to give up due to conflicts with Xerces, supposed CVEs in one of its dependencies (orai18n), and lack of demand from customers.


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -578,6 +587,7 @@ public ModelAndView getView(Object o, BindException errors)
             sb.append("\n<div>This page lists all the data sources defined in your ").append(AppProps.getInstance().getWebappConfigurationFilename()).append(" file that were available when first referenced and the external schemas defined in each.</div><br/>\n");
```
labkey-matthewb: Yuck.  Good candidate for org.labkey.api.util.DOM upgrade.

query/src/org/labkey/query/sql/QuerySelectView.java
```diff
@@ -516,8 +517,12 @@ which columns we can sort on until we validate which columns are available (beca
             if (sortFields != null && !sortFields.isEmpty())
             {
                 if (presentInSort.add(column.getFieldKey()))
+                {
                     // NOTE: we don't need to expando the list here, Sort.getOrderByClause() will do that
                     sort.appendSortColumn(column.getFieldKey(), column.getSortDirection(), false);
+                    if (sort.getSortList().size() >= DEFAULT_SORT_MAX_COLUMNS)
+                        break;
+                }
                 addedSortKeys = true;
```
labkey-matthewb: Is addedSortKeys updated correctly with the break?

api/src/org/labkey/api/data/dialect/LimitRowsSqlGenerator.java
```diff
@@ -8,36 +8,45 @@
  */
 public class LimitRowsSqlGenerator
 {
-    public static SQLFragment limitRows(SQLFragment frag, int rowCount, long offset, boolean supportsOffsetWithoutLimit)
+    public static SQLFragment limitRows(SQLFragment frag, int rowCount, long offset, LimitRowsCustomizer customizer)
```
labkey-matthewb, line 11: Both seem fine, but I'm curious about thoughts on subclass vs customizer.

query/src/org/labkey/query/sql/QuerySelectView.java
labkey-adam: I debated moving this inside the `if()`, meaning previous code was wrong. Sounds like you agree...

api/src/org/labkey/api/data/dialect/LimitRowsSqlGenerator.java
labkey-adam: Either way works. These were static methods so it was easiest to pass in context.

query/src/org/labkey/query/controllers/QueryController.java
labkey-adam: Yes, yes it is.

api/src/org/labkey/api/data/dialect/LimitRowsSqlGenerator.java
labkey-matthewb: Makes sense

labkey-matthewb (approved):

labkey-jeckels (commented):
Looks like Matt did a full review already. Let me know if you want me to dive into anything particular, but otherwise I'll defer to the existing approval.

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -582,148 +592,115 @@ public ModelAndView getView(Object o, BindException errors)
...
+                            catch (SQLException e)
+                            {
+                                status = FONT(cl("labkey-error"), "disconnected");
+                            }
 
-                            sb.append("<a href=\"").append(PageFlowUtil.filter(schemaAdminURL)).append("\">").append(PageFlowUtil.filter(path)).append("</a>");
-                            sb.append("</td></tr>\n");
-                            sb.append("          <tr><td>\n");
-                            sb.append("            <table>\n");
+                            return Stream.of(
```
labkey-matthewb, line 629: Nice!

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -582,148 +592,115 @@ public ModelAndView getView(Object o, BindException errors)
...
+                    TR(TD(TABLE(
+                        defs.stream()
+                            .map(def -> TR(TD(
+                                at(DOM.Attribute.style, "padding-left:20px"),
+                                new LinkBuilder(def.getUserSchemaName() +
+                                    (!StringUtils.equals(def.getSourceSchemaName(), def.getUserSchemaName()) ? " (" + def.getSourceSchemaName() + ")" : ""))
+                                    .href(urls.urlUpdateExternalSchema(c, def))
+                                    .clearClasses())
+                            ))
+                            .toArray()
```
labkey-matthewb: I think just passing the Stream directly works too.  This isn't very obvious since I beleve this argument is basically untyped (Object).

labkey-adam: @labkey-matthewb so it does... magic!

BTW: It would be nice to support `body instanceof Number` as well.

api/src/org/labkey/api/util/DOM.java
```diff
@@ -914,26 +914,38 @@ private static Appendable appendAttribute(Appendable html, Attribute key, Object
...
+        {
+            try
+            {
+                builder.append(body.toString());
+            }
+            catch (IOException e)
+            {
+                throw new RuntimeException(e);
+            }
+        }
```
labkey-adam, line 948: @labkey-matthewb any concern with the above?

labkey-matthewb: Nope.  Seems useful.

PR #4335, opened 2023-04-20
Add Rename Container API Action

RosalineP:
#### Rationale
We are seeking to add the ability to rename a container to our APIs, the first step of which is the addition of some server-side changes.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4335
* https://github.com/LabKey/sampleManagement/pull/1791

#### Changes
* Rename error check that is already being handled in `ContainerManager.rename`
* Add new action


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -6598,12 +6598,6 @@ public class RenameFolderAction extends FormViewAction<ManageFoldersForm>
         @Override
         public void validateCommand(ManageFoldersForm target, Errors errors)
         {
-            Container c = getContainer();
-            if (!ContainerManager.isRenameable(c))
-            {
-                // 16221
-                errors.reject(ERROR_MSG, "This folder may not be renamed as it is reserved by the system.");
-            }
```
cnathe, line 6606: curious why this validation check is being removed here? is it because it is already checked in the UI / JSP or something along those lines? It seems like it is still a valid check.

cnathe: ah, I think I see now. Looks like this validation check and others are already being done within `ContainerManager.rename()`

core/src/org/labkey/core/CoreController.java
```diff
@@ -1665,6 +1666,35 @@ public ApiResponse execute(Object o, BindException errors) throws Exception
         }
     }
 
+    @RequiresPermission(AdminPermission.class)
+    public class RenameContainerAction extends MutatingApiAction<SimpleApiJsonForm>
```
cnathe: I was expecting to see some validation checks here, but then (as per the other comments in this PR) I noticed that they are happening within the `ContainerManager.rename()` method. Maybe worth a comment here saying where the validation is happening.

core/src/org/labkey/core/CoreController.java
```diff
@@ -1665,6 +1666,35 @@ public ApiResponse execute(Object o, BindException errors) throws Exception
...
+            JSONObject json = form.getJsonObject();
+            if (json == null)
+            {
+                throw new NotFoundException("No JSON posted");
+            }
+
+            String name = StringUtils.trimToNull(json.optString("name"));
+            String titleValue = StringUtils.trimToNull(json.optString("title"));
+            String title = Objects.equals(name, titleValue) ? null : titleValue;
+            boolean addAlias = json.optBoolean("addAlias");
```
cnathe: Is there a specific reason for using this form of JSON binding and the `SimpleApiJsonForm`? I was expecting something more like the SampleManagerController.RenameProjectAction which uses `ProjectSettingsForm` or the AdminController.RenameFolderAction which uses `ManageFoldersForm`. Any reason for not creating an explicity java class in this case?

core/src/org/labkey/core/CoreController.java
cnathe: Also, is there a reason why this new action was added to the CoreController? I guess I would have expected it to live right next to the UI version in AdminController.

cnathe (changes requested):

core/src/org/labkey/core/CoreController.java
RosalineP: Added the comment! I think the original thought was that the action could be (in alphabetical order, as the file goes) in the same file as the CreateContainerAction, DeleteContainerAction, MoveContainerAction in CoreController but I think having it next to the UI action in AdminController makes more sense

core/src/org/labkey/core/CoreController.java
RosalineP: I meant for this to be reviewed after getting pulled to real PR (which was unclear, my bad there) so there were some artifacts of me looking at adjacent actions and following their patterns—but I've updated this to use the more normal java class pattern!

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -6590,6 +6591,67 @@ public List<Container> getTargetContainers(final Container currentContainer) thr
         }
     }
 
+    public static class RenameContainerForm
+    {
+        private String name;
+        private String title;
+        private boolean addAlias = true;
```
cnathe, line 6598: does the UI action also default this to `true`?

cnathe (approved):

core/src/org/labkey/core/admin/AdminController.java
RosalineP: So, the UI default checks it to true but the `ManageFoldersForm` defaults it to false. Maybe because the UI is default toggled to true. So I opted for true here

PR #4336, opened 2023-04-21
Issue 47755: Slow file upload via WebDav PUT

labkey-jeckels:
#### Rationale
Writing to a remote SMB mount is very slow with an 8KB array copy between InputStream and OutputStream. Channels are way faster.

#### Changes
* Use `Channel.transferTo()`

### DIFF
--- a/api/src/org/labkey/api/util/FileStream.java
+++ b/api/src/org/labkey/api/util/FileStream.java
@@ -28,6 +28,10 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.channels.Channel;
+import java.nio.channels.Channels;
+import java.nio.channels.FileChannel;
+import java.nio.channels.ReadableByteChannel;
 import java.util.Date;
 
 /**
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/util/FileStream.java
```diff
@@ -63,11 +77,14 @@ static void transferToImpl(FileStream s, File dest) throws IOException
...
-        try (BufferedInputStream in = new BufferedInputStream(s.openInputStream());
-             BufferedOutputStream out = new BufferedOutputStream(new FileOutputStream(dest));
+        try (ReadableByteChannel cIn = s.getInputChannel();
+             FileOutputStream fOut = new FileOutputStream(dest);
+             FileChannel cOut = fOut.getChannel();
         )
         {
-            IOUtils.copy(in, out);
+            LOG.debug("Starting to transfer to " + dest);
+            cOut.transferFrom(cIn, 0, s.getSize());
```
labkey-matthewb: just pass in Long.MAX_VALUE?

labkey-matthewb (approved):
Do you have any idea how important the KeepAlive nonsense is?

labkey-jeckels:
> Do you have any idea how important the KeepAlive nonsense is?

Ha! I was hoping you could tell me.

At least for this scenario, I don't it's helpful. It's presumably trying to keep the session from timing out during a very big upload. However, by the time we wrap the stream, Spring has already written the whole file to disk. Presumably the upload from the client to the webserver is the typical bottleneck, and not our processing of the file (aside from this issue we're currently troubleshooting).

PR #4341, opened 2023-04-21
Issue 47509: If the lookup was defined with an explicit container, verify that the sample is in that container before mapping its value

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47509

See related PR for rationale. This Pr fixes an edge case issue for when the assay field lookup definition is pointed as a specific container and the sample name in the uploaded assay run doesn't exist in that container but does resolve in a different container.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4316

#### Changes
* when resolving samples in assay import, check if the field lookup definition defines a specific lookup container
* PlateMetadataDataHandler update to add missing override to new importFile method signature


api/src/org/labkey/api/assay/plate/PlateMetadataDataHandler.java
```diff
@@ -63,4 +63,10 @@ public void importFile(@NotNull ExpData data, File dataFile, @NotNull ViewBackgr
         // AbstractAssayTsvDataHandler.addAssayPlateMetadata because we need to access the inserted result data to get at the
         // lsids.
     }
+
+    @Override
+    public void importFile(@NotNull ExpData data, File dataFile, @NotNull ViewBackgroundInfo info, @NotNull Logger log, @NotNull XarContext context, boolean allowLookupByAlternateKey) throws ExperimentException
+    {
+        // see description above
```
labkey-susanh, line 70: The "Yeah, what she said." comment. 😄

labkey-susanh (approved):

PR #4342, opened 2023-04-21
Create "Created/CreatedBy/Modified/ModifiedBy" for Assay data rows

### LINKED ISSUE #4328: Create "Created/CreatedBy/Modified/ModifiedBy" for Assay data rows - dev
[Spec](https://docs.google.com/document/d/1v7MpLJPvgF8uU9Hdw78zFPy96ldsSaaqP-MJiCoENh4/edit?usp=sharing)
[Github epic](https://github.com/LabKey/platform/issues/4259)

- [x] Add the columns to the DB table for all existing assay designs at upgrade - 3h
- [x] Add created and created by to default results grid view - 1h
- [x] Special case for assay run insert/create so that we don’t populate results created/etc. fields - 2h
- [x] Wire up special COALESCE behavior to pull data from exp.run and/or exp.data - 3h
- [x] Verify that created/createdBy info is set on API saveBatch that adds new rows - 2h
- [x] Verify that modified/modifiedBy is set for result row on update - 1h
- [x] Verify audit log details are as we expect - 1h
- [x] Unit tests - 3hr
- [x] create test plan
- [x] code review - 2hr @cnathe 
- [x] manual testing - 3hr  @cnathe 📍
- [x] manual testing feedback - 2hr
- [x] tc review and merge - 1hr

XingY:
#### Rationale
Assay results data don't have their own created/modified/by information stored because their creation and updates (reimport) were often tied to run create/update. In certain scenarios, assay data can be created/updated independently of the run create/update, such as when "Editable Results" feature is enabled, or when saveDomain api is used to swap out assay results for an existing run.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4342
* https://github.com/LabKey/commonAssays/pull/611

#### Changes
* migration script to add created/createdby/modified/modifiedby fields to assay designs with Results table
* add created/createdby/modified/modifiedby fields default AssayResultDomainKind
* coalesce result and run table info for created/modified fields for AssayResultTable
* skip created/createdby/modified/modifiedby auto population when results are created at the same time a run is created
* use constant "AliquotedFrom"


api/src/org/labkey/api/assay/AbstractTsvAssayProvider.java
```diff
@@ -41,6 +41,10 @@ public abstract class AbstractTsvAssayProvider extends AbstractAssayProvider
     public static final String ASSAY_SCHEMA_NAME = "assayresult";
     public static final String ROW_ID_COLUMN_NAME = "RowId";
     public static final String DATA_ID_COLUMN_NAME = "DataId";
+    public static final String CREATED_COLUMN_NAME = "Created";
+    public static final String CREATED_BY_COLUMN_NAME = "CreatedBy";
+    public static final String MODIFIED_COLUMN_NAME = "Modified";
+    public static final String MODIFIED_BY_COLUMN_NAME = "ModifiedBy";
```
cnathe: would it be possible to use the existing 4 string constants from `org.labkey.api.data.Table` instead of adding these new ones?

api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
```diff
@@ -140,7 +140,10 @@ public void importFile(@NotNull ExpData data, File dataFile, @NotNull ViewBackgr
         assert(rawData.size() <= 1);
         try
         {
-            importRows(data, info.getUser(), run, protocol, provider, rawData.values().iterator().next(), settings);
+            boolean autoFillDefaultResultColumns = false;
+            if (context instanceof AssayRunUploadContext)
+                autoFillDefaultResultColumns = ((AssayRunUploadContext<?>) context).shouldAutoFillDefaultResultColumns();
```
cnathe: I'm confused by this. The context param is of type `XarContext`. Can that actually ever be an instanceof `AssayRunUploadContext`?

api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
```diff
@@ -249,6 +249,8 @@ public ExpExperiment saveExperimentRun(
         boolean forceSaveBatchProps
     ) throws ExperimentException, ValidationException
     {
+        context.setAutoFillDefaultResultColumns(run.getRowId() > 0);
```
cnathe: should this be a little closer to the usage at line 523 just to make it easier to track where things are set/used? or is this prop used/needed on the context in any of the calls prior to the importRows call at line 523?

assay/src/org/labkey/assay/AssayUpgradeCode.java
```diff
@@ -0,0 +1,142 @@
...
+        try (DbScope.Transaction transaction = ExperimentService.get().ensureTransaction())
+        {
+            Set<ExpProtocol> protocols = new HashSet<>();
+            for (Container container : ContainerManager.getAllChildren(ContainerManager.getRoot()))
+            {
+                if (container != null)
+                    protocols.addAll(AssayService.get().getAssayProtocols(container));
+            }
+
+            Map<Container, List<Pair<ExpProtocol, Domain>>> protocolsByContainer = new HashMap<>();
```
cnathe, line 65: couldn't we just go straight to this Map by container in the for loop above which gets the assay protocols for a single container?

assay/src/org/labkey/assay/AssayUpgradeCode.java
```diff
@@ -0,0 +1,142 @@
...
+            for (ExpProtocol protocol : protocols)
+            {
+                AssayProvider provider = AssayService.get().getProvider(protocol);
```
cnathe, line 68: I can't remember what the spec said, but this will add the created/modified/by fields for all assay types with a data/results domain right (regardless of if they allow editable results)?

assay/src/org/labkey/assay/AssayUpgradeCode.java
```diff
@@ -0,0 +1,142 @@
...
+
+                _log.info("Finished adding result created/modified columns for " + protocolDomains.size() + " assay design(s) in " + container.getPath());
+            }
+
+            transaction.commit();
+        }
+
+    }
+
+    public static void _addAssayResultColumns(ExpProtocol protocol, Domain resultsDomain)
```
cnathe: private?

assay/src/org/labkey/assay/AssayUpgradeCode.java
```diff
@@ -0,0 +1,142 @@
...
+        _ensureColumn(MODIFIED_COLUMN_NAME, resultsDomain, protocol, provisionedTable, kind);
+        _ensureColumn(MODIFIED_BY_COLUMN_NAME, resultsDomain, protocol, provisionedTable, kind);
+    }
+
+    private static void _ensureColumn(String colName, Domain domain, ExpProtocol protocol, SchemaTableInfo provisionedTable, AssayResultDomainKind kind)
+    {
+
+        ColumnInfo col = provisionedTable.getColumn(colName);
+        if (col != null)
+            throw new IllegalStateException("Column '" + colName + "' is already defined in result table for '" + protocol.getName() + "'.");
```
cnathe: should this throw or log a warning/error and continue to the next?

assay/src/org/labkey/assay/TSVProtocolSchema.java
```diff
@@ -56,6 +56,11 @@
 import java.util.Map;
 import java.util.Set;
 
+import static org.labkey.api.assay.AbstractTsvAssayProvider.CREATED_BY_COLUMN_NAME;
+import static org.labkey.api.assay.AbstractTsvAssayProvider.CREATED_COLUMN_NAME;
+import static org.labkey.api.assay.AbstractTsvAssayProvider.MODIFIED_BY_COLUMN_NAME;
+import static org.labkey.api.assay.AbstractTsvAssayProvider.MODIFIED_COLUMN_NAME;
```
cnathe: unused imports? can they be removed?

cnathe (changes requested):

assay/src/org/labkey/assay/AssayUpgradeCode.java
XingY: Yes, because user can toggle editable settings.

api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
XingY: Yes, this is needed before run.save call. I'll add a comment.

assay/src/org/labkey/assay/AssayUpgradeCode.java
cnathe: But "Editable Results" is not an option for all assay types. For example, Luminex and NAb assay designs allow for "editable runs" but not "editable results". I'm not sure if that means we shouldn't add these default columns to those assay's domains or not

XingY: The check for resultsDomain and ExpProtocol.ASSAY_DOMAIN_DATA name space does filter out the other assay design types. I haven't checked Luminex, but NAb doesn't have a assayresult/table.

cnathe: ah, but those assay provider types probably don't have provisioned "Data" results domains as they probably have their own hard tables in the luminex or nab schema.

assay/src/org/labkey/assay/AssayUpgradeCode.java
XingY: Because of assay scope, it's best to put all protocols in a set to get rid of redundant protocols first.

api/src/org/labkey/api/assay/AbstractTsvAssayProvider.java
XingY: good suggestion. done.

api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
XingY: good catch. fixed.

assay/src/org/labkey/assay/AssayUpgradeCode.java
XingY: good point. updated.

cnathe (approved):

PR #4346, opened 2023-04-22
Issue 47704: DataRegionSelection contention on lock object

labkey-jeckels:
#### Rationale
Operations around data region selections on slow DB queries can block all other data region selection changes server-wide

#### Changes
* Switch synchronization to individual selection sets

### DIFF
--- a/api/src/org/labkey/api/data/DataRegionSelection.java
+++ b/api/src/org/labkey/api/data/DataRegionSelection.java
@@ -33,6 +33,7 @@ import org.labkey.api.view.DataView;
 import org.labkey.api.view.NotFoundException;
 import org.labkey.api.view.ViewContext;
 
+import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpSession;
 import java.io.IOException;
 import java.sql.ResultSet;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/DataRegionSelection.java
```diff
@@ -84,15 +94,20 @@
             HttpSession session = request != null ? context.getRequest().getSession(false) : null;
             if (session != null)
             {
-                @SuppressWarnings("unchecked") Set<String> result = (Set<String>)session.getAttribute(key);
-                if (result == null)
+                // Ensure that two different requests don't end up creating two different selection sets
+                // in the same session
+                synchronized (LOCK)
```
labkey-matthewb: See SessionHelper.getAtrribute()

labkey-jeckels: This use case has unusual behavior in terms of always returning a value but not always putting it in the session. I'm not sure why. I switched to use the striped locking from SessionHelper though, even though the time in the critical section is very short here.

labkey-nicka (approved):

PR #4347, opened 2023-04-24
Move or split derivation runs for samples that move

labkey-susanh:
#### Rationale
When derived samples are created, we create experiment runs (and associated exp.objects) in the container in which the sample lives.  When samples change containers, we need to move or split the runs that are used to create those samples.

#### Related Pull Requests
* 

#### Changes
* Update `SampleTypeServiceImpl.moveSamples` method
* Add `ExpProtocoloApplication.removeMaterialInputs` to remove multiple inputs
* Factor out `getDerivationRunName` method


experiment/src/org/labkey/experiment/api/ExpProtocolApplicationImpl.java
```diff
@@ -497,6 +501,27 @@ public void removeMaterialInput(User user, ExpMaterial material)
         ExperimentServiceImpl.get().queueSyncRunEdges(_object.getRunId());
     }
 
+    @Override
+    public void removeMaterialInputs(User user, Collection<Integer> rowIds)
```
cnathe, line 505: thoughts on updating the `removeMaterialInput` method above so that it just calls this method signature to remove some of the duplicate code?
i.e. `removeMaterialInputs(user, Collections.singleton(material.getRowId()));`

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1757,6 +1764,102 @@ public Map<String, Integer> moveSamples(Collection<? extends ExpMaterial> sample
...
+            // Update the run for the samples that have stayed behind. Change the name and remove the moved samples as outputs
+            run.setName(ExperimentServiceImpl.getDerivationRunName(run.getMaterialInputs(), run.getDataInputs(), numStaying, run.getDataOutputs().size()));
+            run.save(user);
+            List<Integer> movingSampleIds = movingSet.stream().map(ExpMaterial::getRowId).toList();
+
+            outputApp.removeMaterialInputs(user, movingSampleIds);
+            if (sourceApplication != null)
+                sourceApplication.removeMaterialInputs(user, movingSampleIds);
+
+            ExperimentService.get().queueSyncRunEdges(run);
```
cnathe: doesn't this `queueSyncRunEdges ` happen at the end of `removeMaterialInputs`? if so, does it matter if it is done twice or should this line go in an else?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -287,7 +287,7 @@
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
-import java.awt.Color;
+import java.awt.*;
```
cnathe: was this change intentional?

cnathe (approved):

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: IntelliJ intended to do it. I don't know why.

experiment/src/org/labkey/experiment/api/ExpProtocolApplicationImpl.java
labkey-susanh: The filtering used in `removeMaterialInput` might be slightly more efficient, so I'm going to leave it as is since it's probably the more common case.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: It might be queued twice anyway since there are potentially two calls to `removeMaterialInputs`, but three times is probably overkill.

PR #4352, opened 2023-04-25
Issue 47661: Assay run sample lookup issue trying to resolve sample for empty string value

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47661

When an assay design has a run domain lookup field to a sample type, we correctly skip trying to resolve the sample if the lookup is to an integer key of the sample and no value is provided. However, if the lookup is to a string name of the sample, we attempt to resolve a sample when the value provided is an empty string. This PR fixes that so we skip sample resolution when the run value is null or empty.

#### Changes
* DefaultAssayRunCreator.addInputMaterials - skip sample resolution when the run value is null or empty


cnathe:
changing to target 23.3 per Steve's request

PR #4357, opened 2023-04-26
Issue 47801: Sample status lookup field should resolve to statuses defined in the current, project, or shared containers

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47801

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4224

#### Changes
* Change default sample status lookup from the exp.materials table to be CurrentPlusProjectAndShared

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
@@ -454,7 +454,7 @@ public class ExpMaterialTableImpl extends ExpRunItemTableImpl<ExpMaterialTable.C
                 ret.setShownInInsertView(statusEnabled);
                 ret.setShownInUpdateView(statusEnabled);
                 ret.setRemapMissingBehavior(SimpleTranslator.RemapMissingBehavior.Error);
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -861,6 +861,13 @@ public void addQueryFieldKeys(Set<FieldKey> keys)
         }
     }
 
+    private ContainerFilter getSampleStatusLookupContainerFilter()
+    {
+        ContainerFilter.Type type = QueryService.get().getContainerFilterTypeForLookups(getContainer());
+        type = type == null ? ContainerFilter.Type.CurrentPlusProjectAndShared : type;
```
labkey-susanh, line 870: Can you document why this is different from other lookups?  I think it's because it's meant to default to CPPAS where otherwise it would default to the current container, right?

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
cnathe: correct. comment added.

PR #4362, opened 2023-04-27
App assay import / re-import fixes for 23.5

cnathe:
#### Rationale
Issue [47749](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47749): Assay re-import run from tmp file to use TSV loader for file preview

Issue [47576](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47576): AssayImportPanels change so that if there are wrapped columns that are marked as userEditable and shownInInsertView, include them in the form / UI

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1183

#### Changes
* UploadWizardAction update to check for wrapped columns for insert form
* InferDomainForm prop addition to allow for guessFormat as TSV


experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -1208,6 +1211,16 @@ public void setDomainKindName(String domainKindName)
         {
             _domainKindName = domainKindName;
         }
+
+        public boolean isGuessFormatAsTSV()
```
XingY, line 1215: Instead of _guessFormatAsTSV, why not just _format?

api/src/org/labkey/api/assay/actions/UploadWizardAction.java
```diff
@@ -827,9 +827,12 @@ protected DataRegion createDataRegionForInsert(TableInfo baseTable, String lsidC
         rgn.setTable(baseTable);
         for (ColumnInfo col : baseTable.getColumns())
         {
+            // Issue 47576: if there are wrapped columns that are marked as userEditable and shownInInsertView, include them in the form / UI
+            boolean shouldInclude = col.getWrappedColumnName() != null && col.isUserEditable() && col.isShownInInsertView();
```
XingY, line 831: is the check of col.isShownInInsertView() needed since dp.isShownInInsertView() is checked below?

XingY (approved):

labkey-klum (approved):

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
cnathe: That was my original plan. I was hoping to just allow the client API to pass in a string "guessFormat" of something like "TSV" and then have that used on the server side. However, I couldn't find a way to get from a string value to a java `FileType`. @labkey-klum do you know how I might be able to do that?

api/src/org/labkey/api/assay/actions/UploadWizardAction.java
cnathe: Yes, I believe it is needed since that original check is part of an OR conditional.

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
labkey-klum: I'm not aware of a way to go from a string to a `FileType` instance, we would need to add some sort of a factory method maybe by file extension, but I think one problem is that there isn't any central registration of `FileType`s

PR #4365, opened 2023-04-27
Create "Created/CreatedBy/Modified/ModifiedBy" for Assay data rows - fix clear flag

XingY:
#### Rationale
To allow setting system fields values on update, https://github.com/LabKey/platform/pull/4358 changed to use a mutable hashmap from Collections.singletonMap for flag values to update. However, the use of Map.of for the initial constructor is problematic for clearing flag value case because flag value is null and Map.of doesn't accept null value.

#### Related Pull Requests
* 

#### Changes
* don't use Map.of since it doesn't accept null value


cnathe (approved):
that unfortunate that Map.of doesn't support null values

PR #4367, opened 2023-04-28
Migrate all remaining uses of the old JSONObject and JSONArray library

labkey-adam:
#### Rationale
Get rid of the last references to the old `JSONObject` / `JSONArray` library and remove the old library.

#### Changes
* Make `ApiSimpleResponse` simple again: switch to new `JSONObject`, stop implementing Map (no need for that), remove unnecessary constructors, etc.
* Switch `AbstractAuditDomainKind.getType()` to return new `JSONObject`. This migrates many of the audit domain kinds plus related infrastructure.
* Remove unused methods
* Update `encodeObject()` and `decodeObject()`. Adjust some bean classes to ignore problematic getters.
* Remove old library


labkey-jeckels (approved):
Looks fine but might as well wait until we branch for 23.5 to avoid needing to fix any tests in that branch?

labkey-adam:
> Looks fine but might as well wait until we branch for 23.5 to avoid needing to fix any tests in that branch?

Yes, I was thinking the same thing.

PR #4356, opened 2023-04-26
JSONObject migration in platform modules

labkey-adam:
#### Rationale
Continue the migration

PR #4369, opened 2023-04-28
Fix QubeQuery to work correctly with new JSONObject

labkey-tchad:
#### Rationale
The new JSONObject does not implement `Map`. Some spots were assuming that it does.

#### Related Pull Requests
* #4356 

#### Changes
* Fix QubeQuery to work correctly with new JSONObject


labkey-jeckels (approved):

labkey-klum (approved):

labkey-adam (approved):
Thanks, @labkey-tchad !

PR #4370, opened 2023-04-28
Clean up JSON operator parsing error message

labkey-jeckels:
#### Rationale
We don't want to include SQLFragment.toString() in a user-facing error message

#### Changes
* Just show the SQL string in the JSON operator parsing error
* Misc other cleanup


labkey-matthewb (approved):
For a little more work this validation can also be moved/duplicated in SqlParser.java (see around line 860).  Then it would look work more like a real parse exception (e.g. Line/Col).

Not a big deal, but looks better in the editor.

labkey-jeckels:
@labkey-matthewb thanks for the suggestion - I'll probably do that if we start seeing more external adoption of these JSON operators.

PR #4372, opened 2023-04-30
Debug RhinoServiceTest failure

bbimber:
This is created for the purpose of getting TeamCity to run the test


labkey-jeckels:
@bbimber thanks for tracking this this failure. As you probably saw, it was running cleanly in some suites but not others. Can we safely close this PR now?

bbimber:
@labkey-jeckels, yes thank you. i fixed this and merged into 22.3 here: https://github.com/LabKey/platform/pull/4373

PR #4373, opened 2023-05-01
Fix windows-specific issue in RhinoService$Test

bbimber:
This is related to a relatively long standing TeamCity windows agent-specific failure in RhinoService$Test, such as:

 https://teamcity.labkey.org/test/-693611466596297028?currentProjectId=LabkeyTrunk&expandTestHistoryChartSection=true&branch=%3Cdefault%3E

https://teamcity.labkey.org/test/-693611466596297028?currentProjectId=LabkeyTrunk&expandTestHistoryChartSection=true&branch=%3Cdefault%3E&expandedTest=build%3A%28id%3A2423836%29%2Cid%3A7097

That test is using parameter substitution in an R script to poke in a filepath. Tests will show if I'm right, but what I think is happening is that these windows agents have a filepath with a dollar sign in the name. As a result, the regex replacement string has that dollar sign, and the Matcher is interpreting this as a bad group reference. 

I dont know if there is a more general-purpose regex escape method to use, but this PR will at least handle the dollar sign case by escaping those in the file path string. 

This needs to run on a windows agent to see if my theory is right and if the tests will pass now.

### DIFF
--- a/api/src/org/labkey/api/reports/report/r/ParamReplacementSvc.java
+++ b/api/src/org/labkey/api/reports/report/r/ParamReplacementSvc.java
@@ -460,6 +460,7 @@ public class ParamReplacementSvc
         {
             throw new IllegalArgumentException("There was a syntax error with the substitution parameter in your script.\n"
                     + (StringUtils.isEmpty(token) ? "" : "Substitution token \"" + token + "\" is used incorrectly.\n")
+                    + (StringUtils.isEmpty(replacementStr) ? "" : "Replacement string was: \"" + replacementStr + "\"\n")
                     + "Error message: " + e.getMessage(), e);
         }
     }
@@ -528,6 +529,10 @@ public class ParamReplacementSvc
                     {
                         resultFileName = resultFile.getAbsolutePath().replaceAll("\\\\", "/");
                     }
+
+                    // NOTE: dollar signs in the path will cause "java.lang.IllegalArgumentException: Illegal group reference" from Matcher.appendReplacement
+                    resultFileName = resultFileName.replaceAll("\\$", "\\\\\\$");
+
                     _log.debug("Found output parameter '" + param.getName() + "'.  Mapping local file '" + resultFile.getAbsolutePath() + "' to '" + resultFileName + "'");
                 }
                 String replacementStr = pattern.getReplacementStr(resultFileName, m.group(0), param.getName());


bbimber:
My theory is right: the agent is using the path "C:/labkey/apps/teamcity/agent/tools/apache-tomcat-9.0.27/temp/EC2AMAZ-CKOB5PR$....". My approach to fix that isnt. I'll have a look tomorrow and see if maybe I can repro it locally now that I know what to try.

bbimber:
Correction: TeamCity is tripping me up. I believe this PR is going to work. I repro'd locally with a dollar sign in the tempdir path and it's working.

I dont know if my user is going to be able to use the LabKey-only pool, but if this job can run it will verify the fix:

https://teamcity.labkey.org/buildConfiguration/LabKey_233Release_Community_ExternalModulesTestSqlserver/2424380?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true

bbimber:
@labkey-tchad or @labkey-adam: my attempt to kick off a windows build ended up running on a linux agent: https://teamcity.labkey.org/buildConfiguration/LabKey_233Release_Community_ExternalModulesTestSqlserver/2424380?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true. Any chance you'd be able to start one? My theory is that this has something to do with me not being able to use the labkey-only pool, but I dont really understand the configuration of TeamCity that well. if someone else starts the job and it's picked up on windows, I am able to restart those jobs. Would one of you mind kicking this off on a windows agent, or perhaps the results of the tests run (which do pass...) are good enough to approve as-is?

As noted above, I figured out how to repro locally. If you update tomcat's temp dir to include a dollar sign (such as editing the run config in intellij), this will repro. This patch should fix that issue by escaping dollar signs in the replacementString. The only real question I have is whether there's a more elegant java regex-escape method to use instead of my one-off replaceAll(). 

Thanks

labkey-tchad:
It looks like you used the correct technique to change the agent type -- modifying the `agent.filter` parameter. You just changed it to to run on Linux. Windows is the default for most SQL Server suites in release branches. @labkey-adam kicked it off this morning and it passed on Windows.
(PS For the "External Modules" suite, Windows is default on develop as well as release branches)

bbimber:
> It looks like you used the correct technique to change the agent type -- modifying the `agent.filter` parameter. You just changed it to to run on Linux. Windows is the default for most SQL Server suites in release branches. @labkey-adam kicked it off this morning and it passed on Windows. (PS For the "External Modules" suite, Windows is default on develop as well as release branches)

I could be completely misinterpreting the UI, but when I kick off the run without configuring anything, it chose linux. When I tried to pick manually, it gives me a bunch of options that are all windows agents. I selected an option to pick from that pool, and it gave me linux.

labkey-tchad:
I think I see what's happening. There was a slight misalignment with that build that was causing it to pick linux agents when you run a custom build. The selector for agent type is on the 'Parameters' tab:
image
(The agent selector on the "General" tab doesn't work how we want)

labkey-adam:
As for regex escaping, I don't know of any standard "escapeSpecialCharactersForRegex()"-type method in the platform or JDK. There is `Pattern.quote()`... not sure if that works in this case.

Ideally, we'd add something like `StringUtilsLabKey.escapeRegexLiteral()` that escapes all regex special chars, a la, https://stackoverflow.com/questions/10664434/escaping-special-characters-in-java-regular-expressions

bbimber:
@labkey-adam, Pattern.quote was my first thought too, and it doesnt cover $. 

If the windows test passed, any concerns with this fix?

bbimber:
thanks for the explanation @labkey-tchad

labkey-adam:
> @labkey-adam, Pattern.quote was my first thought too, and it doesnt cover $.
> 
> If the windows test passed, any concerns with this fix?

Nope

labkey-adam:
Thanks for the fix, @bbimber !

labkey-adam (approved):

bbimber:
@labkey-adam, Excellent - my OCD is now satisfied

PR #4387, opened 2023-05-03
Server-side changes to enable moving data class objects (sources) between folders

labkey-susanh:
#### Rationale
These are the server-side changes to enable moving of data class objects from one container to another, following on the work that enabled moving of samples. This helps users keep their data organized when they make mistakes or change the organizational plan after creating data.

#### Related Pull Requests
* https://github.com/LabKey/sampleManagement/pull/1820
* https://github.com/LabKey/inventory/pull/848

#### Changes
* Add new `MoveDataClassObjects` (aliased to `moveSources`) API action in `ExperimentController` and corresponding method in `ExperimentService`.
* Update `addSummaryAuditEvent` to accept a user comment
* add `removeDataInputs` to `ExpProtocolApplication`
* Move methods common for moving samples and data class objects into `ExperimentService`


experiment/src/org/labkey/experiment/api/ExpProtocolApplicationImpl.java
```diff
@@ -487,6 +487,27 @@ public void removeDataInput(User user, ExpData data)
         ExperimentServiceImpl.get().queueSyncRunEdges(_object.getRunId());
     }
 
+    @Override
+    public void removeDataInputs(User user, Collection<Integer> rowIds)
+    {
+        // Clean up DataInput exp.object and properties
+        List<String> inputLsids = new ArrayList<>();
+        rowIds.forEach(rowId -> {
+            inputLsids.add(MaterialInput.lsid(rowId, getRowId()));
```
cnathe:
```suggestion
            inputLsids.add(DataInput.lsid(rowId, getRowId()));

```

experiment/src/org/labkey/experiment/api/ExpProtocolApplicationImpl.java
```diff
@@ -487,6 +487,27 @@ public void removeDataInput(User user, ExpData data)
...
+        SQLFragment lsidsSql = new SQLFragment().append("SELECT ObjectUri FROM exp.Object WHERE Container = ").appendValue(getContainer())
+                .append(" AND ObjectURI ");
+        expSchema.getSqlDialect().appendInClauseSql(lsidsSql, inputLsids);
+        OntologyManager.deleteOntologyObjects(expSchema, lsidsSql, getContainer(), false);
+
+        SimpleFilter filter = new SimpleFilter();
+        filter.addCondition(FieldKey.fromParts("TargetApplicationId"), getRowId());
+        filter.addCondition(FieldKey.fromParts("DataId"), rowIds, CompareType.IN);
+        Table.delete(ExperimentServiceImpl.get().getTinfoDataInput(), filter);
+        ExperimentServiceImpl.get().queueSyncRunEdges(_object.getRunId());
```
cnathe: I think I said something similar last CR, but there are now 4 methods in this class that are very very similar (which I assume is where that copy-paste error above came from). Thoughts on consolidating them? 

or maybe at least the parts after the `OntologyManager.deleteOntologyObjects()` call could be shared. Something like:
```
    private void removeInput(TableInfo tableInfo, String idColName, Collection<Integer> rowIds)
    {
        SimpleFilter filter = new SimpleFilter();
        filter.addCondition(FieldKey.fromParts("TargetApplicationId"), getRowId());
        filter.addCondition(FieldKey.fromParts(idColName), rowIds, CompareType.IN);
        Table.delete(tableInfo, filter);
        ExperimentServiceImpl.get().queueSyncRunEdges(_object.getRunId());
    }
```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8328,6 +8331,254 @@ public static Pair<Integer, Integer> getCurrentAndCrossFolderDataCount(Collectio
...
+    {
+        TableInfo objectTable = OntologyManager.getTinfoObject();
+        SQLFragment objectUpdate = new SQLFragment("UPDATE ").append(objectTable).append(" SET container = ").appendValue(targetContainer.getEntityId())
+                .append(" WHERE objectid IN (SELECT objectid FROM ").append(tableInfo).append(" WHERE rowid ");
+        objectTable.getSchema().getSqlDialect().appendInClauseSql(objectUpdate, rowIds);
+        objectUpdate.append(")");
+        new SqlExecutor(objectTable.getSchema()).execute(objectUpdate);
+    }
+
+    private int dataObjectAliasMapRowContainerUpdate(List<Integer> sampleIds, Container targetContainer)
```
cnathe: minor: param name change, sampleIds -> dataIds

cnathe (commented):
I only got part way through the CR of this PR today, but will push my comments and get back to it tomorrow

experiment/src/org/labkey/experiment/api/ExpProtocolApplicationImpl.java
labkey-susanh: Thanks. Done.

experiment/src/org/labkey/experiment/api/ExpProtocolApplicationImpl.java
labkey-susanh: Done.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Done.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8328,6 +8331,254 @@ public static Pair<Integer, Integer> getCurrentAndCrossFolderDataCount(Collectio
...
+        return updateCounts;
+    }
+
+    private void addDataClassSummaryAuditEvent(User user, Container container, TableInfo dataClassTable, int rowCount, String auditUserComment)
+    {
+        QueryService queryService = QueryService.get();
+        queryService.getDefaultAuditHandler().addSummaryAuditEvent(user, container, dataClassTable, QueryService.AuditAction.UPDATE, rowCount, AuditBehaviorType.SUMMARY, auditUserComment);
+    }
+
+    private int dataRowContainerUpdate(List<Integer> dataIds, Container targetContainer, User user)
```
cnathe: it looks like this method and `materialRowContainerUpdate` are exactly the same except for the TableInfo. Thoughts on renaming this and making it take in a TableInfo param and then calling it from the SampleTypeServiceImpl case?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8328,6 +8331,254 @@ public static Pair<Integer, Integer> getCurrentAndCrossFolderDataCount(Collectio
...
+    {
+        TableInfo objectTable = OntologyManager.getTinfoObject();
+        SQLFragment objectUpdate = new SQLFragment("UPDATE ").append(objectTable).append(" SET container = ").appendValue(targetContainer.getEntityId())
+                .append(" WHERE objectid IN (SELECT objectid FROM ").append(tableInfo).append(" WHERE rowid ");
+        objectTable.getSchema().getSqlDialect().appendInClauseSql(objectUpdate, rowIds);
+        objectUpdate.append(")");
+        new SqlExecutor(objectTable.getSchema()).execute(objectUpdate);
+    }
+
+    private int dataObjectAliasMapRowContainerUpdate(List<Integer> dataIds, Container targetContainer)
```
cnathe: similar comment here. this looks to be the same as `materialAliasMapRowContainerUpdate` with the exception of the TableInfo. thoughts on passing that in as a prop and calling this from SampleTypeServiceImpl?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8328,6 +8331,254 @@ public static Pair<Integer, Integer> getCurrentAndCrossFolderDataCount(Collectio
...
+            parents.add(new ExpDataClassAttachmentParent(data.getContainer(), lsid));
+
+        }
+        try
+        {
+            AttachmentService.get().moveAttachments(targetContainer, parents, user);
+        }
+        catch (IOException ignored)
+        {
+            // method doesn't actually throw.
```
cnathe, line 8481: good point, looks like it doesn't. should that be removed from the method signature in `AttachmentService.java` then?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8328,6 +8331,254 @@ public static Pair<Integer, Integer> getCurrentAndCrossFolderDataCount(Collectio
...
+                        Map<String, Object> oldRecordMap = new CaseInsensitiveHashMap<>();
+                        oldRecordMap.put("Container", sourceContainer.getName());
+                        oldRecordMap.put("rowId", data.getRowId());
+                        oldRows.add(oldRecordMap);
+                        Map<String, Object> newRecordMap = new CaseInsensitiveHashMap<>();
+                        newRecordMap.put("Container", targetContainer.getName());
+                        newRecordMap.put("rowId", data.getRowId());
+                        newRows.add(newRecordMap);
+                    }
+                    QueryService.get().getDefaultAuditHandler().addAuditEvent(user, targetContainer, dataClassTable, dcAuditBehavior, userComment, QueryService.AuditAction.UPDATE, newRows, oldRows);
```
cnathe, line 8425: interesting, for the sample case we added a new audit event for each sample and in the data class case we can add a single event with all of the objects as params. nice

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8328,6 +8331,254 @@ public static Pair<Integer, Integer> getCurrentAndCrossFolderDataCount(Collectio
...
+                // create new detailed events for each data object that was moved
+                AuditBehaviorType dcAuditBehavior = dataClassTable.getAuditBehavior(auditBehavior);
+                if (dcAuditBehavior == AuditBehaviorType.DETAILED)
+                {
+                    List<Map<String, Object>> oldRows = new ArrayList<>();
+                    List<Map<String, Object>> newRows = new ArrayList<>();
+                    for (ExpData data : classObjects)
+                    {
+                        Map<String, Object> oldRecordMap = new CaseInsensitiveHashMap<>();
+                        oldRecordMap.put("Container", sourceContainer.getName());
```
cnathe, line 8417: should the key for these be "Project" instead of "Container"? that looks like what we went with in the sample case.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8328,6 +8331,254 @@ public static Pair<Integer, Integer> getCurrentAndCrossFolderDataCount(Collectio
...
+                toSplit.add(run);
+        }
+
+        int updateCount = moveExperimentRuns(toUpdate, targetContainer, user);
+        int splitCount = splitExperimentRuns(toSplit, runIdData, targetContainer, user);
+        return Map.of("sourceDerivationRunsUpdated", updateCount, "sourceDerivationRunsSplit", splitCount);
+    }
+
+    @Override
+    public int moveExperimentRuns(List<ExpRun> runs, Container targetContainer, User user)
```
cnathe, line 8513: this is an exact copy of `SampleTypeServiceImpl.updateExperimentRuns`. can that one be removed and call this one?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -7700,60 +7700,186 @@ public ModelAndView getView(Object o, BindException errors) throws SQLException
...
+            }
+
+            _dataClassObjects = ExperimentServiceImpl.get().getExpDatas(dataIds);
+            if (_dataClassObjects.size() != dataIds.size())
+            {
+                errors.reject(ERROR_GENERIC, "Unable to find all sources for the move operation.");
+                return;
+            }
+
+            // verify all samples are from the current container
```
cnathe:
```suggestion
            // verify all sources are from the current container

```

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
```diff
@@ -291,41 +238,37 @@ describe('ExperimentController', () => {
         }
     }
 
-    describe('moveSamples.api', () => {
-        // NOTE: the MoveSamplesAction is in the experiment module, but the sample status related test cases won't
-        // work here because the sample status feature is only "enabled" when the sampleManagement module is available.
-        // See sampleManagement/src/client/test/integration/MoveSamplesAction.ispec.ts for additional test cases.
-
+    describe("moveSamples.api", () => {
```
cnathe, line 241: minor: In the other ispec.ts file, looks like there was an `ACTION_NAME` const defined and used. do the same here?

cnathe (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: No, because "Project" actually means the LKS Project and "Container" means the folder, so we're stuck with confusing fields here, unless we choose to override this in the UI.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Oops. Meant to do that.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: It changes the interface that is relied on elsewhere, so I'm opting not to change this.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Done.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Done.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Oh, looks like I actually did already.

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
labkey-susanh: Didn't seem necessary here since there's only one name for this action. I added it in the other place because (a) the initial name was long and (b) aliasing!  I can do it here too but I'm not sure it helps much.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: done

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
cnathe: no problem. either way is fine with me. I see what you mean about the alias part.

PR #4390, opened 2023-05-04
47775: List of possible export objects doesn't respect source folder when creating from template

labkey-klum:
#### Rationale
When creating a new project or folder based on a template, the initial list of folder objects is based on the current folder (or root). The list of objects is not refreshed when the source folder is updated from the dropdown. More details in this tracking isssue:

[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47775)

#### Changes
- Don't render an initial list of folder object when `create from template folder` is selected.
- Each time the template folder is changed, we update the list of folder objects based on the source folder.
- Remove the hard-coded study selection that appeared regardless of the folder, we believe this was legacy and prior to study folder writers being supported through the API.
- Render any child folder object, for example if the source folder is study, we now render all of the valid study folder objects (not just the top level study object)
- `scrollToBottom` was not working for chrome, I switched it to use a non `Ext4` based method to a javascript using the css `scroll-behavior: smooth` property. I don' know if we use this elsewhere and am fine to just rely on the auto scroll behavior.


core/src/org/labkey/core/admin/createFolder.jsp
```diff
@@ -454,23 +456,69 @@
...
+                                    xtype: "checkbox",
+                                    fieldCls : 'child-checkbox',
+                                    name: "templateWriterTypes",
+                                    hideLabel: true,
+                                    boxLabel: childWriterName,
+                                    inputValue: childWriterName,
+                                    checked: writerConfig.checked,
+                                    parentId: writer.name
+                                });
+                            });
```
cnathe, line 506: I didn't realize when we were discussing this that this code was already including the checkboxes for the child writers, they were just hidden. So as discussed, I assume the reason for that was just to remove extra options from the UI (which was probably desired at the time of initial implementation but it makes sense to show them now).

cnathe (approved):

cnathe:
@labkey-klum I pulled down the FB locally and did some testing, overall it looks good. One comment: in the initial state, it seems a little odd to have the empty "Folder objects to copy:" header showing with nothing below it. Thoughts on adding in some empty state text such as "Select a template folder above to see options."?
Screenshot 2023-05-05 at 8 16 44 AM

labkey-klum:
> @labkey-klum I pulled down the FB locally and did some testing, overall it looks good. One comment: in the initial state, it seems a little odd to have the empty "Folder objects to copy:" header showing with nothing below it. Thoughts on adding in some empty state text such as "Select a template folder above to see options."? Screenshot 2023-05-05 at 8 16 44 AM

@cnathe good suggestion. I added the empty text and also hide the labels and options until a folder is selected.

PR #4391, opened 2023-05-04
Issue 47849: Using import to update samples should fail if a sample in the import file does not belong to the current project.

XingY:
#### Rationale
Import (insert/merge/update) operates on data that exist in the current folder only. StatementUtils checks that  an existing record to be updated has a container that equals the current container during merge/update, without that data permission/security would be compromised. Insert also always insert into the current context container only. [Issue 47849](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47849).

The CF for import target shouldn't be relaxed. Perhaps we should always force CF to be Current.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3479

#### Changes
* remove CF override for import taget tableInfo

### DIFF
--- a/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
+++ b/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
@@ -859,12 +859,12 @@ public class SampleTypeUpdateServiceDI extends DefaultQueryUpdateService
         return ContainerFilter.current(container);
     }
 
-    public record ExistingRowSelect(TableInfo tableInfo, Set<String> columns, boolean includeParent) {}
+    public record ExistingRowSelect(TableInfo tableInfo, Set<String> columns, boolean includeParent, boolean addContainerFilter) {}
 
     public @NotNull ExistingRowSelect getExistingRowSelect(@Nullable Set<String> dataColumns)
     {
         if (!(getQueryTable() instanceof UpdateableTableInfo updatable) || dataColumns == null)
-            return new ExistingRowSelect(getQueryTable(), ALL_COLUMNS, true);
+            return new ExistingRowSelect(getQueryTable(), ALL_COLUMNS, true, false);
 
         CaseInsensitiveHashMap<String> remap = updatable.remapSchemaColumns();
         if (null == remap)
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -859,12 +859,12 @@ else if (!container.isProject() && container.getProject() != null)
         return ContainerFilter.current(container);
     }
 
-    public record ExistingRowSelect(TableInfo tableInfo, Set<String> columns, boolean includeParent) {}
+    public record ExistingRowSelect(TableInfo tableInfo, Set<String> columns, boolean includeParent, boolean addContainerFilter) {}
```
labkey-nicka: nit: Seems like `ExistingRowSelect` and `getExistingRowSelect` should be marked as `private`.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -4129,14 +4129,12 @@ public abstract class AbstractExpDataImportAction extends AbstractQueryImportAct
...
-            if (query.getContainerFilter() == null)
-            {
-                ContainerFilter cf = QueryService.get().getContainerFilterForLookups(getContainer(), getUser());
-                if (cf != null)
-                    query.setContainerFilter(cf);
-            }
             List<QueryException> qpe = new ArrayList<>();
             TableInfo t = query.getTable(form.getSchema(), qpe, true);
+
+            if (query.getContainerFilter() != null) // cross folder import not supported
```
labkey-nicka: The intention of this container filter is to allow resolution of cross-folder lookups. If the API user wants to specify a broader container filter then shouldn't we allow them to resolve those lookups? Or is that now taken care of downstream somewhere else?

If it is handled then rather than adjusting non-null container filters here I'd recommend throwing an error stating that container filtering is not supported for import actions.

XingY: Change made to throw error when cf is not current.

labkey-nicka (approved):

PR #4395, opened 2023-05-05
Allow "Move to Project" for LKB mixture batches

cnathe:
#### Rationale
We recently added the UI and backend support for moving samples to different containers (i.e. projects) within the LKB and LKSM apps. We held off on enabling this for LKB mixture batches until the dev was complete. The related PR enables the "Move to Project" UI for mixture batches from the grid Edit button and from a batch details Manage menu. This PR adds code to update the container value for a few additional object property rows related to the ExperimentRun's protocolapplications and datainputs.

#### Related Pull Requests
- https://github.com/LabKey/biologics/pull/2124

#### Changes
- during moveExperimentRuns, also move any object properties for the run's protocolapplications, datainputs, and materialinputs

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -8332,14 +8332,23 @@ public class ExperimentServiceImpl implements ExperimentService, ObjectReference
     }
 
     @Override
-    public void updateExpObjectContainers(TableInfo tableInfo, List<Integer> rowIds, Container targetContainer)
+    public int updateExpObjectContainers(TableInfo tableInfo, List<Integer> rowIds, Container targetContainer)
     {
         TableInfo objectTable = OntologyManager.getTinfoObject();
         SQLFragment objectUpdate = new SQLFragment("UPDATE ").append(objectTable).append(" SET container = ").appendValue(targetContainer.getEntityId())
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8332,14 +8332,23 @@ public static Pair<Integer, Integer> getCurrentAndCrossFolderDataCount(Collectio
...
                 .append(" WHERE objectid IN (SELECT objectid FROM ").append(tableInfo).append(" WHERE rowid ");
         objectTable.getSchema().getSqlDialect().appendInClauseSql(objectUpdate, rowIds);
         objectUpdate.append(")");
-        new SqlExecutor(objectTable.getSchema()).execute(objectUpdate);
+        return new SqlExecutor(objectTable.getSchema()).execute(objectUpdate);
+    }
+
+    private int updateExpObjectContainers(List<String> lsids, Container targetContainer)
+    {
+        TableInfo objectTable = OntologyManager.getTinfoObject();
```
labkey-nicka, line 8349: This may already be handled by the `SqlExecutor` but it seems like an empty check on `lsids` is warranted.

labkey-nicka (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
cnathe: done

PR #4396, opened 2023-05-05
Make ApiUsageException return a more appropriate response code

labkey-tchad:
#### Rationale
The description on `ApiUsageException` indicate that it should be used when the client makes an invalid request. A `500` response is not appropriate for this sort of error. `ExceptionUtil.handleException` will detect this exception and return `400` status but not all code paths will pass through that.
Looking at usages of `ApiUsageException`; some should be `404`s and many others should be `422`. Changing them all to `4xx` responses seem like a step in the right direction.

#### Related Pull Requests
* N/A

#### Changes
* Make `ApiUsageException` produce a 400 response


api/src/org/labkey/api/action/ApiUsageException.java
```diff
@@ -15,33 +15,28 @@
...
-    public ApiUsageException(String message, Throwable cause)
-    {
-        super(message, cause);
+        this(message, null);
     }
 
     public ApiUsageException(Throwable cause)
     {
-        super(cause.getMessage() == null ? cause.toString() : cause.getMessage(), cause);
+        this(cause.getMessage() == null ? cause.toString() : cause.getMessage(), cause);
```
labkey-jeckels: I think we should just pass `cause.getMessage()` here, though I see this old code that I appear to have committed.

api/src/org/labkey/api/action/ApiUsageException.java
```diff
@@ -15,33 +15,28 @@
...
+import org.labkey.api.view.BadRequestException;
 
 /**
  * Signals the client API caller that they somehow made an invalid request. These errors are not reported to the
  * mothership.
- * User: jeckels
- * Date: Oct 5, 2010
  */
-public class ApiUsageException extends RuntimeException implements SkipMothershipLogging
+public class ApiUsageException extends BadRequestException implements SkipMothershipLogging
```
labkey-jeckels, line 26: I think defaulting to a 422 seems best, and cases that should return a 404 can be switched to use `NotFoundException`.

labkey-jeckels (commented):
I'd like Matt to comment here, as I know he has strong (and totally reasonable) feelings on what response codes are appropriate.

api/src/org/labkey/api/action/ApiUsageException.java
labkey-matthewb: What a mess.  A lot of usages of ApiUsageException do indeed seem equally or better served by NotFoundException or ValidationException.

I like the move to 4xx responses.  I do not like using 400 BAD_REQUEST for any request that is basically well formatted but we can't handle for an explainable reason (e.g. Validation, NotFound, Unauthorized).  

Separately, but not a topic for today.  If I could design this all over again, I would use 200 {success:false} for most of our error cases and reserve non 200 responses for actual protocol errors (e.g. this isn't an legal end point so NOT_FOUND, rather than I can't find that rowid).

labkey-nicka: I disagree. I think we should try to respond with the appropriate HTTP codes as intended. For example, [404](https://en.wikipedia.org/wiki/HTTP_404):

image

404 doesn't indicate that the end point is illegal but rather that the resource was not found. Not finding the thing with "rowId" is precisely the case it is intended to cover.

labkey-tchad: Thank you for taking a look at this. When I have some spare cycles, I'll take a crack at switching `ApiUsageException` to return a 422 and switching around some inappropriate usages.

labkey-matthewb: I'm a bit ambivalent about the 404 case.  I still think that generally any interaction where "I got your request", "I understood your request", "Here's my response to your request" doesn't require out of band error handling.

labkey-nicka: Certainly agree that the "out of band error handling" is not the ideal way to process this kind of thing.

labkey-tchad: The [latest wording for a 400 error](https://www.rfc-editor.org/rfc/rfc9110.html#name-400-bad-request) is a little less specific to malformed requests: `something that is perceived to be a client error` -- compared to [the original wording](https://www.rfc-editor.org/rfc/rfc2616#section-10.4.1): `The request could not be understood by the server due to malformed syntax`.
422 was added as a [WebDAV specific response code](https://www.rfc-editor.org/rfc/rfc4918#section-11.2) but has since been pulled into the [core HTTP standard](https://www.rfc-editor.org/rfc/rfc9110.html#name-422-unprocessable-content).

That being said, I'm inclined to move forward with 422 for `ApiUsageException`. That's what the GitHub API returns for requests that it understands but have some semantic error (e.g. missing a required field).

labkey-tchad:
This is likely to have unintended consequences. Client APIs have some explicit expectations of certain response codes. If we want to do this, it should be a more coordinated effort.

PR #4408, opened 2023-05-09
ApiResponseWriter refactor

labkey-matthewb:
#### Rationale
 1) Refactor exception->json and writing the exception to the response

2) Recover from an exception thrown after a response has been started
  * close open objects/arrays
  * write the exception properties
  * close response

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/action/ApiJsonWriter.java
+++ b/api/src/org/labkey/api/action/ApiJsonWriter.java
@@ -132,7 +132,7 @@ public class ApiJsonWriter extends ApiResponseWriter
             throw new IllegalStateException("close() called without calling endResponse()");
     }
 
-    private void checkClosed()
+    private void ensureNotClosed()
     {
         if (state == WriterState.Closed)
             throw new IllegalStateException("writer is closed");
@@ -141,7 +141,7 @@ public class ApiJsonWriter extends ApiResponseWriter
     @Override
     public void startResponse() throws IOException
     {
-        checkClosed();
+        ensureNotClosed();
         if (state != WriterState.Initialized)
             throw new IllegalStateException("startResponse() has already been called");
         state = WriterState.Started;
@@ -154,7 +154,7 @@ public class ApiJsonWriter extends ApiResponseWriter
     @Override
     public void endResponse() throws IOException
     {
-        checkClosed();
+        ensureNotClosed();
         if (state != WriterState.Started)
             throw new IllegalStateException("startResponse() has not been called");
 
@@ -173,7 +173,7 @@ public class ApiJsonWriter extends ApiResponseWriter
     @Override
     protected void writeObject(Object value) throws IOException
     {
-        checkClosed();
+        ensureNotClosed();
         if (value instanceof String || value instanceof Number || value instanceof Boolean || value == null)
         {
             jg.writeObject(value);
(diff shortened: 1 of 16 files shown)


api/src/org/labkey/api/action/ApiJsonWriter.java
```diff
@@ -106,36 +123,44 @@ protected ObjectMapper createObjectMapper()
...
+        if (state == WriterState.Closed)
+            return;
+        state = WriterState.Closed;
+
         jg.flush();
+        if (state == WriterState.Started)
+            throw new IllegalStateException("close() called without calling endResponse()");
+    }
+
+    private void checkClosed()
```
labkey-jeckels: Naming this `ensureNotClosed()` would be less ambiguous

api/src/org/labkey/api/action/ApiJsonWriter.java
```diff
@@ -245,13 +307,190 @@ public void writeListEntry(Object entry) throws IOException
         writeObject(entry);
     }
 
-    @Override
+    /**
+     * Attempt to reset the http response.  If possible the response will be completely reset.  If not possible,
+     * this code will attempt to return the response to the top-level object in the json response.
+     * Since the response will be in an unspecified state, the caller should only use The caller should use writeExceptionPropertiesToRootAndEndObject() after calling this method.
```
labkey-jeckels:
```suggestion
     * Since the response will be in an unspecified state, the caller should use writeExceptionPropertiesToRootAndEndObject() after calling this method.

```

api/src/org/labkey/api/action/ApiJsonWriter.java
```diff
@@ -245,13 +307,190 @@ public void writeListEntry(Object entry) throws IOException
...
-        initGenerator();
+        var context = jg.getOutputContext();
+
+        // If context.inRoot() is true, then either we haven't started writing and object, or we've finished writing the top-level object.
+        // So either there's nothing to do, or nothing we can do.
+        if (context.inRoot())
+            return;
+
+        // If the entire response so far is buffered in memory, we get a do over.
+        if (!getResponse().isCommitted())
```
labkey-jeckels, line 322: Can this move above the `context.inRoot()` call? Maybe it'll never happen, but as long as we haven't committed the response to the client, we can still reset and start again

api/src/org/labkey/api/view/WebPartView.java
```diff
@@ -150,15 +152,14 @@ public ApiResponse renderToApiResponse()
...
-                writer.endResponse();
-            }
+                JSONObject json = new JSONObject();
+                json.put("html", mr.getContentAsString());
+                json.put("requiredJsScripts", includes);
+                json.put("implicitJsIncludes", implicitIncludes);
+                json.put("requiredCssScripts", cssScripts);
+                json.put("implicitCssIncludes", implicitCssScripts);
+                json.put("moduleContext", PageFlowUtil.getModuleClientContext(getViewContext(), dependencies));
+                writer.writeResponse(json);            }
```
labkey-jeckels:
```suggestion
                writer.writeResponse(json);
            }
```

labkey-jeckels (approved):
Nice! A few minor comments but looks great.

PR #4411, opened 2023-05-10
Issue 47024: Session expiration check on tab focus via WebSocket call to whoAmI.api

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47024

There have been some cases when the current "session expired" check in the WebSocket.js code hasn't been called for the apps. This PR adds another mechanism to check for this by adding a tab focus listener to the apps so that it will check for an expired session when the tab is given focus.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/4411
- https://github.com/LabKey/labkey-ui-components/pull/1193
- https://github.com/LabKey/sampleManagement/pull/1831
- https://github.com/LabKey/biologics/pull/2127
- https://github.com/LabKey/inventory/pull/853
- https://github.com/LabKey/testAutomation/pull/1513

#### Changes
- factor out checkForExpiredSession() function and make it available via WebSocket


labkey-alan (approved):
Looks good to me, though I have not manually tested

PR #4413, opened 2023-05-10
Add document parser for plain text, Issue 43348. Fix greedy CSV parser, Issue 47814.

labkey-adam:
#### Rationale
[Issue 43348: Indexing perf: avoid autodetect parser for known text documents](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43348)
[Issue 47814: Wiki search results show HTML source when document includes a comma](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=47814)


labkey-jeckels (approved):
The code change looks good. As followup, let's be sure to add test coverage for the document preview in the search results since it seems likely to break in the future.

PR #4420, opened 2023-05-12
Refresh schema warnings after recreating views

labkey-jeckels:
#### Rationale
Recreating the DB views often clears up schema validation errors. Don't make the admin bounce Tomcat to verify.

Also, the action coverage TSV was getting truncated during export

#### Changes
* Queue a schema revalidation after recreating views
* Close the ActionsTsvWriter in a try-with-resources
* Minor cleanup

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -7830,8 +7830,6 @@ public class AdminController extends SpringActionController
         public boolean handlePost(Object o, BindException errors)
         {
             ModuleLoader.getInstance().recreateViews();
-            // Hopefully we've cleared whatever schema mismatch errors were showing related to DB views
-            ModuleLoader.getInstance().getModule(CoreModule.class).rerunSchemaCheck();
             return true;
         }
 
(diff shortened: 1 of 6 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -7820,13 +7823,15 @@ public ModelAndView getConfirmView(Object o, BindException errors)
...
-            return new HtmlView("Are you sure you want to drop and recreate all module views?");
+            return new HtmlView(HtmlString.of("Are you sure you want to drop and recreate all module views?"));
         }
 
         @Override
         public boolean handlePost(Object o, BindException errors)
         {
             ModuleLoader.getInstance().recreateViews();
+            // Hopefully we've cleared whatever schema mismatch errors were showing related to DB views
+            ModuleLoader.getInstance().getModule(CoreModule.class).rerunSchemaCheck();
```
labkey-adam:
```suggestion
            WarningService.get().rerunSchemaCheck();

```

labkey-adam (approved):
One suggestion...

core/src/org/labkey/core/admin/AdminController.java
labkey-jeckels: Thanks @labkey-adam. I pushed this change and also realized we are now double-warning about DB views due to the general schema discrepancy check.

labkey-adam (approved):

PR #4423, opened 2023-05-15
Update db consistency check for missing storage column

labkey-martyp:
#### Rationale
Update the db consistency check to handle the case a property descriptor does not have a storage column name. Throw an error explaining the issue and fix the issue by copying the PD name to the storage column name (instead of deleting and re-adding the field).

Related Reports:
http://localhost:8080/labkey/admin-doCheck.view
http://localhost:8080/labkey/<Container>/experiment-types-repair.view?domainUri=<param>

#### Changes
* Report that the PD is missing storage column name (instead of PD missing column and column missing a PD)
* On fix, add the PD name as the storage column name

### DIFF
--- a/experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
+++ b/experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
@@ -1253,9 +1253,10 @@ public class StorageProvisionerImpl implements StorageProvisioner
                 }
                 else if (st.prop != null)
                 {
-                    if (st.prop.getPropertyDescriptor().getStorageColumnName() == null)
+                    TableInfo table = DbSchema.get(report.getSchemaName(), DbSchemaType.Provisioned).getTable(report.getTableName());
+                    PropertyDescriptor pd = st.prop.getPropertyDescriptor();
+                    if (st.prop.getPropertyDescriptor().getStorageColumnName() == null && null != table.getColumn(pd.getName()))
                     {
-                        PropertyDescriptor pd = st.prop.getPropertyDescriptor();
                         pd.setStorageColumnName(pd.getName());
                         OntologyManager.updatePropertyDescriptor(pd);
                         continue;
@@ -1263,7 +1264,7 @@ public class StorageProvisionerImpl implements StorageProvisioner
 
                     if (st.colName == null)
                     {
-                        adds.addColumn(st.prop.getPropertyDescriptor());
+                        adds.addColumn(pd);
                     }
                     if (st.mvColName == null && st.prop.isMvEnabled())
                     {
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
```diff
@@ -1420,7 +1428,15 @@ public ProvisioningReport getProvisioningReport(@Nullable String domainuri)
                 domainReport.getColumns().add(status);
                 status.prop = domainProp;
                 PropertyDescriptor propDescriptor = domainProp.getPropertyDescriptor();
-                if (hardColumnNames.remove(propDescriptor.getStorageColumnName()))
+
+                if (null == propDescriptor.getStorageColumnName())
```
labkey-jeckels: I think this should also check to see if `hardColumnNames` has the storage column. Otherwise, it could end up pointing to a column that doesn't actually exist in the table

experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
```diff
@@ -1253,6 +1253,14 @@ public boolean repairDomain(Container c, String domainUri, BindException errors)
                 }
                 else if (st.prop != null)
                 {
+                    if (st.prop.getPropertyDescriptor().getStorageColumnName() == null)
+                    {
+                        PropertyDescriptor pd = st.prop.getPropertyDescriptor();
+                        pd.setStorageColumnName(pd.getName());
```
labkey-jeckels, line 1260: As commented below, we should only do this when the column is really in the table. Or am I missing a check elsewhere here that will catch it?

labkey-jeckels (changes requested):

experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
labkey-martyp: Ok I added that check. I was going for best effort fix here before just removing and adding the PD.

experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
labkey-martyp: I added that. Fixing the domain refreshes and re-runs the report which would throw a subsequent error in the case you're mentioning here, but we can save that additional step and just check initially if the column exists.

labkey-jeckels: Oh, I hadn't considered a multi-step process to get the domain back to healthy.

labkey-jeckels (approved):

PR #4426, opened 2023-05-15
Switch AssayListAction to ReadOnlyApiAction, which it already was

labkey-jeckels:
#### Rationale
As pointed out in this support forum post:

https://www.labkey.org/home/Support/LabKey%20Support%20Forum/announcements-thread.view?rowId=22966

We claim that assayList.api supports GET requests in our docs, but it doesn't. It easily can.

#### Changes
* Declare that it's a ReadOnlyApiAction
* Misc other cleanup

### DIFF
--- a/assay/src/org/labkey/assay/AssayController.java
+++ b/assay/src/org/labkey/assay/AssayController.java
@@ -151,6 +151,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.EnumMap;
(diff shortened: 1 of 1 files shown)


labkey-adam:
Here's the commit that migrated this action `ApiAction` --> `MutableApiAction`: https://github.com/LabKey/platform/commit/7f3a11bed97da567ceefcd3a49ca3a98b6acc2a7

labkey-adam (approved):
Should change the docs back, if we haven't already

PR #4428, opened 2023-05-15
Add support for moving LKB registry data class objects

cnathe:
#### Rationale
We recently added support for moving non-registry sources / data class objects in the App UI. This can be done from the grid Edit button or for a single item view the details page Manage button. This PR adds similar support for the LKB registry items and for the media data class items. The registry items have cases that use junction tables to store the one-to-many relationship between items. We've added batch trigger events to those data class table trigger factories so that the container column for those related junction list rows will bet updated along with the registry item itself that is being moved.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/4428
- https://github.com/LabKey/biologics/pull/2139 

#### Changes
- update ExperimentService.updateContainer to take generic Collection instead of just a List of Integers for the ids
- during moveDataClassObjects(), fireBatchTrigger to use the registry table trigger classes to move the related junction list rows


cnathe:
@labkey-nicka I still need to add unit/integration tests for these changes and the related PR, but creating a draft PR to get initial feedback if you wanted to take a look while I work on the tests.

api/src/org/labkey/api/data/TableInfo.java
```diff
@@ -435,7 +435,7 @@ default Set<FieldKey> getMethodRequiredFieldKeys()
 
     enum TriggerType
     {
-        INSERT, UPDATE, DELETE, SELECT, TRUNCATE;
+        INSERT, UPDATE, DELETE, SELECT, TRUNCATE, MOVE;
```
labkey-nicka, line 438: Hmm, I'm conflicted about the idea of having a `TriggerType` for `MOVE` as it is effectively an `UPDATE` but for a specific column (i.e. container). It may result in downstream confusion about triggers firing for what things. Also, what would happen if a row is updated and that update includes the container column amongst other things that are changed? Would this result in a `MOVE` being triggered as well? I realize I'm speaking to hypotheticals here. I guess I'm mainly concerned about extending this rather generic enumeration for this rather specific case.

That said, I'd be more in favor of it if Query itself understood this and handled the lifecycle of the `MOVE` trigger type. Just noodling here so happy to discuss.

cnathe: Yes, maybe we can discuss this more during the Frontend meeting today. As far as I know (and the testing I've done on sample types and data classes), you can't use a standard updateRows API call to update the container value for a database row. Stepping through it, if you try to pass in a "container" value to the updateRows it gets replaced during the DefaultQueryUpdateService.setSpecialColumns() method. 

So I get what you are saying about the move being effectively an update for a special column, but that's just it, it is a special column which does require special validation checks, permissions checks, audit entries, etc. So a new enum type seemed to fit so that we didn't mistakenly trigger any existing UPDATE triggers during this move operation. Happy to discuss more.

labkey-nicka (approved):

labkey-susanh (approved):

PR #4430, opened 2023-05-16
ELN: Resolve references via search

labkey-nicka:
#### Rationale
Extend search functionality in support of the capability to search for ELN references via full-text search. Highlights are:

1. Assay Batches and Assay Runs are now indexed with their own search documents.
1. New search categories for batches and runs.
1. Capability to search on specific search fields (e.g. "identifiersHi", "keywordMed", etc) from the API.
1. Capability to index specific fields on custom domains (still requires indexer implements the indexing in the document).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/4430
- https://github.com/LabKey/biologics/pull/2141
- https://github.com/LabKey/sampleManagement/pull/1837
- https://github.com/LabKey/labbook/pull/485

#### Changes
* Add search indexing for Assay Batches and Assay Runs. Batches are only indexed if the underlying Batch Domain has fields defined.
* Add search categories for Assay Batches and Assay Runs.
* Add state to `TableInfo` to capture columns to specifically be indexed. See `indexColumn` and `getSearchIndexColumns`.
* Refactor overloaded search methods on `SearchService` to take a `SearchOptions` object that can be built to only specify necessary parameters.
* Add `AssayBatchDocumentProvider` and `AssayRunDocumentProvider` to create search documents for these types.
* Add `SearchService.ResourceResolver`s for Assay Batches and Assay Runs to retrieve tailored search data for these types.
* Add `fields` as a parameter that can be specified on the `search-json.api` to allow requests to fine-tune which search fields are queried.
* Add `ExperimentService.getExpExperiments` to support fetching multiple Assay Batches.
* Bump `SearchModule` schema to invoke full reindex of search indices.

### DIFF
--- a/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
+++ b/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
@@ -402,7 +402,6 @@ public class DefaultAssayRunCreator<ProviderType extends AbstractAssayProvider> 
             ExperimentService.get().onRunDataCreated(context.getProtocol(), run, container, context.getUser());
 
             int runRowId = run.getRowId();
-            transaction.addCommitTask(() -> indexAssayRun(runRowId), DbScope.CommitTaskOption.POSTCOMMIT);
 
             transaction.commit();
 
@@ -444,15 +443,6 @@ public class DefaultAssayRunCreator<ProviderType extends AbstractAssayProvider> 
         {
             throw new ExperimentException(e);
         }
-    }
-
-    private void indexAssayRun(int expRunRowId)
-    {
-        SearchService ss = SearchService.get();
-        if (ss == null)
-            return;
-
-        AssayService.get().indexAssayRun(ss.defaultTask(), expRunRowId);
     }
 
     private void resolveParticipantVisits(
--- a/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
+++ b/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
@@ -401,8 +401,6 @@ public class DefaultAssayRunCreator<ProviderType extends AbstractAssayProvider> 
 
             ExperimentService.get().onRunDataCreated(context.getProtocol(), run, container, context.getUser());
 
-            int runRowId = run.getRowId();
-
             transaction.commit();
 
             // Inspect the run properties for a “prov:objectInputs” property that is a list of LSID strings.
--- a/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
+++ b/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
@@ -74,6 +74,7 @@ import org.labkey.api.query.SchemaKey;
 import org.labkey.api.query.SimpleValidationError;
 import org.labkey.api.query.ValidationError;
 import org.labkey.api.query.ValidationException;
+import org.labkey.api.search.SearchService;
 import org.labkey.api.security.User;
 import org.labkey.api.security.permissions.UpdatePermission;
 import org.labkey.api.study.assay.ParticipantVisitResolver;
@@ -89,7 +90,6 @@ import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -183,7 +183,7 @@ public class DefaultAssayRunCreator<ProviderType extends AbstractAssayProvider> 
     {
         try
         {
-            // Whether or not we need to save batch properties
+            // Whether we need to save batch properties
             boolean forceSaveBatchProps = false;
             if (batch == null)
             {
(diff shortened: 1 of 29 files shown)


api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
```diff
@@ -401,6 +401,8 @@ public ExpExperiment saveExperimentRun(
 
             ExperimentService.get().onRunDataCreated(context.getProtocol(), run, container, context.getUser());
 
+            int runRowId = run.getRowId();
```
labkey-susanh: Is this used?

api/src/org/labkey/api/data/AbstractTableInfo.java
```diff
@@ -1975,6 +1976,22 @@ public boolean supportsInsertOption(QueryUpdateService.InsertOption option)
...
+    private Map<FieldKey, SearchService.PROPERTY> searchIndexColumns = null;
+
+    public void indexColumn(FieldKey column, SearchService.PROPERTY property)
+    {
+        if (searchIndexColumns == null)
+            searchIndexColumns = new HashMap<>();
+        searchIndexColumns.put(column, property);
+    }
+
+    public SearchService.PROPERTY getSearchIndexColumns(FieldKey column)
```
labkey-susanh: Why is this plural? Seems to be returning just one.

api/src/org/labkey/api/util/StringUtilsLabKey.java
```diff
@@ -55,6 +56,23 @@
...
+        if (null != identifiable)
+            append(sb, identifiable.getName());
+    }
+
+    public static void append(StringBuilder sb, @Nullable String value)
+    {
+        if (null != value)
+        {
+            if (sb.length() > 0)
+                sb.append(" ");
```
labkey-susanh, line 71: Having the append operator be opinionated about adding the separating space is sometimes not so nice, but I suppose if you don't want the space you just use `sb.append(value)`.

api/src/org/labkey/api/util/StringUtilsLabKey.java
```diff
@@ -55,6 +56,23 @@
...
 
+    public static void append(StringBuilder sb, @Nullable Identifiable identifiable)
+    {
+        if (null != identifiable)
+            append(sb, identifiable.getName());
+    }
+
+    public static void append(StringBuilder sb, @Nullable String value)
+    {
+        if (null != value)
```
labkey-susanh: Should this be `!StringUtils.isEmpty(value)`?  Otherwise, you get an extra space when you might not expect it.

assay/src/org/labkey/assay/AssayManager.java
```diff
@@ -684,8 +700,120 @@ public void indexAssay(SearchService.IndexTask task, Container c, ExpProtocol pr
...
+
+        AssayProvider assayProvider = AssayService.get().getProvider(protocol);
+        if (assayProvider == null)
+            return false;
+
+        Domain batchDomain = assayProvider.getBatchDomain(protocol);
+        if (batchDomain == null)
+            return false;
+
+        return !batchDomain.getProperties().isEmpty();
```
labkey-susanh, line 724: Retrieving domains can be relatively expensive, but I guess we're OK with that since search indexing goes on in the background.

labkey-susanh (approved):

api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
labkey-adam: Doesn't look like it is :)

api/src/org/labkey/api/search/SearchService.java
```diff
@@ -465,7 +466,9 @@ public String normalizeHref(Path contextPath, Container c)
          *if the full-text search is deleted, providers may need to clear
          * any stored lastIndexed values.
          */
-        void indexDeleted() throws SQLException;
+        default void indexDeleted() throws SQLException
```
labkey-adam, line 469: This will make it harder to identify providers that don't implement `lastIndexed` handling

search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
```diff
@@ -1427,56 +1434,52 @@ public WebPartView getSearchView(boolean includeSubfolders, int textBoxWidth, bo
...
             iTimer.setPhase(SEARCH_PHASE.createQuery);
 
             BooleanQuery.Builder queryBuilder = new BooleanQuery.Builder();
 
             try
             {
-                QueryParser queryParser = new MultiFieldQueryParser(standardFields, getAnalyzer(), boosts);
-                queryBuilder.add(queryParser.parse(queryString), BooleanClause.Occur.MUST);
+                String[] customFields = null;
+                if (options.fields != null && !options.fields.isEmpty())
```
labkey-adam: Could use `CollectionUtils.isNotEmpty()`

labkey-adam (approved):
A couple suggestions, but no need for re-review

assay/src/org/labkey/assay/AssayManager.java
labkey-nicka: Agreed.

api/src/org/labkey/api/util/StringUtilsLabKey.java
labkey-nicka: Adjusted.

api/src/org/labkey/api/data/AbstractTableInfo.java
labkey-nicka: Updated.

api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
labkey-nicka: Fixed.

search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
labkey-nicka: Adjusted.

PR #4434, opened 2023-05-17
Issue 47289: Export List Archive if the user is an Admin of the folders of the selected Lists, else show a simple error

labkey-bpatel:
#### Rationale
Issue [47289](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47289): Export List Archive if the user is an Admin of the folders of the selected Lists, else show a simple error

Actual fix was done in https://github.com/LabKey/platform/pull/4270

This PR just changes the type of error and re-words the error message.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1517

#### Changes
* Use simple error message instead of an unauthorized exception.
* Re-word error message.

### DIFF
--- a/list/src/org/labkey/list/controllers/ListController.java
+++ b/list/src/org/labkey/list/controllers/ListController.java
@@ -966,7 +966,8 @@ public class ListController extends SpringActionController
                 // Issue 47289: Export List Archive if the user is an Admin of the folders of the selected Lists, else throw Permission error
                 if (!pair.second.hasPermission(getUser(), DesignListPermission.class))
                 {
-                    throw new UnauthorizedException(String.format("You do not have the permission to export List '%s' from Folder '%s'.", listName, pair.second.getPath()));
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):

labkey-jeckels:
This change seems fine, though it doesn't help Molly's scenario of having an export that contains lists from /Shared and then importing them into another folder on the same server

labkey-bpatel:
> This change seems fine, though it doesn't help Molly's scenario of having an export that contains lists from /Shared and then importing them into another folder on the same server

Agreed. @labkey-mohara and I have discussed this scenario, and while this is not ideal, Site Admins with /Shared access should be aware to uncheck Shared Lists before Exporting.

labkey-mohara:
Yes, I agree that either site admins will know/notice and filter out the shared ones, or if they accidentally import them into the new container, they'll be able to delete the local duplicates. It's such an edge case to ever be trampling existing copies of shared lists already in a container where you would import a list archive that I think we can ignore it/document the possibility.  Thanks!

PR #4438, opened 2023-05-17
Issue 47799: don't delete runs when not all the inputs to that run are being deleted

labkey-susanh:
#### Rationale
[Issue 47799](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47799) When we delete the container holding a subset of the inputs to a run (e.g., a sample derivation run), we don't want to delete the full run unless all of the inputs are being deleted.

#### Related Pull Requests
* 

#### Changes
* Change how we select runs for deleting to account for inputs that may not be being deleted.


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -5570,6 +5626,35 @@ private List<ExperimentRun> getRunsForRunIds(SQLFragment runIdsSQL)
         return new SqlSelector(getExpSchema(), sql).getArrayList(ExperimentRun.class);
     }
 
+    /**
+     * Generate a query to get the runIds where the supplied set of data class rowIds were used as inputs
+     * @param dataRowIdSQL -- SQL clause generating material rowIds used to limit results
+     * @return Query to retrieve set of runIds from supplied input material ids
+     */
+    private SQLFragment getTargetRunIdsUsingOtherDataIds(SQLFragment dataRowIdSQL)
```
cnathe: I don't see any usages of this method. Is it needed?

cnathe (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Not anymore.  Removed.

PR #4439, opened 2023-05-18
Issue 47928: NPE in PublishedRecordDataRegion.renderFormEnd()

labkey-jeckels:
#### Rationale
Crawlers keep hitting this, so nice to clean up the error logs.

#### Changes
* Propagate the fact that PublishSource can be null


study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -118,6 +118,7 @@
 import org.labkey.study.query.StudyQuerySchema;
 import org.springframework.beans.factory.InitializingBean;
 
+import javax.validation.constraints.Null;
```
labkey-adam: Unused?

labkey-adam (approved):

PR #4442, opened 2023-05-18
Make use of getContainerFilterForLookups() when applying container filters to QueryForeignKey added via query XML metadata

cnathe:
#### Rationale
This was made evident after the implementation of the "Move Registry Items in the LKB app projects" development was completed. If you have a registry item in a project subfolder that references other registry items in that same subfolder and then you move that item to a sibling subfolder, the lookups are broken for those items that were left behind. This is expected behavior, but in the LKB grids and details pages we have an experimental feature that makes these lookups less restrictive (i.e. allow for resolving lookups across any of the in scope containers within the app's projects (home and subfolders). This PR fixes a case that wasn't taking into account this experimental feature container filter change when applying the QueryForeignKey objects for a foreign key being added via query XML metadata.

Screenshot 2023-05-18 at 4 15 50 PM

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4428

#### Changes
* refactor QueryService.get().getProductContainerFilterForLookups() from PdLookupForeignKey to be reusable
* AbstractTableInfo.initColumnFromXml() to apply getContainerFilterForLookups() on QueryForeignKey

### DIFF
--- a/api/src/org/labkey/api/query/PdLookupForeignKey.java
+++ b/api/src/org/labkey/api/query/PdLookupForeignKey.java
@@ -95,7 +95,7 @@ public class PdLookupForeignKey extends AbstractForeignKey
             cf = new ContainerFilter.SimpleContainerFilterWithUser(user, targetContainer);
         else
         {
-            cf = QueryService.get().getProductContainerFilterForLookups(container, user, null);
+            cf = QueryService.get().getProductContainerFilterForLookups(container, user, cf);
 
             if (cf == null)
                 cf = new ContainerFilter.SimpleContainerFilterWithUser(user, container);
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/query/PdLookupForeignKey.java
```diff
@@ -95,12 +95,7 @@ else if (targetContainer != null)
             cf = new ContainerFilter.SimpleContainerFilterWithUser(user, targetContainer);
         else
         {
-            if (QueryService.get().isProductProjectsAllFolderScopeEnabled())
-            {
-                ContainerFilter lookupCf = QueryService.get().getContainerFilterForLookups(container, user);
-                if (lookupCf != null)
-                    cf = lookupCf;
-            }
+            cf = QueryService.get().getProductContainerFilterForLookups(container, user, null);
```
labkey-nicka: This should be handing through the `cf`, otherwise, we end up dropping the passed in `cf` on the floor.

```java
cf = QueryService.get().getProductContainerFilterForLookups(container, user, cf);
```

labkey-nicka (approved):

PR #4460, opened 2023-05-25
SchemaBrowser: Utilize sub-components for layout

labkey-nicka:
#### Rationale
This PR seeks to address a race condition in the Schema Browser rendering that could result in ExtJS throwing an error in which the view is unable to recover. This has been reproduced by the crawler at the end of test runs resulting a stack trace:

```
TypeError: itemDom is undefined
  isValidParent (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:35361:26)
  renderItems (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:35346:32)
  renderChildren (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:35844:16)
  invalidate (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:119913:28)
  invalidate (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:119917:28)
  flushInvalidates (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:119756:16)
  run (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:120194:12)
  flushLayouts (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:29278:25)
  updateLayout (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:29312:24)
  updateLayout (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:31188:39)
  onContentChange (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:35416:20)
  updateLayout (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:31183:58)
  resumeLayouts (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:29907:20)
  removeAll (http://localhost:8111/labkey/ext-4.2.1/ext-all-sandbox-debug.js:38280:12)
  renderQueryDetails (http://localhost:8111/labkey/query/browser/view/QueryDetails.js:777:14)
  setQueryDetails (http://localhost:8111/labkey/query/browser/view/QueryDetails.js:831:14)
  loadQueryDetails/< (http://localhost:8111/labkey/query/browser/view/QueryDetails.js:141:26)
  success (http://localhost:8111/labkey/query/browser/Caches.js:155:33)
```

I was able to reproduce this locally by slowing down the browser requests and then loading multiple views at the same time. In the end, the problem is in our call to `this.removeAll()` which removes our `loading` component. Generally, this isn't a problem, but if it is called at the right time the layout logic fails to load the underlying DOM. I've addressed this by separating `loading` and `error` rendering as well as creating a `content` component where the primary content of a tab within the SchemaBrowser is rendered.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/1528

#### Changes
* Establish constant components with itemId's `loader`, `error` and `content`.
* Use show/hide logic for `loader` and `error` components rather than removing/injecting from/in the layout. In doing so it removes the path that hit the race condition.

### DIFF
--- a/query/webapp/query/browser/view/QueryDetails.js
+++ b/query/webapp/query/browser/view/QueryDetails.js
@@ -122,37 +122,56 @@ Ext4.define('LABKEY.query.browser.view.QueryDetails', {
         this.items = [{
             xtype: 'box',
             itemId: 'loader',
-            id: 'loaderTag',
             autoEl: {
                 tag: 'p',
                 cls: 'lk-qd-loading',
-                html: 'Loading...'
-            }
+                html: 'Loading...',
+            },
+        },{
+            xtype: 'box',
+            itemId: 'error',
+            autoEl: {
+                tag: 'div',
+                cls: 'lk-qd-error',
+                html: 'There was an error',
+            },
+            hidden: true,
+        },{
+            xtype: 'container',
+            itemId: 'content'
         }];
         this.callParent();
         this.on('afterrender', this.loadQueryDetails, this, {single: true});
         this.parent.on('dependencychanged', this.refreshQueryDependencies, this);
     },
 
-    loadQueryDetails : function(){
-        this.cache.loadQueryDetails(this.schemaName, this.queryName, this.fk,
-                function(queryDetails) {
-                    this.queryDetails = queryDetails;
-                    this.setQueryDetails(this.queryDetails);
-                },
-                this.onLoadError,
-                this);
-
+    getContent : function() {
+        return this.getComponent('content');
     },
 
-    onLoadError : function(errorInfo){
-        if (Ext4.getCmp('loaderTag')) {
-            Ext4.getCmp('loaderTag').update('<div class="lk-qd-error">Error in query: ' + Ext4.htmlEncode(errorInfo.exception) + '</div>');
-        }
+    getError : function() {
+        return this.getComponent('error');
+    },
+
+    getLoading : function() {
+        return this.getComponent('loader');
+    },
+
+    loadQueryDetails : function() {
+        this.getError().hide();
+        this.getLoading().show();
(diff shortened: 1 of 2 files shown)


query/webapp/query/browser/view/QueryDetails.js
```diff
@@ -122,37 +122,56 @@ Ext4.define('LABKEY.query.browser.view.QueryDetails', {
         this.items = [{
             xtype: 'box',
             itemId: 'loader',
-            id: 'loaderTag',
             autoEl: {
                 tag: 'p',
                 cls: 'lk-qd-loading',
-                html: 'Loading...'
-            }
+                html: 'Loading...',
```
labkey-jeckels, line 128: Do you want to keep these trailing commas? I don't think they cause any problems now that don't care about Internet Explorer, but I'm not sure if we have a strong style preference one way or another.

labkey-jeckels (approved):
Wow, thanks for tracking this down.

In my testing, it changes the layout of the error scenario. Instead of two mostly duplicative errors stacked, we get the same error messages separated by the query name. I think that's fine.

Original
Updated

query/webapp/query/browser/view/QueryDetails.js
labkey-nicka: I purposely added those since that it is our preference style (at least within the apps).

labkey-nicka:
> Wow, thanks for tracking this down.
> 
> In my testing, it changes the layout of the error scenario. Instead of two mostly duplicative errors stacked, we get the same error messages separated by the query name. I think that's fine.

Thanks for testing this. I'd like to see if I can get this fixed. Seems kind of lame to show the error twice. I'll see what I can do.

PR #4461, opened 2023-05-25
updateSerialToBigInt

labkey-matthewb:
#### Rationale
Update audit sequences to bigint to match the audit rowids.  I believe this affects databases created pre-22.3.

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/audit/src/org/labkey/audit/query/AuditUpgradeCode.java
+++ b/audit/src/org/labkey/audit/query/AuditUpgradeCode.java
@@ -57,4 +57,16 @@ public class AuditUpgradeCode implements UpgradeCode
             }
         });
     }
+
+    // called from audit-23.000-23.001.sql
+    public static void updateSerialToBigInt(ModuleContext context)
+    {
+        DbScope scope = AuditSchema.getInstance().getSchema().getScope();
+        if (scope.getSqlDialect().isPostgreSQL())
+        {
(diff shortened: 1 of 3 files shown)


labkey-susanh (approved):

audit/src/org/labkey/audit/query/AuditUpgradeCode.java
```diff
@@ -57,4 +57,16 @@ public static void updateRowIdToBigInt(ModuleContext context)
...
         });
     }
+
+    // called from audit-23.000-23.001.sql
+    public static void updateSerialToBigInt(ModuleContext context)
+    {
+        DbScope scope = AuditSchema.getInstance().getSchema().getScope();
+        if (scope.getSqlDialect().isPostgreSQL())
+        {
+            var list = new SqlSelector(scope,"SELECT sequencename FROM pg_sequences WHERE schemaname='audit'").getArrayList(String.class);
```
labkey-adam, line 67: This is totally fine and not worth a change. But `forEach()` and `stream()` are options in cases like this.

labkey-adam (approved):

audit/src/org/labkey/audit/query/AuditUpgradeCode.java
labkey-matthewb: I actually thought about that.  I just wanted to see the list, and didn't bother to keep rewriting.

labkey-adam: Fair enough... thanks for the quick fix!

PR #4462, opened 2023-05-25
Add support for cross-sample type import and update via file

labkey-susanh:
#### Rationale
Often users will have a single manifest for a number of samples of different types. Many of the fields for samples are shared, but some are distinct per sample type.  Currently, we require that users divide the manifests into one file per sample type in order to import them. This work adds support from the server to do that division of the file as part of the import process, when we are told that the file is a cross-sample-type file.  Later work will adjust the UI to allow for this.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-premium/pull/96
* https://github.com/LabKey/sampleManagement/pull/1858
* https://github.com/LabKey/biologics/pull/2151
* https://github.com/LabKey/inventory/pull/868

#### Changes
* Add new `crossSampleType` parameter understood by the import actions
* Add new `MultiSampleTypeDataIterator` to do the work of splitting the given file into multiple files and processing each of these files in turn


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -4084,21 +4087,58 @@ public class ImportSamplesAction extends AbstractExpDataImportAction
         public void validateForm(QueryForm queryForm, Errors errors)
         {
             _form = queryForm;
-            _form.setSchemaName("samples");
             _insertOption = queryForm.getInsertOption();
+            _crossTypeImport = Boolean.valueOf(getParam(Params.crossTypeImport));
+            if (!_crossTypeImport)
+                _form.setSchemaName("samples");
```
XingY: if crossTypeImport with a provided queryName, should we throw error?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -4084,21 +4087,58 @@ public class ImportSamplesAction extends AbstractExpDataImportAction
...
+        @Override
+        protected void initRequest(QueryForm form) throws ServletException
+        {
+            QueryDefinition query = form.getQueryDef();
+            List<QueryException> qpe = new ArrayList<>();
+            TableInfo t = query.getTable(form.getSchema(), qpe, true);
+
+            if (!qpe.isEmpty())
+                throw qpe.get(0);
+            if (!_crossTypeImport && null != t)
```
XingY: can we have getCrossTypeImport() as a method on AbstractExpDataImportAction, which returns false by default? That way we don't need to override initRequest with such duplications

XingY: Also, is merge and update actually not supported for crossTypeImport?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -4084,21 +4087,58 @@ public class ImportSamplesAction extends AbstractExpDataImportAction
...
+            }
+
+            _auditBehaviorType = form.getAuditBehavior();
+        }
+
+        @Override
+        protected int importData(DataLoader dl, FileStream file, String originalName, BatchValidationException errors, @Nullable AuditBehaviorType auditBehaviorType, TransactionAuditProvider.@Nullable TransactionAuditEvent auditEvent) throws IOException
+        {
+            TableInfo tInfo = _crossTypeImport ? new ExpMaterialTableImpl(ExpSchema.TableType.Materials.name(), new SamplesSchema(getUser(), getContainer()), ContainerFilter.current(getContainer())) : _target;
+            QueryUpdateService updateService = _crossTypeImport ? tInfo.getUpdateService() : _updateService;
```
XingY: Why is this needed? I would think _updateService should be set to the correct one on setTarget call above.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -4084,21 +4087,58 @@ public class ImportSamplesAction extends AbstractExpDataImportAction
...
+        {
+            QueryDefinition query = form.getQueryDef();
+            List<QueryException> qpe = new ArrayList<>();
+            TableInfo t = query.getTable(form.getSchema(), qpe, true);
+
+            if (!qpe.isEmpty())
+                throw qpe.get(0);
+            if (!_crossTypeImport && null != t)
+            {
+                setTarget(t);
```
XingY: I think target tableInfo should be set for _crossTypeImport as well.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -317,7 +317,7 @@ protected int _importRowsUsingInsertRows(User user, Container container, DataIte
 
     protected int _importRowsUsingDIB(User user, Container container, DataIteratorBuilder in, @Nullable final ArrayList<Map<String, Object>> outputRows, DataIteratorContext context, @Nullable Map<String, Object> extraScriptContext)
     {
-        if (!hasPermission(user, context.getInsertOption().updateOnly ? UpdatePermission.class : InsertPermission.class))
+        if (!context.isCrossTypeImport() && !hasPermission(user, context.getInsertOption().updateOnly ? UpdatePermission.class : InsertPermission.class))
```
XingY: even with crossTypeImport, user should still need update/insert permission

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -318,7 +317,9 @@ else if (nameObj instanceof Number)
     @Override
     public DataIteratorBuilder createImportDIB(User user, Container container, DataIteratorBuilder data, DataIteratorContext context)
     {
-        assert _sampleType != null : "SampleType required for insert/update, but not required for read/delete";
+        assert context.isCrossTypeImport() || _sampleType != null : "SampleType required for insert/update, but not required for read/delete";
+        if (context.isCrossTypeImport())
```
XingY, line 343: Can the override be at loadRows level? It doesn't seem that all those extra handling associated with standard loadRows are needed so it might be cleaner to override there.

api/src/org/labkey/api/query/AbstractQueryImportAction.java
```diff
@@ -796,7 +798,7 @@ public static int importData(DataLoader dl, TableInfo target, QueryUpdateService
                     throw new RuntimeSQLException(x);
             }
         }
-        else
+        else if (!context.isCrossTypeImport())
```
XingY, line 816: In the case of samples cross type, the target should also be not null, it should be the exp.material table.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2250,4 +2263,325 @@ else if (isMergeOrUpdate)
...
+                        if (updateService == null)
+                        {
+                            _context.getErrors().addRowError(new ValidationException("No update service available for sample type " + typeData.sampleType.getName() + "'."));
+                        }
+                        else
+                        {
+                            try (DataLoader loader = DataLoader.get().createLoader(typeData.dataFile, "text/plain", true, null, null))
+                            {
+                                // We do not need to configure the loader for renamed fields as that has been taken care of when writing the file.
+                                _context.setCrossTypeImport(false);
```
XingY: this can be moved outside of loop and only call once

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2250,4 +2263,325 @@ else if (isMergeOrUpdate)
...
+        {
+            boolean hasNext = super.next();
+
+            if (_context.getErrors().hasErrors())
+                return hasNext;
+
+            if (!hasNext)
+            {
+                Collection<String> importOrderKeys = getImportOrderKeys();
+                if (!_context.getErrors().hasErrors())
```
XingY, line 2331: I think !_context.getErrors().hasErrors() is always false here

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2250,4 +2263,325 @@ else if (isMergeOrUpdate)
...
+            if (_fileNameColIndex == null)
+                _context.getErrors().addRowError(new ValidationException("Sample type cannot be determined. Provide either a '" + StringUtils.join(SAMPLE_TYPE_FIELD_NAMES, "' or '") + "' column in the data."));
+        }
+
+        @Override
+        public boolean next() throws BatchValidationException
+        {
+            boolean hasNext = super.next();
+
+            if (_context.getErrors().hasErrors())
```
XingY, line 2325: If any error, typeData files should be deleted

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2250,4 +2263,325 @@ else if (isMergeOrUpdate)
...
+                                _context.setCrossTypeImport(false);
+                                updateService.loadRows(_user, _container, loader, _context, null);
+                            }
+                            catch (SQLException | IOException e)
+                            {
+                                throw new RuntimeException(e);
+                            }
+                            finally
+                            {
+                                typeData.dataFile.delete();
```
XingY: These files only gets deleted if no other things went wrong. It's better to override close() and have close delete the files.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: Doesn't seem necessary really, but I could be convinced otherwise.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: Yes, good idea.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
labkey-susanh: Yes, but that's checked with the individual imports. No one has insert or update permission on exp.materials, so this fails, even if we set the target table to that table.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
labkey-susanh: We already override loadRows in this class, and we do want some of that processing to happen when it's a cross-type import (e.g., handling the renaming of fields).  Since the thing we are changing is the data iterator, the override here seemed more natural, but I'm happy to discuss.

api/src/org/labkey/api/query/AbstractQueryImportAction.java
labkey-susanh: For the cross-type import, the target table doesn't do much for us since we aren't actually inserting into `exp.materials`, or whatever table we choose.  All that's determined once we have read the file to pick out the types.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: done.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Not always. We can set an error if we find a cycle in the dependencies while doing an insert.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Ah, didn't know there was a `close` method. That's definitely better.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
XingY: If that's the case, then the override should be limited to SampleQUS, not abstractQUS.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: Didn't actually need to override anything since `_crossTypeImport` is defined in the base class.

Yes, merge and update are meant to be supported. Is there something here that suggests they aren't being fully supported?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: It doesn't really work out well for checking permissions. The methods `canUpdate`, `canInsert`, etc. already handle the case of `_target` being null, checking the proper permissions based on the container, which is the most we can hope for as a starting point for a cross-type import.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: See comment above.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: good point. This should now get taken care of by the `close` method.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
labkey-susanh: I think it's safe either way with the current implementation since `_target` will be null when `crossTypeImport` is true so users won't get very far even if trying to bypass the permission check by setting that parameter, but you get a better error message if it's a permissions failure, so I added a method that's overridden in `SampleTyepUpdateServiceDI`.

XingY (approved):

PR #4463, opened 2023-05-25
Export needs to use single Relativizer for all parts of the archive

labkey-matthewb:
#### Rationale
Fixes exports of samples and protocol applications with material inputs which are saved by separated savers.

#### Related Pull Requests
* 

#### Changes
*


labkey-matthewb:
Is it OK to always use LSIDRelativizer.FOLDER_RELATIVE?

labkey-jeckels (approved):
This looks good. Did you make changes to LSIDRelativizer or just move it? I'm assuming the latter but I can't easily tell in the diff.

labkey-matthewb:
> This looks good. Did you make changes to LSIDRelativizer or just move it? I'm assuming the latter but I can't easily tell in the diff.

Just moved it to make it visible to FolderaExportContext.  And hard-coded some strings to avoid having to pull along even more files

labkey-jony:
I tested this locally, but ran into the same issue that was reported, so I'm not sure what is going wrong here.

api/src/org/labkey/api/exp/XarSource.java
```diff
@@ -202,7 +202,7 @@ private String createIllegalReferenceMessage(ExpRun experimentRun, ExpProtocolAp
         return message;
     }
 
-    public ExpMaterial getMaterial(ExpRun experimentRun, ExpProtocolApplication protApp, String materialLSID) throws XarFormatException
+    private ExpMaterial getMaterialOLD(ExpRun experimentRun, ExpProtocolApplication protApp, String materialLSID) throws XarFormatException
```
labkey-jeckels: Can we delete this now?

labkey-jeckels (approved):
Looks OK to me.

PR #4465, opened 2023-05-28
Issue 47268: Export Does Not Include Failed Lookup Values

labkey-bpatel:
#### Rationale
Issue [47268](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47268): Export Does Not Include Failed Lookup Values

#### Changes
* Export broken lookup values just like it is displayed in the UI


api/src/org/labkey/api/data/DisplayColumn.java
```diff
@@ -538,12 +538,29 @@ public String getTsvFormattedValue(RenderContext ctx)
...
+        Object value = getDisplayValue(ctx);
+        if (null == value)
+        {
+            if (this.getColumnInfo().isLookup())
+            {
+                value = ctx.get(this.getColumnInfo().getName());
+
+                // Exports how it is displayed in the UI as per Issue 47268: Export Does Not Include Failed Lookup Values
+                if (null != value)
+                    return "<" + ctx.get(this.getColumnInfo().getName()) + ">";
```
labkey-jeckels:
```suggestion
                    return "<" +value + ">";
```

labkey-jeckels (approved):

labkey-jeckels (changes requested):
This approach means that integer lookup values that do resolve are treated as strings. That means they get left-aligned, don't sort/filter as desired, in Excel, etc.

Can we be more targeted and only flip to string cells for broken lookups?

labkey-bpatel:
> This approach means that integer lookup values that do resolve are treated as strings. That means they get left-aligned, don't sort/filter as desired, in Excel, etc.
> 
> Can we be more targeted and only flip to string cells for broken lookups?

Excellent point! Now updated

labkey-jeckels (approved):

PR #4467, opened 2023-05-29
Project level data type - remove experimental flag

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1200
- https://github.com/LabKey/labkey-ui-premium/pull/97
- https://github.com/LabKey/platform/pull/4467
- https://github.com/LabKey/inventory/pull/869
- https://github.com/LabKey/sampleManagement/pull/1860
- https://github.com/LabKey/biologics/pull/2152

#### Changes
* remove isProductProjectDataTypeSelectionEnabled experimental flag
* allow querying for container data type exclusions by container path
* clean up datatypeexclusion on data type and container delete

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -8199,10 +8199,10 @@ public class ExperimentServiceImpl implements ExperimentService, ObjectReference
     @Override
     public void removeDataTypeExclusion(Collection<Integer> rowIds, DataTypeForExclusion dataType)
     {
-        _removeDataTypeExclusion(rowIds, dataType, null);
+        removeDataTypeExclusion(rowIds, dataType, null);
     }
 
-    public void _removeDataTypeExclusion(Collection<Integer> rowIds, DataTypeForExclusion dataType, @Nullable String excludedContainerId)
+    private void removeDataTypeExclusion(Collection<Integer> rowIds, DataTypeForExclusion dataType, @Nullable String excludedContainerId)
     {
         SQLFragment sql = new SQLFragment("DELETE FROM  ")
                 .append(getTinfoDataTypeExclusion())
@@ -8237,7 +8237,7 @@ public class ExperimentServiceImpl implements ExperimentService, ObjectReference
             String excludedContainerId = excludedContainerIdOrPath;
             if (!GUID.isGUID(excludedContainerIdOrPath))
             {
-                Container container = ContainerManager.ensureContainer(excludedContainerIdOrPath);
+                Container container = ContainerManager.getForPath(excludedContainerIdOrPath);
                 if (container != null)
                     excludedContainerId = container.getId();
             }
@@ -8319,7 +8319,7 @@ public class ExperimentServiceImpl implements ExperimentService, ObjectReference
         }
 
         if (!toRemove.isEmpty())
-            _removeDataTypeExclusion(toRemove, dataType, excludedContainerId);
+            removeDataTypeExclusion(toRemove, dataType, excludedContainerId);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8183,19 +8186,40 @@ public void addDataTypeExclusion(int rowId, DataTypeForExclusion dataType, Strin
...
+
+    public void _removeDataTypeExclusion(Collection<Integer> rowIds, DataTypeForExclusion dataType, @Nullable String excludedContainerId)
     {
         SQLFragment sql = new SQLFragment("DELETE FROM  ")
                 .append(getTinfoDataTypeExclusion())
-                .append(" WHERE DataTypeRowId = ? AND DataType = ? AND ExcludedContainer = ?");
-        sql.add(rowId);
+                .append(" WHERE DataTypeRowId ");
+        sql.appendInClause(rowIds, getExpSchema().getSqlDialect());
+        sql.append(" AND DataType = ?");
```
labkey-susanh, line 8211: Should this be `sql.append(" AND DataType = ").appendValue(dataType.name())` (and similar below)?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8208,8 +8232,16 @@ public void removeDataTypeExclusion(int rowId, DataTypeForExclusion dataType, St
             and = " AND ";
         }
 
-        if (!StringUtils.isEmpty(excludedContainerId))
+        if (!StringUtils.isEmpty(excludedContainerIdOrPath))
         {
+            String excludedContainerId = excludedContainerIdOrPath;
+            if (!GUID.isGUID(excludedContainerIdOrPath))
+            {
+                Container container = ContainerManager.ensureContainer(excludedContainerIdOrPath);
```
labkey-susanh: `ensureContainer` is going to create the container if it doesn't exist. Is that what we want?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8183,19 +8186,40 @@ public void addDataTypeExclusion(int rowId, DataTypeForExclusion dataType, Strin
...
+        new SqlExecutor(getExpSchema()).execute(sql);
+    }
+
+    @Override
+    public void removeDataTypeExclusion(Collection<Integer> rowIds, DataTypeForExclusion dataType)
+    {
+        _removeDataTypeExclusion(rowIds, dataType, null);
+    }
+
+    public void _removeDataTypeExclusion(Collection<Integer> rowIds, DataTypeForExclusion dataType, @Nullable String excludedContainerId)
```
labkey-susanh: Obviously not a functional problem, but the use of `_` as a prefix to a public method name seems not quite right stylistically.  This name doesn't seem like it will conflict with other names, so the `_` doesn't seem necessary.  (Maybe naming conventions have changed for Java, but I don't think `_` is usually used as a prefix for private function names either.)

labkey-susanh (changes requested):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: good catch. fixed

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: Hmm... Is one preferred over the other?

labkey-susanh: I had thought the `appendValue` was becoming the new preferred syntax, but the use of `add` obviously still works.

labkey-susanh (approved):

PR #4468, opened 2023-05-29
Issue 47818 : export options to containing metadata only

labkey-klum:
#### Rationale
Similar to some of the study options to export dataset definitions only, we would like to also offer this option for the following data types:
- lists
- sample types
- data classes
- assay (exclude runs)

**Assay**
When a folder containing assay data is exported, a new (child) option will be available to elect to have the assay runs exported with the assay design, the name of the parent option is : Experiments, Protocols, and Runs. If this option is unchecked then only the assay design should be exported. The assay option is not available when creating a folder from a template:
image

**Lists, Sample Types, and Data Classes**
To make the UI less confusing for users, we are adding top level options to export either the design or data for each of these data types. When exporting from a folder, both design and data will be checked by default. When creating from a template, the data options will be unchecked by default.
image

[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47818)

**Study Child Options**
I added the ability to set the default check state for child writers so that when creating a study folder from a template, the data options for assay, sample type, and study datasets are unchecked.

### Manual Test Plan
- [x] Validate new export options for Assay runs
- [x] Validate new export options for Lists
- [x] Validate new export options for Sample Types
- [x] Validate new export options for Data Classes
- [x] Verify UI default states for create from template versus folder export
- [x] Verify import of sample and data class data from a legacy folder archive. The format of the archive has changed with this PR, but we still need to be able to import archives in the old format. Test by creating an export from labkey.org or some other server without this change and ensure you can still import it.


XingY:
I find the UI a bit weird... Usually unchecking a parent checkbox would uncheck all children. But here the parent can be checked with all children unchecked. I understand this is an existing pattern used for study options...

@labkey-andrewo any thoughts?

labkey-andrewo:
Can I see this somewhere live? I don't know this piece of UI. It does seem wrong to me to allow parent selection without all children selected. For example, if I select "Experiments, Protocols, and Runs" I guess the point is: I can uncheck "Experiment Runs" and get only "Experiments and Protocols"? And I must always get Experiments and Protocols together? I could never only download "Protocols"? I'm not even sure if these are scientifically meaningful questions. :-) 

Lists and Lists Data to me are not distinct enough options. What is a List without data? Does "List Data" somehow omit "lists" themselves? It's fairly unclear what I'd get if I chose one or the other. 

Sample Types and Data Classes definitely should be two independant checkboxes though--the parent option "Sample Types and Data Classes" is a confusing redundancy. This implies I could choose "Sample Types and Data Classes" and _uncheck_ "Data Classes"...but still get both things in the download?

XingY (approved):

cnathe:
> I find the UI a bit weird... Usually unchecking a parent checkbox would uncheck all children. But here the parent can be checked with all children unchecked. I understand this is an existing pattern used for study options...

In looking at the existing export options, I see that in the study case the "Dataset Definitions" and the "Dataset Data" options are siblings instead of the data option being nested under the definitions option. I guess I would have expected the same here (i.e. just adding the new "data" options as siblings to existing items and maybe some text updates to existing items if it made sense).

core/src/org/labkey/core/admin/exportFolder.jsp
```diff
@@ -113,7 +113,7 @@ Ext4.onReady(function(){
                         inputValue: childName,
                         checked: checked,
                         objectType: "child",
-                        parentId: parentName
+                        parentId: parentName.replaceAll(',', '')
```
cnathe, line 116: I was curious what this would be for, but I assume it is for that "Experiments, Protocols, and Runs" option. Is that right?

study/src/org/labkey/study/writer/AssayDatasetData.java
```diff
@@ -18,10 +18,4 @@ public class AssayDatasetData implements InternalStudyWriter
     public void write(StudyImpl object, StudyExportContext ctx, VirtualFile vf) throws Exception
     {
     }
-
-    @Override
-    public boolean includeWithTemplate()
-    {
-        return false;
```
cnathe, line 25: am I missing somewhere that this `false` was moved, or does this mean that these data options will now start showing for the "create from template" case? I believe we want only the "shape / definitions of the data" in the template case and not the data themselves. Does this change that?

study/src/org/labkey/study/writer/SampleTypeDatasetData.java
```diff
@@ -18,10 +18,4 @@ public class SampleTypeDatasetData implements InternalStudyWriter
     public void write(StudyImpl object, StudyExportContext ctx, VirtualFile vf) throws Exception
     {
     }
-
-    @Override
-    public boolean includeWithTemplate()
-    {
-        return false;
```
cnathe, line 25: same question here re: "create from template" case

study/src/org/labkey/study/writer/StudyDatasetData.java
```diff
@@ -18,10 +18,4 @@ public class StudyDatasetData implements InternalStudyWriter
     public void write(StudyImpl object, StudyExportContext ctx, VirtualFile vf) throws Exception
     {
     }
-
-    @Override
-    public boolean includeWithTemplate()
-    {
-        return false;
```
cnathe, line 25: same question here re: "create from template" case

core/src/org/labkey/core/admin/exportFolder.jsp
labkey-klum: yes, this was a bug that didn't become apparent until we had a child element nested under a parent with comma's in it's label.

study/src/org/labkey/study/writer/AssayDatasetData.java
labkey-klum: It would, I thought the original decision to not include these data types for templates was arbitrary and now that lists and sample types & data classes allow the data option via template (due to legacy behavior) that it seemed odd to me that study would be special.

But I would be totally fine if we kept the previous behavior. @labkey-hannah do you have an opinion on this?

cnathe: I don't entirely remember the reasoning off the top of my head, but I believe the data was partly left off because of perf. A study or sample type or other data type could have a lot of data and creating the template could then take a very long time. Also, "template" to me seems to imply definitions and not data, but that could just be me.

api/src/org/labkey/api/admin/FolderArchiveDataTypes.java
```diff
@@ -44,6 +45,10 @@ private FolderArchiveDataTypes() {}
     public static final String FILE_BROWSER_SETTINGS = "File browser admin configurations";
     public static final String ETLS = "ETL Definitions";
     public static final String SAMPLE_TYPES_AND_DATA_CLASSES = "Sample Types and Data Classes";
```
cnathe: is this `SAMPLE_TYPES_AND_DATA_CLASSES` option still relevant now or can it be removed?

core/src/org/labkey/core/admin/exportFolder.jsp
```diff
@@ -100,8 +100,8 @@ Ext4.onReady(function(){
             });
 
             if (Ext4.isArray(children)) {
-                Ext4.each(children, function(childName) {
-                    childName = Ext4.util.Format.htmlEncode(childName);
+                Ext4.each(children, function(child) {
+                    childName = Ext4.util.Format.htmlEncode(child.name);
```
cnathe: minor
```suggestion
                    var childName = Ext4.util.Format.htmlEncode(child.name);
```

experiment/src/org/labkey/experiment/samples/AbstractExpFolderImporter.java
```diff
@@ -161,12 +120,8 @@ else if (file.startsWith(SampleTypeAndDataClassFolderWriter.DATA_CLASS_PREFIX))
                     typesReader.setStrictValidateExistingSampleType(xarCtx.isStrictValidateExistingSampleType());
                     typesReader.parseAndLoad(false, ctx.getAuditBehaviorType());
 
-                    // Import data classes first because the media sample type (RawMaterials) has a lookup to the ingredient data class.
-                    // Registry files need to imported in a particular order because some files rely on data from other files.
```
cnathe, line 166: might be good to bring this comment (about why the sort order is important) over to the DataClassFolderImport.importDataFiles() method

experiment/src/org/labkey/experiment/samples/AbstractExpFolderWriter.java
```diff
@@ -109,178 +79,17 @@ public boolean show(Container c)
...
-            {
-                TableInfo tinfo = userSchema.getTable(sampleType.getName());
-                if (tinfo != null)
-                {
-                    SimpleFilter filter = SimpleFilter.createContainerFilter(ctx.getContainer());
-
-                    // filter only to the specific samples
-                    if (_xarCtx != null && _xarCtx.getIncludedSamples().containsKey(sampleType.getRowId()))
-                        filter.addInClause(FieldKey.fromParts("RowId"), _xarCtx.getIncludedSamples().get(sampleType.getRowId()));
-
```
cnathe, line 237: I don't see this in the SampleTypeFolderWriter.writeSampleTypeDataFiles method that was factored out of this file. Was it no longer needed or relevant?

cnathe (approved):

api/src/org/labkey/api/admin/FolderArchiveDataTypes.java
labkey-klum: yes, those can be removed

experiment/src/org/labkey/experiment/samples/AbstractExpFolderImporter.java
labkey-klum: Yep, I moved it to the `Factory.getPriority` since that controls the import order.

experiment/src/org/labkey/experiment/samples/AbstractExpFolderWriter.java
labkey-klum: Ah, good point. I should have mentioned this previously. The filtering ability was originally added to support Notebook snapshots (via folder archive) in order to reduce the payload of the archive. We have moved away from that technique and there aren't any usages of `XarExportContext` so that class and filtering code could be removed.

PR #4470, opened 2023-05-30
Projects: Selection of data structures to be shown in Projects - Designer UI updates

cnathe:
#### Rationale
Recent PRs have allowed the app project settings page to exclude certain data types (sample types, data classes, assay designs, and storage) from the UI for that project. This PR introduces the reverse, allowing for a data type field editor / designer to show a panel that allow for project exclusions to be made. The two methods both insert/update the exclusions in the same DB table and add similar audit entries.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1202

#### Changes
- domain kind properties support for posting excludedContainerIds
- saveDomain / saveProtocol updates to call ExperimentService.get().ensureDataTypeContainerExclusions
- refactors in ExperimentService to allow for reuse of helpers to add/remove data type exclusions and add audit entries

### DIFF
--- a/api/src/org/labkey/api/exp/api/SampleTypeService.java
+++ b/api/src/org/labkey/api/exp/api/SampleTypeService.java
@@ -104,10 +104,6 @@ public interface SampleTypeService
     ExpSampleType createSampleType(Container container, User user, String name, String description, List<GWTPropertyDescriptor> properties, List<GWTIndex> indices, int idCol1, int idCol2, int idCol3, int parentCol, String nameExpression)
             throws ExperimentException, SQLException;
 
-    @NotNull
-    ExpSampleType createSampleType(Container container, User user, String name, String description, List<GWTPropertyDescriptor> properties, List<GWTIndex> indices, String nameExpression)
-            throws ExperimentException, SQLException;
-
     /**
      * (MAB) todo need a builder interface, or at least  parameter bean
      */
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/exp/api/SampleTypeService.java
```diff
@@ -118,14 +114,13 @@ ExpSampleType createSampleType(Container container, User user, String name, Stri
 
     @NotNull
     ExpSampleType createSampleType(Container container, User user, String name, String description, List<GWTPropertyDescriptor> properties, List<GWTIndex> indices, int idCol1, int idCol2, int idCol3, int parentCol,
-                                   String nameExpression, String aliquotNameExpression, @Nullable TemplateInfo templateInfo, @Nullable Map<String, String> importAliases, @Nullable String labelColor, @Nullable String metricUnit,
-                                   @Nullable Container autoLinkTargetContainer, @Nullable String autoLinkCategory, @Nullable String category)
+                                   String nameExpression, String aliquotNameExpression, @Nullable TemplateInfo templateInfo, @Nullable Map<String, String> importAliases, @Nullable String labelColor, @Nullable String metricUnit)
```
labkey-susanh, line 117: We seriously need to change the signature for these `createSampleType` methods. 21 arguments is just ridiculous.  But that's probably for another day.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8293,6 +8300,67 @@ public void ensureContainerDataTypeExclusions(@Nullable DataTypeForExclusion dat
...
+    @Override
+    public String getDisabledDataTypeAuditMsg(DataTypeForExclusion type, List<Integer> ids, boolean isUpdate)
+    {
+        StringBuilder builder = new StringBuilder();
+        if (ids != null && (isUpdate || !ids.isEmpty()))
+        {
+            if (isUpdate && ids.isEmpty())
+                builder.append(type.name()).append( " exclusion has been cleared.");
+            else
+                builder.append("Excluded ").append(type.name()).append(": ").append(StringUtils.join(ids, ", ")).append(".\n");
```
labkey-susanh, line 8444: Do we always want the new line in the audit message?  Usually seems best to leave the adding of separators to the caller so the caller isn't forced to deal with a separator you don't want.  But if this is going to add the new line, should it also be added for the update case?

labkey-susanh (approved):

api/src/org/labkey/api/exp/api/SampleTypeService.java
cnathe: Agreed, 100%

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
cnathe: Done. I think in this case we actually want the new line for each within this method. Otherwise we need the extra code on the caller side for the null check and adding new line only when not null.

PR #4482, opened 2023-06-02
Clearing list LastIndexed is very slow, Issue 47998

labkey-adam:
#### Rationale
Clearing last indexed is slow... let's speed it up! https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47998

#### Key Changes
* Speed up clearing LastIndexed in lists by adding `WHERE LastIndexed IS NOT NULL` condition to the UPDATE
* Stop clearing LastIndexed on lists that aren't indexed
* Clear LastIndexed column when turning on each-item or attachment indexing
* Clear entire list LastIndexed value when turning on entire-list indexing
* Clear existing docs from index when turning off each-item, entire list, or attachment indexing; stop doing this clearing every time a list is indexed
* Reindex the entire list document when an item is deleted (if data is included in the doc)


labkey-jeckels (approved):

list/src/org/labkey/list/model/ListManager.java
```diff
@@ -457,137 +496,31 @@ private void indexList(@NotNull IndexTask task, ListDefinition list, boolean des
...
-                Runnable r = () ->
-                {
-                    SimpleFilter filter = new SimpleFilter(FieldKey.fromParts(list.getKeyName()), item.getKey());
-                    int count = indexItems(task, list, filter);
-                    if (0 == count)
-                        LOG.info("I should be deleting!");
-                };
-                _addIndexTask(r, SearchService.PRIORITY.item);
+                Runnable r = () -> ss.deleteResource(getDocumentId(list, entityId));
+                task.addRunnable(r, SearchService.PRIORITY.delete);
```
labkey-matthewb: It looks a little odd that one IF considers transactions and the other doesn't.  Feels like a lot probability case, but a roll back would leave this item deleted from the index.

labkey-adam: Moved both checks into a post-commit task for consistency

PR #4486, opened 2023-06-05
Make module dependency sorting deterministic based on module name

labkey-jeckels:
#### Rationale
Module ordering is not deterministic, as it currently relies on the order of the file list, which is OS-specific

#### Changes
* Sort first by module name, and then by module dependencies, to ensure a deterministic order


labkey-susanh (approved):

labkey-martyp (approved):
Looks great!

PR #4497, opened 2023-06-08
Issue 48029: Improve channel-based file copy

labkey-jeckels:
#### Rationale
Channel-based copies aren't guaranteed to copy all bytes in a single call, so we need to loop until we're done.

#### Changes
* Add some logging
* Loop until all bytes have been consumed
* Refactor to allow reuse in WebDav PUT too

### DIFF
--- a/api/src/org/labkey/api/util/FileStream.java
+++ b/api/src/org/labkey/api/util/FileStream.java
@@ -76,39 +76,7 @@ public interface FileStream
         {
             throw new IOException("Destination file [" + dest.getAbsolutePath() + "] already exists and could not be deleted");
         }
-
-        boolean success = false;
-        long expected = s.getSize();
-        long actual = 0;
-        long bytesCopied;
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/util/FileStream.java
```diff
@@ -76,39 +76,7 @@ static void transferToImpl(FileStream s, File dest) throws IOException
...
-            if (success)
-            {
-                LOG.debug("Finished transferring " + actual + " bytes to " + dest);
-            }
-            else
-            {
-                LOG.debug("Failed during transfer, but successfully copied at least " + actual + " bytes to " + dest);
-            }
-        }
+        FileUtil.copyFile(s.getInputChannel(), s.getSize(), dest);
```
labkey-adam: Does somebody need to close the input channel?

labkey-jeckels: Yes. Thank you. Fixed.

labkey-adam (approved):

PR #4500, opened 2023-06-09
Fix IssueObject JSON serialization, #48059

labkey-adam:
#### Rationale
`PageFlowUtil.encodeObject()` was throwing when attempting to serialize an `IssueObject` because `JSONObject` was performing very aggressive interrogation of `User` objects returned from `getNotifyListUserEmail()`. This blocked updating of issues that included a notify list. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48059

The main fix is to annotate `User` to JSON serialize as its int user ID. For good measure, we also mark `getNotifyListUserEmail()` with `@Transient` (so BeanUtils skips this method on the top-level issue object) and `@JSONPropertyIgnore` so `JSONObject` skips this method when encountering embedded issues, for example, issues referenced from comments.

### DIFF
--- a/api/src/org/labkey/api/security/User.java
+++ b/api/src/org/labkey/api/security/User.java
@@ -20,6 +20,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONObject;
+import org.json.JSONString;
 import org.labkey.api.attachments.Attachment;
 import org.labkey.api.attachments.AttachmentService;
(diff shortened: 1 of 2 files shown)


labkey-klum (approved):

api/src/org/labkey/api/security/User.java
```diff
@@ -676,4 +677,10 @@ public void setExpirationDate(Date expirationDate)
     {
         _expirationDate = expirationDate;
     }
+
+    @Override
+    public String toJSONString()
+    {
+        return String.valueOf(getUserId());
```
labkey-matthewb, line 684: Doesn't this need a quote()

PR #4502, opened 2023-06-09
Stop serializing old beans into update forms

labkey-adam:
#### Rationale
Serializing beans into update forms is problematic and seems unnecessary. This does away with serialization in issue, announcement, and visit updates.

- Old code path: serialize old bean (via compressed JSON) into the form, deserialize posted JSON old values into a newly constructed bean, overlay newly posted strings onto that bean
- New code path: after post, read old bean from the database, overlay new posted strings onto it

Not yet tackling optimistic concurrency checks, but that shouldn't be difficult to add.

### DIFF
--- a/announcements/src/org/labkey/announcements/AnnouncementsController.java
+++ b/announcements/src/org/labkey/announcements/AnnouncementsController.java
@@ -1245,24 +1245,23 @@ public class AnnouncementsController extends SpringActionController
         @Override
         public boolean handlePost(AnnouncementForm form, BindException errors)
         {
-            AnnouncementModel ann = form.selectAnnouncement();
+            AnnouncementModel oldAnn = form.selectAnnouncement();
 
-            if (null == ann)
+            if (null == oldAnn)
             {
                 throw new NotFoundException("Announcement");
             }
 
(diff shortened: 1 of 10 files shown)


labkey-matthewb (approved):

labkey-matthewb:
The only potential use for oldValues that I can think of is to provide a diff of changed values in the case of optimistic conflict.  However, we've never had a pattern for merge in the case of conflict, and I doubt we will.  We should kill oldValues until there is a clear use.  Switching to modified for optimistic concurrency seems fine.

labkey-adam:
> The only potential use for oldValues that I can think of is to provide a diff of changed values in the case of optimistic conflict. However, we've never had a pattern for merge in the case of conflict, and I doubt we will. We should kill oldValues until there is a clear use. Switching to modified for optimistic concurrency seems fine.

But it's not nanosecond granularity 😄

labkey-adam:
BVT and Daily suites are green

labkey-jeckels:
I know I'm late to review this, but I'm fine with the approach and am glad to remove the old pattern

PR #4505, opened 2023-06-12
Add support for moving ELNs between containers

labkey-susanh:
#### Rationale
Previous stories have done the work to allow samples and data class objects to be moved to other folders within the application folder structure. This PR and its relatives add support for moving notebooks as well.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/110

#### Changes
* Refactor utility methods used for moving samples and data class objects to make them accessible outside of Experiment
* Add method for moving properties managed by `OntologyManager` from one container to another
* update `moveAttachments` method in AttachmentService to account for attachments that might be stored on disk

### DIFF
--- a/api/src/org/labkey/api/data/ContainerManager.java
+++ b/api/src/org/labkey/api/data/ContainerManager.java
@@ -68,6 +68,7 @@ import org.labkey.api.security.SecurityPolicyManager;
 import org.labkey.api.security.User;
 import org.labkey.api.security.permissions.AdminPermission;
 import org.labkey.api.security.permissions.DeletePermission;
+import org.labkey.api.security.permissions.InsertPermission;
 import org.labkey.api.security.permissions.Permission;
 import org.labkey.api.security.permissions.ReadPermission;
 import org.labkey.api.security.roles.AuthorRole;
@@ -94,6 +95,7 @@ import org.labkey.api.view.ViewContext;
 import org.labkey.api.writer.MemoryVirtualFile;
 import org.labkey.folder.xml.FolderDocument;
 import org.springframework.validation.BindException;
+import org.springframework.validation.Errors;
 
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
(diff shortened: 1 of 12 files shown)


api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -2494,6 +2498,75 @@ private static Container getForPathAlias(String path)
         return ret.length == 0 ? null : ret[0];
     }
 
+    public static Container getMoveTargetContainer(String entityType, Container sourceContainer, User user, String targetIdOrPath, Errors errors)
```
labkey-nicka: nit: Annotate with `@Nullable`

api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -2494,6 +2498,75 @@ private static Container getForPathAlias(String path)
...
+        boolean moveFromProjectToChild = current.isProject() && target.getParent().equals(current);
+        boolean moveFromChildToProject = !current.isProject() && current.getParent().isProject() && current.getParent().equals(target);
+        boolean moveFromChildToSibling = !current.isProject() && current.getParent().isProject() && current.getParent().equals(target.getParent());
+
+        return moveFromProjectToChild || moveFromChildToProject || moveFromChildToSibling;
+    }
+
+    public static int updateContainer(TableInfo dataTable, String idField, Collection<?> ids, Container targetContainer, User user, boolean withModified)
+    {
+        SQLFragment dataUpdate = new SQLFragment("UPDATE ").append(dataTable)
```
labkey-nicka: Wrap in transaction? Otherwise, indicate to callers that this is not a transacted method and must be transacted by them (sorry, I realize that `ContainerManager` is one of the worst examples of transaction boundaries).

core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
```diff
@@ -520,6 +520,18 @@ public void moveAttachments(Container newContainer, List<AttachmentParent> paren
                 for (Attachment att : atts)
                 {
                     filename = att.getName();
+                    if (parent instanceof AttachmentDirectory)
```
labkey-nicka: nit: Consider using `parent instanceof AttachmentDirectory parentDir` and then using `parentDir` below.

labkey-nicka (approved):

api/src/org/labkey/api/data/ContainerManager.java
labkey-susanh: Done. Thanks.

core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
labkey-susanh: Done. Thanks.

PR #4518, opened 2023-06-15
Issue 48074: bulk file update on auto-incrementing 3 key dataset

XingY:
#### Rationale
Issue [48074](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48074): bulk file update on auto-incrementing 3 key dataset

#### Related Pull Requests
* 

#### Changes
* bring back the non-functional merge option for dataset with managed 3rd keys

### DIFF
--- a/study/src/org/labkey/study/StudyModule.java
+++ b/study/src/org/labkey/study/StudyModule.java
@@ -418,8 +418,8 @@ public class StudyModule extends SpringModule implements SearchService.DocumentP
                 "Separate study tables into three groups 'datasets', 'specimens', and 'design'", false);
 
         AdminConsole.addExperimentalFeatureFlag(DatasetQueryView.EXPERIMENTAL_ALLOW_MERGE_WITH_MANAGED_KEYS,
-                "Allow merge of study dataset that uses server manage additional key fields",
-                "Merging of dataset that uses server managed third key (such as GUID or auto RowId) is currently not supported. Unexpected outcome might be experienced when merge is performed.",
+                "Allow merge of study dataset that uses server-managed additional key fields",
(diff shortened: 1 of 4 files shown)


study/src/org/labkey/study/StudyModule.java
```diff
@@ -416,6 +417,11 @@ protected void startupAfterSpringConfig(ModuleContext moduleContext)
         AdminConsole.addExperimentalFeatureFlag(StudyQuerySchema.EXPERIMENTAL_STUDY_SUBSCHEMAS, "Use sub-schemas in Study",
                 "Separate study tables into three groups 'datasets', 'specimens', and 'design'", false);
 
+        AdminConsole.addExperimentalFeatureFlag(DatasetQueryView.EXPERIMENTAL_ALLOW_MERGE_WITH_MANAGED_KEYS,
+                "Allow merge of study dataset that uses server manage additional key fields",
+                "Merging of dataset that uses server managed third key (such as GUID or auto RowId) is currently not supported. Unexpected outcome might be experienced when merge is performed.",
```
labkey-jeckels:
```suggestion
                "Allow merge of study dataset that uses server managed additional key fields",
                "Merging of dataset that uses server managed third key (such as GUID or auto RowId) is not officially supported. Unexpected outcome might be experienced when merge is performed.",

```

labkey-jeckels (approved):

study/src/org/labkey/study/StudyModule.java
labkey-susanh: I would suggest "server-managed" instead of "server managed".

PR #4521, opened 2023-06-16
48037: Change Audit Detail Level behavior for dataset events

labkey-klum:
#### Rationale
Clean up and add more consistency to the study dataset audit logging in order to get the desired behaviors with respect to audit logging XML metadata overrides. The update service for datasets doesn't consistently go through the data iterator framework and dataset auditing occurs in 3 ways:
- Through the `DetailedAuditLogDataIterator`
- Calls directly into the `DatasetAuditHandler.addAuditEvent`
- Add hoc logging of a `DatatsetAuditEvent`

[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48037)

#### Changes
- Be more consistent about passing in the dataset `TableInfo` into `DatasetAuditHandler.addAuditEvent` so that any XML metadata overrides can be respected.
- Move `StudyServiceImpl.addDatasetAuditEvent` into `DatasetAuditHandler`

### DIFF
--- a/study/src/org/labkey/study/controllers/StudyController.java
+++ b/study/src/org/labkey/study/controllers/StudyController.java
@@ -29,7 +29,25 @@ import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONObject;
-import org.labkey.api.action.*;
+import org.labkey.api.action.ApiJsonForm;
+import org.labkey.api.action.ApiResponse;
+import org.labkey.api.action.ApiSimpleResponse;
+import org.labkey.api.action.ConfirmAction;
+import org.labkey.api.action.FormApiAction;
+import org.labkey.api.action.FormHandlerAction;
+import org.labkey.api.action.FormViewAction;
+import org.labkey.api.action.HasAllowBindParameter;
+import org.labkey.api.action.HasViewContext;
+import org.labkey.api.action.Marshal;
+import org.labkey.api.action.Marshaller;
+import org.labkey.api.action.MutatingApiAction;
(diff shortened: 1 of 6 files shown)


study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -2510,10 +2527,12 @@ protected int importData(DataLoader dl, FileStream file, String originalName, Ba
 
             if (!result.getKey().isEmpty())
             {
-                // Log the import
+                // Log the import when SUMMARY is configured, if DETAILED is configured the DetailedAuditLogDataIterator will handle each row change.
+                // It would be nice in the future to replace the DetailedAuditLogDataIterator with a general purpose AuditLogDataIterator
```
labkey-susanh, line 2531: I agree this would be a good thing to do.

study/src/org/labkey/study/model/DatasetDefinition.java
```diff
@@ -1780,6 +1784,29 @@ else if (existingRecord != null && existingRecord.size() > 0)
 
             return event;
         }
+
+        /**
+         * General purpose method to add dataset audit events.
+         * @param requiredAuditType The expected audit behavior type. If this does not match the type set on the
+         *                          dataset, then the event will not be logged.
+         */
+        public void addAuditEvent(User user, Container c, AuditBehaviorType requiredAuditType, String comment, UploadLog ul /*optional*/)
```
labkey-susanh:
```suggestion
        public void addAuditEvent(User user, Container c, AuditBehaviorType requiredAuditType, String comment, @Nullable UploadLog ul)

```

labkey-susanh (approved):

PR #4525, opened 2023-06-20
Remove config options for deprecated Universal Google Analytics

labkey-jeckels:
#### Rationale
The universe was too restricting, so Google has switched to Google Analytics 4. Universal Google Analytics stops accepting new data July 1, so there's no need to retain our integration for 23.7 and beyond

#### Changes
* Purge Universal Google Analytics settings and code
* Minor code cleanup


labkey-jeckels:
@labkey-stuartm @labkey-jony FYI. If a server was configured to use Universal GA, we'll start ignoring it after this change. Startup properties should only reference the GA4 or generic script properties going forward.

labkey-susanh (approved):

PR #4527, opened 2023-06-21
23.3 Fix cache miss logic for RemapCache and reduce initial bulk load size

labkey-susanh:
#### Rationale
The `RemapCache` used during sample imports with derivations is initialized with a query limited to 1,000,000 rows. When a sample type (or data class, or ...) has more rows than that, our cache miss logic was not finding some rows because the empty object returned is an empty list not a null value. Here we reduce the size of that initial query by an order of magnitude and fix the cache miss logic.

#### Changes
* Reduce table selection row limit for initial population of the cache
* Check for empty object as well as null value

### DIFF
--- a/api/src/org/labkey/api/dataiterator/SimpleTranslator.java
+++ b/api/src/org/labkey/api/dataiterator/SimpleTranslator.java
@@ -417,7 +417,7 @@ public class SimpleTranslator extends AbstractDataIterator implements DataIterat
 
         private TableSelector createSelector(ColumnInfo pkCol, ColumnInfo altKeyCol, SimpleFilter filter)
         {
-            return new TableSelector(_targetTable, Arrays.asList(altKeyCol, pkCol), filter, null).setMaxRows(10_000);
+            return new TableSelector(_targetTable, Arrays.asList(altKeyCol, pkCol), filter, null).setMaxRows(100_000);
         }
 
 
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -334,7 +334,7 @@ private Object fetch(Triple<ColumnInfo, ColumnInfo, MultiValuedMap<?,?>> triple,
                     bulkLoaded = map.get(k);
                 }
 
-                if (bulkLoaded == null)
+                if (bulkLoaded == null || bulkLoaded.isEmpty())
```
labkey-jeckels, line 338: This confused me until I found the comment on line 348 describing the behavior when `k` isn't in the map. Maybe a similar comment here would be helpful too

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -417,7 +417,7 @@ private TableSelector createSelector(ColumnInfo pkCol, ColumnInfo altKeyCol, Obj
 
         private TableSelector createSelector(ColumnInfo pkCol, ColumnInfo altKeyCol, SimpleFilter filter)
         {
-            return new TableSelector(_targetTable, Arrays.asList(altKeyCol, pkCol), filter, null).setMaxRows(1_000_000);
+            return new TableSelector(_targetTable, Arrays.asList(altKeyCol, pkCol), filter, null).setMaxRows(100_000);
```
labkey-jeckels, line 423:
```suggestion
            // Load a bunch of rows in the hopes of not needing to fetch for every value we encounter, but if we have a miss, we'll do per-value fetches as needed
            return new TableSelector(_targetTable, Arrays.asList(altKeyCol, pkCol), filter, null).setMaxRows(100_000);
```

labkey-jeckels (approved):

XingY (approved):

PR #4533, opened 2023-06-23
48142: Created and Created By fields are overwritten when existing data is updated via file upload

labkey-klum:
#### Rationale
This [PR](https://github.com/LabKey/platform/pull/3949) altered the handling of `created` and `createdBy` fields for bulk insert/update. It appears that the logic may be inverted and may have been intended to be : 
```
if (context.getInsertOption() == QueryUpdateService.InsertOption.UPDATE)
```
But it seems like it would be better if those columns were always in the don't update list, which was the previous behavior unless that was causing issues.

[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48142)


XingY (approved):

XingY (commented):
The ticket mentioned that list is having a similar issue. Does ListQueryUpdateService need similar treatment?

XingY (commented):
See https://github.com/LabKey/platform/pull/4534 for alternative fix

PR #4535, opened 2023-06-25
Issue 48056: More list indexing fixes, Issue 47886: Poor performance truncating large lists

labkey-bpatel:
#### Rationale
Issue [48056](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48056): More list indexing fixes
Issue [47886](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47886): Poor performance truncating large lists 

#### Changes
* Detect changes to document title, template, and sub-settings like metadata vs. data.
* Combine indexItems() and indexAttachments(); and use post-commit.
* Issue 47886 related change - Skip discussions (if disabled), attachments (if none), and indexing options (if turned off) when truncating Lists


list/src/org/labkey/list/model/ListManager.java
```diff
@@ -540,30 +647,40 @@ private String getDocumentId(ListDefinition list, @Nullable String entityId)
...
 
     // Index all modified items in this list
-    private void indexModifiedItems(@NotNull final IndexTask task, final ListDefinition list, final boolean reindex)
+    private void indexModifiedItems(@NotNull final IndexTask task, final ListDefinition list, final boolean reindex, boolean designChange)
     {
-        if (!list.getEachItemIndex())
-        {
-            return;
-        }
+        TableInfo listTable = list.getTable(User.getSearchUser());
```
labkey-bpatel: I've seen User.getSearchUser() used elsewhere so using it here - unless I am missing something, it does not seem right since the search user might just not have the write permission.

list/src/org/labkey/list/model/ListManager.java
```diff
@@ -826,8 +890,12 @@ public void setLastIndexed(long ms, long modified)
...
     {
-        deleteIndexedEntireListDoc(list);
-        deleteIndexedItems(list);
+        if (list.getEntireListIndex())
+            deleteIndexedEntireListDoc(list);
+
+        if (list.getEachItemIndex())
+            deleteIndexedItems(list);
+
         deleteIndexedAttachments(list);
```
labkey-adam: Shouldn't this call be conditional on `list.getFileAttachmentIndex()`, similar to the others?

list/src/org/labkey/list/model/ListManager.java
```diff
@@ -843,20 +911,25 @@ private void deleteIndexedAttachments(@NotNull ListDefinition list)
...
+                {
+                    as.deleteAttachmentIndexes(parentIds);
+                    parentIds.clear();
+                }
+            });
+
+            // Finish last batch
+            as.deleteAttachmentIndexes(parentIds);
+        }
     }
```
labkey-adam, line 895: FYI: I'm planning to push a change that significantly simplifies this batch handling. It's old code, but worth fixing.

list/src/org/labkey/list/model/ListManager.java
```diff
@@ -816,7 +878,9 @@ public void exec(Results results) throws StopIteratingException
             @Override
             public void setLastIndexed(long ms, long modified)
             {
-                ListManager.get().setLastIndexed(list, ms);
+                list.getTable(User.getSearchUser()).getSchema().getScope().addCommitTask(() -> {
+                    ListManager.get().setLastIndexed(list, ms);
+                }, DbScope.CommitTaskOption.POSTCOMMIT);
```
labkey-adam: This isn't needed or useful. `setLastIndexed()` is called by the indexer thread, so this will never be in a transaction.

list/src/org/labkey/list/model/ListManager.java
labkey-adam: We just need the scope, and we know it's the LabKey scope. Could just call `DbScope.getLabKeyScope()` and call it good.

labkey-adam (changes requested):
See comments

labkey-adam:
See BVT failures as well

list/src/org/labkey/list/model/ListManager.java
labkey-bpatel: Sounds good, thanks!

list/src/org/labkey/list/model/ListManager.java
labkey-bpatel: Yup, i missed that! Thank you!

list/src/org/labkey/list/model/ListManager.java
labkey-bpatel: This was one of my initial additions, I had meant to remove it, thanks for catching it

list/src/org/labkey/list/model/ListManager.java
```diff
@@ -337,16 +336,51 @@ ListDomainKindProperties update(User user, Container c, final ListDomainKindProp
...
         scope.addCommitTask(() -> {
             ListDefinition list = ListDefinitionImpl.of(listDef);
 
             // Turning on each-item indexing or attachment indexing -> clear last indexed column
-            if ((!oldEachItemIndex && newEachItemIndex) || (!oldFileAttachmentIndex && newFileAttachmentIndex))
+            // Also, document title, template, and sub-settings
+            if ((!oldEachItemIndex && newEachItemIndex) ||
+                    (!oldFileAttachmentIndex && newFileAttachmentIndex) ||
+                    (newEachItemTitleTemplate != null && !newEachItemTitleTemplate.equals(oldEachItemTitleTemplate)) ||
+                    (newEachItemBodyTemplate != null && !newEachItemBodyTemplate.equals(oldEachItemBodyTemplate)) ||
```
labkey-adam: I don't think these two checks are correct... switching from custom template to standard template will not trigger a clear, because new template is null. I think you want `Objects.equals()` in both cases.

list/src/org/labkey/list/model/ListManager.java
```diff
@@ -357,7 +391,12 @@ private void handleIndexSettingChanges(DbScope scope, ListDef listDef, boolean o
                 deleteIndexedAttachments(list);
 
             // Turning on entire-list indexing -> clear that list's last indexed column
-            if (!oldEntireListIndex && newEntireListIndex)
+            // Also, changing document title, template, and sub-settings like metadata vs. data
+            if ((!oldEntireListIndex && newEntireListIndex) ||
+                    (newEntireListTitleTemplate != null && !newEntireListTitleTemplate.equals(oldEntireListTitleTemplate)) ||
+                    (newEntireListBodyTemplate != null && !newEntireListBodyTemplate.equals(oldEntireListBodyTemplate)) ||
```
labkey-adam: Same with these two checks: use `Objects.equals()`

labkey-adam (changes requested):

labkey-adam:
@labkey-bpatel This PR is close. One more requested change (use `Objects.equals()`). Please also review my recent commit for sanity. I'm out for a while today, but will be able to review again this afternoon.

list/src/org/labkey/list/model/ListManager.java
labkey-bpatel: good point! done!

labkey-bpatel:
> @labkey-bpatel This PR is close. One more requested change (use `Objects.equals()`). Please also review my recent commit for sanity. I'm out for a while today, but will be able to review again this afternoon.

Requested change is in. Thanks for those updates, they look good, and much cleaner.

labkey-adam (approved):
I think this PR is ready to go, but please review my attachment indexing changes and the test results before merging.

PR #4539, opened 2023-06-27
Issue 48161: Query tracker uses lots of RAM on server w/ big SQL

labkey-jeckels:
#### Rationale
We don't want to spend hundreds of MB of memory remembering long SQL queries

#### Changes
* Limit the number of queries where we remember up to 1 MB of SQL
* Truncate older long queries to a mere 100KB
* Use DOM renderer in Java code
* Easy link to copy execution plan to clipboard
* Use a longer hash instead of Java's `hashCode()` for tracking queries

### DIFF
--- a/api/src/org/labkey/api/data/queryprofiler/Query.java
+++ b/api/src/org/labkey/api/data/queryprofiler/Query.java
@@ -38,6 +38,7 @@ public class Query
     private final @Nullable
     DbScope _scope;
     private final String _sql;
+    private boolean _truncated;
     private final @Nullable
     List<Object> _parameters;
     private final long _elapsed;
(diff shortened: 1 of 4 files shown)


labkey-adam (commented):
Shouldn't `QueryTracker.canShowExecutionPlan()` include a check for `_truncated`?

labkey-jeckels:
> Shouldn't `QueryTracker.canShowExecutionPlan()` include a check for `_truncated`?

I believe that if we've been truncated _validSql should be false, but it was easy to add an explicit check.

labkey-adam:
> > Shouldn't `QueryTracker.canShowExecutionPlan()` include a check for `_truncated`?
> 
> I believe that if we've been truncated _validSql should be false, but it was easy to add an explicit check.

Okay, I see that now, but doesn't hurt

labkey-adam (approved):

PR #4549, opened 2023-06-30
More list indexing improvements

labkey-adam:
#### Rationale
* [Issue 48182: Custom titles on list search results aren't working](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48182)
* Add debug logging to help with investigating [Issue 45624: Search indexer continually visiting subjects, datasets, lists, and navigation](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45624)
* Move a couple operations to post-commit tasks

### DIFF
--- a/list/src/org/labkey/list/model/ListImporter.java
+++ b/list/src/org/labkey/list/model/ListImporter.java
@@ -399,7 +399,7 @@ public class ListImporter
         }
     }
 
-    private boolean createNewList(Container c, User user, String listName, Collection<Integer> preferredListIds, TableType listXml, @Nullable ListsDocument.Lists.List listSettingsXml, Collection<ValidatorImporter> validatorImporters, List<String> errors) throws Exception
+    private boolean createNewList(Container c, User user, String listName, Collection<Integer> preferredListIds, TableType listXml, @Nullable ListsDocument.Lists.List listSettingsXml, Collection<ValidatorImporter> validatorImporters, List<String> errors, Logger log) throws Exception
     {
         final String keyName = listXml.getPkColumnName();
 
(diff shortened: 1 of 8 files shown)


list/src/org/labkey/list/model/ListImporter.java
```diff
@@ -429,21 +429,27 @@ private boolean createNewList(Container c, User user, String listName, Collectio
...
             list.setEntireListTitleTemplate(listSettingsXml.getEntireListTitleTemplate());
             list.setEntireListBodySetting(ListDefinition.BodySetting.getForValue(listSettingsXml.getEntireListBodySetting()));
             list.setEntireListBodyTemplate(listSettingsXml.getEntireListBodyTemplate());
 
             list.setFileAttachmentIndex(listSettingsXml.getFileAttachmentIndex());
             if (listSettingsXml.getCategory() != null)
                 list.setCategory(ListDefinition.Category.valueOf(listSettingsXml.getCategory()));
+
+            // These have been unused for years. Code remnants removed for 23.7, Issue 48182.
+            // TODO: Remove these XSD elements and warnings in 28.7 or before.
```
labkey-jeckels, line 447: 28.7? Do we have a five-year backwards compatibility commitment?

labkey-jeckels (approved):

list/src/org/labkey/list/model/ListImporter.java
labkey-adam: We've committed to nothing publicly. In an email thread last year, we discussed keeping deprecated elements around for somewhere between two and five years. I'd be fine with warning for two years and then failing with a clear error message after that. Especially for settings like these that are rarely (if ever) used.

PR #4553, opened 2023-07-03
GA4 app support

labkey-alan:
#### Rationale
Google Analytics 4 doesn't send particularly useful information for our apps when using the default configuration. This PR updates the injected GA4 script to not send page views for our Apps, so they can manually send page view events in a way that better captures page views.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/123
- https://github.com/LabKey/platform/pull/4553
- https://github.com/LabKey/biologics/pull/2212
- https://github.com/LabKey/inventory/pull/909
- https://github.com/LabKey/sampleManagement/pull/1930
- https://github.com/LabKey/labbook/pull/510

#### Changes
* AnalyticsServiceImpl::getTrackingScript - Do not send page views for app actions

### DIFF
--- a/core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
+++ b/core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
@@ -235,7 +235,7 @@ public class AnalyticsServiceImpl implements AnalyticsService
         String controller = url.getController();
         String action = url.getAction();
         Boolean sendPageView = true;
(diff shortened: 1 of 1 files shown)


labkey-susanh (approved):

labkey-jeckels (commented):
I'm fine with merging this for 23.7, but we need a better way to identify which actions want to opt out of page level data.

We should probably create a new annotation for Java-based actions and an XML-based approach for non-Java actions. Assuming that you don't want to tackle this in this PR, can you open a bug on this, which can be done in develop to remove the hard-coded list of controllers and actions?

PR #4557, opened 2023-07-04
Fix attachment indexing and clearing

labkey-adam:
#### Rationale
Back-port indexing fixes from develop

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4552

### DIFF
--- a/core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
+++ b/core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
@@ -35,7 +35,6 @@ import org.labkey.api.attachments.SpringAttachmentFile;
 import org.labkey.api.audit.AuditLogService;
 import org.labkey.api.audit.provider.FileSystemAuditProvider;
 import org.labkey.api.collections.CaseInsensitiveHashSet;
-import org.labkey.api.collections.CsvSet;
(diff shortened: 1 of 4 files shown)


core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
```diff
@@ -389,38 +389,38 @@ public void deleteAttachmentIndexes(List<String> parentIds)
...
     {
-        if (atts.isEmpty())
-            return;
-
         SearchService ss = SearchService.get();
         if (ss != null)
-            for (Attachment att : atts)
-                ss.deleteResource(makeDocId(parent, att.getName()));
+            ss.deleteResource(makeDocId(parent, name));
+        new SqlExecutor(CoreSchema.getInstance().getSchema()).execute(new SQLFragment(
```
labkey-susanh, line 402: Is this update meant to happen even if there is no search service?

labkey-susanh (approved):

core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
labkey-adam: Yes. If the SS is coming and going then this is the correct thing to do.

PR #4558, opened 2023-07-04
QueryWebPart support for async loading of total rows count in grid pagination header

cnathe:
#### Rationale
We've change the app grid implementation so that it asynchronously loads the total row count for the grid. This allows for the grid data to be rendered while the total count is loading (which is often the slowness when perf issues come up with large data tables). This PR implements a similar approach for the LKS DataRegion async QueryWebPart. We've added a new showPaginationCountAsync prop which will then make a separate query to get the grid total row count rather then including it / waiting for it as part of the grid data query.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/4558
- https://github.com/LabKey/hjfSampleRequests/pull/133
- https://github.com/LabKey/testAutomation/pull/1557

#### Changes
* implementation of async totalRowCount loading with async QWP via showPaginationCountAsync prop
* update pagination menu items after totalRows value is loaded
* don't re-query for totalRows on page change, show all, sort change
* cancel any previous totalRows query when new one is started


api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -3638,6 +3691,55 @@ if (!LABKEY.DataRegions) {
...
+
+        totalRowCountRequest = LABKEY.Query.selectRows({
+            ...region.getQueryConfig(),
+            method: 'POST',
+            containerPath: region.containerPath,
+            filterArray: LABKEY.Filter.getFiltersFromParameters({ ...params, ...jsonData.filters }, params.dataRegionName),
+            sort: undefined,
+            maxRows: 1,
+            offset: 0,
+            includeDetailsColumn: false,
```
labkey-nicka, line 3720: Possibly consider `includeMetadata: false` as well.

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -3638,6 +3691,55 @@ if (!LABKEY.DataRegions) {
...
+    var _loadAsyncTotalRowCount = function(region, params, jsonData) {
+        // if there is a previous request pending, abort it before starting a new one
+        if (totalRowCountRequest !== undefined) {
+            totalRowCountRequest.abort();
+        }
+
+        region.totalRows = undefined;
+        region.loadingTotalRows = true;
+
+        totalRowCountRequest = LABKEY.Query.selectRows({
```
labkey-nicka, line 3711: Clean up `totalRowCountRequest` (i.e. set to `undefined`) in `success` and `failure` cases.

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -3638,6 +3691,55 @@ if (!LABKEY.DataRegions) {
...
+            sort: undefined,
+            maxRows: 1,
+            offset: 0,
+            includeDetailsColumn: false,
+            includeUpdateColumn: false,
+            includeTotalCount: true,
+            success: function(json) {
+                if (json !== undefined && json.rowCount !== undefined) {
+                    region.totalRows = json.rowCount;
+                    region.loadingTotalRows = false;
```
labkey-nicka: Shouldn't this always be set to false regardless of if this `if` condition is true?

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -3638,6 +3691,55 @@ if (!LABKEY.DataRegions) {
...
+            _loadAsyncTotalRowCount(region, params, jsonData);
+        }
+        region.skipTotalRowCount = false;
+    };
+
+    var totalRowCountRequest;
+    var _loadAsyncTotalRowCount = function(region, params, jsonData) {
+        // if there is a previous request pending, abort it before starting a new one
+        if (totalRowCountRequest !== undefined) {
+            totalRowCountRequest.abort();
```
labkey-nicka, line 3705: I would expect an abort here to cause a call to the `failure` callback handler below (from the previous request). Is that not happening? Seems like we need to recognize this case. Here is an [example of something I did](https://github.com/LabKey/labkey-ui-premium/blob/b5a3d824/src/eln/ReferenceSuggester.tsx#L290) that was similar recently.

labkey-nicka (approved):

api/webapp/clientapi/dom/DataRegion.js
cnathe: done

api/webapp/clientapi/dom/DataRegion.js
cnathe: Good idea. I know what didn't set that for the app grid case, but that was because of the parsing code we had for selectRows. I've added it here and the request response went from 22.9kb to 1kb!

api/webapp/clientapi/dom/DataRegion.js
cnathe: yes, moved out

api/webapp/clientapi/dom/DataRegion.js
cnathe: good call. I've updated the failure handler to be:
```
            failure: function(error, request) {
                var aborted = request.status === 0;
                if (!aborted) {
                    console.error(error);
                    totalRowCountRequest = undefined;
                    region.loadingTotalRows = false;
                    _getFormSelector(region).find('a.labkey-paginationText').html(_getPaginationText(region));
                }
            }
```

PR #4559, opened 2023-07-05
Hide names in notebooks and timeline entries when users don't have permission to see them

labkey-susanh:
#### Rationale
When notebooks contain references to entities a user does not have permission to see, we want to hide these notebooks altogether from these users. The initial check here will check permissions on the product projects containers in general so most people won't need to go further than that.

#### Related Pull Requests

- https://github.com/LabKey/labbook/pull/511
- https://github.com/LabKey/labkey-ui-components/pull/1227
- https://github.com/LabKey/labkey-ui-premium/pull/129
- https://github.com/LabKey/sampleManagement/pull/1939
- https://github.com/LabKey/biologics/pull/2222

#### Changes
* add `getProductProjects` method

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -93,6 +93,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.BooleanSupplier;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
 
 /**
@@ -1694,7 +1695,7 @@ public class Container implements Serializable, Comparable<Container>, Securable
     }
 
     /**
-     * Returns the subfolder count of the project container, if product projects feature is enabled in project
+     * Returns true if produt projects feature is enabled and the subfolder count of the project container is > 0
      */
     public boolean hasProductProjects()
     {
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/Container.java
```diff
@@ -1710,6 +1711,18 @@ public boolean hasProductProjects()
         return project.getChildren().stream().anyMatch(c -> c.getContainerType().isInFolderNav());
     }
 
+    public List<Container> getProductProjects()
+    {
+        if (!isProductProjectsEnabled())
+            return Collections.singletonList(this);
```
labkey-nicka: Interesting. What does this mean? Why not return `Collections.emptyList()` here?

labkey-nicka (approved):

api/src/org/labkey/api/data/Container.java
```diff
@@ -1694,7 +1695,7 @@ public boolean isAppHomeFolder()
     }
 
     /**
-     * Returns the subfolder count of the project container, if product projects feature is enabled in project
+     * Returns true if produt projects feature is enabled and the subfolder count of the project container is > 0
```
labkey-nicka: nit: typo "produt"

api/src/org/labkey/api/data/Container.java
labkey-susanh: Ah yes, I'm wanting to get all containers here, not just the productProjects containers, but I'll adjust to align with `hasProductProjects`.

PR #4563, opened 2023-07-07
null check

labkey-matthewb:
#### Rationale
Fix NPE
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48001

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels:
Based on the issue, I think we want a fix for 23.7.

I'd also suggest that we make WrappedColumnInfo.wrap() throw an exception if you pass in a null, which should make it easier to track down future problems.

PR #4568, opened 2023-07-10
Issue 48174: Very slow GET downloads from remote NFS file root

labkey-jeckels:
#### Rationale
One of our cloud instances maps an on-premise file share for its storage. GET downloads from it are excruciatingly slow.

#### Changes
* Switch to Channel-based I/O instead of streams, and use memory mapped files when possible


labkey-matthewb (approved):
Can't say I'm particularly familiar with NIO, but this looks fine.  I don't know that we need to jump from 16K to 8M to get the performance scale up (trade off being how long before data starts showing up a the client).

The next (somewhat extreme step) would be to get some asynchronous handling of the read and write, at the expense of complexity.

labkey-jeckels:
@labkey-matthewb I chose 8MB since that's also the default in sun.nio.ch.FileChannelImpl that other NIO methods end up using. This pointer is for JDK 8, but it's the same in JDK 17

https://github.com/frohoff/jdk8u-jdk/blob/master/src/share/classes/sun/nio/ch/FileChannelImpl.java#L497

labkey-jeckels:
> Can't say I'm particularly familiar with NIO, but this looks fine. I don't know that we need to jump from 16K to 8M to get the performance scale up (trade off being how long before data starts showing up a the client).
> 
> The next (somewhat extreme step) would be to get some asynchronous handling of the read and write, at the expense of complexity.

I should have paid more attention to your comment on buffer size. I wasn't considering the case of many rapid-fire small file downloads.

PR #4574, opened 2023-07-12
Plate schema refactor

labkey-klum:
#### Rationale
This represents an initial phase of AB Discovery work to support some of the initial scenarios regarding plate creation as well as the plating of samples into wells. Changes made in this FB include:
- Exposing the tables in the plate schema to `Query`. Note that a plate schema with tables had already existed, but the virtual tables were broken and there were no usages of that schema/table in the server so everything was refactored. The result is a new plate schema with plate, wellGroup, and well tables.
- Implement a `QueryUpdateService` for each of the exposed tables.
- Refactor the `PlateManager` to go through the update service when inserting, updating, or deleting (when possible) rows in those tables.

#### Related Pull Requests
https://github.com/LabKey/commonAssays/pull/623


assay/src/org/labkey/assay/plate/query/PlateTable.java
```diff
@@ -16,75 +16,231 @@
...
+                {
+                    int runsInUse = PlateManager.get().getRunCountUsingPlateTemplate(container, plate);
+                    if (runsInUse > 0)
+                        throw new QueryUpdateServiceException(String.format("Plate template is used by %d runs and cannot be deleted", runsInUse));
+
+                    PlateManager.get().beforePlateDelete(container, plateId);
+                    Map<String, Object> returnMap = super.deleteRow(user, container, oldRowMap);
+
+                    transaction.addCommitTask(() -> PlateManager.get().clearCache(), DbScope.CommitTaskOption.POSTCOMMIT);
+                    transaction.commit();
```
labkey-nicka, line 238: Should this commit be done outside of this if statement?

labkey-nicka (approved):

assay/src/org/labkey/assay/plate/query/PlateTable.java
labkey-klum: yes, I've narrowed the scope of the try block

PR #4577, opened 2023-07-12
Ratchet down reindexing of lists and datasets

labkey-adam:
#### Rationale
For a variety of reasons, lists and datasets are being reindexed unnecessarily. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45624

#### Changes
* Lists:
  * Fix overly aggressive definition of `reindex`
  * Clear `ListDef` cache entry after setting `LastIndexed` to avoid checks on stale values
  * When turning on attachment indexing, clear `core.Documents.LastIndexed` rows associated with that list, not the list's own `LastIndexed` column
  * Conditionalize the handling of subsettings on the primary indexing settings (e.g., don't reindex if a template changes but the corresponding indexing setting is off)
  * Remove unused `designChange` boolean parameter
  * Rework list indexing settings transition checks for better readability (IMO)
* Datasets: Filter indexed datasets on `Modified` > `modifiedSince`.
* Participants: Simplify batching of participants. Fix completely broken alias filter in participant query; move to `SimpleFilter` instead of manually building `WHERE` clause via `SQLFragment`.
* General: Ensure that `setLastIndexed()` is always called with a value that's greater than or equal to Modified, to avoid pointless reindexing

### DIFF
--- a/list/src/org/labkey/list/model/ListManager.java
+++ b/list/src/org/labkey/list/model/ListManager.java
@@ -373,13 +373,14 @@ public class ListManager implements SearchService.DocumentProvider
         scope.addCommitTask(() -> {
             ListDefinition list = ListDefinitionImpl.of(listDef);
 
-            // Turning on each-item indexing or attachment indexing, or changing document title template, body template,
-            // or body setting means reindexing all items, so clear last indexed column
-            if ((!oldEachItemIndex && newEachItemIndex) ||
-                    (!oldFileAttachmentIndex && newFileAttachmentIndex) ||
-                    (!Objects.equals(newEachItemTitleTemplate, oldEachItemTitleTemplate)) ||
-                    (!Objects.equals(newEachItemBodyTemplate, oldEachItemBodyTemplate)) ||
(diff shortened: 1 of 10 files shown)


labkey-matthewb (approved):

list/src/org/labkey/list/model/ListManager.java
```diff
@@ -373,45 +373,77 @@ private void handleIndexSettingChanges(DbScope scope, ListDef listDef, ListIndex
...
-                    .append(" SET LastIndexed = NULL WHERE ListId = ? AND LastIndexed IS NOT NULL")
-                    .add(list.getListId());
+                // Turning off each-item indexing -> clear item docs from the index
+                if (oldEachItemIndex)
+                    deleteIndexedItems(list);
+            }
 
-                new SqlExecutor(scope).execute(sql);
+            // Is attachment indexing turned on?
+            if (newFileAttachmentIndex)
```
labkey-matthewb, line 400: much easier to read.

PR #4580, opened 2023-07-13
Fix ext4 storeid for lookups

ankurjuneja:
#### Rationale
The change in related PR to include viewName in the lookup storeId caused the store Id to end in "||" as the lookups may not have a viewName defined in metadata that caused some lookups to not resolve to the displayColumn.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4344

#### Changes
* check for existence of viewName in lookup definition


api/webapp/clientapi/ext4/Util.js
```diff
@@ -865,13 +865,14 @@
             }
 
             if (c.lookup) {
-                return c.lookup.storeId || [
-                    c.lookup.schemaName || c.lookup.schema,
+                let storeId = c.lookup.storeId || [c.lookup.schemaName || c.lookup.schema,
```
labkey-jeckels: If `c.lookup.storeId` is defined, we'll use it. Presumably it's a string, and calling `join()` on it will fail.

labkey-jeckels (changes requested):

labkey-jeckels:
Do we need to fix this in 23.7?

api/webapp/clientapi/ext4/Util.js
ankurjuneja: yes, correct. Fixed it

labkey-klum (approved):

PR #4584, opened 2023-07-14
Issue 48156: Perf issues with file watchers on EFS deployments

labkey-jeckels:
#### Rationale
We're seeing excessive file metadata I/O on some deployments. Throttle back a bit on the polling file system listener and add some logging to gain better insights.

#### Changes
* Log when we finish a poll or hit an exception
* Drop back to 2 threads doing the polling, every 30 seconds instead of 10 seconds

### DIFF
--- a/api/src/org/labkey/api/files/FileSystemWatcherImpl.java
+++ b/api/src/org/labkey/api/files/FileSystemWatcherImpl.java
@@ -84,7 +84,7 @@ public class FileSystemWatcherImpl implements FileSystemWatcher
     private final ConcurrentMap<Path, PathListenerManager> _listenerMap = new ConcurrentHashMap<>(1000);
     private final PathWatchService _pollingWatcher;
 
-    private long _lastPollingExceptionTime = 0;
+    private volatile long _lastPollingExceptionTime = 0;
 
     FileSystemWatcherImpl() throws IOException
     {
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/files/FileSystemWatcherImpl.java
```diff
@@ -78,12 +78,14 @@
...
 {
     private static final Logger LOG = LogHelper.getLogger(FileSystemWatcherImpl.class, "Open file system handlers and listeners");
-    private static final long POLLING_PERIOD_SECONDS = 10L;
+    private static final long POLLING_PERIOD_SECONDS = 30L;
 
     private final WatchService _watcher;
     private final ConcurrentMap<Path, PathListenerManager> _listenerMap = new ConcurrentHashMap<>(1000);
     private final PathWatchService _pollingWatcher;
 
+    private long _lastPollingExceptionTime = 0;
```
labkey-adam: `volatile`?

labkey-adam (approved):

api/src/org/labkey/api/files/FileSystemWatcherImpl.java
labkey-jeckels: Sure!

PR #4585, opened 2023-07-14
DataRegion fix for tracking totalRowCountRequest per dataRegion for cancel/abort

cnathe:
#### Rationale
This was found during testing for a client app that uses QueryWebPart with the async total row count loading enabled. In the case where multiple QWPs are on the page / in scope, we need to track the totalRowCountRequest for each dataRegion.name separately so that we don't cancel/abort the request for one dataRegion when another one creates the request.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4558

#### Changes
* update DataRegion.js to track per dataRegion.name requests via totalRowCountRequests object


RosalineP (approved):
Manually tested and reviewed, looks good!

PR #4594, opened 2023-07-18
Issue 48278: Short fails to convert to boolean in SQLServer ETL

labkey-jeckels:
#### Rationale
An ETL is failing to coerce a `SHORTINT` value to a `BIT` in a SQLServer-based ETL

#### Changes
* Allow any numeric type that's equal to 0 or 1 for conversion
* Add unit test coverage for recent fixes to `ArrayListMap.toString()`

### DIFF
--- a/api/src/org/labkey/api/collections/ArrayListMap.java
+++ b/api/src/org/labkey/api/collections/ArrayListMap.java
@@ -503,24 +503,26 @@ public class ArrayListMap<K, V> extends AbstractMap<K, V> implements Iterable<V>
             assertEquals(a.get("Z"), "one");
             assertEquals(b.get("Z"), "ONE");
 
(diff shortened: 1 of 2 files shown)


labkey-matthewb (approved):

labkey-jeckels:
@labkey-matthewb can you also take a look at the unit test updates I just pushed?

My change made the `containsKey()` implementation more consistent with standard Map behavior. Previously, we grew the map's ArrayList with nulls. We now expand it with `DOES_NOT_CONTAINKEY` instead. We still share the findMap across rows. I'll also run ETL tests, but let me know if you anticipate any problems

labkey-matthewb:
containsKey() is usually not too interesting for resultset like usages, but it definitely better to not be gratuitously inconsistent!

PR #4597, opened 2023-07-18
Add support for ancestor lookups for data classes

labkey-susanh:
#### Rationale
We are finding there is a demand for supporting source lineage in order to model real-world data. This PR adds support for using Ancestor lookups in query grids in the same way we support it for samples. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1236
* https://github.com/LabKey/labkey-ui-premium/pull/138
* https://github.com/LabKey/sampleManagement/pull/1970
* https://github.com/LabKey/biologics/pull/2240

#### Changes
* Add metric to count the number of data class objects that have data class parents
* Add Ancestor lookup column to DataClass tables and update `ClosureQueryHelper` to also compute closure for data class objects.


experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
```diff
@@ -232,16 +231,16 @@ public ColumnInfo createLookupColumn(ColumnInfo foreignKey, String displayField)
     public static SQLFragment getValueSql(UserSchema userSchema, TableType type, String sourceLSID, SQLFragment objectId, ExpObject target)
     {
         if (target instanceof ExpSampleType st)
-            return getValueSql(userSchema, type, sourceLSID, objectId, "m" + st.getRowId());
+            return getValueSql(userSchema, type, sourceLSID, objectId, "m" + st.getRowId(), true);
```
labkey-matthewb: Doesn't the last boolean parameter here depend on sourceType (TableType type) rather than targetType?

If so, maybe just pass sourceType all the way through instead of converting to a boolean?

labkey-matthewb (changes requested):

experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -624,6 +624,8 @@ public void containerDeleted(Container c, User user)
 
                 results.put("dataClassCount", new SqlSelector(ExperimentService.get().getSchema(), "SELECT COUNT(*) FROM exp.dataclass").getObject(Long.class));
                 results.put("dataClassRowCount", new SqlSelector(ExperimentService.get().getSchema(), "SELECT COUNT(*) FROM exp.data WHERE classid IN (SELECT rowid FROM exp.dataclass)").getObject(Long.class));
+                results.put("dataWithDataParentsCount", new SqlSelector(ExperimentService.get().getSchema(), "SELECT COUNT(DISTINCT d.sourceApplicationId) FROM exp.data d\n" +
```
XingY, line 627: Would this include Biologics registry related entities? Is this what we want?

labkey-susanh: The related entities are generally not represented by experiment runs, so I think this will be fine. Question out to @labkey-hannah about the particulars here, though.

labkey-susanh: Hannah indicated that this count should be fine for our usage evaluation purposes.

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
labkey-susanh: I've made this change, though I'm still a bit baffled why I've seen no difference in behavior when using one or the other implementation. I must not be looking at the right combination of source and target tables...

labkey-matthewb (approved):

PR #4598, opened 2023-07-18
Fix user details button order & support identity maps

labkey-adam:
#### Rationale
Fix two unrelated issues:
* User details / My Account page: Move "Done" button to far right, per our best practices. Also, "Show Users" / "Show Project Users" button is effectively an alternative "Done" button (sometimes the only "Done" button), so move it to the right as well (just to the left of the "Done" button, if present). https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48276
* Adjust `Selector` methods `getValueMap()`, `fillValueMap()`, `getMultiValuedMap()`, and `fillMultiValuedMap()` to support single-column queries instead of throwing `IllegalStateException`. A single column result set implies an identify map (each key maps to a value that equals the key). https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=48054

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4292

### DIFF
--- a/api/src/org/labkey/api/data/BaseSelector.java
+++ b/api/src/org/labkey/api/data/BaseSelector.java
@@ -569,7 +569,7 @@ public abstract class BaseSelector<SELECTOR extends BaseSelector<?>> extends Jdb
         return factory;
     }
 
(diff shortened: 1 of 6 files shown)


labkey-jeckels (approved):
Note the unit test failure about the formerly expected exception.

PR #4599, opened 2023-07-19
Issue 48288: Crashes linked to download of many small JS files

labkey-jeckels:
#### Rationale
Many small rapid-fire downloads can put too much memory pressure on the VM after the recent changes to improve I/O for large files.

#### Changes
* Reduce buffer size, especially for non-memory-mapped I/O transfers


labkey-martyp (approved):

labkey-matthewb:
This also suggests that we're not using the org.apache.tomcat.sendfile.support code path.  That might also be worth investigating.  I'm not sure what environments are supposed to support that or if it requires a JNI lib or something like that.

PR #883, opened 2020-02-26
Issue 39805: uq_plate_container_name violation when upgrading

labkey-kevink:
- plate templates must have unique names, but plate instances may not
- use filtered unique index over assay.plate (container, name) for plate templates

PR #4604, opened 2023-07-19
Combine Plate/PlateTemplate and WellGroup/WellGroupTemplate interfaces

labkey-klum:
#### Rationale
Previously, assay plates had maintained the following hierarchy:
- `Plate\PlateTemplate` interfaces and classes
- `WellGroup\WellGroupTemplate` interfaces and classes

The intended distinction between the template versus non-template was that a template implied that no data had been associated with a plate's wells and the minute you added data to the plate it was no longer a template. However, well data associated with a plate was never stored to the database and while there was a `template` field in the `PlateTable` that was also never associated with the distinction. It wasn't too problematic since in existing implementations this never mattered. With `AbDisc` this now changes.

The abstraction also never worked and these classes tended to be used interchangeably with plenty of downcasting and implementation assumptions.

#### Related Pull Requests
https://github.com/LabKey/commonAssays/pull/625

#### Changes
- Combine the `Plate\PlateTemplate` interfaces so only `Plate` exists
- Combine the `PlateImpl\PlateTemplateImpl` classes so only `PlateImpl` exists
- Combine the `WellGroup\WellGroupTemplate` interfaces so only `WellGroup` exists
- Combine the `WellGroupImpl\WellGroupTemplateImpl` classes so only `WellGroupImpl` remains
- Clean up `PlateService` to remove most distinctions between `Plate` and `PlateTemplate`.
- Replace the plate cache in `PlateManager` with a `BlockingCache` implementation, the semantics are a bit different and the cache key is `Container` this is to facilitate resolving `Plate` instances by : `RowId, Lsid, Name`.
- Add an additional unit test to verify assumptions about templates.


assay/api-src/org/labkey/api/assay/plate/PlateBasedAssayProvider.java
```diff
@@ -37,8 +37,8 @@ public interface PlateBasedAssayProvider extends AssayProvider
 {
     String VIRUS_NAME_PROPERTY_NAME = "VirusName";
 
-    void setPlateTemplate(Container container, ExpProtocol protocol, PlateTemplate template);
-    PlateTemplate getPlateTemplate(Container container, ExpProtocol protocol);
+    void setPlateTemplate(Container container, ExpProtocol protocol, Plate template);
```
labkey-nicka: nit: Should we consider renaming `setPlateTemplate` and `getPlateTemplate` on the `PlateBasedAssayProvider` interface to `setPlate` and `getPlate`?

assay/api-src/org/labkey/api/assay/plate/AbstractPlateBasedAssayProvider.java
```diff
@@ -89,10 +89,10 @@ public void setPlateTemplate(Container container, ExpProtocol protocol, PlateTem
     }
 
     @Override
-    public PlateTemplate getPlateTemplate(Container container, ExpProtocol protocol)
+    public Plate getPlateTemplate(Container container, ExpProtocol protocol)
```
labkey-nicka: nit: Consider marking the interface and this override as `@Nullable`.

assay/api-src/org/labkey/api/assay/plate/PlateService.java
```diff
@@ -58,26 +59,50 @@ static PlateService get()
...
+    /**
+     * Creates a new plate
+     * @param container The template's container.
+     * @param templateType The type of plate, if associated with a particular assay.
+     * @param rowCount The number of columns in the plate.
+     * @param columnCount The number of rows in the plate.
+     * @return A newly created plate instance
+     * @throws IllegalArgumentException Thrown if a template of the specified name already exists in the container.
+     */
+    Plate createPlate(Container container, String templateType, int rowCount, int columnCount);
```
labkey-nicka: nit: Consider annotating as `@NotNull`. Same with the next interface `createPlateTemplate`.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -339,32 +284,42 @@ public int getRunCountUsingPlateTemplate(@NotNull Container c, @NotNull PlateTem
...
+
+    @Override
+    public @Nullable Plate getPlate(Container container, Lsid lsid)
+    {
+        return PlateCache.getPlate(container, lsid);
     }
 
-    public @Nullable PlateImpl getPlate(String lsid)
+    /**
+     * Note that this does not use the cache nor does it return a fully materialized plate.
```
labkey-nicka: Interesting. I understand it not using the cache since this is kind of a "one off" scenario where a container is not specified. That said, what would it take for it to be fully materialized? A call to `populatePlate()`? Possibly worth doing so we don't have this discrepancy? I'm guessing you analyzed this and determined it wasn't necessary but figured I'd call it out.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -1029,39 +986,9 @@ public void registerPlateTypeHandler(PlateTypeHandler handler)
...
-        PLATE_TEMPLATE_CACHE.put(getPlateTemplateCacheKey(template.getContainer(), template.getLSID()), template);
-    }
-
-    public void clearCache()
-    {
-        PLATE_TEMPLATE_CACHE.removeUsingFilter(new Cache.StringPrefixFilter(PlateTemplateImpl.class.getName()));
-    }
-
-    private PlateTemplateImpl getCachedPlateTemplate(Container container, int rowId)
+    public void clearCache(Container c)
```
labkey-nicka, line 1005: This is very satisfying to see.

assay/src/org/labkey/assay/plate/WellGroupImpl.java
```diff
@@ -41,10 +44,10 @@
 
 /**
  * User: brittp
-* Date: Oct 20, 2006
-* Time: 10:26:38 AM
-*/
-public class WellGroupImpl extends WellGroupTemplateImpl implements WellGroup
+ * Date: Oct 20, 2006
+ * Time: 4:32:11 PM
```
labkey-nicka: nit: Lol, updating time stamps from 2006 are we?

assay/src/org/labkey/assay/plate/query/PlateTable.java
```diff
@@ -199,33 +199,37 @@ public DataIteratorBuilder createImportDIB(User user, Container container, DataI
...
                 return Collections.emptyMap();
 
-            int runsInUse = PlateManager.get().getRunCountUsingPlateTemplate(container, plate);
+            int runsInUse = PlateManager.get().getRunCountUsingPlate(container, plate);
             if (runsInUse > 0)
                 throw new QueryUpdateServiceException(String.format("%s is used by %d runs and cannot be updated", plate.isTemplate() ? "Plate template" : "Plate", runsInUse));
 
-            return super.updateRow(user, container, row, oldRow);
+            Map<String, Object> newRow = super.updateRow(user, container, row, oldRow);
+            PlateManager.get().clearCache(container);
```
labkey-nicka, line 252: Given that we're updating a single plate is it possible to just uncache that plate? Same comment for `deleteRow`.

assay/src/org/labkey/assay/plate/model/WellGroupBean.java
```diff
@@ -0,0 +1,92 @@
+package org.labkey.assay.plate.model;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import org.labkey.assay.plate.WellGroupImpl;
+
+/**
+ * Used to serialize to the WellGroup table
```
labkey-nicka, line 7: Interesting. Where are we serializing this presently?

assay/src/org/labkey/assay/plate/query/PlateTable.java
labkey-nicka: Never mind, I see that we're caching plates by container as the lowest level of cache granularity.

assay/src/org/labkey/assay/plate/PlateCache.java
```diff
@@ -0,0 +1,106 @@
...
+            Map<Integer, Plate> rowIdMap = new HashMap<>();
+            Map<String, Plate> nameMap = new HashMap<>();
+            Map<Lsid, Plate> lsidMap = new HashMap<>();
+            List<Plate> templates = new ArrayList<>();
+
+            new TableSelector(AssayDbSchema.getInstance().getTableInfoPlate(), SimpleFilter.createContainerFilter(c), null).forEach(PlateImpl.class, plate -> {
+                PlateManager.get().populatePlate(plate);
+
+                rowIdMap.put(plate.getRowId(), plate);
+                nameMap.put(plate.getName(), plate);
```
labkey-nicka, line 49: I don't believe we enforce any uniqueness on plate names within a container (or within any scope for that matter). We have a couple options:

1. Consider adding this as a restriction. I could imagine this getting confusing if multiple plates have the same name.
2. If not, then should probably have `getForName` return a `Collection<Plate>`.
3. May the best call win. If you're the last plate to be queried with that name then you win the cache game. Lol.

labkey-nicka (commented):
This is great. Certainly a step in the right direction for clarity of modeling. The caching layer is much _much_ improved.

assay/src/org/labkey/assay/plate/PlateManager.java
labkey-nicka: Another reason I ask is if we're creating plate documents to search upon we'll likely want to key them by LSID. At that point we will want to materialize them. However, we could leave it up to the caller to call `populatePlate`.

assay/src/org/labkey/assay/plate/PlateImpl.java
```diff
@@ -68,83 +104,397 @@ else if (wellValues.length != template.getRows() && wellValues[0].length != temp
...
+
+    @Override
+    public int getColumns()
     {
-        return (List<? extends WellGroupImpl>)getWellGroupTemplates(type);
+        return _columns;
+    }
+
+    @Override
+    public String getName()
```
labkey-nicka, line 300: Hmm, for some reason `name` isn't round-tripping:
image

assay/api-src/org/labkey/api/assay/plate/PlateBasedAssayProvider.java
cnathe: Agreed

assay/api-src/org/labkey/api/assay/plate/PlateService.java
```diff
@@ -58,26 +59,50 @@ static PlateService get()
     }
 
     /**
-     * Instantiates a new plate instance based on the specified plate template and well data.
+     * Instantiates a new plate instance based on the specified plate and well data.
      * This plate is not persisted to the database.
-     * @param template The template that this instance is based upon.
+     * This plate is not persisted to the database.
```
cnathe: is this line of the comment a duplicate from the line above?

assay/api-src/org/labkey/api/assay/plate/PlateService.java
```diff
@@ -58,26 +59,50 @@ static PlateService get()
...
      */
-    @Nullable Plate createPlate(PlateTemplate template, double[][] wellValues, boolean[][] excludedWells, int runId, int plateNumber);
+    @Nullable Plate createPlate(Plate plate, double[][] wellValues, boolean[][] excludedWells, int runId, int plateNumber);
 
     /**
-     * Instantiates a new plate instance based on the specified plate template and well data.
+     * Instantiates a new plate instance based on the specified plate and well data.
      * This plate is not persisted to the database.
-     * @param template The template that this instance is based upon.
+     * This plate is not persisted to the database.
```
cnathe: same here re: duplicate line in comment

cnathe (commented):
Great refactor, this should hopefully be much clearer now. I'll defer the final CR approval for Nick after his comments are addressed.

assay/src/org/labkey/assay/plate/PlateManager.java
labkey-klum: Actually a valid point, I'll go ahead and populate the plate just so we don't have this weird case.

assay/src/org/labkey/assay/plate/WellGroupImpl.java
labkey-klum: 🤣

assay/src/org/labkey/assay/plate/model/WellGroupBean.java
labkey-klum: ah right, this is in `PlateManager.savePlateImpl` to limit the fields that the `ObjectFactory` attempts to introspect.

assay/api-src/org/labkey/api/assay/plate/PlateService.java
labkey-klum: lol, I guess we felt like we had to say it twice.

assay/src/org/labkey/assay/plate/PlateCache.java
labkey-klum: On very old databases there is the possibility that we have duplicate name/container combos. Kevin tried to add the constraint in 2020 and ran into problems, but settled on a unique filtered index on name/container for templates.
- https://github.com/LabKey/platform/pull/883

Prior to 2020 when you edited a copy of an existing template we would set the template bit to `false`, but as part of that work we implemented:
- java check to enforce name uniqueness
- copies are made as templates

While we permitted duplicate plate names to be created in the past, you wouldn't have been able to use them in an assay since the assays resolved them assuming a single plate for the name:
```
PlateTemplateImpl template = new TableSelector(AssayDbSchema.getInstance().getTableInfoPlate(), filter, null).getObject(PlateTemplateImpl.class);
```

I'm leaning towards enforcing container/name at the update service level and logging a warning in the `PlateCache` if we encounter plates with duplicate names.

labkey-nicka: I agree. At the QUS level is sufficient for now with a warning from `PlateCache`.

labkey-nicka (approved):

PR #4606, opened 2023-07-19
Only site admins should see encryption key warnings

labkey-adam:
#### Rationale
Encryption key warning needs a permission check. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48296


labkey-jeckels (approved):
Your call on whether to sneak this into 23.7.0 or wait. Seems low risk, but also not a must-have

labkey-adam:
> Your call on whether to sneak this into 23.7.0 or wait. Seems low risk, but also not a must-have

I will wait for the code freeze to be over, because I'm a rule follower. Unlike some people.

PR #4607, opened 2023-07-19
Add Project level date/time to App Settings page

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1238
- https://github.com/LabKey/platform/pull/4607
- https://github.com/LabKey/biologics/pull/2243
- https://github.com/LabKey/sampleManagement/pull/1975

#### Changes
* Create the api version of ProjectSettingsAction

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -10382,6 +10382,183 @@ public class AdminController extends SpringActionController
     }
 
     @RequiresPermission(AdminPermission.class)
+    public static class UpdateProjectSettingsAction extends MutatingApiAction<SimpleApiJsonForm>
+    {
+        @Override
+        public Object execute(SimpleApiJsonForm form, BindException errors)
+        {
+            ApiSimpleResponse response = new ApiSimpleResponse();
+            response.put("success", false);
+            JSONObject json = form.getJsonObject();
+            Container c = getContainer();
+            WriteableLookAndFeelProperties props = LookAndFeelProperties.getWriteableInstance(c);
+            boolean hasAdminOpsPerm = c.hasPermission(getUser(), AdminOperationsPermission.class);
+
+            if (json.has("themeName"))
+            {
+                String themName = json.optString("themeName");
+                try
+                {
+                    if (themName == null)
+                    {
+                        if (!c.isRoot())
+                            props.clearThemeName();
+                    }
+                    else
+                    {
+                        props.setThemeName(themName);
+                    }
+                }
+                catch (IllegalArgumentException ignored)
+                {
+                }
+            }
+
+            if (json.has("shouldInherit"))
+            {
+                boolean shouldInherit = json.optBoolean("shouldInherit");
+                if (shouldInherit != SecurityManager.shouldNewSubfoldersInheritPermissions(c))
+                {
+                    SecurityManager.setNewSubfoldersInheritPermissions(c, getUser(), shouldInherit);
+                }
+            }
+
+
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -10381,6 +10381,183 @@ protected TYPE getType()
         }
     }
 
+    @RequiresPermission(AdminPermission.class)
+    public static class UpdateProjectSettingsAction extends MutatingApiAction<SimpleApiJsonForm>
```
cnathe: could this use `ProjectSettingsForm` instead of just `SimpleApiJsonForm`?

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -10381,6 +10381,183 @@ protected TYPE getType()
...
     }
 
+    @RequiresPermission(AdminPermission.class)
+    public static class UpdateProjectSettingsAction extends MutatingApiAction<SimpleApiJsonForm>
+    {
+        @Override
+        public Object execute(SimpleApiJsonForm form, BindException errors)
+        {
+            ApiSimpleResponse response = new ApiSimpleResponse();
+            response.put("success", false);
```
cnathe: why is this set to false here? seems like setting it at the end of the execute method would be enough.

core/src/org/labkey/core/admin/AdminController.java
cnathe: ah, I suppose that might not work for being able to tell the difference between a property that is being cleared out vs one that just isn't being updated.

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -10381,6 +10381,183 @@ protected TYPE getType()
...
+                    {
+                        // this will throw an InvalidEmailException for invalid email addresses
+                        ValidEmail email = new ValidEmail(systemEmailAddress);
+                        props.setSystemEmailAddress(email);
+                    }
+                    catch (ValidEmail.InvalidEmailException e)
+                    {
+                        errors.reject(SpringActionController.ERROR_MSG, "Invalid System Email Address: ["
+                                + e.getBadEmail() + "]. Please enter a valid email address.");
+                        return response;
```
cnathe: is it necessary to return after each of these error cases? could we instead build up the error object and then only call save, write audit, etc. if errors is empty at the end?

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -10381,6 +10381,183 @@ protected TYPE getType()
...
+                    props.setRestrictedColumnsEnabled(json.optBoolean("restrictedColumnsEnabled"));
+                }
+                catch (IllegalArgumentException e)
+                {
+                    errors.reject(ERROR_MSG, "Invalid restricted columns flag: " + e.getMessage());
+                    return response;
+                }
+            }
+
+            props.save();
```
cnathe: This approach will work for sure...but it is a lot of duplicate code. Is the main reason for the duplication because we need to support the "update partial settings" in this new action? It looks like the `saveFolderSettings` method is an attempt to share some of this code between a couple actions, but I can't think of a good way to refactor this to remove the duplicate code. Bummer.

cnathe: Maybe the way to handle it would be to convert the existing UI form action to use the new API action and then we could remove that save/execute part of the UI action. Thoughts? maybe that would be a TODO for after this PR

XingY: good suggestion. changed pushed.

cnathe (approved):

PR #4608, opened 2023-07-19
New Line Support for Multi Value Filters

RosalineP:
#### Rationale
Whereas previously it was necessary to use a semicolon delimited list of values with multi-value filters such as 'Contains One Of', with these changes it is possible to use a newline as a delimiter. As such, filter value inputs for relevant multi-value filters have been changed from field inputs to textarea inputs. See [spec](https://docs.google.com/document/d/1148zLiMiyYkYbZd2mODGhH2kQK7XIdZGsK3lktnpWh0/edit?usp=sharing) for more details.

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/156
* https://github.com/LabKey/labkey-ui-premium/pull/140
* https://github.com/LabKey/labkey-ui-components/pull/1239
* https://github.com/LabKey/platform/pull/4608
* https://github.com/LabKey/sampleManagement/pull/1976
* https://github.com/LabKey/biologics/pull/2244
* https://github.com/LabKey/inventory/pull/931
* https://github.com/LabKey/testAutomation/pull/1578

#### Changes
* Where relevant, use secondary value seperator while parsing values
* Ext updates to support frontend textarea filter-value input (thanks Xing!)
* Update packages


api/webapp/clientapi/ext3/FilterDialog.js
```diff
@@ -716,71 +765,80 @@ LABKEY.FilterDialog.View.Default = Ext.extend(LABKEY.FilterDialog.ViewPanel, {
...
+                filterIndex   : idx,
+                id            : 'value_'+(idx + 1) + (typeId ? '-' + typeId: ''),   //for compatibility with tests...
+                width         : 330,
+                blankText     : 'You must enter a value.',
+                validateOnBlur: true,
+                value         : null,
+                altFormats    : this.getAltDateFormats(),
+                hidden: typeId === 1,
+                disabled: typeId === 1,
+                emptyText: xType === 'textarea' ? 'Use semicolon or new line to separate entries' : null,
```
RosalineP: ToDo: Update as Adam suggested to 'Use new line or semicolon to separate entries'

query/src/org/labkey/query/reports/getdata/FilterClauseBuilder.java
```diff
@@ -142,7 +142,7 @@ public void testNullToSQL()
         @Test
         public void testInClauseToSQL()
         {
-            assertEquals("(((\"Field1\" IN ('value1', 'value2'))))", toLabKeySQL(CompareType.IN, "value1;value2", JdbcType.VARCHAR));
+            assertEquals("(((\"Field1\" IN ('value2', 'value1'))))", toLabKeySQL(CompareType.IN, "value1;value2", JdbcType.VARCHAR));
```
XingY: Why did the ordering change?

api/webapp/clientapi/ext3/FilterDialog.js
XingY: still todo?

api/src/org/labkey/api/data/CompareType.java
```diff
@@ -1086,6 +1122,11 @@ public String getValueSeparator()
         return _valueSeparator;
     }
 
+    public String getSecondaryValueSeparator()
+    {
+        return _secondaryValueSeparator;
```
XingY: _secondaryValueSeparator is never set and always null. Can getSecondaryValueSeparator() just return null and skipping adding _secondaryValueSeparator?

api/src/org/labkey/api/data/CompareType.java
```diff
@@ -972,16 +997,27 @@ protected static void parseParams(String value, String separator, Collection<Str
...
             return Collections.emptySet();
+
+        if (secondarySeparator != null)
+            value = value.replace(secondarySeparator, separator);
+
         // check for JSON marker
         Set<String> values = new LinkedHashSet<>();
         parseParams(value, separator, values);
-        return values;
+        return (secondarySeparator != null)
```
XingY: Seems random to me that trimming is only done when there is a secondarySeparator. What about instead of using String secondarySeparator,  use boolean allowNewlineSep?

api/src/org/labkey/api/data/CompareType.java
XingY: public boolean isNewLineSeparatorAllowed()
    {
        return false;
    }

api/src/org/labkey/api/data/CompareType.java
```diff
@@ -972,16 +997,27 @@ protected static void parseParams(String value, String separator, Collection<Str
...
+    protected static Set<String> parseParams(Object value_, String separator, @Nullable String secondarySeparator)
     {
         if (value_ == null)
             return Collections.emptySet();
         String value = value_.toString();
         if (StringUtils.isBlank(value))
             return Collections.emptySet();
+
+        if (secondarySeparator != null)
+            value = value.replace(secondarySeparator, separator);
```
XingY: This should check !value.startsWith(JSON_MARKER_START)

XingY (changes requested):

api/webapp/clientapi/ext3/FilterDialog.js
RosalineP: Done!

api/src/org/labkey/api/data/CompareType.java
RosalineP: Good catch, thanks!

query/src/org/labkey/query/reports/getdata/FilterClauseBuilder.java
RosalineP: At first I thought this was specific to the helper functions and unimportant, but it turns out that despite parseParams() returning a set, here we do rely on the ordering. I've updated the code to preserve order

api/src/org/labkey/api/data/CompareType.java
```diff
@@ -971,17 +995,23 @@ protected static void parseParams(String value, String separator, Collection<Str
         }
     }
 
-    protected static Set<String> parseParams(Object value_, String separator)
+    protected static Set<String> parseParams(Object value_, String separator, boolean isNewLineSeparatorAllowed)
```
RosalineP, line 1003: I ended up getting rid of the other signature so that other usages would actually pass in their isNewLineSeparatorAllowed() fn, but let me know if that wasn't what you were picturing!

api/src/org/labkey/api/data/CompareType.java
RosalineP: Like so?

XingY (approved):

PR #4613, opened 2023-07-20
Fix participant indexing when participant aliases are configured

labkey-adam:
#### Rationale
The filter to select participants for full-text search indexing is too strict when a participant alias dataset is configured.

When the PTID list is null we should select all participants. Currently, however, when an alias table is present, we're adding an overly restrictive condition about modified aliases. Instead, we should add this condition only if we're selecting specific participants (i.e., the `lastIndexedClause` is not empty).

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4577

### DIFF
--- a/study/src/org/labkey/study/model/StudyManager.java
+++ b/study/src/org/labkey/study/model/StudyManager.java
@@ -44,7 +44,8 @@ import org.labkey.api.compliance.ComplianceService;
 import org.labkey.api.data.*;
 import org.labkey.api.data.DbScope.CommitTaskOption;
 import org.labkey.api.data.DbScope.Transaction;
-import org.labkey.api.data.SimpleFilter.FilterClause;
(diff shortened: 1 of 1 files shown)


labkey-adam:
All previously failing search tests are now passing on this FB:

* https://teamcity.labkey.org/buildConfiguration/LabKey_237Release_Community_DailySuites_DailyEPostgres/2547256?buildTab=tests
* https://teamcity.labkey.org/buildConfiguration/LabKey_237Release_Community_ModuleSuites_SearchPostgres/2547258?buildTab=tests
* https://teamcity.labkey.org/buildConfiguration/LabKey_237Release_Community_ModuleSuites_SearchSqlserver/2547260?buildTab=tests

labkey-matthewb (approved):

PR #4615, opened 2023-07-21
Transform script support for plate based standard assay

### LINKED ISSUE #4610: Issue 48293: Standard Assay with Plate Metadata & Transformation Script throws an error
[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48293)

labkey-klum:
#### Rationale
This addresses a few issues when trying to use the standard assay with plate metadata.

When attempting to use a transform script the server was throwing an error when trying to serialize the metadata JSON file. More seriously, the run data that gets passed back to the transform script did not contain any of the plate metadata. With this PR any uploaded plate metadata will get merged with the incoming data and passed back to the transform script in the `runDataFile.tsv`. This should work for either the assay import UI and via the `saveBatch` or `importRun` APIs.

Also fixed is the download sample metadata link from the assay import wizard was not always resolving the configured plate template.

#### Related Issues
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48293
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48292


labkey-nicka (approved):
Looks good. We should have a follow on story for automated regression coverage of these features.

PR #4622, opened 2023-07-25
Stop serializing entire rows/beans on update forms

labkey-adam:
#### Rationale
JSON serialization is unnecessary and can be problematic. We now serialize/deserialize just the version and PK, as a simple JSON map string that's HTML encoded. Legacy JSON serialization is left behind under an experimental feature flag for now out of an abundance of caution.


labkey-adam:
In addition to BVT, BVT-EHR, and Study Postres (above), Daily Postgres ran completely green on this FB: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/2769509?buildTab=dependencies&mode=list&type=snapshot

labkey-matthewb (approved):

PR #4625, opened 2023-07-26
Issue 46427: Slow queries for runs related to datas and materials

labkey-jeckels:
#### Rationale
We can avoid joining to the `exp.data` and `exp.material` tables for these queries because they're only invoked for single rows at a time, and we already have the data object so we know if it has a `sourceApplicationId`

#### Changes
* Drop the extra join and don't do an `OR` unless we need to


labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -5458,33 +5458,33 @@ public List<ExpRunImpl> getRunsUsingDataIds(List<Integer> ids)
...
                         INNER JOIN exp.ProtocolApplication pa ON pa.RunId = r.RowId
                         INNER JOIN exp.DataInput di ON di.targetApplicationId = pa.RowId
-                        LEFT OUTER JOIN exp.Data d ON pa.RowId = d.sourceApplicationID
-                        WHERE di.dataId\s""");
-        getExpSchema().getSqlDialect().appendInClauseSql(sql, ids);
-        sql.append("\n");
-        sql.append("OR d.RowId ");
-        getExpSchema().getSqlDialect().appendInClauseSql(sql, ids);
-        sql.append("\n");
+                        WHERE di.dataId = ?""");
```
```suggestion
                        WHERE di.dataId = ? """);
```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -5458,33 +5458,33 @@ public List<ExpRunImpl> getRunsUsingDataIds(List<Integer> ids)
...
+        sql.add(data.getRowId());
```
labkey-susanh: Should this use `sql.appendValue` instead?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -5531,33 +5531,33 @@ private SQLFragment getAppendInClause(Collection<?> ids)
...
                         INNER JOIN exp.MaterialInput mi ON mi.targetApplicationId = pa.RowId
-                        LEFT OUTER JOIN exp.Material m ON pa.RowId = m.sourceApplicationID
-                        WHERE mi.materialId\s""");
-        getExpSchema().getSqlDialect().appendInClauseSql(sql, ids);
-        sql.append("\n");
-        sql.append("OR m.RowId ");
-        getExpSchema().getSqlDialect().appendInClauseSql(sql, ids);
-        sql.append("\n");
+                        WHERE mi.materialId = ?""");
+        sql.add(material.getRowId());
```
labkey-susanh: Same comments here.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-jeckels: I fixed the whitespace problem with a different commit.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-jeckels: `appendValue()` inlines the value into the SQL itself. `add()` supplies a value for the `?` parameter. The latter means that the DB can reuse the same query plan over multiple executions, and that we don't end up with a bunch of variations with different rowId values in the query profiler.

labkey-susanh: Ah right.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -5531,33 +5531,33 @@ private SQLFragment getAppendInClause(Collection<?> ids)
...
-        sql.append("\n");
-        sql.append("OR m.RowId ");
-        getExpSchema().getSqlDialect().appendInClauseSql(sql, ids);
-        sql.append("\n");
-        sql.append("ORDER BY r.Created DESC");
+                        WHERE mi.materialId = ?""");
+        sql.add(material.getRowId());
+        if (material.getSourceApplication() != null)
+        {
+            sql.append(" OR pa.RowId = ?");
```
labkey-matthewb: OR where the two sides reference different tables are _terrible_
Better to just UNION the two cases

labkey-matthewb (changes requested):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-jeckels: OK. New version pushed.

labkey-matthewb (approved):

PR #4627, opened 2023-07-26
Issue 47359: Support updating registry from file with merge option

XingY:
#### Rationale
Issue [47359](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47359): Update vectors from file does not update selection methods as expected

DataClass data rowId and lsid are auto generated columns during import. For merge operations, the auto-generated rowId and lsid that's available in the data map will not match the existing data's rowid and lsid and won't participate in the insert/update action. However, the auto generated temp rowid/lsid is being used in triggers for handling related entities or lineage actions, which resulted in incorrect or failed merge for existing data.

This PR allow triggers to use the existingRecord for the data row. When available, existingRecord is favored over newRow for determining the data row's key values.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4627
* https://github.com/LabKey/biologics/pull/2260
* https://github.com/LabKey/wnprc-modules/pull/367

#### Changes
*


api/src/org/labkey/api/data/triggers/Trigger.java
```diff
@@ -101,17 +101,17 @@ default void complete(TableInfo table, Container c, User user, TableInfo.Trigger
     {
     }
 
-
     default void rowTrigger(TableInfo table, Container c, User user, TableInfo.TriggerType event, boolean before, int rowNumber,
                             @Nullable Map<String, Object> newRow, @Nullable Map<String, Object> oldRow,
-                            ValidationException errors, Map<String, Object> extraContext) throws ValidationException
+                            ValidationException errors, Map<String, Object> extraContext,
+                            @Nullable Map<String, Object> existingRecord/*existing record for the row, used for merge operation to differentiate new vs existing row*/) throws ValidationException
```
labkey-nicka: nit: Not sure this one-off comment here is useful given the comment on the `fireRowTrigger` interface.

labkey-nicka (approved):

labkey-martyp:
@XingY is this a biologics specific use case? Or something that needs to be supported more broadly in triggers?

XingY:
> @XingY is this a biologics specific use case? Or something that needs to be supported more broadly in triggers?

This implementation uses existing record, which is only available when audit level is set to detailed. The cud actions from apps have auditbehavior set. Other usages are not guaranteed.

PR #4629, opened 2023-07-27
Issue 48193: LKS: Creating file directories with same name & different casing on hosted servers

XingY:
#### Rationale
Issue [48193](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48193): LKS: Creating file directories with same name & different casing on hosted servers

The per request resourceCache uses Path as key. Path is case insensitive. For a file system is case sensitive, such as linux, the case insensitive resourceCache is failing to return resources with different casing. 

This PR create a subclass DavPath, that checks OS to guess desired case-sensitivity for the file system.

Additionally, a new subclass S3Path is created for S3 files, to accommodate scenarios of using S3 storage for on premise deployments.  

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4629
* https://github.com/LabKey/cloud/pull/130

#### Changes
* Use DavPath as resourceCache key, which checks OS to determine case-sensitivity
* (Optional change) Special casing S3 resources to use a separate S3Path that's always case-sensitive


labkey-jeckels (commented):
It looks like we might be able to add a `toPath()` method on `WebdavResource`. Most resources can return a `DavPath` but the S3 resources can override and return `S3Path`. I think that might let us get rid of the separate `s3ResourceCache`.

We already have more code than I'd like that's special-casing for S3, so I'm looking for ways to avoid adding more checks.

XingY:
> It looks like we might be able to add a `toPath()` method on `WebdavResource`. Most resources can return a `DavPath` but the S3 resources can override and return `S3Path`. I think that might let us get rid of the separate `s3ResourceCache`.
> 
> We already have more code than I'd like that's special-casing for S3, so I'm looking for ways to avoid adding more checks.

I thought about using a single cache, but couldn't think a way that won't break equals() contract. For example, there are 2 resources, one on macOs ("/labkey_webdav/folder1"), one on S3 ("/labkey_webdav/FOLDER1"). macOs.equals(S3) would return true, but S3.equals(macOs) would return false. Though in reality this scenario is unlikely to happen because a file root is either mounted or not, but not a mix.

api/src/org/labkey/api/webdav/WebdavResource.java
```diff
@@ -243,4 +243,8 @@ public interface WebdavResource extends Resource
      */
     void setLastModified(long time) throws IOException;
 
+    default DavPath toDavPath(org.labkey.api.util.Path path)
```
labkey-jeckels: Do we need the `path` argument here?

labkey-jeckels: And can you add a JavaDoc comment about how this differs from `getPath()`

labkey-jeckels (approved):
Nice! See comments for a couple of minor suggestions but no need for additional review unless you want it

PR #4635, opened 2023-07-29
Add a proactive startup test for encryption key changes

labkey-adam:
#### Rationale
Our cloud team would like the server to perform a test at startup that detects encryption key changes. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=48346

We also now display an admin warning if an encryption key is not configured.


api/src/org/labkey/api/security/Encryption.java
```diff
@@ -81,16 +86,104 @@ public void addDynamicWarnings(@NotNull Warnings warnings, @NotNull ViewContext
...
+                    LOG.info("Encryption key test property was generated, encrypted, and stored. It will be tested on subsequent server startups.");
+                }
+                else
+                {
+                    String base64 = map.get(TEST_BYTES_NAME);
+                    byte[] combined = Base64.decodeBase64(base64); // This doesn't seem to throw, no matter what garbage you throw at it
+                    if (combined.length != TEST_BYTES_LENGTH + SHA1_LENGTH)
+                    {
+                        // Base64 decoding problem -- log it and treat as a decryption failure
+                        LOG.error("Encryption key test failed: Base64 decoding failed");
```
labkey-jeckels, line 154: I know admins will see more details in the UI when there's a failure, but maybe we should give some similar guidance in the log too.

labkey-jeckels (approved):

api/src/org/labkey/api/security/Encryption.java
labkey-adam: Logged guidance

PR #4642, opened 2023-08-02
Backport ApiXmlWriter JSON library update - DO NOT MERGE FORWARD

labkey-martyp:
#### Rationale
EHR controller was converted to the new JSON object library in 23.3. The ApiXmlWriter didn't get converted until 23.7. There is one use case in the EHR using the ApiXmlWriter for an EHR action. This PR partially backports the JSON conversion of ApiXmlWriter to 23.3 so it will handle both the new and old JSON library. 

#### Related Pull Requests
* https://github.com/LabKey/onprcEHRModules/pull/748

#### Changes
* Handle new JSONObject type
* Handle new JSONArray type


labkey-adam (approved):
I would put "DO NOT MERGE FORWARD" comments in the code as well. No need for re-review.

PR #4649, opened 2023-08-03
48185: Update of the User table clears any non-specified custom fields

labkey-klum:
#### Rationale
I'm a bit surprised that we haven't run into this earlier but perhaps it's because form updates wouldn't expose this, many datatypes already override this method and maybe there aren't that many extensible tables running this code?

[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48185)


labkey-matthewb (approved):
This looks correct to me.  It's a shame we haven't converted on using StatementUtils everywhere.

PR #4661, opened 2023-08-10
Issue 48388: QuerySnapshotDependencyThread not setting Query environment

labkey-jeckels:
#### Rationale
We should support query snapshots on servers that have the Compliance module deployed

#### Changes
* Set the environment so that we can figure out if/how to log

### DIFF
--- a/study/src/org/labkey/study/dataset/DatasetSnapshotProvider.java
+++ b/study/src/org/labkey/study/dataset/DatasetSnapshotProvider.java
@@ -700,6 +700,9 @@ public class DatasetSnapshotProvider extends AbstractSnapshotProvider implements
                     SnapshotDependency.SourceDataType data = QUEUE.take();
                     try
                     {
+                        // Issue 48388 - set the environment for compliance logging purposes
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-matthewb I don't love the use of the search user here. If you have ideas on a better approach, let me know.

labkey-matthewb (approved):

labkey-matthewb:
> 

ETL must stash the user that enabled the job.  We could copy that pattern, but that can also have problems.  It's easy to create a system user with a mkrw appropriate name if that name is showing up somewhere.

labkey-jeckels:
> ETL must stash the user that enabled the job. We could copy that pattern, but that can also have problems. It's easy to create a system user with a mkrw appropriate name if that name is showing up somewhere.

Thinking about this scenario a little more, I don't think we should log anything in this case, since the user who set up the snapshot won't see any PHI as part of refreshing the snapshot. They'd have to go look at the snapshot.

I'm sure that there are scenarios that this PR won't fully address, but it should enable a bunch that will currently cause an error.

PR #4666, opened 2023-08-11
Issue 48437: Error rendering NAb results grid with PercentNeutralizationInitialDilution when well group doesn't have a single min dilution row

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48437

#### Changes
- use AVG() since we can't guarantee that there will be just a single DilutionData row for the min dilution


cnathe:
@labkey-klum could I get your code review on this change?

labkey-klum (approved):

PR #4667, opened 2023-08-11
Support assay run import of plate based metadata

### LINKED ISSUE #4636: AbDisc, Connect Plate Metadata to Assay data - dev
[spec](https://docs.google.com/document/d/1kS4hU7EvPnkXjs5dnNWhjhPbbSS3w0p8ydOJ0k7fgeg/edit#heading=h.hpvwefcmri7g)

**Rationale**
We would like to support adding metadata to plates, on a per well basis. While we have some support for this in the standard assay, it doesn't currently support all of the scenarios required for AB Discovery.

This phase of the development work will focus on creating a domain that can be attached to the `Assay.Well` table and be able to query and insert data associated with the vocabulary domain properties.

**Tasks**
- [x] Implemenation plan - 5
- [x] Conditionally add `WellLsid` field to the assay results domain - 3
- [x] `AssayPlateMetadataService` updates
  - [x] Implement an `ImportHelper` class to add additional bindings to join `assay.well` to results - 5
  - [x] Map incoming well location rows to existing `LSID` values on the `assay.well` table - 3
- [x] Add metric to track assays with plate metadata - 2 
- [x] Align with existing standard assay plate support - 3
- [x] Transform script support, merge in plate metadata - 5 📍 
- [x] Don't require metadata JSON file - 2
- [x] Plate endpoints
  - [x] Set Fields : 3

labkey-klum:
#### Rationale
This iteration of plate metadata support builds on prior work which added a vocabulary domain in order to support custom fields for wells on a plate. In this PR we now fully support the ability to add custom fields to a plate and add values to the custom fields on a per well basis. This plate can now be imported into an assay run and the custom well values are joined to the assay results. This is in contrast to the prior plate metadata support where JSON metadata was needed to at run import time in order to combine plate and assay results. For now both methods are supported but in the apps, the JSON file is no longer required if the `app plate` experimental flag is enabled.

Additionally transform script support is added for metadata introduced directly via a plate. Similar to the legacy JSON metadata transform script support, the plate metadata is combined with the assay results and made available to the transform script via the `runData.tsv` file. The plate metadata values cannot but updated by a transform script but those values can be used to help compute or validate assay result field values. While not a primary scenario, plate metadata furnished both by JSON or plate wells can be provided in the same run. 

A metric was added to determine how many protocols have plate support enabled which will help guide future support of the legacy JSON method.

#### Changes
- Add a new endpoint to set `WellCustomField` values for a specific well location on a plate. A `WellCustomField` is a field that has been configured for a plate and contains the value to set for the well.
- On assay import, if plate based metadata exists, the `lsid` for the `Well` row is injected into the assay result row. The vocab domain is already joined to the `Well` table in order to pull in vocab properties. This occurs via the `ImportHelper` class that is created by the `AssayPlateMetadataService`.
- Inject plate metadata values into the run data that is handed to transform scripts.
- Fixed a bug where JSON based plate metadata was not working for transform scripts for import via the apps. See line 187 of `AbstractAssayTsvDataHandler`.

### DIFF
--- a/api/src/org/labkey/api/assay/plate/AssayPlateMetadataService.java
+++ b/api/src/org/labkey/api/assay/plate/AssayPlateMetadataService.java
@@ -13,6 +13,7 @@ import org.labkey.api.exp.api.ExpProtocol;
 import org.labkey.api.exp.api.ExpRun;
 import org.labkey.api.exp.property.Domain;
 import org.labkey.api.security.User;
+import org.labkey.api.settings.ExperimentalFeatureService;
 
 import java.io.File;
 import java.util.HashMap;
@@ -32,6 +33,11 @@ public interface AssayPlateMetadataService
         }
         else
             throw new RuntimeException("The specified assay data type is null");
+    }
+
+    static boolean isExperimentalAppPlateEnabled()
+    {
+        return ExperimentalFeatureService.get().isFeatureEnabled("experimental-app-plate-support");
     }
 
     @Nullable
@@ -64,7 +70,7 @@ public interface AssayPlateMetadataService
      *
      * @return the merged rows
      */
-    List<Map<String, Object>> mergePlateMetadata(Lsid plateLsid, List<Map<String, Object>> rows, Map<String, MetadataLayer> plateMetadata,
+    List<Map<String, Object>> mergePlateMetadata(Container container, User user, Lsid plateLsid, List<Map<String, Object>> rows, @Nullable Map<String, MetadataLayer> plateMetadata,
                                                  ExpProtocol protocol) throws ExperimentException;
 
     /**
--- a/api/src/org/labkey/api/assay/plate/AssayPlateMetadataService.java
+++ b/api/src/org/labkey/api/assay/plate/AssayPlateMetadataService.java
@@ -1,5 +1,6 @@
 package org.labkey.api.assay.plate;
 
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONObject;
 import org.labkey.api.assay.AssayDataType;
@@ -83,6 +84,7 @@ public interface AssayPlateMetadataService
      * Returns an import helper to help join assay results data to well data and metadata that is associated
      * with the plate used in the assay run import
      */
+    @NotNull
     OntologyManager.UpdateableTableImportHelper getImportHelper(Container container, User user, ExpRun run, ExpData data, ExpProtocol protocol, AssayProvider provider) throws ExperimentException;
 
     interface MetadataLayer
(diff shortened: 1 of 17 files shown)


api/src/org/labkey/api/assay/plate/AssayPlateMetadataService.java
```diff
@@ -33,6 +35,11 @@ static void registerService(AssayDataType dataType, AssayPlateMetadataService ha
             throw new RuntimeException("The specified assay data type is null");
     }
 
+    static boolean isExperimentalAppPlateEnabled()
+    {
+        return ExperimentalFeatureService.get().isFeatureEnabled("experimental-app-plate-support");
```
labkey-nicka, line 41: When I declared the experimental flag I declared it in the Biologics module since we were thinking the server-side wouldn't be looking at this flag at all. If we're going to be using this flag here then we should consider moving the experimental flag into the assay module.

api/src/org/labkey/api/assay/plate/AssayPlateMetadataService.java
```diff
@@ -72,6 +79,12 @@ List<Map<String, Object>> mergePlateMetadata(Lsid plateLsid, List<Map<String, Ob
     Map<String, MetadataLayer> parsePlateMetadata(JSONObject json) throws ExperimentException;
     Map<String, MetadataLayer> parsePlateMetadata(File jsonData) throws ExperimentException;
 
+    /**
+     * Returns an import helper to help join assay results data to well data and metadata that is associated
+     * with the plate used in the assay run import
+     */
+    OntologyManager.UpdateableTableImportHelper getImportHelper(Container container, User user, ExpRun run, ExpData data, ExpProtocol protocol, AssayProvider provider) throws ExperimentException;
```
labkey-nicka, line 88: nit: Consider annotating this and the underlying implementation as `@NotNull`.

assay/src/org/labkey/assay/plate/PlateImpl.java
```diff
@@ -500,6 +511,15 @@ protected WellGroupImpl createWellGroup(String name, WellGroup.Type type, List<P
     public void setWells(WellImpl[][] wells)
     {
         _wells = wells;
+
+        // create a rowId to well map
+        _wellMap = new HashMap<>();
+        Arrays.stream(_wells).toList().forEach(w -> {
```
labkey-nicka: nit: I don't believe either call to `.toList()` is necessary here as `forEach` is supported on `Stream`.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -1452,6 +1469,45 @@ public List<PlateField> removeFields(Container container, User user, Integer pla
...
+        if (well == null)
+            throw new IllegalArgumentException("Failed to set plate custom field values. Well id \"" + wellId + "\" not found.");
+
+        Domain domain = getPlateMetadataDomain(container, user);
+        if (domain == null)
+            throw new IllegalArgumentException("Failed to set plate custom field values. Custom fields domain does not exist. Try creating fields first.");
+
+        // verify the fields have been configured for the plate
+        Set<String> configuredProps = new HashSet<>();
+        SQLFragment sql = new SQLFragment("SELECT PropertyURI FROM ").append(AssayDbSchema.getInstance().getTableInfoPlateProperty(), "")
```
labkey-nicka: Should this query happen within the transaction boundary?

labkey-nicka (approved):

PR #4668, opened 2023-08-12
Reduce calls to OntologyManager.getPropertyObjects() by reusing defaults

labkey-jeckels:
#### Rationale
We call `OntologyManager.getPropertyObjects()` a lot. It's usually fast, but not always. One big source is from calling it once per property instead of a single time for the whole domain.

#### Changes
* Use a `CachingSupplier` to make it easy to fetch defaults once per domain instead of once per property
* Adjust callers to reuse the defaults
* Eliminate some deprecated methods and collapse some confusing overloaded methods
* Minor cleanup

### DIFF
--- a/api/src/org/labkey/api/data/BaseColumnInfo.java
+++ b/api/src/org/labkey/api/data/BaseColumnInfo.java
@@ -183,16 +183,9 @@ public class BaseColumnInfo extends ColumnRenderPropertiesImpl implements Mutabl
         setAlias(rsmd.getColumnName(col));
     }
 
-    @Deprecated
-    public BaseColumnInfo(String name, TableInfo parentTable)
-    {
-//        assert -1 == name.indexOf('/');
-        this(new FieldKey(null, name), parentTable);
-    }
-
     public BaseColumnInfo(String name, TableInfo parentTable, JdbcType type)
     {
-        this(new FieldKey(null, name), parentTable, type);
+        this(FieldKey.fromParts(name), parentTable, type);
     }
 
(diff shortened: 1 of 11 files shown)


api/src/org/labkey/api/data/BaseColumnInfo.java
```diff
@@ -183,16 +183,9 @@ public BaseColumnInfo(ResultSetMetaData rsmd, int col) throws SQLException
...
-    public BaseColumnInfo(String name, TableInfo parentTable)
-    {
-//        assert -1 == name.indexOf('/');
-        this(new FieldKey(null, name), parentTable);
-    }
-
     public BaseColumnInfo(String name, TableInfo parentTable, JdbcType type)
     {
-        this(new FieldKey(null, name), parentTable, type);
+        this(FieldKey.fromParts(name), parentTable, type);
```
labkey-matthewb, line 188: This feels like a big change to me.  Name is the most common way to construct a BaseColumnInfo, and now we're running them all through a (somewhat strange) decoder.  I would prefer that callers explicitly use the FieldKey constructor if they need to.

labkey-jeckels: I've always considered `fromParts` with a single argument to be equivalent the constructor with a `null` parent. I can revert, but do I have bad understanding? I thought that `fromString` does decoding.

labkey-matthewb: My mistake.  I'm so used to the (null,name) idiom that I read that as fromString().

labkey-matthewb (approved):

PR #4675, opened 2023-08-15
Update eigenbase libs and transitive dependencies

Sigmonia:
#### Rationale
https://teamcity.labkey.org/viewLog.html?buildId=2584515&buildTypeId=LabKey_Trunk_Premium_InternalSuites_OwaspDependencyCheck

#### Related Pull Requests
* https://github.com/LabKey/server/pull/548

#### Changes
* Update versions of Eigenbase libraries

### DIFF
--- a/query/build.gradle
+++ b/query/build.gradle
@@ -43,7 +43,7 @@ dependencies {
     BuildUtils.addExternalDependency(
         project,
         new ExternalDependency(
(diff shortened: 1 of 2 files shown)


labkey-susanh (approved):
Curious that the group changed and the availability of other versions is spotty. I'm glad these versions are on mavenCentral.

PR #4676, opened 2023-08-16
Create XmlToObject subclass

bbimber:
Rationale
With the XStream update in 23.7, we need to consistently use a custom deserializer subclass, analogous to the subclass for ObjectToXml

Changes
Add XmlToObject subclass


labkey-jeckels (approved):
Verified on my dev machine.

PR #4685, opened 2023-08-19
TOTP Phase 2 enhancements

labkey-adam:
#### Rationale
Support TOTP Phase 2 enhancements

#### Related Pull Requests
* https://github.com/LabKey/mfa/pull/2

#### Changes
* Add methods for global user-specific property sets
* Rework `UserDetailsButtonProvider.addCustomButtons()` to take details user in addition to current user
* Add a mechanism to disallow insert of new authentication configurations

### DIFF
--- a/core/src/org/labkey/core/login/LoginController.java
+++ b/core/src/org/labkey/core/login/LoginController.java
@@ -2721,6 +2721,7 @@ public class LoginController extends SpringActionController
             if (null != testLink)
                 m.put("testLink", testLink);
             m.put("settingsFields", ap.getSettingsFields());
+            m.put("allowInsert", ap.allowInsert());  // TODO: Update client to gray out disallowed providers
             return m;
         }
     }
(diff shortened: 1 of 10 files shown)


core/src/org/labkey/core/login/LoginController.java
```diff
@@ -2721,6 +2721,7 @@ private Map<String, Object> getProviderMap(AuthenticationProvider ap)
             if (null != testLink)
                 m.put("testLink", testLink);
             m.put("settingsFields", ap.getSettingsFields());
+            m.put("allowInsert", ap.allowInsert());  // TODO: Update client to gray out disallowed providers
```
labkey-bpatel, line 2724: Looks like this comment is similar-ish to what I had mentioned in mfa [PR](https://github.com/LabKey/mfa/pull/2), I was thinking not to display if already present, but graying out sounds better.

labkey-bpatel (approved):

PR #4305, opened 2023-04-14
Improve ETL support for remote data and delete sources

labkey-martyp:
#### Rationale
Originally opened using the bbimber/platform, here: https://github.com/LabKey/platform/pull/4239

This is related to https://github.com/LabKey/dataintegration/pull/202, which is designed to refactor dataintegration to better support remote data sources. In platform, this change will include jdbcType in the column JSON, in addition to SqlType. This is needed to allow the server to interrogate the remote data source, because the existing sqlType property is not detailed enough. Column having sqlType = timestamp can be either binary or datetime. Knowing that difference is required in the ETL for incremental updates.

PR #4686, opened 2023-08-21
Update JSONDataLoader to use jdbcType

labkey-martyp:
#### Rationale
When loading JSON data via remote api some types are being coerced into types that lose precision or in some cases change values. Using "type" in column metadata is very general and does not specify the Java type to use to prevent precision issues. The metadata value "jdbcType" was added in the related PR. This more accurately identifies the correct type. This PR backports that additional metadata field and uses that to identify the correct Java class to use when converting JSON.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4305

#### Changes
* Backport jdbcType metadata addition
* JSONDataLoader checks for "jdbcType" first, if that doesn't exist then fallback to "type"
* Update some unit tests

### DIFF
--- a/api/src/org/labkey/api/reader/JSONDataLoader.java
+++ b/api/src/org/labkey/api/reader/JSONDataLoader.java
@@ -36,6 +36,7 @@ import org.labkey.api.action.ExtendedApiQueryResponse;
 import org.labkey.api.action.ExtendedApiQueryResponse.ColMapEntry;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.DisplayColumn;
+import org.labkey.api.data.JdbcType;
 import org.labkey.api.dataiterator.DataIterator;
 import org.labkey.api.dataiterator.DataIteratorContext;
 import org.labkey.api.iterator.CloseableIterator;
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
Thanks for adding the test coverage

api/src/org/labkey/api/reader/JSONDataLoader.java
```diff
@@ -415,11 +418,32 @@ protected static ColumnDescriptor parseField(JsonParser parser, Container mvIndi
...
                 parser.nextToken();
             }
+            else if (fieldName.equals("jdbcType"))
+            {
+                String value = parser.getValueAsString();
+
+                if (value == null)
+                    throw new JsonParseException("Failed to find jdbc type property", loc);
+
+                JdbcType jdbc = JdbcType.valueOf(value);
```
labkey-matthewb: This naming is little odd.  Maybe JdbcType jdbcType, and Class jdbcTypeClass?

labkey-matthewb (approved):

api/src/org/labkey/api/reader/JSONDataLoader.java
labkey-martyp: good call, updated.

PR #4694, opened 2023-08-22
Refactor Plate cache

### LINKED ISSUE #4673: AbDisc, Refactor Plate Cache, cache custom props
[spec](https://docs.google.com/document/d/1kS4hU7EvPnkXjs5dnNWhjhPbbSS3w0p8ydOJ0k7fgeg/edit#heading=h.hpvwefcmri7g)

**Rationale**
In order to reduce the number of queries we have to make for the Antibody Discovery MVP scenarios we want to try to push more information that can be cached. This includes well and potentially well metadata.

One thing that needs to be refactored is that currently the cache is scoped at the container level so when a plate is changed all plates will need to be re-queried. We want to change the cache to per-plate granularity, but one thing that complicates this is that we currently reference plates by:
- name
- lsid
- rowId
This makes it more complicated since a `BlockingCache` key is a single object.

**Tasks**
- [x] Cache at the per plate scope
  - [x] Implement cache key for name, lsid, rowid : 3
  - [x] Implement internal cache maps : 3
  - [x] serialize/deserialize cache key : 3
  - [x] Update endpoints : 2
  - [x] Implement cache removals : 3
  - [x] Use the plate cache where possible : 3 
- [x] Additional objects to cache
  - [x] Plate custom fields : 3
  - [x] Metadata domain : 2
  - [x] Ensure creation of metadata domain for Biologics folders (with experimental flag) : 3 
- [x] Move app plate experimental flag to assay module : 1
- [x] Update via Data Iterator not working for vocab properties : 3 
- [x] Plate metadata unit tests : 3
- [x] Remove well metadata endpoints : 2

labkey-klum:
#### Rationale
Previously, we were caching plates using `container` as a key. This meant that each time a plate was un-cached, all plates in the container needed to be re-cached. We wanted to move to a finer grained per-plate cache which should improve performance as we anticipate the creation of many plates. One complication that needed to be solved is that `Plates` can be referenced by:
- row ID
- LSID
- plate name

This version of the cache uses a cache key that can be generated from each of the various reference units. When a plate is added to the cache we populate the cache with the same plate instance using the other 2 cache key variants to avoid having to re-query for a plate we already know about. The same thing happens in reverse when we un-cache a plate. An internal collection is maintained to help with the un-caching of all plates for a container.

#### Other Changes
- Moves the app plate experimental feature flag from the `biologics` module.
- Add new endpoints to get the `domainId` of the plate metadata domain and `getPlate.api`
- Add a unit test case for setting plate metadata and values
- Fixed a problem where plate metadata values could not be set using `Query.updateRows`. This now works however metadata fields are only resolved using their `PropertyURI` values.


assay/src/org/labkey/assay/plate/query/WellTable.java
```diff
@@ -107,6 +115,36 @@ protected boolean acceptColumn(ColumnInfo col)
         return super.acceptColumn(col) && !ignoredColumns.contains(col.getName());
     }
 
+    /**
+     * Override to resolve Property URIs for vocabulary columns during update. Consider adding the
+     * capability to resolve vocabulary columns by field key or name.
+     */
+    @Override
+    protected ColumnInfo resolveColumn(String name)
```
labkey-nicka, line 132: I'm curious what you think it would take to allow us to resolve these by another alias like `Properties/columnName` or `PlateMetadata/columnName`? Is that just a whole bag of worms?

labkey-nicka (approved):
Nicely refactored!

assay/src/org/labkey/assay/plate/query/WellTable.java
labkey-klum: Yeah, this should be straight forward and I'll add it to this PR as well as update the test case to verify resolving by either `PropertyURI` or by `FieldKey (Properties/columnName)`

PR #4698, opened 2023-08-24
Issue 48551: remove explicit declaration of properties for exp data/materials

labkey-nicka:
#### Rationale
This PR addresses [Issue 48551](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48551) by removing explicit declaration of the `properties` passed to `ExperimentJSONConverter.serializeExpObject()` and passes `null` instead. The reason for this is that `ExperimentJSONConverter.serializeOntologyProperties()` processes the value any explicitly provided properties against the experiment object/property store and subsequently ignores values from the provisioned tables since those properties have already been "seen". This regression was introduced with #4430 where I had made the change to hand in explicit props unknowingly changing how the values for these properties would be sourced.

#### Changes
* Update serialization of experiment data and materials to pass `null` properties rather than fetching properties from the data type's domain.
* Regression coverage added in `LineageTest`
* Updated JavaDoc comment to a buyer beware.

### DIFF
--- a/api/src/org/labkey/api/exp/api/ExperimentJSONConverter.java
+++ b/api/src/org/labkey/api/exp/api/ExperimentJSONConverter.java
@@ -678,15 +678,14 @@ public class ExperimentJSONConverter
     @NotNull
     public static JSONObject serializeData(@NotNull ExpData data, @Nullable User user, @NotNull Settings settings)
     {
-        JSONObject jsonObject = null;
+        JSONObject jsonObject = serializeExpObject(data, null, settings, user);
+
         if (settings.isIncludeProperties())
         {
             ExpDataClass dataClass = data.getDataClass(user);
 
             if (dataClass != null)
             {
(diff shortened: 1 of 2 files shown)


XingY (approved):
serializeRunGroup and serializeRun are also passing in explicit domainproperties, do they need to be updated as well?

labkey-nicka:
> serializeRunGroup and serializeRun are also passing in explicit domainproperties, do they need to be updated as well?

Thanks. I purposely left those as they were before my most recent changes.

labkey-jeckels (commented):
Can you add some JavaDoc comments to help capture the current behavior for `serializeOntologyProperties()`, which isn't at all intuitive

labkey-nicka:
> Can you add some JavaDoc comments to help capture the current behavior for `serializeOntologyProperties()`, which isn't at all intuitive

Added with 47e522b.

PR #4704, opened 2023-08-28
Issue 48499: Use preferred SecurityPolicyManager.savePolicy() variant

labkey-jeckels:
#### Rationale
We're uneven in terms of the validation and auditing we do for saving SecurityPolicies and related updates, and want to be more consistent.

#### Changes
* Let modules register ContainerSecurableResourceProviders instead of having Container know about them all
* Remove the `savePolicy` and `createContainer` methods that don't take a user, check permissions, or log for audit purposes
* Introduce `User.getAdminServiceUser()` for `sudo` like scenarios or when we're doing an operation not initiated by a user, like bootstrapping the server
* Update callers


study/src/org/labkey/study/StudyModule.java
```diff
@@ -32,6 +32,7 @@
 import org.labkey.api.audit.AuditLogService;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
+import org.labkey.api.data.ContainerSecurableResourceProvider;
```
labkey-nicka: Unused imports

study/src/org/labkey/study/StudyServiceImpl.java
```diff
@@ -124,12 +125,15 @@
 /**
  * User: jgarms
  */
-public class StudyServiceImpl implements StudyService
+public class StudyServiceImpl implements StudyService, ContainerSecurableResourceProvider
```
labkey-nicka, line 128: nit: Can/should `getSecurableResources` be removed from the `StudyService` interface?

labkey-nicka (approved):

PR #4705, opened 2023-08-28
Issue 48332: add transaction audit event cache for faster retrieval of audit events

labkey-susanh:
#### Rationale
[Issue 48332](https://www.labkey.org/home/Developer/issues/issues-update.view?issueId=48332): Currently, we use the transaction id in audit events to be able to reference all entities that were involved in a single import or update transaction. When querying for this data from audit logs, it can be quite expensive, but our applications need this only immediately after the transaction is done, so we can bypass the query to the database by caching the related audit events in memory. The cache doesn't need to be large, and could even be set up to age out very quickly, but this implementation is a simple one that gets the expected performance boost.

#### Related Pull Requests
- https://github.com/LabKey/inventory/pull/989

#### Changes
* Move `_transactionId` into bae `AuditTypeEvent` class.
* Add `_transactionEventCache` to `AuditLogImpl`, which is updated as each new event is added and queried when looking for the related events.


api/src/org/labkey/api/audit/AuditLogService.java
```diff
@@ -32,6 +32,7 @@
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
```
cnathe: is this import used in this file?

audit/src/org/labkey/audit/AuditLogImpl.java
```diff
@@ -17,13 +17,16 @@
 package org.labkey.audit;
 
 import org.apache.logging.log4j.Logger;
+import org.apache.xmlbeans.impl.soap.Detail;
```
cnathe: looks like this and a few other imports are unused and can be removed

experiment/src/org/labkey/experiment/api/data/LineageHelper.java
```diff
@@ -36,7 +36,7 @@ static ExpRunItem getStart(String lsid)
 
         List<String> lsidsToInvestigate = new ArrayList<>();
         lsidsToInvestigate.add(start.getLSID());
-        lsidsToInvestigate.addAll(svc.collectRunsToInvestigate(start, options));
+//        lsidsToInvestigate.addAll(svc.collectRunsToInvestigate(start, options));
```
cnathe: is this commented line expected? should it be removed?
seems unrelated to other changes in this PR

cnathe (approved):

experiment/src/org/labkey/experiment/api/data/LineageHelper.java
labkey-susanh: Yeah, unrelated here.  It was part of another investigation. I still think this could be removed and it might improve performance elsewhere since I see no difference in behavior, but that shouldn't be in this PR.

audit/src/org/labkey/audit/AuditLogImpl.java
labkey-susanh: Done

PR #4709, opened 2023-08-29
ScriptReport: Honor containerFilter param (Issue 48517)

labkey-alan:
#### Rationale
This PR fixes [Issue 48517](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48517)

#### Related Pull Requests
* 

#### Changes
* ScriptReport::createQueryView: Add containerFilter to QuerySettings

### DIFF
--- a/api/src/org/labkey/api/reports/report/ScriptReport.java
+++ b/api/src/org/labkey/api/reports/report/ScriptReport.java
@@ -27,6 +27,7 @@ import org.labkey.api.data.BooleanFormat;
 import org.labkey.api.data.ColumnHeaderType;
 import org.labkey.api.data.CompareType;
 import org.labkey.api.data.Container;
+import org.labkey.api.data.ContainerFilter;
 import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.DataRegion;
 import org.labkey.api.data.DisplayColumn;
@@ -102,6 +103,7 @@ public abstract class ScriptReport extends AbstractReport
         final String queryName = descriptor.getProperty(QueryParam.queryName.toString());
         final String viewName = descriptor.getProperty(QueryParam.viewName.toString());
         final String dataRegionName = descriptor.getProperty(QueryParam.dataRegionName.toString());
+        final String containerFilter = context.cloneActionURL().getParameter("containerFilter");
 
         if (context != null && schemaName != null)
         {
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/reports/report/ScriptReport.java
```diff
@@ -102,6 +103,7 @@ protected QueryView createQueryView(ViewContext context, ReportDescriptor descri
         final String queryName = descriptor.getProperty(QueryParam.queryName.toString());
         final String viewName = descriptor.getProperty(QueryParam.viewName.toString());
         final String dataRegionName = descriptor.getProperty(QueryParam.dataRegionName.toString());
+        final String containerFilter = context.cloneActionURL().getParameter("containerFilter");
```
labkey-nicka, line 106: There is a `QueryParam.containerFilterName`. Should we use that instead? I realize this changes the API prop to be `containerFilterName` which may not align with what is currently being supplied.

labkey-alan: That doesn't align with what selectRows and other APIs expect though, which is what I think we should support, since we're otherwise honoring selectRows style query params.

labkey-nicka: Agreed. I feel like we should deprecate `QueryParam.containerFilterName` and add `QueryParam.containerFilter` (not in this story). Sounds good. Resolving.

api/src/org/labkey/api/reports/report/ScriptReport.java
```diff
@@ -116,6 +118,12 @@ protected QueryView createQueryView(ViewContext context, ReportDescriptor descri
                 // need to reset the report id since we want to render the data grid, not the report
                 settings.setReportId(null);
 
+                if (containerFilter != null)
```
labkey-nicka, line 121: Doesn't this come through on `QuerySettings`? It seems like we should be able to get it from there rather than having to parse from the URL here. Additionally, this will respect any `dataRegionName` type prefixing.

labkey-alan: If it came through on the settings then we wouldn't need this PR, I'll take a closer look at how we initialize the settings to see why it's otherwise not getting populated.

labkey-alan:
This API supports `containerFilterName` already, if you prefix it with the appropriate `dataRegionName`. So I am going to close this PR and update our client instead. I would like to understand why some APIs use `containerFilter` and others use `containerFilterName` this inconsistency is less than ideal.

PR #4712, opened 2023-08-31
LABKEY.Query.schemaSelectInput - allow for config.filterFn to be applied to filter the schemas to show in the select options

cnathe:
#### Rationale
A client recently ran into a recursive type error after inadvertently setting the schemaName for a QC Trend Report configuration to the "qctrendreport" schema itself. This PR allows for those schemas to be filtered out of the options for the QC Trend Report configuration page.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4712
* https://github.com/LabKey/premium/pull/408

#### Changes
* allow for config.filterFn to be applied to filter the schemas to show in the select options

### DIFF
--- a/api/webapp/clientapi/dom/Query.js
+++ b/api/webapp/clientapi/dom/Query.js
@@ -358,7 +358,12 @@ LABKEY.Query = new function(impl, $) {
         LABKEY.Query.getSchemas({
             includeHidden: false,
             success: function(data) {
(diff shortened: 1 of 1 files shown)


labkey-susanh (approved):
Seems fine to me, but the specificity of the fix does make me wonder if we've got a need for something like this elsewhere.

PR #4715, opened 2023-09-01
Restore attribute to target web part as anchor within page

labkey-jeckels:
#### Rationale
A recent PR (https://github.com/LabKey/platform/pull/4683 to be specific) switched from `<a name="x">` to `<span name="x">` in the title of framed web parts. The means it's no longer a targetable anchor. We should keep using `<a>` but switch to `id` as the attribute

#### Changes
* Restore the `<a>` and use `id` instead of `name`

### DIFF
--- a/core/src/org/labkey/core/view/template/bootstrap/ViewServiceImpl.java
+++ b/core/src/org/labkey/core/view/template/bootstrap/ViewServiceImpl.java
@@ -610,7 +610,7 @@ public class ViewServiceImpl implements ViewService
                 out.print(PageFlowUtil.filter(title));
                 out.print("\"");
             }
-            out.print("><a id=\"" + PageFlowUtil.filter(title) + "\" class=\"labkey-anchor-disabled\">");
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/view/template/bootstrap/ViewServiceImpl.java
```diff
@@ -610,12 +610,12 @@ protected void renderWebpartHeaderStart(PrintWriter out, String title)
                 out.print(PageFlowUtil.filter(title));
                 out.print("\"");
             }
-            out.print("><span name=\"" + PageFlowUtil.filter(title) + "\" class=\"labkey-anchor-disabled\">");
+            out.print("><a id=\"" + PageFlowUtil.filter(title) + "\" class=\"labkey-anchor-disabled\">");
```
labkey-nicka: Should we consider using `PageFlowUtil.makeHtmlId(title)` instead? This is effectively the same problem we were discussing the other day in dev chat but this is relevant to the user navigation so contextualizing the id makes sense.

labkey-nicka (approved):

PR #4723, opened 2023-09-06
Issue 48602: Use a proper blocking cache for transaction audit events

labkey-susanh:
#### Rationale
[Issue 48602](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48602) Previous implementation using a poor man's cache resulted in a concurrent modification error for some code paths, so we'll use a full (but still small) cache here.

#### Related Pull Requests
* #[4705](https://github.com/LabKey/platform/pull/4705)

#### Changes
* Update transaction audit event cache implementation

### DIFF
--- a/audit/src/org/labkey/audit/AuditLogImpl.java
+++ b/audit/src/org/labkey/audit/AuditLogImpl.java
@@ -24,6 +24,8 @@ import org.labkey.api.audit.AuditLogService;
 import org.labkey.api.audit.AuditTypeEvent;
 import org.labkey.api.audit.DetailedAuditTypeEvent;
 import org.labkey.api.audit.SampleTimelineAuditEvent;
+import org.labkey.api.cache.Cache;
+import org.labkey.api.cache.CacheManager;
 import org.labkey.api.data.CompareType;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerFilter;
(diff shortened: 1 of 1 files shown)


audit/src/org/labkey/audit/AuditLogImpl.java
```diff
@@ -74,14 +75,10 @@ public class AuditLogImpl implements AuditLogService, StartupListener
...
-    {
-        @Override
-        public boolean removeEldestEntry(Map.Entry eldest)
-        {
-            return size() > 50;
-        }
-    };
+    private static final Cache<Long, List<AuditTypeEvent>> TRANSACTION_EVENT_CACHE = CacheManager.getBlockingCache(50, CacheManager.DAY,
+            "Transaction Audit Event Cache",
+            (key, argument) -> new ArrayList<>()
```
labkey-adam, line 80: Each `ArrayList` is accessed by only one thread, right? So it doesn't have to be thread-safe?

labkey-adam (approved):
Good to go, assuming `ArrayList` access is guaranteed to be single threaded

audit/src/org/labkey/audit/AuditLogImpl.java
labkey-susanh: For current usages, yes, should be accessed by a single thread, unless I'm forgetting about something...

PR #4727, opened 2023-09-07
Add support for generating assay data and more specifics for sample types

labkey-susanh:
#### Rationale
For testing application behavior and performance in the face of large sample types, we want to be able to do more incremental data generation, specifying existing sample types and data classes to use in the process where it makes sense. We also want to have assay run data for the samples to be able to assess application performance with this data. 

#### Related Pull Requests
- https://github.com/LabKey/inventory/pull/1003
- https://github.com/LabKey/labbook/pull/534
- https://github.com/LabKey/biologics/pull/2346
- https://github.com/LabKey/sampleManagement/pull/2069

#### Changes
* Add new `DataType.AssayRunData` for data generation
* Update `AssayDesignDataGenerator` to properly wire up the sampleID fields
* Add `AssayRunDataGenerator`

### DIFF
--- a/api/src/org/labkey/api/data/generator/DataGenerator.java
+++ b/api/src/org/labkey/api/data/generator/DataGenerator.java
@@ -402,7 +402,7 @@ public class DataGenerator<T extends DataGenerator.Config>
     {
         if (_config.getMaxAliquotsPerParent() <= 0)
         {
-            _log.info(String.format("Generating no aliquots because maxAliquotsPerParent is %d", _config.getMaxAliquotsPerParent()));
+            _log.info(String.format("Generating no aliquots because %s is %d", Config.MAX_ALIQUOTS_PER_SAMPLE, _config.getMaxAliquotsPerParent()));
             return 0;
         }
 
@@ -427,6 +427,7 @@ public class DataGenerator<T extends DataGenerator.Config>
             List<Map<String, Object>> parents = getRandomSamples(sampleType, Math.min(10, Math.max(quantity, quantity / 100)));
             numGenerated = generateAliquotsForParents(parents, svc, quantity - totalAliquots, 0, 1, randomInt(1, _config.getMaxGenerations()), sampleStatuses);
             totalAliquots += numGenerated;
+            _log.info("... " + totalAliquots);
             iterations++;
         }
         while (totalAliquots < quantity && numGenerated > 0);
@@ -479,7 +480,7 @@ public class DataGenerator<T extends DataGenerator.Config>
             if (errors.hasErrors())
                 throw errors;
         }
-        _log.info(String.format("... %d (generation %d)", (numGenerated + generatedCount), generation));
+//        _log.info(String.format("... %d (generation %d)", (numGenerated + generatedCount), generation));
         // for some of the aliquots, possibly generate further aliquot generations
         if (generatedCount < quantity && generation < maxGenerations)
         {
--- a/api/src/org/labkey/api/data/generator/DataGenerator.java
+++ b/api/src/org/labkey/api/data/generator/DataGenerator.java
@@ -323,6 +323,7 @@ public class DataGenerator<T extends DataGenerator.Config>
             return;
         }
         List<String> parentTypes = new ArrayList<>(config.getSampleTypeParents());
+        boolean hasParentTypes = !parentTypes.isEmpty();
         List<String> dataClassParents = new ArrayList<>(config.getDataClassParents());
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -439,7 +480,7 @@ private int generateAliquotsForParents(List<Map<String, Object>> parents, QueryU
             if (errors.hasErrors())
                 throw errors;
         }
-        _log.info(String.format("... %d (generation %d)", (numGenerated + generatedCount), generation));
+//        _log.info(String.format("... %d (generation %d)", (numGenerated + generatedCount), generation));
```
cnathe: commented line to be removed or added back?

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -448,32 +489,93 @@ private int generateAliquotsForParents(List<Map<String, Object>> parents, QueryU
...
+
+        TableInfo tableInfo = getSamplesSchema().getTable(sampleType.getName(), containerFilter);
+        SQLFragment sql = limitFromOffsetSqlFrag(tableInfo, "RowId", limit, totalSampleCount);
+        return new SqlSelector(tableInfo.getSchema(), sql).getArrayList(Integer.class);
+    }
+
+    private SQLFragment limitFromOffsetSqlFrag(TableInfo tableInfo, String columns, int limit, long totalCount)
+    {
+        if (limit <= 0)
+            return null;
```
cnathe, line 515: it looks like this check already happens in the callers above. here if this method would return null, then the `sql` variable would be null and I assume cause a problem in `SqlSelector`

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -448,32 +489,93 @@ private int generateAliquotsForParents(List<Map<String, Object>> parents, QueryU
...
+                .append(tableInfo, "dc")
                 .append(" LIMIT ")
                 .appendValue(limit);
-        if (totalSampleCount > limit)
+        if (totalCount > limit)
             sql.append(" OFFSET ")
-                .appendValue(randomLong(0, totalSampleCount-limit));
-        return new SqlSelector(tableInfo.getSchema(), sql).getArrayList(Integer.class);
+                    .appendValue(randomLong(0, totalCount-limit));
+        return sql;
```
cnathe, line 524: this is fine to leave, just asking for my own info. I was surprised to see SqlSelector here instead of just a TableSelector. Does TableSelector not support telling it an offset and limit?

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -448,32 +489,93 @@ private int generateAliquotsForParents(List<Map<String, Object>> parents, QueryU
...
+    }
+
+    public List<Map<String, Object>> selectExistingSamples(ExpSampleType sampleType, int limit, long totalSampleCount)
```
cnathe, line 527: just curious, why is there two methods for `selectExistingSamplesIds` and `selectExistingSamples`? One returns a list of integers and the other a list of maps, but otherwise they are very similar. Could selectExistingSamplesIds just call selectExistingSamples and then get out the ids?

cnathe (approved):

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: Unclear what's most useful. Probably I'll remove it.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: Indeed it would cause a problem if the callers weren't already checking. They could check the limit themselves, or check for a null value. Either way works, I think.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: TableSelector does support those and likely could be used here, but doesn't seem worth changing.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: First I needed just integers, then I needed more than that. I'm in the mode of selecting the fewest columns possible.

labkey-danield (approved):

PR #4732, opened 2023-09-08
Improve full-text search export, history, and logging

labkey-adam:
#### Rationale
Using the "Export Index Contents" feature while investigating https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48415 revealed a number of problems.

Added the option to log indexing of documents within one or more specific categories to a separate log file.

The recently added accumulated history feature on the admin page was less than perfect, in that it included the current hour in the totals, typically meaning 24 completed hours plus a 25th partial hour.

#### Changes
* Export "Category" (very useful) instead of "Type" (less useful and already part of the unique ID)
* Stop including "Summary," which isn't very useful and seems to confuse Excel
* Switch to using a `StoredFields`, which now seems to be the sanctioned way to retrieve stored data
* Filter out deleted docs
* Export to Excel using CSV, which seems to work better than TSV (eliminates errors about format conversion, hanging, and errors about content too large)
* Fix some discrepancies with accumulated history feature
* Add option to log document indexing on a per-category basis

### DIFF
--- a/search/src/org/labkey/search/model/IndexInspector.java
+++ b/search/src/org/labkey/search/model/IndexInspector.java
@@ -55,6 +55,13 @@ public class IndexInspector
             {
                 setDelimiterCharacter(DELIM.COMMA);
             }
+
+            // We want the "Export to Excel" option to open Excel and load the file. We're providing Excel content type
+            // but CSV file name and format because that seems to work and exporting in an Excel-native format would
+            // require a bit of a rewrite to the below, since ExcelWriter and TSVWriter don't share any useful
+            // interfaces. I haven't found a definitive reference that says this content type + extension discrepancy
+            // always works, but here's an old post that suggests it's reasonable:
+            // https://stackoverflow.com/questions/393647/response-content-type-as-csv#1719233
             @Override
             protected String getContentType()
             {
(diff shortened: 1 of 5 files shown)


search/src/org/labkey/search/model/IndexInspector.java
```diff
@@ -63,7 +64,7 @@ protected String getContentType()
             @Override
             protected String getFilenameExtension()
             {
-                return "xls";
+                return "csv";
```
labkey-jeckels, line 74: Adding a comment similar to your notes in the PR description would be good, as it seems odd to use XLS as the content type but use CSV as the extension.

labkey-jeckels (approved):

search/src/org/labkey/search/model/IndexInspector.java
labkey-adam: The format is CSV because it didn't seem worth it to rewrite the export code to handle both TSV and native Excel formats, given that CSV works. Note that this used to be TSV format with XLS extension, which arguably is even worse. I'll add a comment, but I haven't found a definitive source that states this is the way to cause Excel to open CSV format files.

labkey-adam: Added

labkey-jeckels (approved):

PR #4736, opened 2023-09-12
Issue 48657: LKSM/LKB: Loading selections for large grids is very slow

### LINKED ISSUE #4737: Issue 48657: LKSM/LKB: Loading selections for large grids is very slow
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48657

### Tasks
- [x] Investigate repro steps
- [x] Troubleshoot
- [x] Fix
- [ ] Manual testing 
- [x] Code review @labkey-susanh 
- [x] Merge

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48657

When setting the grid selection or clearing selections, there are a couple scenarios where we query for all PK values in a data region context so that we can verify if the selected keys are valid. This causes perf issues for large tables which are querying for a large set of data just to look for a small set. This PR fixes that by using the selected keys as a query filter for the "validate if the pks being selected are still valid" scenario on select or clear.

#### Changes
* DataRegionSelection.getSelectedItems() changes
  * no need to query the region result set if we have no selectedValues
  * no need to query for all region results if we are only interested in a subset, filter for just those
* QueryController.SelectNoneAction update to replace setSelectionForAll() with query for just the selected IDs to be removed from the context

### DIFF
--- a/api/src/org/labkey/api/data/DataRegionSelection.java
+++ b/api/src/org/labkey/api/data/DataRegionSelection.java
@@ -468,6 +468,7 @@ public class DataRegionSelection
         var ctx = dataRegionContext.second;
 
         // Issue 48657: no need to query for all region results if we are only interested in a subset, filter for just those
+        // Note: this only currently applies for tables with a single PK col. Consider altering this for multi-pk tables.
         List<ColumnInfo> pkCols = rgn.getTable().getPkColumns();
         if (pkCols.size() == 1)
         {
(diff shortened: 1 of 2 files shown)


labkey-susanh (approved):
Nice! Were you able to get a handle on what the actual impact on performance is (just curious if we can know what to expect)?

cnathe:
> Nice! Were you able to get a handle on what the actual impact on performance is (just curious if we can know what to expect)?

Yes, the perf improvement should be pretty dramatic. Instead of a 2 million row sample type taking approx a minute to find out that it has no selections, it should take less than a second. Same goes for loading and clearing selections for that size of a table (should be seconds now, not minutes). If the user decides to select all 2 million rows...it will still take a minute to set and clear that though.

PR #4739, opened 2023-09-13
DatabaseCache replay of put operations into shared cache

labkey-adam:
#### Rationale
When a `DatabaseCache` is used within a transaction, all remove and clear operations are replayed into the shared cache after successful commit. Put operations are not replayed, however. The inherent assumption has been that those objects will be loaded at some point by non-transacted code paths. It turns out we have code paths (e.g., dataset provisioned table metadata loading) that always access DatabaseCaches inside a transaction, which means their objects are effectively never cached. This is discussed in detail here: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48663

This PR makes a small change to `BlockingCache` that allows subclasses to intercept load operations that occur before put() is called. `DatabaseCache` then overrides this `load()` method to add a reload commit task on the BlockingCache if a transaction is active. This ensures that entries added during the transaction will be loaded into the shared cache immediately after commit.

This also moves many `DatabaseCache` usages to a more standard `BlockingCache` with static `CacheLoader` pattern that's required for put() replays.

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -1252,6 +1252,7 @@ public class DbScope
         return getTableInfoCache(options.getSchema().getType()).get(options);
     }
 
+    // This is expensive... use only for debugging and troubleshooting
     public <OptionType extends SchemaTableOptions> boolean isCached(OptionType options)
     {
         return getTableInfoCache(options.getSchema().getType()).isCached(options);
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -1252,7 +1252,7 @@ public class DbScope
         return getTableInfoCache(options.getSchema().getType()).get(options);
     }
 
-    // This is expensive... use only for debugging and troubleshooting
+    /** This is expensive... use only for debugging and troubleshooting */
     public <OptionType extends SchemaTableOptions> boolean isCached(OptionType options)
     {
         return getTableInfoCache(options.getSchema().getType()).isCached(options);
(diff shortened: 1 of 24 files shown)


api/src/org/labkey/api/data/DbScope.java
```diff
@@ -1252,6 +1252,12 @@ public <OptionType extends SchemaTableOptions> SchemaTableInfo getTable(OptionTy
         return getTableInfoCache(options.getSchema().getType()).get(options);
     }
 
+    // This is expensive... use only for debugging and troubleshooting
```
labkey-jeckels: Using JavaDoc makes it more likely people will see this comment in statement completion, etc

```suggestion
    /** This is expensive... use only for debugging and troubleshooting */
```

api/src/org/labkey/api/data/SchemaTableInfoCache.java
```diff
@@ -41,7 +35,20 @@ public class SchemaTableInfoCache
...
-        _blockingCache = new SchemaTableInfoBlockingCache(scope, provisioned);
+        _blockingCache = createCache(scope, provisioned);
+        _blockingCache.setCacheTimeChooser(TABLE_CACHE_TIME_CHOOSER);
+    }
+
+    // Ask the DbSchemaType how long to cache each table
+    @SuppressWarnings("DataFlowIssue")
+    private static final CacheTimeChooser<String> TABLE_CACHE_TIME_CHOOSER = (key, argument) -> ((SchemaTableOptions)argument).getSchema().getType().getCacheTimeToLive();
+
+    // This is expensive... use only for debugging and troubleshooting
```
labkey-jeckels:
```suggestion
    /** This is expensive... use only for debugging and troubleshooting */
```

labkey-jeckels (approved):

PR #4744, opened 2023-09-15
Account for different path separator character for Windows

labkey-susanh:
#### Rationale
Tests are failing on Windows because of the check for path updates. Thanks, Windows.

#### Changes
* Make all file paths have forward slashes before comparing in the tests


cnathe (approved):
Nice!

PR #4745, opened 2023-09-15
SampleType refactor

labkey-matthewb:
#### Rationale
Optimize the schema used to implement ExpMaterialTableImpl (SampleType tables) to improve performance.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1290
- https://github.com/LabKey/labkey-ui-premium/pull/191
- https://github.com/LabKey/platform/pull/4745
- https://github.com/LabKey/biologics/pull/2374
- https://github.com/LabKey/sampleManagement/pull/2098

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/data/BaseColumnInfo.java
+++ b/api/src/org/labkey/api/data/BaseColumnInfo.java
@@ -1461,7 +1461,7 @@ public class BaseColumnInfo extends ColumnRenderPropertiesImpl implements Mutabl
     @Override
     public boolean isRequiredForInsert(@Nullable DomainProperty dp)
     {
-        if (isCalculated() || isAutoIncrement() || isVersionColumn() || null != getJdbcDefaultValue())
+        if (isCalculated() || !isUserEditable() || isAutoIncrement() || isVersionColumn() || null != getJdbcDefaultValue())
             return false;
         return !isNullable() || (null != dp && dp.isRequired());
     }
--- a/experiment/resources/schemas/dbscripts/postgresql/exp-23.009-23.010.sql
+++ b/experiment/resources/schemas/dbscripts/postgresql/exp-23.009-23.010.sql
@@ -1,0 +1,12 @@
+UPDATE exp.material
+    SET rootmateriallsid = lsid
+    WHERE rootmateriallsid IS NULL;
+
+ALTER TABLE exp.material
+    ALTER COLUMN rootmateriallsid SET NOT NULL;
+
+CREATE INDEX IF NOT EXISTS uq_material_rootlsid
+    on exp.material (rootmateriallsid);
+
+-- this is duplicative of constraint uq_material_lsid
+DROP INDEX IF EXISTS exp.idx_material_lsid;
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -137,6 +137,7 @@ import static org.labkey.api.exp.api.ExpRunItem.INPUTS_PREFIX_LC;
 import static org.labkey.api.exp.api.ExperimentService.ALIASCOLUMNALIAS;
 import static org.labkey.api.exp.api.ExperimentService.QueryOptions.SkipBulkRemapCache;
 import static org.labkey.api.exp.query.ExpMaterialTable.Column.AliquotedFromLSID;
+import static org.labkey.api.exp.query.ExpMaterialTable.Column.LSID;
 import static org.labkey.api.exp.query.ExpMaterialTable.Column.RootMaterialLSID;
 import static org.labkey.api.query.AbstractQueryImportAction.configureLoader;
 import static org.labkey.experiment.api.SampleTypeUpdateServiceDI.PARENT_RECOMPUTE_LSID_COL;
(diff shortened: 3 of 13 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2262,11 +2263,28 @@ else if (isMergeOrUpdate)
...
+
+            // add "rootmateriallsid" if it does not exist
+            DataIteratorBuilder step2b = new DataIteratorBuilder()
+            {
+                @Override
+                public DataIterator getDataIterator(DataIteratorContext context)
+                {
+                    DataIterator in = step2a.getDataIterator(context);
+                    var map = DataIteratorUtil.createColumnNameMap(in);
+                    if (map.containsKey(RootMaterialLSID.toString()) || !map.containsKey(LSID.toString()))
```
XingY, line 2276: Should RootMaterialLsid be added to NOT_FOR_UPDATE? Also, what if user provide an bad rootmateriallsid for creating a non-aliquot sample, will it get checked?

experiment/resources/schemas/dbscripts/postgresql/exp-23.009-23.010.sql
```diff
@@ -0,0 +1,12 @@
+UPDATE exp.material
```
XingY, line 1: Will importing an old exported folder archive with samples be problematic after this?

XingY (commented):
Here are some additional missed checks:

SampleTypeUpdateServiceDI
`boolean isAliquot = !StringUtils.isEmpty(oldAliquotedFromLSID);`
`String aliquotRoot = (String) oldRow.get(RootMaterialLSID.name());` 

SampleManagerWorkflowSampleCompareType
`if ("aliquots".equalsIgnoreCase(aliquotType))
                    sql.append(" AND M.rootmateriallsid = ?").add(lsid);`

experiment/resources/schemas/dbscripts/postgresql/exp-23.009-23.010.sql
labkey-susanh: The folder export for samples doesn't explicitly include the rootmateriallsid, so I think it should be OK as long as we do the right thing when we create the new derivation runs that come from the export.

cnathe:
As mentioned in the chat thread, this doesn't look to be working quite as expected. After switching to the fb and starting up my server with a sample type that has ~2 million rows, I was not able to get the data grid to load at all within the app (I waited approx 10 min):
Screenshot 2023-09-15 at 2 44 40 PM

api/src/org/labkey/api/data/BaseColumnInfo.java
```diff
@@ -1461,7 +1461,7 @@ public String toString()
     @Override
     public boolean isRequiredForInsert(@Nullable DomainProperty dp)
     {
-        if (isCalculated() || isAutoIncrement() || isVersionColumn() || null != getJdbcDefaultValue())
+        if (isCalculated() || !isUserEditable() || isAutoIncrement() || isVersionColumn() || null != getJdbcDefaultValue())
```
labkey-susanh, line 1464: The `isUserEditable` check always worries me after I discovered that is is what is used to determine which fields are shown in a details view (see [Issue 43760](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43760)) and thus may not actually correspond to a normal interpretation of "user-editable".  I'm sure you know better than I about such things, but is a check for `isReadOnly` in addition or instead warranted here?

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-matthewb: It is/was already added.  See just above before the displayed diff at line 2242.

api/src/org/labkey/api/data/BaseColumnInfo.java
labkey-matthewb: I also find the overlapping properties options confusing.  I was surprised to not find one of isUserEditable or isReadOnly here.  Thanks for the issue pointer.

experiment/resources/schemas/dbscripts/postgresql/exp-23.009-23.010.sql
labkey-matthewb: I just made a fix in this code XarReader.updateSourceInfo()

api/src/org/labkey/api/data/BaseColumnInfo.java
labkey-matthewb: I agree that getFieldKeysToShow() seems wrong.  However, it also seems right that a field that can't be user-edited can't be required for insert.

cnathe:
> As mentioned in the chat thread, this doesn't look to be working quite as expected. After switching to the fb and starting up my server with a sample type that has ~2 million rows, I was not able to get the data grid to load at all within the app (I waited approx 10 min): Screenshot 2023-09-15 at 2 44 40 PM

To close the loop. I went back to my large DB with the fb and this time things performed much better. Maybe my DB was performing some other actions or the stats weren't updated yet. After the upgrade, which took 50 minutes, the sample type with 2 million rows loaded within a few seconds (which was the case before the upgrade as well) but now when I applied a "contains filter" to one of the provisioned columns for the sample type the grid loaded in < 10 seconds (note the same action without this change took > 2 minutes).

labkey-matthewb:
> As mentioned in the chat thread, this doesn't look to be working quite as expected. After switching to the fb and starting up my server with a sample type that has ~2 million rows, I was not able to get the data grid to load at all within the app (I waited approx 10 min): Screenshot 2023-09-15 at 2 44 40 PM

The obvious alternative strategy is to try a SELECT INTO/RENAME instead of UPDATE in place.  A quick search of the code shows about a half-dozen FOREIGN KEYS outside the exp module that point to exp.material.  That makes the SELECT INTO strategy much more challenging.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -1084,33 +1085,32 @@ public SQLFragment getFromSQL(String alias, Set<FieldKey> selectedColumns)
                     sql.append(comma);
                     boolean rootField = StringUtils.isEmpty(propertyColumn.getDerivationDataScope())
                             || ExpSchema.DerivationDataScopeType.ParentOnly.name().equalsIgnoreCase(propertyColumn.getDerivationDataScope());
-                    if (!rootField
-                            || "genid".equalsIgnoreCase(propertyColumn.getColumnName())
-                            || propertyColumn.isUniqueIdField())
+                    if ("genid".equalsIgnoreCase(propertyColumn.getColumnName()) || propertyColumn.isUniqueIdField())
+                    {
+                        sql.append(propertyColumn.getValueSql("m_aliquot")).append(" AS ").append(propertyColumn.getSelectName());
```
XingY: Shouldn't it be `.appendIdentifier(propertyColumn.getSelectName())` instead

XingY (changes requested):

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
labkey-matthewb: Nice catch.  We should probably switch getSelectName() to return a SQLFragment or have another method.

XingY (approved):

PR #4746, opened 2023-09-15
Issue 48688: ModuleLoader synchronization bottleneck

labkey-jeckels:
#### Rationale
Methods like `getModules()` and `getModule()` are called dozens or hundreds of times for a single page load. We need the underlying collections to be thread safe, but we can remove the synchronized blocks to reduce contention between threads. The collections are only modified during startup or when a module is redeployed, which is very infrequent.

#### Changes
* Use a `CopyOnWriteArrayList` and manually copy the maps when they're modified
* Remove synchronized blocks for read operations

### DIFF
--- a/api/src/org/labkey/api/module/ModuleLoader.java
+++ b/api/src/org/labkey/api/module/ModuleLoader.java
@@ -119,6 +119,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
@@ -130,6 +131,7 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


labkey-martyp (approved):
Yeay!

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -203,19 +205,29 @@ public enum ModuleState
...
+     */
+    private volatile Map<String, Module> _moduleMap = Collections.emptyMap();
+    private volatile Map<Class<? extends Module>, Module> _moduleClassMap = Collections.emptyMap();
+    /** Use a CopyOnWriteArrayList since the underlying collection is frequently accessed and very infrequently mutated */
+    private final List<Module> _modules = new CopyOnWriteArrayList<>();
+    /**
+     * Immutable wrapper over _modules so that we don't need to create a new wrapper every time we return it, which
+     * can be thousands of times per request
+     */
+    private final List<Module> _modulesImmutable = Collections.unmodifiableList(_modules);
```
labkey-matthewb, line 220: Is this just to avoid calling Collections.unmodifiableList() is getModule()?

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -396,16 +412,22 @@ public void updateModuleDirectory(File dir, File archive)
                 throw new IllegalStateException("Module name should not have changed.  Found '" + moduleCreated.getName() + "' and '" + moduleExisting.getName() + "'");
 
             _moduleContextMap.put(context.getName(), context);
-            _moduleMap.put(moduleCreated.getName(), moduleCreated);
+
+            Map<String, Module> newModuleMap = new CaseInsensitiveHashMap<>(_moduleMap);
+            newModuleMap.put(moduleCreated.getName(), moduleCreated);
+            _moduleMap = Collections.unmodifiableMap(newModuleMap);
```
labkey-matthewb, line 418: Shouldn't writing to _moduleMap and _moduleClassMap be synchronized?

labkey-matthewb (approved):

PR #4747, opened 2023-09-15
Fix a few more tests failing because of DatabaseCache changes

labkey-adam:
#### Rationale
After my `DatabaseCache` changes, `StudyPHIExportTest` and `DataViewsTest` are failing with `ERROR: insert or update on table "propertydomain" violates foreign key constraint "fk_propertydomain_property"`. Switch the property descriptor cache back to a plain old `DatabaseCache` for now.

`AuditInsertionFailureTestCase` is now failing because it intentionally trashes the connection which means the replay tasks throw. Catch and log these exceptions in `handleAuditFailure()` so the test continues to see the exception it expects.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4739

### DIFF
--- a/api/src/org/labkey/api/audit/AuditLogService.java
+++ b/api/src/org/labkey/api/audit/AuditLogService.java
@@ -16,6 +16,7 @@
 
 package org.labkey.api.audit;
 
+import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerFilter;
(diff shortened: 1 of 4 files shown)


labkey-adam:
`StudyPHIExportTest` and `DataViewsTest` both passed on this branch: https://teamcity.labkey.org/buildConfiguration/bt7/2637417?buildTab=overview&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true

labkey-jeckels (approved):

PR #4748, opened 2023-09-15
buildTime and serverStartupTime substitutions for page headers

labkey-jeckels:
#### Rationale
Some new properties that admins can use for a dynamic site short name: `buildTime` (from the Core module) and `serverStartTime`

#### Changes
* Add the new properties
* Don't return null (yielding NPEs) if there's an unrecognized substitution in the header expression
* New StringExpression option to pass unrecognized substitution properties through as-is
* Minor code cleanup

### DIFF
--- a/api/src/org/labkey/api/util/StringExpressionFactory.java
+++ b/api/src/org/labkey/api/util/StringExpressionFactory.java
@@ -663,13 +663,49 @@ public class StringExpressionFactory
         public enum NullValueBehavior
         {
             // Any null field results in a null eval (good for URLs)
-            NullResult(StringExpressionType.ReplaceMissing.NULL_RESULT),
+            NullResult(StringExpressionType.ReplaceMissing.NULL_RESULT)
+                    {
+                        @Override
+                        public String handleNull(StringExpressionFactory.StringPart part) throws StopIteratingException
+                        {
+                            throw new StopIteratingException();
+                        }
+                    },
 
             // Null or missing fields get replaced with blank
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/util/StringExpressionFactory.java
```diff
@@ -693,6 +729,12 @@ public static AbstractStringExpression.NullValueBehavior fromXML(StringExpressio
                 return null;
             }
 
+            public abstract String handleNull(StringExpressionFactory.StringPart part) throws StopIteratingException;
+
+            public String handleUndefined(StringExpressionFactory.StringPart part) throws StopIteratingException
```
labkey-adam, line 734: Thanks for pushing these into the enum!

labkey-adam:
Would be nice to display the new properties on the Server Information page, in export diagnostics, etc. for consistency

labkey-adam (approved):

labkey-jeckels:
> Would be nice to display the new properties on the Server Information page, in export diagnostics, etc. for consistency

I added them to Server Info and exported diagnostics. I couldn't find any other spots that use the existing properties, so let me know if I missed anything to cover the "etc"

labkey-adam:
> > Would be nice to display the new properties on the Server Information page, in export diagnostics, etc. for consistency
> 
> I added them to Server Info and exported diagnostics. I couldn't find any other spots that use the existing properties, so let me know if I missed anything to cover the "etc"

Looks good... the other usages I was remembering appear to be covered by reflection-based map creation

PR #4755, opened 2023-09-19
Fix StudyLotsOfParticipantsTest.

labkey-danield:
#### Rationale
This test was out of date and had been failing for a long time. This should fix it and get it running again.

#### Related Pull Requests
* None.

#### Changes
* Do the correct steps to get the chart.


study/test/src/org/labkey/test/tests/study/StudyLotsOfParticipantsTest.java
```diff
@@ -107,6 +125,7 @@ public void testTimeChart()
         timeChartWizard.saveReport("Test time chart", "");
     }
 
+    // This method is not called, should it be deleted?
```
cnathe: yes, delete

cnathe (approved):

study/test/src/org/labkey/test/tests/study/StudyLotsOfParticipantsTest.java
labkey-danield: Deleted.

PR #4761, opened 2023-09-20
Automation test for TOTP 2FA integration with LabKey Server

labkey-sweta:
#### Rationale
Automation test for TOTP 2FA integration with LabKey Server

https://docs.google.com/document/d/1HvnsKeXBNhX1D5Tq7DkEnlN9m09GoJtkCj56oxhwQIA/edit

#### Related Pull Requests
https://github.com/LabKey/mfa/pull/4

#### Changes
*


devtools/test/src/org/labkey/test/params/devtools/SecondaryAuthenticationTOTPProvider.java
```diff
@@ -0,0 +1,33 @@
+package org.labkey.test.params.devtools;
```
labkey-tchad: You should be able to put this class in the `mfa` repository.
```suggestion
package org.labkey.test.params.mfa;
```

labkey-tchad (changes requested):

devtools/test/src/org/labkey/test/params/devtools/SecondaryAuthenticationTOTPProvider.java
labkey-sweta: I added this class here, because everything was part of devtools

Still makes sense to move?

labkey-tchad: Yes, it should move. `SecondaryAuthenticationProvider` is in here because the only secondary auth test we had previously was for the trivial one defined in the `devtools` module

labkey-tchad (changes requested):

labkey-tchad (changes requested):
I don't think any of these changes are needed.

PR #4764, opened 2023-09-20
use addHandlerForQuerySelector

labkey-matthewb:
#### Rationale
use addHandlerForQuerySelector when rendering helpPopup to reduce number or registered events.
48714

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/util/PageFlowUtil.java
+++ b/api/src/org/labkey/api/util/PageFlowUtil.java
@@ -156,6 +156,7 @@ import java.util.regex.Pattern;
 import java.util.zip.DeflaterOutputStream;
 import java.util.zip.InflaterInputStream;
 
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.apache.commons.lang3.StringUtils.startsWith;
 import static org.labkey.api.util.DOM.A;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/util/PageFlowUtil.java
```diff
@@ -1611,6 +1612,10 @@ public static HelpPopupBuilder popupHelp(@NotNull HtmlString helpHtml, String ti
     }
 
 
+    // NOTE: these are attached vial query selector, so they must be the same for every popup
```
labkey-martyp:
```suggestion
    // NOTE: these are attached via query selector, so they must be the same for every popup
```

labkey-martyp (approved):
Looks good. I did a quick test to make sure the popups were still working. And this cut the event handler count way down.

labkey-jeckels:
Would it be worthwhile to cap the number of handlers in `PageConfig` to something that's ridiculously high but enough to protect the server? 100,000? 1,000,000? We could log an error/warning.

labkey-matthewb:
> Would it be worthwhile to cap the number of handlers in `PageConfig` to something that's ridiculously high but enough to protect the server? 100,000? 1,000,000? We could log an error/warning.

Probably a good idea.

PR #4767, opened 2023-09-21
Issue 48598: already saved query metadata XML is overridden / removed on a second save

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48598

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/4767
- https://github.com/LabKey/testAutomation/pull/1643

#### Changes
* AssayProtocolSchema.createTable() to take an includeExtraMetadata param and check it before calling overlayMetadata()

### DIFF
--- a/api/src/org/labkey/api/query/UserSchema.java
+++ b/api/src/org/labkey/api/query/UserSchema.java
@@ -271,7 +271,7 @@ abstract public class UserSchema extends AbstractSchema implements MemTrackable
                 return table;
         }
         if (null == table)
-            table = createTable(name, cf);
+            table = createTable(name, cf, includeExtraMetadata);
         Object torq;
 
         if (table != null)
@@ -323,8 +323,12 @@ abstract public class UserSchema extends AbstractSchema implements MemTrackable
         return createTable(name, null);
     }
 
-    public abstract @Nullable TableInfo createTable(String name, ContainerFilter cf);
+    public @Nullable TableInfo createTable(String name, ContainerFilter cf, boolean includeExtraMetadata)
+    {
+        return createTable(name, cf);
+    }
 
+    public abstract @Nullable TableInfo createTable(String name, ContainerFilter cf);
 
     @Override
     abstract public Set<String> getTableNames();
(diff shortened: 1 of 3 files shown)


labkey-jeckels (commented):
This doesn't feel quite right. I believe that it fixes the original repro scenario, but I feel like it would ignore certain edits to the wrapped column made as part of the second save.

I think a more robust fix would be to propagate more of the underlying column's properties when we initialize the `rawColumnInfo` around line 185.

cnathe:
> This doesn't feel quite right. I believe that it fixes the original repro scenario, but I feel like it would ignore certain edits to the wrapped column made as part of the second save.
> 
> I think a more robust fix would be to propagate more of the underlying column's properties when we initialize the `rawColumnInfo` around line 185.

Thanks for the suggestion. Upon further review, this looks like an issue that was just affecting assay schema tables (or at least no a general issue affecting all tables). It looks like in the assay table case, there is an `overlayMetadata()` call happening in the `AssayProtocolSchema.createTable()` method. This then means that even though the `MetadataTableJSON.saveMetadata()` method is asking for the rawTableInfo without includeExtraMetadata, that assay table case really is applying the previously saved custom metadata....which then means that the next save looks like it should unset those things that haven't changed. I have a proposed fix that I'll push for a review.

api/src/org/labkey/api/query/UserSchema.java
```diff
@@ -323,8 +323,12 @@ protected void overlayTableMetadata(TableInfo table, String name, Collection<Que
         return createTable(name, null);
     }
 
-    public abstract @Nullable TableInfo createTable(String name, ContainerFilter cf);
+    public @Nullable TableInfo createTable(String name, ContainerFilter cf, boolean includeExtraMetadata)
```
labkey-jeckels, line 331: Adding a comment here for the intended use case would be nice. I think it's for subclasses that may apply metadata that comes from atypical locations (like the assay-provider-scoped example here)

labkey-jeckels (approved):

PR #4768, opened 2023-09-21
Issue 48678: Don't rely on folder type to determine if a feature is enabled

labkey-susanh:
#### Rationale
Issue [48678](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48678): We had been relying on a folder type check to determine if certain features in the application were enabled. This prevents the Sample Manager application from working well if someone just enables the requisite modules in a folder but does not change the folder type. This PR removes the check for the folder type altogether since we currently have nothing that prevents enabling the product modules from any type of folder.

It should be noted that the Biologics product will NOT work just because the module is enabled. There is logic within the `BiologicsFolderType::configureContainer` method that initializes registry and media types, and this won't get triggered when the module is just enabled. This was true before this change as well.

#### Related Pull Requests

- https://github.com/LabKey/sampleManagement/pull/2105
- https://github.com/LabKey/biologics/pull/2377

#### Changes
* Remove `FolderType.isProduct` method.


labkey-nicka (approved):

api/src/org/labkey/api/data/Container.java
```diff
@@ -39,6 +39,7 @@
 import org.labkey.api.pipeline.PipeRoot;
 import org.labkey.api.pipeline.PipelineService;
 import org.labkey.api.portal.ProjectUrls;
+import org.labkey.api.products.ProductRegistry;
```
labkey-nicka, line 42: Looks good. Not necessarily for this PR but does this seem like the kind of thing that would be good to cover with Junit integration tests? Seems like we could enumerate the many scenarios we're looking to support and help to ensure we don't regress with future changes.

api/src/org/labkey/api/data/Container.java
```diff
@@ -1650,40 +1652,40 @@ public boolean hasRestrictedActiveModule(Set<Module> activeModules)
...
-        boolean enabledAtProject = project.getFolderType().isProductFeatureEnabled(feature);
-        if (atProjectOnly || enabledAtProject)
-            return enabledAtProject;
-
-
-        return getFolderType().isProductFeatureEnabled(feature);
+        return AdminConsole.isProductFeatureEnabled(feature);
     }
 
     public boolean isFeatureEnabled(ProductFeature feature)
```
labkey-nicka, line 1669: Might be worth adding a description for when this method should be used. At first glance it seems kind of odd that `Container.isFeatureEnabled()` is different from `Container.isFeatureEnabledForContainer()`.

labkey-nicka (approved):

api/src/org/labkey/api/data/Container.java
labkey-susanh: I did add a couple in the sampleManagement module. Placement of the tests is a little tricky because of our reliance on modules to know which features are enabled.

api/src/org/labkey/api/data/Container.java
labkey-susanh: Yes, naming was difficult too. I'll add a comment that tries to help.

PR #4769, opened 2023-09-21
Expose key active usage data via new JMX beans

labkey-jeckels:
#### Rationale
Exposing key utilization data via JMX beans will let us monitor via Datadog and other tools, improving operational processes and our understanding of usage patterns

#### Changes
* Two new JMX beans that expose status for the search indexer and the number of active users, most recent activity, pipeline queue, and more
* Log admin-oriented warnings after being queried via JMX
* Support getting warnings outside of an HTTP request
* Minor code cleanup


search/src/org/labkey/search/SearchModule.java
```diff
@@ -133,6 +137,57 @@ public HttpView<?> getCustomSearchResult(User user, @NotNull String resourceIden
                 return null;
             }
         });
+
+        LabKeyManagement.register(new StandardMBean(new SearchMXBean()
```
labkey-matthewb: Why not have SearchService implement SearchMXBean?  
That approach could be used for PipelineService as well. 
I can see the use for OperationsMXBean, but also like the idea of services handling their own MXBean properties and methods.

labkey-matthewb (approved):

search/src/org/labkey/search/SearchModule.java
labkey-jeckels: Good idea. I started hacking on an implementation before really thinking about the scoping. I pushed updates

PR #4772, opened 2023-09-24
Provide execution plans for more queries

labkey-adam:
#### Rationale
We want the ability to produce execution plans on as many queries as possible

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48725

#### Changes
* Make `SELECT` queries that use CTEs (`WITH` statements) eligible for execution plans
* Make non-`SELECT` queries eligible for estimated execution plans, since they don't require query execution
* Skip block comments in addition to line comments when determining `SELECT` queries
* Bypass `SQLFragment` validation when rendering queries with their parameters
* `isSelect()` junit test


api/src/org/labkey/api/data/Table.java
```diff
@@ -160,34 +160,101 @@ public static void setParameters(PreparedStatement stmt, Collection<?> parameter
...
-        boolean select = false;
+        // Remove comments and trim
+        String strippedSql = new SqlScanner(sql).stripComments().toString().trim();
+
+        // If present, remove WITH statement, which should leave behind a SELECT
+        if (StringUtils.startsWithIgnoreCase(strippedSql, "WITH"))
+        {
+            SqlScanner scanner = new SqlScanner(strippedSql);
+            SkipWithStatementHandler handler = new SkipWithStatementHandler();
+            scanner.scan(4, handler);
```
labkey-matthewb: I sometimes do "WITH".length() for self-documenting.

api/src/org/labkey/api/collections/CaseInsensitiveHashSet.java
```diff
@@ -115,6 +115,10 @@ public static Set<String> of(String s1, String s2, String s3, String s4, String
         return unmodifiableSet(new CaseInsensitiveHashSet(s1, s2, s3, s4, s5));
     }
 
+    public static Set<String> of(String s1, String s2, String s3, String s4, String s5, String s6)
```
labkey-jeckels: Any reason we can't use var-args here instead?

api/src/org/labkey/api/data/Table.java
```diff
@@ -160,34 +160,101 @@ public static void setParameters(PreparedStatement stmt, Collection<?> parameter
...
+        public void test()
+        {
+            assertTrue(isSelect("SELECT * FROM foo.bar"));
+            assertTrue(isSelect("WITH Blue AS (SELECT * FROM foo.bar) SELECT * FROM Blue"));
+            assertTrue(isSelect("WITH cte AS (SELECT CAST('1' AS INTEGER) AS i) SELECT * FROM cte"));
+
+            assertFalse(isSelect("WITH Nothing SELECT * FROM foo.bar"));
+            assertFalse(isSelect("WITH Mismatched AS (SELECT * FROM (SELECT * FROM foo.bar) SELECT * FROM Mismatched"));
+            assertFalse(isSelect("WITH Incomplete AS (SELECT * FROM foo.bar)"));
+        }
```
labkey-matthewb, line 261: Maybe a non-SELECT case?  WITH cte AS () UPDATE

labkey-matthewb (approved):

api/src/org/labkey/api/data/Table.java
```diff
@@ -160,34 +160,101 @@ public static void setParameters(PreparedStatement stmt, Collection<?> parameter
...
+    public static class IsSelectTestCase extends Assert
+    {
+        @Test
+        public void test()
+        {
+            assertTrue(isSelect("SELECT * FROM foo.bar"));
+            assertTrue(isSelect("WITH Blue AS (SELECT * FROM foo.bar) SELECT * FROM Blue"));
+            assertTrue(isSelect("WITH cte AS (SELECT CAST('1' AS INTEGER) AS i) SELECT * FROM cte"));
+
+            assertFalse(isSelect("WITH Nothing SELECT * FROM foo.bar"));
```
labkey-jeckels, line 255: Let's throw in an `UPDATE` or `DELETE` that uses a CTE too

labkey-jeckels (approved):

api/src/org/labkey/api/data/Table.java
labkey-matthewb: Really "WITH\s", but I can't imagine where this would matter (unless WITHOUT is a keyword)

labkey-adam: Addressed both of these... and we're now looking for "SELECT\s" and "\sFROM\s" as well

api/src/org/labkey/api/data/Table.java
labkey-adam: Done

api/src/org/labkey/api/collections/CaseInsensitiveHashSet.java
labkey-adam: Because they hadn't been invented in 2020 when Kevin added these factory methods?

labkey-adam: Addressed

api/src/org/labkey/api/data/Table.java
labkey-adam: Added

PR #4776, opened 2023-09-25
User count metrics for site roles and special groups

labkey-jeckels:
#### Rationale
Knowing the numbers of users assigned to these site-wide roles and groups will help understand how they are used, and monitor for unusual assignments.

#### Changes
* New UsageMetricsProvider to capture user counts in site-wide roles and the special site admin and developer groups
* Minor code cleanup

### DIFF
--- a/api/src/org/labkey/api/security/SecurityManager.java
+++ b/api/src/org/labkey/api/security/SecurityManager.java
@@ -17,6 +17,7 @@
 package org.labkey.api.security;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.collections4.bag.HashBag;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.Logger;
@@ -95,7 +96,6 @@ import org.labkey.api.util.Pair;
 import org.labkey.api.util.SessionHelper;
 import org.labkey.api.util.TestContext;
 import org.labkey.api.util.URLHelper;
-import org.labkey.api.util.emailTemplate.EmailTemplate;
 import org.labkey.api.util.emailTemplate.EmailTemplateService;
 import org.labkey.api.util.emailTemplate.UserOriginatedEmailTemplate;
 import org.labkey.api.util.logging.LogHelper;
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -182,6 +183,42 @@ public class SecurityManager
...
+                if (siteRoles.contains(role))
+                {
+                    int count = roleCounts.getOrDefault(role.getName(), 0);
+                    UserPrincipal principal = SecurityManager.getPrincipal(assignment.getUserId());
+                    if (principal != null && principal.isActive())
+                    {
+                        switch (principal.getPrincipalType())
+                        {
+                            case USER -> count++;
+                            case GROUP -> getAllGroupMembers((Group)principal, MemberType.ACTIVE_USERS);
```
labkey-adam: Seems like a no-op. Don't you want to do something like `count += getAllGroupMembers().size()` here?

api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -182,6 +183,42 @@ public class SecurityManager
         EmailTemplateService.get().registerTemplate(PasswordResetAdminEmailTemplate.class);
     }
 
+    public static UsageMetricsProvider getMetricsProvider()
+    {
+        return () -> {
+            Map<String, Object> result = new HashMap<>();
+
+            Map<String, Integer> roleCounts = new HashMap<>();
```
labkey-adam, line 191: Optional: using a `HashBag` might be cleaner. It handles the initialization and accumulation, and provides `add(value, count)`.

labkey-adam (changes requested):
Unless I'm misreading, the metrics provider group handling needs to be adjusted

api/src/org/labkey/api/security/SecurityManager.java
labkey-jeckels: Indeed

api/src/org/labkey/api/security/SecurityManager.java
labkey-jeckels: I didn't see an easy way to get the role name -> count map from it so I left as-is

labkey-adam (approved):

PR #4778, opened 2023-09-25
Cap the total number of JS handlers we're tracking per page

labkey-jeckels:
#### Rationale
We store and later render all inline JavaScript handlers to support stronger Content Security Policy directives. Some page renders, especially huge grids, may produce enough of them to overwhelm the heap. See issue 48714

#### Changes
* Don't track more than 100,000 handlers
* Clarify reason for not preloading fonts

### DIFF
--- a/api/src/org/labkey/api/view/template/PageConfig.java
+++ b/api/src/org/labkey/api/view/template/PageConfig.java
@@ -601,7 +601,7 @@ public class PageConfig
         {
             if (_eventHandlers.size() == MAX_EVENT_HANDLERS)
             {
-                LOG.warn("Limit of " + MAX_EVENT_HANDLERS + " JavaScript event handlers reached. Subsequent handlers will be dropped. Current handler for " + eh.event + ": " + eh.handler);
+                LOG.error("Limit of " + MAX_EVENT_HANDLERS + " JavaScript event handlers reached. Subsequent handlers will be dropped. Current handler for " + eh.event + ": " + eh.handler);
             }
             var prev = _eventHandlers.put(eh.getKey(), eh);
             assert null == prev || prev.handler.equals(eh.handler) : "Duplicate handler registered. event:" + eh.getKey() + " handler:" + eh.handler();
--- a/api/src/org/labkey/api/view/template/PageConfig.java
+++ b/api/src/org/labkey/api/view/template/PageConfig.java
@@ -113,9 +113,13 @@ public class PageConfig
     }
 
     // collected javascript handlers
-    private final LinkedHashMap<String,EventHandler> _eventHandlers = new LinkedHashMap<>();
-    private final ArrayList<String> _onDomLoaded = new ArrayList<>();
-    private final ArrayList<String> _onDocumentLoaded = new ArrayList<>();
+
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/view/template/PageConfig.java
```diff
@@ -584,8 +597,15 @@ public HtmlString getScriptTagStart()
...
-        assert null==prev || prev.handler.equals(eh.handler) : "Duplicate handler registered. event:" + eh.getKey() + " handler:" + eh.handler();
+        if (_eventHandlers.size() <= MAX_EVENT_HANDLERS)
+        {
+            if (_eventHandlers.size() == MAX_EVENT_HANDLERS)
+            {
+                LOG.warn("Limit of " + MAX_EVENT_HANDLERS + " JavaScript event handlers reached. Subsequent handlers will be dropped. Current handler for " + eh.event + ": " + eh.handler);
+            }
+            var prev = _eventHandlers.put(eh.getKey(), eh);
+            assert null == prev || prev.handler.equals(eh.handler) : "Duplicate handler registered. event:" + eh.getKey() + " handler:" + eh.handler();
+        }
```
labkey-matthewb, line 608: My only concern is that this could be a very confusing support call.  Should we add a document.onload event to show the page banner or something?  Or make it an error in the log?  (Trying to think what would help the AM that gets this ticket).

labkey-matthewb (changes requested):

api/src/org/labkey/api/view/template/PageConfig.java
```diff
@@ -420,8 +424,17 @@ private String getCanonicalLink(URLHelper current)
...
     {
-        final List<String> fonts = List.of();
+        // Preloading is sometimes great but is a waste if the resource isn't actually used in a way
+        // that impacts the initial page render. Need to refine the list to just the ones that will be used
+        // above the fold if we want to reintroduce this. Largely moot after the initial render due to caching.
+        final List<String> fonts = List.of(
+//                "/fonts/Roboto/Roboto-Regular.ttf",
+//                "/fonts/TitilliumWeb/TitilliumWeb-Regular.ttf",
+//                "/fonts/TitilliumWeb/TitilliumWeb-Bold.ttf",
+//                "/fonts/Roboto/Roboto-Bold.ttf"
```
labkey-martyp, line 434: Are these commented out on purpose?

labkey-martyp (approved):

api/src/org/labkey/api/view/template/PageConfig.java
labkey-jeckels: Yes. They're the ones we used to always send, and the ones to consider reintroducing per the comment. They don't need to there in comments, but I thought it might be useful to not need to dive into the history to find them

api/src/org/labkey/api/view/template/PageConfig.java
labkey-jeckels: Switched to error level

labkey-matthewb (approved):

labkey-martyp (approved):

PR #4784, opened 2023-09-26
Don't include cpastype filter when it is redundant

labkey-matthewb:
#### Rationale
Don't include cpastype filter when it is redundant.  This may help the postgres optimizer.  We'll see.

#### Related Pull Requests
* 

#### Changes
*


experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -532,8 +532,15 @@ protected ColumnInfo getPkColumn(TableInfo table)
         return ret;
     }
 
-    @Override
+    @Deprecated
```
labkey-susanh, line 535: Should probably add a comment about what to use instead and why.

labkey-susanh (approved):

PR #4791, opened 2023-09-28
Exempt API Key authentications from 2FA

labkey-adam:
#### Rationale
Related PR made API key use a first class authentication mechanism which meant, among other things, they are subject to two-factor authentication mechanisms like TOTP and Duo. This PR removes the 2FA requirement when API keys are used to authenticate.

https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=48764

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4751

### DIFF
--- a/api/src/org/labkey/api/security/SecurityManager.java
+++ b/api/src/org/labkey/api/security/SecurityManager.java
@@ -17,7 +17,6 @@
 package org.labkey.api.security;
 
 import org.apache.commons.codec.binary.Base64;
-import org.apache.commons.collections4.bag.HashBag;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -17,7 +17,6 @@
 package org.labkey.api.security;
 
 import org.apache.commons.codec.binary.Base64;
-import org.apache.commons.collections4.bag.HashBag;
```
labkey-jeckels, line 20: And I thought you liked HashBags

labkey-jeckels (approved):

PR #4795, opened 2023-09-29
Issue 48773: diagnostic export too slow for clients with giant file roots

labkey-jeckels:
#### Rationale
Servers with large or high-latency file systems as their file root can take many minutes to recursively call them to get the total file size. This is packaged into the diagnostics export, and when submitting usage metrics.

#### Changes
* Time out after 20 seconds of crawling
* Count the files visited
* Track whether we completed or not

### DIFF
--- a/api/src/org/labkey/api/data/StopIteratingRuntimeException.java
+++ b/api/src/org/labkey/api/data/StopIteratingRuntimeException.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) 2016 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.labkey.api.data;
+
+/**
+ * Can be thrown from stream operations, iterators, etc to signal that we reached our limit.
+ */
+public class StopIteratingRuntimeException extends RuntimeException
+{
+}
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/data/StopIteratingRuntimeException.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.labkey.api.data;
+
+/**
+ * Can be thrown from stream operations, iterators, etc to signal that we reached our limit.
+ */
+public class StopIteratingRuntimeException extends RuntimeException
```
labkey-adam, line 21: We already have a `StopIteratingException`, though it's checked. Your call if you want to use the existing one or create a new one.

filecontent/src/org/labkey/filecontent/FileContentModule.java
```diff
@@ -184,9 +190,38 @@ public void doStartup(ModuleContext moduleContext)
...
-                results.put("fileRootSize", FileUtils.sizeOfDirectory(root));
-                results.put("fileRootMillisecondsToCalculateSize", System.currentTimeMillis() - currentTime);
+                long startTime = HeartBeat.currentTimeMillis();
+                MutableLong totalSize = new MutableLong(0);
+                MutableLong fileCount = new MutableLong(0);
+                boolean timedOut = false;
+                boolean succeeded = true;
+                try
+                {
+                    FileUtils.streamFiles(root, true, (String[])null).forEach(f ->
```
labkey-adam: I believe you need to call this with try-with-resources. Sadly, `FileUtils.streamFiles()` doesn't document this nor does `PathUtils.walk()` (which it calls), but `Files.walk()` (called by PathUtils) states: "This method must be used within a try-with-resources statement or similar control structure to ensure that the stream's open directories are closed promptly after the stream's operations have completed."

labkey-adam (changes requested):
Please close the Stream and consider using the existing exception

filecontent/src/org/labkey/filecontent/FileContentModule.java
labkey-jeckels: Thank you. I remember a leak in a similar call. Fixed.

api/src/org/labkey/api/data/StopIteratingRuntimeException.java
labkey-jeckels: I can't throw a checked exception from inside of the Consumer lambda, which is why I created the new one.

labkey-adam: Yes, of course

labkey-adam (approved):

PR #4798, opened 2023-10-02
Issue 48118: Always display an aliquot naming pattern

labkey-susanh:
#### Rationale
[Issue 48118](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48118): There's always a naming pattern for aliquots, so we want to always display one. New sample types now save the default naming pattern value in the database. The upgrade script here sets the naming pattern for sample types created before this change was made.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-premium/pull/138

#### Changes
* Upgrade script to populate the aliquotNameExpression for all sample types that don't have it set.

### DIFF
--- a/experiment/resources/schemas/dbscripts/postgresql/exp-23.010-23.011.sql
+++ b/experiment/resources/schemas/dbscripts/postgresql/exp-23.010-23.011.sql
@@ -1,0 +1,3 @@
+UPDATE exp.materialsource
+    SET aliquotnameexpression = '${${AliquotedFrom}-:withCounter}'
+    WHERE aliquotnameexpression IS NULL;
--- a/experiment/resources/schemas/dbscripts/sqlserver/exp-23.010-23.011.sql
+++ b/experiment/resources/schemas/dbscripts/sqlserver/exp-23.010-23.011.sql
@@ -1,0 +1,3 @@
+UPDATE exp.materialsource
+    SET aliquotnameexpression = '${${AliquotedFrom}-:withCounter}'
+    WHERE aliquotnameexpression IS NULL;
--- a/experiment/src/org/labkey/experiment/ExperimentModule.java
+++ b/experiment/src/org/labkey/experiment/ExperimentModule.java
@@ -175,7 +175,7 @@ public class ExperimentModule extends SpringModule implements SearchService.Docu
     @Override
     public Double getSchemaVersion()
     {
-        return 23.010;
+        return 23.011;
     }
 
     @Nullable


XingY (changes requested):
For the purpose of fixing the scenario that details view displays empty but designer view displays the default aliquot pattern, I don't think this change works. If a user manually delete the aliquot pattern and save, the issue will surface again. The NameGenerator will always use default aliquot pattern if no one is present on the db, so I think the change should be on the UI instead of db. Otherwise, we need to change aliquot generation to require the presence of a pattern or explicit name.

labkey-susanh:
> For the purpose of fixing the scenario that details view displays empty but designer view displays the default aliquot pattern, I don't think this change works. If a user manually delete the aliquot pattern and save, the issue will surface again. The NameGenerator will always use default aliquot pattern if no one is present on the db, so I think the change should be on the UI instead of db. Otherwise, we need to change aliquot generation to require the presence of a pattern or explicit name.

I considered updating the UI, but it's quite a hack. When there's no aliquot naming pattern, we also won't pull through the prefix for the aliquot name in the UI, so we'd have to hack something there as well.  We don't actually require the naming pattern, but effectively make it difficult to not have one explicitly. When you go to the domain editor, we'll populate the default naming pattern if one doesn't exist, so if you save the domain, the default will be saved if you don't deliberately remove it at that time.  

This change seems like it will handle the regular use cases without much fuss. If someone deliberately deletes the aliquot naming pattern, they are presumably not planning to create aliquots and so don't want to be bothered with seeing this preview? If they do create aliquots, then they'll get the default pattern and that seems as it should be.  Because this is a fairly low-priority (in my opinion) update just to make it more clear how names are generated for aliquots by default, I'm opting for a small-effort delta.

XingY:
> > For the purpose of fixing the scenario that details view displays empty but designer view displays the default aliquot pattern, I don't think this change works. If a user manually delete the aliquot pattern and save, the issue will surface again. The NameGenerator will always use default aliquot pattern if no one is present on the db, so I think the change should be on the UI instead of db. Otherwise, we need to change aliquot generation to require the presence of a pattern or explicit name.
> 
> I considered updating the UI, but it's quite a hack. When there's no aliquot naming pattern, we also won't pull through the prefix for the aliquot name in the UI, so we'd have to hack something there as well. We don't actually require the naming pattern, but effectively make it difficult to not have one explicitly. When you go to the domain editor, we'll populate the default naming pattern if one doesn't exist, so if you save the domain, the default will be saved if you don't deliberately remove it at that time.
> 
> This change seems like it will handle the regular use cases without much fuss. If someone deliberately deletes the aliquot naming pattern, they are presumably not planning to create aliquots and so don't want to be bothered with seeing this preview? If they do create aliquots, then they'll get the default pattern and that seems as it should be. Because this is a fairly low-priority (in my opinion) update just to make it more clear how names are generated for aliquots by default, I'm opting for a small-effort delta.

For LKS, aliquot pattern is not surfaced on the designer. So for shared types or types created in LKS but used in LKSM, there would be similar display discrepancy.

labkey-susanh:
After chatting with Hannah, we're opting to leave the current behavior as it is. Users can update existing sample types to save the aliquot naming pattern in order to see the value displayed. Any new sample types created in the app already have the default naming pattern populated, so will get saved with a value unless users deliberately decide to remove that.

If we find there is confusion here, we can consider what makes the most sense to provide consistency in the handling of this naming pattern for LKS sample types and app sample types and with the handling of the ID naming pattern.

PR #4809, opened 2023-10-05
Issue 48331: Find by barcode performance

XingY:
#### Rationale
Barcode fields are stored in the provisioned sample table. When finding sample by barcode, the material table is joined to the provisioned table (as well as other tables such as user, container, inventory.item) to filter by barcode. This is performed per sample type per barcode field. Since barcode and lsid are both available on provisioned table, it's more performant to query for lsid from the provisioned table without any joins, and then query for other sample fields using the matched lsids.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/207
- https://github.com/LabKey/platform/pull/4809
- https://github.com/LabKey/sampleManagement/pull/2142
- https://github.com/LabKey/biologics/pull/2417

#### Changes
* modify SaveOrderedSamplesQueryAction to first query for lsid using barcode from provisioned table only, then filter samples table by lsids


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -9089,6 +9089,62 @@ public int moveAuditEvents(Container targetContainer, List<String> runLsids)
...
+
+        for (ExpSampleTypeImpl type : sampleTypes)
+        {
+            TableInfo provisioned = type.getTinfo();
+            TableInfo tableInfo = samplesUserSchema.getTable(type.getName());
+            if (tableInfo == null)
+                continue;
+            List<ColumnInfo> uniqueIdCols = tableInfo.getColumns().stream().filter(ColumnInfo::isScannableField).collect(Collectors.toList());
+            numUniqueIdCols += uniqueIdCols.size();
+            for (ColumnInfo col : uniqueIdCols)
```
labkey-matthewb, line 9117: If the uniqueId columns are not indexed, it may be faster to do an OR rather than UNION.  The entire expression can be done in one pass of the table.  With indexes, this SELECT per column makes sense.

XingY: Interesting thought...
Here I need to get not just the list of lsid from the list of barcode, but also the 1:1 mapping of the lsid and the barcode. In order to do that with OR, I probably need to return a concatenated allBarcodes in the result, which will later get parsed and filtered based on the original set of barcodes. This probably won't add too much cost. But my understanding is most sample types have at most one barcode fields, so it's not very clear OR approach would benefit in most scenarios.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -7270,21 +7282,26 @@ private SQLFragment getOrderedRowsSql()
                     sampleIdValuesSql.append(sampleIdComma).append("\t(").appendValue(index);
                     sampleIdValuesSql.append(", ");
                     sampleIdValuesSql.append(LabKeySql.quoteString(id.substring(SAMPLE_ID_PREFIX.length())));
+                    sampleIdValuesSql.append(", ");
+                    sampleIdValuesSql.append(LabKeySql.quoteString("null"));
```
labkey-matthewb, line 7286: Is this really supposed to be the string **"null"** rather than the literal **null**?

labkey-matthewb (approved):
I really don't know entirely what the use case is here, but this seems like a good optimization.  My only question is how big can this request get?  Is it worth stashing the getUniqueIdLsids() result in a temp table, rather than copying down to the webserver and then writing back into a VALUES clause?

PR #4811, opened 2023-10-07
Add Impersonating Troubleshooter role

labkey-adam:
#### Rationale
An Impersonating Troubleshooter normally acts as a Troubleshooter, but can impersonate site roles, including Site Administrator.

https://docs.google.com/document/d/1aXDfb7_E7WnyOKOPXactMzheOir9T-hmgLxBRQl7xJA/edit

#### Changes
* Add `ImpersonatingTroubleshooterRole` and associated permission
* Designate Site Admin, Platform Developer, and Impersonating Troubleshooter as "privileged" roles
* Adjust permission checks to allow Impersonating Troubleshooters to impersonate privileged roles
* Provide Troubleshooters with:
  * View admin warnings
  * Monitor upgrades via the web UI
  * Adjust and reset mini profiler settings
  * View module errors
* Revamp "ensure at least one site admin" handling
* Remove Application Admins' ability to impersonate privileged roles
* Implement consistent restrictions on privileged roles
* Lots of cleanup


api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -1497,19 +1522,47 @@ public static void deleteMembers(Group group, List<UserPrincipal> membersToDelet
...
-        new SqlExecutor(core.getSchema()).execute("DELETE FROM " + core.getTableInfoMembers() + "\n" +
-            "WHERE GroupId = ? AND UserId = ?", groupId, principal.getUserId());
-        fireDeletePrincipalFromGroup(groupId, principal);
+        deleteMembers(group, Set.of(principal));
+    }
+
+    /**
+     * Throws if the site current has no Site Admins and no Impersonating Troubleshooters
+     */
+    public static void ensureAtLeastOneSiteAdminExists()
```
labkey-jeckels, line 1547: This method name isn't really accurate, but I don't have a good suggestion that's not even more verbose. Thanks for adding the comment to describe its behavior

api/src/org/labkey/api/security/SecurityPolicyManager.java
```diff
@@ -238,11 +235,21 @@ private static void savePolicyToDBAndValidate(@NotNull MutableSecurityPolicy pol
                 Table.insert(null, table, assignment);
             }
 
-            //commit transaction
+            // Ensure at least one site admin will remain if attempting to modify the root container's policy
+            if (policy.getResourceId().equals(ContainerManager.getRoot().getResourceId()))
```
labkey-jeckels, line 245: Do we not need the post-commit task here too?

api/src/org/labkey/api/security/UserManager.java
```diff
@@ -1026,11 +1038,20 @@ public static void setUserActive(User currentUser, User userToAdjust, boolean ac
             // Call update unconditionally to ensure Modified & ModifiedBy are always updated
             Table.update(currentUser, CoreSchema.getInstance().getTableInfoUsers(), map, userId);
 
+            // If deactivating a site admin or impersonating troubleshooter, ensure at least one site admin remains
+            if (!active && userToAdjust.hasRootPermission(CanImpersonatePrivilegedSiteRolesPermission.class))
```
labkey-jeckels, line 1042: Similar question - no need for post-commit tasks?

core/src/org/labkey/core/admin/miniprofiler/MiniProfilerController.java
```diff
@@ -174,7 +163,7 @@ public URLHelper getSuccessURL(Object o)
 
     // Invoked by test framework
     @RequiresSiteAdmin
```
labkey-jeckels: Any reason to not move these to TroubleshooterPermission too for consistency?

labkey-jeckels (approved):

api/src/org/labkey/api/security/SecurityManager.java
labkey-adam: I debated `ensureAtLeastOneSiteAdminLikeUserExists()` but decided just to add the comment

api/src/org/labkey/api/security/SecurityPolicyManager.java
labkey-adam: I don't think so, but please check my logic. We need to clear caches in the rollback case, the non-root case, and even the successful ensure case (small chance of a race before commit). At first, I added a commit task on post-commit & rollback, but switched it to a `finally()` block since that seems effectively identical but simpler.

api/src/org/labkey/api/security/UserManager.java
labkey-adam: I don't think so... see `finally()`

core/src/org/labkey/core/admin/miniprofiler/MiniProfilerController.java
labkey-adam: Nope, that seems fine

api/src/org/labkey/api/security/SecurityPolicyManager.java
labkey-jeckels: It should be functionally equivalent, but only if isn't an outer transaction pending.

api/src/org/labkey/api/security/UserManager.java
labkey-jeckels: Ah, I see. It hadn't changed so it was out of view in the diff.

Like above, it's fine as long as we're not worried about outer transaction wrapping this method.

labkey-adam: I can't guarantee that it'll never be wrapped by an outer, so I switched to commit tasks to be safe

api/src/org/labkey/api/security/SecurityPolicyManager.java
labkey-adam: Switched to commit tasks

labkey-jeckels (approved):

PR #4813, opened 2023-10-09
Changes to Short URL admin page

vagisha:
#### Rationale
panoramaweb.org has over 1K short URLs.  The admin page renders all existing short URLs in a JSP page.  We would like replace this with a LabKey grid to enable filtering and sorting.  This should also speed up the time taken to fully load the page on panoramaweb.org

#### Changes
Replaced the JSP page that lists all the short URLs with a LabKey grid.   Added a ShortUrlTableInfo class.  The shorturl table is not exposed through the query schema browser.


core/src/org/labkey/core/query/CoreQuerySchema.java
```diff
@@ -177,6 +178,9 @@ public TableInfo createTable(String name, ContainerFilter cf)
             return new ViewCategoryTable(ViewCategoryManager.getInstance().getTableInfoCategories(), this, cf);
         if (MISSING_VALUE_INDICATOR_TABLE_NAME.equalsIgnoreCase(name))
             return getMVIndicatorTable(cf);
+        if (SHORTURL_TABLE_NAME.equalsIgnoreCase(name))
```
labkey-jeckels: While the schema isn't advertising this as an available table, it will happily serve up the table by name if asked. Short URLs are scoped site-wide so they don't have a container. This means that anyone with read access to any container will be able to get the full list.

I don't think that would likely be a problem, but we also don't need to be that permissive. I suggest checking if the container is the root and that the user has AdminOperationsPermission before returning it.

labkey-jeckels (changes requested):

core/src/org/labkey/core/query/ShortUrlTableInfo.java
```diff
@@ -0,0 +1,145 @@
...
+    private static final String UPDATE_SHORT_URL_COL = "UpdateShortUrl";
+    private static final String TO_CLIPBOARD_COL = "CopyToClipboard";
+
+    public ShortUrlTableInfo(@NotNull CoreQuerySchema userSchema)
+    {
+        super(CoreSchema.getInstance().getTableInfoShortURL(), userSchema);
+        wrapAllColumns(true);
+        setInsertURL(LINK_DISABLER);
+        setImportURL(LINK_DISABLER);
+        setDeleteURL(LINK_DISABLER);
```
labkey-jeckels: Can we switch to a more standard update and delete pattern with this grid change? Update should be easy - point at the action. Delete's probably not much harder. That would also clean up the UI a bit by reducing the number of links per row.

core/src/org/labkey/core/query/CoreQuerySchema.java
vagisha: Done.

core/src/org/labkey/core/query/ShortUrlTableInfo.java
vagisha: Done

labkey-jeckels:
FB mirrored in the LK repo

https://github.com/LabKey/platform/tree/23.7_fb_shorturl-admin

I also queued a run for the Panorama Public tests. I know that workflow uses short URLs but it may not go through the admin page.

https://teamcity.labkey.org/buildConfiguration/LabKey_237Release_Community_ModuleSuites_MacCossLabModulesPostgres/2681154?hideProblemsFromDependencies=false&hideTestsFromDependencies=false

labkey-jeckels (approved):

PR #4814, opened 2023-10-09
Validate values when saving/loading DB login properties

labkey-tchad:
#### Rationale
I accidentally flipped the strength and expiration values when saving settings via API. This made my server unable to start up. Adding validation to save and falling back to `Strong`/`Never` if invalid values are found in the database.

#### Related Pull Requests
* N/A

#### Changes
* Implementing `SaveDbLoginPropertiesAction.validateForm` to prevent saving invalid values
* Use default rule (`Strong`) or expiration (`Never`) if invalid value(s) are found in database


labkey-adam (approved):
I'm fine with this as is. Just a thought that using `EnumUtils.getEnum()` with the default option in place of `stringProperties.getOrDefault()` in the original code would address this in a much more succinct way. The downside would be no logging in the invalid case. Up to you.

PR #4815, opened 2023-10-09
Materialized SampleType table

labkey-matthewb:
#### Rationale
Materialize SampleType table into a temp table.  Possibly helping the query execution engine do less repeated work and the query optimizer generate better execution plans.

#### Related Pull Requests
* 

#### Changes
*


experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -2177,4 +2177,9 @@ public long getCurrentCount(NameGenerator.EntityCounter counterType, Container c
 
         return getProjectSampleCount(container, counterType == NameGenerator.EntityCounter.rootSampleCount);
     }
+
+    void refreshSampleTypeMaterializedView(@NotNull ExpSampleType st, boolean schemaChange)
+    {
+        ExpMaterialTableImpl.refreshMaterializedView(st.getLSID(), schemaChange);
```
XingY, line 2187: SampleTypeUpdateServiceDI.handleRecalc can happen in a background thread independent from the main transaction, depending on the size of the data that needs recalc. Seems like the refresh of the materialized view might happen before recalc is complete.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -1044,8 +1023,130 @@ public SQLFragment getFromSQL(String alias)
...
+                new AtomicLong(System.currentTimeMillis())
+        );
+    }
+
+
+    /* SELECT and JOIN, does not include WHERE, same as get`JoinSQL() */
+    private SQLFragment getMaterializedSQL()
+    {
+        if (null == _ss)
+            return getJoinSQL(null);
```
XingY, line 1127: This is probably a silly question... Is PHI handled the same on materialized view vs the original table?

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -1044,8 +1023,130 @@ public SQLFragment getFromSQL(String alias)
...
+            return new MaterializedQueryHelper.Builder("", getExpSchema().getDbSchema().getScope(), viewSql)
+                .addIndex("CREATE UNIQUE INDEX uq_${NAME}_rowid ON temp.${NAME} (rowid)")
+                .addIndex("CREATE UNIQUE INDEX uq_${NAME}_lsid ON temp.${NAME} (lsid)")
+                .addIndex("CREATE INDEX idx_${NAME}_container ON temp.${NAME} (container)")
+                .addIndex("CREATE INDEX idx_${NAME}_root ON temp.${NAME} (rootmateriallsid)")
+                .addInvalidCheck(() -> String.valueOf(getInvalidateCounter(_ss.getLSID()).get()))
+                .upToDateSql(new SQLFragment("SELECT CAST(COUNT(*) AS VARCHAR) FROM ").append(provisioned))  // MAX(modified) would probably be better if it were a) in the materailized table b) indexed
+                .build();
+        });
+        return new SQLFragment("SELECT * FROM ").append(mqh.getFromSql("_cached_view_"));
```
XingY, line 1148: is the materialized view over all records in all folders?

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
labkey-matthewb: Not a silly question!  PHI handling is handled completely in metadata.  For instance, which columns are made available in the TableInfo, or are replaced with NULL columns.  This should not affect PHI.  But worth double checking.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
labkey-matthewb: Yes, I decided to materialize the entire SampleType across folders.  That means the we don't need to do anything tricky for varying container filters.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
labkey-matthewb: NOTE: if we use incremental updates, using a container for scoping updates could be relevant.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-matthewb: Thank you, I will look at this.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -1056,21 +1157,19 @@ public SQLFragment getFromSQL(String alias, Set<FieldKey> selectedColumns)
...
         selectedColumns = computeInnerSelectedColumns(selectedColumns);
 
         SQLFragment sql = new SQLFragment();
-        sql.appendComment("<ExpMaterialTableImpl.getFromSQL()>", getSqlDialect());
-        sql.append("(SELECT ");
+        sql.appendComment("<ExpMaterialTableImpl.getJoinSQL(" + (null==_ss ? "" : _ss.getName()) + ")>", getSqlDialect());
+        sql.append("SELECT ");
         String comma = "";
-        for (String dataCol : dataCols)
+        for (String naterialCol : materialCols)
```
```suggestion
        for (String materialCol : materialCols)

```

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Does this case get handled by the removal of the cache asa post-commit task?

labkey-susanh (commented):
Seems OK to me, but I'm going to defer to Xing.

labkey-susanh (approved):
Since this is still behind an experimental feature flag, I'm going to approve so we can have this in the mix for 23.11

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
XingY: The removal of cache is only done when the schema is changed? I'm still not understanding how recalc is going to work with this

labkey-matthewb: This deserves a code comment.  Thanks.  MaterializedQueryHelper takes a fixed SQL to define the materialized view.  The "helper" part is that it can detect changes to the data and rematerialize the data. This has worked really well in previous usages which used non-user tables.

This usage involves provisioned tables so the schema can change.  In this case, we toss the MQH so that we will regenerate it with new SQL to account for new/removed/renamed columns.

labkey-matthewb: > SampleTypeUpdateServiceDI.handleRecalc can happen in a background thread independent from the main transaction, depending on the size of the data that needs recalc. Seems like the refresh of the materialized view might happen before recalc is complete.

@XingY can recomputeSampleTypeRollup() affect multiple SampleTypes or just the one that is passed in?

XingY: recompute is done for one sampletype

XingY: > This deserves a code comment. Thanks. MaterializedQueryHelper takes a fixed SQL to define the materialized view. The "helper" part is that it can detect changes to the data and rematerialize the data. This has worked really well in previous usages which used non-user tables.
> 
> This usage involves provisioned tables so the schema can change. In this case, we toss the MQH so that we will regenerate it with new SQL to account for new/removed/renamed columns.

Re "it can detect changes to the data", does it detect changes to any column? Can it bypass columns such as lastIndexed?

labkey-matthewb: Thanks

labkey-matthewb: Whoever constructs the MQH, provides callbacks for this purpose.  In this example, (for the moment at least) there are two callbacks provided.

                .addInvalidCheck(() -> String.valueOf(getInvalidateCounter(_ss.getLSID()).get()))
                .upToDateSql(new SQLFragment("SELECT CAST(COUNT(*) AS VARCHAR) FROM ").append(provisioned))  // MAX(modified) would probably be better if it were a) in the materailized table b) indexed

The first checks that an in-memory counter has not changed.  This is the primary mechanism.  refreshMaterializedView() bumps up this counter, so the MQH will consider any currently materialized table as being no longer up-to-date.  The other is just a count of rows in the provisioned table.  This is generally very fast to compute as the PK index can be inspected to find this count.  (Using MAX(modified) would be interesting too, but we don't have an index on modified).

PR #4821, opened 2023-10-13
Issue 48501: MultiValuedDisplayColumn doesn't handle rows with mismatched null/non-nulls

labkey-jeckels:
#### Rationale
Multi valued lookups get very confused when the child rows have columns with different numbers of nulls. Separately but relatedly, our SQL Server implementation didn't handle CTEs in the underlying SQL

#### Changes
* Handle nulls for child rows in the SQL we generate and how we parse the flattened string with all the values
* Use expected delimiter for SQL Server
* Be CTE-aware in SQL Server query generation

### DIFF
--- a/api/src/org/labkey/api/data/MultiValuedDisplayColumn.java
+++ b/api/src/org/labkey/api/data/MultiValuedDisplayColumn.java
@@ -135,9 +135,13 @@ public class MultiValuedDisplayColumn extends DisplayColumnDecorator implements 
 
             while (mvCtx.next())
             {
-                out.append(sep);
-                super.renderGridCellContents(mvCtx, out);
-                sep = ", ";
+                Object o = getValue(mvCtx);
+                if (o != null)
+                {
+                    out.append(sep);
+                    super.renderGridCellContents(mvCtx, out);
+                    sep = ", ";
+                }
             }
         }
         finally
@@ -176,7 +180,7 @@ public class MultiValuedDisplayColumn extends DisplayColumnDecorator implements 
     @Override
     public String getTsvFormattedValue(RenderContext ctx)
     {
-        return getTsvFormattedValues(ctx).stream().collect(Collectors.joining(", "));
+        return String.join(", ", getTsvFormattedValues(ctx));
     }
 
     @Override
--- a/api/src/org/labkey/api/data/MultiValuedRenderContext.java
+++ b/api/src/org/labkey/api/data/MultiValuedRenderContext.java
@@ -59,6 +59,8 @@ public class MultiValuedRenderContext extends RenderContextDecorator
             }
             else
             {
+                // Use -1 as the limit so that we pick up back-to-back delimiters as empty strings in the returned array,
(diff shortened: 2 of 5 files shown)


api/src/org/labkey/api/data/MultiValuedDisplayColumn.java
```diff
@@ -176,7 +180,7 @@ public List<Object> getDisplayValues(RenderContext ctx)
     @Override
     public String getTsvFormattedValue(RenderContext ctx)
     {
-        return getTsvFormattedValues(ctx).stream().collect(Collectors.joining(", "));
+        return String.join(", ", getTsvFormattedValues(ctx));
```
labkey-susanh, line 183: Just a simplification, right? Not intending to change the actual output.

bigiron/src/org/labkey/bigiron/mssql/BaseMicrosoftSqlServerDialect.java
```diff
@@ -728,9 +728,9 @@ else if (c == ')')
         {
             throw new IllegalArgumentException("Can't handle SQL: " + sql);
         }
-        ret.insert(fromIndex, "AS NVARCHAR) AS [text()] ");
+        ret.insert(fromIndex, "AS NVARCHAR), '') AS [text()] ");
```
labkey-susanh, line 733: Is the COALESCE corresponding to this change the one inserted below?  Might be worth a comment.

labkey-susanh (approved):
Seems OK to me, if a little bit crazy to read.

api/src/org/labkey/api/data/MultiValuedDisplayColumn.java
labkey-jeckels: Correct, this was an IntelliJ auto-refactor suggestion. No functional difference.

bigiron/src/org/labkey/bigiron/mssql/BaseMicrosoftSqlServerDialect.java
labkey-jeckels: Comment added

api/src/org/labkey/api/data/MultiValuedRenderContext.java
```diff
@@ -59,7 +59,7 @@ public MultiValuedRenderContext(RenderContext ctx, Set<FieldKey> requiredFieldKe
             }
             else
             {
-                String[] values = value.toString().split(VALUE_DELIMITER_REGEX);
+                String[] values = value.toString().split(VALUE_DELIMITER_REGEX, -1);
```
labkey-matthewb, line 64: You made me look at the java docs.

labkey-matthewb (approved):

api/src/org/labkey/api/data/MultiValuedRenderContext.java
labkey-jeckels: I don't want that. I added a comment.

PR #4824, opened 2023-10-13
RFC 5987

labkey-matthewb:
#### Rationale
ResponseHelper.setContentDisposition()
create common code-path to conform to RFC 5987 encoding for "Content-Disposition" header

#### Related Pull Requests
There are still direct calls to addHeader("Content-Disposition") in other repositories.

#### Changes
*


api/src/org/labkey/api/util/ResponseHelper.java
```diff
@@ -83,6 +135,36 @@ public static void setPublicStatic(HttpServletResponse response, int days)
...
     }
 
+    public static void setContentDisposition(HttpServletResponse response, String type)
+    {
+        switch (type.toLowerCase())
+        {
+            case "attachment":
+            case "inline":
+            case "form-data":
+                 response.setHeader("Content-Disposition", "form-data");
```
labkey-susanh: Why do these all get converted to `form-data`?

api/src/org/labkey/api/util/ResponseHelper.java
```diff
@@ -18,18 +18,70 @@
...
+        ContentDispositionType(String type)
+        {
+            this._type = type;
+        }
+
+        abstract ContentDisposition.Builder builder();
+
+        public String toHeaderValue()
+        {
+            return _type;
```
labkey-susanh, line 74: Wouldn't we still want this to go through `builder()`?

api/src/org/labkey/api/writer/ZipFile.java
```diff
@@ -96,7 +97,7 @@ private static OutputStream getOutputStream(Path root, String name) throws IOExc
     public static OutputStream getOutputStream(HttpServletResponse response, String name) throws IOException
     {
         response.setContentType("application/zip");
-        response.setHeader("Content-Disposition", "attachment; filename=\"" + _makeLegalName(name) + "\";");
+        ResponseHelper.setContentDisposition(response, ResponseHelper.ContentDispositionType.attachment, _makeLegalName(name));
```
labkey-susanh: Seems like the call to `_makeLegalName` could be removed since that's being done in `setContentDisposition` now.

core/src/org/labkey/core/attachment/AttachmentServiceImpl.java
```diff
@@ -1029,7 +1028,7 @@ private void writeDocument(DocumentWriter writer, AttachmentParent parent, Strin
                     throw new NotFoundException("Could not find file " + alias);
 
                 if (asAttachment)
-                    writer.setContentDisposition("attachment; filename=\"" + alias + "\"");
+                    writer.setContentDisposition(ContentDisposition.builder("attachment").filename(alias, StandardCharsets.UTF_8).build());
```
labkey-susanh: Could do `ContentDisposition.attachment()` instead. though I know it's the same thing in the end.

api/src/org/labkey/api/util/ResponseHelper.java
labkey-matthewb: Because I added ContentDispositionType.__type so I wouldn't need this switch at all, and SQUIRREL!

api/src/org/labkey/api/util/ResponseHelper.java
labkey-matthewb: Seemed like a lot of churn to out put a string we already have.

api/src/org/labkey/api/writer/ZipFile.java
labkey-matthewb: I added that later (once I found bad usages elsewhere).  Good call.

labkey-susanh (approved):

PR #4826, opened 2023-10-16
Support for marketing content

labkey-klum:
#### Rationale
We want to be able to push marketing updates to community edition server instances that all users are able to view. This PR introduces the following changes:

- Update the mothership configuration UI to contain an optional marketing update message. This is similar to the release update message and can contain HTML.
- Add platform updates for servers running a `community` distribution to check for marketing updates every 24 hours and display them in the ribbon banner at the top of every page.
- Include a default message if there is no configured message available from mothership or if the mothership server cannot be reached.
- Add a hidden experimental flag to allow for the self test of the marketing updates. In this mode the message is fetched from the local server running mothership and the distribution is not filtered.
- For servers running a CE distribution, we no longer provide a way to opt-out of updates and usage reporting. We will also force the reporting level to `ON` if it was previously turned off. The site settings for these servers will explain the behavior but no longer allow them to control it.

image


labkey-tchad (commented):
Looks good from manual test pass

core/src/org/labkey/core/admin/customizeSite.jsp
```diff
@@ -180,10 +181,23 @@ Click the Save button at any time to accept the current settings and continue.</
     </td>
     <td>
         <table>
+<%
+        if (MothershipReport.shouldReceiveMarketingUpdates(MothershipReport.getDistributionName()))
+        {
+%>
+            <tr>
+                <td><span>Update checks and usage reporting are automatically <strong>on</strong> for servers running a community edition distribution.</span></td>
```
labkey-jeckels:
```suggestion
                <td><span>Update checks and usage reporting are automatically <strong>on</strong> for servers running a LabKey Server Community Edition.</span></td>

```

labkey-jeckels (approved):

PR #4829, opened 2023-10-17
Assay Design Renaming

### LINKED ISSUE #3705: Assay Design Renaming - dev
[Spec](https://docs.google.com/document/d/1Cy0p98tLqgqAhu6NEGKn5nulGVKk0RJw3SgDhmO9iww/edit?usp=sharing)
[Github epic](https://github.com/LabKey/platform/issues/3703)

**Tasks**
- [x] Use DBSeq instead of name in LSID for TSVAssay protocol and domains - 3h
   - [x] Only query for new DB Seq for protocol
   - [x] Pass the same seq to .core and .output protocol
   - [x] Pass the same seq to 3 domains (batch, run and result) 
   - [x] no need to migrate existing lsids
- [x] Update assay export/import with new LSID- 5h
   - [x] BatchProtocolLSID should no longer reference name
   - [x] domainUrl for 3 domains should no longer reference name
   - [x] action LSID
   - [x] OntologyEntryURISupport 
   - [x] folder import with name collide
- [x] updating name for TSVAssay - 5h
   - [x] Find all 3 related exp.protocol and update name
   - [x] Update all 3 exp.protocolparameter.applicationname record
   - [x] Update exp.domaindescriptor.name batch, run and result
- [x] Enable GPAT name update on designer UI - 1h
- [x] Fix audit log for assay name update - 1h
   - [x] An audit entry will be added per domain (batch, run and result) for the renaming event
- [x] Update custom view reference for renamed assays - 2h
   - [x] Instead of queryName, assay renaming will change part of the schema name, so can’t use handleQueryNameChange directly
- [x] Fix audit log msg for sample type and dataclass renaming - 1h
- [x] Fix stale import aliases - 2h
- [x] Fix ELN reference for sample type, dataclass and assays - 3h
- [x] Fix Sample Finder reference - 2h
- [X] [Manual test plan](https://docs.google.com/document/d/1Cy0p98tLqgqAhu6NEGKn5nulGVKk0RJw3SgDhmO9iww/edit#heading=h.e6gj012203y) - 1hr
- [x] update selenium tests - 2h
- [x] Feature verification - .5hr @labkey-hannah 
- [x] Manual testing/Manual Testing adjustments - 3hr @cnathe 
- [x] Code review/Code review feedback - 2hr @cnathe 
- [x] Automated test plan - 1h
- [x] Teamcity review and merge - 1hr
- [x] Docs handoff - .5hr

XingY:
#### Rationale
This PR adds support for renaming GPAT assay designs.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1318
- https://github.com/LabKey/labkey-ui-premium/pull/222
- https://github.com/LabKey/platform/pull/4829
- https://github.com/LabKey/commonAssays/pull/653
- https://github.com/LabKey/labbook/pull/555
- https://github.com/LabKey/sampleManagement/pull/2172
- https://github.com/LabKey/biologics/pull/2450

#### Changes
* Modified QueryChangeListener and extended reports/views to support schema name update
* Support GPAT assay design renaming
* Modified domain audit log to include renaming msg
*


api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -665,8 +665,11 @@ public static ValidationException updateDomainDescriptor(GWTDomain<? extends GWT
             return validationException;
         }
 
-        if (updateDomainName)
+        if (updateDomainName && !d.getName().equals(update.getName()))
```
cnathe, line 668: I'm curious why we have this `updateDomainName` param if we can/do just check if the names are equal from `d` and `update`?

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
```diff
@@ -361,7 +365,8 @@ private void setPropertyDomainURIs(ExpProtocol protocol, Set<String> uris)
         {
             if (!props.containsKey(uri))
             {
-                ObjectProperty prop = new ObjectProperty(protocol.getLSID(), protocol.getContainer(), uri, uri);
+                ObjectProperty prop = new ObjectProperty(protocol.getLSID(), protocol.getContainer(), uri, uri);//
```
cnathe:
```suggestion
                ObjectProperty prop = new ObjectProperty(protocol.getLSID(), protocol.getContainer(), uri, uri);
```

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
```diff
@@ -451,12 +462,28 @@ public GWTProtocol saveChanges(GWTProtocol assay, boolean replaceIfExisting) thr
...
 
+                    if (hasNameChange)
+                    {
+                        ExperimentService.get().handleAssayNameChange(assay.getName(), oldAssayName, assayProvider,  protocol,getUser(), getContainer());
+                        if (assayProvider != null)
+                        {
+                            SchemaKey newSchema = SchemaKey.fromParts(AssaySchema.NAME, assayProvider.getResourceName(), assay.getName());
+                            SchemaKey oldSchema = SchemaKey.fromParts(AssaySchema.NAME, assayProvider.getResourceName(), oldAssayName);
+                            QueryChangeListener.QueryPropertyChange.handleSchemaNameChange(oldSchema.toString(), newSchema.toString(), newSchema, getUser(), getContainer());
+                        }
```
cnathe: this is the only place that calls `handleAssayNameChange`, right? If so, could these lines and the call to `handleSchemaNameChange` just be moved within `handleAssayNameChange`?

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
```diff
@@ -436,6 +445,8 @@ public GWTProtocol saveChanges(GWTProtocol assay, boolean replaceIfExisting) thr
                         if (!protocol.getContainer().equals(getContainer()))
                             throw new AssayException("Assays can only be edited in the folder where they were created.  " +
                                     "This assay was created in folder " + protocol.getContainer().getPath());
+                        oldAssayName = protocol.getName();
+                        hasNameChange = !assay.getName().equals(oldAssayName);
```
cnathe, line 449: should we be using `provider.hasDomainNameChanged()` here?

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
```diff
@@ -555,7 +582,7 @@ public GWTProtocol saveChanges(GWTProtocol assay, boolean replaceIfExisting) thr
                     StringBuilder errors = new StringBuilder();
                     for (GWTDomain<GWTPropertyDescriptor> domain : assay.getDomains())
                     {
-                        ValidationException domainErrors = updateDomainDescriptor(domain, protocol, provider);
+                        ValidationException domainErrors = updateDomainDescriptor(domain, protocol, provider, isNew);
```
cnathe: could we just pass through `hasNameChange` here instead of recalculating it inside of `updateDomainDescriptor`?

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
```diff
@@ -596,8 +623,18 @@ private ValidationException updateDomainDescriptor(GWTDomain<GWTPropertyDescript
...
+        boolean hasNameChange = !isNew && provider.hasDomainNameChanged(protocol, domain);
+        String oldName = domain.getName();
+        String newName = protocol.getName();
         provider.changeDomain(getUser(), protocol, previous, domain);
-        return DomainUtil.updateDomainDescriptor(previous, domain, getContainer(), getUser());
+
+        String auditComment = null;
+        if (hasNameChange)
+        {
+            auditComment = "The name of the assay domain '" + oldName + "' was changed to '" + newName + "'.";
```
cnathe: as mentioned in manual testing, the `oldName` here has the protocol name and the domain name (i.e. "Titer Run Fields") whereas the `newName` is just the protocol name. It was a little confusing to me when looking at the audit entries.

assay/src/org/labkey/assay/TsvAssayProvider.java
```diff
@@ -202,7 +269,7 @@ public AssayTableMetadata getTableMetadata(@NotNull ExpProtocol protocol)
     @Override
     public List<Pair<Domain, Map<DomainProperty, Object>>> createDefaultDomains(Container c, User user)
     {
-        List<Pair<Domain, Map<DomainProperty, Object>>> result = super.createDefaultDomains(c, user);
+        List<Pair<Domain, Map<DomainProperty, Object>>> result = super.createDefaultDomains(c, user);//
```
cnathe:
```suggestion
        List<Pair<Domain, Map<DomainProperty, Object>>> result = super.createDefaultDomains(c, user);
```

assay/src/org/labkey/assay/TsvAssayProvider.java
```diff
@@ -212,7 +279,7 @@ public List<Pair<Domain, Map<DomainProperty, Object>>> createDefaultDomains(Cont
 
     protected Pair<Domain,Map<DomainProperty,Object>> createResultDomain(Container c, User user)
     {
-        Domain dataDomain = PropertyService.get().createDomain(c, getPresubstitutionLsid(ExpProtocol.ASSAY_DOMAIN_DATA), "Data Fields");
+        Domain dataDomain = PropertyService.get().createDomain(c, getPresubstitutionLsid(ExpProtocol.ASSAY_DOMAIN_DATA, ASSAY_DBSEQ_SUBSTITUTION), "Data Fields");//
```
cnathe:
```suggestion
        Domain dataDomain = PropertyService.get().createDomain(c, getPresubstitutionLsid(ExpProtocol.ASSAY_DOMAIN_DATA, ASSAY_DBSEQ_SUBSTITUTION), "Data Fields");
```

assay/src/org/labkey/assay/TsvAssayProvider.java
```diff
@@ -340,6 +407,11 @@ public void changeDomain(User user, ExpProtocol protocol, GWTDomain<GWTPropertyD
     {
         super.changeDomain(user, protocol, orig, update);
 
+        if (!(protocol.getName() + getDomainNameSuffix(orig)).equals(orig.getName()))
```
cnathe: this looks to be a duplicate of `hasDomainNameChanged`, can that be called here instead?

experiment/src/org/labkey/experiment/XarReader.java
```diff
@@ -724,14 +727,18 @@ private ExpDataClass loadDataClass(DataClassType dataClassType) throws Experimen
...
         Domain domain = loaded.getKey();
         Map<DomainProperty, Object> newDefaultValues = loaded.getValue();
         String lsid = domain.getTypeURI();
 
         DomainDescriptor existingDomainDescriptor = OntologyManager.getDomainDescriptor(lsid, getContainer());
+        if (existingDomainDescriptor == null && useName)
+            existingDomainDescriptor = OntologyManager.fetchDomainDescriptorFromDB(xDomain.getName(), getContainer(), true);
         DomainImpl existingDomain = existingDomainDescriptor == null ? null : new DomainImpl(existingDomainDescriptor);
-        if (existingDomain != null)
+        if (existingDomain != null)//
```
cnathe:
```suggestion
        if (existingDomain != null)
```

experiment/src/org/labkey/experiment/XarReader.java
```diff
@@ -917,7 +924,10 @@ private void loadExperiment(ExperimentType exp) throws SQLException, XarFormatEx
             if (exp.isSetBatchProtocolLSID())
             {
                 String batchProtocolLSID = LsidUtils.resolveLsidFromTemplate(exp.getBatchProtocolLSID(), getRootContext(), "Protocol");
-                ExpProtocol batchProtocol = ExperimentService.get().getExpProtocol(batchProtocolLSID);
+
+                ExpProtocol batchProtocol = _loadedProtocols.get(batchProtocolLSID);
+                if (batchProtocol == null)
+                    batchProtocol = ExperimentService.get().getExpProtocol(batchProtocolLSID);
```
cnathe: this set of lines looks like it happens several times in this file. Could it be refactored to a helper method which does the "first check _loadedProtocols" check?

experiment/src/org/labkey/experiment/samples/ExperimentQueryChangeListener.java
```diff
@@ -0,0 +1,169 @@
...
+            queryNameChangeMap.put((String)qpc.getOldValue(), (String)qpc.getNewValue());
+        }
+
+        String prefix = isSamples ? "materialInputs/" : "dataInputs/";
+
+        for (String oldQueryName : queryNameChangeMap.keySet())
+        {
+            String newQueryName = queryNameChangeMap.get(oldQueryName);
+            if (oldQueryName.equals(newQueryName))
+                continue;
```
cnathe: minor: similar comment here re: moving this check up to the for loop at line 49

experiment/src/org/labkey/experiment/samples/ExperimentQueryChangeListener.java
```diff
@@ -0,0 +1,169 @@
...
+        if (!property.equals(QueryProperty.Name))
+            return;
+
+        Map<String, String> queryNameChangeMap = new CaseInsensitiveHashMap<>();
+        for (QueryPropertyChange qpc : changes)
+        {
+            queryNameChangeMap.put((String)qpc.getOldValue(), (String)qpc.getNewValue());
+        }
+
+        String prefix = isSamples ? "materialInputs/" : "dataInputs/";
```
cnathe: Do we have these strings as constants some where? I seem to recall that we do...but maybe not.

experiment/src/org/labkey/experiment/samples/ExperimentQueryChangeListener.java
```diff
@@ -0,0 +1,169 @@
...
+        for (String removed : queries)
+        {
+            String searchStr = "\"" + prefix + removed + "\"";
+
+            SimpleFilter sampleTypeFilter = SimpleFilter.createContainerFilter(container, "Container");
+            sampleTypeFilter.addCondition(FieldKey.fromParts("materialparentimportaliasmap"), searchStr, CompareType.CONTAINS);
+            TableInfo sampleTypeTable = ExperimentServiceImpl.get().getTinfoSampleType();
+            List<ExpSampleTypeImpl> sampleTypes = new TableSelector(sampleTypeTable, sampleTypeFilter, null).stream(MaterialSource.class)
+                    .map(ExpSampleTypeImpl::new)
+                    .toList();
```
cnathe: this looks like a duplicate from the `queryChanged` method above. Could they be shared? 
Same comment below for the data class filtering.

query/src/org/labkey/query/persist/QueryManager.java
```diff
@@ -506,6 +506,14 @@ public void fireQueryChanged(User user, Container container, ContainerFilter sco
             l.queryChanged(user, container, scope, schema, property, changes);
     }
 
+//    public void fireQueryChanged(User user, Container container, ContainerFilter scope, SchemaKey schema, @NotNull QueryChangeListener.QueryProperty property, @NotNull Collection<QueryPropertyChange> changes)
```
cnathe: remove?

cnathe (changes requested):

api/src/org/labkey/api/exp/property/DomainUtil.java
XingY: I'm not positive update.getName will always be populated. For example, if update using api, the name might not be? Also, we want to limit the ability to rename to the set of domains that we explicitly support renaming, and from the code path that we support (UI). This could probably be improved, but adding the redundancy here for safety.

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
XingY: I think not. This one is for protocol.name, not domain name

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
XingY: good point. then we no longer need provider.hasDomainNameChanged()

cnathe (approved):

PR #4830, opened 2023-10-17
Issue 48788: Add index for container + created for audit tables

labkey-susanh:
#### Rationale
[Issue 48788](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48788): We want queries to audit logs to be faster, so let's index some of the fields we know are commonly used for querying.

#### Related Pull Requests

- https://github.com/LabKey/inventory/pull/1064

#### Changes
* Add index on (container, created) by default
* Add index on lsid for dataset audit logs
* Add index on sampleId for sample timeline audit logs


api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -935,6 +938,18 @@ public Config(Properties properties)
 
         }
 
+        public Map<Enum, Object> getImportConfig()
```
labkey-matthewb, line 945: Is this related or a different fix/patch?

labkey-matthewb (approved):

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: It's somewhat related here because I wanted to augment the data generators to be able to specify the audit level so we could generate lots of audit logs.

PR #4833, opened 2023-10-18
Issue 48347: Field editor "User" data type to allow lookup validator option

cnathe:
#### Rationale
Issue [48347](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48347): LKSM: Not validating user field in sample manager

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1321

#### Changes
- SimpleTranslator.createConverColumn update to use error in the case when a lookup validator is present
- update @labkey/build package version


api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -1294,8 +1294,9 @@ private SimpleConvertColumn createConvertColumn(@NotNull ColumnInfo col, int fro
         if (fk != null && _context.isAllowImportLookupByAlternateKey() && fk.allowImportByAlternateKey())
         {
             RemapMissingBehavior missing = remapMissingBehavior;
+            boolean hasValidator = pd != null && pd.getValidators().size() > 0;
             if (missing == null)
-                missing = col.isRequired() ? RemapMissingBehavior.Error : RemapMissingBehavior.Null;
+                missing = col.isRequired() || hasValidator ? RemapMissingBehavior.Error : RemapMissingBehavior.Null;
```
labkey-susanh, line 1302: Do validators ever allow for a value to be null? If so, it seems that this will report that as an error.

cnathe: `missing` here is the variable that determines whether a null value should be considered an error or should be allowed (i.e. RemapMissingBehavior.Null). However, I did update the check to more explicitly look for a pd validator that is of type `PropertyValidatorType.Lookup`

labkey-susanh (approved):

PR #4835, opened 2023-10-18
Issue 48530: 'Contains One Of' with 1250 samples results in Request header is too large

XingY:
#### Rationale
See issue [here](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48530).

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4608

#### Changes
* Ext changes to disallow long filter values


RosalineP (approved):
Looks good, thanks Xing! Could we align on a value of 2000 instead of 1000, though? We also might want to pull that value into a constant, and I can add a comment so that the length limiter number stays aligned across LKS and LKB/LKSM (platform and ui-components)

PR #4836, opened 2023-10-18
Provide "report-to" support for Content-Security-Policy

labkey-matthewb:
#### Rationale
CSP supports a mechanism to report problems to a server.  This might be very useful for detecting issues using our test automation framework.

#### Related Pull Requests
* 

#### Changes
*


api/src/org/labkey/api/action/BaseApiAction.java
```diff
@@ -312,7 +312,7 @@ private Pair<FORM, BindException> populateForm() throws Exception
             String contentType = getViewContext().getRequest().getContentType();
             if (null != contentType)
             {
-                if (contentType.contains(ApiJsonWriter.CONTENT_TYPE_JSON))
+                if (contentType.contains(ApiJsonWriter.CONTENT_TYPE_JSON) || contentType.equals("application/csp-report"))
```
labkey-jeckels: Probably good to introduce a constant for the MIME type.

labkey-jeckels (approved):
Is the idea that clients will report back to the client they're talking to, or a centralized place like labkey.org?

labkey-jeckels (approved):
Oops, I reviewed but must have failed to actually approve.

PR #4839, opened 2023-10-19
Update list search timestamp to only update if resource not deleted

Sigmonia:
#### Rationale
[Issue #48878](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48878) Search indexer has race condition

#### Changes
* Update list indexer to only update if item resource is present

### DIFF
--- a/list/src/org/labkey/list/model/ListManager.java
+++ b/list/src/org/labkey/list/model/ListManager.java
@@ -76,6 +76,7 @@ import org.labkey.api.webdav.WebdavResource;
 import org.labkey.list.controllers.ListController;
 import org.labkey.list.model.ListImporter.ValidatorImporter;
 import org.labkey.list.view.ListItemAttachmentParent;
+import org.springframework.jdbc.BadSqlGrammarException;
 
 import java.sql.Timestamp;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


list/src/org/labkey/list/model/ListManager.java
```diff
@@ -1088,8 +1088,13 @@ private void setItemLastIndexed(ListDefinition list, Object pk, TableInfo ti, lo
             if (null != keyColumn)
             {
                 String keySelectName = keyColumn.getSelectName();
-                new SqlExecutor(sti.getSchema()).execute("UPDATE " + getListTableName(sti) + " SET LastIndexed = ? WHERE " +
-                        keySelectName + " = ?", new Timestamp(ms), pk);
+
```
labkey-matthewb: I think this additional condition in the WHERE clause is redundant.  The original already checks that the row exists.  Wasn't the bug that the actual list got deleted?

labkey-matthewb (changes requested):

list/src/org/labkey/list/model/ListManager.java
labkey-adam: Correct... I discussed this with Ian. My suggestion is to catch, swallow constraint violations, and re-throw anything else.

labkey-adam (approved):

list/src/org/labkey/list/model/ListManager.java
```diff
@@ -695,7 +695,18 @@ private void indexItems(@NotNull final IndexTask task, final ListDefinition list
...
                     {
-                        ListManager.get().setItemLastIndexed(list, pk, listTable, ms, modified);
+                        try
+                        {
+                            ListManager.get().setItemLastIndexed(list, pk, listTable, ms, modified);
+                        }
+                        catch (RuntimeSQLException e)
+                        {
+                            // This may occur due to a race condition between enumeration and list deletion. Issue #48878
+                            if (e.isConstraintException())
```
labkey-matthewb: Has anyone tested that table not found throws an exception that is flagged as a constraint exception???  I might be, but I wouldn't expect that.

Sigmonia: tested newest updated against both MS-SQL and psql.

labkey-matthewb (approved):

PR #4841, opened 2023-10-19
use document-uri instead of source-file

labkey-matthewb:
#### Rationale
document-uri is better for avoiding duplicates

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -10959,7 +10959,7 @@ public class AdminController extends SpringActionController
                 JSONObject cspReport = jsonObj.optJSONObject("csp-report");
                 if (cspReport != null)
                 {
-                    String urlString = cspReport.optString("document-uri");
+                    String urlString = cspReport.optString("document-uri", null);
                     if (urlString != null)
                     {
                         String path = new URLHelper(urlString).deleteParameters().getPath();
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -10955,7 +10955,7 @@ public class AdminController extends SpringActionController
             var jsonObj = form.getJsonObject();
             if (null != jsonObj)
             {
-                var jsonStr = jsonObj.toString();
+                var jsonStr = jsonObj.toString(2);
                 JSONObject cspReport = jsonObj.optJSONObject("csp-report");
                 if (cspReport != null)
                 {
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -182,7 +182,6 @@ import org.labkey.core.security.BlockListFilter;
 import org.labkey.core.security.SecurityController;
 import org.labkey.data.xml.TablesDocument;
 import org.labkey.security.xml.GroupEnumType;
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -10952,10 +10952,16 @@ public Object execute(SimpleApiJsonForm form, BindException errors) throws Excep
...
+                if (cspReport != null)
+                {
+                    String urlString = cspReport.getString("document-uri");
+                    if (urlString != null)
+                    {
+                        String path = new URLHelper(urlString).deleteParameters().getPath();
+                        if (null == reports.put(path, Boolean.TRUE))
+                            _log.warn("ContentSecurityPolicy warning on page: " + urlString + "\n" + jsonStr);
+                    }
+                }
```
labkey-tchad: Should use `JSONObject.opt*` for items that might not be there; `get*` will just throw an exception if they don't find anything. Also, `JSONObject.toString` can take an indent factor to make it look pretty.
```suggestion
                JSONObject cspReport = jsonObj.optJSONObject("csp-report");
                if (cspReport != null)
                {
                    String urlString = cspReport.optString("document-uri", null);
                    if (urlString != null)
                    {
                        String path = new URLHelper(urlString).deleteParameters().getPath();
                        if (null == reports.put(path, Boolean.TRUE))
                        {
                            var jsonStr = jsonObj.toString(2);
                            _log.warn("ContentSecurityPolicy warning on page: " + urlString + "\n" + jsonStr);
                        }
                    }
                }
```

labkey-tchad (changes requested):
I thought of another way to 'disable' this endpoint. It could no-op if the logger is turned off. Like:
```
if (!_log.isWarnEnabled())
{
    return new JSONObject().put("success", false);
}
``` 

Also, we should update `server/webapps/log4j2.xml` to put these warning into a separate file. That would make it possible for tests to check for new warnings (and fail if we want).

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -10946,18 +10945,30 @@ public class ContentSecurityPolicyReportAction extends ReadOnlyApiAction<SimpleA
...
                 var jsonStr = jsonObj.toString();
-                JSONObject cspReport = jsonObj.getJSONObject("csp-report");
-                String urlString = cspReport.getString("source-file");
-                String path = new URLHelper(urlString).deleteParameters().getPath();
-                if (null == reports.put(path,Boolean.TRUE))
-                    _log.warn("ContentSecurityPolicy warning:\n" + jsonStr);
+                JSONObject cspReport = jsonObj.optJSONObject("csp-report");
+                if (cspReport != null)
+                {
+                    String urlString = cspReport.optString("document-uri");
```
labkey-tchad: Unless specified, `optString(key)`'s default value is `""`, not `null`.
```suggestion
                    String urlString = cspReport.optString("document-uri", null);
```

`org.json` has numerous questionable design decisions with regard to `null` values.

labkey-tchad (approved):
One small fix

core/src/org/labkey/core/admin/AdminController.java
labkey-matthewb: Thanks, now I know.

PR #4843, opened 2023-10-20
LabKeyUIComponentsPage.tsx update to limit included @labkey/components

cnathe:
#### Rationale
To avoid confusion we are limiting the components from @labkey/components that are included in the LabKeyUIComponentsPage.tsx to just those from /public and those included in the selenium tests.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1690

#### Changes
- only include the subset of @labkey/components that are used in selenium tests
- linting


labkey-alan (approved):
I am glad to see the non-public components go, but personally I do think we should limit the components on this page to only those that are needed by our tests. If we feel we need examples of the public components we are better off writing documentation in the ui-components repo, rather than spreading the docs between here and ui-components.

PR #4856, opened 2023-10-24
Issue 48925: Indexing pulls all samples in a container in a single list

labkey-jeckels:
#### Rationale
Some servers have more samples in a single container than we can comfortably fit into memory at once for indexing purposes.

#### Changes
* Break samples and datas into blocks of 10,000 for indexing purposes
* Make selector methods return the number of rows processed
* Shift all responsibility for search indexing from ExperimentModule to ExperimentServiceImpl


labkey-adam (commented):
Looks good to me. Might add a comment about not overwhelming the queue when you check rowCount.

labkey-adam (approved):
Get rid of `EXPERIMENTAL_DOMAIN_DESIGNER` constant as well?

PR #4870, opened 2023-10-25
GetUsersWithPermissionsAction: Add includeInactive flag

labkey-alan:
#### Rationale
Users want to be able to search for ELNs via author, even if the author is deactivated ([Issue 48250](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48250)).

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4870
* https://github.com/LabKey/labkey-api-js/pull/164
* https://github.com/LabKey/labkey-ui-components/pull/1328
* https://github.com/LabKey/labkey-ui-premium/pull/235
* https://github.com/LabKey/biologics/pull/2475
* https://github.com/LabKey/inventory/pull/1076
* https://github.com/LabKey/sampleManagement/pull/2191

#### Changes
* GetUsersWithPermissionsAction: version the API, add includeInactive flag
* SecurityManager: Add includeInactive flag to getUsersWithPermissions method

#### Additional Context
***Note:*** The below context has been addressed by versioning the API. I'm keeping it here since it's useful history.

I am really not sure that this is the correct way to go, but it seems like our best bet if we want to remain backwards compatible, and I'd really appreciate some feedback if there's maybe a better approach. Here are the current downsides as I see them:

1. We already have an `active` flag, so this new flag is kind of confusing since it's the opposite. However, even though we have the active flag, it's currently ignored if you're not requesting users of a group. This means that if we start to use it all existing consumers of this API will need to be updated to pass the flag as `true`, or they'll start getting inactive users returned. It also doesn't feel appropriate to use the `active` flag for this behavior, the behavior we want is to return all users, active or deactivated, or active users. So we'd be treating the `active` flag as meaning `activeUsersOnly`, which seems odd.
2. The active flag is used if you're trying to get users of a group (if the `group` attribute is passed), this means we cannot use the `includeDeactivated` flag for this usage of the GetUsersWithPermissionsAction without also breaking backwards compatibility (all consumers passing the active flag would have to be updated to use the `includeDeactivated` flag).

Here are the upsides:
1. It works
2. It's backwards compatible

Maybe we should version this API like we do for selectRows? We could make the default version `23.10`, and if `23.11` is passed we honor the `includeDeactivated` flag, but not the `active` flag.

### DIFF
--- a/api/src/org/labkey/api/security/SecurityManager.java
+++ b/api/src/org/labkey/api/security/SecurityManager.java
@@ -2173,11 +2173,11 @@ public class SecurityManager
     /**
      * @return an immutable list of Users who have been assigned all the requested permissions in the given container
      */
-    public static List<User> getUsersWithPermissions(Container c, boolean includeDeactivated, Set<Class<? extends Permission>> perms)
+    public static List<User> getUsersWithPermissions(Container c, boolean includeInactive, Set<Class<? extends Permission>> perms)
     {
         // No cache right now, but performance seems fine. After the user list and policy are cached, no other queries occur.
         SecurityPolicy policy = c.getPolicy();
-        return UserManager.getUsers(includeDeactivated).stream()
+        return UserManager.getUsers(includeInactive).stream()
             .filter(user -> SecurityManager.hasAllPermissions(null, policy, user, perms, Set.of()))
             .toList();
     }
--- a/api/src/org/labkey/api/security/SecurityManager.java
+++ b/api/src/org/labkey/api/security/SecurityManager.java
@@ -2173,13 +2173,21 @@ public class SecurityManager
     /**
      * @return an immutable list of Users who have been assigned all the requested permissions in the given container
      */
-    public static List<User> getUsersWithPermissions(Container c, Set<Class<? extends Permission>> perms)
+    public static List<User> getUsersWithPermissions(Container c, boolean includeDeactivated, Set<Class<? extends Permission>> perms)
     {
         // No cache right now, but performance seems fine. After the user list and policy are cached, no other queries occur.
         SecurityPolicy policy = c.getPolicy();
-        return UserManager.getActiveUsers().stream()
+        return UserManager.getUsers(includeDeactivated).stream()
             .filter(user -> SecurityManager.hasAllPermissions(null, policy, user, perms, Set.of()))
             .toList();
+    }
+
+    /**
+     * @return an immutable list of Users who have been assigned all the requested permissions in the given container
+     */
+    public static List<User> getUsersWithPermissions(Container c, Set<Class<? extends Permission>> perms)
+    {
+        return getUsersWithPermissions(c, false, perms);
     }
 
     /**
--- a/api/src/org/labkey/api/security/SecurityManager.java
+++ b/api/src/org/labkey/api/security/SecurityManager.java
@@ -2183,7 +2183,7 @@ public class SecurityManager
     }
 
     /**
-     * @return an immutable list of Users who have been assigned all the requested permissions in the given container
+     * @return an immutable list of active Users who have been assigned all the requested permissions in the given container
      */
     public static List<User> getUsersWithPermissions(Container c, Set<Class<? extends Permission>> perms)
     {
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -2173,15 +2173,23 @@ public static Collection<Integer> getFolderUserids(Container c)
...
         // No cache right now, but performance seems fine. After the user list and policy are cached, no other queries occur.
         SecurityPolicy policy = c.getPolicy();
-        return UserManager.getActiveUsers().stream()
+        return UserManager.getUsers(includeDeactivated).stream()
             .filter(user -> SecurityManager.hasAllPermissions(null, policy, user, perms, Set.of()))
             .toList();
     }
 
+    /**
+     * @return an immutable list of Users who have been assigned all the requested permissions in the given container
```
labkey-adam:
```suggestion
     * @return an immutable list of active Users who have been assigned all the requested permissions in the given container

```

core/src/org/labkey/core/user/UserController.java
```diff
@@ -2501,6 +2501,7 @@ public static class GetUsersForm
         private boolean _active; // should we get only active members (relevant only if permissions is empty)
         private Permission[] _permissions; // the  permissions each user must have (They must have all of these)
         private Set<Class<? extends Permission>> _permissionClasses = Collections.emptySet(); // the set of permission classes corresponding to the permissions array
+        private boolean _includeDeactivated; // Should we include inactive members (relevant if permissions is not empty)
```
labkey-nicka: nit: Add a note to this comment that this flag is ignored when the "groups" property is specified.

core/src/org/labkey/core/user/UserController.java
```diff
@@ -2752,7 +2763,7 @@ public ApiResponse execute(GetUsersForm form, BindException errors)
             }
             else
             {
-                users = SecurityManager.getUsersWithPermissions(container, form.getPermissionClasses());
+                users = SecurityManager.getUsersWithPermissions(container, form.getIncludeDeactivated(), form.getPermissionClasses());
```
labkey-nicka: In general, this approach seems reasonable. I agree with you that a versioning of the API to clean-up some of the illogical patterns with `active` and `groups` would be nice.

If you go that route, are you planning on throwing if someone specified `active` in the `23.11` API versioin?

labkey-alan: I'm not sure we should throw in that case, I have updated the documentation in api-js to reflect the differences between the versions of the API so I think that is good enough.

labkey-nicka (approved):
Looks good. Endpoint test coverage would be nice.

core/src/org/labkey/core/user/UserController.java
```diff
@@ -2570,6 +2572,16 @@ public void setActive(boolean active)
         {
             _active = active;
         }
+
+        public boolean getIncludeDeactivated()
```
labkey-adam: In the product and (most) other code we call these "inactive" users, so I'd rather the flag were `includeInactive`. I think `completeUsers.api` might be the only other API action that has this kind of flag, and it's called `includeInactive` there. There are examples of "deactivated" in tests and commands... and I think you've already made changes in other repos... so I'm not going to go to the mat for this. But consistency with the other API and product terminology would be my preference.

labkey-adam (approved):
See my comment about preferring `includeInactive`, but not a required change

core/src/org/labkey/core/user/UserController.java
labkey-alan: I'd rather be consistent, so I'll update the flag to be `includeInactive` on Monday before I merge. I really only used `includeDeactivated` because that's the name of the flag in the `UserManager`.

labkey-adam: Thanks, let me know if I can help... I know it's a pain at this point. And yeah, we're not consistent internally, that's for sure.

If you're okay with it, I'll drop in a line in this PR that will count calls to the 23.10 variant and send it to mothership.

labkey-alan: I've renamed the flag across all branches

PR #4871, opened 2023-10-25
Issue 48926: Unable to delete or truncate a huge sample type

labkey-jeckels:
#### Rationale
Deleting a large sample type takes a long time and can use a lot of memory. There's ample room for optimization

#### Changes
* Batch the delete so that we don't need to pull all samples back into a single list
* Switch SQL to use JDBC parameters to make tracking the overall query perf easier
* Fix problem where `BatchForEachBlock` could fire a duplicate handling via `close()` if an exception is thrown
* Do better at reusing objects like `ExpRun` instead of fetching multiple times
* Combine queries to reduce DB round trips
* Add missing indices
* Don't clear run graph caches more than needed, leveraging a `CommitTask`
* Misc code cleanup


api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -1017,7 +1017,7 @@ public static void deleteOntologyObjects(Container c, boolean deleteOwnedObjects
 
         try
         {
-            // if uris is long, split it up
+            // if it's a long list, split it up
```
labkey-adam, line 1020: `<rant>`Oh man I dislike the pattern below. Using `ListUtils.partition().forEach()` is so much cleaner. We're dealing with arrays here, but converting to/from `ArrayList` is cheap and easy. I know it's not your code and you have better things to do... but it's still painful to look at. `</rant>`

labkey-adam (approved):

PR #4873, opened 2023-10-25
Import Excel datetimes with seconds

labkey-adam:
#### Rationale
We're failing to parse seconds from Excel datetime cells. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48930

We also want more consistent and obvious names for date-time formatting methods and constants.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/2942

### DIFF
--- a/api/src/org/labkey/api/reader/ExcelLoader.java
+++ b/api/src/org/labkey/api/reader/ExcelLoader.java
@@ -73,9 +73,6 @@ import java.util.function.Consumer;
 
 /**
  * Data loader for Excel files -- can infer columns and return rows of data
- *
- * User: jgarms
- * Date: Oct 22, 2008
  */
 public class ExcelLoader extends DataLoader
 {
@@ -436,11 +433,11 @@ public class ExcelLoader extends DataLoader
     {
         try
         {
-                var md = getWorkbookMetadata(null);
-                if (md.isSpreadSheetML())
-                    return new XlsxIterator();
-                else
-                    return new ExcelIterator();
+            var md = getWorkbookMetadata(null);
+            if (md.isSpreadSheetML())
+                return new XlsxIterator();
+            else
+                return new ExcelIterator();
         }
         catch (InvalidFormatException | IOException e)
         {
@@ -702,6 +699,7 @@ public class ExcelLoader extends DataLoader
             assertTrue(isExcel(new File(excelSamplesRoot, "SimpleExcelFile.xls")));
             assertTrue(isExcel(new File(excelSamplesRoot, "SimpleExcelFile.xlsx")));
             assertTrue(isExcel(new File(excelSamplesRoot, "fruits.xls")));
+            assertTrue(isExcel(new File(excelSamplesRoot, "DatesWithSeconds.xlsx")));
 
             // Issue 22153: detect xls file without extension
             assertTrue(isExcel(JunitUtil.getSampleData(null, "Nab/seaman/MS010407")));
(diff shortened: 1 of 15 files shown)


api/src/org/labkey/api/reader/ExcelLoader.java
```diff
@@ -970,7 +1001,8 @@ final Date getDateFromExcelDouble(double value)
 
         final String getIsoStringDateFromExcelDouble(double value)
         {
-            return DateUtil.formatDateTimeISO8601(getDateFromExcelDouble(value));
+            // Format as date time with seconds, Issue #48930
+            return DateUtil.formatDateTimeISO8601WithSeconds(getDateFromExcelDouble(value));
```
labkey-matthewb: This is an odd function name.  ISO8601 includes seconds.  The version without seconds is just derived from LabKey's old default display format.

api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -1048,6 +1048,15 @@ public static String formatDateTimeISO8601(Date date)
...
     }
 
+    /**
+     * Format date and time using ISO 8601 pattern including seconds. This is appropriate only for persisting dates in
+     * machine-readable form, for example, for export or in filenames. Most callers should use
+     * formatDateTime(Container c, Date d) instead.
+     */
+    public static String formatDateTimeISO8601WithSeconds(Date date)
+    {
+        return formatDateTime(date, ISO_DATE_FORMAT_STRING + " " + LONG_TIME_FORMAT_STRING);
```
labkey-matthewb: As per note above LONG_TIME_FORMAT_STRING really should be called ISO_TIME_FORMAT_STRING.  What we call ISO_TIME_FORMAT_STRING isn't.

api/src/org/labkey/api/reader/ExcelLoader.java
```diff
@@ -770,16 +773,44 @@ private static void checkData(ExcelLoader loader)
...
+                assertEquals(String.valueOf(Integer.MAX_VALUE), xl._toString(Integer.MAX_VALUE, true));
+                assertEquals("1.572", xl._toString("1.5720000000000001", true));
+                assertEquals("2.441", xl._toString("2.4409999999999998", true));
+                assertEquals("1.5720000000000001", xl._toString("1.5720000000000001", false));
+                assertEquals("2.4409999999999998", xl._toString("2.4409999999999998", false));
+                assertEquals("0.000001572", xl._toString("0.000001572", true));
+            }
+        }
+
+        @Test
```
labkey-matthewb, line 789: Nice

labkey-matthewb (changes requested):

api/src/org/labkey/api/reader/ExcelLoader.java
labkey-susanh: ISO8601 also apparently includes milliseconds. Do we abandon this name or actually want to use IS08601?

labkey-matthewb: Yes, thanks for the correction.

labkey-adam: My understanding is ISO 8601 is not a single format; the standard defines many, many date, time, and date+time formats. Formats can include space, T, Z, colons, etc. or not. Times are **not** required to have seconds, milliseconds, or even minutes. I believe all of the ISO constants are valid ISO formats and the ISO methods format using valid ISO formats.

labkey-adam: I'm fine if we don't want to reference ISO 8601 and agree we should come up with a better naming system so we're clear which format is which being used. But let's not pretend there's only one true ISO format and the others are not ISO.

api/src/org/labkey/api/util/DateUtil.java
labkey-adam: They're both ISO

labkey-matthewb (approved):

labkey-susanh (approved):

PR #4874, opened 2023-10-25
Issue 48968 and 48959

XingY:
Issue [48968](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48968): LKSM: Updating samples with sample IDs that look like integers are getting confused
Issue [48959](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48959): Sample Manager: Ancestor naming pattern is not respecting specified parent

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/data/NameGenerator.java
+++ b/api/src/org/labkey/api/data/NameGenerator.java
@@ -1271,7 +1271,8 @@ public class NameGenerator
 
         if (!ancestorTypes.isEmpty())
         {
-            if (!hasLookupColumn || fieldParts.size() < 2 || fieldParts.size() > 4)
+            if (!hasLookupColumn || fieldParts.size() < 2 /* should be at least 2 parts, /alias/Name */
+                    || fieldParts.size() > 4 /* should be at most 3 parts, for example /MaterialInputs/Type1/Name */)
             {
                 _syntaxErrors.add("Invalid substitution token, lookup column name not specified: ${" + fkTokDisplay + "}.");
                 return fieldParts;
--- a/api/src/org/labkey/api/data/NameGenerator.java
+++ b/api/src/org/labkey/api/data/NameGenerator.java
@@ -962,7 +962,7 @@ public class NameGenerator
                 if (token instanceof FieldKey fkTok)
                 {
                     int previousErrorCount = _syntaxErrors.size();
-                    List<String> fieldParts = processFieldParts(fkTok, partAncestorOptions, user, dataClassLSIDs, sampleTypeLSIDs);
+                    List<String> fieldParts = processFieldParts(fkTok, partAncestorOptions, user, dataClassLSIDs, sampleTypeLSIDs, importAliases);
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1256,9 +1256,9 @@ private List<String> processFieldParts(FieldKey fkTok, Map<String, NameExpressio
                     return fieldParts;
                 }
 
-                if (ancestorLevel > 4) // 1 generation of direct parent + 4 extra generations of ancestors
+                if (ancestorLevel > 9) // 1 generation of direct parent + 9 extra generations of ancestors
```
labkey-susanh, line 1259: The 9 here is somewhat arbitrary, right?  Might be worth a comment somewhere.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1271,7 +1271,7 @@ private List<String> processFieldParts(FieldKey fkTok, Map<String, NameExpressio
 
         if (!ancestorTypes.isEmpty())
         {
-            if (!hasLookupColumn)
+            if (!hasLookupColumn || fieldParts.size() < 2 || fieldParts.size() > 4)
```
labkey-susanh: Might want a comment here about the size limit as well and how it relates to the ancestorLevel.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1283,7 +1283,24 @@ private List<String> processFieldParts(FieldKey fkTok, Map<String, NameExpressio
...
+            String parentTypeName = null;
+            if (parentParts.size() == 1) // alias, or one of Inputs/, MaterialInputs/, DataInputs/
+            {
+                String alias = parentParts.get(0);
+                boolean isParentAlias = importAliases != null && importAliases.containsKey(alias);
+                if (isParentAlias)
+                {
+                    String dataTypeToken = importAliases.get(alias);
+                    parentTypeName = dataTypeToken.substring(dataTypeToken.indexOf("/") + 1);
+                }
```
labkey-susanh, line 1297: if `isParentAlias` is false (meaning we have something like "MaterialInputs/Blood"), `parentTypeName` will still be null, right?  Is that what we want?

labkey-susanh (approved):

api/src/org/labkey/api/data/NameGenerator.java
XingY: If not alias, then it's one of those general Inputs/, MaterialInputs/, DataInputs/, in which case we want to parentTypeName to be null to accept all types.

api/src/org/labkey/api/data/NameGenerator.java
XingY: it is in fact arbitrary... someone wanted 6, so Hannah decided to make it 10 lol

PR #4876, opened 2023-10-27
ElevatedUser

labkey-adam:
#### Rationale
Creating a fake user with strictly limited permissions and augmenting a real user's permissions with contextual roles are very different actions that deserve separate (but related) classes. They also deserve junit tests. New class `ElevatedUser` effectively wraps a user to correctly convey its underlying groups, impersonation status, roles, etc. Both classes now dictate the new user's security properties by setting `ImpersonationContext`, which is preferred vs. overriding `User` methods. ElevatedUser leverages existing class `WrappedImpersonationContext`, reconciling what were two separate approaches to escalating permissions.

This fixes `BiologicsManagerTest`, which was failing because SeqPart cache loading expected wrapping one `LimitedUser` with another to work.

Note: Deprecated "elevate user" methods in `LimitedUser` are left temporarily to decouple PRs in multiple repos. These will be removed shortly.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4865

### DIFF
--- a/api/src/org/labkey/api/security/ElevatedUser.java
+++ b/api/src/org/labkey/api/security/ElevatedUser.java
@@ -14,9 +14,9 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 /**
- * A wrapped user that possesses all the security aspects (groups, roles, impersonation status, etc.) of the underlying
- * user and adds one or more contextual roles. Use this class as a last resort; preference is to use contextual roles
- * in individual permissions checks or LimitedUser.
+ * A wrapped user that possesses all the security properties (groups, roles, impersonation status, etc.) of the
+ * underlying user and adds one or more contextual roles. Use this class as a last resort; preference is to use
+ * contextual roles in individual permission checks or LimitedUser.
  */
 public class ElevatedUser extends ClonedUser
 {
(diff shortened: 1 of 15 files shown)


api/src/org/labkey/api/security/ElevatedUser.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * A wrapped user that possesses all the security properties (groups, roles, impersonation status, etc.) of the
+ * underlying user and adds one or more contextual roles. Use this class as a last resort; preference is to use
+ * contextual roles in individual permission checks or LimitedUser.
+ */
```
labkey-matthewb, line 22: Maybe an additional comment that the creator of the LimitedUser is responsible for making sure the scope in which this object is used to evaluate permissions is constrained appropriately.

labkey-matthewb (approved):

api/src/org/labkey/api/security/ElevatedUser.java
labkey-adam: Agreed. I'll add this in my follow-on cleanup pass.

labkey-adam: Actually, added to this PR

PR #4877, opened 2023-10-27
fix typo

labkey-matthewb:
#### Rationale
missing ```nonce=```

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/core/src/org/labkey/core/login/setPassword.jsp
+++ b/core/src/org/labkey/core/login/setPassword.jsp
@@ -149,7 +149,7 @@
     <% } %>
     </div>
 </labkey:form>
-<script type="application/javascript" <%=getScriptNonce()%>>
+<labkey:script>
     let renderBarFunction;
 
     LABKEY.Utils.onReady(function() {
@@ -263,4 +263,4 @@
 
         return ratio;
     }
-</script>
+</labkey:script>


labkey-adam (approved):
So this wasn't correct? My FB passed the BVT and I thought BVTs ran with CSP in place.

labkey-matthewb:
> So this wasn't correct? My FB passed the BVT and I thought BVTs ran with CSP in place.

Maybe Trey switched from "enforce" to "report" so all the tests run with one configuration?

labkey-tchad:
BVTs are set to enforce. Enforcing just makes the javascript not execute; it would only cause a test failure if the javascript not executing would interfere with the tests. In this case, the difference is purely cosmetic as we don't yet test the strength meter.

labkey-matthewb:
> BVTs are set to enforce. Enforcing just makes the javascript not execute; it would only cause a test failure if the javascript not executing would interfere with the tests. In this case, the difference is purely cosmetic as we don't yet test the strength meter.

Right, I didn't actually look at what was in the <script> block.

labkey-adam:
> > BVTs are set to enforce. Enforcing just makes the javascript not execute; it would only cause a test failure if the javascript not executing would interfere with the tests. In this case, the difference is purely cosmetic as we don't yet test the strength meter.
> 
> Right, I didn't actually look at what was in the <script> block.

You should... there's excellent work in there! 😄 

Thanks for the clarification, Trey... that makes sense. Time for me to turn my local CSP back on.

PR #4879, opened 2023-10-27
Issue 41677: Include single field uniqueness constraint option in field editor

cnathe:
#### Rationale
Part 1 of https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41677

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1331

#### Changes
* Refactor for a shared setPropertyIndices() across domain kinds
* Include current domain indices info with getDomainDescriptor()
* Support for updating (add and remove) domain indices as part of the updateDomainDescriptor()
* Add domain kind property for allowUniqueConstraintProperties

### DIFF
--- a/api/src/org/labkey/api/exp/property/DomainKind.java
+++ b/api/src/org/labkey/api/exp/property/DomainKind.java
@@ -312,6 +312,7 @@ abstract public class DomainKind<T>  implements Handler<String>
     public boolean allowTextChoiceProperties() { return true; }
     public boolean allowSampleSubjectProperties() { return true; }
     public boolean allowTimepointProperties() { return false; }
+    public boolean allowUniqueConstraintProperties() { return false; }
     public boolean showDefaultValueSettings() { return false; }
 
     public List<String> getDisabledSystemFields(List<String> disabledSystemFields)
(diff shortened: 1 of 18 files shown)


api/src/org/labkey/api/exp/property/DomainKind.java
```diff
@@ -312,6 +312,7 @@ public boolean matchesTemplateXML(String templateName, DomainTemplateType templa
     public boolean allowTextChoiceProperties() { return true; }
     public boolean allowSampleSubjectProperties() { return true; }
     public boolean allowTimepointProperties() { return false; }
+    public boolean allowUniqueConstraintProperties() { return false; }
```
XingY, line 315: I assume other domains such as assays didn't support creating unique index using api prior to this change?

XingY (approved):

api/src/org/labkey/api/exp/property/DomainKind.java
cnathe: Correct. Only the domain kinds mentioned in the test plan were previously calling the domain.setPropertyIndices or even looking for "indices" during domain creation.

PR #4896, opened 2023-10-31
re-enable stricter type checks

labkey-matthewb:
#### Rationale
Re-enable the strict column type checking code in SchemaXMLTestCase

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/data/dialect/SqlDialect.java
+++ b/api/src/org/labkey/api/data/dialect/SqlDialect.java
@@ -1371,13 +1371,29 @@ public abstract class SqlDialect
                 StringUtils.equalsIgnoreCase("uniqueidentifier", sqlTypeName);
     }
 
+    public static boolean isJSONType(String sqlTypeName)
+    {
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/data/dialect/SqlDialect.java
```diff
@@ -1371,13 +1371,29 @@ public static boolean isGUIDType(String sqlTypeName)
...
-            if (isGUIDType(typeName))
-                t = JdbcType.GUID;
+            if (t == JdbcType.VARCHAR || t == JdbcType.CHAR)
+            {
+                if (isGUIDType(typeName))
+                    t = JdbcType.GUID;
+            }
+            else if (t == JdbcType.OTHER)
+            {
+                // CONSIDER is it useful to have an internal JSON type?
```
labkey-jeckels, line 1393: I thought about adding it when I did some of the earlier JSON work but obviously didn't go for it. It could help us provide better error messages for JSON-specific functions, etc

labkey-jeckels (approved):

PR #4897, opened 2023-10-31
Metric to track invocations of old version of GetUsersWithPermissions

labkey-adam:
#### Rationale
We want to track invocations of the old version of this API with an eye toward perhaps removing support for the old format

Also, switch `includeDeactivated` -> `includeInactive` in a few places

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4870


labkey-alan (approved):
Looks good!

PR #4903, opened 2023-11-02
Issues 48991, 48994: Improve file root behavior

labkey-jeckels:
#### Rationale
Things can go badly if the server's configured file root isn't available and we fall back to the default location

#### Changes
* Log when the configured file root isn't available (exactly once per web app, at WARN level)
* Log when we create the file root directory
* Report a new metric if the configured file root isn't available
* Show file root in Admin Console->Server Information

### DIFF
--- a/filecontent/src/org/labkey/filecontent/FileContentServiceImpl.java
+++ b/filecontent/src/org/labkey/filecontent/FileContentServiceImpl.java
@@ -69,6 +69,7 @@ import org.labkey.api.query.BatchValidationException;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.QueryUpdateService;
 import org.labkey.api.security.User;
+import org.labkey.api.services.ServiceRegistry;
 import org.labkey.api.settings.AppProps;
 import org.labkey.api.settings.RandomSiteSettingsPropertyHandler;
 import org.labkey.api.settings.StartupPropertyEntry;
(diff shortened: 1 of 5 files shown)


filecontent/src/org/labkey/filecontent/FileContentServiceImpl.java
```diff
@@ -589,11 +591,26 @@ public void setIsUseDefaultRoot(Container c, boolean useDefaultRoot)
...
+                _lastLoggedProblematicFileRoot = null;
+            }
 
             if (!root.exists())
             {
-                FileUtil.mkdirs(root);
+                if (FileUtil.mkdirs(root))
+                {
+                    _log.info("Created site-wide file root " + root);
+                }
```
labkey-susanh, line 626: Log error if `mkdirs` fails?

filecontent/src/org/labkey/filecontent/FileContentServiceImpl.java
```diff
@@ -589,11 +591,26 @@ public void setIsUseDefaultRoot(Container c, boolean useDefaultRoot)
         try
         {
             if (root == null || !root.exists())
+            {
+                File configuredRoot = root;
                 root = getDefaultRoot();
+                if (configuredRoot != null && !configuredRoot.equals(root) && !configuredRoot.equals(_lastLoggedProblematicFileRoot))
```
labkey-susanh: Any chance that the default root doesn't exist either?

labkey-susanh (approved):

filecontent/src/org/labkey/filecontent/FileContentServiceImpl.java
labkey-jeckels: Might as well. If we fail to create it, you'll also get other error messaging about the directory not existing when you try to view/write files.

filecontent/src/org/labkey/filecontent/FileContentServiceImpl.java
labkey-jeckels: We'll try to create it a few lines later. And now log if that fails

PR #4910, opened 2023-11-03
Avoid unneeded map copying

labkey-jeckels:
#### Rationale
Profiling shows that iterating through large result sets and converting rows to strongly typed beans churns a lot of memory copying into CaseInsensitiveHashMap. RowMap is already case insensitive, so that's a waste

#### Changes
* Introduce marker interface instead of checking for specific implementations
* Don't bother copying RowMap
* Misc code cleanup

### DIFF
--- a/api/src/org/labkey/api/collections/CaseInsensitiveArrayListValuedMap.java
+++ b/api/src/org/labkey/api/collections/CaseInsensitiveArrayListValuedMap.java
@@ -27,7 +27,7 @@ import java.util.List;
  *
  * Created by Nick on 10/1/2016.
  */
-public class CaseInsensitiveArrayListValuedMap<V> extends AbstractListValuedMap<String, V>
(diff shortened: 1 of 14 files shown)


labkey-matthewb (approved):
Passing thought.  Worth making a marker interface CaseInsensitiveMap?

labkey-adam:
> Passing thought. Worth making a marker interface CaseInsensitiveMap?

I like that... maybe just `CaseInsensitive` so it can apply to case-insensitive sets and multi-maps as well

labkey-jeckels:
@labkey-matthewb @labkey-adam feel free to check my work. I introduced the marker interface and tagged a bunch of collections.

labkey-adam:
> @labkey-matthewb @labkey-adam feel free to check my work. I introduced the marker interface and tagged a bunch of collections.

Looks good to me... thanks!

PR #4913, opened 2023-11-03
Issue 48999: Unable to "Select them in the grid" when created trigger script runs after samples are added

XingY:
#### Rationale
Trigger script can call [labkey/api/Query](https://labkey.github.io/labkey-api-js/modules/Query.html) to additionally do insert/update/delete at any of the before/after calls. Those nested actions do belong to the same transaction and shouldn't trigger separate creation of TransactionAuditEvents. Changes made in this PR to attach the transactionAuditEvent on Transaction, so it can be reused. 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1340
- https://github.com/LabKey/labkey-ui-premium/pull/246
- https://github.com/LabKey/platform/pull/4913
- https://github.com/LabKey/biologics/pull/2501
- https://github.com/LabKey/sampleManagement/pull/2220

#### Changes
* add get/setAuditEvent to Transaction
* add addComment function to allow appending of comment per action
* check existing auditEvent for query actions and import actions
* add reselectRowCount flag in response to indicate the returned rowCount might not reflect the total rows affected

### DIFF
--- a/api/src/org/labkey/api/audit/TransactionAuditProvider.java
+++ b/api/src/org/labkey/api/audit/TransactionAuditProvider.java
@@ -1,5 +1,7 @@
 package org.labkey.api.audit;
 
+import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.Nullable;
 import org.labkey.api.audit.query.AbstractAuditDomainKind;
 import org.labkey.api.audit.query.DefaultAuditTypeTable;
 import org.labkey.api.data.ContainerFilter;
@@ -102,7 +104,7 @@ public class TransactionAuditProvider extends AbstractAuditTypeProvider implemen
         private QueryService.AuditAction _auditAction;
         private String _transactionType;
 
-        private int rowCount;
+        private boolean _multiActions; // if the audit event comment is updated/appended multiple times, for example, the original insert triggers additional insert/update via trigger scripts
 
         public TransactionAuditEvent()
         {
@@ -111,7 +113,7 @@ public class TransactionAuditProvider extends AbstractAuditTypeProvider implemen
 
         public TransactionAuditEvent(String container, QueryService.AuditAction auditAction, long transactionId)
         {
-            super(EVENT_TYPE, container, String.format(auditAction.getCommentSummary(), "Some"));
+            super(EVENT_TYPE, container, auditAction.getDefaultCommentSummary());
             _auditAction = auditAction;
             _transactionType = auditAction.name();
             _startTime = new Date();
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/audit/TransactionAuditProvider.java
```diff
@@ -138,15 +140,36 @@ public void setTransactionType(String transactionType)
             _transactionType = transactionType;
         }
 
-        public int getRowCount()
+        public void setRowCount(int rowCount)
```
labkey-susanh: Since we're getting rid of `rowCount`, I might call this `updateComment` or something like that.

api/src/org/labkey/api/audit/TransactionAuditProvider.java
```diff
@@ -138,15 +140,36 @@ public void setTransactionType(String transactionType)
...
-            return rowCount;
+            setComment(String.format(_auditAction.getCommentSummary(), rowCount));
         }
 
-        public void setRowCount(int rowCount)
+        public void addComment(@Nullable QueryService.AuditAction action, int rowCount)
+        {
+            String existingComment = this.getComment();
+            QueryService.AuditAction newAction = action == null ? _auditAction : action;
+            boolean isDefaultOrNullComment = StringUtils.isEmpty(existingComment) || newAction.getDefaultCommentSummary().equals(existingComment);
```
labkey-susanh: Doing string comparisons probably works here, but seems potentially fragile. I would probably add a simple counter here that is incremented with each call.  This could replace the `_multiActions` member as you could then check the value of the counter instead to know if more than one comment has been added.

It seems a little side-effect-y that we rely on calls to `addComment` to know that multiple inserts have happened. It also seems possible that a trigger script would, say, remove some rows as well.

XingY: good point. changed to _commentCount instead. Yes trigger script can do anything and the comment will not assume subsequence action matches the originating action, rather, it would append to the previous comment.

api/src/org/labkey/api/audit/TransactionAuditProvider.java
```diff
@@ -138,15 +140,40 @@ public void setTransactionType(String transactionType)
             _transactionType = transactionType;
         }
 
-        public int getRowCount()
+        @Override
+        public void setComment(String comment)
+        {
+            if (!StringUtils.isEmpty(comment))
```
labkey-susanh: I can't think what the case is that someone calls setComment with an empty comment (and should this be asking about the trimmed value?), but if this is being called for any reason, should that count as one comment for purposes of understanding what's gone on in the transaction?

XingY: Realistically this empty check is probably not needed. I added it so if someone initialize this to null it wouldn't count as 1, but that's not likely. Removed the empty check

labkey-susanh (approved):

PR #4915, opened 2023-11-04
Issue 49035: DELETE from exp.edges during sample type deletion never completes

labkey-jeckels:
#### Rationale
Deleting from a 1.5 million row sample type took more than 24 hours to delete from `exp.edge` with the current code (and was still cranking). Switching to use `ObjectId` in the temp table instead of `ObjectURI` and using three separate `DELETE` statements is *way* faster in local testing with a restore of the DB in question.

#### Changes
* Use `ObjectId` in temp table. It's 3x faster to create the temp table, and for most uses, faster in queries that reference it
* Support deleting directly by `ObjectId` SQL in `OntologyManager`
* Improve generics a bit to avoid a cast

### DIFF
--- a/api/src/org/labkey/api/exp/OntologyManager.java
+++ b/api/src/org/labkey/api/exp/OntologyManager.java
@@ -975,8 +975,6 @@ public class OntologyManager
 
     public static int deleteOntologyObjectsByObjectIdSql(DbSchema schema, SQLFragment objectIdSql)
     {
-        // we have different levels of optimization possible here deleteOwned=true/false, scope=/<>exp
-
         if (!schema.getScope().equals(getExpSchema().getScope()))
             throw new UnsupportedOperationException("can only use with same DbScope");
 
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -958,50 +957,41 @@ public static void deleteOntologyObjects(Container c, String... uris)
...
+        }
+        objectIdSQL.append("ObjectUri IN (");
+        objectIdSQL.append(objectUriSql);
+        objectIdSQL.append(")");
+        return deleteOntologyObjectsByObjectIdSql(schema, objectIdSQL);
+    }
+
+    public static int deleteOntologyObjectsByObjectIdSql(DbSchema schema, SQLFragment objectIdSql)
     {
         // we have different levels of optimization possible here deleteOwned=true/false, scope=/<>exp
```
labkey-susanh: The comment now seems out of date.

labkey-susanh (approved):

api/src/org/labkey/api/exp/OntologyManager.java
labkey-jeckels: Yes, it's even less helpful now than it was before

PR #4924, opened 2023-11-06
Sample deletion test fix - stop deleting rows from other containers

labkey-jeckels:
#### Rationale
My commit last night to optimize sample deletion misinterpreted the `isTruncate` argument as indicating the whole sample type was being truncated. It's actually just being "truncated" for that single container. Thus, a DB-level `TRUNCATE` is not appropriate, as it will clear out rows from other containers too.

https://teamcity.labkey.org/buildConfiguration/LabKey_2311Release_Premium_ProductSuites_SampleManager_SampleManagerCPostgres/2733502?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildTestsSection=true&expandBuildChangesSection=true

#### Changes
* Revert the change to use `TRUNCATE`. It's faster, but not doing what we need


labkey-jeckels:
The tests are now passing the FB.

https://teamcity.labkey.org/buildConfiguration/LabKey_2311Release_Premium_ProductSuites_SampleManager_SampleManagerCPostgres/2733761?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true

labkey-matthewb (approved):

PR #4937, opened 2023-11-08
Prevent multiple webapps pointed at the same database

labkey-adam:
#### Rationale
Our customers occasionally point multiple instances of LabKey Server at the same database; this never goes well and often results in problems that are difficult to diagnose.

This PR adds detection at server startup of any existing LabKey Server instance(s) operating on the same database; if found, the more recent server disables itself to prevent damage and alert administrators. Approach:

* On the very first connection to the primary database, query the application name.
* If the query returns the driver's default name then set "LabKey Server" as the application name on all subsequent connections, otherwise allow the custom name.
* On the next connection, query for other connections to the same database with this instance's application name.
* If any other connections exist, abort server startup. 

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46149

```
org.labkey.api.util.ConfigurationException: There are 3 connections to database "labkey23.11" with the application name "LabKey Server"! This likely means another LabKey Server deployment is already using this database.
	org.labkey.api.data.DbScope.ensureDatabase(DbScope.java:1705)
	org.labkey.api.data.DbScope.initializeDataSources(DbScope.java:1581)
	org.labkey.api.module.ModuleLoader.doInit(ModuleLoader.java:527)
	org.labkey.api.module.ModuleLoader.init(ModuleLoader.java:254)
	org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:171)
	org.apache.catalina.startup.HostConfig.deployDescriptor(HostConfig.java:689)
```

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -307,9 +307,9 @@ public class DbScope
         private final String _driverClassName;
         private final SqlDialect _dialect;
         private final Class<Driver> _driverClass;
-        private final @Nullable String _applicationName;
         private final String _url;
 
+        private @Nullable String _applicationName; // Null if a custom application name is set (JDBC URL) or initial connection hasn't been made yet
         private boolean _logQueries = false;
         private String _displayName = null;
         private boolean _primary = false;
@@ -323,7 +323,6 @@ public class DbScope
             _driverClassName = null;
             _dialect = null;
             _driverClass = null;
-            _applicationName = null;
             _url = null;
         }
 
@@ -337,7 +336,6 @@ public class DbScope
             _dialect = SqlDialectManager.getFromDriverClassname(_dsName, _driverClassName);
             MemTracker.get().remove(_dialect);
             _driverClass = initializeDriver();
-            _applicationName = ensureApplicationName();
             _url = _dsPropertyReader.getUrl();
 
             // Validate that data source is using a supported connection pool
@@ -367,60 +365,63 @@ public class DbScope
             _displayName = displayName;
         }
 
-        public DataSource getDataSource()
+        DataSource getDataSource()
         {
             return _ds;
         }
 
-        public String getDsName()
+        private String getDsName()
         {
             return _dsName;
         }
 
-        public DataSourcePropertyReader getDataSourcePropertyReader()
+        DataSourcePropertyReader getDataSourcePropertyReader()
         {
             return _dsPropertyReader;
         }
 
-        public String getDriverClassName()
+        private String getDriverClassName()
         {
             return _driverClassName;
         }
 
-        public SqlDialect getDialect()
+        private SqlDialect getDialect()
         {
             return _dialect;
         }
 
-        public Class<Driver> getDriverClass()
+        private Class<Driver> getDriverClass()
         {
             return _driverClass;
         }
 
-        public String getUrl()
+        private String getUrl()
         {
             return _url;
         }
 
-        public void setPrimary(boolean primary)
+        private void setPrimary()
         {
-            _primary = primary;
+            _primary = true;
         }
 
-        public boolean isPrimary()
+        private boolean isPrimary()
         {
             return _primary;
         }
 
-        public @Nullable String getApplicationName()
+        private @Nullable String getApplicationName()
         {
             return _applicationName;
         }
 
-        // Ensure an application name has been specified for this data source
-        private @Nullable String ensureApplicationName()
+        // Set the default application name for all connections on this data source
+        private String setDefaultApplicationName()
         {
-            String applicationName = null;
+            // Used by getRawConnection()
+            _applicationName = DEFAULT_APPLICATION_NAME;
+
+            // Push application name into the connection properties are used to create pooled connections
             Properties connectionProps = _dsPropertyReader.getConnectionProperties();
 
             if (connectionProps != null)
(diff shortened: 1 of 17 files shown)


api/src/org/labkey/api/data/dialect/SqlDialect.java
```diff
@@ -1291,6 +1285,21 @@ public Long getMaxWaitMillis()
                 return null;
             }
         }
+
+        public Properties getConnectionProperties()
```
labkey-jeckels: Annotate as nullable? Or throw an UnexpectedException instead?

api/src/org/labkey/api/data/DbScope.java
```diff
@@ -323,22 +365,161 @@ public void setDisplayName(String displayName)
...
+        // Ensure an application name has been specified for this data source
+        private @Nullable String ensureApplicationName()
+        {
+            Properties connectionProps = _dsPropertyReader.getConnectionProperties();
+            String paramName = _dialect.getApplicationNameParameter();
+            String applicationName = null;
+
+            if (paramName != null)
+            {
+                applicationName = (String) connectionProps.get(paramName);
```
labkey-jeckels:
```suggestion
                applicationName = connectionProps.getProperty(paramName);

```

api/src/org/labkey/api/data/DbScope.java
```diff
@@ -1596,78 +1691,77 @@ public static boolean ensureDataBase(String dsName, DataSource ds) throws Servle
...
-                // Load and prepare the JDBC driver
-                @SuppressWarnings("unchecked")
-                Class<Driver> driverClass = (Class<Driver>)Class.forName(props.getDriverClassName());
-                dialect.prepareDriver(driverClass);
+                LOG.debug("Successful connection to \"" + ds.getDsName() + "\" at " + ds.getUrl());
+
+                // Very first connection (and we haven't sent the application name for our connection), so ask the
+                // dialect how to count connections with our database and application name.
+                String dbName = dialect.getDatabaseName(ds.getUrl());
+                int count = dialect.getApplicationConnectionCount(conn, dbName, ds.getApplicationName());
```
labkey-jeckels: There's a chance that some deployments will have permissions locked down such that we can't run the query for all connections. Maybe we should catch SQLException here and log instead of treating it as fatal? Or have a way to opt out of the check (system property, data source setting, etc)?

labkey-jeckels (approved):
Looks good. One potential concern about permissions to query for connections. See comment on the call to get the count.

api/src/org/labkey/api/data/dialect/SqlDialect.java
labkey-adam: Added annotation and null checks

api/src/org/labkey/api/data/DbScope.java
labkey-adam: Thanks

api/src/org/labkey/api/data/DbScope.java
labkey-adam: We now use straight JDBC to make this early query (SqlSelector exception handling wasn't tolerant of scopes not being initialized yet), catch SQLException, and log a one-line warning.

labkey-jeckels (approved):

api/src/org/labkey/api/data/DbScope.java
```diff
@@ -1768,15 +1716,104 @@ public static void ensureDatabase(LabKeyDataSource ds) throws ServletException
...
+            try (PreparedStatement stmt = conn.prepareStatement(dialect.getApplicationNameSql()); ResultSet rs = stmt.executeQuery())
+            {
+                if (rs.next())
+                {
+                    applicationName = rs.getString(1);
+                    String message = "Application name detected on first database connection: \"" + applicationName + "\"";
+
+                    // If connection shows the default application name (i.e., application name is not set on the JDBC URL)
+                    // then set our default application name on the data source so it's set on every connection.
+                    if (applicationName.equals(dialect.getDefaultApplicationName()))
```
labkey-jeckels, line 1784: Will `applicationName` always be non-null at this point?

api/src/org/labkey/api/data/DbScope.java
```diff
@@ -1768,15 +1716,104 @@ public static void ensureDatabase(LabKeyDataSource ds) throws ServletException
...
+                    else
+                    {
+                        throw new IllegalStateException("Application connection count query returned no rows!");
+                    }
+                }
+            }
+        }
+        catch (SQLException | ServletException e)
+        {
+            LOG.warn("Attempt to detect other LabKey Server instances using this database failed: " + e.getMessage());
```
labkey-jeckels: Unless there's a good reason to exclude it, I'd vote for capturing the full stack.

I'm not getting the UI to easily suggest an alternative line, but I mean:

`LOG.warn("Attempt to detect other LabKey Server instances using this database failed", e);`

labkey-jeckels (approved):

api/src/org/labkey/api/data/DbScope.java
labkey-adam: I'm fine either way. I was trying to avoid making it seem like it's the end of the world and must be fixed, like most stack traces.

api/src/org/labkey/api/data/DbScope.java
labkey-adam: Should be... but I can add an assert. Because I know how much you like asserts instead of NPEs.

PR #4941, opened 2023-11-09
Metric to assess context path variability across deployments

labkey-jeckels:
#### Rationale
To help understand how the web app is deployed in practice, let's collect info on what context paths people are using.

#### Changes
* New 'webappContextPath` metric


labkey-adam (approved):
This is lovely as is. However, it would be great to also collect the other (non-LabKey) context paths, to identify clients who may have difficulty migrating to embedded. `CoreWarningProvider.collectAllDeployedApps()` and then remove the LabKey context should do the trick.

labkey-jeckels:
> This is lovely as is. However, it would be great to also collect the other (non-LabKey) context paths, to identify clients who may have difficulty migrating to embedded. `CoreWarningProvider.collectAllDeployedApps()` and then remove the LabKey context should do the trick.

I had a similar thought but forgot that we already had the code to query the other web apps, so I stopped short. Thanks for the pointer. I pushed code that adds a second metric, the comma-separated non-LK context paths.

PR #4948, opened 2023-11-10
fix inline event

labkey-matthewb:
#### Rationale
The crawler likes this page

#### Related Pull Requests
* 

#### Changes
*


labkey-adam:
I migrated this drop-down to `SelectBuilder`, which I think is even cleaner, yesterday, but neglected to create the PR. https://github.com/LabKey/platform/pull/4949

labkey-adam (approved):

PR #4949, opened 2023-11-10
Eliminate old helpPopup() method and fix specimen actions

labkey-adam:
#### Rationale
Use the builder directly and eliminate some inline script complaints

### DIFF
--- a/specimen/src/org/labkey/specimen/actions/SpecimenController.java
+++ b/specimen/src/org/labkey/specimen/actions/SpecimenController.java
@@ -99,6 +99,8 @@ import org.labkey.api.util.MailHelper;
 import org.labkey.api.util.PageFlowUtil;
 import org.labkey.api.util.Pair;
 import org.labkey.api.util.URLHelper;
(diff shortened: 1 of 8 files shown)


specimen/src/org/labkey/specimen/actions/SpecimenController.java
```diff
@@ -528,28 +530,20 @@ public void render(RenderContext ctx, Writer out) throws IOException
...
+
             out.write("Hide Previously Requested By ");
-            out.write("<select ");
-            out.write("onchange=\"window.location=");
-            out.write(PageFlowUtil.jsString(url + "&amp;excludeRequestedBySite="));
-            out.write(" + this.options[this.selectedIndex].value;\"");
-            out.write(">");
-            out.write("<option value=''>&lt;Show All&gt;</option>");
+
             String excludeStr = ctx.getRequest().getParameter(SpecimenQueryView.PARAMS.excludeRequestedBySite.name());
```
labkey-matthewb, line 536: So much better

labkey-matthewb (approved):

PR #4951, opened 2023-11-11
Address specimen/study inline

labkey-adam:
#### Rationale
Migrate `specimen-manageRequestInputs`, `specimen-specimens`, and `study-manageQCStates` away from inline script

### DIFF
--- a/api/src/org/labkey/api/qc/view/manageQCStates.jsp
+++ b/api/src/org/labkey/api/qc/view/manageQCStates.jsp
@@ -138,24 +138,24 @@
 </labkey:form>
 <script type="text/javascript" nonce="<%=getScriptNonce()%>">
     function addRow() {
-        var table = document.getElementById("qcStatesTable");
-        var numberOfRows = table.rows.length;  // we'll use this as an ID later too
-        var newRow = table.insertRow(numberOfRows);
-        var nameCell = newRow.insertCell(newRow.cells.length);
-        var descriptionCell = newRow.insertCell(newRow.cells.length);
-        var publicDataCell = newRow.insertCell(newRow.cells.length);
-        var inUseCell = newRow.insertCell(newRow.cells.length);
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/qc/view/manageQCStates.jsp
```diff
@@ -150,7 +150,8 @@
         descriptionCell.innerHTML = '<input type="text" name="newDescriptions" size="50"></td>';
         publicDataCell.innerHTML = '<input type="checkbox" value="' + numberOfRows + '" id="' + numberOfRows +'_public" name="newPublicData" CHECKED>';
         publicDataCell.style.textAlign = 'center';
-        inUseCell.innerHTML = '<span style="color:black;padding-left: 30%;" class="fa fa-circle-o"></span> <span style="color:red;" class="fa fa-times" onclick="deleteRow(this)"></span>';
+        inUseCell.innerHTML = '<span style="color:black;padding-left: 30%;" class="fa fa-circle-o"></span> <span style="color:red;" class="fa fa-times"></span>';
+        inUseCell.childNodes[2]['onclick'] = function(){ deleteRow(this) };
```
labkey-matthewb: I think this is the same as: 
inUseCell.childNodes[2]['onclick'] = deleteRow;

labkey-matthewb (approved):

api/src/org/labkey/api/qc/view/manageQCStates.jsp
labkey-adam: Just about... function is passed the event instead of the element, but that's an easy adjustment

labkey-matthewb: Right, now the same, since deleteRow() would need to look at **this**, intead of using **deleteButton** argument.  Not worth changing.

PR #4954, opened 2023-11-13
Query for other LabKey instances on a pooled connection

labkey-adam:
#### Rationale
During startup, there are very short windows when the server has no connections to the database. These windows could allow two servers starting in parallel to not detect each other. Moving the detection code to a pooled connection should eliminate these races. First server to create a pooled connection will likely succeed and leave that connection open (in the pool) for the second server to detect (and fail startup). It's possible the servers will create their initial pooled connections simultaneously, in which case both will fail to start up. That's certainly preferable to letting them both start up.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4937

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -1729,45 +1729,32 @@ public class DbScope
     {
         assert _applicationName != null;
 
-        // First pooled connection on the primary scope. Application name should be set on this connection in both the
-        // default and custom cases
-        try (Connection conn = primaryScope.getConnection())
+        SqlDialect dialect = primaryScope.getSqlDialect();
+        String databaseName = primaryScope.getDatabaseName();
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
This change seems OK but is getting very close to the wire for 23.11.0 for such a critical codepath. I don't know how much testing we've gotten with and without this change yet, so I'll defer the final decision on when to merge.

labkey-matthewb (approved):

labkey-adam:
> This change seems OK but is getting very close to the wire for 23.11.0 for such a critical codepath. I don't know how much testing we've gotten with and without this change yet, so I'll defer the final decision on when to merge.

Matt has been testing... he discovered the race condition and has tested with the change on PG. I think he's planning SS next. We can discuss in standup tomorrow to make the merge call.

PR #4957, opened 2023-11-14
purgeParticipantVisit()

labkey-matthewb:
#### Rationale
fixes related to https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49028
fix "manage visits/recalculate visit dates"
add purgeParticipantVisit() to said page
fix performance on "manage visits/delete unused visits" (getUnusedVisits())

#### Related Pull Requests
* 

#### Changes
*


labkey-matthewb:
Note, this implies that you would want to execute "recalculate visit dates" before executing "delete unused visits".  That's not obvious.  We do this automatically, but it would have to be done in getConfirmView() which is a bit odd.

labkey-klum (approved):
I also did some manual testing using the export from Joe's repro.

PR #4964, opened 2023-11-15
Don't fire container events before fireCreateContainer()

labkey-matthewb:
#### Rationale
This is in an attempt to reduce unnecessary work (and database queries) that may contribute to deadlocks. 

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels (approved):
I checked many but not all ContainerListener implementations for those that might be relying on the property change to trigger for new containers. CoreContainerListener was the only one that seemed suspect, which you already handled.

PR #4978, opened 2023-11-22
Remove references to Group.groupAdministrators

labkey-adam:
#### Rationale
Most code that directly references the "Site Admin" group is wrong; typically, it should be checking for `SiteAdminPermission` or assigning the `SiteAdminRole` instead.

In the process, remove pointless code, simplify some overly complex handling, add and use a standard method for enumerating site admins, fix null handling for "primary admin," and add a test to confirm some assumptions.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48879

### DIFF
--- a/api/src/org/labkey/api/security/UserManager.java
+++ b/api/src/org/labkey/api/security/UserManager.java
@@ -20,7 +20,9 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.junit.AfterClass;
 import org.junit.Assert;
+import org.junit.BeforeClass;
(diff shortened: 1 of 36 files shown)


api/src/org/labkey/api/security/UserManager.java
```diff
@@ -1286,4 +1289,49 @@ public static ActionURL getUserDetailsURL(Container container, User currentUser,
 
         return null;
     }
+
+    public static class TestCase extends Assert
+    {
+        @Test
+        public void testPermissionsRetrieval()
+        {
+            List<User> appAdmins = getAppAdmins();
```
labkey-jeckels, line 1324: Should the test make a test user an app admin and site admin to make sure there's something to actually verify?

labkey-jeckels (approved):

PR #4844, opened 2023-10-20
Sample Types: Add RootMaterialRowId

labkey-nicka:
#### Rationale
This introduces a new `RootMaterialRowId` column into the `exp.material` table which emulates the `RootMaterialLSID` column. This column identifies the root material for a sample which is useful for derivates and aliquots. Additionally, in materialized Sample Type tables the `RowId` for the sample is now replicated into the materialized table. When joining to materialized Sample Type tables `RootMaterialRowId` is now used instead of `RootMaterialLSID` which has been shown to result in moderate to significant gain in query execution times for Sample Types with a large number of samples.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4844
* https://github.com/LabKey/labkey-ui-components/pull/1324
* https://github.com/LabKey/labkey-ui-premium/pull/230
* https://github.com/LabKey/biologics/pull/2464
* https://github.com/LabKey/sampleManagement/pull/2182
* https://github.com/LabKey/inventory/pull/1070

#### Changes
- Upgrade SQL scripts to add `RootMaterialRowId` column to `exp.material` table.
- Upgrade code to add `RowId` to materialized Sample Type tables.
- Refactor `SampleTypeService.recomputeSampleTypeAvailableAliquotRollup` into `ExperimentUpgradeCode` as it is only utilized during upgrade. Remove from service interface.
- Update XAR export/import to translate rowId to LSID and back again for `RootMaterialRowId`.
- Update aliquot recalculation to utilize RowId instead of LSID.
- Remove `RootMaterialLSID`.
- Bump `@labkey` packages

PR #4979, opened 2023-11-22
Issue 49150: Fix sample type upgrade from 23.07 to 23.12

labkey-nicka:
#### Rationale
This seeks to address [Issue 49150](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49150) where an upgrade directly from a 23.07 (or earlier) instance to 23.12 can fail due to some shared code that generates queries using `RootMaterialLSID`/`RootMaterialRowId`. This updates the shared code to accept an additional argument to switch the target column for the joins. The intent is for this to be effectively reverted when the upgrade path goes defunct with the release of 25.2.0.

#### Related Pull Requests
* #4844 (upgrade that introduces `RootMaterialRowId`)

#### Changes
* Introduce `useRootMaterialLSID` and generate subsequent JOINs in queries based off this flag.

### DIFF
--- a/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
+++ b/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
@@ -357,12 +357,15 @@ public class ExperimentUpgradeCode implements UpgradeCode
         List<Integer> rootSamplesWithAvailableAliquotVolume = new ArrayList<>(s1);
 
         // Exposed as "public" only for upgrade. When removing this code make this signature "private".
+        // Issue 49150: Additionally, remove the "useRootMaterialLSID" flag and all of it's related logic as it
+        // is only necessary for this upgrade path.
         return SampleTypeServiceImpl.get().recomputeSamplesRollup(
             Collections.emptyList(),
             rootSamplesWithAvailableAliquot,
             rootSamplesWithAvailableAliquotVolume,
             sampleType.getMetricUnit(),
-            container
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -357,12 +357,15 @@ private static int recomputeSampleTypeAvailableAliquotRollup(ExpSampleType sampl
         List<Integer> rootSamplesWithAvailableAliquotVolume = new ArrayList<>(s1);
 
         // Exposed as "public" only for upgrade. When removing this code make this signature "private".
+        // Issue 49150: Additionally, remove the "useRootMaterialLSID" flag and all of it's related logic as it
```
labkey-susanh, line 360:
```suggestion
        // Issue 49150: Additionally, remove the "useRootMaterialLSID" flag and all of its related logic as it
```

labkey-susanh (approved):

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
labkey-nicka: I'll change this in a subsequent PR.

labkey-adam:
@labkey-nicka FYI this upgrade code will actually get removed in 25.2.0. We've switched to an annual consolidation process that clears out old upgrade code every January.

PR #4983, opened 2023-11-28
Regression test for password strength gauge

labkey-tchad:
#### Rationale
I believe this was intended to set the placeholder text that shows up if canvas is disabled/unavailable. If so, `innerText` is the correct thing to set.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1736

#### Changes
* Use `innerText` to set canvas placeholder text

### DIFF
--- a/core/src/org/labkey/core/login/setPassword.jsp
+++ b/core/src/org/labkey/core/login/setPassword.jsp
@@ -225,7 +225,7 @@
                     const textIndex = Math.floor(percent * 6);
                     const text = showPlaceholderText ?  "Password Strength Gauge" : ["Very Weak", "Very Weak", "Weak", "Weak", "Strong", "Very Strong"][textIndex];
                     ctx.fillText(text, centerX, centerY + textHeightFix);
-                    canvas.innerText = showPlaceholderText ? text : ("Password Strength: " + text);
+                    canvas.innerText = text;
                 })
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/login/setPassword.jsp
```diff
@@ -225,7 +225,7 @@
                     const textIndex = Math.floor(percent * 6);
                     const text = showPlaceholderText ?  "Password Strength Gauge" : ["Very Weak", "Very Weak", "Weak", "Weak", "Strong", "Very Strong"][textIndex];
                     ctx.fillText(text, centerX, centerY + textHeightFix);
-                    canvas.text = "Password Strength: " + text;
+                    canvas.innerText = showPlaceholderText ? text : ("Password Strength: " + text);
```
labkey-adam: Maybe it should just assign `text`? I think this line preceded the placeholder text and wasn't updated after it was introduced.

labkey-adam (approved):

core/src/org/labkey/core/login/setPassword.jsp
labkey-tchad: Sure. That simplifies the test code a bit and the tests are almost certainly the only thing that will ever see this text.

PR #4992, opened 2023-11-29
Issue 49099: resolve material input/output container scope via ExperimentService

labkey-nicka:
#### Rationale
This addresses [Issue 49099](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49099) by fixing the resolution of container scope for input/output materials to always rely on `ExperimentService` to resolve the material based on container configuration. Previously, the assay run creator utilized `ExpSchema.getSearchContainers` which does not hold with our expectations for scoping of input/output materials. Instead, we will now leverage `ExperimentService.getExpMaterial()` resolves the container scope internally via `getContainerFilterTypeForFind` which will only return a material if the material is found in the configured container scope.

#### Changes
* Remove `ExpSchema.getSearchContainers` container scoping for assay material input/output and always rely on `ExperimentService` to resolve the material with the configured container scope.

### DIFF
--- a/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
+++ b/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
@@ -606,8 +606,7 @@ public class DefaultAssayRunCreator<ProviderType extends AbstractAssayProvider> 
         @NotNull Map<Integer, ExpMaterial> materialCache
     ) throws ExperimentException, ValidationException
     {
-        Set<Container> searchContainers = ExpSchema.getSearchContainers(context.getContainer(), null, null, context.getUser());
-        addMaterials(context, inputMaterials, context.getInputMaterials(), searchContainers, cache, materialCache);
+        addMaterials(context, inputMaterials, context.getInputMaterials(), cache, materialCache);
 
         // Find lookups to a SampleType and add the resolved material as an input sample
         for (Map.Entry<DomainProperty, String> entry : context.getRunProperties().entrySet())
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
```diff
@@ -606,8 +606,7 @@ protected void addInputMaterials(
         @NotNull Map<Integer, ExpMaterial> materialCache
     ) throws ExperimentException, ValidationException
     {
-        Set<Container> searchContainers = ExpSchema.getSearchContainers(context.getContainer(), null, null, context.getUser());
-        addMaterials(context, inputMaterials, context.getInputMaterials(), searchContainers, cache, materialCache);
+        addMaterials(context, inputMaterials, context.getInputMaterials(), cache, materialCache);
```
XingY, line 609: For this usage it's probably sufficient to rely on ExperimentService.findExpMaterial to use the correct cf. But should ExpSchema.getSearchContainers be updated to respect the experimental flags as well?

XingY (approved):
A question about ExpSchema.getSearchContainers

api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
labkey-nicka: I wanted to avoid scope creep and changing all the other usages of that method. I'd probably prefer to just get rid of it when we can.

PR #4995, opened 2023-11-29
experiment cycle check

labkey-matthewb:
#### Rationale
We have an API to report cycles in the experiment graph.  This is an action that presents the same info in HTML (with links).

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
+++ b/experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
@@ -58,6 +58,7 @@ import org.labkey.api.attachments.BaseDownloadAction;
 import org.labkey.api.audit.AuditLogService;
 import org.labkey.api.audit.TransactionAuditProvider;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
+import org.labkey.api.data.Container;
 import org.labkey.api.data.*;
 import org.labkey.api.dataiterator.DataIteratorContext;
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -7837,4 +7841,91 @@ protected void updateSelections(MoveEntitiesForm form)
...
+        @Override
+        public boolean handlePost(Object o, BindException errors) throws Exception
+        {
+            var edges = new SqlSelector(ExperimentService.get().getSchema(), "SELECT fromObjectId, toObjectId FROM exp.Edge")
+                    .resultSetStream()
+                    .map(r -> { try { return new Pair<>(r.getInt(1), r.getInt(2)); } catch (SQLException x) { throw new RuntimeException(x); } })
+                    .collect(toList());
+            var cyclesEdges = (new GraphAlgorithms<Integer>()).detectCycleInDirectedGraph(edges);
+
+            var set = new LinkedHashSet<Integer>();
```
labkey-nicka, line 7910: Seems like it'd be helpful to know the `cyclesEdges` grouping themselves. Possibly consider breaking this down in the display as different groupings?

labkey-nicka (approved):
Nice. It's certainly a start. I could imagine the view supporting showing the edges grouped together. Additionally, seems like it could be wired into GraphVis DOT graph pretty quickly.

PR #4996, opened 2023-11-29
separate list of active vs waiting threads in thread dump

labkey-matthewb:
#### Rationale
This should make scanning mothership reports easier.  Note if/when we think this is working exactly as intended we could consider not dumping the "waiting" threads.

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels (approved):
Nice. I've wanted this in the past, both as a convenience for those well versed in interpreting thread dumps and to help orient the less savvy.

PR #5003, opened 2023-12-01
Don't render bad link for mismatched DbSchema/UserSchema names

labkey-jeckels:
#### Rationale
The crawler found a bad link for the raw schema metadata in a data class table. It's a mishmash of DbSchema and UserSchema metadata and naming.

Example: https://teamcity.labkey.org/buildConfiguration/LabKey_2311Release_Community_DailySuites_DailyAPostgres/2760149?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildTestsSection=true&expandBuildChangesSection=true

#### Changes
* Don't offer a link when the UserSchema isn't exposed under the same name.

### DIFF
--- a/query/src/org/labkey/query/controllers/QueryController.java
+++ b/query/src/org/labkey/query/controllers/QueryController.java
@@ -1648,8 +1648,13 @@ public class QueryController extends SpringActionController
             {
                 VBox result = new VBox();
 
-                ActionURL url = new ActionURL(RawSchemaMetaDataAction.class, getContainer());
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-matthewb @labkey-adam let me know if you think we need to expose the `expdataclass` schema raw metadata somehow.

labkey-adam (approved):
Tests are suggesting another null check

labkey-matthewb (approved):

PR #4969, opened 2023-11-17
Upgrade TinyMCE and Wiki

Sigmonia:
#### Rationale
Clean up a lot of CSP errors, and upgrade the TinyMCE editor that we've been using forever. Pseudo-spec: https://docs.google.com/document/d/1U0rXcOUUWnJVPFeFTHX0YMNzSej-pu2AKInPGPYoA9Q/edit

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/1731

#### Changes
* Swap Tiny_Mce (v3) for TinyMce (v6)
* Update `Wiki` module to use new version
* Toolbar/Menu updates

PR #5004, opened 2023-12-01
Fix the Wiki crawler tests

Sigmonia:
#### Rationale
Fix the crawler tests that are complaining about the TinyMCE's logo attribute

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4969

#### Changes
* Applied the Branding attribute


labkey-tchad (approved):
Crawler looks happy

PR #5016, opened 2023-12-05
Fix inline handler in study designer

labkey-adam:
#### Rationale
Clean up CSP logs. Code is deprecated but we still test it.


labkey-adam:
`VaccineProtocolTest` passed on this FB: https://teamcity.labkey.org/buildConfiguration/bt7/2780078?buildTab=tests

labkey-matthewb (approved):

PR #5017, opened 2023-12-05
Issue 49216: Set JDBC preparedThreshold=0 for PostgreSQL

labkey-nicka:
#### Rationale
This seeks to address [Issue 49216](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49216) by setting JDBC parameter `preparedThreshold=0` for all PostgreSQL primary database connections. Setting the parameter in this way [disables prepared statement caching](https://jdbc.postgresql.org/documentation/server-prepare/#deactivation) which is known to cause problems when schema changes occur.

#### Changes
- Expose setting of connection properties on `DbScope` and consolidate setting these properties.
- Introduce `prepare(LabKeyDataSource dataSource)` on `SqlDialect` to support dialect-specific preparation of the data source prior to connections being established.
- Set `preparedThreshold=0` for the primary data source on PostgreSQL to address [Issue 49216](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49216).
- Retrieve driver `Class` from data source rather than from connection.

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -304,12 +304,12 @@ public class DbScope
         private final String _dsName; // DataSource name from labkey.xml
         private final DataSource _ds;
         private final DataSourcePropertyReader _dsPropertyReader;
+        private final @Nullable Properties _connectionProperties;
         private final String _driverClassName;
         private final SqlDialect _dialect;
         private final Class<Driver> _driverClass;
         private final String _url;
 
-        private @Nullable String _applicationName; // Null if a custom application name is set (JDBC URL) or initial connection hasn't been made yet
         private boolean _logQueries = false;
         private String _displayName = null;
         private boolean _primary = false;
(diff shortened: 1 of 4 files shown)


labkey-adam (changes requested):
I have some local changes. Will commit some time today and add my approval.

labkey-adam (approved):

labkey-adam:
@labkey-nicka I've approved, but you should review my changes. I hope they (along with my comment in the issue) are self-explanatory. Note that since `disablePreparedStatementCaching()` was super specific to PostgreSQL, I opted to remove the SqlDialect method, loosen some visibility on LabKeyDataSource, and let PG prepare() set the property directly. I also got rid of a sneaky, but unnecessary, connection getting made before the no-other-connections check; not sure why this wasn't causing problems with the check, but it's gone now so I'm not going to worry about it.

If you and TeamCity are satisfied, feel free to merge. Thanks for your help!

labkey-nicka (commented):
Looks good. This is much cleaner.

PR #5018, opened 2023-12-05
It's a bit too soon to be using NULLS NOT DISTINCT

labkey-adam:
#### Rationale
`NULLS NOT DISTINCT` syntax was just added in PostgreSQL 15, so drop and recreate the unique index with the `COALESCE()` hack.


labkey-nicka (approved):
Looks good. Upgrade ran successfully locally on PostgreSQL 13.

PR #5022, opened 2023-12-06
Issue 49227: Default lineage maximum depth

### LINKED ISSUE #5021: Issue 49227: Add capability to configure maximum default lineage depth, reduce default
[Issue 49227](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49227)

There are a couple of different ways we utilize lineage queries in LKS:

1. Via the experiment-lineage.api (this backs things like the lineage views in our apps).
2. Via LineageForeignKey which backs Input/* and Output/* columns in queries for experiment related entities.

The (1) way supports a depth parameter which callers can specify. The (2) way, however, does not support a depth parameter and falls back to a default maximum depth. This default maximum depth is 100 for SQL Server and 1,000 for PostgreSQL (see [Issue 37332](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=37332) for the reason these differ). On systems with a lot of data, and specifically a lot of data across a large number of generations, (2) can become a performance bottleneck due to it processing up to 1,000 generations. In general, this amount of information is not useful for day-day usage.

The request here is to either (and/or):

A. Add a configuration option that allows for users to set the default maximum depth.
B. Reduce the default maximum depth on PostgreSQL to a lower value (100?).

Personally, I'm leaning towards both A and B since it will reduce the default but users can opt-in to a higher default. Otherwise, I'd lean towards B before A. For A, I'm thinking a ModuleProperty on the ExperimentModule could work but open to other ideas.

labkey-nicka:
#### Rationale
This seeks to address [Issue 49227](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49227) by a) lowering the default maximum lineage depth to 100 (from 1,000 on PostgreSQL) and b) introducing a new site-wide Experiment module ModuleProperty that allows for users to override this setting.

image

#### Changes
- Set default maximum depth to 100.
- Introduce Experiment module ModuleProperty `LineageMaximumDepthModuleProperty` to support user override of the default depth.
- Add validation of module properties prior to save save.
- Maximum allowed depth is 100 for SQL Server and 1,000 for PostgreSQL.

### DIFF
--- a/api/src/org/labkey/api/exp/api/ExpLineageOptions.java
+++ b/api/src/org/labkey/api/exp/api/ExpLineageOptions.java
@@ -17,13 +17,17 @@ package org.labkey.api.exp.api;
 
 import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.Nullable;
+import org.labkey.api.module.ModuleLoader;
 
 /**
  * Captures options for doing a lineage search
- * Created by Nick Arnold on 2/12/2016.
  */
 public class ExpLineageOptions extends ResolveLsidsForm
 {
+    // Issue 37332: SQL Server can hit max recursion depth over 100 generations.
+    // Note: If this is adjusted higher, then consider separate default values for SQL Server and PostgreSQL.
+    public static final int LINEAGE_DEFAULT_MAXIMUM_DEPTH = 100;
+
     public enum LineageExpType
     {
         ALL,
@@ -177,5 +181,22 @@ public class ExpLineageOptions extends ResolveLsidsForm
     public void setSourceKey(String sourceKey)
     {
         _sourceKey = sourceKey;
+    }
+
+    public int getConfiguredDepth()
+    {
+        if (_depth != 0)
+            return _depth;
+
+        var module = ModuleLoader.getInstance().getModule(ExperimentService.MODULE_NAME);
+        var property = module.getModuleProperties().get(ExperimentService.LINEAGE_DEFAULT_MAXIMUM_DEPTH_PROPERTY_NAME);
+        if (property != null)
+        {
+            String sDepth = property.getEffectiveValue(null);
+            if (!StringUtils.isEmpty(sDepth))
+                return Integer.parseInt(sDepth);
+        }
+
+        return LINEAGE_DEFAULT_MAXIMUM_DEPTH;
     }
 }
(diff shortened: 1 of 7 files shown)


experiment/src/org/labkey/experiment/LineageMaximumDepthModuleProperty.java
```diff
@@ -0,0 +1,45 @@
...
+    {
+        if (StringUtils.isEmpty(value))
+            return;
+
+        try
+        {
+            var depth = Integer.parseInt(value);
+            if (depth < MINIMUM_DEPTH)
+                throw new IllegalArgumentException(String.format("Invalid value for \"%s\". Minimum value is %d.", getLabel(), MINIMUM_DEPTH));
+            else if (depth > MAXIMUM_DEPTH)
```
labkey-matthewb: should maximum depth be dialect specific?

api/src/org/labkey/api/exp/api/ExpLineageOptions.java
```diff
@@ -178,4 +182,21 @@ public void setSourceKey(String sourceKey)
     {
         _sourceKey = sourceKey;
     }
+
+    public int getConfiguredDepth()
+    {
+        if (_depth != 0)
```
labkey-matthewb, line 188: I see there is validation for this, but... ">" instead of "!="?

labkey-matthewb (approved):

experiment/src/org/labkey/experiment/LineageMaximumDepthModuleProperty.java
labkey-matthewb: Or we could use this moment to inject the dialect specific syntax into the recursive CTE.

api/src/org/labkey/api/exp/api/ExpLineageOptions.java
labkey-nicka: I was carrying this over from how the queries were previously set up (they had `depth == 0`). I can switch this to `>` and we'll find out if anyone has been supplying negative depth values (which the query is partially defensive against  `AND _Graph.depth >= <%= (-1 * Math.abs(depth)) + 1 %>` but not fully).

labkey-nicka: Could also consider:

```java
if (_depth != 0)
    return Math.abs(_depth);
```

Edit: went with this approach. See cb0661f.

experiment/src/org/labkey/experiment/LineageMaximumDepthModuleProperty.java
labkey-nicka: Elected to stick with a dialect specific depth for now. See ea01a5d.

PR #5023, opened 2023-12-06
Issue 48243: Period in Sample ID full of numbers expects an integer during Assay Import and is handled poorly during update

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48243

For the assay import of lookup columns, we want to allow for the original value to be used when attempting to resolve the lookup (for example, the SampleId column might be type int but the original value could be a string or double and we want that original value to have a chance to resolve instead of giving a type conversion error.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/1750

#### Changes
- AbstractAssayTsvDataHandler to setBsetEfforConversion(true) for DataLoaderSettings
- Fix tests to check for update to assay import data type conversion error message

### DIFF
--- a/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
+++ b/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
@@ -146,7 +146,7 @@ public abstract class AbstractAssayTsvDataHandler extends AbstractExperimentData
         // Issue 48243: For the assay import of lookup columns, we want to allow for the original value to be used when
         // attempting to resolve the lookup (for example, the SampleId column might be type int but the original value
         // could be a string or double and we want that original value to have a chance to resolve instead of giving a
-        // type conversion error.
+        // type conversion error).
         settings.setBestEffortConversion(true);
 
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
```diff
@@ -143,6 +143,12 @@ public void importFile(@NotNull ExpData data, File dataFile, @NotNull ViewBackgr
         DataLoaderSettings settings = new DataLoaderSettings();
         settings.setAllowLookupByAlternateKey(allowLookupByAlternateKey);
 
+        // Issue 48243: For the assay import of lookup columns, we want to allow for the original value to be used when
+        // attempting to resolve the lookup (for example, the SampleId column might be type int but the original value
+        // could be a string or double and we want that original value to have a chance to resolve instead of giving a
+        // type conversion error.
```
labkey-susanh:
```suggestion
        // type conversion error).
```

labkey-susanh (approved):

PR #5024, opened 2023-12-07
Resolve MenuBarTest Outage

RosalineP:
#### Rationale
Previously, a `var` was changed to a `const`. This `var` is updated in code below its declaration and also seems to be referenced in other .js files, meriting a `var` and not a `let`. This update should also fix MenuBarTest.testSteps.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5008

#### Changes
* Change const to var
* Add newline

### DIFF
--- a/core/src/org/labkey/core/view/template/bootstrap/navigation.jsp
+++ b/core/src/org/labkey/core/view/template/bootstrap/navigation.jsp
@@ -261,6 +261,7 @@
 </nav>
 <% if (!model.getCustomMenus().isEmpty()) { %>
 <script type="text/javascript" nonce="<%=getScriptNonce()%>">
+    <%--  note __menus must be a var, not a let or const  --%>
     var __menus = {};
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/view/template/bootstrap/navigation.jsp
```diff
@@ -261,7 +261,7 @@
 </nav>
 <% if (!model.getCustomMenus().isEmpty()) { %>
 <script type="text/javascript" nonce="<%=getScriptNonce()%>">
-    const __menus = {};
+    var __menus = {};
```
Sigmonia, line 265: If this needs to be `var` vs `let` I would add comment so someone doesn't come around with the linter later and change it.

Sigmonia (approved):

labkey-adam (approved):
Thanks for fixing my mistake

PR #5030, opened 2023-12-09
PostgreSQL connections: set preparedStatementCacheQueries=0

labkey-nicka:
#### Rationale
There are [test failures](https://teamcity.labkey.org/buildConfiguration/LabKey_2311Release_Premium_ProductSuites_Biologics_BiologicsAPostgres/2786728?hideProblemsFromDependencies=false&expandBuildTestsSection=true&hideTestsFromDependencies=false&expandBuildChangesSection=true) in CI due to the server logging an exception:

image

Originally, I'd thought that setting `preparedThreshold=0` should suffice in disabling the caching behavior, however, that doesn't appear to be correct. From the [PostgreSQL JDBC docs](https://jdbc.postgresql.org/documentation/use/#connection-parameters):

<p align="center">
image
</p>

I've elected to additionally add `preparedStatementCacheQueries=0`, however, it could be sufficient for [Issue 49216](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49216) that we only need to set this parameter and can remove `preparedThreshold=0`.

#### Related Pull Requests
* #5017

#### Changes
* Set `preparedStatementCacheQueries=0` for PostgreSQL connections

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -1768,7 +1768,7 @@ public class DbScope
                     // then set our default application name on the data source so it's set on every connection.
                     if (applicationName.equals(dialect.getDefaultApplicationName()))
                     {
-                        // Set LabKey's default application name ("LabKey Server") into the connectinon properties
+                        // Set LabKey's default application name ("LabKey Server") into the connection properties
                         applicationName = ds.setDefaultApplicationName();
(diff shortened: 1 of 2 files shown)


labkey-adam (approved):
Interesting... `preparedThreshold=0` worked for the upgrade scenario and is documented as shutting off the cache. But I guess we've been using `preparedStatementCacheQueries` for years.

PR #5033, opened 2023-12-11
Issue 49036: Include containerFilter in sample table info for NAb PercentNeutralizationInitialDilution calculation

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49036

The related PR fixed the issue for the NAb PercentNeutralizationInitialDilution calculation by joining to the NAb samples table to get the initial dilution for the calc where clause. However, the container filter was not passed through to the TableInfo for the samples table, so this resulted in blank value in the NAb Link to Study dataset when the NAb run was in a subfolder of the assay design container.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4936

#### Changes
* Include containerFilter in sample table info for NAb PercentNeutralizationInitialDilution calculation


cnathe:
NOTE: I am checking with the client if we need to backport this fix all the way back to 23.3 or if we can put it in 23.11.

cnathe:
closing this PR and targeting 23.11 with this fix

PR #5034, opened 2023-12-11
Preprocess custom participant view with ModuleHtmlView

labkey-matthewb:
#### Rationale
Always treat custom participant view as ModuleHtmlView.  Rationale we script nonce handling, and it seems correct for the participantview to execute the same whether defined in a module or via the UI.  (There are features other than the nonce substitution)

#### Related Pull Requests
* 

#### Changes
*


labkey-adam (commented):
Nice!

labkey-adam (approved):
Could have sworn I approved this early this morning...

PR #5037, opened 2023-12-11
handle mysql quoted identifiers

labkey-matthewb:
#### Rationale
SQLFragment.appendIdentifier() did not recognized MySql-style quotes

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/data/SQLFragment.java
+++ b/api/src/org/labkey/api/data/SQLFragment.java
@@ -414,7 +414,7 @@ public class SQLFragment implements Appendable, CharSequence
         else
             malformed = StringUtils.containsAny(identifier, "*/\\'\"`?;- \t\n");
         if (malformed && !AppProps.getInstance().isExperimentalFeatureEnabled(FEATUREFLAG_DISABLE_STRICT_CHECKS))
-            throw new IllegalArgumentException("SQLFragment.appendIdentifier(String) value appears to incorrectly formatted");
+            throw new IllegalArgumentException("SQLFragment.appendIdentifier(String) value appears to be incorrectly formatted");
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/SQLFragment.java
```diff
@@ -406,12 +406,15 @@ public SQLFragment appendIdentifier(CharSequence charseq)
...
-        }
+        boolean malformed;
+        if (identifier.length() >= 2 && identifier.startsWith("\"") && identifier.endsWith("\""))
+            malformed = (StringUtils.countMatches(identifier, '\"') % 2) != 0;
+        else if (identifier.length() >= 2 && identifier.startsWith("`") && identifier.endsWith("`"))
+            malformed = (StringUtils.countMatches(identifier, '`') % 2) != 0;
+        else
+            malformed = StringUtils.containsAny(identifier, "*/\\'\"`?;- \t\n");
+        if (malformed && !AppProps.getInstance().isExperimentalFeatureEnabled(FEATUREFLAG_DISABLE_STRICT_CHECKS))
+            throw new IllegalArgumentException("SQLFragment.appendIdentifier(String) value appears to be incorrectly formatted");
```
labkey-jeckels: Can we include the value in this error message to make future troubleshooting easier, or is that somehow problematic?

labkey-jeckels (approved):

api/src/org/labkey/api/data/SQLFragment.java
labkey-matthewb: Not problematic.  But it's not really input related, it's just a bad code path.

PR #5038, opened 2023-12-12
Issues 49200 (standardized sample unit) and 49246 (remove redundant index)

labkey-susanh:
#### Rationale
Issue [49200:](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49200) - When importing with units that do not match the standard, we were matching properly, but leaving the non-standard casing in the data. This PR updates the conversion to always return a Units value.
Issue [49246](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49246) - inadvertently created a redundant index for the User Audit Domain when adding the `Created` index in the base class

#### Related Pull Requests
* #4830 

#### Changes
* In `UserAuditProvider`, don't call the base class method that creates the redundant index
* Update `SampleUnitsConvertColumn` to have the `convert` method return the standardized measurement unit.

### DIFF
--- a/core/src/org/labkey/core/query/UserAuditProvider.java
+++ b/core/src/org/labkey/core/query/UserAuditProvider.java
@@ -143,13 +143,12 @@ public class UserAuditProvider extends AbstractAuditTypeProvider implements Audi
         @Override
         public Set<PropertyStorageSpec.Index> getPropertyIndices(Domain domain)
         {
-            Set<PropertyStorageSpec.Index> indexes = super.getPropertyIndices(domain)
-                    .stream()
-                    .filter(index -> index.columnNames.length != 1 || !index.columnNames[0].equalsIgnoreCase(COLUMN_NAME_CREATED))
-                    .collect(Collectors.toSet());
-           indexes.add(new PropertyStorageSpec.Index(false, COLUMN_NAME_USER));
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/query/UserAuditProvider.java
```diff
@@ -142,9 +142,11 @@ public Set<PropertyDescriptor> getProperties()
         @Override
         public Set<PropertyStorageSpec.Index> getPropertyIndices(Domain domain)
         {
-            Set<PropertyStorageSpec.Index> indexes = super.getPropertyIndices(domain);
-            indexes.add(new PropertyStorageSpec.Index(false, COLUMN_NAME_USER));
-            indexes.add(new PropertyStorageSpec.Index(false, COLUMN_NAME_CREATED, COLUMN_NAME_USER));
+            Set<PropertyStorageSpec.Index> indexes = PageFlowUtil.set(
```
labkey-jeckels: I'd be inclined to keep the call to `super` and instead remove the redundant index (which would require the simple change to make `PropertyStorageSpec.Index` implement `equals()` and `hashCode()`). But I guess it depends on whether it's better to default to opting-in or opting-out of the common indices.

labkey-jeckels (approved):

core/src/org/labkey/core/query/UserAuditProvider.java
labkey-susanh: OK. Updated with an equals and hash and a remove call.

PR #5039, opened 2023-12-12
Issue 49131: Always calculate the values for the first N-1 samples, based on their sequence

ankurjuneja:
#### Rationale
This PR changes the trailing means and trailing cvs calculation helper to always calculate the values for the first N-1 samples, based on their sequence. For example, the first sample's data point for trailing mean should be the raw value of the first point, and the CV should be 0. The second sample's trailing mean should be the mean of the first and second point, and their CVs. And so on until you hit N, when we start using the moving window.

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/798

### DIFF
--- a/api/src/org/labkey/api/visualization/Stats.java
+++ b/api/src/org/labkey/api/visualization/Stats.java
@@ -166,7 +166,7 @@ public class Stats
      */
     public static Double[] getTrailingMeans(Double[] values, int N)
     {
-        if (values == null || values.length <= 1 || values.length <= N)
+        if (values == null || values.length <= 1)
             return new Double[0];
 
         Double[] trailingMeans = new Double[values.length];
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/visualization/Stats.java
```diff
@@ -195,34 +205,54 @@ public static Double[] getTrailingCVs(Double[] values, int N)
         if (values == null || values.length <= 1 || values.length <= N)
```
labkey-jeckels: The last part of this check is no longer valid - we should now calculate the CV for all the points, even if there are less than N

api/src/org/labkey/api/visualization/Stats.java
```diff
@@ -195,34 +205,54 @@ public static Double[] getTrailingCVs(Double[] values, int N)
...
+        trailingCVs[0] = 0.0;
+        for (int i = 1; i < values.length; i++)
         {
-            Double[] vals = getValuesFromRange(values, start, end);
-            double sd = getStdDev(vals, false);
-            double mean = getMean(vals);
-            if (mean == 0)
+            // Issue 49131: At the very beginning of the folder's date range, always calculate the values for the first N-1 samples,
+            // based on their sequence. And so on until you hit N, when we start using the moving window.
+            if (i < N)
```
labkey-jeckels: The if/else branches here (and the trailing mean calc) can be almost completely combined. They only differ in whether they increment `start` or not.

api/src/org/labkey/api/visualization/Stats.java
```diff
@@ -169,15 +169,25 @@ public static Double[] getTrailingMeans(Double[] values, int N)
         if (values == null || values.length <= 1 || values.length <= N)
```
labkey-jeckels: The last part of this check is no longer valid - we should now calculate the mean for all the points, even if there are less than N

labkey-jeckels (approved):
One important change, another optional suggestion. Let me know if you want another review.

PR #5048, opened 2023-12-14
deadlock retry

labkey-matthewb:
#### Rationale
attempt to mitigate some reported deadlocks.  
a) use executeWithRetry 
b) actually batch in indexDataClassData()

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -1009,13 +1009,11 @@ public class ExperimentServiceImpl implements ExperimentService, ObjectReference
             .append(" AND (d.lastIndexed IS NULL OR d.lastIndexed < ? OR (d.modified IS NOT NULL AND d.lastIndexed < d.modified))")
             .add(dataClass.getModified());
 
-        new SqlSelector(table.getSchema().getScope(), sql).forEachBatch(Data.class, 1000, batch -> {
-            for (Data data : batch)
-            {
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -1009,13 +1009,11 @@ private void indexDataClassData(ExpDataClassImpl dataClass, SearchService.IndexT
...
             .add(dataClass.getModified());
 
-        new SqlSelector(table.getSchema().getScope(), sql).forEachBatch(Data.class, 1000, batch -> {
-            for (Data data : batch)
-            {
-                ExpDataImpl impl = new ExpDataImpl(data);
-                impl.index(task, null);
-            }
-        });
+        new SqlSelector(table.getSchema().getScope(), sql).forEachBatch(Data.class, 1000, batch ->
```
labkey-jeckels, line 1012: The Postgres driver will cache the full ResultSet. I haven't checked if/when this is called inside of a transaction (where our cache-disabling doesn't work), but probably worth it to at least an attempt to stop caching on the `SqlSelector`.

labkey-jeckels (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-matthewb: Yes, I was just trying at least not create all the Documents (and property queries) synchronously.

PR #5049, opened 2023-12-14
Convert move entities between projects APIs to use new query-moveRows.api

cnathe:
#### Rationale
After we enabled moving of entities between projects, we knew that moving via the API would eventually be helpful. This is particularly useful since we’ve limited the number of entities that can be moved via the UI for performance purposes. We also realized after the implementation that instead of having individual controller actions for each entity type, we could consolidate them into a single MoveRowsAction in the QueryController and have the implementations for each handled within the specific QueryUpdateService for the entities table.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5049
- https://github.com/LabKey/labkey-ui-components/pull/1369
- https://github.com/LabKey/labkey-ui-premium/pull/279
- https://github.com/LabKey/biologics/pull/2582
- https://github.com/LabKey/labbook/pull/584
- https://github.com/LabKey/inventory/pull/1129
- https://github.com/LabKey/sampleManagement/pull/2302

#### Changes
- Initial QueryController.MoveRowsAction and QueryUpdateService moveRows() 
  - require targetContainerPath param in addition to standard save rows action params
  - extend ApiSaveRowsForm to support getting rowIds from the selectionKey
  - allow for target container param as targetContainer or targetContainerPath and allow for auditUserComment or userComment
  - add new MoveEntitiesPermission to be used for permission checks in table QueryUpdateServices that implement moveRows()
- Remove MoveSamplesAction and move implementation to SampleTypeUpdateServiceDI moveRows()
- Remove MoveDataClassObjectsAction and move implementation to ExpDataClassDataTableImpl.DataClassDataUpdateService moveRows()
- Remove MoveAssayRunsAction and move implementation to ExpRunTableImpl.RunTableUpdateService moveRows()
- Remove AbstractMoveEntitiesAction and related form class


api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -2585,7 +2585,7 @@ public static Container getMoveTargetContainer(@NotNull String entityType, @NotN
 
         if (!_targetContainer.hasPermission(user, InsertPermission.class))
         {
-            errors.reject(ERROR_GENERIC, "You do not have permission to move " + entityType + " to the target container: " + targetIdOrPath + ".");
+            errors.reject(ERROR_GENERIC, "You do not have permission to move rows to this target container: " + targetIdOrPath + ".");
```
labkey-susanh:
```suggestion
            errors.reject(ERROR_GENERIC, "You do not have permission to move rows to the target container: " + targetIdOrPath + ".");

```
Do we have the query name that we could pass in here for better user feedback ("permission to move rows from 'My Sample Type' to the target container: X")?

experiment/src/client/test/integration/MoveAssayRunsAction.ispec.ts
```diff
@@ -182,24 +185,28 @@ describe(CONTROLLER_NAME, () => {
...
+                schemaName: 'assay.General.assayA',
+                queryName: 'Runs',
+                rows: []
+            }, {...topFolderOptions, ...editorUserOptions}).expect(500);
 
             // Assert
             const {exception, success} = response.body;
             expect(success).toBe(false);
-            expect(exception).toEqual('Run IDs must be specified for the move operation.');
+            expect(exception).toEqual('No \'rows\' array supplied!');
```
labkey-susanh: Does this message really need the `!`? It's consistent, I'm sure, but not all that astonishing (to me).

experiment/src/client/test/integration/MoveAssayRunsAction.ispec.ts
```diff
@@ -419,17 +438,19 @@ describe(CONTROLLER_NAME, () => {
...
+            const response = await server.post('query', 'moveRows.api', {
                 targetContainer: subfolder1Options.containerPath,
-                rowIds: [runId],
+                schemaName: 'assay.General.assayA',
+                queryName: 'Runs',
+                rows: [{ RowId: runId }],
                 auditBehavior: "DETAILED",
                 userComment
-            }, {...topFolderOptions, ...editorUserOptions}).expect(200);
+            }, {...topFolderOptions, ...editorUserOptions}).expect(500);
```
labkey-susanh: I wouldn't expect a 500 error here. That says something went wrong on the server, not that the request was bad.

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
```diff
@@ -315,36 +337,42 @@ describe('ExperimentController', () => {
...
+        it('error, missing required rows param', async () => {
             // Act
-            const response = await server.post('experiment', 'moveSamples.api', {
-                targetContainer: subfolder1Options.containerPath
-            }, {...topFolderOptions, ...editorUserOptions}).expect(400);
+            const response = await server.post('query', 'moveRows.api', {
+                targetContainer: subfolder1Options.containerPath,
+                schemaName: 'samples',
+                queryName: SAMPLE_TYPE_NAME_1,
+            }, {...topFolderOptions, ...editorUserOptions}).expect(500);
```
labkey-susanh: Also won't expect a 500 response here. This isn't a server error.

experiment/src/client/test/integration/MoveSourcesAction.ispec.ts
```diff
@@ -308,36 +317,42 @@ describe('ExperimentController', () => {
...
+                targetContainer: subfolder1Options.containerPath,
+                schemaName: 'exp.data',
+                queryName: SOURCE_TYPE_NAME_1,
+            }, { ...topFolderOptions, ...editorUserOptions }).expect(500);
 
             // Assert
             const { exception, success } = response.body;
             expect(success).toBe(false);
-            expect(exception).toEqual('Source IDs must be specified for the move operation.');
+            expect(exception).toEqual('No \'rows\' array supplied!');
```
labkey-susanh: Using double quotes here would be more readable than escaping the single quotes.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -9268,18 +9269,15 @@ public Map<String, Integer> moveAssayRuns(@NotNull List<? extends ExpRun> assayR
...
             transaction.addCommitTask(() -> {
-                int fileMoveCount = 0;
                 for (List<AbstractAssayProvider.AssayFileMoveData> runFileRenameData : assayMoveData.fileMovesByRunId().values())
                 {
                     for (AbstractAssayProvider.AssayFileMoveData renameData : runFileRenameData)
-                    {
-                        if (moveFile(renameData))
-                            fileMoveCount++;
-                    }
+                        moveFile(renameData);
```
labkey-susanh, line 9279: If the move fails we'll still be counting that as a moved file. Not a big deal since we don't rely on these counts too much, but not entirely accurate.

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -4255,6 +4257,21 @@ public List<Map<String, Object>> saveRows(QueryUpdateService qus, List<Map<Strin
                 return Collections.emptyList();
             }
         },
+        moveRows(MoveEntitiesPermission.class, QueryService.AuditAction.UPDATE)
```
labkey-susanh, line 4260: Is this separate permissions class needed? Do we need to distinguish it from `UpdatePermission`?

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -4402,6 +4429,15 @@ protected Container getContainerForCommand(JSONObject json)
             return container;
         }
 
+        protected String getTargetContainerProp()
+        {
+            JSONObject json = getJsonObject();
+            if (json.has(PROP_TARGET_CONTAINER) && !json.has(PROP_TARGET_CONTAINER_PATH))
```
labkey-susanh: Do we need to support both of these?

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -4720,8 +4771,118 @@ public ApiResponse execute(ApiSaveRowsForm apiSaveRowsForm, BindException errors
...
+        private void updateSelections(MoveRowsForm form)
+        {
+            String selectionKey = form.getDataRegionSelectionKey();
+            if (selectionKey != null)
+            {
+                Set<String> rowIds = form.getIds(getViewContext(), false)
+                        .stream().map(Object::toString).collect(Collectors.toSet());
+                DataRegionSelection.setSelected(getViewContext(), selectionKey, rowIds, false);
+
+                // if moving run items from a type, the selections from other selectionKeys in that container will
```
labkey-susanh:
```suggestion
                // if moving entities from a type, the selections from other selectionKeys in that container will

```

labkey-susanh (changes requested):

experiment/src/client/test/integration/MoveAssayRunsAction.ispec.ts
cnathe: done

api/src/org/labkey/api/data/ContainerManager.java
cnathe: Yes, added queryName to error message

experiment/src/client/test/integration/MoveAssayRunsAction.ispec.ts
cnathe: catch IllegalArgumentException and rethrow as ValidationException

experiment/src/client/test/integration/MoveSamplesAction.ispec.ts
cnathe: catch IllegalArgumentException and rethrow as ValidationException

experiment/src/client/test/integration/MoveSourcesAction.ispec.ts
cnathe: done

query/src/org/labkey/query/controllers/QueryController.java
cnathe: no, changed to only support "targetContainerPath" and "auditUserComment"

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
cnathe: Yeah, agreed. Since the file moves happen in an post commit task, this count wasn't getting included in the response object. I couldn't come up with a better way to make this work but open to suggestions.

query/src/org/labkey/query/controllers/QueryController.java
cnathe: I didn't think it was going to be...the move samples, sources, and assay runs all worked fine using UpdatePermission as the moveRows action check. However, when I got to the move notebooks case, I couldn't find another way to get past the NotebookTable NotebookUpdateService.hasPermission check. I couldn't find a way to get the TableInfo QUS for this NotebookTable in the general moveRows case in a way that would `_allowInsertUpdate`. @labkey-nicka @labkey-susanh if either of you have a suggestion about how to do with without adding a new permission class, that would be awesome.

labkey-susanh: I see how that's a problem now, yes and am also not seeing a better way out of this.

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -4402,25 +4428,31 @@ protected Container getContainerForCommand(JSONObject json)
...
-                throw new IllegalArgumentException("Empty request");
+                throw new ValidationException("Empty request");
 
             JSONArray rows;
             try
             {
                 rows = json.getJSONArray(PROP_ROWS);
                 if (rows.length() < 1)
-                    throw new IllegalArgumentException("No '" + PROP_ROWS + "' array supplied!");
+                    throw new ValidationException("No '" + PROP_ROWS + "' array supplied.");
```
labkey-susanh, line 4451: Thank you for this!

labkey-susanh (approved):

query/src/org/labkey/query/controllers/QueryController.java
labkey-nicka: Indeed this is intentional. We don't want API users to be able to insert/update against the notebook table even if they have insert/update permissions. The intention is that they use the notebook APIs which will perform other necessary checks. Adding a new class seems like a reasonable approach to me but I'd also be curious what @labkey-klum thinks.

labkey-klum: Agreed, introducing a new permission class is the cleanest way to deal with the notebook table permissions.

PR #5056, opened 2023-12-16
log user agent with csp report.

labkey-matthewb:
#### Rationale
Just a little more info

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels:
Do we want this in 23.12?

PR #5057, opened 2023-12-16
Cross folder samples import and merge

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1371
- https://github.com/LabKey/labkey-ui-premium/pull/282
- https://github.com/LabKey/platform/pull/5057
- https://github.com/LabKey/sampleManagement/pull/2305
- https://github.com/LabKey/biologics/pull/2586

#### Changes
*


experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -1168,18 +1155,20 @@ else if (name != null && materialSourceId != null)
...
-            Map<String, Object>[] cfRows = new TableSelector(tInfo, existingDataFilter, null).getMapArray();
+            existingDataFilter.addCondition(FieldKey.fromParts(useLsid ? "LSID" : "Name"), allKeys, CompareType.IN);
+            Map<String, Object>[] cfRows = new TableSelector(ExperimentService.get().getTinfoMaterial(), existingDataFilter, null).getMapArray();
             for (Map<String, Object> row : cfRows)
             {
-                String dataContainer = (String) row.get("folder");
+                String dataContainer = (String) row.get("container");
                 if (!dataContainer.equals(container.getId()))
-                    throw new InvalidKeyException("Sample does not belong to the current container: " + (String) row.get("name") + ".");
+                    throw new InvalidKeyException("Sample does not belong to " + container.getName() + " container: " + (String) row.get("name") + ".");
```
labkey-susanh, line 1171: I would probably word this like a sentence ("Sample <Name> does not belong to <ContainerName>.") but it's not worth changing if it breaks lots of expectations somewhere else or makes this messaging inconsistent with other messaging in some significant way.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2291,54 +2298,231 @@ else if (isMergeOrUpdate)
...
+         */
+        private boolean handleCrossFolderUpdate(TypeData typeData)
+        {
+            String headerRow = typeData.dataRows.get(0);
+            String dataFileName = typeData.dataFile.getName();
+
+            ExpSampleTypeImpl sampleType = typeData.sampleType;
+
+            SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("MaterialSourceId"), sampleType.getRowId());
+            filter.addCondition(FieldKey.fromParts("Name"), typeData.sampleIds, CompareType.IN);
```
labkey-susanh, line 2447: This looks like we are specifically not supporting cross-folder/cross-type updates using rowIds instead of names, or does that handling happen elsewhere? Probably supporting only name is OK, but we'll at least want to communicate that that's the case.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2291,54 +2298,231 @@ else if (isMergeOrUpdate)
...
+                {
+                    containerRows.put(dataContainer, new ArrayList<>());
+                    hasCrossFolderData = hasCrossFolderData || !dataContainer.equalsIgnoreCase(_container.getId());
+                }
+
+                containerRows.get(dataContainer).add(sampleRowId);
+            }
+
+            if (!hasCrossFolderData)
+                return false;
```
labkey-susanh, line 2469: This is probably because my brain wants to be on holiday too soon, but I'm confused by the kind of dual purpose of the return value for this function.  The false value is clear, but the true value is returned even when there's an error, and that just makes my brain hurt a bit. Seems like the functionality above this line could be factored out into a `getContainerRowsData` method with the return value from that method then passed to the method that does the rest of this.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2291,54 +2298,231 @@ else if (isMergeOrUpdate)
...
+                containerIndexes.put(containerId, dataRowIndexes.get(0));
+                for (Integer dataRowIndex : dataRowIndexes)
+                    dataRows.add(typeData.dataRows.get(dataRowIndex));
+
+                TypeData containerTypeData = new TypeData(container, sampleType, samplesTable, dataFile, typeData.fieldIndexes, typeData.dependencyIndexes, dataRows, new ArrayList<>());
+                containerTypeDatas.put(containerId, containerTypeData);
+            }
+
+            List<String> orderedContainer = new ArrayList<>(containerRows.keySet());
+            orderedContainer.sort(Comparator.comparing(containerIndexes::get));
```
labkey-susanh: Can you add a comment here about why this sort is needed?

labkey-susanh (approved):
I'll approve since I'm going to be away, but would appreciate that little refactor if it makes sense to you.

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: We don't support "Update from file" or "merge" using lsid/rowid, the only scenarios supported so far is "Name". For "Update" from ui, we do support "lsid" instead of name. But we are using a different approach for updating from UI: saveRows.

PR #5058, opened 2023-12-16
Issue 48715: Limit Data Region "Show all" to a maximum number of rows

labkey-nicka:
#### Rationale
This addresses [Issue 48715](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48715) by switching "Show all" functionality in Data Regions to set a `maxRows=1000` parameter instead of setting `showRows=all`. The intention is to a) avoid unwieldy large requests generated by our UI that can consume significant server resources and b) avoid crashing the user's browser attempting to render too many results. This is a client-side change only (so `showRows=all` is still supported but just not generated by our UI).

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/1760

#### Changes
- Switch selecting "Show all" or calling `DataRegion.showAllRows()` to setting the URL parameter `maxRows=1000` instead of `showRows=all`
- Display a message to the user iff the "Show all" max rows boundary is selected and has been reached.

### DIFF
--- a/api/webapp/clientapi/dom/DataRegion.js
+++ b/api/webapp/clientapi/dom/DataRegion.js
@@ -3539,7 +3539,7 @@ if (!LABKEY.DataRegions) {
 
             msg += "&nbsp;" + "<span class='labkey-button select-none'>Select None</span>";
             var showOpts = [];
-            if (region.showRows !== 'all')
+            if (region.showRows !== 'all' && region.maxRows !== ALL_ROWS_MAX)
                 showOpts.push("<span class='labkey-button show-all'>Show All</span>");
             if (region.showRows !== 'selected')
                 showOpts.push("<span class='labkey-button show-selected'>Show Selected</span>");
--- a/api/webapp/clientapi/dom/DataRegion.js
+++ b/api/webapp/clientapi/dom/DataRegion.js
@@ -12,8 +12,10 @@ if (!LABKEY.DataRegions) {
     //
     // CONSTANTS
     //
+    // Issue 48715: Limit the number of rows that can be displayed in a data region
+    var ALL_ROWS_MAX = 1_000;
     var CUSTOM_VIEW_PANELID = '~~customizeView~~';
-    var DEFAULT_TIMEOUT = 30000;
+    var DEFAULT_TIMEOUT = 30_000;
     var PARAM_PREFIX = '.param.';
     var SORT_ASC = '+';
     var SORT_DESC = '-';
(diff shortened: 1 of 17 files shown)


api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -1659,7 +1663,11 @@ if (!LABKEY.DataRegions) {
                     //bind functions to menu items
                     _getShowFirstSelector(this).click(_firstPage.bind(this));
                     _getShowLastSelector(this).click(_lastPage.bind(this));
-                    _getShowAllSelector(this).click(_showRows.bind(this, this, 'all'));
+                    _getShowAllSelector(this).click(this.showAllRows.bind(this));
+
+                    if (this.maxRows === ALL_ROWS_MAX && this.totalRows > this.maxRows) {
+                        this.addMessage('<span><b>Show all:</b> Displaying the first ' + ALL_ROWS_MAX.toLocaleString() + ' rows. Use paging to see more results.</span>');
```
labkey-nicka, line 1669: Certainly open to other messaging here. The main thing is I think we should display a message, otherwise, it just seems silly (borderline lying) when someone clicks "Show all" and they only see 1,000 rows.

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -3531,7 +3539,7 @@ if (!LABKEY.DataRegions) {
 
             msg += "&nbsp;" + "<span class='labkey-button select-none'>Select None</span>";
             var showOpts = [];
-            if (region.showRows !== 'all')
+            if (region.showRows !== 'all' && region.maxRows !== ALL_ROWS_MAX)
```
labkey-nicka: Reviewing this logic makes it apparent that are other ways to easily overburden the server/browser with results. The "Show Selected" and "Show Unselected" options should likely be limited in the same way.

labkey-jeckels (approved):
Looks OK to me.

The only question in my mind is what limit to apply for "all". Anything in the 1,000-10,000 range seems fine.

labkey-nicka:
> The only question in my mind is what limit to apply for "all". Anything in the 1,000-10,000 range seems fine.

I played around with different values and it seems like anything over 5,000 is getting into sluggish page territory. Obviously, this is also influenced by the number of columns in the table but for the sake of this setting I'd just assume at least a few columns.

**Edit:** I have increased the limit to 5,000 (Note: This PR is currently hung "Processing Updates" on GitHub...).

api/webapp/clientapi/dom/DataRegion.js
cnathe: I agree. Having a message like this will be helpful to the user to know what's going on.

api/webapp/clientapi/dom/DataRegion.js
cnathe: Yeah, good point. I'm curious how often those are used with large sets of rows though.

cnathe (approved):

labkey-nicka:
Screwed up merge. Closing in favor of #5067.

PR #5060, opened 2023-12-18
Issue 48810: Field editor should check reserved field names based on the domain kind during create as it does during update

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48810

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/1759

#### Changes
- DomainUtils.validateProperties update to pass in domain kind for the create case to better check reserved field names

### DIFF
--- a/api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
+++ b/api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
@@ -255,7 +255,10 @@ public class SampleTypeDomainKind extends AbstractDomainKind<SampleTypeDomainKin
         Set<String> reserved = new CaseInsensitiveHashSet(RESERVED_NAMES);
 
         if (domain == null)
+        {
+            reserved.remove("Name"); // Issue 48810: See SampleTypeService.createSampleType hasNameProperty
             return reserved;
+        }
 
         ExpSampleType st = getSampleType(domain);
         if (st == null)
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
```diff
@@ -251,11 +251,23 @@ public Set<PropertyStorageSpec> getBaseProperties(Domain domain)
...
+    @Override
+    public Set<String> getReservedPropertyNames(Domain domain, User user, boolean forCreate)
     {
         Set<String> reserved = new CaseInsensitiveHashSet(RESERVED_NAMES);
 
         if (domain == null)
+        {
+            // Issue 48810: See SampleTypeService.createSampleType hasNameProperty
+            if (forCreate)
+                reserved.remove("Name");
```
labkey-susanh, line 267: This is all because of the legacy support for idCols, which we no longer allow in the UI, as far as I know. Do we really need to continue to special-case this?  Do we still support creation of sample types using the id columns from our APIs?

cnathe: Well this is "sort of" related to that idCols thing. The SampleTypeDesigner.tsx has, since the beginning I believe, has had code to add in the "name" col to the domain design right before the saveDomain API is called. If I take this out of the client side code, I get an error that `Either a 'Name' property or an index for idCol1 is required`. So the create domain for sample types actually requires the name column be included in the posted domain object, so I had to include this if statement that says "in the create case, remove 'name' from the set of reserved names to fail for".

labkey-susanh: Ah, I see. Maybe an update to the comment to add more of that explanation would be a good idea?

labkey-susanh (approved):

api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
cnathe: oops, didn't see this message before merging. I'll get that comment updated.

PR #5082, opened 2023-12-26
Issue 49328: Ensure RootMaterialRowId is set for all materials

labkey-nicka:
#### Rationale
This addresses [Issue 49328](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=49328) by updating the upgrade script to repoint row's with invalid LSIDs, LSIDs of samples that do not exist in the system, to themselves for their `RootMaterialRowId` value. This can be done as an update to an existing script because upgrades will fail if there is `NULL` values for `RootMaterialRowId` which will mean this script will need to be re-run.

#### Related Pull Requests
* #4844

#### Changes
* Update upgrade script to account for `NULL` values in `RootMaterialRowId`


labkey-nicka:
It is currently up in the air if we'll target 23.12 for this and create a point release or defer to develop/24.01.

labkey-jeckels (approved):
Assuming we confirm that this is the right target branch, looks good to me.

PR #5086, opened 2023-12-28
Fix cross project sample import for sql server

XingY:
#### Rationale
Sample import now accepts "Container" column. For SQL server, GUID/uniqueidentifier datatype is incompatible with character string datatype. This PR adds the check if the _pkColumnLookupMap key is not GUID for GUID columns during remap cache fetch, in which case it should fetch based on _titleColumnLookupMap instead.

https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SampleManagerCSqlserver/2818132?buildTab=tests&status=failed&name=SMProSamplesCreateAndUpdateTest

#### Related Pull Requests
* 

#### Changes
* skip fetch by _pkColumnLookupMap when pk is non-guid string for guid columns
* switch from IllegalStateException to IllegalArgumentException

### DIFF
--- a/api/src/org/labkey/api/dataiterator/SimpleTranslator.java
+++ b/api/src/org/labkey/api/dataiterator/SimpleTranslator.java
@@ -387,6 +387,14 @@ public class SimpleTranslator extends AbstractDataIterator implements DataIterat
             }
             else
             {
+                if (pkCol.getJdbcType() == JdbcType.GUID)
+                {
+                    if (k == null)
+                        return null;
+                    if (k instanceof String strKey && !GUID.isGUID(strKey))
+                        return null;
+                }
+
                 try
                 {
                     if (_allowBulkLoads && _bulkLoads.add(Pair.of(pkCol, pkCol)))
--- a/api/src/org/labkey/api/dataiterator/SimpleTranslator.java
+++ b/api/src/org/labkey/api/dataiterator/SimpleTranslator.java
@@ -387,16 +387,24 @@ public class SimpleTranslator extends AbstractDataIterator implements DataIterat
             }
             else
             {
-                if (_allowBulkLoads && _bulkLoads.add(Pair.of(pkCol, pkCol)))
+                try
                 {
-                    TableSelector ts = createSelector(pkCol, pkCol);
-                    ts.forEach(pkCol.getJavaObjectClass(), (Object pk) -> map.put(pk, pk));
+                    if (_allowBulkLoads && _bulkLoads.add(Pair.of(pkCol, pkCol)))
+                    {
+                        TableSelector ts = createSelector(pkCol, pkCol);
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -387,16 +387,32 @@ private Object fetch(Pair<ColumnInfo, Map<?,?>> pair, Object k)
...
+                    }
+                    else
+                    {
+                        TableSelector ts = createSelector(pkCol, pkCol, k);
+                        ts.forEach(pkCol.getJavaObjectClass(), (Object pk) -> map.put(pk, pk));
+                    }
                 }
+                catch (Exception e)
+                {
+                    // ignore
```
labkey-susanh: Where is the exception thrown?  We didn't have a try-catch before.

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -387,16 +387,32 @@ private Object fetch(Pair<ColumnInfo, Map<?,?>> pair, Object k)
...
             {
-                if (_allowBulkLoads && _bulkLoads.add(Pair.of(pkCol, pkCol)))
+                if (pkCol.getJdbcType() == JdbcType.GUID)
                 {
-                    TableSelector ts = createSelector(pkCol, pkCol);
-                    ts.forEach(pkCol.getJavaObjectClass(), (Object pk) -> map.put(pk, pk));
+                    if (k == null)
+                        return null;
+                    if (k instanceof String strKey && !GUID.isGUID(strKey))
+                        return null;
```
labkey-susanh, line 395: these two early returns seem to skip over adding the MISS value into the map. Is that intentional?

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
XingY: forEach calls rs.next, which throws SQLException

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
XingY: good point. updated to set MISS

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
labkey-susanh: Is that really something we want to ignore?

XingY: It's debatable. This is the call to query by PK, one can argue it's ok for it to silently fail and then the next attempt to query by title will be tried. For my specific GUID/string issue, the check added above should prevent the exception. I've removed the ignore, to minimize the scope of changes.

labkey-susanh (approved):

PR #5091, opened 2024-01-01
When creating entities, check name requirements against current container only

labkey-susanh:
#### Rationale
The setting for whether to allow users to supply names for entities is folder-specific. When creating new entity types, we need to check all containers in context to determine if any do not allow users to supply names and thus require a naming pattern for all entity types, but when creating the entities themselves, we should always be checking the current container's setting to determine if user-supplied names are allowed.

#### Related Pull Requests
* #5036
* https://github.com/LabKey/biologics/pull/2609

#### Changes
* Update checks on whether to allow names to be provided during insert and update


XingY (approved):
working for me now

PR #5101, opened 2024-01-05
Add support for generating API keys in apps and for deleting existing keys

labkey-susanh:
#### Rationale
For LKSM Prof+ clients, we support usage of APIs, so we want to bring the ability to generate and delete API keys into the applications.  Here we replace the existing JSP for generating API tokens with the react component also used in the apps, which brings with it the ability to delete existing API keys.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/1385
* https://github.com/LabKey/testAutomation/pull/1778

#### Changes
* Add ProductFeature.apiKeys 
* Replace `apiKey.jsp` with `apiKeys` react entry point
* Add a UserApiKeysTableInfo, a filtered table over the existing APIKeysTable to the Core schema, but not to the Schema browser


core/src/org/labkey/core/security/SecurityController.java
```diff
@@ -104,6 +113,7 @@
 import org.labkey.core.user.SecurityAccessView;
 import org.labkey.core.user.UserController;
 import org.labkey.core.user.UserController.AccessDetailRow;
+import org.springframework.dao.DataIntegrityViolationException;
```
labkey-adam: Just a bunch of unused imports?

core/src/org/labkey/core/query/CoreQuerySchema.java
```diff
@@ -161,6 +167,8 @@ public TableInfo createTable(String name, ContainerFilter cf)
             return getDataStatesTable();
         if (API_KEYS_TABLE_NAME.equalsIgnoreCase(name) && getUser().hasRootPermission(UserManagementPermission.class))
             return new ApiKeysTableInfo(this);
+        if (USER_API_KEYS_TABLE_NAME.equalsIgnoreCase(name))
```
labkey-adam, line 171: Optional: could add a comment we're intentionally not adding this to schema browser tree

labkey-adam (approved):

core/src/client/APIKeys/APIKeys.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+import React, { FC } from 'react';
+import { APIKeysPanel, AppContextProvider, ServerContextProvider, withAppUser, } from '@labkey/components';
+import { getServerContext } from '@labkey/api';
+
+import '../DomainDesigner.scss';
+
+export const App: FC<any> = () => {
+    const serverContext = withAppUser(getServerContext());
```
cnathe, line 23: glad this pattern worked and wasn't too much overhead

core/src/client/entryPoints.js
```diff
@@ -74,5 +74,9 @@ module.exports = {
         title: 'Query Select Input',
         generateLib: true, // used in TypeAheadSelectDisplayColumn.java
         path: './src/client/QuerySelectInput'
+    }, {
+        name: 'apiKeys', // used in ApiKeyViewProvider.java
```
cnathe, line 78: shouldn't this entryPoint config set `permission: 'admin'`? 
The main usage is in the ApiKeyViewProvider.java which likely checks the perm (haven't look yet) but a user could still technically access this page via the /views.

labkey-susanh: No. Everyone gets to look at their own API keys and generate them if enabled.

core/src/org/labkey/core/security/SecurityController.java
labkey-susanh: Removed.

core/src/org/labkey/core/query/CoreQuerySchema.java
labkey-susanh: Done.

PR #5102, opened 2024-01-05
Initial support for Plate Sets - server side

labkey-klum:
#### Rationale
This change adds initial support for plate sets according to this [spec](https://docs.google.com/document/d/1i6qP3inaj5fFD6z0M-ZpaD4HLPmj1Z4wuj3BvmKRqFg/edit). Plate sets are intended to associate multiple plates together. This PR contains the following changes:
- Initial schema and changes for plate sets and plate support
- Expose schema changes to query
- `QueryUpdateService` implementation for the `PlateSet` table
- `NameExpression` support for both plate sets and plates. Generated names are site unique and non-configurable for this initial implementatation.
- Ensure that all plates must be assigned to a plate set. A default plate set is created if no existing plate set is specified
- Upgrade script to establish new plate sets for every plate in the system

### DIFF
--- a/assay/src/org/labkey/assay/plate/query/PlateTable.java
+++ b/assay/src/org/labkey/assay/plate/query/PlateTable.java
@@ -20,21 +20,26 @@ import org.apache.commons.lang3.ObjectUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.assay.plate.Plate;
+import org.labkey.api.assay.plate.PlateService;
 import org.labkey.api.collections.CaseInsensitiveHashSet;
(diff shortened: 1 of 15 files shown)


assay/src/org/labkey/assay/plate/query/PlateTable.java
```diff
@@ -110,8 +115,10 @@ public MutableColumnInfo wrapColumn(ColumnInfo col)
         var columnInfo = super.wrapColumn(col);
 
         // require a plate name
+/*
```
labkey-nicka: Should these lines now be removed?

labkey-nicka (approved):

labkey-nicka (approved):

PR #5110, opened 2024-01-09
Automated test fix - faulty logic for undefined visits

labkey-klum:
#### Rationale
One last [test failure](https://teamcity.labkey.org/buildConfiguration/LabKey_241RElease_Community_DailySuites_DailyAPostgres/2831423?buildTab=tests&status=failed&name=DatasetPublishTest.testSteps)...

I'm surprised this didn't get exposed earlier because the logic wasn't correct in determining when a new visit would be created (the rowId would be zero if the visit already existed).


labkey-danield (approved):
I know the title says an "Automated test fix", but this looks more like a product fix.

It looks ok to me but probably should get @cnathe to validate as well.

PR #5121, opened 2024-01-11
Add nonce to tracking script template

labkey-tchad:
#### Rationale
Tracking scripts are triggering our new content security policy. They need the nonce.

#### Related Pull Requests
* N/A

#### Changes
* Add nonce to Google analytics tracking script template
* Add instructions for adding nonce to custom tracking scripts

PR #5122, opened 2024-01-11
Add nonce to tracking script template

labkey-tchad:
#### Rationale
Tracking scripts are triggering our new content security policy. They need the nonce.

#### Related Pull Requests
* Replaces #5121 

#### Changes
* Add nonce to Google analytics tracking script template
* Add instructions for adding nonce to custom tracking scripts


labkey-susanh (commented):
Seems right to me, but I'm no expert. I defer to Matt on approval here.

labkey-willm:
> Seems right to me, but I'm no expert. I defer to Matt on approval here.

same.

labkey-matthewb (approved):

PR #5123, opened 2024-01-12
Fix EHR sql server error

XingY:
#### Rationale
#5089 introduced a change to use the same transaction in a nested way. This turns out to be causing issues for sql server, see [related failures](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_EhrSqlserver/2840238?buildTab=overview&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true
)

This PR reverts the !transacted change to use NO_OP transaction for the inner transaction, and seek to handle transaction audit on an alternative current transaction. 
Good [run](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_EhrSqlserver/2848313?hideProblemsFromDependencies=false&expandBuildTestsSection=true&hideTestsFromDependencies=false&expandBuildChangesSection=true) 

#### Related Pull Requests
* 

#### Changes
* revert !transacted changes made in #5089 
* use alternative non NO_OP transaction for audit event, if available

### DIFF
--- a/query/src/org/labkey/query/controllers/QueryController.java
+++ b/query/src/org/labkey/query/controllers/QueryController.java
@@ -4436,6 +4436,11 @@ public class QueryController extends SpringActionController
 
         protected JSONObject executeJson(JSONObject json, CommandType commandType, boolean allowTransaction, Errors errors) throws Exception
         {
+            return executeJson(json, commandType, allowTransaction, errors, null);
+        }
+
+        protected JSONObject executeJson(JSONObject json, CommandType commandType, boolean allowTransaction, Errors errors, @Nullable TransactionAuditProvider.TransactionAuditEvent tAuditEvent) throws Exception
+        {
             JSONObject response = new JSONObject();
             Container container = getContainerForCommand(json);
             User user = getUser();
@@ -4550,6 +4555,8 @@ public class QueryController extends SpringActionController
                 {
                     if (transaction.getAuditEvent() != null)
                         auditEvent = transaction.getAuditEvent();
+                    else if (tAuditEvent != null)
+                        auditEvent = tAuditEvent;
                     else
                     {
                         auditEvent = AbstractQueryUpdateService.createTransactionAuditEvent(container, commandType.getAuditAction());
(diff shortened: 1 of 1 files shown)


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -4976,7 +4980,7 @@ else if (scope != tableInfo.getSchema().getScope())
                     }
                     commandObject.put("extraContext", commandExtraContext);
 
-                    JSONObject commandResponse = executeJson(commandObject, command, transacted, errors);
+                    JSONObject commandResponse = executeJson(commandObject, command, !transacted, errors);
```
labkey-susanh: Don't quite get why this got switched to `!transacted`.   This says if it's not transacted we'll allow transaction if the JSON property is true, but if it is transacted here and that property is true, we don't allow transactions?

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -4548,12 +4548,16 @@ protected JSONObject executeJson(JSONObject json, CommandType commandType, boole
             {
                 if (behaviorType != null && behaviorType != AuditBehaviorType.NONE)
                 {
-                    if (transaction.getAuditEvent() != null)
-                        auditEvent = transaction.getAuditEvent();
+                    DbScope.Transaction auditTransaction = transacted ? transaction : table.getSchema().getScope().getCurrentTransaction();
+                    if (auditTransaction == null)
+                        auditTransaction = transaction;
```
labkey-susanh: I'm not sure I understand this. If `transacted` is true, we use the ensured transaction from the try block and if it's false we get the transaction from the table schema. Presumably if it's transacted, these are the same thing. If there is no transaction on the table schema, we use the `NO_OP_TRANSACTION`.  So could this be
`
```suggestion
DbScope.Transaction auditTransaction = table.getSchema().getScope().getCurrentTransaction();
if (auditTransaction == null)
    auditTransaction = NO_OP_TRANSACTION;
```

query/src/org/labkey/query/controllers/QueryController.java
XingY: It has been like this for 12 years. I think it got flipped because it was trying to avoid nested transaction and have the outer transaction used only. Attempt to reverse is resulted in sql server transaction already rolled back error.

query/src/org/labkey/query/controllers/QueryController.java
XingY: I've modified the logic a little bit more to only get scope transaction when the outer caller is indeed transacted.

labkey-susanh (approved):
Thanks. That's a bit more understandable

PR #5125, opened 2024-01-12
Sample Type Exclusion from Dashboard Insights Panel

cnathe:
#### Rationale
The app dashboard features a Dashboard Insights Panel, which includes bar charts like "Sample Count By Status" and "Sample Count By Sample Type." Clients using Sample Types for inventory purposes have identified a need to exclude certain types from these charts to avoid clutter and enhance data relevance. This set of PRs allow for sample type exclusions from the dashboard graphs in a similar way to how sample types are excluded from Projects (i.e. via the Project Settings page or from the Sample Type Designer).

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1393

#### Changes
- Add DashboardSampleType to DataTypeExclusion types and DB persistence
- remove any related dashboard exclusions for a sample type when it is excluded from the project


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8660,7 +8661,13 @@ public void ensureContainerDataTypeExclusions(@NotNull DataTypeForExclusion data
         if (!toAdd.isEmpty())
         {
             for (Integer add : toAdd)
+            {
                 addDataTypeExclusion(add, dataType, excludedContainerId, user);
+
+                // if this exclusion type has a related exclusion, remove the related exclusion on add
```
XingY: I'm not quite understanding this.. Is the idea to make sure there is no double exclusion (excluded by project and also excluded on dashboard)? 

What about the following scenario:
- exclude SampleType1 from "data in project" section on Projects page, save, but don't refresh the page.
- then exclude SampleType1 from dashboard on Project1. The option is still there before you refresh. 

is there hard in having the data type double excluded? would it make sense for this to be a client-side sync/clean up?

cnathe: Yes, this is to prevent "double exclusions" (for a sample type, if it is excluded for the project then it doesn't need to also be excluded for the dashboard of that project). Having "double exclusions" aren't really an issue and won't cause issues in the app, they just aren't necessary so I added this code to remove them. 

As you mentioned, yes ideally we'd handle this on the client side and prevent it in the first place, but I spoke to Hannah and she said that the dev for that just isn't worth while until we find out how users are going to use this. It seems unlikely that they'll actually use both the project exclusions and the dashboard exclusions together (more likely to use one or the other).

XingY (approved):

PR #5142, opened 2024-01-18
Support all labkey.xml/ROOT.xml settings via application.properties

### LINKED ISSUE #5140: [LK CI] Ensure we handle all properties currently configurable through ROOT.xml/labkey.xml
[Tomcat 10 upgrade Github epic](https://github.com/LabKey/platform/issues/5140)

### Tasks 📍

- [x] Propagate JMS, LDAP, and other settings from application.properties
- [x] Update comments and error messages
- [x] Redirector for legacy context paths
- [x] Handle HTTP and HTTPS concurrently
- [x] Handle session timeout correctly
- [x] Debug config for free IntelliJ
- [x] Fix Mondrian unit tests
- [x] Fix JSP recompilation
- [x] Revisit application.properties handling for JMS, etc
- [x] Fix additional web apps📍
- [x] doc handoff

labkey-jeckels:
#### Rationale
We're enabling full compatibility for configuration via embedded Tomcat.

#### Changes
* Update references to config files in comments

### DIFF
--- a/api/src/org/labkey/api/settings/AppPropsImpl.java
+++ b/api/src/org/labkey/api/settings/AppPropsImpl.java
@@ -469,14 +469,12 @@ class AppPropsImpl extends AbstractWriteableSettingsGroup implements AppProps
         }
     }
 
-    // Get the name of the webapp configuration file, e.g., labkey.xml, cpas.xml, or ROOT.xml.  Used in some error messages
+    // Get the name of the webapp configuration file, now always application.properties.  Used in some error messages
     //  to provide suggestions to the admin.
     @Override
     public String getWebappConfigurationFilename()
     {
-        String path = getContextPath();
-
-        return "".equals(path) ? "ROOT.xml" : path.substring(1) + ".xml";
+        return "application.properties";
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/settings/AppPropsImpl.java
```diff
@@ -469,14 +469,12 @@ public UsageReportingLevel getUsageReportingLevel()
...
-    // Get the name of the webapp configuration file, e.g., labkey.xml, cpas.xml, or ROOT.xml.  Used in some error messages
+    // Get the name of the webapp configuration file, now always application.properties.  Used in some error messages
     //  to provide suggestions to the admin.
     @Override
     public String getWebappConfigurationFilename()
     {
-        String path = getContextPath();
-
-        return "".equals(path) ? "ROOT.xml" : path.substring(1) + ".xml";
+        return "application.properties";
```
labkey-susanh, line 477: Is this overridden somewhere so we can still return the .xml files? I think the tomcat10 branch will get merged before we are able to all be using embedded all the time.

labkey-jeckels: No, it's now hard-coded to always return `application.properties`. I could do work to be able to differentiate between standalone and embedded Tomcat, but for the moment I'm proceeding optimistically that by 24.3 on-premise deployments (where users are most likely to see these messages) will all be configured via the properties file. If you think that's a bad idea, let me know

labkey-susanh: Seems OK (now that I've looked at what this method actually does) and easy enough to rework if need be.

labkey-susanh (approved):

PR #5143, opened 2024-01-19
Server side plate type support

### LINKED ISSUE #5114: Plate Sets - Plate Types (server) - dev
[Spec](https://docs.google.com/document/d/1UapOPmM0QJ3vWlBbxYFm7vHe2oycWs9f_8xSede8udc/edit#heading=h.x8pj9t8hkau)
[Github epic](https://github.com/LabKey/biologics/issues/2531)
fb_plate_type_schema (platform, commonAssays)

#### Background
Plate types represent the physical plate that an experiment might be performed upon and typically are described in the number of wells on the plate (rows and columns). We want to be able to support a plate type column on plates that could be a lookup to a plate type table. LabKey server does have the notion of plate types but they are not defined in the database and, instead are defined in java code and registered by a `PlateTypeHandler`. The motivation for this abstraction is to allow plate type handlers to create plates (or templates) which have predefined well groups (controls, replicates, samples, etc.) to simplify the workflow during plate template creation for our supported assay types (NAb, ELISpot, ELISA, Standard). This is less important in the application workflow and will become obsolete when we have a modern plate designer/editor.

For this story, the goal is to decouple plate type registration from the handlers and move to a model where types are defined and sourced from the database but still maintain handlers for the legacy workflow where well group creation is important.

### Tasks 
- [x] Plate type schema - 2hr
- [x] Update plate schema - 1hr
- [x] Query support - 2hr
- [x] Refactor
  - [x] Separate PlateType from PlateTypeHandler - 3hr
  - [x] Create `PlateLayout` and `PlateLayoutHandler` - 2hr
  - [x] Replace rows/columns with `PlateType` param - 3hr
  - [x] Update `PlateService` with new signatures - 2hr
- [x] Clean up endpoints - 2hr
- [x] Plate upgrade script
  - [x] Replace type with plate type ID - 1hr
  - [x] Remove rows & columns fields? - 1hr
- [x] Junit tests - .5hr
- [x] Rename "type" to "assayType". - 1hr 
- [x] Repurpose "PlateType" as what is currently "PlateTypeId" - 1hr
- [x] Introduce "archived" column to plate.PlateType and set 1536 and invalid plate types to archived = true - .5hr
- [x] Add a "PlateID" column to plate.Plate and replicate name expression value into "PlateID" column - 1hr
- [x] Add a "PlateSetID" column to plate.PlateSet and replicate name expression value into "PlateSetID" column - 1hr
- [x] Add a "WellsFilled" column to plate.Plate - 1hr
- [x] Code review - 1hr
- [x] Teamcity review and merge - 1hr

labkey-klum:
#### Rationale
Introduces plate types to the server. Plate types represent the physical plate that an experiment might be performed upon and typically are described in the number of wells on the plate (rows and columns). 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1395
- https://github.com/LabKey/labkey-ui-premium/pull/307
- https://github.com/LabKey/biologics/pull/2655
- https://github.com/LabKey/platform/pull/5143
- https://github.com/LabKey/commonAssays/pull/726

#### Changes
- Creates an `assay.PlateType` table and populates it with the standard sizes we plan to support.
- Adds column `PlateType` to the `assay.Plate` table.
- Drops columns : rows & columns from `assay.Plate`
- Add column `assay.Plate.PlateId` to be populated with the plate name expression. Name is an optional column which will be populated with the plate ID if it is not provided.
- Adds column `PlateSet.PlateSetId` to be populated with the plate set name expression. Name is an optional column which will be populated with the plate set ID if it is not provided.
- Refactor server side code to use `PlateType` in places where rows and columns were used.


assay/src/org/labkey/assay/AssayUpgradeCode.java
```diff
@@ -157,7 +158,7 @@ private static void _ensureColumn(String colName, Domain domain, ExpProtocol pro
 
     /**
      * Called from assay-23.002-23.003.sql
-     *
+     * <p>
```
labkey-nicka, line 161: nit: Just curious, what purpose does this serve? Is IntelliJ recommending it?

assay/src/org/labkey/assay/plate/model/PlateTypeImpl.java
```diff
@@ -1,21 +1,25 @@
...
     private Integer _rows;
     private String _type;
 
-    public PlateType()
+    public PlateTypeImpl()
     {
     }
 
-    public PlateType(String assayType, String type, String description, Integer rows, Integer cols)
+    public PlateTypeImpl(String assayType, String type, String description, Integer rows, Integer cols)
```
labkey-nicka: I believe `_assayType` and `_type` can both be removed from this class.

nit: This constructor is not used.

assay/src/org/labkey/assay/plate/PlateImpl.java
```diff
@@ -50,16 +51,16 @@ public class PlateImpl extends PropertySetImpl implements Plate, Cloneable
...
     private int _createdBy;
     private long _created;
     private int _modifiedBy;
     private long _modified;
     private String _dataFileId;
-    private String _type;
+    private String _assayType;
     private boolean _isTemplate;
     private Integer _plateSetId;
+    private Integer _plateType;
```
labkey-nicka, line 63: You might consider holding onto the rich `PlateType` instance (instead? as well? I think this depends on database serialization needs). It can still serialize the plate types rowId but it will mean that `getPlateTypeObject()` will not need to query the database each time it is called. Additionally, avoids the NPE implications of calling out for the plate type.

assay/src/org/labkey/assay/plate/PlateLayout.java
```diff
@@ -0,0 +1,42 @@
+package org.labkey.assay.plate;
+
+import org.labkey.api.assay.plate.PlateType;
+
+/**
+ * Plate layouts encompass plate types but also include assay specific information
+ */
+public class PlateLayout
```
labkey-nicka: nit: Could consider using `record` (will require updating getter usages):

```java
public record PlateLayout(String name, PlateType plateType, String assayType, String description) { }
```

Otherwise, all class variables can be marked as `final`.

labkey-nicka (approved):

assay/src/org/labkey/assay/AssayUpgradeCode.java
labkey-klum: yes this came from IntelliJ but I don't know why?

assay/src/org/labkey/assay/plate/model/PlateTypeImpl.java
labkey-klum: I'll update that bean and delete the constructor, archived is also missing.

assay/src/org/labkey/assay/plate/PlateImpl.java
labkey-klum: Agreed, we probably want to set the `PlateType` object in `PlateManager.populatePlate` to be consistent with the caching pattern that is used by `PlateCache`

assay/src/org/labkey/assay/plate/PlateLayout.java
labkey-klum: yeah, good suggestion, I've haven't gotten a lot of opportunity to use the `record` keyword.

PR #5154, opened 2024-01-23
Add option to require comments for data changes

labkey-susanh:
#### Rationale
We currently give users with the option to provide comments on certain changes being made to their data. In regulated environments, the comments are not optional, so for products beyond our LKSM Starter, we are adding a configuration for the application that will make these comments required in the various places that we offer this option. (At later story will add the ability to provide comments in more places.)

#### Related Pull Requests
- https://github.com/LabKey/sampleManagement/pull/2401
- https://github.com/LabKey/biologics/pull/2665
- https://github.com/LabKey/inventory/pull/1171
- https://github.com/LabKey/labkey-ui-premium/pull/311
- https://github.com/LabKey/labkey-ui-components/pull/1399

#### Changes
* Add `DataChangeCommentRequirement` product feature
* Add property set and property for storing the option to require audit comments on data changes
* add `auditCommentsRequiredContainerCount` metric
* Update `moveRows` operation to verify a reason is provided if required


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -4787,12 +4787,18 @@ public void validateForm(MoveRowsForm form, Errors errors)
...
             {
-                errors.reject(ERROR_MSG, "Empty request");
+                errors.reject(ERROR_GENERIC, "Empty request");
             }
             else
             {
-                String queryName = json.optString(PROP_QUERY_NAME, null);
-                _targetContainer = ContainerManager.getMoveTargetContainer(queryName, getContainer(), getUser(), getTargetContainerProp(), errors);
+                // Since we are moving between containers, we know we have product projects enabled
+                if (getContainer().getProject().getAuditCommentsRequired() && StringUtils.isBlank(json.optString("auditUserComment")))
```
cnathe, line 4795: do we know that we always want to be doing `getContainer().getProject()`? what if the container is already a project? oh maybe that will just always return itself.

cnathe (approved):

query/src/org/labkey/query/controllers/QueryController.java
labkey-susanh: Yes, getProject for a project returns itself.

PR #5156, opened 2024-01-24
Support additional context paths

labkey-jeckels:
#### Rationale
Some server configurations have historically used non-root context paths, and/or deployed additional web apps for delivering static content.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/687

#### Changes
* Property to control LabKey's context path: `context.contextPath`
* Property to automatically redirect/forward requests from a legacy context path to the root: `context.legacyContextPath`
* Properties to deploy extra web apps: `webapps.contextPath` and `webapps.docBase`

### DIFF
--- a/api/src/org/labkey/api/view/RedirectorServlet.java
+++ b/api/src/org/labkey/api/view/RedirectorServlet.java
@@ -1,0 +1,51 @@
+//
+// Source code recreated from a .class file by IntelliJ IDEA
+// (powered by FernFlower decompiler)
+//
+
+package org.labkey.api.view;
+
+import java.io.IOException;
+
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServlet;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+
+/** Simple redirector to redirect and forward from legacy context paths to the root context */
+public class RedirectorServlet extends HttpServlet
+{
+    private final String _legacyContextPath;
+
+    public RedirectorServlet(String legacyContextPath)
+    {
+        if (!legacyContextPath.startsWith("/") || legacyContextPath.length() < 2)
+        {
+            throw new IllegalArgumentException("Invalid legacy context path: " + legacyContextPath);
+        }
+        _legacyContextPath = legacyContextPath;
+    }
+
+    @Override
+    protected void service(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException
+    {
+
+        if ("get".equalsIgnoreCase(request.getMethod()))
+        {
+            // Send a redirect to let the client know there's a new preferred URL
+            String originalUrl = request.getRequestURL().toString();
+            String redirectUrl = originalUrl.replaceFirst(_legacyContextPath, "");
+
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/view/RedirectorServlet.java
```diff
@@ -0,0 +1,51 @@
+//
+// Source code recreated from a .class file by IntelliJ IDEA
+// (powered by FernFlower decompiler)
+//
```
labkey-susanh: Interesting...

api/src/org/labkey/api/view/RedirectorServlet.java
```diff
@@ -0,0 +1,51 @@
...
+/** Simple redirector to redirect and forward from legacy context paths to the root context */
+public class RedirectorServlet extends HttpServlet
+{
+    private final String _legacyContextPath;
+
+    public RedirectorServlet(String legacyContextPath)
+    {
+        if (!legacyContextPath.startsWith("/") || legacyContextPath.length() < 2)
+        {
+            throw new IllegalArgumentException("Invalid legacy context path: " + legacyContextPath);
```
labkey-susanh: Should this error message use the property name to make it easier to track down where the error is?

api/src/org/labkey/api/view/RedirectorServlet.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    protected void service(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException
+    {
+
+        if ("get".equalsIgnoreCase(request.getMethod()))
+        {
+            // Send a redirect to let the client know there's a new preferred URL
+            String originalUrl = request.getRequestURL().toString();
+            String redirectUrl = originalUrl.replaceFirst(_legacyContextPath, "");
```
labkey-susanh, line 31: Should there be some error message somewhere if both the legacyContextPath and contextPath are provided and/or if they are the same thing?

api/src/org/labkey/api/view/RedirectorServlet.java
labkey-jeckels: Hah, IntelliJ had helpfully collapsed this section so I hadn't noticed it. I started with a redirector servlet that I wrote many years ago but I think the source code got lost in the SVN->GitHub migration. I still had the binaries in email though.

api/src/org/labkey/api/view/RedirectorServlet.java
labkey-jeckels: I added a check to LabKeyServer.java. It's only useful to set a legacyContextPath if you're deploying to the root, so the error checks for that.

api/src/org/labkey/api/view/RedirectorServlet.java
labkey-jeckels: Updated.

api/src/org/labkey/api/view/RedirectorServlet.java
labkey-susanh: Thanks!

labkey-susanh (approved):

PR #5162, opened 2024-01-26
Plate cache updates and DB serialization clean up.

labkey-klum:
#### Rationale
Removes support for plate name in the cache. Plate by name is `deprecated` and could eventually be removed. Instead the more recent `plateId` has been added to the cache.

This PR also introduces a `PlateBean` to help with serialization to and from the database. This allows us to be a bit cleaner with the `Plate` instance which is a richer, composed object so we don't have to have methods like:
- `getPlateType` & `getPlateTypeObject`
- `getPlateSet` & `getPlateSetObject`

Which were needed to support DB binding plus returning the underlying object instances.

### DIFF
--- a/assay/api-src/org/labkey/api/assay/plate/Plate.java
+++ b/assay/api-src/org/labkey/api/assay/plate/Plate.java
@@ -40,9 +40,9 @@ public interface Plate extends PropertySet, Identifiable
 
     boolean isTemplate();
 
-    @NotNull PlateType getPlateTypeObject();
+    @NotNull PlateType getPlateType();
 
-    @Nullable PlateSet getPlateSetObject();
+    @Nullable PlateSet getPlateSet();
 
     /**
      * Returns an existing well, or creates a new well if one
--- a/assay/api-src/org/labkey/api/assay/plate/Plate.java
+++ b/assay/api-src/org/labkey/api/assay/plate/Plate.java
@@ -44,6 +44,8 @@ public interface Plate extends PropertySet, Identifiable
 
     @Nullable PlateSet getPlateSet();
 
+    @NotNull String getPlateId();
+
     /**
      * Returns an existing well, or creates a new well if one
      * had not previously existed.
--- a/assay/src/org/labkey/assay/AssayUpgradeCode.java
+++ b/assay/src/org/labkey/assay/AssayUpgradeCode.java
@@ -205,7 +205,7 @@ public class AssayUpgradeCode implements UpgradeCode
     {
         // resolve plate by the legacy deprecated plate name method
         ObjectProperty prop = protocol.getObjectProperties().get(protocol.getLSID() + AbstractPlateBasedAssayProvider.PLATE_TEMPLATE_SUFFIX);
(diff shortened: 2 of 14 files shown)


assay/api-src/org/labkey/api/assay/plate/Plate.java
```diff
@@ -40,9 +40,11 @@ public interface Plate extends PropertySet, Identifiable
 
     boolean isTemplate();
 
-    @NotNull PlateType getPlateTypeObject();
+    @NotNull PlateType getPlateType();
```
labkey-nicka, line 43: Yay, much more straightforward.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -1276,7 +1292,7 @@ public DilutionCurve getDilutionCurve(List<WellGroup> wellGroups, boolean assume
     @Override
     public List<? extends PlateType> getPlateTypes()
     {
-        return new TableSelector(AssayDbSchema.getInstance().getTableInfoPlateType()).getArrayList(PlateTypeImpl.class);
+        return new TableSelector(AssayDbSchema.getInstance().getTableInfoPlateType()).getArrayList(PlateTypeBean.class);
```
labkey-nicka, line 1296: nit: The interface method is annotated with `@NotNull`. Consider adding here as well.

assay/src/org/labkey/assay/AssayUpgradeCode.java
```diff
@@ -205,7 +205,7 @@ private static Plate getPlate(ExpProtocol protocol)
     {
         // resolve plate by the legacy deprecated plate name method
         ObjectProperty prop = protocol.getObjectProperties().get(protocol.getLSID() + AbstractPlateBasedAssayProvider.PLATE_TEMPLATE_SUFFIX);
-        return prop != null ? PlateManager.get().getPlate(protocol.getContainer(), prop.getStringValue()) : null;
+        return prop != null ? PlateManager.get().getPlateByName(protocol.getContainer(), prop.getStringValue()) : null;
```
labkey-nicka, line 208: Just curious, did you verify these changes through the upgrade(s)?

assay/src/org/labkey/assay/plate/model/PlateBean.java
```diff
@@ -0,0 +1,140 @@
+package org.labkey.assay.plate.model;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import org.labkey.api.data.Entity;
+import org.labkey.assay.plate.PlateImpl;
+
+/**
+ * Serializes a row in the assay.plate table.
+ */
+public class PlateBean extends Entity
```
labkey-nicka, line 10: This is great. Splitting these up will serve us well I believe and leave us less prone to improper serialization on either side.

labkey-nicka (approved):

assay/src/org/labkey/assay/AssayUpgradeCode.java
labkey-klum: Yes I ran the upgrade manually, but you have to revert the protocol property to use the plate name unless you upgrade from an older version. Did you see an error on your end?

PR #5163, opened 2024-01-27
restore mondrian dependencies

labkey-matthewb:
#### Rationale
Mondrian seems to work fine if we restore its dependencies (which no longer conflict with tomcat).

#### Related Pull Requests
* 

#### Changes
*


labkey-jeckels:
This approach seems reasonable. I think the alternative would be to run the javax -> jakarta migration tool on the Mondrian JARs and host them in our Artifactory.

Something doesn't look right in the PR though. I think it's missing your commit. Also, if we're going to have both versions of the servlet API on the class path, we should configure IntelliJ to not suggest the javax version for auto-import

labkey-jeckels:
Replaced with https://github.com/LabKey/platform/pull/5173

PR #5164, opened 2024-01-27
Add server version info to CSP reports

labkey-adam:
#### Rationale
We want CSP reports to include the originating server's version information (and potentially other properties in the future). https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49489

#### Related Pull Requests
* https://github.com/LabKey/server/pull/693

#### Changes
* Move `ContentSecurityPolicyFilter` to API to give it access to properties and helpers
* Register ContentSecurityPolicyFilter in `ContextListener.contextInitialized()` (embedded case). It seems that filters must be registered very, very early (e.g.,`ModuleLoader.init()` is too late).
* Add `${CSP.REPORT.PARAMS}` substitution to inject server properties into the CSP report JSON
* Use `StringExpression` to perform CSP substitutions
* Add `NullValueBehavior.ReplaceNullAndMissingWithBlank` and use it to ensure unknown substitutions are stripped from the CSP. Older servers ignoring unknown substitutions should ease the roll-out of future CSP changes.
* Fix comment on `NullValueBehavior.ReplaceNullWithBlank` for accuracy
* Temporarily append the ${CSP.REPORT.PARAMS} substitution on all `report-uri` URLs that lack query parameters. Will be removed once we propagate this substitution into our CSPs.

### DIFF
--- a/api/src/org/labkey/api/util/StringExpressionFactory.java
+++ b/api/src/org/labkey/api/util/StringExpressionFactory.java
@@ -672,7 +672,8 @@ public class StringExpressionFactory
                 }
             },
 
-            // Null fields get replaced with blank. Any missing field results in null eval.
+            // Null fields get replaced with blank. Any missing field results in null eval of the whole expression
+
             ReplaceNullWithBlank(StringExpressionType.ReplaceMissing.BLANK_VALUE)
             {
                 @Override
--- a/api/src/org/labkey/api/util/StringExpressionFactory.java
+++ b/api/src/org/labkey/api/util/StringExpressionFactory.java
@@ -664,34 +664,50 @@ public class StringExpressionFactory
         {
             // Any null field results in a null eval (good for URLs)
             NullResult(StringExpressionType.ReplaceMissing.NULL_RESULT)
-                    {
-                        @Override
-                        public String handleNull(StringExpressionFactory.StringPart part) throws StopIteratingException
-                        {
-                            throw new StopIteratingException();
-                        }
-                    },
+            {
+                @Override
+                public String handleNull(StringExpressionFactory.StringPart part) throws StopIteratingException
+                {
+                    throw new StopIteratingException();
+                }
+            },
 
-            // Null or missing fields get replaced with blank
+            // Null fields get replaced with blank. Any missing field results in null eval.
             ReplaceNullWithBlank(StringExpressionType.ReplaceMissing.BLANK_VALUE)
-                    {
-                        @Override
-                        public String handleNull(StringExpressionFactory.StringPart part)
-                        {
-                            return "";
-                        }
-                    },
+            {
+                @Override
+                public String handleNull(StringExpressionFactory.StringPart part)
+                {
+                    return "";
+                }
+            },
 
+            // Null and missing fields get replaced with blank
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/util/StringExpressionFactory.java
```diff
@@ -664,34 +664,50 @@ public enum NullValueBehavior
...
+            {
+                @Override
+                public String handleNull(StringExpressionFactory.StringPart part) throws StopIteratingException
+                {
+                    throw new StopIteratingException();
+                }
+            },
 
-            // Null or missing fields get replaced with blank
+            // Null fields get replaced with blank. Any missing field results in null eval.
```
labkey-jeckels:
```suggestion
            // Null fields get replaced with blank. Any missing field results in null eval of the whole expression

```

api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
```diff
@@ -0,0 +1,231 @@
...
+                s = s.replace((char)0x2019, (char)0x027);     // RIGHT SINGLE QUOTATION MARK -> APOSTROPHE
+
+                // This is temporary. TODO: Remove once we've propagated this substitution into our CSPs
+                String directive = "report-uri";
+                int reportUriIdx = StringUtils.indexOfIgnoreCase(s, directive);
+                if (reportUriIdx != -1)
+                {
+                    int semicolonIdx = s.indexOf(';', reportUriIdx);
+
+                    if (semicolonIdx != -1)
```
labkey-jeckels, line 140: Is there always a trailing semicolon, or might that be omitted if report-uri is the last directive?

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -11101,30 +11101,32 @@ public void setHourDelta(Integer hourDelta)
...
-            var ret = new JSONObject().put("success",true);
+            var ret = new JSONObject().put("success", true);
 
             // fail fast
             if (!_log.isWarnEnabled())
                 return ret;
 
-            var userAgent = getViewContext().getRequest().getHeader("User-Agent");
+            var request = getViewContext().getRequest();
+            assert null != request;
```
labkey-jeckels, line 11123: Feel free to keep it, but I've never understood the value of asserts like this. We'll either fail here with an AssertionError or on the next line with a NPE. Is that a valuable difference when asserts are enabled?

labkey-jeckels (approved):

core/src/org/labkey/core/admin/AdminController.java
labkey-matthewb: This avoid the IntelliJ yellow squiggle telling you that the code could throw a NPE.

labkey-adam: Exactly... seems to be the best way to suppress the warning without incurring needless overhead (in production)

api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
labkey-adam: It might be omitted, in which case we won't append the substitution param, which is fine. I believe all the CSPs that LabKey deploys and tests include the semicolon.

PR #5170, opened 2024-01-30
Issue #49400: Fix RServe credential issue

Sigmonia:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=49400

#### Changes
* Added allowBindParameter override to allow User parameter in script engine config


Sigmonia:
@labkey-jony and I confirmed the fix with our test RServe

labkey-klum (approved):

PR #5175, opened 2024-01-31
Switch ImportPanel from client-side LABKEY.requiresScript() to server-side XML

bbimber:
Starting a little while ago (~3-4 months), we started to get error logs from query-importData, which suggested some kind of problem with loading timing of the Ext4 dependencies. I have been able to repro this intermittently as well. Looking at importData.html, one thing I see is that the problem components, ExcelUploadPanel and Ext4FormPanel, were being loaded using the client-side requiresScript(). I dont think there is any good reason to use this mechanism. I dont have a great answer as to the root cause, but my guess is that some change in browser JS loading behavior might cause instances where requireScript() properly load the dependency.

This PR is a simple change to remove requiresScript() and replace it with ClientDependencies in the XML file. This should be a much more reliable mechanism. The only usage I see of LABKEY.ext.ImportPanel is in importData.html, so the impact of this change should be limited.

### DIFF
--- a/core/webapp/extWidgets/ImportPanel.js
+++ b/core/webapp/extWidgets/ImportPanel.js
@@ -4,9 +4,9 @@
  * Licensed under the Apache License, Version 2.0: http://www.apache.org/licenses/LICENSE-2.0
  */
 
-LABKEY.requiresScript("/extWidgets/ExcelUploadPanel.js");
-LABKEY.requiresScript("/extWidgets/Ext4FormPanel.js");
-
+/*
+ * NOTE: ExcelUploadPanel.js and Ext4FormPanel.js are required for this component
+ */
 Ext4.define('LABKEY.ext.ImportPanel', {
     extend: 'Ext.tab.Panel',
     initComponent: function(){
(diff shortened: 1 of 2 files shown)


labkey-martyp:
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49339

bbimber:
thanks @labkey-martyp. Finding the actual cause is probably a good idea (since the problem is probably within requiresScript(); however, this change might not be a bad idea either. The only reason I can think that ImportPanel is using requiresScript() is that it pre-dates ClientDependency.

labkey-nicka:
> Finding the actual cause is probably a good idea (since the problem is probably within requiresScript(); however, this change might not be a bad idea either. The only reason I can think that ImportPanel is using requiresScript() is that it pre-dates ClientDependency.

The cause is known. Moving to server side is definitely the recommend pattern. Using `requiresScript` without a dependency callback is not recommend as it can provide inconsistent results. It logs a warning in the console about this as well.

bbimber:
@labkey-nicka: so are you good approving this PR then? I see no reason to update these to use a callback when we can just use the server-side route as proposed here...

labkey-nicka (approved):

labkey-nicka:
> @labkey-nicka: so are you good approving this PR then? I see no reason to update these to use a callback when we can just user the server-side route as proposed here...

Sorry, didn't see the approval status. Was just commenting on the thread. Approved.

labkey-adam (approved):

PR #5176, opened 2024-01-31
WebSocket: ensure single notification websocket

labkey-nicka:
#### Rationale
Initialization of the notifications web socket on the client needs to ensure there is only one notification web socket connection at a time.

#### Changes
- Dereference `_websocket` when the connection is closed
- Move defensive checks to `openWebsocket()`


cnathe (approved):
manual tested as well, looks good

PR #5177, opened 2024-01-31
Issue 49440: App user defined field is mapped incorrectly if the user doesn't have active read+ access

cnathe:
#### Rationale
Issue [49440](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49440): App "User" user defined field is mapped incorrectly if the user doesn't have active read+ access

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1405

#### Changes
- If user display name isn't available via the core.Users lookup, try to get it from UserManager


api/src/org/labkey/api/query/UserIdRenderer.java
```diff
@@ -88,4 +89,20 @@ public String renderURL(RenderContext ctx)
 
         return null;
     }
+
+    @Override
+    public Object getDisplayValue(RenderContext ctx)
+    {
+        Object displayValue = super.getDisplayValue(ctx);
+
+        // Issue 49440: If user display name isn't available via the core.Users lookup, try to get it from UserManager
```
labkey-susanh, line 98: This seems fine, but I wonder if we should always just get it from UserManager?

labkey-susanh (approved):

PR #5186, opened 2024-02-02
FR47180 Holistic instrument tracking folders - outlier level setting ability

ankurjuneja:
#### Rationale
This PR provides support for displaying lower bound and upper bound outlier level setting with related PR.

#### Related Pull Requests
* https://github.com/LabKey/targetedms/pull/822


core/webapp/vis/src/plot.js
```diff
@@ -2175,41 +2206,71 @@ boxPlot.render();
...
+                            geom: new LABKEY.vis.Geom.ErrorBar({size: 1, color: 'red', dashed: true, width: barWidth, topOnly: true}),
+                            data: meanStdDevData,
+                            aes: {
+                                error: function (row) {
+                                    return row[config.properties.stdDev] * config.properties.lowerBound;
+                                },
+                                yLeft: config.properties.mean
+                            }
+                        }));
+                        // Are we OK suppressing the 1x and 2x bars for all use cases?
```
labkey-jeckels: I think we decided that we want to retain the 1x and 2x bars for non-Panorama use cases. Is that happening already or do we need to handle that? @cnathe can provide non-Panorama use cases if that's helpful.

labkey-jeckels (approved):
Question on the +/- 1 and 2 SD lines. Please either remove the comment or make sure that they are retained for non-Panorama use cases.

core/webapp/vis/src/plot.js
ankurjuneja: I retained those with a checkbox option but 2nd item in feature verification suggested to remove those. CC: @wendyinnis

labkey-jeckels: Right - we don't want to show them in Panorama. But we do want to show them in other places that use the Levey-Jennings plot but always use +/- 3 SDs as the cutoffs

PR #5194, opened 2024-02-05
remove usages of new ObjectMapper

labkey-matthewb:
#### Rationale
Make sure shared DEFAULT_MAPPER is not mutable

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/api/src/org/labkey/api/util/JsonUtil.java
+++ b/api/src/org/labkey/api/util/JsonUtil.java
@@ -23,11 +23,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.core.JsonpCharacterEscapes;
 import com.fasterxml.jackson.core.io.CharacterEscapes;
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.MappingJsonFactory;
-import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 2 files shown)


labkey-adam:
@labkey-matthewb I like this a lot, but here's a suggestion: Instead of the `checkLocked()` pattern, create a generic `ImmutableObjectMapper` that wraps `ObjectMapper`, but every setter throws. `LabKeyObjectMapper` extends `ObjectMapper` with our special settings. `DEFAULT_MAPPER` is an ImmutableObjectMapper that wraps LabKeyObjectMapper. I like splitting immutability from our custom settings, and I prefer this to using checkLocked(), but my other motivation is we might get the Jackson crew to accept a generic an ImmutableObjectMapper into their code base (and maintain it instead of us).

labkey-matthewb:
> @labkey-matthewb I like this a lot, but here's a suggestion: Instead of the `checkLocked()` pattern, create a generic `ImmutableObjectMapper` that wraps `ObjectMapper`, but every setter throws. `LabKeyObjectMapper` extends `ObjectMapper` with our special settings. `DEFAULT_MAPPER` is an ImmutableObjectMapper that wraps LabKeyObjectMapper. I like splitting immutability from our custom settings, and I prefer this to using checkLocked(), but my other motivation is we might get the Jackson crew to accept a generic an ImmutableObjectMapper into their code base (and maintain it instead of us).

This was my first thought too.  However, ObjectMapper is a class not an interface, so it would be a bit awkward?  We'd have to instantiate a dummy class that delegates to the wrapped class.

labkey-adam (approved):

PR #5197, opened 2024-02-06
Issue 49338: LKSM: Sample ID generation is only sometimes case sensitive

XingY:
#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5197
- https://github.com/LabKey/sampleManagement/pull/2442

* https://github.com/LabKey/platform/pull/4175

#### Changes
*


api/src/org/labkey/api/data/DbSequenceManager.java
```diff
@@ -328,9 +325,8 @@ static Pair<Long,Long> reserve(DbSequence sequence, int count)
 
         addValueSql(sql, tinfo, sequence);
 
-        sql.append(") + ? WHERE Container = ? AND RowId = ?");
+        sql.append(") + ? WHERE RowId = ?");
```
labkey-susanh, line 313: I'm curious if the removal of the Container clause was a necessary thing or just an optimization.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -2384,6 +2384,52 @@ public String getValue(Map map)
...
+        {
+            String prefix = prefixRaw.trim().toLowerCase(); // Issue 49338: withCounter should be case-insensitive
+            DbSequence counterSeq = null;
+            if (noCache || !counterSequences.containsKey(prefix) || _strictIncremental)
+            {
+                long existingCount = -1;
+
+                if (_strictIncremental || AppProps.getInstance().isExperimentalFeatureEnabled(EXPERIMENTAL_WITH_COUNTER))
+                {
+                    // TODO: use DbSequence.ReclaimableDbSequence for 23.3 and investigate enabling ReclaimablePreallocateSequence in develop
```
labkey-susanh: I see this got moved from down below. Still a TODO worth keeping around?

core/src/org/labkey/core/CoreUpgradeCode.java
```diff
@@ -201,4 +205,97 @@ public static void uniquifyPrincipalsName(ModuleContext context)
         UpgradeUtils.uniquifyValues(name, ownerId, new SimpleFilter(type.getFieldKey(), "u"), new Sort("UserId"), false, false);
         UpgradeUtils.uniquifyValues(name, ownerId, new SimpleFilter(type.getFieldKey(), "m"), new Sort("UserId"), false, false);
     }
+
+    /**
+     * Remove WithCounter (SampleNameGenCounter-) core.DBSequences records with case-insensitive names, keep the one with the largest Value
+     */
+    private static void removeDuplicateWithCountSeqs(Container container)
```
labkey-susanh: I might call this `removeDuplicateWithCounterSeqs`

core/src/org/labkey/core/CoreUpgradeCode.java
```diff
@@ -201,4 +205,97 @@ public static void uniquifyPrincipalsName(ModuleContext context)
...
+            if (!toRemove.isEmpty())
+            {
+                SQLFragment deleteSql = new SQLFragment("DELETE FROM ").append(tableInfo).append(" WHERE RowId");
+                deleteSql = tableInfo.getSqlDialect().appendInClauseSql(deleteSql, toRemove);
+                new SqlExecutor(tableInfo.getSchema()).execute(deleteSql);
+            }
+        }
+    }
+
+    private static void toLowerCaseWithCountSeqs(Container container)
```
labkey-susanh:
```suggestion
    private static void toLowerCaseWithCounterSeqs(Container container)

```

labkey-susanh (approved):

api/src/org/labkey/api/data/DbSequenceManager.java
labkey-adam: I'm curious too. Including Container is usually better from a security standpoint and often improves performance, depending on what indexes exist.

XingY: I made this change here since @labkey-matthewb questioned in https://github.com/LabKey/platform/pull/4175 the need of inclusion of Container. I think the idea is the fewer columns involved, the less likely a deadlock would occur.  Though for my scenarios, keeping/removing makes no difference in the outcome of deadlocks. I'm ok with adding them back.

XingY: In the case of performance, since RowId is already PK, the inclusion of "container" probably won't add much benefit.

labkey-matthewb: I don't think security plays any role here.  The container is not being inferred from the URL for instance.  It is part of the DbSequence that we created.

api/src/org/labkey/api/data/NameGenerator.java
XingY: Good point. I removed the TODO as well as related code. In my testing, ReclaimablePreallocateSequence didn't seem to provide perf benefit anyways.

PR #5199, opened 2024-02-06
Convert well properties to a provisioned domain

### LINKED ISSUE #5166: Plate Architecture : Convert well properties to provisioned table
[Spec](https://docs.google.com/document/d/1i6qP3inaj5fFD6z0M-ZpaD4HLPmj1Z4wuj3BvmKRqFg/edit)

 In a previous story, custom plate properties were implemented using a `vocabulary domain` which was joined to the `assay.well` table. The domain was scoped at the project level. Vocabulary domain are simple to add support for a table because of previous work that Kevin did in both the `UserSchema` and `QueryUpdateService`. However, vocabulary domains rely on `Ontology` to store and resolve values and is not as efficient as a provisioned solution. The conclusion was made that vocab domains are not a good match for plates and the decision was made to move towards a solution that joins the well data to a domain backed by a provisioned table.

Additionally, since the existing behavior was never exposed outside of an experimental flag, we plan to not support migrating from one domain to another and instead will drop the vocabulary domains on upgrade.

#### Tasks
- [x] WellDomainKind - 2 hr
- [x] Refactor PlateService endpoints to use new domain - 2 hr
- [x] Remove support in WellTable for vocabulary domains - 1 hr
- [x] Join WellTable to provisioned table - 1 hr
- [x] Remove support in WellUpdateService for vocab domains - 1 hr
- [x] Update WellUpdateService to insert in to both well and properties table - 3 hr
- [x] Clean up provisioned rows on plate or container delete - 3 hr
- [x] Drop well vocabulary domains - 2 hr
- [x] Drop assay.plateProperty values - 1 hr
- [x] Update junit tests, api endpoints - 2 hr
- [x] code reviews - 1
- [x] Manual Testing - 2 @labkey-chrisj 
- [ ] tc review and merge -1

labkey-klum:
#### Rationale
The initial implementation of custom well properties was implemented using a `vocabulary` domain. Due to perf considerations plus discounting the need to support scenarios that `vocabulary` domains were designed for, this PR transitions to using a provisioned domain architecture.

Overall, this is a two table architecture, similar to sample types, data classes, issues etc. where there is a hard table with the base well columns : `assay.well`. Custom properties are modeled using a project scoped domain and the provisioned table is joined to the hard table using the `lsid` value.

#### Changes
- Implement `PlateMetadataDomainKInd` to represent custom well properties.
- Transition endpoints to refer to the new domain.
- Update the well `QueryUpdateService` to handle the second table.
- Adjust plate and container cleanup code to use the new domain.
- Upgrade script to create the new provisioned schema, drop existing vocabulary domains. Since the plate feature is still under an experimental flag, we elected not to support an upgrade path of custom properties.
- Plate unique name validator [not working](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49665)


labkey-klum:
One thing I've noticed is that the `LSID` column on the provisioned table shows up in the UI. I do mark it as hidden but we may need to filter it out on the client. 
image

assay/src/org/labkey/assay/AssayModule.java
```diff
@@ -107,7 +108,7 @@ public String getName()
     @Override
     public Double getSchemaVersion()
     {
-        return 24.002;
+        return 24.003;
```
labkey-nicka, line 113: Does this feel like the right time to change how/when we initialize the plate metadata domain? Currently, the server does not guarantee that the domain is initialized for a particular folder so the client has to be defensive against this state (and it is not in all cases at the moment).

Perhaps now is the time to just start ensuring the domain at startup of the biologics module? This would have to be done for all Biologics top-level folders so is this better done somewhere/how else? Could also argue for the assay module but then we'll need to make sure we don't initialize for non-supported server configurations.

labkey-nicka:
> One thing I've noticed is that the `LSID` column on the provisioned table shows up in the UI. I do mark it as hidden but we may need to filter it out on the client.

Hmm, I'll have to investigate. I'm hoping there is a way to hide this via metadata that the application is already respecting.

**Edit:** The `Lsid` column is not returning as `hidden` nor `isHidden` via `query-getQueryDetails.api`:

image

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -786,7 +795,7 @@ public Object execute(CustomFieldsForm form, BindException errors) throws Except
         }
     }
 
-    @RequiresPermission(DesignVocabularyPermission.class)
+    @RequiresPermission(AddUserPermission.class)
```
labkey-nicka: I don't think `AddUserPermission` is intended here.

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -671,7 +680,7 @@ public void setFields(List<GWTPropertyDescriptor> fields)
         }
     }
 
-    @RequiresPermission(DesignVocabularyPermission.class)
+    @RequiresPermission(AdminPermission.class)
```
labkey-nicka: Let's relax this to editor permissions.

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -714,7 +723,7 @@ public void setFields(List<PlateCustomField> fields)
...
```
labkey-nicka: Let's relax this to editor permissions.

assay/src/org/labkey/assay/plate/PlateMetadataDomainKind.java
```diff
@@ -0,0 +1,156 @@
...
+import org.labkey.api.security.permissions.AdminPermission;
+import org.labkey.api.util.UnexpectedException;
+import org.labkey.api.view.ActionURL;
+import org.labkey.api.writer.ContainerUser;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+public class PlateMetadataDomainKind extends BaseAbstractDomainKind
```
labkey-nicka, line 31: Bug. I added a couple fields to the new metadata table and then attempted to add them to a specific plate:

image

assay/src/org/labkey/assay/plate/PlateMetadataDomainKind.java
```diff
@@ -0,0 +1,156 @@
...
+            return PropertyService.get().getDomain(container, domainURI);
+        }
+        catch (Exception e)
+        {
+            throw UnexpectedException.wrap(e);
+        }
+    }
+
+    @Override
+    public DbSchemaType getSchemaType()
```
labkey-nicka: nit: This can be removed as it is identical to the parent implementation.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -955,6 +953,8 @@ public void afterPlateDelete(Container container, Plate plate)
     public void beforePlateDelete(Container container, Integer plateId)
     {
         final AssayDbSchema schema = AssayDbSchema.getInstance();
+        DbScope scope = schema.getSchema().getScope();
+        assert scope.isTransactionActive();
```
labkey-nicka: Rather than asserting feels like this should throw an error if called from outside a transaction.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -1030,20 +1035,24 @@ public void deleteAllPlateData(Container container)
...
                 .append(" WHERE Container = ? )")
                 .add(container);
-        new SqlExecutor(AssayDbSchema.getInstance().getSchema()).execute(sql2);
+        new SqlExecutor(AssayDbSchema.getInstance().getSchema()).execute(sql);
+
+        // delete any plate metadata values from the provisioned table
+        TableInfo provisionedTable = getPlateMetadataTable(container, User.getAdminServiceUser());
+        if (provisionedTable != null)
+        {
+            SQLFragment sql2 = new SQLFragment("DELETE FROM ").append(provisionedTable, "")
```
labkey-nicka, line 1169: This statement is never executed against the database.

assay/src/org/labkey/assay/plate/query/WellTable.java
```diff
@@ -100,45 +102,76 @@ public void addColumns()
...
             }
         }
     }
 
-    private MutableColumnInfo addVocabularyDomainColumns(Domain domain, @NotNull String lookupColName)
+    /**
+     * Alternatively, if we wish to join the provisioned fields into the well table without introducing another hierarchy
+     * level we just wrap the fields. This also requires the code : getFromSql to be uncommented to handle the join.
+     */
+    private void addWellPropertiesFlattened()
```
labkey-nicka, line 150: This method is not used.

assay/src/org/labkey/assay/plate/query/WellTable.java
```diff
@@ -100,45 +102,76 @@ public void addColumns()
...
-        colProperty.setFk(new PropertyForeignKey(_userSchema, getContainerFilter(), domain));
-        colProperty.setUserEditable(false);
-        colProperty.setIsUnselectable(true);
-        colProperty.setCalculated(true);
+        Domain wellDomain = PlateManager.get().getPlateMetadataDomain(getContainer(), getUserSchema().getUser());
+        FieldKey lsidFieldKey = FieldKey.fromParts("lsid");
+        Supplier<Map<DomainProperty, Object>> defaultsSupplier = null;
 
-        return addColumn(colProperty);
+        for (ColumnInfo col : PlateManager.get().getPlateMetadataTable(getContainer(), _userSchema.getUser()).getColumns())
```
labkey-nicka: (If this method ends up being used) Should be defensive here against a null result from `getPlateMetadataTable()`.

assay/src/org/labkey/assay/plate/query/WellTable.java
```diff
@@ -195,34 +210,109 @@ public MutableColumnInfo wrapColumn(ColumnInfo col)
...
 
     @Override
-    public ObjectUriType getObjectUriType()
+    public List<FieldKey> getDefaultVisibleColumns()
     {
-        return schemaColumn;
+        return defaultVisibleColumns;
     }
 
+/*
```
labkey-nicka, line 222: Remove?

assay/src/org/labkey/assay/AssayModule.java
labkey-klum: Sure, currently we only guarantee it on folder creation time, but we can initialize existing Biologics folders as well. Should we do this regardless of the experimental flag, given that it will be removed at some point?

assay/src/org/labkey/assay/plate/PlateMetadataDomainKind.java
labkey-klum: This should be fixed now, it was a problem with the casing for the `FieldKey` returned by the server.

assay/src/org/labkey/assay/plate/query/WellTable.java
labkey-klum: yes, keeping it here for the moment if we change our minds about how the properties are exposed.

assay/src/org/labkey/assay/plate/PlateMetadataDomainKind.java
labkey-klum: Additionally, the `lsid` field is returning `hidden=true` but is still showing up in the dialog. Maybe if you can take a look and let me know why it's still showing up?

assay/src/org/labkey/assay/AssayModule.java
labkey-nicka: Yes, I think regardless of the experimental flag.

labkey-nicka (approved):

PR #5200, opened 2024-02-06
Automatically add Google Analytics URLs to CSP when enabled

labkey-jeckels:
#### Rationale
We know the external URLs that will be used for Google Analytics tracking, so they're safe and easy to register automatically with the content security policy.

#### Changes
* Inject Google URLs into CSP when GA tracking is enabled

### DIFF
--- a/core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
+++ b/core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
@@ -46,8 +46,7 @@ public class AnalyticsServiceImpl implements AnalyticsService
 {
     private static final String SEPARATOR = ",";
     private static final String GOOGLE_TAG_MANAGER_URL = "https://www.googletagmanager.com";
-    private static final String ANALYTICS_CSP_KEY_1 = AnalyticsServiceImpl.class.getName() + "_1";
-    private static final String ANALYTICS_CSP_KEY_2 = AnalyticsServiceImpl.class.getName() + "_2";
+    private static final String ANALYTICS_CSP_KEY = AnalyticsServiceImpl.class.getName();
 
     public static AnalyticsServiceImpl get()
     {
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
```diff
@@ -109,6 +113,18 @@ public String getDescription()
...
     }
 
+    public void resetCSP()
+    {
+        ContentSecurityPolicyFilter.unregisterAllowedConnectionSource(ANALYTICS_CSP_KEY_1);
+        ContentSecurityPolicyFilter.unregisterAllowedConnectionSource(ANALYTICS_CSP_KEY_2);
+
+        if (getTrackingStatus().contains(TrackingStatus.ga4FullUrl))
+        {
+            ContentSecurityPolicyFilter.registerAllowedConnectionSource(ANALYTICS_CSP_KEY_1, GOOGLE_TAG_MANAGER_URL);
```
labkey-matthewb: Too bad this method isn't multi-valued

labkey-matthewb (changes requested):
I don't think this is solving any problem.  a) AllowedConnectionSource is for connect-src not script-src b) script-src uses strict dynamic so the problem is with including the nonce, not the presense/absence of the host.

labkey-jeckels:
> I don't think this is solving any problem. a) AllowedConnectionSource is for connect-src not script-src b) script-src uses strict dynamic so the problem is with including the nonce, not the presense/absence of the host.

Without these changes, on my dev machine where I have GA enabled (the built-in option, not the custom analytics script option), I get CSP violations reported. They go away when I run with this code. We have these two URLs in the `connect-src` for the CSPs we're deploying on our container instances today. Is that a wrong example to model this on?

labkey-matthewb:
> > I don't think this is solving any problem. a) AllowedConnectionSource is for connect-src not script-src b) script-src uses strict dynamic so the problem is with including the nonce, not the presense/absence of the host.
> 
> Without these changes, on my dev machine where I have GA enabled (the built-in option, not the custom analytics script option), I get CSP violations reported. They go away when I run with this code. We have these two URLs in the `connect-src` for the CSPs we're deploying on our container instances today. Is that a wrong example to model this on?

Sorry, I'm still thinking about the problem with today's deployments caused by not having the the script-nonce set.  Once we get pass that, we still have to make the API calls to do the actually analytics tracking... Those do need the connect-src.

So yes this would be useful.  mea culpa

labkey-matthewb (approved):

labkey-adam (approved):

core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
labkey-jeckels: I thought the same thing. Your comment inspired me to make it var-arg, now pushed into this FB.

labkey-jeckels:
> > > I don't think this is solving any problem. a) AllowedConnectionSource is for connect-src not script-src b) script-src uses strict dynamic so the problem is with including the nonce, not the presense/absence of the host.
> > 
> > 
> > Without these changes, on my dev machine where I have GA enabled (the built-in option, not the custom analytics script option), I get CSP violations reported. They go away when I run with this code. We have these two URLs in the `connect-src` for the CSPs we're deploying on our container instances today. Is that a wrong example to model this on?
> 
> Sorry, I'm still thinking about the problem with today's deployments caused by not having the the script-nonce set. Once we get pass that, we still have to make the API calls to do the actually analytics tracking... Those do need the connect-src.
> 
> So yes this would be useful. mea culpa

OK, great. Yes, this doesn't do anything for the custom `<script>` scenario

PR #5208, opened 2024-02-08
Fix 49234: CreatedBy is not displayed for some files

Sigmonia:
#### Rationale
[Secure Issue 49234: FileContent isn't always displaying CreatedBy despite having CreatedBy in the audit log](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=49234)

Notes: 
* Chose to use the first across the board to have a consistent source
* The usage column links use the expDatas link to runs and the analysis 

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/731
* https://github.com/LabKey/platform/pull/5208

#### Changes
* Modified the set of metadata properties to use the first/oldest always

### DIFF
--- a/api/src/org/labkey/api/webdav/AbstractWebdavResource.java
+++ b/api/src/org/labkey/api/webdav/AbstractWebdavResource.java
@@ -726,6 +726,8 @@ public abstract class AbstractWebdavResource extends AbstractResource implements
                     list.addAll(provider.getExpDataByPath(path, container));
                 }
             }
+
+            //Sort the results by creation date so the original is used for metadata display
             _data = list.stream().sorted(Comparator.comparing(ExpObject::getCreated)).toList();
         }
         return _data;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/webdav/AbstractWebdavResource.java
```diff
@@ -715,7 +726,7 @@ protected List<ExpData> getExpDatasHelper(@Nullable java.nio.file.Path path, Con
                     list.addAll(provider.getExpDataByPath(path, container));
                 }
             }
-            _data = list;
+            _data = list.stream().sorted(Comparator.comparing(ExpObject::getCreated)).toList();
```
labkey-jeckels, line 731: Since other methods are relying on this ordering, adding a comment here and to `getExpData()` to make the expectation clear would be nice.

labkey-jeckels (approved):

PR #5210, opened 2024-02-08
Follow Google's CSP guidance for connect-src

labkey-jeckels:
#### Rationale
My previous change was too restrictive in the Google Analytics hosts being allowed for the Content Security Policy

#### Changes
* Allow wildcards and one other domain


labkey-adam (approved):
Good thing this is varags now!

PR #5219, opened 2024-02-09
Plates: switch to non-hidden experimental flag

labkey-nicka:
#### Rationale
Retain and display publicly the experimental support for plating samples.

#### Changes
- Default to displaying the experimental flag publicly


labkey-klum (approved):

labkey-chrisj:
Manual testing- was able to confirm that the plates in biologics switch is no longer hidden

PR #5220, opened 2024-02-09
Fix cloned user display name

labkey-adam:
#### Rationale
New `ClonedUser` constructor was confused about which parameter was which, causing 
`org.labkey.inventory.query.InventoryBoxTable$TriggerTest` to fail and likely other problems

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5203


labkey-jeckels (approved):
Always good to lead with friendly

labkey-adam:
> Always good to lead with friendly

Is it time to add `getPreferredPronouns()`?

PR #5222, opened 2024-02-10
Fix ClusterStartup test cases

labkey-jeckels:
#### Rationale
We're not getting a pipeline-lib directory and need to get the servlet API onto the class path for the forked process

#### Changes
* Extract the servlet API JAR, and treat its home as the pipeline-lib directory


labkey-tchad (approved):
Looks good on TeamCity. There are a few similar references in `DiscvrLabKeyModules` as well.

PR #5223, opened 2024-02-12
Substitute report parameters during init()

labkey-adam:
#### Rationale
Now that we ensure `ModuleLoader.init()` is called before any filter initialization, we can substitute the report URL parameters once, during filter `init()` instead of on every request

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5178

### DIFF
--- a/api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
+++ b/api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
@@ -1,5 +1,13 @@
 package org.labkey.filters;
 
+import jakarta.servlet.Filter;
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.FilterConfig;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
 import org.apache.commons.lang3.StringUtils;
 import org.labkey.api.settings.AppProps;
 import org.labkey.api.settings.ExperimentalFeatureService;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
```diff
@@ -112,11 +112,11 @@ public class ContentSecurityPolicyFilter implements Filter
     private static final String ALLOWED_CONNECT_SUBSTITUTION = "LABKEY.ALLOWED.CONNECTIONS";
     private static final String REPORT_PARAMETER_SUBSTITUTION = "CSP.REPORT.PARAMS";
     private static final String HEADER_NONCE = "org.labkey.filters.ContentSecurityPolicyFilter#NONCE";  // needs to match PageConfig.HEADER_NONCE
-    private static final String CONTENT_SECURITY_POLICY_HEADER_NAME = "Content-Security-Policy";
-    private static final String CONTENT_SECURITY_POLICY_REPORT_ONLY_HEADER_NAME = "Content-Security-Policy-Report-Only";
-    private static final Map<String, List<String>> allowedConnectionSources = new ConcurrentHashMap<>();
+    private static final Map<String, List<String>> ALLOWED_CONNECTION_SOURCES = new ConcurrentHashMap<>();
```
labkey-matthewb, line 115: Do you use ALL_CAPS for all static final?  I usually only do that for string/number (or otherwise immutable)

labkey-matthewb (approved):

api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
labkey-adam: I usually do (like `LOG`), but I understand why some only do this for "constants"

PR #5226, opened 2024-02-12
Adjust embedded server info property

labkey-adam:
#### Rationale
Switching "embedded/standalone" from boolean to String means the value tells the whole story and substitutions are an option

#### Related Pull Requests
* https://github.com/LabKey/cloudServices/pull/24

### DIFF
--- a/core/src/org/labkey/core/admin/admin.jsp
+++ b/core/src/org/labkey/core/admin/admin.jsp
@@ -104,8 +104,8 @@
                 <tr class="<%=getShadeRowClass(row++)%>"><td>Mode</td><td><%=h(AdminBean.mode)%></td></tr>
                 <tr class="<%=getShadeRowClass(row++)%>"><td>Asserts</td><td><%=h(AdminBean.asserts)%></td></tr>
                 <tr class="<%=getShadeRowClass(row++)%>"><td>Servlet Container</td><td><%=h(AdminBean.servletContainer)%></td></tr>
-                <tr class="<%=getShadeRowClass(row++)%>"><td>Embedded Tomcat</td><td><%=h(AppProps.getInstance().isEmbeddedTomcat())%></td></tr>
-                <tr class="<%=getShadeRowClass(row++)%>"><td>Session Timeout (minutes)</td><td><%=h(AdminBean.sessionTimeout)%></td></tr>
+                <tr class="<%=getShadeRowClass(row++)%>"><td>Servlet Configuration</td><td><%=h(AdminBean.servletConfiguration)%></td></tr>
+                <tr class="<%=getShadeRowClass(row++)%>"><td>Session Timeout (hours)</td><td><%=h(AdminBean.sessionTimeout)%></td></tr>
                 <tr class="<%=getShadeRowClass(row++)%>"><td>Java Runtime Vendor</td><td><%=h(AdminBean.javaVendor)%></td></tr>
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/admin/admin.jsp
```diff
@@ -104,8 +104,8 @@
                 <tr class="<%=getShadeRowClass(row++)%>"><td>Mode</td><td><%=h(AdminBean.mode)%></td></tr>
                 <tr class="<%=getShadeRowClass(row++)%>"><td>Asserts</td><td><%=h(AdminBean.asserts)%></td></tr>
                 <tr class="<%=getShadeRowClass(row++)%>"><td>Servlet Container</td><td><%=h(AdminBean.servletContainer)%></td></tr>
-                <tr class="<%=getShadeRowClass(row++)%>"><td>Embedded Tomcat</td><td><%=h(AppProps.getInstance().isEmbeddedTomcat())%></td></tr>
-                <tr class="<%=getShadeRowClass(row++)%>"><td>Session Timeout (minutes)</td><td><%=h(AdminBean.sessionTimeout)%></td></tr>
+                <tr class="<%=getShadeRowClass(row++)%>"><td>Servlet Configuration</td><td><%=h(AdminBean.servletConfiguration)%></td></tr>
+                <tr class="<%=getShadeRowClass(row++)%>"><td>Session Timeout (hours)</td><td><%=h(AdminBean.sessionTimeout)%></td></tr>
```
labkey-susanh: Hours seems like a strange unit to use here, given that the setting is specified in minutes. If we're wanting to save someone having to do the time math, perhaps we could *also* display it in hours, but report the minutes as given.

labkey-adam: The property is set in minutes, and then reported in minutes in one place (`ServletContext`) and seconds in another (`HttpSession` itself). The fact that labkey.org is currently showing 1,440 minutes led me to think hours would be more useful. I doubt folks are setting these timeouts to arbitrary numbers of minutes... I suspect it's typically some multiple of 60. But if there's disagreement I can restore minutes. FWIW: Josh didn't care either way.

labkey-jeckels: As noted, I don't have strong personal feelings here. I suspect the most common value is Tomcat's default of 30 minutes (as a courtesy, I'll translate that here to 0.5 hours). And as a further wrinkle, Spring Boot's config lets you specify the units, so `application.properties` accepts values like `30m`, `2h`, or `15346s`

labkey-susanh: Given that little wrinkle, perhaps reporting as given is the safer route?  Is the default unit for Spring Boot minutes?  We could see if the string is a number and then do math or report a unit and otherwise report just what is given. Not a huge preference, so I won't hold up the PR, but if we're wanting to troubleshoot with this data, the less manipulation that occurs the better, I think.

labkey-jeckels: Inside the web app we're disconnected from however the value was specified, and are pulling the default session timeout from Tomcat (in minutes) via `ServletContext.getSessionTimeout()`

https://jakarta.ee/specifications/platform/9/apidocs/jakarta/servlet/servletcontext#getSessionTimeout--

labkey-jeckels (approved):

labkey-susanh (approved):

PR #5233, opened 2024-02-13
Jakarta JspWriter

labkey-adam:
#### Rationale
Tomcat 10


labkey-jeckels (approved):
Which copy gets picked up at runtime? I don't think we've been updating this copy every time we bump Tomcat versions.

labkey-adam:
> Which copy gets picked up at runtime? I don't think we've been updating this copy every time we bump Tomcat versions.

Tomcat version gets picked up in every scenario I've seen. That's what Tomcat passes into every JSP, and then we wrap it with our wrapper which also resolves to the Tomcat version as its parent. The class is completely abstract... no implementation at all... and really should be an interface; I see no risk even if some code path uses our version. This thing hasn't changed in ages. If it did, we'd likely see compilation issues with our wrapper.

PR #5235, opened 2024-02-14
Move LastIndexed to separate table for exp.Data and exp.Material

labkey-jeckels:
#### Rationale
It's slow to null out a column's value for large tables with lots of indices.

Issue 49577: Clearing last indexed for provider : ExperimentServiceImpl extremely slow

#### Changes
* Move LastIndexed value into a side table, which can be quickly truncated instead of updated in-place


labkey-jeckels:
We don't typically use `ON DELETE CASCADE` for our FKs, but I'm trying it here. It's helped clear up race conditions in the tests due to the multi-threaded nature of indexing. Let me know if you have concerns.

labkey-adam (approved):

PR #5240, opened 2024-02-15
Update next crawl time calculation to match expectations

labkey-adam:
#### Rationale
Previous calculation caused the crawler to visit folders every nine hours (on average). Every 12 hours +/- 10% seems sufficient.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48415


labkey-jeckels (approved):

labkey-matthewb:
I the intention was that changes would be noticed within 12 hrs, but if this solves some problem great.

labkey-adam:
> I the intention was that changes would be noticed within 12 hrs, but if this solves some problem great.

Average just makes it easier to understand if we're indexing more than expected (i.e., if there are 1,100 folder then we should see 1,100 folders reindexed every 12 hours, on average).

labkey-matthewb (approved):

PR #5246, opened 2024-02-19
Issue 49731: Assay import from subfolder-defined plate fails to resolve plate

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49731

PlateManager.getPlateContainerFilter() was using getProductContainerFilterForLookups() to create the container filter for resolving a plate. This is the helper that we use for the isProductProjectsAllFolderScopeEnabled() setting and to resolve lookup fields. Since the plate set grid for a app subfolder shows the parent project plate sets based on the getContainerFilterForLookups() helper, I believe that is the one we should be using here when resolving the plate set.

#### Related Pull Requests
- https://github.com/LabKey/biologics/pull/2727

#### Changes
- PlateManager.getPlateContainerFilter update to use getContainerFilterForLookups() instead of getProductContainerFilterForLookups()

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateManager.java
+++ b/assay/src/org/labkey/assay/plate/PlateManager.java
@@ -581,8 +581,9 @@ public class PlateManager implements PlateService
      */
     public ContainerFilter getPlateContainerFilter(@Nullable ExpProtocol protocol, Container container, User user)
     {
-        ContainerFilter defaultCf = ContainerFilter.Type.Current.create(protocol != null ? protocol.getContainer() : container, user);
-        return QueryService.get().getProductContainerFilterForLookups(container, user, defaultCf);
+        ContainerFilter lookupCf = QueryService.get().getContainerFilterForLookups(container, user);
+        ContainerFilter currentCf = ContainerFilter.Type.Current.create(protocol != null ? protocol.getContainer() : container, user);
+        return lookupCf != null ? lookupCf : currentCf;
     }
 
     @Override


labkey-chrisj:
Just as a heads-up, with this fix `PlateSetTest.testCrossfolderPlateLoading` is able to do some things it asserts it should not:

In the plate detail view, a plate defined in the top folder can view and select samples from subfolders into the wells grid. I think we intended to filter those out because we ran into issues deleting plates when they reference samples in another container

cnathe:
> Just as a heads-up, with this fix `PlateSetTest.testCrossfolderPlateLoading` is able to do some things it asserts it should not:
> 
> In the plate detail view, a plate defined in the top folder can view and select samples from subfolders into the wells grid. I think we intended to filter those out because we ran into issues deleting plates when they reference samples in another container

@labkey-chrisj PlateSetTest.testCrossFolderPlateLoading passes for me locally with this change. Do you have local changes or did you see this failure on TeamCity?

labkey-chrisj:
> > Just as a heads-up, with this fix `PlateSetTest.testCrossfolderPlateLoading` is able to do some things it asserts it should not:
> > In the plate detail view, a plate defined in the top folder can view and select samples from subfolders into the wells grid. I think we intended to filter those out because we ran into issues deleting plates when they reference samples in another container
> 
> @labkey-chrisj PlateSetTest.testCrossFolderPlateLoading passes for me locally with this change. Do you have local changes or did you see this failure on TeamCity?

It was local results, it turns out I had the 'Less restrictrive product project lookups' feature enabled. Runs green locally now with it disabled. Now I'll push test changes to https://github.com/LabKey/biologics/pull/2727 verify the success of this fix going forward

labkey-nicka (approved):

PR #5247, opened 2024-02-19
class change due to commons-compress upgrade

Sigmonia:
#### Rationale
Upgrade version of commons-compress

#### Related Pull Requests
* https://github.com/LabKey/server/pull/726

#### Changes
* Change ArchiveEntry to TarArchiveEntry

### DIFF
--- a/api/src/org/labkey/api/reader/ExcelFactory.java
+++ b/api/src/org/labkey/api/reader/ExcelFactory.java
@@ -20,6 +20,7 @@ import org.apache.logging.log4j.Logger;
 import org.apache.poi.UnsupportedFileFormatException;
 import org.apache.poi.ooxml.POIXMLException;
 import org.apache.poi.openxml4j.exceptions.InvalidFormatException;
+import org.apache.poi.openxml4j.exceptions.InvalidOperationException;
 import org.apache.poi.openxml4j.opc.OPCPackage;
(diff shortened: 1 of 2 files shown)


labkey-adam (approved):

api/src/org/labkey/api/reader/ExcelFactory.java
```diff
@@ -921,7 +922,7 @@ public void testExcelFileImportShouldFail() throws Exception
 
             attemptImportExpectError(new File(dataloading, "doesntexist.xls"), FileNotFoundException.class);
             attemptImportExpectError(new File(dataloading, ""), FileNotFoundException.class);
-            attemptImportExpectError(new File(dataloading, "notreallyexcel.xls"), InvalidFormatException.class);
+            attemptImportExpectError(new File(dataloading, "notreallyexcel.xls"), InvalidOperationException.class);
```
Sigmonia, line 925: AFAICT we handle this exception type similarly

labkey-tchad (approved):
This worries me a little bit. I hope this isn't a sign of some deeper incompatibility between poi and the new version of commons-compress.

Sigmonia:
I don't disagree it is a little trouble some. I did check by upgrading poi to the latest and it seemed happy at least there were no additional test failures/changes that I noticed, but it didn't address the existing one so I didn't include it on this PR.

PR #5249, opened 2024-02-20
Enable editing of workflow templates with no associated jobs

labkey-susanh:
#### Rationale
When a job template has no jobs associated with it, we want to allow users to edit the template so they don't have to know everything about the template before they click "Save" the first time. Previously, we had allowed editing of the name, description, and any domain fields. Now we allow users to edit attachments and tasks as well when there are no jobs using the template. When there are jobs, we still allow editing of the name, description, and fields (consistent with our treatment of other domains), and will display read-only versions of the attachments and tasks.

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-premium/pull/337

#### Changes
* Add `ExperimentService.updateProtocol` method

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -8199,9 +8199,15 @@ public class ExperimentServiceImpl implements ExperimentService, ObjectReference
     @Override
     public ExpProtocol updateProtocol(@NotNull ExpProtocol wrappedProtocol, @Nullable List<ExpProtocol> steps, @Nullable Map<String, List<String>> predecessors, User user) throws ExperimentException
     {
-        Protocol baseProtocol = ((ExpProtocolImpl) wrappedProtocol).getDataObject();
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8196,6 +8196,14 @@ public List<ExpDataImpl> getExpDatasUnderPath(@NotNull String path, @Nullable Co
 
     }
 
+    @Override
+    public ExpProtocol updateProtocol(@NotNull ExpProtocol wrappedProtocol, @Nullable List<ExpProtocol> steps, @Nullable Map<String, List<String>> predecessors, User user) throws ExperimentException
+    {
+        Protocol baseProtocol = ((ExpProtocolImpl) wrappedProtocol).getDataObject();
+        insertProtocolSteps(baseProtocol, steps, predecessors, user, true);
+        return getExpProtocol(baseProtocol.getRowId());
```
cnathe: the insertProtocol() method wraps the inserts in a transaction. any reason this needs to do the same?

cnathe (approved):

PR #5214, opened 2024-02-08
Plate Set Assay Import - Assay Run PlateSet lookup and Results Plate lookup

cnathe:
#### Rationale
This is the first story in the epic to move from having an assay run reference a single plate at the run level to having an assay run reference the plate set at the run level and then reference the plates from that set within the results level.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5214 
- https://github.com/LabKey/labkey-ui-components/pull/1414
- https://github.com/LabKey/labkey-ui-premium/pull/329
- https://github.com/LabKey/biologics/pull/2705
- https://github.com/LabKey/inventory/pull/1196
- https://github.com/LabKey/sampleManagement/pull/2449

#### Changes
- Add Standard Assay run domain field for PlateSet lookup and result domain field for Plate lookup
- Change assay import plate resolution to use result plate lookup value instead of run prop value
- Null check for wellLocation string before determining PositionImpl
- Resolve assay result plate from the plate set based on rowId, plateId, or name
- add check for domain field existence on assay import
- Support import of Plate and WellLocation using import aliases from DomainProperty
- Add metrics for assay run count for "runs with plate template lookup"
- PlateManager.getRunIdsUsingPlateInResults to get run count for references to the results domain Plate field

PR #5260, opened 2024-02-22
Plate Set: Assay Import - Story 2: Assay Import Graphical Layouts / Formats (single measure)

cnathe:
#### Rationale
The next step in the Plate Set Assay Import epic is to support more data file formats then just the tall Table format. This means that for an assay import into an assay design with the "plate metadata" property enabled, we will try a second file parsing that looks for graphical/plate layout data matrices in the file that match the expected plate size/type of the plates for the selected plate set. This story only covers the single measure case. We will extend this to multiple measure support / parsing in the future.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5214
- https://github.com/LabKey/limsModules/pull/6

#### Changes
- only applies when the assay has plate metadata prop enabled
- data file parsing based on the plate type/size defined by the plates in the selected plate set
- allows for data grids with and without a plate identifier
- if any data grids have plate identifiers, only those are processed for import
- if no data grids have plate identifiers, verify that the number of grids doesn't exceed the number of plates in the plate set
- update PlateUtils to use LinkedHashMap to keep order of data grids when parsing
- NOTE: currently only supports numeric data grid values
- NOTE: currently only supports single measure data grid parsing


api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
```diff
@@ -260,6 +234,61 @@ public Map<DataType, List<Map<String, Object>>> getValidationDataMap(ExpData dat
...
+                Object lsid = runProps.getOrDefault(propertyPlateTemplate, null);
+                Lsid templateLsid = lsid != null && !StringUtils.isEmpty(String.valueOf(lsid)) ? Lsid.parse(String.valueOf(lsid)) : null;
+                Integer plateSetId = getPlateSetValueFromRunProps(context, provider, protocol);
+                return svc.mergePlateMetadata(context.getContainer(), context.getUser(), templateLsid, plateSetId, dataRows, plateMetadata, provider, protocol);
+            }
+        }
+        return dataRows;
+    }
+
+    private Integer getPlateSetValueFromRunProps(XarContext context, AssayProvider provider, ExpProtocol protocol) throws ExperimentException
```
labkey-klum, line 279: add the nullable annotation to this method?

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -508,6 +514,87 @@ else if (propEntry.getValue().isBoolean())
...
 
+    @Override
+    public List<Map<String, Object>> parsePlateGrids(Container container, AssayProvider provider, ExpProtocol protocol, Integer plateSetId, File dataFile) throws ExperimentException
+    {
+        // NOTE: currently only supporting single measure assay protocols
+        List<DomainProperty> measureProperties = provider.getResultsDomain(protocol).getProperties().stream().filter(DomainProperty::isMeasure).collect(Collectors.toList());
+        String measureName = !measureProperties.isEmpty() ? measureProperties.get(0).getName() : "Value";
+
+        // get the ordered list of plates for the plate set
+        List<Plate> plates = PlateManager.get().getPlatesForPlateSet(container, plateSetId);
```
labkey-klum: We should verify that the `PlateSet` is resolvable from the passed in `rowId`

assay/src/org/labkey/assay/plate/PlateImpl.java
```diff
@@ -684,4 +692,16 @@ public void setPlateId(String plateId)
...
     }
+
+    @Override
+    public boolean isIdentifierMatch(String id)
+    {
+        return id != null && !id.isEmpty() && (id.equals(getRowId() + "") || id.equalsIgnoreCase(getPlateId()) || id.equalsIgnoreCase(getName()));
+    }
+
+    @Override
+    public boolean isPlateSizeMatch(int rows, int cols)
```
labkey-klum: Rather than introducing a new method, we could just retrieve the `PlateType` from the plate and compare the shape using that object. I believe `PlateType` already supports `hashCode` and `equals`. You could introduce a constructor taking row & col for convenience.

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -508,6 +514,90 @@ else if (propEntry.getValue().isBoolean())
...
+        if (hasPlateIdentifiers)
+        {
+            for (Map.Entry<PlateType, Map<String, double[][]>> plateTypeMapEntry : plateTypeGrids.entrySet())
+            {
+                for (Map.Entry<String, double[][]> entry : plateTypeMapEntry.getValue().entrySet())
+                {
+                    // find the plate set plate for this identifier so we can skip those that don't match the expected plate type
+                    Plate matchingPlate = plates.stream().filter(p -> p.isIdentifierMatch(entry.getKey())).findFirst().orElse(null);
+                    double[][] plateGrid = entry.getValue();
+                    if (matchingPlate != null && matchingPlate.isPlateSizeMatch(plateGrid.length, plateGrid[0].length))
```
labkey-klum: If we think we have plate identifiers but none (or some) of them match then we just take the matching ones? Probably a question for Rob/Nick but I wonder if none match then we fall back to indexed mode for all the data?

labkey-klum (approved):

api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
cnathe: done

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
cnathe: done

assay/src/org/labkey/assay/plate/PlateImpl.java
cnathe: Good idea. Done

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
cnathe: I will ask this during feature verification with Nick and Bernie today.

labkey-chrisj:
FYI I've observed that some of the error messages for importing data from plates are different from what the tests expect- I can address these when they cause test failures.  Also let's chat about what to cover in selenium tests

cnathe:
> FYI I've observed that some of the error messages for importing data from plates are different from what the tests expect- I can address these when they cause test failures. Also let's chat about what to cover in selenium tests

@labkey-chrisj I have a biologics branch created with the fix for those PlateSetTest failures from the error message changes. I'll open a PR after I merge your latest changes to that fb. I'll also ping you to sync on selenium testing for this story.

PR #5263, opened 2024-02-22
49665: Plate names should be unique within a plate set

labkey-klum:
[tracking issue
](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49665)

#### Rationale
Modify the existing checks for plate name uniqueness:
- For Biologics projects or subfolders, the names within the plate set must be unique.
- For everything else, it will default to container uniqueness.

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateManager.java
+++ b/assay/src/org/labkey/assay/plate/PlateManager.java
@@ -621,11 +621,14 @@ public class PlateManager implements PlateService
     }
 
     /**
-     * Checks to see if there is a plate with the same name in the folder, or for
+     * Issue 49665 : Checks to see if there is a plate with the same name in the folder, or for
      * Biologics folders if there is a duplicate plate name in the plate set.
      */
     public boolean isDuplicatePlate(Container c, User user, String name, @Nullable PlateSet plateSet)
     {
+        // Identifying the "Biologics" folder type as the logic we pivot this behavior on is not intended to be
+        // a long-term solution. We will be looking to introduce plating as a ProductFeature which we can then
+        // leverage instead.
         boolean isBiologicsProject = c.getProject() != null && "Biologics".equals(ContainerManager.getFolderTypeName(c.getProject()));
         if (isBiologicsProject && plateSet != null)
         {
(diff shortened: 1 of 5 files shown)


assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -621,12 +621,26 @@ public ContainerFilter getPlateContainerFilter(@Nullable ExpProtocol protocol, C
     }
 
     /**
-     * Checks to see if there is a plate with the same name in the folder
+     * Checks to see if there is a plate with the same name in the folder, or for
```
labkey-nicka: Let's add this associated issue number here.

```java
/**
 * Issue 49665: Checks to see if there is a plate with the same name in the folder, or for
 * Biologics folders if there is a duplicate plate name in the plate set.
 */
```

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -621,12 +621,26 @@ public ContainerFilter getPlateContainerFilter(@Nullable ExpProtocol protocol, C
...
+     * Biologics folders if there is a duplicate plate name in the plate set.
      */
-    public boolean plateExists(Container c, String name)
+    public boolean isDuplicatePlate(Container c, User user, String name, @Nullable PlateSet plateSet)
     {
-        Plate plate = getPlateByName(c, name);
-        return plate != null && plate.getName().equals(name);
+        boolean isBiologicsProject = c.getProject() != null && "Biologics".equals(ContainerManager.getFolderTypeName(c.getProject()));
```
labkey-nicka, line 632: I'd like to give caution about this type of check and what our intentions are. Let's add a comment. Something like:

```java
public boolean isDuplicatePlate(Container c, User user, String name, @Nullable PlateSet plateSet)
{
    // Identifying the "Biologics" folder type as the logic we pivot this behavior on is not intended to be
    // a long-term solution. We will be looking to introduce plating as a ProductFeature which we can then
    // leverage instead.
    boolean isBiologicsProject = c.getProject() != null && "Biologics".equals(ContainerManager.getFolderTypeName(c.getProject()));
    // ...
}
```

labkey-nicka (approved):
Looks good. When adding integration tests I hope this case can be covered, at least in the Biologics case, in `PlateController.ispec.ts` via the APIs.

PR #5265, opened 2024-02-23
Option to continue server startup if existing connections are detected

labkey-adam:
#### Rationale
Some deployments want to bravely continue in the face of existing database connections. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=49494

LabKey will continue to terminate if existing connections are detected to the same database with the same application name, but setting this property will skip termination:

`-DterminateOnExistingConnections=false`

Also added a simple mechanism to test the detection of unexpected connections. This is especially helpful when using embedded Tomcat, since the trivial standalone repro of "copy the context file" no longer works in embedded.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4937

### DIFF
--- a/api/src/org/labkey/api/util/ResultSetUtil.java
+++ b/api/src/org/labkey/api/util/ResultSetUtil.java
@@ -26,6 +26,7 @@ import org.labkey.api.collections.ResultSetRowMapFactory;
 import org.labkey.api.data.CachedResultSet;
 import org.labkey.api.data.CachedResultSets;
 import org.labkey.api.data.ResultSetMetaDataImpl;
+import org.labkey.api.data.RuntimeSQLException;
 import org.labkey.api.query.AliasManager;
 
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/util/ResultSetUtil.java
```diff
@@ -150,45 +151,53 @@ public static void logData(ResultSet rs, Logger log)
...
-                for (int i = 1; i <= columnCount; i++)
-                {
-                    sb.append(md.getColumnName(i)).append(" ");
-                }
+            sb.append('\n');
 
-                sb.append('\n');
+            for (int i = 1; i <= columnCount; i++)
+            {
+                sb.append(md.getColumnName(i)).append(" ");
```
labkey-jeckels: Consider tab as the separator.

api/src/org/labkey/api/util/ResultSetUtil.java
```diff
@@ -150,45 +151,53 @@ public static void logData(ResultSet rs, Logger log)
...
-                    }
+            sb.append('\n');
 
-                    sb.append('\n');
+            while (rs.next())
+            {
+                for (int i = 1; i <= columnCount; i++)
+                {
+                    Object value = rs.getObject(i);
+                    sb.append(null == value ? "-" : value.toString().trim()).append(" ");
```
labkey-jeckels: Consider tab as the separator.

labkey-jeckels (approved):

api/src/org/labkey/api/util/ResultSetUtil.java
labkey-adam: Done

api/src/org/labkey/api/util/ResultSetUtil.java
labkey-adam: Done

PR #5271, opened 2024-02-26
Update project paths for consolidated repositories

labkey-tchad:


labkey-jeckels (commented):
This seems unrelated to the migration. I'm not against the change, but what's it intended to improve?

labkey-tchad:
Accidental commit

PR #5272, opened 2024-02-26
Execute any reindexing operations in the background to not block startup

labkey-adam:
#### Rationale
[Issue 48960: Search: Consider not blocking server startup on search index deletion](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48960)

### DIFF
--- a/search/src/org/labkey/search/SearchModule.java
+++ b/search/src/org/labkey/search/SearchModule.java
@@ -39,9 +39,8 @@ import org.labkey.api.settings.AdminConsole;
 import org.labkey.api.settings.StandardStartupPropertyHandler;
 import org.labkey.api.settings.StartupPropertyEntry;
 import org.labkey.api.usageMetrics.UsageMetricsService;
-import org.labkey.api.util.ContextListener;
+import org.labkey.api.util.JobRunner;
 import org.labkey.api.util.PageFlowUtil;
-import org.labkey.api.util.StartupListener;
 import org.labkey.api.util.logging.LogHelper;
 import org.labkey.api.view.ActionURL;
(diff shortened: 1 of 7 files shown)


search/src/org/labkey/search/SearchModule.java
```diff
@@ -195,25 +193,27 @@ public void handle(Map<SearchStartupProperties, StartupPropertyEntry> properties
...
         SearchService ss = SearchService.get();
 
-        if (null != (ss))
+        if (null != ss)
         {
-            ss.start();
-            DavCrawler.getInstance().start();
+            // Execute any reindexing operations in the background to not block startup, Issue #48960
+            JobRunner.getDefault().execute(() -> {
+                _searchIndexStartupHandler.reindexIfNeeded(ss);
```
labkey-jeckels, line 207: What happens if someone does a search or other action that touches the index before this background process runs?

labkey-adam: If you manage to kick off a search before indexes are cleared (difficult to do after your exp change), the search page will return with this message: `Error: Search is disabled because the search index is misconfigured. Contact the system administrator of this server.` It may be easy to detect startup vs. misconfiguration and provide a slightly better message, but the current behavior seems reasonable.

labkey-adam: After the last commit, the message is now `Error: Search is unavailable because the search service is in the process of starting up.` Actual misconfiguration message is now `Error: Search is unavailable because the search index is misconfigured. Contact the system administrator of this server.`

labkey-jeckels (approved):

PR #5282, opened 2024-02-28
remove experimental feature USE_MATERIALIZED_SAMPLETYPE

labkey-matthewb:
use materialized sampletypes by default

#### Related Pull Requests
* 

#### Changes
*


labkey-susanh (commented):
We had talked about possibly wanting a way to clear out the materialized table if we got into a state where the code didn't do that as it should have (possibly through the clear caches page). Do we want that or is it already available?

labkey-matthewb:
> We had talked about possibly wanting a way to clear out the materialized table if we got into a state where the code didn't do that as it should have (possibly through the clear caches page). Do we want that or is it already available?

I'll check.  I forget if MaterializedQueryHelper is automatically wired up to clear cache events.

labkey-matthewb:
> > We had talked about possibly wanting a way to clear out the materialized table if we got into a state where the code didn't do that as it should have (possibly through the clear caches page). Do we want that or is it already available?
> 
> I'll check. I forget if MaterializedQueryHelper is automatically wired up to clear cache events.

This uses a regular BlockingCache (ExpMaterialTableImpl#1070).

labkey-susanh (approved):

PR #5257, opened 2024-02-21
Refactor logging configuration for embedded Tomcat

labkey-adam:
#### Rationale
We want to reconcile logging between the Spring Boot / embedded Tomcat startup process and the webapp

#### Related Pull Requests
* https://github.com/LabKey/server/pull/735

PR #5283, opened 2024-02-29
Fix package to match log4j2.xml packages attribute

labkey-adam:
#### Rationale
@labkey-willm flagged a log4j2 error on Tomcat standalone which pointed out that this class is in the wrong package for that deployment scenario

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5257

### DIFF
--- a/api/src/org/labkey/api/util/LabKeyDeleteAction.java
+++ b/api/src/org/labkey/api/util/LabKeyDeleteAction.java
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.labkey.api.util.logging;
+package org.labkey.api.util;
 
 import org.apache.logging.log4j.core.Core;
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
I agree the path and declared package should match, but I'd prefer to use the `logging` subpackage or get rid of it completely. I guess my attempt at moving it in the original FB didn't go well.

labkey-adam:
> I agree the path and declared package should match, but I'd prefer to use the `logging` subpackage or get rid of it completely. I guess my attempt at moving it in the original FB didn't go well.

Since log4j2.xml has now been updated in the docker PR, this class and `SessionAppender` will disappear once we're embedded-only, and branch time is quickly approaching, I'm going to leave it like this for now. We can clean up the logging package in a subsequent PR if we want to.

PR #5295, opened 2024-03-04
Improve upgrade workflow for embedded Tomcat

### LINKED ISSUE #725: [LabKey CI] Improve upgrade workflow for embedded Tomcat
[[Issue 49706](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49706)]

LabKey embedded currently has very simple logic for when to extract modules into place. If it finds the 'labkeywebapp' directory or if the 'context.webAppLocation' property is defined, it assumes that everything is already in place.
This works fine for our current primarily scenarios, test deployments and containerized deployments, because they are always setting up a fresh installation. If we want to encourage folks to switch existing production servers to embedded, the embedded server should be able to update modules in an existing deployment.

### Tasks
- [x] [Spec](https://docs.google.com/document/d/1aGvAw7QZGx38ql4Dja4H1SAIqyQZ2HB6W4CM7LRiQz8/edit) - 2
- [x] Spec review - 1 @labkey-jony
- [x] Create `application.properties` template for distributions
- [x] Remove version from 'labkeyServer.jar' included in distributions
- [x] Remove 'server' subdirectory creation
  - Update Dockerfile repo
  - Update LabKeyServer
- [x] Move VERSION file into 'labkeywebapp' directory
- [x] Compare VERSION files during startup
- [x] Add 'application.properties' to embedded distributions
- [x] Update default full-text-search index location
- [x] Backup existing deployment before extracting new distribution
- [x] Add example service scripts to embedded distributions
- [x] Code review 
  - https://github.com/LabKey/Dockerfile/pull/62
  - https://github.com/LabKey/gradlePlugin/pull/200
  - https://github.com/LabKey/testAutomation/pull/1845
  - https://github.com/LabKey/platform/pull/5295
  - https://github.com/LabKey/server/pull/752
- [x] Document upgrade process
- [x] Merge

labkey-tchad:
#### Rationale
Making migration from standalone to embedded tomcat a little smoother

#### Related Pull Requests
* https://github.com/LabKey/server/pull/752

#### Changes
* Update default full-text search index location
* Make heap and thread dump trigger files disappear when server shuts down

### DIFF
--- a/api/src/org/labkey/api/util/DebugInfoDumper.java
+++ b/api/src/org/labkey/api/util/DebugInfoDumper.java
@@ -87,6 +87,7 @@ public class DebugInfoDumper
             {
                 LOG.error("Failed to create file " + _threadDumpFile.getAbsolutePath(), e);
             }
+            _threadDumpFile.deleteOnExit();
         }
 
         // Ensure there's a heap dump request file
@@ -104,6 +105,7 @@ public class DebugInfoDumper
             {
                 LOG.error("Failed to create file " + _heapDumpFile.getAbsolutePath(), e);
             }
+            _heapDumpFile.deleteOnExit();
         }
 
         _threadDumpLastModified = _threadDumpFile.lastModified();
--- a/search/src/org/labkey/search/model/SearchPropertyManager.java
+++ b/search/src/org/labkey/search/model/SearchPropertyManager.java
@@ -18,6 +18,7 @@ package org.labkey.search.model;
 import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.data.PropertyManager;
+import org.labkey.api.files.FileContentService;
 import org.labkey.api.search.SearchService;
 import org.labkey.api.security.User;
 import org.labkey.api.settings.AppProps;
@@ -63,7 +64,17 @@ public class SearchPropertyManager
         String path = getProperty(INDEX_PATH);
 
         // Use path if set, otherwise fall back to temp directory.
-        return null != path ? path : new File(FileUtil.getTempDirectory(), "labkey_full_text_index").getPath();
+        if (path != null)
+        {
+            return path;
+        }
+        else
+        {
+            File indexParent = FileContentService.get() != null
+                    ? FileContentService.get().getSiteDefaultRoot()
+                    : FileUtil.getTempDirectory();
+            return new File(indexParent, ".labkey_full_text_index").getPath();
+        }
     }
 
     public static void setIndexPath(User user, String path)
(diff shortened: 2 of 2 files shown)


api/src/org/labkey/api/util/DebugInfoDumper.java
```diff
@@ -87,6 +87,7 @@ public DebugInfoDumper(File modulesDir)
             {
                 LOG.error("Failed to create file " + _threadDumpFile.getAbsolutePath(), e);
             }
+            _threadDumpFile.deleteOnExit();
```
labkey-jeckels, line 90: I don't see any harm in deleting these files but I'm not sure what benefit it brings either. We don't need to ask admins to delete these files either.

search/src/org/labkey/search/model/SearchPropertyManager.java
```diff
@@ -63,7 +64,17 @@ public static String getUnsubstitutedIndexDirectory()
...
+        if (path != null)
+        {
+            return path;
+        }
+        else
+        {
+            File indexParent = FileContentService.get() != null
+                    ? FileContentService.get().getSiteDefaultRoot()
+                    : FileUtil.getTempDirectory();
+            return new File(indexParent, ".labkey_full_text_index").getPath();
```
labkey-jeckels: @labkey-stuartm @labkey-jony any preferences on how to name this file? With or without the `.` prefix?

labkey-jeckels (approved):

search/src/org/labkey/search/model/SearchPropertyManager.java
labkey-tchad: I mostly added that so that it wouldn't show up in the `webdav` browser. (I _think_ it excludes dot files)

labkey-jeckels: Yes, I think you're right - those are hidden in the webdav UI.

However, I hope these files won't be exposed anyway. They're not in an `@files` subdirectory, are they?

labkey-tchad: You're correct, they are not exposed. I was just confused by seeing the `@files` directories in the webdav UI and assumed it was directly showing the file system. Further investigation does make me realize that there is a possibility of colliding with a project folder.
In order to avoid that, I see a few obvious solutions.
1. Use a folder name that is already reserved (e.g. `@labkey_full_text_index`)
2. Add `.labkey_full_text_index` to the list of reserved project names (consider removing the `.`)
3. Choose an entirely different location, maybe a sibling of `files`

labkey-jeckels: I vote for `@labkey_full_text_index`. On some deployments, the file root is a mount point on the file system. I think by default having the full text search land on the same file system is a good default.

labkey-tchad: That was my thought as well. Thank you

PR #5312, opened 2024-03-08
Separate the execution plan from the SQL a bit more

labkey-jeckels:
#### Rationale
We often show the SQL and execution plan on back-to-back lines in the Admin Console's Queries page. They're different things, so it's nice to have a little separation.

#### Changes
* Two beautiful newlines


labkey-adam (approved):
Seems risky, but if you insist...

PR #5313, opened 2024-03-08
Add nonce and onMouseOver() option

labkey-adam:
#### Rationale
`LinkBuilder` now supports `onmouseover` events

### DIFF
--- a/api/src/org/labkey/api/util/ExceptionUtil.java
+++ b/api/src/org/labkey/api/util/ExceptionUtil.java
@@ -1114,7 +1114,7 @@ public class ExceptionUtil
             if (response.isCommitted())
             {
                 PrintWriter out = response.getWriter();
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/util/ExceptionUtil.java
```diff
@@ -1114,7 +1114,7 @@ private static void doErrorRedirect(HttpServletResponse response, String url, in
             if (response.isCommitted())
             {
                 PrintWriter out = response.getWriter();
-                out.println("\"'>--></script><script type=\"text/javascript\">");
+                out.println("\"'>--></script><script type=\"text/javascript\" nonce=\"" + HttpView.currentPageConfig().getScriptNonce() + "\">");
```
labkey-jeckels: Hopefully we keep the nonce values simple, but we should HTML-encode anyway.

labkey-jeckels (approved):

PR #5314, opened 2024-03-09
Junit test that detects CSP violations in HTML views

labkey-adam:
#### Rationale
We can detect and flag module HTML views that contain nonce-less script tags and inline handlers. For now, just log. Once we've sanitized all views we'll adjust this test to throw.


labkey-jeckels (approved):
Nice!

PR #5321, opened 2024-03-12
Introduce "Primary" Plate Sets

labkey-nicka:
#### Rationale
This introduces the concept of "primary" and "assay" Plate Sets as a type of Plate Set that can be utilized in plate operations. 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1447
- https://github.com/LabKey/labkey-ui-premium/pull/359
- https://github.com/LabKey/limsModules/pull/50
- https://github.com/LabKey/platform/pull/5321

#### Changes
- Introduce `plate.PlateSetEdge` table for persisting relationships between plate sets
- Add `Type`, `PrimaryPlateSetId` and `RootPlateSetId` columns to `plate.PlateSet`
- Default all pre-existing Plate Sets to type of "assay"
- Serialize `type` on `PlateSet`
- Introduce `PlateController.LineageAction` to support requesting all edges for a given seed
- Handle clean-up/deletion of plate sets


assay/resources/schemas/dbscripts/postgresql/assay-24.003-24.004.sql
```diff
@@ -0,0 +1,27 @@
+ALTER TABLE assay.PlateSet ADD COLUMN Type VARCHAR(64);
+ALTER TABLE assay.PlateSet ADD COLUMN RootPlateSetId INT;
+ALTER TABLE assay.PlateSet ADD COLUMN PrimaryPlateSetId INT;
+ALTER TABLE assay.PlateSet ADD CONSTRAINT FK_PlateSet_RootPlateSetId FOREIGN KEY (RootPlateSetId) REFERENCES assay.PlateSet(RowId);
+ALTER TABLE assay.PlateSet ADD CONSTRAINT FK_PlateSet_PrimaryPlateSetId FOREIGN KEY (PrimaryPlateSetId) REFERENCES assay.PlateSet(RowId);
+
+-- Update all pre-existing plate sets to type "assay"
+UPDATE assay.PlateSet SET type = 'assay';
+
+ALTER TABLE assay.PlateSet ALTER COLUMN Type SET NOT NULL;
```
labkey-klum, line 10: Did you consider just creating the type column as `NOT NULL` with `assay` as the default?

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -1950,6 +1977,19 @@ public PlateSetImpl createPlateSet(Container container, User user, @NotNull Plat
...
+        if (parentPlateSetId != null)
+        {
+            parentPlateSet = (PlateSetImpl) getPlateSet(container, parentPlateSetId);
+            if (parentPlateSet == null)
+                throw new ValidationException(String.format("Failed to create plate set. Parent plate set with rowId (%d) is not available.", parentPlateSetId));
+            if (parentPlateSet.getRootPlateSetId() == null)
+                throw new ValidationException(String.format("Failed to create plate set. Parent plate set with rowId (%d) does not have a root plate set specified.", parentPlateSetId));
+        }
+
+        if (plateSet.getType() == null)
```
labkey-klum, line 2003: You could also initialize `PlateSetImpl._type` to `assay`.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -1123,13 +1110,21 @@ public void beforePlateDelete(Container container, Integer plateId)
...
         Table.delete(schema.getTableInfoWellGroup(), plateIdFilter);
     }
 
+    public void beforePlateSetDelete(Container container, User user, Integer rowId)
+    {
+        // We'll want to consider having a way of stitching up the lineage when deleting plate sets
+        SQLFragment sql = new SQLFragment("DELETE FROM ").append(AssayDbSchema.getInstance().getTableInfoPlateSetEdge(), "PSE")
+                .append("WHERE fromPlateSetId = ? OR toPlateSetId = ? OR RootPlateSetId = ?").addAll(rowId, rowId, rowId);
+        new SqlExecutor(AssayDbSchema.getInstance().getSchema()).execute(sql);
+    }
```
labkey-klum, line 1148: Should we also be cleaning up any references to this `PlateSet`? Similar to what you are doing in the container delete method.

labkey-klum (approved):

assay/resources/schemas/dbscripts/postgresql/assay-24.003-24.004.sql
labkey-nicka: As discussed I looked at this but our system doesn't seem to be able to distinguish well between required (due to `NOT NULL`) and `DEFAULT 'assay'` in the database. System ends up verifying the field's value is non-null without recognizing the fact the database will furnish a default value. Going to leave as-is.

assay/src/org/labkey/assay/plate/PlateManager.java
labkey-nicka: I think I'd rather be explicit here and not have the model default to anything. I wouldn't want models being written that end up changing their type because the default differs.

assay/src/org/labkey/assay/plate/PlateManager.java
labkey-nicka: We certainly should. c04c3e2.

PR #5323, opened 2024-03-13
QC plot changes for fixed deviation from mean outlier option

ankurjuneja:
#### Rationale
This PR adds necessary support changes in plots api for fixed deviation from mean outlier option.

#### Related Pull Requests
- 

#### Changes
- add lowerBound and upperBound layers for new outlier option


core/webapp/vis/src/plot.js
```diff
@@ -1731,7 +1731,7 @@ boxPlot.render();
         }
 
         config.properties.boundType = config.properties.boundType || 'stddev';
-        if (config.properties.boundType !== 'stddev' && config.properties.boundType !== 'absolute') {
+        if (config.properties.boundType !== 'stddev' && config.properties.boundType !== 'absolute' && config.properties.boundType !== 'meanDev') {
```
labkey-jeckels: This might be a good time to introduce constants instead of having the string hard-coded repeatedly.

labkey-jeckels (approved):

PR #5326, opened 2024-03-13
Canonical list of themes

labkey-adam:
#### Rationale
Fix misspelling and add valid values to theme startup property docs. While we're at it, consolidate the list of supported themes into a proper enum.


labkey-susanh (approved):
Nice. I like enums.

PR #5333, opened 2024-03-14
Plateset perf & miscellaneous fixes

labkey-klum:
#### Rationale
This is a slight optimization for the creation of plates and platesets. Instead of inserting new plate well rows one at a time, we now insert them in bulk and this has helped scenarios where we are creating a large number of plates (and wells) as part of the LKB plateset workflow.

Additionally, this fixes a transform script regression caused by the transition to provisioned tables for plate metadata from vocabulary domains. We need to source the values for the well custom fields appropriately.

Also, re-enables the unit tests which validate custom well metadata.


assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -859,11 +861,6 @@ private int savePlateImpl(Container container, User user, PlateImpl plate) throw
 
             if (updateExisting)
             {
-                // special case if the plate name changes, we want to remove the cache key with the old name
-                Plate oldPlate = getPlate(container, plateId);
-                if (!oldPlate.getName().equals(plate.getName()))
-                    clearCache(container, oldPlate);
-
```
labkey-klum, line 866: No longer necessary since we don't support plates by name in the cache.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -923,66 +920,59 @@ private int savePlateImpl(Container container, User user, PlateImpl plate) throw
...
-                    if (updateExisting)
-                    {
-                        position = existingPositionMap.get(Pair.of(row, col));
-                        assert position != null;
-                        assert position.getRowId() != null && position.getRowId() > 0;
-                        assert position.getLsid() != null;
-
-                        Lsid lsid = Lsid.parse(position.getLsid());
-                        String objectId = lsid.getObjectId();
-                        assert objectId.endsWith("-well-" + row + "-" + col);
```
labkey-klum, line 953: These extra checks for existing wells don't add much value.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -1084,8 +1076,6 @@ public void beforePlateDelete(Container container, Integer plateId)
         lsids.add(plate.getLSID());
         for (WellGroup wellgroup : plate.getWellGroups())
             lsids.add(wellgroup.getLSID());
-        for (Well well : plate.getWells())
-            lsids.add(well.getLsid());
```
labkey-klum, line 1088: Was only necessary when well metadata was using vocabulary.

labkey-nicka (approved):

PR #5337, opened 2024-03-18
Miscellaneous issues

labkey-klum:
#### Related issues
- https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49335
- https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=49371

#### Rationale
- Return `HTMLString` objects for the QC state UI interfaces
- Do a better job of reporting errors when using a visit map to translate strings to visits.

### DIFF
--- a/study/src/org/labkey/study/controllers/StudyController.java
+++ b/study/src/org/labkey/study/controllers/StudyController.java
@@ -16,6 +16,9 @@
 
 package org.labkey.study.controllers;
 
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpSession;
 import org.apache.commons.beanutils.ConversionException;
 import org.apache.commons.collections4.FactoryUtils;
 import org.apache.commons.collections4.MapUtils;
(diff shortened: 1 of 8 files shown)


study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -3486,7 +3532,7 @@ public String getQcStateDefaultsPanel(Container container, DataStateHandler qcSt
             panelHtml.append("      </tr>");
             panelHtml.append("  </table>");
 
-            return panelHtml.toString();
```
labkey-matthewb, line 3489: really need to convert getQcStateHtml() as well.  It seems to append string without using filter().  HtmlStringBuilder makes it harder to miss filtering strings

labkey-matthewb (changes requested):

study/src/org/labkey/study/controllers/StudyController.java
labkey-klum: done

labkey-matthewb (approved):

PR #5339, opened 2024-03-18
Check access for dataset change notifications

labkey-klum:
[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49635)

#### Rationale
Need to do an additional access check for datasets in case study level security has been configured for the study. Currently the notification provider is only checking for container level access.


labkey-klum:
@cnathe I don't know if you worked on or are familiar with this feature, but this just adds an extra dataset check for notifications.

cnathe (approved):

PR #5343, opened 2024-03-19
Update cross-folder moves to allow moving a subset you have permissions for

labkey-susanh:
#### Rationale
In the [related PR](#5325), we added the ability to move entities from a selection that crosses containers with the server side returning an error if you included one or more entities from containers you don't have permission to move from. This PR and its relatives update the behavior so we warn about any entities you don't have permissions for and send across only the entities you do have permission for for the move operation.

#### Related Pull Requests
- #5325 
- https://github.com/LabKey/limsModules/pull/66
- https://github.com/LabKey/labkey-ui-components/pull/1452
- https://github.com/LabKey/labkey-ui-premium/pull/363

#### Changes
- Add method for getting container rows from a selection.


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -7632,6 +7632,37 @@ public void setSeqType(String seqType)
...
+    @Marshal(Marshaller.Jackson)
+    @RequiresPermission(ReadPermission.class)
+    public static class GetSelectedContainersAction extends ReadOnlyApiAction<CrossFolderSelectionForm>
+    {
+        @Override
+        public void validateForm(CrossFolderSelectionForm form, Errors errors)
+        {
+            if (form.getDataRegionSelectionKey() == null && form.getRowIds() == null)
+                errors.reject(ERROR_REQUIRED, "You must provide either a set of rowIds or a dataRegionSelectionKey.");
+            if (!"samples".equalsIgnoreCase(form.getDataType()) && !"exp.data".equalsIgnoreCase(form.getDataType())&& !"assay".equalsIgnoreCase(form.getDataType()))
```
cnathe: minor: could the `ExperimentServiceImpl.getTableInfo(form.getDataType())` be used here instead of hardcoding the schema lists in multiple places?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -7641,7 +7672,7 @@ public void validateForm(CrossFolderSelectionForm form, Errors errors)
         {
             if (form.getDataRegionSelectionKey() == null && form.getRowIds() == null)
                 errors.reject(ERROR_REQUIRED, "You must provide either a set of rowIds or a dataRegionSelectionKey.");
-            if (!"sample".equalsIgnoreCase(form.getDataType()) && !"data".equalsIgnoreCase(form.getDataType())&& !"assayrun".equalsIgnoreCase(form.getDataType()))
+            if (!"samples".equalsIgnoreCase(form.getDataType()) && !"exp.data".equalsIgnoreCase(form.getDataType())&& !"assay".equalsIgnoreCase(form.getDataType()))
```
cnathe, line 7661: same question here about the getTableInfo helper

cnathe (approved):

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
labkey-susanh: Getting a TableInfo is relatively expensive, so I'd rather not do that. I could make a helper method for use in both these actions, but, as discussed, may try to get rid of one of the actions soon anyway, so may not be worthwhile.

PR #5344, opened 2024-03-19
Prevent login form from posting

labkey-adam:
#### Rationale
Related pull request removed `action="javascript:void(0)"` from all login page `<form>` elements. That seems to have broken the use of `Enter` on the login form. I believe after that change, `authenticateUser()` was issuing an AJAX post to the login action AND the form was posting to the action, which interfered with the AJAX post and generated a CSRF error. Returning `false` blocks the form post.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5334


labkey-klum (approved):
yup, returning false in the click handler prevents the browser from performing the default action (form submit).

PR #5346, opened 2024-03-20
Turn off JDBC caching when requesting an uncached result set

labkey-adam:
#### Rationale
We're still seeing out-of-memory errors while exporting audit log rows. We previously instructed `QueryService` not to cache, but that didn't turn off PostgreSQL JDBC caching.

A simple change to `QueryServiceImpl.select()` (`cache == false` implies turn off JDBC caching) exposed some bad logic in `SqlExecutingSelector`: in short, it would happily use a new connection from the pool even when a transaction was active. (Issue indexing, I don't know why you turned off caching for a single issue select, but thank you for blowing up.) Pushing all the JDBC caching logic into the PostgreSQL dialect allows us to check all the conditions in one place and get a new connection only when we're really sure about it. Also, we don't need so many dialect methods and classes worrying about this very PostgreSQL-specific concept.

#### Related Pull Requests
- https://github.com/LabKey/professional/pull/21

### DIFF
--- a/api/src/org/labkey/api/data/ConnectionWrapper.java
+++ b/api/src/org/labkey/api/data/ConnectionWrapper.java
@@ -16,7 +16,6 @@
 
 package org.labkey.api.data;
 
-import org.apache.commons.collections.map.MultiValueMap;
 import org.apache.commons.collections4.multimap.HashSetValuedHashMap;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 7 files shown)


labkey-jeckels (approved):

labkey-adam:
@labkey-jeckels @labkey-matthewb this was approved previously, but all the reworking is worth another review

labkey-adam:
PostgreSQL Dailies are all green as well: https://teamcity.labkey.org/buildConfiguration/LabKey_243Release_Community_DailyPostgres2/2960514?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true

labkey-jeckels (approved):

PR #5354, opened 2024-03-21
use SQLFragment in Table.update

labkey-matthewb:
#### Rationale
Convert Table.update() to use SQLFragment.  Avoids problem with SQLFragment validity check later.

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/data/Table.java
+++ b/api/src/org/labkey/api/data/Table.java
@@ -1083,7 +1083,7 @@ public class Table
         {
             // UNDONE: reselect
             SQLFragment updateSQL = new SQLFragment()
-                    .append("UPDATE " + table.getSelectName() + "\n\t")
+                    .append("UPDATE ").append(table.getSQLName()).append("\n\t")
                     .append("SET ").append(setSQL).append("\n\t")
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/Table.java
```diff
@@ -996,17 +994,16 @@ else if (pkVals instanceof Map)
...
         }
 
         for (ColumnInfo col : columnPK)
         {
             whereSQL.append(whereAND);
-            whereSQL.append(col.getSelectName());
+            whereSQL.appendIdentifier(col.getSelectName());
             whereSQL.append("=?");
-            parametersWhere.add(keys.get(col.getName()));
+            whereSQL.add(keys.get(col.getName()));
```
labkey-jeckels, line 1006: I guess we never support NULL values in a PK field (unchanged from original code)

labkey-jeckels (approved):

api/src/org/labkey/api/data/Table.java
labkey-matthewb: By definition PK != NULL.

PR #5355, opened 2024-03-21
Issue 48500: Fix display for multi-valued column with single amount that is empty

labkey-susanh:
#### Rationale
Issue [48500](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48500)

#### Changes
- update `AmountDataColumn` to handle a string value (in particular, an empty string value) well.

### DIFF
--- a/api/src/org/labkey/api/data/measurement/AmountDataColumn.java
+++ b/api/src/org/labkey/api/data/measurement/AmountDataColumn.java
@@ -58,7 +58,16 @@ public class AmountDataColumn extends DataColumn
 
     private Object convertToDisplayUnits(RenderContext ctx)
     {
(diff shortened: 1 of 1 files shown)


XingY (approved):

labkey-jeckels (commented):
Do we want this in a 24.3 maintenance release? I'm not sure of the urgency for the customer(s) who hit it

labkey-susanh:
> Do we want this in a 24.3 maintenance release? I'm not sure of the urgency for the customer(s) who hit it

It has not been requested for 24.3 because it comes up rarely (since the functionality in our apps is not officially supported) and currently has happened for customers who get monthly releases.  it's easy enough to back port if we find we need it there, but I'll leave it targeting develop for now.

PR #5357, opened 2024-03-22
 Issue 49884, 49901, 49906, 49823

XingY:
#### Rationale
Issue 49884: Time field values in an assay run field do not survive a round trip of a folder export-import
 Issue 49901: Time-only field stores incorrect time after daylight savings time if milliseconds and am/pm are used
 Issue 49906: Time only fields in datasets do not honor milliseconds
 Issue 49823: Time only fields are interpreted as text fields if a tsv,csv file is used to define field types in a domain

Selenium coverage for 49884, 49901, and 49906 are already in place:
AssayRenameExportImportTest.testProjectImport
ListDateAndTimeTest.testDateAndTimeFormat
DataClassTest.testDateAndTimeFields

#### Related Pull Requests
- 

#### Changes
- new SimpleTime marker class to use for parsing string fields that matches a set of time format when inferring domain fields
- fix time tz offset, should always use the offset for the specified date
- fix folder export for time fields


api/src/org/labkey/api/reader/DataLoader.java
```diff
@@ -49,11 +49,13 @@
 import org.labkey.api.query.ValidationException;
 import org.labkey.api.util.ExceptionUtil;
 import org.labkey.api.util.Filter;
+import org.labkey.api.util.SimpleTime;
 
 import java.io.Closeable;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.sql.Time;
```
labkey-susanh, line 58: unused?

api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -80,8 +80,8 @@ private DateUtil()
     private static final String ISO_DATE_SHORT_TIME_FORMAT_STRING = ISO_DATE_FORMAT_STRING + " " + ISO_SHORT_TIME_FORMAT_STRING;
     private static final String ISO_TIME_FORMAT_STRING = "HH:mm:ss";
     private static final String ISO_LONG_TIME_FORMAT_STRING = "HH:mm:ss.SSS";
-    private static final String[] SIMPLE_TIME_FORMATS_WITH_AMPM = {"hh:mm:ss a", "hh:mm a"};
-    private static final String[] SIMPLE_TIME_FORMATS_NO_AMPM = {"HH:mm:ss", "HH:mm"};
+    private static final String[] SIMPLE_TIME_FORMATS_WITH_AMPM = {"hh:mm:ss.SSS a", "hh:mm:ss a", "hh:mm a"};
```
labkey-susanh, line 83: Any need to support just "hh a" as well?

api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -80,8 +80,8 @@ private DateUtil()
...
+    private static final String[] SIMPLE_TIME_FORMATS_NO_AMPM = {"HH:mm:ss.SSS", "HH:mm:ss", "HH:mm"};
```
labkey-susanh, line 84: Similar question here.

api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -1056,8 +1069,9 @@ public static Time fromTimeString(@NotNull String s, Container container, boolea
             }
         }
 
-        int timezoneDiffSec = ZonedDateTime.now().getOffset().getTotalSeconds();
-        return new Time(DateUtil.parseTime(s, strict) - 1000 * timezoneDiffSec);
+        long time = DateUtil.parseTime(s, strict);
+        int timezoneDiffSec = TimeZone.getDefault().getOffset(time);
+        return new Time(time - 1000 * timezoneDiffSec);
```
labkey-susanh, line 1075: Is this converting to UTC? I'm not seeing why that's needed/wanted.

labkey-susanh (approved):

api/src/org/labkey/api/util/DateUtil.java
XingY: Not UTC, but the tz the server is at. The reason this is needed is because DateUtil.parseTime throws away the timezone part during parsing, so we need to add that back.

api/src/org/labkey/api/util/DateUtil.java
XingY: It's kinda risky. Those formats are used for inferring domain field types, it's debatable if we want those values to be evaluated to Time imo.

api/src/org/labkey/api/util/DateUtil.java
labkey-susanh: Ah, thanks for that info, That seems like it might be worth a comment.

api/src/org/labkey/api/util/DateUtil.java
labkey-susanh: I'm not sure what else it would be if not a time, but is likely very uncommon for time data to just be "3 am". Asking only because it's the natural progression. We can always add that in later if it's needed.

PR #5358, opened 2024-03-22
Improve server-side CSP violation detection

labkey-adam:
#### Rationale
Proactive server-side detection of CSP violations will be helpful

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5314

#### Changes
- Factor CSP violation detection within HTML Documents out of junit test
- Add detection of `javascript:` hrefs and actions
- Create wiki site validation provider that enumerates all wikis and reports rendering and CSP issues
- ModuleHtmlView junit test now fails if any CSP violation is detected

### DIFF
--- a/wiki/src/org/labkey/wiki/WikiValidationProvider.java
+++ b/wiki/src/org/labkey/wiki/WikiValidationProvider.java
@@ -16,11 +16,13 @@ import org.w3c.dom.Document;
 
 import java.util.Collection;
 import java.util.LinkedList;
+import java.util.Map;
 import java.util.Set;
 
 public class WikiValidationProvider implements SiteValidationProvider
 {
     private static final Logger LOG = LogHelper.getLogger(WikiValidationProvider.class, "Wiki rendering exceptions");
+
     @Override
     public String getName()
     {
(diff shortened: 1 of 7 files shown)


wiki/src/org/labkey/wiki/WikiValidationProvider.java
```diff
@@ -0,0 +1,85 @@
...
+                }
+            }
+        }
+
+        return list;
+    }
+
+    private void addResult(SiteValidationResultList list, Wiki wiki, String title, String message)
+    {
+        list.addWarn(title + " (" + wiki.getName() + "): " + message, wiki.getPageURL());
```
labkey-jeckels, line 83: Does the container also get included when this is surfaced to the user?

labkey-jeckels (approved):

wiki/src/org/labkey/wiki/WikiValidationProvider.java
labkey-adam: Yes, results (if any) are organized by folder:

image

It would be easy to add UI to validate a single folder (or subtree or choose the validators to run), but that's not currently an option.

PR #5360, opened 2024-03-22
Use simpler check for authenticated user

labkey-matthewb:
#### Rationale
AuthenticatedRequest.isGuestSession() returns false for robots causing us to track sessions that won't ever fire SessionListener.sessionDestroyed()

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/security/UserManager.java
+++ b/api/src/org/labkey/api/security/UserManager.java
@@ -100,6 +100,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
+import static org.labkey.api.security.SecurityManager.USER_ID_KEY;
 import static org.labkey.api.security.permissions.AbstractActionPermissionTest.APPLICATION_ADMIN_EMAIL;
 import static org.labkey.api.security.permissions.AbstractActionPermissionTest.SITE_ADMIN_EMAIL;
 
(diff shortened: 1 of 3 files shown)


labkey-matthewb:
Not right yet.  It seems we do want to track "real" guests?

labkey-matthewb:
I think AuthenticatedRequest.isGuestSession() isn't very meaningful here.  I think these changes are correct if the intention is that getActiveUserSessionCount() returns the count authenticated users sessions.  If that's not the case, then I'm confused about what the previous code was trying to do.

labkey-jeckels (approved):

labkey-jeckels (changes requested):
The new assert on line 509 is failing.

api/src/org/labkey/api/security/UserManager.java
```diff
@@ -506,14 +500,24 @@ public static int getUniqueUsersCount(Date since)
...
 
-    private static final Map<String, HttpSession> _activeSessions = Collections.synchronizedMap(new HashMap<>());
+    private static final Set<String> _activeSessions = Collections.synchronizedSet(new HashSet<>());
+
+    public static void ensureSessionTracked(HttpSession s)
+    {
+        if (s != null)
+        {
+            Integer userId = (Integer)s.getAttribute(USER_ID_KEY);
+            if (null != userId && 0 != userId)
```
labkey-jeckels:
```suggestion
            if (null != userId && getGuestUser().getUserId() != userId)
```

labkey-jeckels (approved):

PR #5361, opened 2024-03-24
Plates: Hit Selection Persistence/API

labkey-nicka:
#### Rationale
Introduces hit selection capabilities to plates. This includes persistence, query, and API support as well as a "Hit Selection" column on plate-based assay results.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/77

#### Changes
- Introduce initial interface for `AssayListener` which currently includes one "event" `beforeResultDelete`. Assay listeners can be registered via `AssayService.registerListener()`.
- Introduce `assay.Hit` table for persisting hits. This is exposed to LK query as well.
- Adds `assay-hit.api` for marking and unmarking hits based on either supplied assay result rowIds or by data region selection key.
- Adds a "Hit Selection" `ExprColumn` on plate-based assay results.

### DIFF
--- a/api/src/org/labkey/api/assay/AssayListener.java
+++ b/api/src/org/labkey/api/assay/AssayListener.java
@@ -1,0 +1,12 @@
+package org.labkey.api.assay;
+
+import org.labkey.api.data.Container;
+import org.labkey.api.exp.api.ExpRun;
+import org.labkey.api.security.User;
+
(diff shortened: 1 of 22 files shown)


labkey-alan (approved):
Looks good to me, but I think @labkey-klum should still take a look

labkey-klum (approved):

PR #5365, opened 2024-03-25
Issue 49785: Search Crawler fills error logs when it can't get DB conn

labkey-jeckels:
#### Rationale
It's OK to log errors if the DB is offline, but there's no need to scream continuously.

#### Changes
- When we hit a `ConfigurationException` (which includes the DB being unavailable but could represent other serious problems), wait 1-10 minutes before we attempt again

### DIFF
--- a/search/src/org/labkey/search/model/DavCrawler.java
+++ b/search/src/org/labkey/search/model/DavCrawler.java
@@ -30,6 +30,7 @@ import org.labkey.api.query.FieldKey;
 import org.labkey.api.resource.Resource;
 import org.labkey.api.search.SearchService;
 import org.labkey.api.security.User;
+import org.labkey.api.util.ConfigurationException;
 import org.labkey.api.util.ContextListener;
 import org.labkey.api.util.DateUtil;
 import org.labkey.api.util.ExceptionUtil;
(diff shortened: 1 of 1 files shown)


search/src/org/labkey/search/model/DavCrawler.java
```diff
@@ -547,10 +548,17 @@ public void run()
...
+                catch (InterruptedException ignored) {}
+                catch (ConfigurationException e)
                 {
-                    continue;
+                    // Issue 49785. Avoid spinning in a tight loop if the DB is unresponsive.
+                    consecutiveConfigExceptionCount++;
+                    _log.error("Unexpected error, delaying next attempt" + (consecutiveConfigExceptionCount > 1 ? (". " + consecutiveConfigExceptionCount + " consecutive ConfigurationExceptions") : ""), e);
+
+                    // Fallback strategy based on the number of consecutive failed attempts
+                    _wait(_crawlerEvent, TimeUnit.MINUTES.toMillis(Math.min(10, consecutiveConfigExceptionCount)));
```
labkey-matthewb, line 561: This won't wait the expected time if a container or folder is created.  Not a a practical problem.  Just a note that this is a little different that the common usage this.wait()

labkey-matthewb (approved):

PR #5366, opened 2024-03-25
Move unused pivot.jsp

labkey-matthewb:
#### Rationale
Moving unused pivot.jsp to devtools module (because it is a useful example).

#### Related Pull Requests
- 

#### Changes
-


labkey-adam (changes requested):
I know this is a (broken) example, but could you migrate the three inline event handlers and HTML encode the content? These are the last inline handlers in our repositories. And would be good to follow best practices in sample code.

labkey-matthewb:
> I know this is a (broken) example, but could you migrate the three inline event handlers and HTML encode the content? These are the last inline handlers in our repositories. And would be good to follow best practices in sample code.

Hmm, I must have overwritten my changes when I refactored/moved the file...

labkey-matthewb:
> > I know this is a (broken) example, but could you migrate the three inline event handlers and HTML encode the content? These are the last inline handlers in our repositories. And would be good to follow best practices in sample code.
> 
> Hmm, I must have overwritten my changes when I refactored/moved the file...

Found the changes hanging out over in the specimens module

labkey-adam (approved):
Code assumes that `member.uniqueName` and `member.name` are HTML-safe without any encoding, but that seems like a reasonable assumption.

PR #5368, opened 2024-03-25
Issue 49830: Improve file property handling

labkey-jeckels:
#### Rationale
Improve consistency of how we resolve file paths and exp.data references

#### Changes
- Consolidate validation, annotate parameters, and remove unneeded null checks

### DIFF
--- a/api/src/org/labkey/api/data/ExpDataFileConverter.java
+++ b/api/src/org/labkey/api/data/ExpDataFileConverter.java
@@ -30,11 +30,14 @@ import org.labkey.api.exp.api.ExperimentService;
 import org.labkey.api.files.FileContentService;
 import org.labkey.api.pipeline.PipeRoot;
 import org.labkey.api.pipeline.PipelineService;
+import org.labkey.api.query.QueryService;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/ExpDataFileConverter.java
```diff
@@ -210,89 +211,121 @@ public Object convert(Class type, Object value)
...
+        {
+            File result = data.getFile();
+            if (result != null)
+            {
+                return result;
+            }
+        }
         // Otherwise, treat it as a plain path
-        return FILE_CONVERTER.convert(type, value);
+        return FILE_CONVERTER.convert(File.class, value);
```
labkey-matthewb, line 329: I see that checks in convert() validate this returned file.  Still not sure why we would allow this in the API.  It encourages/allows implementation leaking.

labkey-matthewb (approved):
This is a great improvement.  I would still argue that we don't need to register this functionality with ConvertHelper at all.  I think it would be better to make make the implicit users of this functionality call this explicitly.

PR #5374, opened 2024-03-28
AbDisc: multi-measure assay plateset import

labkey-klum:
#### Rationale
This PR adds to the previous work to allow the import of plate sets into the standard assay:
- Tabular formats without plate identifiers can now be imported provided the following criteria is met:
  - All rows in the data must not have a plate identifier, either the column must be missing (plate, plateID, plate ID), or the field must be blank.
  - All plates in the plate set must be of the same type (same size).
  - The data cannot have more rows than the total number of wells available in the plate set.
  - The data must fit evenly within a plate boundary. For example, if you have a plate set with four 96 well plates, a data file with 192 rows would work (2 plates) but 193 would be illegal because the 3rd plate would have only 1 well populated.
  - Well location is required from the data file.
  - Within a plate each well location must be unique.
- Multiple measures from a grid input format can be imported by annotating the region above the grid header row. Single and multiple plates can support more than a single measure. As an open question, there is currently no validation to ensure that an annotated measure exists in the results domain. If it doesn't then the imported measure will be ignored through the normal import process. Additionally, if more than one annotation is added to a grid, the following prefix (or key) must exist:
  - The plate identifier must have the prefix `plate` followed by a colon and the identifier, upper or lower case is acceptable.
  - The measure name must have the prefix `measure` followed by a colon and the measure name, upper or lower case is acceptable.

image

#### Related Pull Requests
https://github.com/LabKey/commonAssays/pull/743


assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -533,24 +548,206 @@ public List<Map<String, Object>> parsePlateGrids(Container container, User user,
...
+            User user,
+            AssayProvider provider,
+            ExpProtocol protocol,
+            List<Plate> plates,
+            List<Map<String, Object>> data,
+            Pair<String, Boolean> plateIdAdded
+    ) throws ExperimentException
+    {
+        DomainProperty plateProp = provider.getResultsDomain(protocol).getPropertyByName(AssayResultDomainKind.PLATE_COLUMN_NAME);
+        Set<String> importAliases = new CaseInsensitiveHashSet(plateProp.getImportAliasSet());
```
cnathe, line 586: does this `plateProps. getImportAliasSet()` include the field name and label as well as the importAliases?

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -533,24 +548,206 @@ public List<Map<String, Object>> parsePlateGrids(Container container, User user,
...
+                throw new ExperimentException(String.format(ERROR_MESSAGE, "the number of rows in the data (" + data.size() + ") exceeds the total number of wells available in the plate set (" + (plates.size() * plateSize) + ")"));
+
+            // attempt to add the plate identifier into the data rows in the order that they appear in the plate set
+            List<Map<String, Object>> newData = new ArrayList<>();
+            int rowCount = 0;
+            int curPlate = 0;
+            Set<Position> positions = new HashSet<>();
+            for (Map<String, Object> row : data)
+            {
+                // well location field is required, return it will fail downstream
```
cnathe:
```suggestion
                // well location field is required, return if not provided or it will fail downstream
```

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -533,24 +548,206 @@ public List<Map<String, Object>> parsePlateGrids(Container container, User user,
...
+            List<Plate> plates = PlateManager.get().getPlatesForPlateSet(plateSet);
+            List<String> annotations = getAnnotations();
+
+            // single annotation can only be a plate identifier
+            if (annotations.size() == 1)
+                _plate = getPlateForId(annotations.get(0), plates);
+            else
+            {
+                // multiple annotation must have an annotation prefix
+                for (String annotation : getAnnotations())
```
cnathe:
```suggestion
                for (String annotation : annotations)
```

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -533,24 +548,206 @@ public List<Map<String, Object>> parsePlateGrids(Container container, User user,
...
+                }
+            }
+        }
+
+        private @Nullable Plate getPlateForId(String annotation, List<Plate> platesetPlates)
+        {
+            return platesetPlates.stream().filter(p -> p.isIdentifierMatch(annotation)).findFirst().orElse(null);
+        }
+
+        private @Nullable String getPrefixedValue(String annotation, String prefix)
```
cnathe, line 691: I like the use of the "Plate:" and "Measure:" prefix for this initial supported format.

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -559,41 +756,73 @@ public List<Map<String, Object>> parsePlateGrids(Container container, User user,
...
+                    for (Map.Entry<Plate, Collection<PlateGridInfo>> entry : plateMaps.asMap().entrySet())
                     {
-                        Plate dataForPlate = PlateService.get().createPlate(matchingPlate, plateGrid, null);
-                        for (Well well : dataForPlate.getWells())
-                            dataRows.add(getDataRowFromWell(entry.getKey(), well, measureName));
+                        Set<String> measures = new HashSet<>();
+                        Map<Position, Map<String, Object>> measureDataRows = new LinkedHashMap<>();
+                        for (PlateGridInfo gridInfo : entry.getValue())
+                        {
+                            String measureName = gridInfo.getMeasureName() != null ? gridInfo.getMeasureName() : defaultMeasureName;
```
cnathe: This is within the `multipleMeasures` case so I would have expected that each grid would have to define their measure name here instead of making use of the `defaultMeasureName`.  So does this mean that your file could theoretically define measure names for all but one of the measures in the assay design?

cnathe (approved):

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
labkey-klum: it's just the import aliases, I think for plate there are two that are added to the domain field : `PlateID` and `Plate ID`.

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
labkey-klum: It does, but maybe that is too much of an edge case since it would be the first measure in the domain. Maybe for this initial case I'll make it more restrictive so that if you have multi-measures all grids need to be explicit with the measure name.

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
cnathe: yeah, that sounds right. I just wasn't sure if we also needed to support the field name (i.e. Plate).

PR #5375, opened 2024-03-29
Turn off PostgreSQL JDBC driver caching for assay data export

labkey-adam:
#### Rationale
Attempting to export very large assays can result in OutOfMemoryErrors due to PG JDBC driver caching. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=49913

Also, delete the temp file after copying it.

### DIFF
--- a/api/src/org/labkey/api/assay/TsvDataHandler.java
+++ b/api/src/org/labkey/api/assay/TsvDataHandler.java
@@ -18,13 +18,12 @@ package org.labkey.api.assay;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
+import org.apache.logging.log4j.Logger;
 import org.labkey.api.data.ColumnHeaderType;
 import org.labkey.api.data.ColumnInfo;
 import org.labkey.api.data.DisplayColumn;
-import org.labkey.api.data.Results;
 import org.labkey.api.data.SimpleFilter;
 import org.labkey.api.data.Sort;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/assay/TsvDataHandler.java
```diff
@@ -169,15 +161,21 @@ public void exportFile(ExpData data, Path dataFile, User user, OutputStream out)
...
+                            try (TSVGridWriter writer = new TSVGridWriter(() -> ts.getResults(false), displayColumns))
                             {
                                 writer.setColumnHeaderType(ColumnHeaderType.FieldKey);
-                                writer.write(tempFile);
+                                rowCount = writer.write(tempFile);
                             }
 
-                            FileUtils.copyFile(tempFile, out);
+                            if (rowCount > 0)
+                                FileUtils.copyFile(tempFile, out);
```
labkey-jeckels, line 173: Moving the file may be significantly faster if the source and target are on the same file system. Since the data is big enough to blow the heap in some cases, maybe it would be noticeable. But your call.

labkey-jeckels (approved):

api/src/org/labkey/api/assay/TsvDataHandler.java
labkey-adam: We're copying to an `OutputStream`, which I believe is a `ZipOutputStream` wrapping the response OutputStream in some cases. I think it would require refactoring these classes and callers to use move in the appropriate cases.

PR #5381, opened 2024-04-01
Issue 50015: Upgrade banner causing confusion for LabKey Cloud clients

labkey-jeckels:
#### Rationale
LabKey Cloud customers who are app admins sometimes see the banner message advising them to upgrade to the latest major release, ahead of their scheduled upgrade process. This causes confusion with the separate upgrade communication that's already happening.

#### Changes
- Only show the upgrade banner to Site Admins, who are the ones with full access to the system and in a position to do an upgrade


labkey-adam (approved):
Fine the way it is. But seems unfortunate that context is `@Nullable`... I guess that's for `OperationsMXBeanImpl`? Maybe that caller should pass in a mock context instead.

labkey-jeckels:
> Fine the way it is. But seems unfortunate that context is `@Nullable`... I guess that's for `OperationsMXBeanImpl`? Maybe that caller should pass in a mock context instead.

Indeed. Per the commitment in WarningProvider.addDynamicWarnings(), if we have a ViewContext it should have an HTTP request. I guess we could also fake up a mock request but that seems like a lot

PR #5383, opened 2024-04-01
AbDisc: Replating

RosalineP:
#### Rationale
See spec [here](https://docs.google.com/document/d/1Gzrl64LuXAXQNZJOjoFJ6MRmEmbUYBswG02IkXxNlww/edit#heading=h.6dlamxczq719).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5383
- https://github.com/LabKey/labkey-ui-premium/pull/377
- https://github.com/LabKey/limsModules/pull/100

#### Changes
- Add cases to CreatePlateSetAction
- Support replating in CretaePlateSetAction
- Pull processing server-side, add helper fns

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateManager.java
+++ b/assay/src/org/labkey/assay/plate/PlateManager.java
@@ -2499,7 +2499,7 @@ public class PlateManager implements PlateService, AssayListener, ExperimentList
 
                 Map<String, Object> wellData = Map.of(
                         "sampleId", sampleIdsSorted[sampleIdsCounter],
-                        "wellLocation", PlateManager.get().createPosition(c, rowIdx,colIdx + 1).getDescription()
+                        "wellLocation", PlateManager.get().createPosition(c, rowIdx, colIdx).getDescription()
                 );
                 wellSampleDataForPlate.add(wellData);
                 sampleIdsCounter++;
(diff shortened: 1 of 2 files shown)


assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -166,6 +166,8 @@ public class PlateManager implements PlateService, AssayListener, ExperimentList
     private static final String PLATE_SET_NAME_EXPRESSION = "PLS-${now:date('yyyyMMdd')}-${RowId}";
     private static final String PLATE_NAME_EXPRESSION = "${${PlateSet/PlateSetId}-:withCounter}";
 
+    private static final List<String> POSITION_PREFIXES = Arrays.stream("A,B,C,D,E,F,G,H,I,J,K,L,M,N,O,P,Q,R,S,T,U,V,W,X,Y,Z,AA,AB,AC,AD,AE,AF".split(",")).toList();
```
RosalineP: Copied from Karl's comment from previous PR before branch:
> Code already exists that knows how to translate between well location syntax and row/column format. See PositionImpl and PlateManager.createPosition. I would be tempted to have just a single place in the code where we handle this type of translation.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2487,6 +2487,99 @@ private void requireActiveTransaction()
...
 
+    private Pair<Integer, List<Map<String, Object>>> getWellSampleData(int[] sampleIdsSorted, Integer rowCount, Integer columnCount, int sampleIdsCounter, Container c)
+    {
+        List<Map<String, Object>> wellSampleDataForPlate = new ArrayList<>();
+        for (int rowIdx = 0; rowIdx < rowCount; rowIdx++)
+        {
+            for (int colIdx = 0; colIdx < columnCount; colIdx++)
+            {
+                if (sampleIdsCounter >= sampleIdsSorted.length)
+                    break;
```
labkey-klum: Once we hit our limit of available samples, don't we want to exit the function? The existing code just looks like it will break out of the inner loop.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2487,6 +2487,99 @@ private void requireActiveTransaction()
...
+
+            PlateType plateType;
+            if (plateTypesHash.get(plateTypeId) != null)
+            {
+                plateType = plateTypesHash.get(plateTypeId);
+            }
+            else
+            {
+                plateType = PlateManager.get().getPlateType(plateTypeId);
+                plateTypesHash.put(plateTypeId, plateType);
```
labkey-klum, line 2542: Suggestion: let's protect against a `null` plate type returned. We should probably throw in this case.

labkey-klum (approved):

PR #5384, opened 2024-04-01
Plate metrics

labkey-alan:
#### Rationale
This PR adds the PlateMetricsProvider which returns metrics for Plate Sets, Plates, and related assays when the `experimental-app-plate-support flag` is enabled.

#### Related Pull Requests
- N/A

#### Changes
- Add PlateMetricsProvider

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateMetricsProvider.java
+++ b/assay/src/org/labkey/assay/plate/PlateMetricsProvider.java
@@ -1,0 +1,194 @@
+package org.labkey.assay.plate;
+
+import org.labkey.api.assay.AssayProtocolSchema;
+import org.labkey.api.assay.AssayProvider;
+import org.labkey.api.assay.AssayService;
+import org.labkey.api.assay.plate.PlateSetType;
+import org.labkey.api.data.Container;
+import org.labkey.api.data.ContainerFilter;
+import org.labkey.api.data.ContainerManager;
+import org.labkey.api.data.DbSchema;
+import org.labkey.api.data.SQLFragment;
+import org.labkey.api.data.SqlSelector;
+import org.labkey.api.data.TableInfo;
+import org.labkey.api.exp.api.ExpProtocol;
+import org.labkey.api.exp.api.ExperimentService;
+import org.labkey.api.security.User;
+import org.labkey.api.usageMetrics.UsageMetricsProvider;
+import org.labkey.assay.TsvAssayProvider;
+import org.labkey.assay.query.AssayDbSchema;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class PlateMetricsProvider implements UsageMetricsProvider
+{
+    private SQLFragment plateSetPlatesSQL(TableInfo plateSetTable, TableInfo plateTable)
+    {
+        return new SQLFragment("SELECT ps.name, COUNT(p.rowid) FROM ")
+                .append(plateSetTable, "ps")
+                .append(" LEFT OUTER JOIN ").append(plateTable, "p")
+                .append(" ON ps.rowid = p.plateset")
+                .append(" GROUP BY ps.rowId");
+    }
+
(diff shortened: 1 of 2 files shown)


assay/src/org/labkey/assay/plate/PlateMetricsProvider.java
```diff
@@ -0,0 +1,196 @@
...
+            count += PlateManager.get().getPlateMetadataFields(c, User.getSearchUser()).size();
+        }
+
+        return count;
+    }
+
+    @Override
+    public Map<String, Object> getUsageMetrics()
+    {
+        if (!ExperimentalFeatureService.get().isFeatureEnabled("experimental-app-plate-support"))
```
labkey-nicka: Replace with a call to `AssayPlateMetadataService.isExperimentalAppPlateEnabled()`

assay/src/org/labkey/assay/plate/PlateMetricsProvider.java
```diff
@@ -0,0 +1,196 @@
...
+                .append(" ON p.platetype = pt.rowid")
+                .append(" WHERE pt.columns = ? AND pt.rows = ?")
+                .add(cols)
+                .add(rows);
+        return new SqlSelector(schema, sql).getObject(Long.class);
+    }
+
+    private List<Container> getBiologicsFolders()
+    {
+        return ContainerManager.getAllChildren(ContainerManager.getRoot())
```
labkey-nicka: I think for now we should constrain this to `ContainerManager.getProjects()` and then filter for "Biologics" projects. The main reason is that there are almost certainly subfolders out there that are of type "Biologics" that belong to a "Biologics" project. Iterating over these uniquely ends up counting some things more than once. I think the change could be as follows:

1. Use `ContainerManager.getProjects()` here instead of `ContainerManager.getAllChildren(ContainerManager.getRoot())` keeping the filtering.
2. Replace subsequent usages of `ContainerFilter.EVERYTHING` with `ContainerFilter.Type.AllInProject.create(protocol.getContainer(), User.getSearchUser())`

assay/src/org/labkey/assay/plate/PlateMetricsProvider.java
```diff
@@ -0,0 +1,196 @@
...
+                "96WellCount", plateTypeCount(schema.getSchema(), plateTable, plateTypeTable, 12, 8),
+                "384WellCount", plateTypeCount(schema.getSchema(), plateTable, plateTypeTable, 24, 16)
+        ));
+
+        TableInfo hitTable = schema.getTableInfoHit();
+        List<ExpProtocol> plateEnabledProtocols = getPlateEnabledAssayProtocols();
+        plateMetrics.put("assays", Map.of(
+                "hitCount", new SqlSelector(schema.getSchema(), new SQLFragment("SELECT COUNT(*) FROM ").append(hitTable, "h")).getObject(Long.class),
+                "assaysWithPlateMetadataEnabled", plateEnabledProtocols.size(),
+                "assayRunsRount", getPlateBasedAssayRunsCount(plateEnabledProtocols),
```
labkey-nicka: With or without the change recommended above I'm seeing two runs counted for each run I import. I wonder if somehow two records end up in the associated "runs" table.

labkey-nicka (approved):
Looks good. Just some feedback about how we coalesce the data.

assay/src/org/labkey/assay/plate/PlateMetricsProvider.java
labkey-alan: I can't reproduce this issue.

labkey-alan: lol, did you notice it was misspelled as `assayRunsRount`

labkey-alan: I also looked at the emitted SQL used by `getPlateBasedAssayRunsCount` and it seems to check out. I looked at the rows returned (instead of just `COUNT(*)` and they are definitely unique assay runs. I am not really sure what would cause what you're seeing.

labkey-nicka: It looks like this was only the case when we were counting the same container(s) twice. Counts look good now.

PR #5385, opened 2024-04-01
Add ViewUsageStatistics view

labkey-alan:
#### Rationale
This PR adds a new action to the AdminController, ViewUsageStatisticsAction, which renders a react component that fetches the current usage statistics and renders them. The react component features a text box that lets users type a JSON path to filter down the results, as well as a list of keys in the current view that users can click to further filter the results. You can also navigate to the view with the query arg `?jsonPath=` to prefill the jsonPath text box e.g. `admin-viewUsageStatistics.view?jsonPath=modules.Assay`

This view is most useful when adding or modifying metrics to a module, as it allows you to easily filter the usage statistics object down to the module you care about, and then with the click of a button reload the usage statistics.

Screenshot 2024-04-01 at 5 42 22 PM

#### Related Pull Requests
- N/A

#### Changes
- Add ViewUsageStatisticsAction
- Add UsageStatsViewer.tsx and related component infrastructure
- Add link to ViewUsageStatisticsAction in `cusomizeSite.jsp`

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -89,6 +89,7 @@ import org.labkey.api.module.FolderTypeManager;
 import org.labkey.api.module.IgnoresForbiddenProjectCheck;
 import org.labkey.api.module.Module;
 import org.labkey.api.module.ModuleContext;
+import org.labkey.api.module.ModuleHtmlView;
 import org.labkey.api.module.ModuleLoader;
 import org.labkey.api.module.SimpleModule;
 import org.labkey.api.moduleeditor.api.ModuleEditorService;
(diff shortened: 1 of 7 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -11175,6 +11176,22 @@ public void setHourDelta(Integer hourDelta)
...
+        @Override
+        public ModelAndView getView(Object o, BindException errors)
+        {
+            return ModuleHtmlView.get(ModuleLoader.getInstance().getModule("core"), ModuleHtmlView.getGeneratedViewPath("ViewUsageStatistics"));
+        }
+
+        @Override
+        public void addNavTrail(NavTree root)
+        {
+            // TODO: What would be appropriate here?
```
labkey-jeckels: See `CustomizeSiteAction.addNavTrail()`. I think you could do the same, using "Usage Stats" as the child title instead.

core/src/org/labkey/core/admin/customizeSite.jsp
```diff
@@ -222,7 +222,7 @@ Click the Save button at any time to accept the current settings and continue.</
 %>
             <tr>
                 <td style="padding: 5px 0 5px;" colspan="2">
-                            <%=button("View").id("testUsageReport").onClick("testUsageReport(false); return false;")%>
```
labkey-jeckels, line 225: Could simplify `testUsageReport` to take no parameters since it just has a single caller now - the download scenario.

labkey-jeckels (commented):
I'll let Nick review the TSX and SCSS.

labkey-jeckels:
Nice! Two small bugs to report:

1. String values offer the individual characters as indexed keys which is a little confusing and isn't very useful.
2. Whenever you drill into a terminal child node in the JSON the `Keys` listing in the header switches to say `No valid keys starting with`. Better to omit this completely or switch to a "You're on a leaf JSON node" message or something similar.

labkey-nicka (approved):

PR #5388, opened 2024-04-02
49954: Add PHI options for folder templates

labkey-klum:
[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49954)

#### Rationale
Allow users to select PHI levels when creating folders from a template. The PHI selection combo will render in the Options section when the template folder option (and folder) are selected during the new folder creation wizard.

#### Changes
- Create a shared component for PHI options that could be reused between the existing export and create folder forms.
- Moved over some other utilities that previously existed in the folder export code for registering help popup event listeners that were previously refactored for CSP inline event handlers.

### DIFF
--- a/core/webapp/AdminWizardForm.js
+++ b/core/webapp/AdminWizardForm.js
@@ -33,3 +33,113 @@ Ext4.define('LABKEY.ext4.AdminWizardForm', {
     }
 
 });
+
+Ext4.define('LABKEY.ext4.PHIStore', {
+    extend: 'Ext.data.Store',
+
+    constructor: function (config) {
+        Ext4.apply(this, config);
+
+        this.fields = ['label', 'value'];
+
+        if (config.maxAllowedLevel) {
+            let data = [];
+
+            switch (config.maxAllowedLevel) {
+                case 'Restricted':
+                    data.push({ value: 'Restricted', label: 'Restricted, Full and Limited PHI' });
+                case 'PHI':
+                    data.push({ value: 'PHI', label: 'Full and Limited PHI' });
+                case 'Limited':
+                    data.push({ value: 'Limited', label: 'Limited PHI' });
+                    break;
+            }
+            // want the PHI to be ordered from least to most restrictive
+            this.data = data.reverse();
+        }
+        this.callParent(config);
+    }
+});
+
+Ext4.define('LABKEY.ext4.PHIExport', {
+    extend: 'Ext.container.Container',
+    alias: 'widget.labkey_phi_option',
+
+    constructor: function (config) {
+        Ext4.applyIf(config, {
+            maxAllowedLevel: 'NotPHI',
+            fieldName: 'exportPhiLevel'
+        });
+
+        this.phiHelp = LABKEY.export.Util.helpPopup("Include PHI Columns", "Include all dataset and list columns, study properties, and specimen data that have been tagged with this PHI level or below.");
+        let isIncludePhiChecked = config.maxAllowedLevel !== 'NotPHI';
+        let phiStore = Ext4.create('LABKEY.ext4.PHIStore', {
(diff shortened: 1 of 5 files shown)


core/webapp/AdminWizardForm.js
```diff
@@ -32,4 +32,114 @@ Ext4.define('LABKEY.ext4.AdminWizardForm', {
...
+
+    constructor: function (config) {
+        Ext4.apply(this, config);
+
+        this.fields = ['label', 'value'];
+
+        if (config.maxAllowedLevel) {
+            let data = [];
+
+            switch (config.maxAllowedLevel) {
```
cnathe, line 48: Changing this to a switch seems like we'd now always only end up with one data row in the PHIStore. Am I missing something? Previously it had checks like `maxAllowedPhiLevel >= fullPhi` and `maxAllowedPhiLevel >= limitedPhi` etc. which allowed the store to get populated with more than one option based on the maxAllowedLevel.

core/webapp/AdminWizardForm.js
```diff
@@ -32,4 +32,114 @@ Ext4.define('LABKEY.ext4.AdminWizardForm', {
...
+
+    constructor: function (config) {
+        Ext4.applyIf(config, {
+            maxAllowedLevel: 'NotPHI',
+            fieldName: 'exportPhiLevel'
+        });
+
+        this.phiHelp = LABKEY.export.Util.helpPopup("Include PHI Columns", "Include all dataset and list columns, study properties, and specimen data that have been tagged with this PHI level or below.");
+        let isIncludePhiChecked = config.maxAllowedLevel !== 'NotPHI';
+        let phiStore = Ext4.create('LABKEY.ext4.PHIStore', {
```
cnathe: I believe these can both be `const` instead of `let`

core/webapp/AdminWizardForm.js
```diff
@@ -32,4 +32,114 @@ Ext4.define('LABKEY.ext4.AdminWizardForm', {
...
+
+        this.listeners = {
+            render : function(cmp){cmp.phiHelp.callback();}
+        };
+        this.callParent(config);
+    }
+});
+
+LABKEY.export = LABKEY.export || {};
+LABKEY.export.Util = new function () {
```
cnathe: in develop it looks like this code exists in api/webapp/clientapi/dom/Utils.js. I assume this was put here in AdminWizardForm.js because this is targeting 24.3? We should be able to remove this in develop and use that version instead, right?

cnathe (changes requested):

core/webapp/AdminWizardForm.js
labkey-klum: no, perhaps I'm being too tricky but note there are no break statements in each case statement. So depending on where it starts, it will fall through to the lower ones. This is why it is organized from the highest to the lowest (and then reversed below).

core/webapp/AdminWizardForm.js
labkey-klum: Oh interesting, I did not see that good catch. I could also merge Adam's change to `Utils.js` to my branch and use that version in 24.3

core/webapp/AdminWizardForm.js
cnathe: ahhh, yup I missed the fact that there are no break statements. I saw the reverse part being added but didn't catch on. Looks good.

core/webapp/AdminWizardForm.js
cnathe: that might make the merge forward easier

cnathe (approved):

core/webapp/AdminWizardForm.js
labkey-klum: done

PR #5389, opened 2024-04-03
Address various issues with export to script

labkey-adam:
#### Rationale
Bit of a rewrite due to incorrect quoting, improper escaping, missing features, IntelliJ warnings, etc. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50035

#### Changes
- Fix `PageFlowUtil.jsString()` to escape `/` only when preceeded by `<`
- Force every `ExportScriptModel` to define and use a `quote()` method
- Reduce duplication by introducing a generalized `getFilters()` method
- Remove a couple redundant `getColumns()` overrides
- Switch Python export to use Java double quote string literals instead of JavaScript
- Output column list for Python (was missing!)
- Add proper quoting to parameters for Python
- Correct quoting for Perl filter expression
- Address some IntelliJ warnings

### DIFF
--- a/api/src/org/labkey/api/query/ExportScriptModel.java
+++ b/api/src/org/labkey/api/query/ExportScriptModel.java
@@ -15,6 +15,7 @@
  */
 package org.labkey.api.query;
 
+import jakarta.servlet.http.HttpServletResponse;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
@@ -35,7 +36,6 @@ import org.labkey.api.view.WebPartView;
 import org.labkey.api.view.template.PageConfig;
 import org.springframework.web.servlet.ModelAndView;
 
-import jakarta.servlet.http.HttpServletResponse;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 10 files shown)


labkey-alan (approved):
The Python Changes look correct to me, and I did some light manual testing and found no issues. Let me know if there are specific use cases you'd like me to test.

labkey-adam:
> The Python Changes look correct to me, and I did some light manual testing and found no issues. Let me know if there are specific use cases you'd like me to test.

Thanks, Alan! You probably noticed and covered it, but I added the column list (missing for some reason), so that's brand new and untested. Using customize grid to pull columns from a different tables (e.g., joining two datasets), sorting & filtering on the joined in columns, and then testing export to script would be a good smoke test.

labkey-alan:
I spent some additional time this morning verifying the script output for the scenarios outlined above, and everything appears to be working as intended.

labkey-adam:
> I spent some additional time this morning verifying the script output for the scenarios outlined above, and everything appears to be working as intended.

Awesome, thanks, Alan!

labkey-adam:
Note to no one in particular: Dailies are unhappy with these changes because `AbstractExportTest` expects the syntactically incorrect scripts that we used to export. https://teamcity.labkey.org/buildConfiguration/LabKey_243Release_Community_DailyPostgres2/2970271?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildTestsSection=true&expandBuildChangesSection=true

PR #5390, opened 2024-04-03
Issue 50030: sample update data iterator

XingY:
#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5390
- https://github.com/LabKey/limsModules/pull/115

#### Changes
-

### DIFF
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -2972,6 +2972,7 @@ public class ExpDataIterators
                 "ModifiedBy",
                 "Created",
                 "_rowNumber",
+                RowId.name(),
                 AliquotedFromLSID.name(),
                 ALIQUOTED_FROM_INPUT,
                 ROOT_RECOMPUTE_ROWID_COL,
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -136,6 +136,7 @@ import java.util.stream.Collectors;
 
 import static org.apache.commons.lang3.StringUtils.equalsIgnoreCase;
 import static org.labkey.api.data.CompareType.IN;
+import static org.labkey.api.dataiterator.SampleUpdateAddColumnsDataIterator.CURRENT_SAMPLE_STATUS_COLUMN_NAME;
 import static org.labkey.api.exp.api.ExpData.DATA_INPUTS_PREFIX_LC;
 import static org.labkey.api.exp.api.ExpData.DATA_INPUT_PARENT;
 import static org.labkey.api.exp.api.ExpMaterial.ALIQUOTED_FROM_INPUT;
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2930,4 +2937,88 @@ public DataIterator getDataIterator(DataIteratorContext context)
...
+            {
+                _context.getErrors().addRowError(new ValidationException(String.format("Updating sample data when status is %s is not allowed.", oldStatus.getLabel())));
+                return hasNext;
+            }
+
+            Integer newState = (Integer) get(_sampleStateCol);
+            DataState newStatus = _allStates.get(newState);
+            boolean newAllowsOp = SampleStatusService.get().isOperationPermitted(newStatus, SampleTypeService.SampleOperations.EditMetadata);
+
+            if (!newAllowsOp)
```
labkey-susanh: I'm not sure this is quite right.  We'll get here if we had an old status and a new status and the old status was a Locked status. If the new status is the same as the old status, that should be allowed (the case of providing the column so some rows can change values but not all), so I think a check of `newState != oldState` is also needed here.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -722,7 +723,7 @@ else if (!isAliquot && isAliquotField)
             }
         }
         // had a locked status before and either not updating the status or updating to a new locked status
-        if (hasNonStatusChange && !oldAllowsOp && (newStatus == null || !newAllowsOp))
+        if (hasNonStatusChange && !oldAllowsOp && !newAllowsOp)
```
labkey-susanh: I think this doesn't handle the case when the status has not changed. (Started out Locked and didn't change that, so newAllowsOp will be true because the value new `newStatus` is `null`.) I think we'll need to check if the `newRow` has a `SampleState` column at all to distinguish these cases.

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: Good point about `new != old`. I see the _update case allows "modified/by/samplestate/folder" column updates, but none of "modified/by/folder" is user updatable. So for the DIB case, we just need to check that "samplestate" is the only provided column, besides "sampleId".

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: I thought that's what `hasNonStatusChange` does?

labkey-susanh: `hasNonStatusChange` will indicate if there's something other than status changing. If I have a locked sample and provide an update to a Data field but no SampleState field, then `hasNonStatusChange` is true, `oldAllowsOp` is false but `newAllowsOp` is true (because `newStatus` is `null`), so I believe the validation exception will not be thrown when it should be.

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: updated so status update is allowed when no other columns are provided

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: got it. should be fixed now

labkey-susanh (approved):

PR #5395, opened 2024-04-03
Add "Require Secondary Authentication" role

labkey-adam:
#### Rationale
We want administrators to have the option to require Two-Factor Authentication (e.g., TOTP or Duo) for only a subset of highly privileged users, such as Site Administrators and Developers. This may be done on a long-term basis but the ability also allows a progressive roll-out of 2FA. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=50024

#### Related Pull Requests
- https://github.com/LabKey/premiumModules/pull/9

#### Changes
- Introduce a standard option (essentially, "Require 2FA for all users OR only users assigned the special 2FA role") to all secondary authentication configuration dialogs. Persist and respect it.
- Stop excluding "Site Administrators" group from all role assignments... this group is no longer special plus Site Admins are not granted all roles automatically

### DIFF
--- a/api/src/org/labkey/api/security/AuthenticationManager.java
+++ b/api/src/org/labkey/api/security/AuthenticationManager.java
@@ -1455,9 +1455,9 @@ public class AuthenticationManager
                     if (notRequired)
                     {
                         if (bypass)
-                            _log.info("Per configuration, bypassing secondary authentication for provider: " + provider.getClass());
+                            _log.info("Per application.properties configuration, bypassing secondary authentication for provider: " + provider.getClass());
                         else
-                            _log.info("Bypassing secondary authentication since authenticated user lacks the \"Require Secondary Authentication\" role: " + primaryAuthUser.getDisplayName(null));
+                            _log.debug("Bypassing secondary authentication since authenticated user lacks the \"Require Secondary Authentication\" role: " + primaryAuthUser.getDisplayName(null));
 
                         setSecondaryAuthenticationUser(session, configuration.getRowId(), primaryAuthUser);
                         continue;
--- a/api/src/org/labkey/api/security/AuthenticationManager.java
+++ b/api/src/org/labkey/api/security/AuthenticationManager.java
@@ -15,6 +15,8 @@
  */
 package org.labkey.api.security;
 
+import jakarta.servlet.http.HttpServletRequest;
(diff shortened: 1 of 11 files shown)


api/src/org/labkey/api/security/AuthenticationManager.java
```diff
@@ -1450,9 +1450,15 @@ public URLHelper getRedirectURL()
...
                 {
                     SecondaryAuthenticationProvider<?> provider = configuration.getAuthenticationProvider();
-                    if (provider.bypass())
+                    boolean bypass = provider.bypass();
+                    boolean notRequired = bypass || !configuration.isRequired(primaryAuthUser);
+                    if (notRequired)
                     {
-                        _log.info("Per configuration, bypassing secondary authentication for provider: " + provider.getClass());
+                        if (bypass)
+                            _log.info("Per configuration, bypassing secondary authentication for provider: " + provider.getClass());
```
labkey-jeckels: These seem more like debug level logging. This will get very noisy on servers like labkey.org if we enable 2FA for site admins, for example.

api/src/org/labkey/api/security/BaseSecondaryAuthenticationConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+    public boolean isRequired(User user)
+    {
+        return _requiredFor.isRequired(user);
+    }
+
+    @Override
+    public @NotNull Map<String, Object> getCustomProperties()
+    {
+        Map<String, Object> map = new HashMap<>();
+        map.put("requiredFor", getRequiredFor().name());
```
labkey-jeckels: Worth pulling out as a constant since it's sprinkled across a few files

labkey-jeckels (approved):

api/src/org/labkey/api/security/AuthenticationManager.java
labkey-adam: I've left the site-level bypass logging at info level, since that's meant to be temporary. Switched the pre-user bypass to debug.

api/src/org/labkey/api/security/BaseSecondaryAuthenticationConfiguration.java
labkey-adam: Done

PR #5401, opened 2024-04-05
Sample status colors

labkey-susanh:
#### Rationale
Because we see a good amount of usage of sample statuses, we want to allow users to visually distinguish the statuses in the UI by providing custom color choices used to display the status pills in grids and detail views and for use in the dashboard and insight graphs.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/121
- https://github.com/LabKey/labkey-ui-premium/pull/381
- https://github.com/LabKey/labkey-ui-components/pull/1467

#### Changes
- Add color field to `core.dataStates` table and all its relatives


api/src/org/labkey/api/qc/export/AbstractDataStateImporter.java
```diff
@@ -44,7 +44,9 @@ public static void importQCStates(ImportExportContext<?> ctx, StudyqcDocument do
                         state.setLabel(xmlState.getName());
                         state.setDescription(xmlState.getDescription());
                         state.setPublicData(xmlState.getPublic());
-                        state.setStateType(xmlState.getType() == null ? null : xmlState.getType().toString());
+                        String stateType = xmlState.getType() == null ? null : xmlState.getType().toString();
+                        state.setStateType(stateType);
```
cnathe: minor: I assume this change was made assuming you were going to use the stateType for the default color somehow. Since it isn't used, this could be reverted.

cnathe (approved):

api/src/org/labkey/api/qc/export/AbstractDataStateImporter.java
labkey-susanh: Done

PR #5405, opened 2024-04-05
Improve report metrics

labkey-adam:
#### Rationale
Primary goal is to add metrics for chart count by type, for example:

```
  "genericChartCountsByRenderType": {
    "BOX_PLOT": 4,
    "LINE_PLOT": 7,
    "SCATTER_PLOT": 47,
    "BAR_PLOT": 3,
    "PIE_CHART": 2
  }
```

It also moves the metrics handling from study to query, so that all deployments will report this information.

And there's some spring cleaning... removing unused code, migrating away from deprecated methods, fixing IntelliJ warnings, etc.


query/src/org/labkey/query/reports/ReportServiceImpl.java
```diff
@@ -1061,4 +1062,26 @@ public void run(Logger log)
...
     }
+
+    public static void registerUsageMetrics(String moduleName)
+    {
+        UsageMetricsService svc = UsageMetricsService.get();
+        if (null != svc)
+        {
+            // Iterate all the database reports once and produce two occurrence maps: all reports by type and just the charts by render type
+            MultiSet<GenericChartReport.RenderType> chartCountsByRenderType = new HashMultiSet<>();
+            Map<String, Long> countsByType = ContainerManager.getAllChildren(ContainerManager.getRoot()).stream()
```
labkey-jeckels: Hopefully this isn't too slow on servers with tens of thousands of containers. I imagine that it's not easy to get these aggregate counts via a DB query since it's all stored as XML.

labkey-jeckels (approved):

query/src/org/labkey/query/reports/ReportServiceImpl.java
labkey-adam: Iterating all containers is common and tends to be fast, since they're all cached. Same with database reports. Though these caches do have a DAY time-to-live set, which is probably not ideal for a (typically) once-a-day operation like this. I did make the effort to ensure we collect both metrics with a single iteration. And we were doing all this work before (in the study module), so at least there's no change.

labkey-jeckels: Right, as long as the query/queries we do for each container are fast too.

labkey-adam: Sure, and in this case, no queries happen when the reports are cached

PR #5406, opened 2024-04-06
Relax builders to take Renderable instead of HtmlString

labkey-adam:
#### Rationale
Builder methods that take HTML as `Renderable` integrate better with `DOM` constructions, since they avoid unnecessary `HtmlString` conversion.


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -151,20 +219,70 @@
 import org.labkey.api.settings.WriteableAppProps;
 import org.labkey.api.settings.WriteableFolderLookAndFeelProperties;
 import org.labkey.api.settings.WriteableLookAndFeelProperties;
-import org.labkey.api.util.*;
 import org.labkey.api.util.Button;
+import org.labkey.api.util.ConfigurationException;
```
labkey-jeckels, line 223: Someday we should standardize our import preferences in IntelliJ.

labkey-jeckels (approved):

core/src/org/labkey/core/admin/AdminController.java
labkey-adam: I think Susan made a change semi-recently to eliminate or greatly reduce importing *

labkey-matthewb (approved):

PR #5408, opened 2024-04-08
Improve site validation

labkey-adam:
#### Rationale
Current site validation UI is very primitive and inflexible. It's available only in the root and full validation (which can take a long time, even timing out on large sites) takes place immediately after clicking the admin console link.

#### Changes
- Configuration page to provide guidance and options
- Folder management tab to run validation in a project or folder
- Include subfolders (or not) option
- Choose which validators to run
- Various results formatting improvements

image

### DIFF
--- a/core/src/org/labkey/core/CoreModule.java
+++ b/core/src/org/labkey/core/CoreModule.java
@@ -33,8 +33,6 @@ import org.labkey.api.admin.FolderSerializationRegistry;
 import org.labkey.api.admin.HealthCheck;
 import org.labkey.api.admin.HealthCheckRegistry;
 import org.labkey.api.admin.notification.NotificationService;
-import org.labkey.api.admin.sitevalidation.SiteValidationProvider;
-import org.labkey.api.admin.sitevalidation.SiteValidationResultList;
 import org.labkey.api.admin.sitevalidation.SiteValidationService;
 import org.labkey.api.analytics.AnalyticsService;
 import org.labkey.api.attachments.AttachmentService;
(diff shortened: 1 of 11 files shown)


core/src/org/labkey/core/CoreModule.java
```diff
@@ -479,6 +481,54 @@ public QuerySchema createSchema(DefaultSchema schema, Module module)
         if (null != svc)
         {
             svc.registerProvider("core", new PermissionsValidator());
+            svc.registerProvider("core", new SiteValidationProvider()
```
labkey-jeckels: Are these going to be committed? I'm not seeing how they're enabled/disabled for testing purposes.

core/src/org/labkey/core/admin/sitevalidation/SiteValidationServiceImpl.java
```diff
@@ -48,97 +45,126 @@ public void registerProvider(String module, SiteValidationProvider provider)
...
-            }
+            Set<SiteValidationProvider> moduleSet = containerValidators.computeIfAbsent(module, k -> new ConcurrentSkipListSet<>());
             moduleSet.add(provider);
         }
     }
 
+    @Override
+    public Map<String, Set<SiteValidationProvider>> getSiteProviders()
+    {
+        return siteValidators;
```
labkey-jeckels: Return immutable wrappers?

core/src/org/labkey/core/admin/sitevalidation/SiteValidationServiceImpl.java
```diff
@@ -48,97 +45,126 @@ public void registerProvider(String module, SiteValidationProvider provider)
...
         {
-            Map<SiteValidatorDescriptor, Map<String, Map<String, SiteValidationResultList>>> moduleResults = new TreeMap<>();
-            allContainerResults.put(moduleValidators.getKey(), moduleResults);
+            final List<Container> parentList;
 
-            // Initialize maps for each provider
-            for (SiteValidationProvider validator : moduleValidators.getValue())
+            if (topLevel.isRoot())
+            {
+                parentList = ContainerManager.getProjects();
```
labkey-jeckels, line 123: I know this is existing behavior, but we never validate the root container? Is there any use in doing it too?

labkey-jeckels (approved):

core/src/org/labkey/core/CoreModule.java
labkey-adam: No, my final commit to remove them (and do a bit of cleanup) didn't get pushed. Pushed a few minutes ago.

core/src/org/labkey/core/admin/sitevalidation/SiteValidationServiceImpl.java
labkey-adam: Sure

core/src/org/labkey/core/admin/sitevalidation/SiteValidationServiceImpl.java
labkey-adam: Yes, existing. Some validators are not very compelling for root, but technically we should since (e.g.) nothing prevents you from creating wikis there. I think I'll leave that for a follow-on PR, though, since we may want to switch from `includeSubfolders` to a tri-state (root only, root + projects, all folders).

labkey-klum (approved):

PR #5422, opened 2024-04-10
Refactor Resource interface

labkey-matthewb:
#### Rationale
Remove ambiguous methods in interface Resource.
Prefer Path and Path.Part to ambiguous usage of String.

#### Related Pull Requests
- 

#### Changes
Path.getPathPart()
FileUtil.appendName()

### DIFF
--- a/query/src/org/labkey/query/QueryServiceImpl.java
+++ b/query/src/org/labkey/query/QueryServiceImpl.java
@@ -2383,7 +2383,7 @@ public class QueryServiceImpl implements QueryService
         // TODO: Schema names can contains slashes (e.g. assay names)
         // TODO: Not sure how the callers deal with this, when they map these into file-system directories.
         // TODO: We could throw IllegalArgument and let the caller handle it.
-        return Path.rawParts(subDirs.toArray(new String[0]));
+        return new Path(subDirs.toArray(new String[0]));
     }
 
     // Look for file-based definitions in modules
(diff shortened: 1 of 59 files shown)


labkey-jeckels:
Did this break the FileContent tests?

query/src/org/labkey/query/QueryServiceImpl.java
```diff
@@ -2380,7 +2380,10 @@ private Path createSchemaPath(SchemaKey schemaKey)
         List<String> subDirs = new ArrayList<>(schemaKey.size() + 1);
         subDirs.add(QueryService.MODULE_QUERIES_DIRECTORY);
         subDirs.addAll(schemaKey.getParts());
-        return new Path(subDirs);
+        // TODO: Schema names can contains slashes (e.g. assay names)
+        // TODO: Not sure how the callers deal with this, when they map these into file-system directories.
+        // TODO: We could throw IllegalArgument and let the caller handle it.
```
labkey-jeckels, line 2385: I'd vote for replacing `/` with`_`. If we throw a runtime exception, it's unlikely that callers will know they may need to deal with it.

labkey-jeckels (approved):
This looks good, with the possible exception of the FileContent test failure.

labkey-jeckels (commented):
Should we add a warning on callers of `new File(File, String)`?

query/src/org/labkey/query/QueryServiceImpl.java
labkey-matthewb: I didn't want to break things in release branch.

labkey-matthewb:
> Should we add a warning on callers of `new File(File, String)`?

Yes, on develop

labkey-matthewb:
> Did this break the FileContent tests?

Hmm, passed several times... Let's see what teamcity says...

labkey-matthewb:
> Did this break the FileContent tests?

The sendDigest functionality seems to be sensitive to database/webserver time mismatch.
And also a missing null check

PR #5425, opened 2024-04-11
Switch Strict withCounter experimentalFlag behavior

XingY:
#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5425
- https://github.com/LabKey/limsModules/pull/138

#### Changes
-


experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -225,8 +226,16 @@ protected void init()
...
-                "When withCounter or rootSampleCount is used in name expression, make sure the count increments one-by-one and does not jump.", true);
+        if (CoreSchema.getInstance().getSqlDialect().isSqlServer())
+        {
+            AdminConsole.addExperimentalFeatureFlag(NameGenerator.EXPERIMENTAL_WITH_COUNTER, "Use strict incremental withCounter and rootSampleCount expression",
+                    "When withCounter or rootSampleCount is used in name expression, make sure the count increments one-by-one and does not jump.", true);
+        }
+        else
+        {
+            AdminConsole.addExperimentalFeatureFlag(NameGenerator.EXPERIMENTAL_ALLOW_GAP_COUNTER, "Allow gap with withCounter and rootSampleCount expression",
+                    "Check this option if gaps in the count generated by withCounter or rootSampleCount name expression is allowed.", true);
```
labkey-susanh:
```suggestion
                    "Check this option if gaps in the count generated by withCounter or rootSampleCount name expression are allowed.", true);

```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -9712,6 +9715,19 @@ public void handleAssayNameChange(String newAssayName, String oldAssayName, Assa
...
+    @Override
+    public boolean useStrictCounter()
+    {
+        if (CoreSchema.getInstance().getSqlDialect().isSqlServer())
+        {
+            return AppProps.getInstance().isExperimentalFeatureEnabled(EXPERIMENTAL_WITH_COUNTER);
+        }
+        else
+        {
+            return !AppProps.getInstance().isExperimentalFeatureEnabled(EXPERIMENTAL_ALLOW_GAP_COUNTER);
```
labkey-susanh, line 9727: The `NoGap` token in the naming pattern will still cause there to be no gaps for that individual entity type, right? This just allows other entities to have gaps.

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: yep, it's for the type that specifically use the naming pattern only

PR #5427, opened 2024-04-11
Improve-cross project actions - Edit via Grid

cnathe:
#### Rationale
As part of the effort to make "cross folder" action improvements in the apps, this set of PRs takes on the "Edit in Grid" action. This will allow for users with proper permission to edit selected rows in a sample, source, and assay runs/results grid when the set of selections includes rows from multiple containers.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1472

#### Changes
- add comment about SampleOperations that have null for permissionClass
- GetAssayResultsOperationConfirmationDataAction to determine notPermitted for assay results Edit in Grid
- add DataOperations.Edit to enum


assay/src/org/labkey/assay/AssayController.java
```diff
@@ -1724,8 +1725,45 @@ public Object execute(AssayRunOperationConfirmationForm form, BindException erro
...
+        @Override
+        public Object execute(AssayOperationConfirmationForm form, BindException errors) throws Exception
+        {
+            Collection<Integer> allowedIds = form.getIds(false);
+
+            ExperimentService service = ExperimentService.get();
+            ExpProtocol protocol = service.getExpProtocol(form.getProtocolId());
+            AssayProvider provider = AssayService.get().getProvider(protocol);
+            AssaySchema schema = provider.createProtocolSchema(getUser(), getContainer(), protocol, null);
+            TableInfo tableInfo = schema.getTable(AssayProtocolSchema.DATA_TABLE_NAME, ContainerFilter.EVERYTHING);
```
labkey-nicka: nit: Consider using `schema.getTableOrThrow()` instead.

assay/src/org/labkey/assay/AssayController.java
```diff
@@ -1724,8 +1725,45 @@ public Object execute(AssayRunOperationConfirmationForm form, BindException erro
...
+
+            // need to query to get the dataIds for the data rowIds so that we can check container permissions on that exp.data table
+            SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("RowId"), allowedIds, CompareType.IN);
+            List<Map<String, Object>> dataRows = new TableSelector(tableInfo, PageFlowUtil.set("rowid", "dataid"), filter, null).getMapCollection().stream().toList();
```
labkey-nicka: nit: `getMapCollection().stream().toList()` seems like it could just be `getMapCollection()` as the return type is `Collection<Map<String, Object>>` which can be iterated on just the same.

assay/src/org/labkey/assay/AssayController.java
```diff
@@ -1724,8 +1725,45 @@ public Object execute(AssayRunOperationConfirmationForm form, BindException erro
...
+            List<Integer> permittedRowIds = new ArrayList<>();
+            for (Map<String, Object> dataRow : dataRows)
+            {
+                Integer rowId = (Integer) dataRow.get("rowid");
+                ExpData data = ExperimentService.get().getExpData((Integer) dataRow.get("dataid"));
```
labkey-nicka, line 1750: nit: Consider aggregation for this query by using `getExpDatas(...)`

assay/src/org/labkey/assay/AssayController.java
```diff
@@ -1724,8 +1725,45 @@ public Object execute(AssayRunOperationConfirmationForm form, BindException erro
...
     }
 
+    @Marshal(Marshaller.Jackson)
+    @RequiresPermission(ReadPermission.class)
+    public static class GetAssayResultsOperationConfirmationDataAction extends ReadOnlyApiAction<AssayOperationConfirmationForm>
+    {
+        @Override
+        public Object execute(AssayOperationConfirmationForm form, BindException errors) throws Exception
+        {
+            Collection<Integer> allowedIds = form.getIds(false);
```
labkey-nicka, line 1735: nit: Just curious, has there been any discussion about bounding these results? Like 10k max?

labkey-nicka (approved):

assay/src/org/labkey/assay/AssayController.java
cnathe: done

assay/src/org/labkey/assay/AssayController.java
cnathe: yup, that works. done

assay/src/org/labkey/assay/AssayController.java
cnathe: I originally had `getExpDatas` but in the end I need the set of assay results "RowIds"  but the permission check is on the data.getRun(). When I was using getExpDatas(), I had to do a second loop over the dataRows to get the rowIds for those that were in the allowed ExpData collection. The current implementation seemed cleaner.

assay/src/org/labkey/assay/AssayController.java
cnathe: I assume your talking about a server side check, right?
The client side has the check for "max of 1000 selected" for many of these grid button actions (including the one that calls this action). 
@labkey-susanh do you know if there has been any consideration on this or any perf assessment for large selections to these actions?

labkey-susanh: IIRC, in at least one case, we limited the number of items to be acted upon after observing that the call to get the operation confirmation data was somewhat expensive, but no, we have not considered limiting this from the server side (just oversight, really). Since these actions are currently exclusively used by the applications, I think we could implement these limits on the server side as well without interfering anywhere else. I don't think there are places in the apps that currently have no limits, but I do think there might be different limits for different actions.

PR #5437, opened 2024-04-16
calculated columns

labkey-matthewb:
#### Rationale
https://github.com/LabKey/limsModules/issues/95

#### Related Pull Requests
- 

#### Changes
-


api/src/org/labkey/api/data/AbstractTableInfo.java
```diff
@@ -350,8 +376,32 @@ public SQLFragment getFromSQL(String alias)
             return new SQLFragment().append("(").append(getFromSQL()).append(") ").append(alias);
     }
 
+    /** When a table a) overrides (String alias, Set<FieldKey> cols) b) has CalculatedColumsn we need to make sure that
```
cnathe:
```suggestion
    /** When a table a) overrides (String alias, Set<FieldKey> cols) b) has CalculatedColumns we need to make sure that

```

api/src/org/labkey/api/data/AbstractTableInfo.java
```diff
@@ -1319,10 +1378,15 @@ protected void loadAllButCustomizerFromXML(QuerySchema schema, @Nullable TableTy
                 if (null == xmlColumn.getColumnName())
                     throw new ConfigurationException("Table schema has column with empty columnName attribute: ", xmlTable.getTableName());
 
-                if (xmlColumn.getWrappedColumnName() != null)
+                if (xmlColumn.isSetWrappedColumnName() && isNotBlank(xmlColumn.getWrappedColumnName()))
                 {
                     wrappedColumns.add(xmlColumn);
                 }
+                else if (allowCalculatedColumns && xmlColumn.isSetValueExpression() && isNotBlank(xmlColumn.getValueExpression()))
```
cnathe, line 1386: should this be `!allowCalculatedColumns`? I assume that when allowCalculatedColumns is true we don't want to treat these value expression as "wrappedColumns" but when it is false we do. Maybe I am misunderstanding that allowCalculatedColumns variable though.

api/src/org/labkey/api/data/AbstractTableInfo.java
```diff
@@ -1338,23 +1402,50 @@ protected void loadAllButCustomizerFromXML(QuerySchema schema, @Nullable TableTy
...
+                {
+                    sql = xmlColumn.getValueExpression();
+                }
+                else if (xmlColumn.isSetWrappedColumnName() && isNotBlank(xmlColumn.getWrappedColumnName()))
                 {
-                    var wrappedColumn = new WrappedColumn(column, wrappedColumnXml.getColumnName());
-                    initColumnFromXml(schema, wrappedColumn, wrappedColumnXml, errors);
+                    sql = LabKeySql.quoteIdentifier(xmlColumn.getWrappedColumnName());
+                }
+                try
```
cnathe, line 1438: the similar code in SchemaColumnMetaData.java has a check like `if (isNotBlank(sql))`. Do we need that here? or will that just get handled in the catch block?

api/src/org/labkey/api/data/StatementUtils.java
```diff
@@ -748,7 +748,7 @@ public ParameterMapStatement createStatement(Connection conn, @Nullable Containe
 
         for (ColumnInfo column : table.getColumns())
         {
-            if (column instanceof WrappedColumn)
+            if (column instanceof WrappedColumn || column.isCalculated())
```
cnathe, line 751: are WrappedColumn's also set as isCalculated true? if so, seems like the check could drop the instanceof part.

api/src/org/labkey/api/query/QueryParseException.java
```diff
@@ -113,6 +127,8 @@ public JSONObject toJSON(String sql)
             String errorStr = lines[getLine() - 1];
             error.put("errorStr", errorStr);
         }
+        if (!isBlank(_field))
```
cnathe: minor: use `isNotBlank`?

core/api-src/org/labkey/api/admin/TableXmlUtils.java
```diff
@@ -211,24 +211,20 @@ private static void compareTableDocuments(TablesDocument dbTablesDoc,
...
-                    }
-                    else
+                    if (isCalculated)
                     {
                         if (isColumnInDatabase)
-                        {
                             rlOut.addError("ERROR: Table \"").append(xmlTable.getTableName()).append("\", column \"").append(xmlColName).append("\" found in database but shouldn't be, since it's a wrapped column");
-                        }
-
+                        // TODO validate wrappedColumnName and valueExpression
```
cnathe: the check that was removed was for "xml mentions a wrappedColumnName that is not in database". would that check go here in this TODO?

api/src/org/labkey/api/data/AbstractTableInfo.java
```diff
@@ -1338,23 +1402,50 @@ protected void loadAllButCustomizerFromXML(QuerySchema schema, @Nullable TableTy
...
+            for (var col : _columnMap.values())
+                existingColumns.put(col.getFieldKey(), col);
+
+            for (ColumnType xmlColumn : wrappedColumns)
             {
-                var column = getMutableColumn(wrappedColumnXml.getWrappedColumnName());
+                // expression column can't hide built-in column
+                if (null != getMutableColumn(xmlColumn.getColumnName()))
+                {
+                    warnings.add(new QueryParseWarning("A wrapped or calculated column cannot hide an existing column",null,0,0).setFieldName(xmlColumn.getColumnName()));
```
cnathe:
```suggestion
                    warnings.add(new QueryParseWarning("A wrapped or calculated column cannot hide an existing column.",null,0,0).setFieldName(xmlColumn.getColumnName()));

```

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -8059,6 +8059,134 @@ public Object execute(GetQueryDetailsAction.Form form, BindException errors) thr
...
+        {
+            if (errors.hasErrors())
+                return errors;
+            JSONObject result = new JSONObject(Map.of("success",true));
+            var requiredColumns = new HashSet<FieldKey>();
+            JdbcType jdbcType = JdbcType.OTHER;
+            try
+            {
+                var schema = DefaultSchema.get(getViewContext().getUser(), getViewContext().getContainer()).getUserSchema("core");
+                var table = new VirtualTable(schema.getDbSchema(), "EXPR", schema){};
```
cnathe, line 8153: when I first saw this action, I was assuming that you'd pass in a schema/query name that would be used in some way to determine valid columns. But this is a level above / before that. It requires that the client / caller know the column names from the schema/query they are dealing with and this is just parsing the expression itself against that column map. Does that sounds right? So this would work for the domain designer UI which could know about columns being added in the same transaction and aren't yet saved to the DB for that schema/query.

query/src/org/labkey/query/sql/CalculatedExpressionColumn.java
```diff
@@ -0,0 +1,371 @@
...
+    private QExpr _resolveFields(QExpr expr, Map<FieldKey, ? extends ColumnInfo> columnMap)
+    {
+        if (expr instanceof QQuery || expr instanceof QUnion)
+        {
+            setResolveException("SELECT and UNION are not allowed in calculated columns.");
+            return null;
+        }
+        if (expr instanceof QRowStar)
+        {
+            setResolveException("SELECT and UNION are not allowed in calculated columns.");
```
cnathe: minor: should this have a slightly different error message? if not, it can be added to the above if block

query/src/org/labkey/query/sql/CalculatedExpressionColumn.java
```diff
@@ -0,0 +1,371 @@
...
+            {
+                setResolveException(key.toSQLString() + " has PHI, it cannot be used in a calculated expression.");
+                return null;
+            }
+            return new _BoundColumn(key);
+        }
+
+        if (expr instanceof QAggregate)
+        {
+            setResolveException("SELECT and UNION are not allowed in calculated columns.");
```
cnathe: minor: same comment here, can this be moved to the top if block? or should it have a different error message?

cnathe (approved):

api/src/org/labkey/api/data/StatementUtils.java
labkey-matthewb: I don't think WrappedColumn enforces anything (and there's at least one subclass)

api/src/org/labkey/api/data/AbstractTableInfo.java
labkey-matthewb: This is just allowing the wrappedColumnName="" syntax for backward-compatibility.  But _for now_ disallowing <valueExpression> for queries  (limit testing surface and design work for SQL integration for a low value scenario)

api/src/org/labkey/api/data/AbstractTableInfo.java
labkey-matthewb: I see.  Since both "if" statement check isNotBlank() I think the isNotBlank() in SchemaColumnMetaData is redundant.

core/api-src/org/labkey/api/admin/TableXmlUtils.java
labkey-matthewb: This is doing the same validation as before.  The comment suggests we could do more (e.g. syntax check)

query/src/org/labkey/query/controllers/QueryController.java
labkey-matthewb: My though was that for "saved" tables, getQueryDetails gives you validation warnings.  In the designer, presumably the users may have added/removed/renamed columns.  The server doesn't know what the table will look like after it is saved.

labkey-matthewb: We could add an optional schema/table to pre-populate  the column map.

query/src/org/labkey/query/sql/CalculatedExpressionColumn.java
labkey-matthewb: Yup

query/src/org/labkey/query/sql/CalculatedExpressionColumn.java
labkey-matthewb: Yup

PR #5438, opened 2024-04-17
Issues 50133 & 50175: Fix some date handling issues

labkey-susanh:
#### Rationale
Issue [50133](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50133): When an entity has more than one ancestor of a particular type, our Ancestor display columns return string values such as "2 values", which can't be cast to dates or times for writing to Excel workbooks, so we check the type before trying to cast to a Date or Time (which is a good idea anyway, I think).

Issue [50175](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50175): When splitting up an input file that may span sample types, source types, or containers, we use the default toString() methods for the fields involved. For Date fields, this means we will convert an input string such as "10-Apr-2024" to something like "Wed Apr 10 00:00:00 BST 2024". Generally, this will round trip and come back in as the 10th of April 2024, but the three-letter abbreviations for timezones are ambiguous and, in particular, "BST" can mean "British Summer Time" (as written by the default `toString()` method for `java.util.Date`) but it can also mean "Bangladesh Standard Time" (as read by Java), and that causes an unwanted time shift to occur. Here we update the method that writes out values to the separate files so it will use `DateUtil.formatIsoDateLongTime`, which does not include the timezone (since the value will have already been converted to the local timezone upon reading it from the original file) for writing out `Date` values.

#### Related Pull Requests
- 

#### Changes
- Update `ExcelColumn.writeCell` to check types before casting
- Update `MultiDataTypeCrossProjectDataIterator` to better handle round-tripping of Date and DateTime values

### DIFF
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -96,6 +96,7 @@ import org.labkey.api.security.permissions.InsertPermission;
 import org.labkey.api.security.permissions.UpdatePermission;
 import org.labkey.api.study.publish.StudyPublishService;
 import org.labkey.api.usageMetrics.SimpleMetricsService;
+import org.labkey.api.util.DateUtil;
 import org.labkey.api.util.FileUtil;
 import org.labkey.api.util.GUID;
 import org.labkey.api.util.Pair;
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2865,6 +2868,13 @@ else if (aliasMap.containsKey(name) && _dataIdIndex != -1)
             return new TypeData(container, sampleType, samplesTable, dataFile, fieldIndexes, dependencyIndexes, dataRows, new ArrayList<>());
         }
 
+        private Object getSerializingObject(Object data)
+        {
+            if (data instanceof Date d && !(data instanceof Time))
+                return DateUtil.formatIsoDateLongTime(d);
```
XingY: Should we use formatIsoDateLongTime(d, true) instead to preserve ms?

XingY (approved):

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: Yeah, good idea.

PR #5446, opened 2024-04-22
Use server GUID to tie database to full-text search index

labkey-adam:
#### Rationale
Improve the ability to link a full-text-search index with the database that produced it. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=50201

#### Changes
- Add ability to save arbitrary global properties to the Lucene index
- On every crawler start, retrieve the saved GUID. If it doesn't match the database GUID, delete the index (and lastIndexed).
- Log a reason for every delete index and reindex operation

### DIFF
--- a/api/src/org/labkey/api/search/SearchService.java
+++ b/api/src/org/labkey/api/search/SearchService.java
@@ -68,11 +68,6 @@ import java.util.Set;
 import java.util.concurrent.Future;
 import java.util.function.Function;
 
-/**
- * User: matthewb
- * Date: Nov 12, 2009
- * Time: 12:54:01 PM
- */
 public interface SearchService extends SearchMXBean
 {
     // create logger for package which can be set via logger-manage.view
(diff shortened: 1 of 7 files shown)


labkey-matthewb (approved):
This seems good to me.  Because this area seems to generate confusion.  My only comment would be that maybe we should document in the code how this is expected to behave in known common/problem scenarios (even if the expected behavior is nothing happens).  E.g.  Blue-Green,  Restore Production database to Staging,  "retargeting" an instance after creation, etc.   Might as well document it here.

labkey-adam:
> This seems good to me. Because this area seems to generate confusion. My only comment would be that maybe we should document in the code how this is expected to behave in known common/problem scenarios (even if the expected behavior is nothing happens). E.g. Blue-Green, Restore Production database to Staging, "retargeting" an instance after creation, etc. Might as well document it here.

The linked issue includes extensive narrative about this (maybe too much), but I agree that some comments in the code could be helpful. I'll keep this basic and point to the issue for details.

PR #5452, opened 2024-04-23
Issue 50212: Check if parent directory isWritable before calling FileUtil.createDirectories()

cnathe:
#### Rationale
When the FileSystemAttachmentParent attempts to ensure that the @scripts dir exists, it will create it if not already present. however if the file system for the parent dir is read-only, this can results in a AccessDeniedException. This PR adds a Files.isWritable() check for the parent dir before the call to FileUtil.createDirectories().

#### Changes
- Check if parent directory isWritable before calling FileUtil.createDirectories()


labkey-matthewb (approved):

cnathe:
Updating this to target release24.3-SNAPSHOT

PR #5455, opened 2024-04-24
Issue 50241: legacyContextPath redirect can mangle URL

labkey-jeckels:
#### Rationale
We're replacing the first instance of the context path in the request URL, which can match against the host name instead of the file/path

#### Changes
- Parse the URL and only replace in the file/path
- Also carry the GET parameters forward, as they're not included in `HttpServletRequest.getRequestURL()`

### DIFF
--- a/api/src/org/labkey/api/view/RedirectorServlet.java
+++ b/api/src/org/labkey/api/view/RedirectorServlet.java
@@ -1,6 +1,7 @@
 package org.labkey.api.view;
 
 import java.io.IOException;
+import java.net.URL;
 
 import jakarta.servlet.ServletException;
 import jakarta.servlet.http.HttpServlet;
@@ -27,11 +28,20 @@ public class RedirectorServlet extends HttpServlet
         if ("get".equalsIgnoreCase(request.getMethod()))
         {
             // Send a redirect to let the client know there's a new preferred URL
-            String originalUrl = request.getRequestURL().toString();
-            String redirectUrl = originalUrl.replaceFirst(_legacyContextPath, "");
+            StringBuffer rawUrl = request.getRequestURL();
+            // getRequestURL() doesn't include GET parameters, so append them if needed
+            if (request.getQueryString() != null)
+            {
+                rawUrl.append("?").append(request.getQueryString());
+            }
+            URL originalUrl = new URL(rawUrl.toString());
+            URL redirectUrl = new URL(originalUrl.getProtocol(),
+                    originalUrl.getHost(),
+                    originalUrl.getPort(),
+                    originalUrl.getFile().replaceFirst(_legacyContextPath, ""));
 
             response.setStatus(HttpServletResponse.SC_MOVED_PERMANENTLY);
-            response.setHeader("Location", redirectUrl);
+            response.setHeader("Location", redirectUrl.toString());
         }
         else
         {


labkey-adam (approved):

api/src/org/labkey/api/view/RedirectorServlet.java
```diff
@@ -27,11 +28,20 @@ protected void service(HttpServletRequest request, HttpServletResponse response)
...
         {
             // Send a redirect to let the client know there's a new preferred URL
-            String originalUrl = request.getRequestURL().toString();
-            String redirectUrl = originalUrl.replaceFirst(_legacyContextPath, "");
+            StringBuffer rawUrl = request.getRequestURL();
+            // getRequestURL() doesn't include GET parameters, so append them if needed
+            if (request.getQueryString() != null)
+            {
+                rawUrl.append("?").append(request.getQueryString());
+            }
```
labkey-tchad, line 36: Does this need to handle app paths (e.g. `samplemanager-app.view#/samples/type`)?

api/src/org/labkey/api/view/RedirectorServlet.java
```diff
@@ -27,11 +28,20 @@ protected void service(HttpServletRequest request, HttpServletResponse response)
...
+            URL originalUrl = new URL(rawUrl.toString());
+            URL redirectUrl = new URL(originalUrl.getProtocol(),
+                    originalUrl.getHost(),
+                    originalUrl.getPort(),
+                    originalUrl.getFile().replaceFirst(_legacyContextPath, ""));
```
labkey-tchad, line 41: Should we ensure that we only replace the legacy context path if the URL file begins with it?
```suggestion
                    originalUrl.getFile().replaceFirst("^" + _legacyContextPath, ""));
```

Which leads me to the thought: What if there's a project or folder that happens to match the legacy context path? The legacy URL `/labkey/labkey/project-begin.view` would redirect to what you want but the "correct" URL would be broken.

labkey-tchad (commented):
How could the legacyContextPath match the hostname? It must start with `/` which I don't think is a legal hostname character (`www./labkey.org`)

labkey-jeckels:
> How could the legacyContextPath match the hostname? It must start with `/` which I don't think is a legal hostname character (`www./labkey.org`)

When you have the full URL, the hostname is preceded by a slash. `https://labkey.org/labkey/home/project-begin.view` gets redirected to `https://.org/labkey/home/project-begin.view` if you have `/labkey` as your legacy context path

api/src/org/labkey/api/view/RedirectorServlet.java
labkey-jeckels: Per the HTTP spec, the browser never sends the hash part of the URL to the server, so it will unfortunately get dropped after the redirect.

api/src/org/labkey/api/view/RedirectorServlet.java
labkey-jeckels: This is indeed a limitation to the current approach but more complicated to try to redirect only if we're otherwise going to serve a 404. If someone hits it, they can keep deploying to their old context path and we can regroup on the best approach to migrate them.

labkey-jeckels: This servlet is registered for URLs that start with the legacy context path, so we don't need to separately check. Your proposed change would be fine, but it won't have a functional difference.

https://github.com/LabKey/platform/blob/5649b28c6498d0ab292bdadd95218cc966940e32/api/src/org/labkey/api/ApiModule.java#L261

PR #5458, opened 2024-04-24
Improve-cross project actions - Edit in Bulk

cnathe:
#### Rationale
As part of the effort to make "cross folder" action improvements in the apps, this set of PRs takes on the "Edit in Bulk" and "Edit Parents/Sources" actions. This will allow for users with proper permission to edit selected rows in a sample, source, and assay runs/results grid when the set of selections includes rows from multiple containers.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1479

#### Changes
- Add container information (id, path, permitted) to the response for OperationConfirmationData
- Issue 50245: use substring to get second part of the alias instead of relying on parts[]
- Remove ExperimentService.getUniqueContainers() and ExperimentService.getIdsNotPermitted() and use exp rows to get their container instead

### DIFF
--- a/assay/src/org/labkey/assay/AssayController.java
+++ b/assay/src/org/labkey/assay/AssayController.java
@@ -1683,27 +1683,35 @@ public class AssayController extends SpringActionController
         @Override
         public Object execute(AssayOperationConfirmationForm form, BindException errors) throws Exception
         {
+            ExperimentService service = ExperimentService.get();
+
             Collection<Integer> allowedIds = form.getIds(false);
+            List<? extends ExpRun> allRuns = service.getExpRuns(allowedIds);
 
             Set<Integer> notAllowedIds = new HashSet<>();
-            ExperimentService service = ExperimentService.get();
             Map<String, Object> response = new HashMap<>();
 
-            TableInfo tableInfo = ExperimentService.get().getTinfoExperimentRun();
+            Collection<Container> containers = new HashSet<>();
+            Collection<Integer> notPermittedIds = new ArrayList<>();
             Class<? extends Permission> permClass = form.getDataOperation().getPermissionClass();
-            Collection<Container> containers = service.getUniqueContainers(allowedIds, tableInfo);
-            response.put("containers", containers == null ? Collections.emptyList(): containers.stream().map(c -> Map.of(
+            for (ExpRun expRun : allRuns)
+            {
+                Container c = expRun.getContainer();
+                containers.add(c);
+                if (permClass != null && !c.hasPermission(getUser(), permClass))
+                    notPermittedIds.add(expRun.getRowId());
+            }
+
+            response.put("containers", containers.stream().map(c -> Map.of(
                     "id", c.getEntityId(),
                     "path", (Object) c.getPath(),
                     "permitted", permClass == null || c.hasPermission(getUser(), permClass)
             )).toList());
 
-            Collection<Integer> notPermittedIds = service.getIdsNotPermitted(getUser(), containers, allowedIds, tableInfo, permClass);
-            List<Map<String, Object>> notPermitted = notPermittedIds == null ? Collections.emptyList() : notPermittedIds.stream().map(id -> Map.of("RowId", (Object) id)).toList();
+            response.put("notPermitted", notPermittedIds.stream().map(id -> Map.of("RowId", (Object) id)).toList());
+
             if (form.getDataOperation() == AssayRunOperations.Delete)
             {
-                List<? extends ExpRun> allRuns = service.getExpRuns(allowedIds);
-
                 service.getObjectReferencers().forEach(referencer ->
                         notAllowedIds.addAll(referencer.getItemsWithReferences(allowedIds, "assay")));
                 allowedIds.removeAll(notAllowedIds);
@@ -1721,13 +1729,11 @@ public class AssayController extends SpringActionController
 
                 response.put("allowed", allowedRows);
                 response.put("notAllowed", notAllowedRows);
-                response.put("notPermitted", notPermitted);
                 return success(response);
             }
 
             response.put("allowed", allowedIds.stream().map(id -> Map.of("RowId", id)).toList());
             response.put("notAllowed", notAllowedIds);
-            response.put("notPermitted", notPermitted);
             return success(response);
         }
     }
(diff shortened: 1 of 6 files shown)


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -3481,7 +3482,16 @@ public Object execute(DataOperationConfirmationForm form, BindException errors)
 
             Map<String, Collection<Map<String, Object>>> response = ExperimentServiceImpl.partitionRequestedOperationObjects(requestIds, notAllowedIds, allData);
 
-            Collection<Integer> notPermittedIds = service.getIdsNotPermitted(getUser(), requestIds, ExperimentService.get().getTinfoData(), form.getDataOperation().getPermissionClass());
+            TableInfo tableInfo = ExperimentService.get().getTinfoData();
+            Class<? extends Permission> permClass = form.getDataOperation().getPermissionClass();
+            Collection<Container> containers = service.getUniqueContainers(requestIds, tableInfo);
```
XingY: Not sure why this is needed. ExperimentServiceImpl.partitionRequestedOperationObjects already returns containerPath for every allowed and notAllowed record. This extra query seems redundant and might slow down the perf.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -3536,8 +3546,16 @@ public Object execute(MaterialOperationConfirmationForm form, BindException erro
 
             Map<String, Collection<Map<String, Object>>> response = ExperimentServiceImpl.partitionRequestedOperationObjects(requestIds, notAllowedIds, allMaterials);
 
-            Collection<Integer> notPermittedIds = service.getIdsNotPermitted(getUser(), requestIds, ExperimentService.get().getTinfoMaterial(), form.getSampleOperation().getPermissionClass());
+            TableInfo tableInfo = ExperimentService.get().getTinfoMaterial();
+            Class<? extends Permission> permClass = form.getSampleOperation().getPermissionClass();
+            Collection<Container> containers = service.getUniqueContainers(requestIds, tableInfo);
```
XingY: same comment about not doing this duplicate query.

assay/src/org/labkey/assay/AssayController.java
```diff
@@ -1687,7 +1687,18 @@ public Object execute(AssayOperationConfirmationForm form, BindException errors)
 
             Set<Integer> notAllowedIds = new HashSet<>();
             ExperimentService service = ExperimentService.get();
-            Collection<Integer> notPermittedIds = service.getIdsNotPermitted(getUser(), allowedIds, ExperimentService.get().getTinfoExperimentRun(), form.getDataOperation().getPermissionClass());
+            Map<String, Object> response = new HashMap<>();
+
+            TableInfo tableInfo = ExperimentService.get().getTinfoExperimentRun();
+            Class<? extends Permission> permClass = form.getDataOperation().getPermissionClass();
+            Collection<Container> containers = service.getUniqueContainers(allowedIds, tableInfo);
```
XingY: same comment here

assay/src/org/labkey/assay/AssayController.java
```diff
@@ -1687,7 +1687,18 @@ public Object execute(AssayOperationConfirmationForm form, BindException errors)
...
+            response.put("containers", containers == null ? Collections.emptyList(): containers.stream().map(c -> Map.of(
+                    "id", c.getEntityId(),
+                    "path", (Object) c.getPath(),
+                    "permitted", permClass == null || c.hasPermission(getUser(), permClass)
+            )).toList());
+
+            Collection<Integer> notPermittedIds = service.getIdsNotPermitted(getUser(), containers, allowedIds, tableInfo, permClass);
```
XingY: This is not new to the current branch, but I think the set of 3 related apis are a bit problematic. We shouldn't return name for id that user doesn't have read perm to.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
cnathe: Yes, the containerPath is in each record for allowed and notAllowed, but since we were already doing this `getUniqueContainers` call on the server side, I figured it was easier to just return the info/data in the response and not have to do another iteration/parsing client side. It shouldn't change perf as this is still just done once (I just moved it outside of the `getIdsNotPermitted()` method and pass in the `containers` collection as a param.

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
cnathe: see above response

assay/src/org/labkey/assay/AssayController.java
cnathe: see above response

assay/src/org/labkey/assay/AssayController.java
cnathe: I don't believe we are returning any names, just RowIds for the notPermitted rows. The user already sent use the set of ids, so they must have at least read perm to the row's container to get that id. notPermitted here means they don't have higher than read permission for that row, so I don't think we are exposing any info they wouldn't already see in the app.

XingY: This is true for notPermitted, but `allowed` and `notAllowed` does have name and containerPath info, and they overlap with notPermitted

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
XingY: I meant that because we already queried for `List<? extends ExpMaterial> allMaterials`, we can already know which ids are permitted vs not, as material.container is known and we know which contains are permitted. So the `getUniqueContainers` and `getIdsNotPermitted` calls are not needed since all information is already in memory.

cnathe: Good point. I refactored things as suggested here: https://github.com/LabKey/platform/pull/5458/commits/3a8238e075bf213d432717d4938404e0942f9a88

assay/src/org/labkey/assay/AssayController.java
cnathe: I see what you mean.

XingY (approved):

PR #5459, opened 2024-04-24
AbDisc: Plate Set Worklist & Instrument Instructions

RosalineP:
#### Rationale
See spec [here](https://docs.google.com/document/d/1TpINpd9UlIJCPYzrAbWxNzvTIqJUw1ZbNN_lvzW8MrQ/edit#heading=h.6oglt1vd831o).
#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5459
- https://github.com/LabKey/limsModules/pull/189
- https://github.com/LabKey/labkey-ui-components/pull/1480
- https://github.com/LabKey/labkey-ui-premium/pull/395

#### Changes
- Create ArrayExcelWriter
- Modify ExcelWriter to support setting full file name
- Create WorklistAction and InstrumentInstructionAction, alongside related supporting fns

### DIFF
--- a/api/src/org/labkey/api/data/ArrayExcelWriter.java
+++ b/api/src/org/labkey/api/data/ArrayExcelWriter.java
@@ -1,19 +1,3 @@
-/*
- * Copyright (c) 2008-2019 LabKey Corporation
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
 package org.labkey.api.data;
 
 import org.apache.poi.ss.usermodel.Sheet;
@@ -24,11 +8,6 @@ import java.io.Writer;
 import java.util.ArrayList;
 import java.util.List;
 
-/**
- * User: Mark Igra
- * Date: May 4, 2007
- * Time: 3:08:34 PM
- */
 public class ArrayExcelWriter extends ExcelWriter
 {
     private final List<List<Object>> data;
@@ -79,7 +58,6 @@ public class ArrayExcelWriter extends ExcelWriter
         @Override
         public Object getValue(RenderContext ctx)
         {
-            //Ignore the context.
             return data.get(currentRow).get(position);
         }
 
@@ -90,7 +68,7 @@ public class ArrayExcelWriter extends ExcelWriter
         }
 
 
-        //NOTE: Methods beyond here are unimplemented, just abstract in base class!
+        //NOTE: Methods beyond here are unimplemented, just abstract in base class
         @Override
         public void renderGridCellContents(RenderContext ctx, Writer out)
         {
--- a/api/src/org/labkey/api/data/ArrayExcelWriter.java
+++ b/api/src/org/labkey/api/data/ArrayExcelWriter.java
@@ -8,12 +8,18 @@ import java.io.Writer;
 import java.util.ArrayList;
 import java.util.List;
 
+// This class supports generating Excel files with duplicate column names. Consider using MapArrayExcelWriter if
+// multiple identical column names is not an implementation concern.
 public class ArrayExcelWriter extends ExcelWriter
 {
-    private final List<List<Object>> data;
+    private final List<Object[]> data;
     private int currentRow = 0;
 
-    public ArrayExcelWriter(List<List<Object>> data, ColumnDescriptor[] cols)
+    /**
+     * @param data The data rows, in which index position of a value corresponds to the desired respective column index
+     * @param cols The columns, in which ordering determines the left-to-right column ordering in the generated Excel
+     */
+    public ArrayExcelWriter(List<Object[]> data, ColumnDescriptor[] cols)
     {
         super(ExcelDocumentType.xls);
         this.data = data;
@@ -22,7 +28,7 @@ public class ArrayExcelWriter extends ExcelWriter
         for (int i = 0; i < cols.length; i++)
         {
             ColumnDescriptor col = cols[i];
-            xlcols.add(new MapArrayDisplayColumn(col.name, col.clazz, i));
+            xlcols.add(new ArrayDisplayColumn(col.name, col.clazz, i));
         }
 
         setDisplayColumns(xlcols);
@@ -37,17 +43,17 @@ public class ArrayExcelWriter extends ExcelWriter
         }
     }
 
-    public class MapArrayDisplayColumn extends DisplayColumn
+    public class ArrayDisplayColumn extends DisplayColumn
     {
         Class valueClass;
         int position;
 
-        public MapArrayDisplayColumn(String name, Class valueClass, int position)
+        public ArrayDisplayColumn(String name, Class valueClass, int position)
         {
             this(name, name, valueClass, position);
         }
 
-        public MapArrayDisplayColumn(String name, String caption, Class valueClass, int position)
+        public ArrayDisplayColumn(String name, String caption, Class valueClass, int position)
         {
             setName(name);
             setCaption(caption);
@@ -58,7 +64,7 @@ public class ArrayExcelWriter extends ExcelWriter
         @Override
         public Object getValue(RenderContext ctx)
         {
-            return data.get(currentRow).get(position);
+            return data.get(currentRow)[position];
         }
 
         @Override
--- a/api/src/org/labkey/api/data/ArrayExcelWriter.java
+++ b/api/src/org/labkey/api/data/ArrayExcelWriter.java
@@ -14,39 +14,37 @@
  * limitations under the License.
  */
 
-package org.labkey.api.study;
+package org.labkey.api.data;
 
 import org.apache.poi.ss.usermodel.Sheet;
-import org.labkey.api.data.ColumnInfo;
-import org.labkey.api.data.DisplayColumn;
-import org.labkey.api.data.ExcelColumn;
-import org.labkey.api.data.ExcelWriter;
-import org.labkey.api.data.RenderContext;
 import org.labkey.api.reader.ColumnDescriptor;
 import org.labkey.api.view.ActionURL;
 
 import java.io.Writer;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
 
 /**
  * User: Mark Igra
  * Date: May 4, 2007
  * Time: 3:08:34 PM
  */
-public class MapArrayExcelWriter extends ExcelWriter
+public class ArrayExcelWriter extends ExcelWriter
 {
-    private final List<Map<String,Object>> maps;
+    private final List<List<Object>> data;
     private int currentRow = 0;
 
-    public MapArrayExcelWriter(List<Map<String, Object>> maps, ColumnDescriptor[] cols)
+    public ArrayExcelWriter(List<List<Object>> data, ColumnDescriptor[] cols)
     {
         super(ExcelDocumentType.xls);
-        this.maps = maps;
+        this.data = data;
         List<DisplayColumn> xlcols = new ArrayList<>();
-        for (ColumnDescriptor col : cols)
-            xlcols.add(new MapArrayDisplayColumn(col.name, col.clazz));
+
+        for (int i = 0; i < cols.length; i++)
+        {
+            ColumnDescriptor col = cols[i];
+            xlcols.add(new MapArrayDisplayColumn(col.name, col.clazz, i));
+        }
 
         setDisplayColumns(xlcols);
     }
@@ -54,7 +52,7 @@ public class MapArrayExcelWriter extends ExcelWriter
     @Override
     public void renderGrid(RenderContext ctx, Sheet sheet, List<ExcelColumn> visibleColumns) throws MaxRowsExceededException
     {
-        for (currentRow = 0; currentRow < maps.size(); currentRow++)
+        for (currentRow = 0; currentRow < data.size(); currentRow++)
         {
             renderGridRow(sheet, ctx, visibleColumns);
         }
@@ -63,24 +61,26 @@ public class MapArrayExcelWriter extends ExcelWriter
     public class MapArrayDisplayColumn extends DisplayColumn
     {
         Class valueClass;
+        int position;
 
-        public MapArrayDisplayColumn(String name, Class valueClass)
+        public MapArrayDisplayColumn(String name, Class valueClass, int position)
         {
-            this(name, name, valueClass);
+            this(name, name, valueClass, position);
         }
 
-        public MapArrayDisplayColumn(String name, String caption, Class valueClass)
+        public MapArrayDisplayColumn(String name, String caption, Class valueClass, int position)
         {
             setName(name);
             setCaption(caption);
             this.valueClass = valueClass;
+            this.position = position;
         }
 
         @Override
         public Object getValue(RenderContext ctx)
         {
             //Ignore the context.
-            return maps.get(currentRow).get(getName());
+            return data.get(currentRow).get(position);
         }
 
         @Override
--- a/assay/src/org/labkey/assay/PlateController.java
+++ b/assay/src/org/labkey/assay/PlateController.java
@@ -43,10 +43,10 @@ import org.labkey.api.data.ArrayExcelWriter;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerFilter;
 import org.labkey.api.data.ContainerManager;
-import org.labkey.api.data.MapArrayExcelWriter;
 import org.labkey.api.gwt.client.model.GWTPropertyDescriptor;
 import org.labkey.api.gwt.server.BaseRemoteService;
 import org.labkey.api.query.FieldKey;
+import org.labkey.api.query.ValidationException;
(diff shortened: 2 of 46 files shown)


RosalineP:
Pending manual testing, as well as server-side and client-side unit tests

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2597,6 +2611,162 @@ public List<PlateManager.CreatePlateSetPlate> getPlates(Integer plateSetId, Cont
...
+        if (!prefix.equals("Destination "))
+            baseColumns.add(rs.getString(FKMap.get("sampleId")));
+
+        for (FieldKey col : includedMetaDataCols)
+            baseColumns.add(rs.getString(col));
+
+        return baseColumns;
+    }
+
+    public ColumnDescriptor[] getColumnDescriptors(String prefix, Set<FieldKey> includedMetadataCols)
```
RosalineP: Considered doing an enum or such for the prefix, but was waffling on it

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -166,6 +169,17 @@ public class PlateManager implements PlateService, AssayListener, ExperimentList
     private static final String PLATE_SET_NAME_EXPRESSION = "PLS-${now:date('yyyyMMdd')}-${RowId}";
     private static final String PLATE_NAME_EXPRESSION = "${${PlateSet/PlateSetId}-:withCounter}";
 
+    private final Map<String, FieldKey> FKMap = Map.of(
```
RosalineP: So this thing I've done is kind of weird. Hopefully by looking at the usages it's intelligible why I thought defining this constant might be useful—open to alternate structures, though!

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -1243,4 +1249,117 @@ public Object execute(PlateSetAssaysForm form, BindException errors) throws Exce
...
+        }
+
+        public void setDestinationPlateSetId(int destinationPlateSetId)
+        {
+            _destinationPlateSetId = destinationPlateSetId;
+        }
+    }
+
+    @RequiresPermission(ReadPermission.class)
+    public static class WorkListAction extends ReadOnlyApiAction<WorklistForm>
```
RosalineP, line 1282: (For both of these actions) I just saw that ExportAction exists—if that's the right class to use here I can update

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -1243,4 +1249,117 @@ public Object execute(PlateSetAssaysForm form, BindException errors) throws Exce
...
+    {
+        @Override
+        public Object execute(WorklistForm form, BindException errors) throws Exception
+        {
+            try
+            {
+                Set<FieldKey> sourceIncludedMetadataCols = PlateManager.get().getMetadataColumns(form.getSourcePlateSetId(), getContainer(), getUser());
```
RosalineP: Could possibly use some comments? Not totally sure how well this parses to the outside eye now

assay/src/org/labkey/assay/plate/PlateManager.java
labkey-klum: Consider making the map case insensitive

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2597,6 +2611,162 @@ public List<PlateManager.CreatePlateSetPlate> getPlates(Integer plateSetId, Cont
...
+            Set<FieldKey> destinationIncludedMetadataCols,
+            Container c,
+            User u
+    ) throws RuntimeSQLException, ValidationException
+    {
+        TableInfo wellTable = getWellTable(c, u);
+        List<List<Object>> plateDataRows = new ArrayList<>();
+        Map<String, List<List<Object>>> sampleIdToDestinationRow = new LinkedHashMap<>();
+
+        try (Results rs = QueryService.get().select(wellTable, getWellColumns(wellTable, destinationIncludedMetadataCols), new SimpleFilter(FKMap.get("plateSetId"), destinationPlateSetId), new Sort("rowid")))
```
labkey-klum: Are we relying on this sort?

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2597,6 +2611,162 @@ public List<PlateManager.CreatePlateSetPlate> getPlates(Integer plateSetId, Cont
...
+        {
+            while (rs.next())
+            {
+                String sampleId = rs.getString(FKMap.get("sampleId"));
+                if (sampleId == null)
+                    continue;
+
+                if (sampleIdToDestinationRow.containsKey(sampleId))
```
labkey-klum: would `computeIfAbsent` be more concise?

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2597,6 +2611,162 @@ public List<PlateManager.CreatePlateSetPlate> getPlates(Integer plateSetId, Cont
...
+                }
+            }
+        }
+        catch (SQLException e)
+        {
+            throw new RuntimeException(e);
+        }
+
+        Set<String> originSamples = new HashSet<>();
+        try (Results rs = QueryService.get().select(wellTable, getWellColumns(wellTable, sourceIncludedMetadataCols), new SimpleFilter(FKMap.get("plateSetId"), sourcePlateSetId), new Sort("rowid")))
```
labkey-klum: Keep in mind that sorting this way may not always guarantee row/column ordering. Consider either adding a sorter or add test coverage that will fail if this were ever to change.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2597,6 +2611,162 @@ public List<PlateManager.CreatePlateSetPlate> getPlates(Integer plateSetId, Cont
...
+            throw new ValidationException("Unable to resolve plate set of id " + plateSetId);
+
+        List<Plate> plates = plateSet.getPlates(u);
+
+        Set<FieldKey> includedMetadataCols = new HashSet<>();
+        for (Plate plate : plates)
+        {
+            List<String> metadataColNames = getFields(c, plate.getRowId()).stream().map(PlateCustomField::getName).collect(Collectors.toCollection(ArrayList::new));
+            for (String name : metadataColNames)
+                includedMetadataCols.add(FieldKey.fromParts("properties", name));
```
labkey-klum: The `well` table is responsible for how metadata is joined into the schema. To make this more maintainable, consider moving this to a static method on `WellTable`.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2597,6 +2611,162 @@ public List<PlateManager.CreatePlateSetPlate> getPlates(Integer plateSetId, Cont
...
+    {
+        List<Object> baseColumns = new ArrayList<>(
+            Arrays.asList(
+                rs.getString(FKMap.get("plateName")),
+                rs.getInt(FKMap.get("row")) * rs.getInt(FKMap.get("col")) + "-well",
+                rs.getString(FKMap.get("position"))
+            )
+        );
+
+        if (!prefix.equals("Destination "))
```
labkey-klum: The prefix should probably be converted to a string constant.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2597,6 +2611,162 @@ public List<PlateManager.CreatePlateSetPlate> getPlates(Integer plateSetId, Cont
         return plates;
     }
 
+    public Set<FieldKey> getMetadataColumns(int plateSetId, Container c, User u) throws ValidationException
```
labkey-klum: These new methods that were added to `PlateManager` seem very tailored to be useful to only worklist and instrument instruction generation and might be confusing to other users. Recommendation is to move these to another class and expose the methods needed from `PlateManager`, it doesn't look there is much needed additionally, maybe just `getWellTable`. Consider adding some comments to the code as well.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2597,6 +2611,162 @@ public List<PlateManager.CreatePlateSetPlate> getPlates(Integer plateSetId, Cont
...
+                }
+                else
+                {
+                    sampleIdToDestinationRow.put(sampleId, new ArrayList<>(Arrays.asList(getDataRow("Destination ", rs, destinationIncludedMetadataCols))));
+                }
+            }
+        }
+        catch (SQLException e)
+        {
+            throw new RuntimeException(e);
```
labkey-klum: In these cases you are almost always better off using : `UnexpectedException.wrap(e)` to improve the generated stack trace.

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -1243,4 +1249,117 @@ public Object execute(PlateSetAssaysForm form, BindException errors) throws Exce
...
+                ColumnDescriptor[] destinationXlCols = PlateManager.get().getColumnDescriptors("Destination ", destinationIncludedMetadataCols);
+                ColumnDescriptor[] xlCols = ArrayUtils.addAll(sourceXlCols, destinationXlCols);
+
+                List<List<Object>> plateDataRows = PlateManager.get().getWorklist(form.getSourcePlateSetId(), form.getDestinationPlateSetId(), sourceIncludedMetadataCols, destinationIncludedMetadataCols, getContainer(), getUser());
+
+                ArrayExcelWriter xlWriter = new ArrayExcelWriter(plateDataRows, xlCols);
+                PlateSet plateSetSource = PlateManager.get().getPlateSet(getContainer(), form.getSourcePlateSetId());
+                PlateSet plateSetDestination = PlateManager.get().getPlateSet(getContainer(), form.getDestinationPlateSetId());
+
+                if (plateSetSource == null || plateSetDestination == null)
```
labkey-klum, line 1291: Consider moving this check up (or in validateForm) before we do all the work to generate the data. Similar feedback for the `InstrumentInstructionAction`.

api/src/org/labkey/api/data/ArrayExcelWriter.java
```diff
@@ -0,0 +1,162 @@
...
+
+import org.apache.poi.ss.usermodel.Sheet;
+import org.labkey.api.reader.ColumnDescriptor;
+import org.labkey.api.view.ActionURL;
+
+import java.io.Writer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ArrayExcelWriter extends ExcelWriter
```
labkey-klum, line 13: Maybe some javadoc to describe what this class is and does. Is this really an array writer? It seems to accept a list of lists so just curious what prompted the name.

api/src/org/labkey/api/data/ArrayExcelWriter.java
```diff
@@ -0,0 +1,162 @@
...
+{
+    private final List<List<Object>> data;
```
labkey-klum: If you don't care whether the thing inside of each list entry is another list you could also use a data structure like : `List<Object[]>`.

api/src/org/labkey/api/data/ArrayExcelWriter.java
```diff
@@ -0,0 +1,162 @@
...
+    @Override
+    public void renderGrid(RenderContext ctx, Sheet sheet, List<ExcelColumn> visibleColumns) throws MaxRowsExceededException
+    {
+        for (currentRow = 0; currentRow < data.size(); currentRow++)
+        {
+            renderGridRow(sheet, ctx, visibleColumns);
+        }
+    }
+
+    public class MapArrayDisplayColumn extends DisplayColumn
```
labkey-klum: Doesn't seem related to a map array, perhaps copied from `MapArrayExcelWriter`. Consider renaming to something more accurate.

labkey-klum (changes requested):
please consider some of the suggested changes.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -166,6 +169,17 @@ public class PlateManager implements PlateService, AssayListener, ExperimentList
     private static final String PLATE_SET_NAME_EXPRESSION = "PLS-${now:date('yyyyMMdd')}-${RowId}";
     private static final String PLATE_NAME_EXPRESSION = "${${PlateSet/PlateSetId}-:withCounter}";
 
+    private final Map<String, FieldKey> FKMap = Map.of(
+            "sampleId", FieldKey.fromParts("sampleid", "name"),
+            "position", FieldKey.fromParts("position"),
```
RosalineP: define and use constants where possible on the table classes

assay/src/org/labkey/assay/plate/PlateManager.java
RosalineP: Yes—see other comment thread around line 2719!

assay/src/org/labkey/assay/plate/PlateManager.java
RosalineP: Yes, I remember you bringing this up earlier, and I chatted with Nick about it—we thought that in the eventuality we change row insertion into a plate to be sparse (such that we cannot rely on rowid for order anymore for row/columns) this will be another spot where we will have to update the code. I'll be sure to add some test coverage if it shakes out for the server-side unit tests, if not I'll pop this into the selenium test recommendations

api/src/org/labkey/api/data/ArrayExcelWriter.java
RosalineP: Added some documentation! I think I used arrays in an initial implementation and then switched to more lists. Now that I've updated so that `data` is `List<Object[]>` the name seems a little more reasonable

assay/src/org/labkey/assay/plate/PlateManager.java
RosalineP: Like so? (Also added some comments!)

assay/src/org/labkey/assay/plate/PlateManager.java
RosalineP: Elided this since I started using constants in the FKMap, good to know this exists though!

assay/src/org/labkey/assay/plate/PlateManager.java
RosalineP: @labkey-klum Take a look at my idea here for making this a bit nicer, let me know what you think... got tripped up on the lookup cols that didn't seem appropriate to put as constants on WellTable. The FieldKey parts are untouched because it felt like there would have to be a lot of making-into-undercase and bits where things don't quite align, like 'Row' being the const on WellTable but in FKMap within the lookup I need 'rows'

assay/src/org/labkey/assay/plate/PlateSetExport.java
```diff
@@ -0,0 +1,189 @@
...
+        return Stream.concat(baseColumns.stream(), metadataColumns.stream()).toList().toArray(new ColumnDescriptor[baseColumCount + includedMetadataCols.size()]);
+    }
+
+    public List<Object[]> getWorklist(
+            TableInfo wellTable,
+            int sourcePlateSetId,
+            int destinationPlateSetId,
+            Set<FieldKey> sourceIncludedMetadataCols,
+            Set<FieldKey> destinationIncludedMetadataCols
+    ) throws RuntimeSQLException, ValidationException
```
labkey-klum: minor: RuntimeSQLException is not a checked exception so you don't need to declare it here.

labkey-klum (approved):

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3075,5 +3086,116 @@ public void getWellSampleData()
...
+
+        @Test
+        public void getWorklist() throws Exception
+        {
+            // Arrange
+            final ExpMaterial sample1 = ExperimentService.get().createExpMaterial(container, sampleType.generateNextDBSeqLSID().toString(), "sampleA");
+            sample1.setCpasType(sampleType.getLSID());
+            sample1.save(user);
+            final ExpMaterial sample2 = ExperimentService.get().createExpMaterial(container, sampleType.generateNextDBSeqLSID().toString(), "sampleB");
+            sample2.setCpasType(sampleType.getLSID());
```
labkey-nicka, line 3153: nit: There is a comment in the `LineageTest` about not using `generateNextDbSeqLSID` in tests.

image

That said, it is used in tests all over:

image

Seems like either that comment needs to go away or there is actually a reason to not use it and use `generateSampleLSID().setObject(sampleName).toString()` instead.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2639,12 +2643,19 @@ public static void setupTest() throws Exception
...
+            props.add(new GWTPropertyDescriptor("col1", "string"));
+            props.add(new GWTPropertyDescriptor("col2", "string"));
+            sampleType = SampleTypeService.get().createSampleType(container, user, "SampleType1", null, props, emptyList(), 0, -1, -1, -1, null, null);
         }
 
         @After
         public void cleanupTest() throws Exception
         {
             PlateManager.get().deleteAllPlateData(container);
+            SampleTypeService.get().deleteSampleType(sampleType.getRowId(), container, user, null);
```
labkey-nicka: nit: Seems odd that the `PlateManager.TestCase` is individually cleaning up these structures. Other tests normally just call `JunitUtil.deleteTestContainer()` and then dereference `container` and `user`. Something like:

```java
@AfterClass
public static void cleanup()
{
    deleteTestContainer();
    _container = null;
    _user = null;
}
```

assay/src/org/labkey/assay/plate/PlateManager.java
RosalineP: Huh, interesting, good catch—not using generateNextDbSeqLSID works fine so I've updated to the generateSampleLSID() method

PR #5462, opened 2024-04-29
Issue 50132: importing folder archive with identical sample IDs in diferent sample types fails

XingY:
#### Rationale
The sample lookup columns in assay data are exported as sample names during folder export. When there are duplicate sample names across different samples types, the folder import fails with unable to resolve sample error. 

This PR modifies assay result data import to first try to resolve assay result samples names from already processed material inputs information (which was exported using lsid and hence resolvable). 

This PR does not address sample resolution for assay Run fields, which currently exports sample row id, and doesn't resolve (but doesn't error out either).

#### Related Pull Requests
- 

#### Changes
- try to use assay run's input material to resolve assay result samples during import


labkey-klum (approved):
That's a lot of code in `checkData`, it's unfortunate that assays don't use `DataIterators`.

PR #5464, opened 2024-04-30
Cache property descriptors on demand

labkey-adam:
#### Rationale
Property descriptor cache is the last one using a deprecated `DatabaseCache` constructor; convert to use the standard factory method.

Domain properties cache loader is doing a lot more work than it needs to, querying all the property descriptor columns and proactively filling that cache as well. Simplify its query to retrieve just PropertyURI and Required, which is all it returns.

### DIFF
--- a/api/src/org/labkey/api/data/DatabaseCache.java
+++ b/api/src/org/labkey/api/data/DatabaseCache.java
@@ -47,8 +47,8 @@ public class DatabaseCache<K, V> implements Cache<K, V>
     private final Cache<K, V> _sharedCache;
     private final DbScope _scope;
 
-    @Deprecated // Use the factory methods that return a BlockingDatabaseCache instead
-    public DatabaseCache(DbScope scope, int maxSize, long defaultTimeToLive, String debugName)
+    // Use the factory methods that return a BlockingDatabaseCache instead
+    private DatabaseCache(DbScope scope, int maxSize, long defaultTimeToLive, String debugName)
     {
         _sharedCache = createSharedCache(maxSize, defaultTimeToLive, debugName);
         _scope = scope;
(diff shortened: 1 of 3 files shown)


labkey-adam:
In addition to BVT, etc., the Daily suites all passed, except for a few failures appearing on `develop`:

* https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySqlServer/3008623?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildTestsSection=true&expandBuildChangesSection=true
* https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/3008633?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildTestsSection=true&expandBuildChangesSection=true

labkey-matthewb (approved):
This seems better, but I have questions...
Why are we caching this separately?  I mean I get that it's a different table from property descriptor (don't remember why), but what scenario are we accessing this separately from caching whole domain definitions? Seems too low-level.

labkey-jeckels (approved):

labkey-adam:
> This seems better, but I have questions... Why are we caching this separately? I mean I get that it's a different table from property descriptor (don't remember why), but what scenario are we accessing this separately from caching whole domain definitions? Seems too low-level.

I also don't like that we're caching these separately. I don't know the history. I didn't write any of this, I'm just trying to move it toward best practices. Since property descriptors can be shared between domains, the cached ones don't know sort order or required, so that also seems sketchy. There are 62 callers of the two public `OntologyManager.getPropertyDescriptor()` variants. I'd be happy to require the domain for retrieving and caching these, but it's a project.

PR #5466, opened 2024-04-30
include the sample ID for the well in the transform script result data.

labkey-klum:
#### Rationale
[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50276)

Make the sample ID from the well table available in the results file that is handed to the transform script. This is only available when the LKB plate experimental feature is enabled since it is the only way that samples can be plated into plates/plate sets.

### DIFF
--- a/assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
+++ b/assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
@@ -45,8 +45,10 @@ import org.labkey.api.exp.Lsid;
 import org.labkey.api.exp.OntologyManager;
 import org.labkey.api.exp.PropertyDescriptor;
 import org.labkey.api.exp.api.ExpData;
+import org.labkey.api.exp.api.ExpMaterial;
 import org.labkey.api.exp.api.ExpProtocol;
 import org.labkey.api.exp.api.ExpRun;
+import org.labkey.api.exp.api.ExperimentService;
 import org.labkey.api.exp.property.Domain;
 import org.labkey.api.exp.property.DomainKind;
 import org.labkey.api.exp.property.DomainProperty;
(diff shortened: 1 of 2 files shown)


assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -364,8 +364,12 @@ public List<Map<String, Object>> mergePlateMetadata(
 
                     if (positionToWell.containsKey(well))
                     {
-                        for (WellCustomField customField : PlateManager.get().getWellCustomFields(user, plate, positionToWell.get(well).getRowId()))
+                        WellBean wellBean = positionToWell.get(well);
+                        for (WellCustomField customField : PlateManager.get().getWellCustomFields(user, plate, wellBean.getRowId()))
                             row.put(customField.getName(), customField.getValue());
+
+                        // include the sample ID field from the well
+                        row.put("SampleID", wellBean.getSampleId());
```
cnathe: I tested this locally, and this does give me the sample RowId in the runDataFile.tsv for the transform script. But just as we include the Plate ID instead of the plate RowId, I was expecting the Sample ID (name). Would that be possible to include instead of the rowId?

cnathe (changes requested):

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -364,8 +376,17 @@ public List<Map<String, Object>> mergePlateMetadata(
 
                     if (positionToWell.containsKey(well))
                     {
-                        for (WellCustomField customField : PlateManager.get().getWellCustomFields(user, plate, positionToWell.get(well).getRowId()))
+                        WellBean wellBean = positionToWell.get(well);
+                        for (WellCustomField customField : PlateManager.get().getWellCustomFields(user, plate, wellBean.getRowId()))
                             row.put(customField.getName(), customField.getValue());
+
+                        // include the sample information from the well (Issue 50276)
```
cnathe, line 383: @labkey-klum in my local testing, I have the first set of wells in my plate empty (i.e. blanks). This means that the mergedRows for those didn't have sampleId and sampleName values. This resulted in the runDataFile not having those columns (likely because it was using the first row to get the column headers. 

I made a slight change with this commit: https://github.com/LabKey/platform/pull/5466/commits/3b778ee812f01ba505512ab30322f41969e5648e
Can you take a look?

cnathe (approved):

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
labkey-klum: Yes, looks great, thanks for making that change.

PR #5469, opened 2024-05-01
Issue 50190: Plate Well Position Label Update for Expected Sorting

labkey-nicka:
#### Rationale
This seeks to address [Issue 50190](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50190) by sorting by Well.RowId instead of Well.Position in the hit selection queries. This assumes that the rows are fully populated (as opposed to sparsely) in order in the well table.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/218

#### Changes
- Add a "sort column" of plate.Well.RowId for the plate.Well.Position column
- Add defensive creation of the plate metadata domain before inserting/updating rows.


labkey-klum (approved):
Makes sense, we'll just need to be aware if we move away from the well being fully populated on creation. I noticed this ordering assumption in Rosaline's worklist PR as well.

PR #5472, opened 2024-05-02
Migrate DbCache use to DatabaseCache

### LINKED ISSUE #5518: [LabKey CI] Eliminate DbCache use
`DbCache` uses a very inefficient invalidation mechanism (the entire cache is cleared on insert, update, or delete of any row in any container). It's time to migrate uses to a more standard and efficient caching pattern, and eliminate this class.

- [x] Migrate experiment runs cache DbCache -> DatabaseCache
- [x] Migrate participant cache DbCache -> DatabaseCache
- [x] Migrate StudyCache/QueryHelper DbCache -> DatabaseCache
- [x] Add validation checking (old vs. new return values) and logging
- [x] Fix failing tests and flagged validation problems
- [x] Code review - @labkey-jeckels 
- [x] Merge with validation in place
- [x] Fix tests (participant cache invalidation, delete specimen tables, bad insert button)
- [x] Code review 2
- [x] Merge test fixes
- [x] Remove validation code and DbCache
- [x] Code review 3
- [x] Merge final PR

labkey-adam:
#### Rationale
`DbCache` has been an anti-pattern for many years since it uses a very inefficient invalidation mechanism (the entire cache is cleared on every insert, update, or delete). It's time to eliminate it.

#### Changes
* Migrate experiment runs cache DbCache -> DatabaseCache
* Migrate participant cache DbCache -> DatabaseCache
* Migrate StudyCache/QueryHelper DbCache -> DatabaseCache
* Move various specimen methods study -> specimen. `deleteAllSpecimenData()` move is needed to allow explicit invalidation of specimen helpers; other moved methods aren't related to caching, but there's no reason for them to stay in study.
* Implement `equals()` and `hashCode()` on many classes to accommodate cache verification.

Note: This PR leaves DbCache and a large amount of validation & logging code in place. While many test suites have been run, my preference is to merge the verification code to develop and run all the tests with it in place for a few days, then follow up with a PR that removes DbCache, validation, logging, etc.


labkey-tchad:
Preparing review...

labkey-tchad:
Preparing review...

labkey-adam:
Daily suites are clear, except for two expected failures that appear in develop:

- https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/3036229?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true
- https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySqlServer/3036238?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildTestsSection=true

api/src/org/labkey/api/query/DefaultQueryUpdateService.java
```diff
@@ -334,7 +335,9 @@ protected Map<String, Object> _insert(User user, Container c, Map<String, Object
 
         try
         {
-            return Table.insert(user, getDbTable(), row);
+            Map<String, Object> ret = Table.insert(user, getDbTable(), row);
+            if (getDbTable().getName().equals("AssaySpecimen")) { DbCache.trackRemove(getDbTable()); }
```
labkey-jeckels, line 339: I assume these will be completely going away, and will be easy to remember since `DbCache` will be deleted and the compiler will flag them?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -298,6 +299,7 @@ public class ExperimentServiceImpl implements ExperimentService, ObjectReference
 
     private final Cache<String, ExpProtocolImpl> PROTOCOL_LSID_CACHE = DatabaseCache.get(getExpSchema().getScope(), CacheManager.UNLIMITED, CacheManager.HOUR, "Protocol by LSID",
         (key, argument) -> toExpProtocol(new TableSelector(getTinfoProtocol(), new SimpleFilter(FieldKey.fromParts("LSID"), key), null).getObject(Protocol.class)));
+    static final Cache<String, ExperimentRun> EXPERIMENT_RUN_CACHE = DatabaseCache.get(getExpSchema().getScope(), CacheManager.UNLIMITED, "Experiment Runs", new ExperimentRunCacheLoader());
```
labkey-jeckels: Unlimited seems too big for servers that have hundreds of thousands or millions of runs. Also consider adding a comment that the key is an LSID?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -3880,6 +3882,11 @@ public TableInfo getTinfoExperiment()
 
     @Override
     public TableInfo getTinfoExperimentRun()
+    {
+        return getTableInfoExperimentRun();
+    }
+
+    private static TableInfo getTableInfoExperimentRun()
```
labkey-jeckels: Do we need both of these? It is to have static and non-static variants?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -4141,6 +4148,35 @@ public List<ExpExperimentImpl> getExperiments(Container container, User user, bo
     }
 
     public ExperimentRun getExperimentRun(String LSID)
```
labkey-jeckels: I know the argument was originally `LSID` but can we make it `lsid`?

specimen/src/org/labkey/specimen/SpecimenServiceImpl.java
```diff
@@ -461,4 +469,73 @@ public void fireSpecimensChanged(Container c, User user, Logger logger)
...
+        SpecimenRequestManager.get().clearRequestStatusHelper(c);
+        assert set.add(SpecimenSchema.get().getTableInfoSampleRequestStatus());
+
+        new SpecimenTablesProvider(c, null, null).deleteTables();
+        LocationCache.clear(c);
+
+        Table.delete(SpecimenSchema.get().getTableInfoSampleAvailabilityRule(), containerFilter);
+        assert set.add(SpecimenSchema.get().getTableInfoSampleAvailabilityRule());
+
+        SpecimenMigrationService SMS = SpecimenMigrationService.get();
```
labkey-jeckels: `sms` instead of `SMS`

api/src/org/labkey/api/cache/CachingTestCase.java
```diff
@@ -2,79 +2,12 @@
...
-            key2 = "RowId" + rowId4;
-            DbCache.put(testTable, key2, umm);
-        }
-        m2 = (Map) DbCache.get(testTable, key2);
-        assertNull(m2);
-
-        // Clean up
-        Table.delete(testTable, rowId2);
-        Table.delete(testTable, rowId3);
+        DbCache.logUnmatched();
```
labkey-jeckels, line 11: Is this interesting? Wouldn't it need to do something to potentially get into an unmatched state?

labkey-jeckels (approved):

api/src/org/labkey/api/query/DefaultQueryUpdateService.java
labkey-adam: Correct. I have a local change that removes DbCache and all the validation code. I'll invite you to review that as well :)

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-adam: Switched to using TableInfo.getCacheSize(). This table doesn't specify a size, so it will default to 5,000 entries, as before. LMK if you think another size is more appropriate.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-adam: The cache & cache loader were defined statically and referenced outside of this class, so the static method was convenient. A small refactor to define these instance-level members and abstract the cache invalidation allowed for removing the static variant.

api/src/org/labkey/api/cache/CachingTestCase.java
labkey-adam: This was left as a convenient way to log all unmatched invalidations (Table-layer-invoked DbCache invalidation with no corresponding DatabaseCache explicit invalidation) that have occurred since the server started. This test will go away shortly.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-adam: Yes we can!

specimen/src/org/labkey/specimen/SpecimenServiceImpl.java
labkey-adam: Better yet, I eliminated these references to `SpecimenMigrationService` by moving the corresponding code into the specimen module.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-jeckels: That seems reasonable.

PR #5476, opened 2024-05-02
Export Plates to TSV, CSV, and Excel

labkey-alan:
#### Rationale
This PR adds a `PlateExportAction` to the `PlateController` which allows callers to export one or more plates as TSV, CSV, or Excel "map" files.

The CSV and TSV exports are in the same format as the Plate Editable Grid, one row per cell. The Excel "map" files are laid out in the format of the actual underlying plate, with a summary sheet that includes all data, a sheet for the Sample ID, and one sheet per Plate Custom Field.

*Note for reviewers*: with the changes to ExcelColumn in this PR I am concerned that some regressions may have been introduced. I am currently unsure of what our test coverage looks like for Excel export, so an especially eagle eyed approach would be appreciated when reviewing this PR.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5476
- https://github.com/LabKey/labkey-ui-premium/pull/407
- https://github.com/LabKey/limsModules/pull/229

#### Changes
- Add PlateExportAction
- Add `exportPlateData` and `exportPlateMaps` to PlateManager
- Add `PlateMapExcelWriter`, an ExcelWriter that exports plates to excel in their "map" form
- Add `ExcelCellUtils`, a helper class containing methods for writing properly formatted values to Excel cells
  - The majority of the code in this class was directly extracted from `ExcelColumn` so it could be re-used in `PlateMapExcelWriter`
- Refactor `ExcelColumn` to use `ExcelCellUtils`

### DIFF
--- a/api/src/org/labkey/api/data/ExcelCellUtils.java
+++ b/api/src/org/labkey/api/data/ExcelCellUtils.java
@@ -15,6 +15,11 @@ import java.text.Format;
 import java.util.Date;
 import java.util.GregorianCalendar;
 
+/**
+ * This is a utility class that contains the necessary methods for writing properly formatted values to Excel cells.
+ * This logic was previously housed directly in ExcelColumn, however with the introduction of PlateMapExcelWriter we
+ * needed a shared location for rendering formatted values to Excel cells.
+ */
 public class ExcelCellUtils
 {
     public static final int TYPE_UNKNOWN = 0;
--- a/api/src/org/labkey/api/data/ExcelCellUtils.java
+++ b/api/src/org/labkey/api/data/ExcelCellUtils.java
@@ -1,0 +1,213 @@
+package org.labkey.api.data;
+
+import jakarta.annotation.Nullable;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.time.FastDateFormat;
+import org.apache.poi.ss.usermodel.Cell;
+import org.apache.poi.ss.usermodel.CellStyle;
+import org.apache.poi.ss.usermodel.Workbook;
+import org.labkey.api.util.DateUtil;
+
+import java.io.File;
+import java.math.BigDecimal;
+import java.sql.Time;
+import java.text.Format;
+import java.util.Date;
+import java.util.GregorianCalendar;
+
+public class ExcelCellUtils
+{
+    public static final int TYPE_UNKNOWN = 0;
+    public static final int TYPE_INT = 1;
+    public static final int TYPE_DOUBLE = 2;
+    public static final int TYPE_STRING = 3;
+    public static final int TYPE_MULTILINE_STRING = 4;
+    public static final int TYPE_DATE = 5;
+    public static final int TYPE_BOOLEAN = 6;
+    public static final int TYPE_FILE = 7;
+    public static final int TYPE_TIME = 8;
+
+    private static final Date EXCEL_DATE_0 = (new GregorianCalendar(1900, 1, 1)).getTime();
+
+    public static int getSimpleType(DisplayColumn dc)
+    {
+        Class valueClass = dc.getDisplayValueClass();
+
+        if (Integer.class.isAssignableFrom(valueClass) || Integer.TYPE.isAssignableFrom(valueClass) ||
+                Long.class.isAssignableFrom(valueClass) || Long.TYPE.isAssignableFrom(valueClass) ||
+                Short.class.isAssignableFrom(valueClass) || Short.TYPE.isAssignableFrom(valueClass))
+            return TYPE_INT;
+        else if (Float.class.isAssignableFrom(valueClass) || Float.TYPE.isAssignableFrom(valueClass) ||
+                Double.class.isAssignableFrom(valueClass) || Double.TYPE.isAssignableFrom(valueClass) ||
+                BigDecimal.class.isAssignableFrom(valueClass))
+            return TYPE_DOUBLE;
+        else if (String.class.isAssignableFrom(valueClass))
+        {
+            if (dc.getColumnInfo() != null && dc.getColumnInfo().getInputRows() > 1)
+            {
+                return TYPE_MULTILINE_STRING;
+            }
+            return TYPE_STRING;
+        }
+        else if (Date.class.isAssignableFrom(valueClass))
+        {
+            if (Time.class.isAssignableFrom(valueClass))
+                return TYPE_TIME;
+            else
+                return TYPE_DATE;
+        }
+        else if (Boolean.class.isAssignableFrom(valueClass) || Boolean.TYPE.isAssignableFrom(valueClass))
+            return TYPE_BOOLEAN;
+        else if (File.class.isAssignableFrom(valueClass))
+            return TYPE_FILE;
+        else
+        {
+            return TYPE_UNKNOWN;
+        }
+    }
+
+    public static String getFormatString(int simpleType, @Nullable String formatString)
+    {
+        if (formatString != null)
+        {
+            if (simpleType == TYPE_DATE || simpleType == TYPE_TIME)
+            {
+                formatString = formatString.replaceAll("aa", "a").replaceAll("a", "AM/PM");
+            } else if (simpleType == TYPE_INT || simpleType == TYPE_DOUBLE)
+            {
+                // Excel has a different idea of how to represent scientific notation, so be sure that we
+                // transform the Java format if needed.
+                // https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=17735
+                formatString = formatString.replaceAll("[eE][^\\+]", "E+0");
+            }
+
+            return formatString;
+        }
+
+        switch (simpleType)
+        {
+            case(TYPE_DATE):
+                return DateUtil.getStandardDateFormatString();
+            case(TYPE_TIME):
+                return DateUtil.getStandardTimeFormatString();
+            case(TYPE_INT):
+                return "0";
+            case(TYPE_DOUBLE):
+                return "0.0000";
+            default:
+                return null;
+        }
+    }
+
+    public static CellStyle createCellStyle(Workbook workbook, int simpleType, @Nullable String formatString)
+    {
+        if (formatString == null)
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/data/ExcelCellUtils.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    public static String getFormatString(int simpleType, @Nullable String formatString)
+    {
+        if (formatString != null)
+        {
+            if (simpleType == TYPE_DATE || simpleType == TYPE_TIME)
+            {
+                formatString = formatString.replaceAll("aa", "a").replaceAll("a", "AM/PM");
+            } else if (simpleType == TYPE_INT || simpleType == TYPE_DOUBLE)
```
labkey-nicka: nit: Formatting

api/src/org/labkey/api/data/ExcelCellUtils.java
```diff
@@ -0,0 +1,218 @@
...
+                // Excel has a different idea of how to represent scientific notation, so be sure that we
+                // transform the Java format if needed.
+                // https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=17735
+                formatString = formatString.replaceAll("[eE][^\\+]", "E+0");
+            }
+
+            return formatString;
+        }
+
+        switch (simpleType)
```
labkey-nicka: nit: Consider enhanced "switch":

```java
return switch (simpleType)
{
    case (TYPE_DATE) -> DateUtil.getStandardDateFormatString();
    case (TYPE_TIME) -> DateUtil.getStandardTimeFormatString();
    case (TYPE_INT) -> "0";
    case (TYPE_DOUBLE) -> "0.0000";
    default -> null;
};
```

api/src/org/labkey/api/data/ExcelCellUtils.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    public static CellStyle createCellStyle(Workbook workbook, int simpleType, @Nullable String formatString)
+    {
+        if (formatString == null)
+            formatString = getFormatString(simpleType, formatString);
+
+        CellStyle style = workbook.createCellStyle();
+
+        switch (simpleType)
```
labkey-nicka: nit: Consider enhanced "switch":

```java
return switch (simpleType)
{
    case (TYPE_INT), (TYPE_DOUBLE), (TYPE_DATE), (TYPE_TIME) ->
    {
        short formatIndex = workbook.createDataFormat().getFormat(formatString);
        style.setDataFormat(formatIndex);
        yield style;
    }
    case (TYPE_MULTILINE_STRING) ->
    {
        style.setWrapText(true);
        yield style;
    }
    default -> null;
};
```

api/src/org/labkey/api/data/ExcelCellUtils.java
```diff
@@ -0,0 +1,218 @@
...
+    public static final int TYPE_INT = 1;
+    public static final int TYPE_DOUBLE = 2;
+    public static final int TYPE_STRING = 3;
+    public static final int TYPE_MULTILINE_STRING = 4;
+    public static final int TYPE_DATE = 5;
+    public static final int TYPE_BOOLEAN = 6;
+    public static final int TYPE_FILE = 7;
+    public static final int TYPE_TIME = 8;
+
+    private static final Date EXCEL_DATE_0 = (new GregorianCalendar(1900, 1, 1)).getTime();
```
labkey-nicka: I realize you are just copying this over from `ExcelColumn` but the month in `GregorianCalendar` is zero-based. To get what we want we could use the constant provided:

```java
new GregorianCalendar(1900, Calendar.JANUARY, 1)
```

Also, the original constant in `ExcelColumn` is no longer referenced and can be removed.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2621,6 +2635,146 @@ public List<Object[]> getInstrumentInstructions(int plateSetId, Set<FieldKey> in
...
+                }
+
+                files.add(tempFile);
+            }
+        }
+
+        return files;
+    }
+
+    public List<File> exportPlateMaps(Container c, User user, ContainerFilter cf, List<Integer> plateIds, ViewContext viewContext)  throws Exception
```
labkey-nicka: The `ViewContext` parameter is unused and can be removed.

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -1362,4 +1371,152 @@ public Object execute(InstrumentInstructionForm form, BindException errors) thro
...
+            if (form.getExportType() == PlateExportType.CSV)
+                files = PlateManager.get().exportPlateData(getContainer(), getUser(), cf, form.getPlateIds(), TSVWriter.DELIM.COMMA);
+            else if (form.getExportType() == PlateExportType.TSV)
+                files = PlateManager.get().exportPlateData(getContainer(), getUser(), cf, form.getPlateIds(), TSVWriter.DELIM.TAB);
+            else
+                files = PlateManager.get().exportPlateMaps(getContainer(), getUser(), cf, form.getPlateIds(), getViewContext());
+
+            if (files.isEmpty())
+            {
+                // TODO: what does the empty response look like? Should it be an error?
```
labkey-nicka: I think `null` is fine.

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -1362,4 +1371,152 @@ public Object execute(InstrumentInstructionForm form, BindException errors) thro
...
+                Files.deleteIfExists(file.toPath());
+                return null;
+            }
+
+            String filename = form.getFilename();
+
+            if (filename == null)
+                filename = "plates.zip";
+            else
+                filename = filename + ".zip";
```
labkey-nicka: This should be wrapped in a call to `FileUtil.makeLegalName()`. Consider:

```java
String filename = form.getFilename();
if (filename == null)
    filename = "plates";
filename = FileUtil.makeLegalName(filename + ".zip");
```

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -1362,4 +1371,152 @@ public Object execute(InstrumentInstructionForm form, BindException errors) thro
...
+        public Object execute(PlateExportForm form, BindException errors) throws Exception
+        {
+            ContainerFilter cf = ContainerFilter.Type.Current.create(getViewContext());
+            // if an optional container filter is specified
+            if (form.getContainerFilter() != null)
+                cf = form.getContainerFilter().create(getViewContext());
+
+            List<File> files;
+
+            if (form.getExportType() == PlateExportType.CSV)
```
labkey-nicka, line 1513: We may want to limit the number of plates that can be requested to a reasonable size. Seems like `PlateSet.MAX_PLATES` could be a good starting number.

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -1362,4 +1371,152 @@ public Object execute(InstrumentInstructionForm form, BindException errors) thro
...
+                files = PlateManager.get().exportPlateData(getContainer(), getUser(), cf, form.getPlateIds(), TSVWriter.DELIM.COMMA);
+            else if (form.getExportType() == PlateExportType.TSV)
```
labkey-nicka, line 1518: Speaking of size have you tried this with exporting a full plate set worth of plates? What kind of performance times are you seeing?

assay/api-src/org/labkey/api/assay/plate/PlateMapExcelWriter.java
```diff
@@ -0,0 +1,364 @@
...
+import org.labkey.api.view.ActionURL;
+import org.labkey.api.view.HttpView;
+
+import java.io.IOException;
+import java.io.Writer;
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
```
labkey-nicka: Unused import

api/src/org/labkey/api/data/ExcelColumn.java
```diff
@@ -120,7 +106,7 @@ public class ExcelColumn extends RenderColumn
     private static final Date EXCEL_DATE_0 = (new GregorianCalendar(1900, 1, 1)).getTime();
 
     // CONSIDER: Add support for left/right/center alignment (from DisplayColumn)
-    private int _simpleType = TYPE_UNKNOWN;
+    private int _simpleType = ExcelCellUtils.TYPE_UNKNOWN;
```
labkey-nicka, line 106: There are at least some integration tests already in `ExcelColumn`. Consider adding additional cases to ensure coverage of the methods you've extracted into `ExcelCellUtils`.

assay/src/org/labkey/assay/PlateController.java
labkey-alan: I've tested with a few different plate sets and haven't hit anything that seemed unbearable, but I'm not really sure what the average plate set size will be.

api/src/org/labkey/api/data/ExcelColumn.java
labkey-alan: If I am interpreting the existing test cases correctly I believe they should cover all of the extracted methods as-is

assay/src/org/labkey/assay/PlateController.java
labkey-alan: I think if we find this to be slower than we want we could consider rendering a modal that shows progress like we do in other parts of our app (e.g. ELN export). We could do that as part of adding the export functionality to the Plateset page.

api/src/org/labkey/api/data/ExcelCellUtils.java
```diff
@@ -0,0 +1,211 @@
+package org.labkey.api.data;
+
+import jakarta.annotation.Nullable;
```
XingY: Use `import org.jetbrains.annotations.Nullable;` instead to be consistent with other usages?

XingY (approved):

labkey-nicka (approved):

PR #5479, opened 2024-05-02
query recursion check doesn't clean up properly when it throws.

labkey-matthewb:
#### Rationale
Customer started hitting the recursion limit check inconsistently with the same query.  That behavior is consistent with the discovered bug.

Are we still hot-fixing 23.11?

#### Related Pull Requests
- 

#### Changes
-


labkey-tchad (commented):
We could do a hotfix if release triage approves it. Do you have an issue number?

labkey-matthewb:
> We could do a hotfix if release triage approves it. Do you have an issue number?

Sorry, didn't have bug yet when I created the PR.  There is a 24.3 version as well.

PR #5480, opened 2024-05-03
Issue 50212: ScriptsResourceProvider.resolve to catch AccessDeniedException for @scripts dir

cnathe:
#### Rationale
See related PR for rationale. This is a second attempt to fix the issue. This PR updates ScriptsResourceProvider.resolve to catch AccessDeniedException for @scripts dir attempted creation.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5453

#### Changes
- ScriptsResourceProvider.resolve to catch AccessDeniedException


labkey-jeckels (approved):
This seems good as a 24.3 path. Longer term it would be nice to not rely on the RuntimeException wrapping the underlying cause. Perhaps we should make more of these methods throw IOException so callers know that problems might pop up and deal with them as needed.

PR #5481, opened 2024-05-03
Issues 50253 and 50195: Add missing import aliases for Sample ID and SampleID

labkey-susanh:
#### Rationale
See related PR

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/220

#### Changes
- Make sure renamed column is in `columnInfoMap` before trying to use it.


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -4171,6 +4172,8 @@ protected Map<String, String> getRenamedColumns()
 
                 renameColumns.put(paramName.substring(renameParamPrefix.length()), (String) pv.getValue());
             }
+            if (_target instanceof ExpMaterialTable)
```
XingY: Instead of here, how about `ImportSamplesAction` override `getRenamedColumns` method?

labkey-susanh: Yes, of course that's better.

XingY (approved):

PR #5488, opened 2024-05-03
Avoid inline script for ExtJS label tooltips

labkey-jeckels:
#### Rationale
We can't use `href="javascript:void(0)"` with a strict CSP so we need a new strategy for ExtJS-based tooltips. And we want to show tooltips on top of all open ExtJS windows/dialogs.

#### Changes
- Use newer tooltip approach for ExtJS labels
- Fix some HTML encoding problems

### DIFF
--- a/core/webapp/util.js
+++ b/core/webapp/util.js
@@ -47,13 +47,13 @@ function getHelpDiv()
     return _helpDiv;
 }
 
-function showHelpDivDelay(elem, titleText, bodyText, width, delay)
+function showHelpDivDelay(elem, titleText, bodyText, width, delay, zIndex)
 {
     // IE support
-    _showTimer = setTimeout(function() { showHelpDiv(elem, titleText, bodyText, width); }, delay ? delay : 400);
+    _showTimer = setTimeout(function() { showHelpDiv(elem, titleText, bodyText, width, zIndex); }, delay ? delay : 400);
 }
 
-function showHelpDiv(elem, titleText, bodyHtml, width)
+function showHelpDiv(elem, titleText, bodyHtml, width, zIndex)
 {
     var posLeft = 12;
     var posTop = 8;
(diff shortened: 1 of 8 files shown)


core/webapp/util.js
```diff
@@ -47,13 +47,13 @@ function getHelpDiv()
     return _helpDiv;
 }
 
-function showHelpDivDelay(elem, titleText, bodyText, width, delay)
+function showHelpDivDelay(elem, titleText, bodyText, width, delay, zIndex)
 {
     // IE support
-    _showTimer = setTimeout(function() { showHelpDiv(elem, titleText, bodyText, width); }, delay ? delay : 400);
+    _showTimer = setTimeout(function() { showHelpDiv(elem, titleText, bodyText, width, zIndex); }, delay ? delay : 400);
```
labkey-adam, line 53: Still needed? "IE support"?

wiki/src/org/labkey/wiki/view/customizeWiki.jsp
```diff
@@ -110,7 +110,7 @@ function disableSubmit()
     var btn = document.getElementById("btnSubmit");
     btn.disabled = true;
     btn.className = "labkey-disabled-button primary";
-    btn.href = "javascript:return false;";
+    btn.href = "#";
```
labkey-adam, line 113: `#` vs. `return false` behave differently in some cases (refresh/navigate vs. no navigate). But probably fine.

labkey-adam (approved):
Consider updating "IE support" comment or code

wiki/src/org/labkey/wiki/view/customizeWiki.jsp
labkey-jeckels: I couldn't get the CSP to error on this one but decided to be conservative and keep an href

PR #5489, opened 2024-05-06
Issue 49989: "Show metric in plots, but don't identify outliers" should not display bound lines

ankurjuneja:
#### Rationale
Check new plot config property to display bound lines

#### Related Pull Requests
- https://github.com/LabKey/targetedms/pull/896

### DIFF
--- a/core/webapp/vis/src/plot.js
+++ b/core/webapp/vis/src/plot.js
@@ -2215,7 +2215,9 @@ boxPlot.render();
                 config.layers = [];
 
                 if (config.properties.mean !== undefined) {
-                    if (config.properties.stdDev !== undefined && config.properties.boundType === LABKEY.vis.PlotProperties.BoundType.StandardDeviation) {
+                    if (config.properties.stdDev !== undefined &&
+                            config.properties.boundType === LABKEY.vis.PlotProperties.BoundType.StandardDeviation &&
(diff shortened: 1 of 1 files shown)


core/webapp/vis/src/plot.js
```diff
@@ -2215,7 +2215,9 @@ boxPlot.render();
                 config.layers = [];
 
                 if (config.properties.mean !== undefined) {
-                    if (config.properties.stdDev !== undefined && config.properties.boundType === LABKEY.vis.PlotProperties.BoundType.StandardDeviation) {
+                    if (config.properties.stdDev !== undefined &&
+                            config.properties.boundType === LABKEY.vis.PlotProperties.BoundType.StandardDeviation &&
+                            config.properties.showBoundLines) {
```
labkey-jeckels, line 2222: This won't be backwards compatible for non-Panorama use cases. We need to make sure that `showBoundLines` defaults to true in the absence of the property.

labkey-jeckels (changes requested):

core/webapp/vis/src/plot.js
ankurjuneja: done

labkey-jeckels (approved):

PR #5490, opened 2024-05-06
Fix group membership cache invalidation

labkey-adam:
#### Rationale
Repeated runs of the DRT suite failed, pointing out an issue with group membership cache invalidation. Fix this and improve the test to fail faster.

Repeated runs also failed due to discrepancies between the `deleteGroups()` and `deleteGroup()` implementations. (The latter didn't invalidate most of the caches.) deleteGroups() now simply iterates the groups and calls deleteGroup() on each, for consistency.

Enhance `PrincipalArray` with more convenience methods to reduce use of internal array.

Move some long-running tests (>10s) from DRT to BVT.

### DIFF
--- a/api/src/org/labkey/api/security/SecurityManager.java
+++ b/api/src/org/labkey/api/security/SecurityManager.java
@@ -1676,9 +1676,8 @@ public class SecurityManager
         {
             int id = recurse.removeFirst();
             groupSet.add(id);
-            int[] groups = GroupMembershipCache.getGroupMemberships(id).getPrincipals();
 
-            for (int g : groups)
+            for (int g : GroupMembershipCache.getGroupMemberships(id))
             {
                 if (g == root.getUserId())
                     return true;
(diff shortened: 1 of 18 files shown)


api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -1503,6 +1498,8 @@ static void deleteGroup(int groupId)
                 ProjectAndSiteGroupsCache.uncache(c);
                 // 20329 SecurityPolicy cache still has the role assignments for deleted groups.
                 SecurityPolicyManager.notifyPolicyChanges(resources);
+                // Need to invalidate all computed group lists. This isn't quite right, but it gets the job done.
```
labkey-jeckels: I know this is a preexisting comment, but if you know why this isn't quite right, it would be nice to clarify.

labkey-jeckels (approved):

api/src/org/labkey/api/security/SecurityManager.java
labkey-adam: Updated... I don't think there's a real problem here

PR #5503, opened 2024-05-08
Issue 50315: graal issue causing markdown text to not render

labkey-jeckels:
#### Rationale
When Graal isn't available and we can't render Markdown, we don't want to let the server keep running

#### Changes
- In prod mode, fail the startup if Graal isn't available. Continue to be tolerant of this in dev mode


labkey-adam:
Would it be better to simply not allow webapp reloading in production mode? I thought that used to be the case.

labkey-jeckels:
We shouldn't be allowing reloads on production servers with embedded Tomcat via the lack of `spring.devtools` values in `application.properties`

We could disallow this for standalone too by making sure that `ROOT.xml` doesn't have `reloadable="true"` in it, but I don't want to spend time on standalone-specific updates at this point.

This change is also useful for cases like we saw on trial instances, where Graal failed to extract the `.so` and therefore couldn't load the JNI library on its first try either.

labkey-susanh (approved):

PR #5520, opened 2024-05-21
Make PrincipalArray immutable

labkey-adam:
#### Rationale
Caching and stashing immutable classes is strongly preferred. `PrincipalArray` previously exposed its array of principal IDs to other security classes. None of that code has been changing the array recently, but that wasn't obvious and, regardless, it's dangerous and unnecessary. This PR converts callers to use an immutable list of principal IDs and removes external access to the array.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5490

### DIFF
--- a/api/src/org/labkey/api/security/PrincipalArray.java
+++ b/api/src/org/labkey/api/security/PrincipalArray.java
@@ -1,7 +1,6 @@
 package org.labkey.api.security;
 
 import com.google.common.primitives.Ints;
-import org.apache.commons.collections4.iterators.ArrayIterator;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/security/PrincipalArray.java
```diff
@@ -12,69 +11,67 @@
...
-        return new ArrayIterator<>(_principals);
+        return _list.iterator();
     }
 
-    // Returns an immutable list that wraps the principals array
+    // Returns an immutable list of principals
     public List<Integer> getList()
     {
-        return Ints.asList(_principals);
+        return _list;
```
labkey-jeckels, line 74: I haven't tested, but it looks like `Ints.asList()` returns a `IntArrayAsList` which has an implementation of `set()`. Should we hold an immutable wrapper instead?

labkey-jeckels (approved):

api/src/org/labkey/api/security/PrincipalArray.java
labkey-adam: Yes, looks like it's not immutable. It's fixed size (add() and remove() throw), but values can be set(). I've wrapped it in the constructor.

PR #5521, opened 2024-05-21
Allow authentication and email change with an invalid address

labkey-adam:
#### Rationale
Users can end up with invalid email addresses saved in their accounts. Overly aggressive validation blocks 1) admins from changing these to valid addresses and 2) users from logging in and executing the self-change workflow. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=50188

Also, rewrite the validation code for self-serve email changes.

### DIFF
--- a/api/src/org/labkey/api/security/SecurityManager.java
+++ b/api/src/org/labkey/api/security/SecurityManager.java
@@ -1013,7 +1013,6 @@ public class SecurityManager
         }
     }
 
-
     public static class UserManagementException extends Exception
     {
         private final String _email;
@@ -1337,8 +1336,8 @@ public class SecurityManager
         return getPasswordHash(email.getEmailAddress());
     }
 
-    // Look up email in Logins table and return the corresponding password hash
-    private static String getPasswordHash(String email)
+    // For internal use only, plus database login and change email workflows, where existing email could be invalid.
+    public static String getPasswordHash(String email)
     {
         SqlSelector selector = new SqlSelector(core.getSchema(), new SQLFragment("SELECT Crypt FROM " + core.getTableInfoLogins() + " WHERE Email = ?", email));
         return selector.getObject(String.class);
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/security/UserManager.java
```diff
@@ -338,12 +337,19 @@ public static void addCustomButtons(UserDetailsButtonCategory category, ButtonBa
...
     }
 
-
     public static @Nullable User getUser(ValidEmail email)
     {
         return UserCache.getUser(email);
     }
 
+    // Use only for database login when existing email is invalid
+    public static @Nullable User getUser(String email)
```
labkey-matthewb: If this is a not a "normal" code path, maybe it should have a name that doesn't look like a regular call?  E.g. getUserForInvalidEmail() or something?

api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -1337,8 +1336,8 @@ public static String getPasswordHash(ValidEmail email)
         return getPasswordHash(email.getEmailAddress());
     }
 
-    // Look up email in Logins table and return the corresponding password hash
-    private static String getPasswordHash(String email)
+    // For internal use only, plus database login and change email workflows, where existing email could be invalid.
```
labkey-matthewb: The use of invalid in the comments might a bit ambiguous to a casual reader?  I think we simply mean not convertible to a ValidEmail as opposed to invalid/wrong in any other way.  Maybe "non-conformant".  A nit obviously.

labkey-matthewb (approved):

api/src/org/labkey/api/security/UserManager.java
labkey-adam: Better yet, I deleted the method. I moved the implementation into `DbLoginProvider` which, after some churn, was left as the only caller.

api/src/org/labkey/api/security/SecurityManager.java
labkey-adam: Clarified

PR #5523, opened 2024-05-21
Update deduping code for warn message.  Ignore x-amzn-trace-id.

labkey-matthewb:
#### Rationale
Reduce noise in server logs.

#### Related Pull Requests
-

### DIFF
--- a/api/src/org/labkey/api/security/AuthenticatedRequest.java
+++ b/api/src/org/labkey/api/security/AuthenticatedRequest.java
@@ -469,6 +469,7 @@ public class AuthenticatedRequest extends HttpServletRequestWrapper implements A
 
     private static void _logGuestSession(String ip, String msg)
     {
+        /* try to not flood the log with repeat messages if we have a badly behaved client (e.g. not remembering JSESSIONID) */
         ip = StringUtils.defaultString(ip,"unknown");
         String prevMessage = logMessages.get(ip);
         if (StringUtils.equals(prevMessage, msg))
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/security/AuthenticatedRequest.java
```diff
@@ -465,7 +465,7 @@ HttpSession guestSession()
     }
 
     // helper to avoid filling the _log
-    private static final Cache<String, String> logMessages = CacheManager.getCache(100, TimeUnit.MINUTES.toMillis(1), "Guest session messages");
+    private static final Cache<String, String> logMessages = CacheManager.getCache(100, TimeUnit.MINUTES.toMillis(5), "Guest session messages");
 
     private static void _logGuestSession(String ip, String msg)
```
labkey-jeckels, line 470: Is the IP in this scenario the IP of the client or the ALB?

labkey-jeckels (approved):
Seems like we don't need to change the log level with this change? It's hard for me follow the intent of the code in terms of logging. Maybe a comment or two to help explain?

api/src/org/labkey/api/security/AuthenticatedRequest.java
labkey-matthewb: Good question...

labkey-matthewb: If there is an x-forwarded-for header we should probably use that.  In this case, it looks like there is no x-forwarded-for and the logged remote ips look like routable addresses.

PR #5526, opened 2024-05-22
Issue 50354: LKSM: Include indicator of unavailable file in apps

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1500
- https://github.com/LabKey/platform/pull/5526
- https://github.com/LabKey/limsModules/pull/298
- https://github.com/LabKey/testAutomation/pull/1946

#### Changes
- Append " (unavailable)" suffix for file displayValues for missing files


api/src/org/labkey/api/data/AbstractFileDisplayColumn.java
```diff
@@ -77,9 +79,10 @@ protected boolean isImage(String filename)
 
     protected void renderIconAndFilename(RenderContext ctx, Writer out, String filename, @Nullable String fileIconUrl, @Nullable String popupIconUrl, boolean link, boolean thumbnail) throws IOException
     {
-        if (null != filename)
+        if (null != filename && !StringUtils.isEmpty(filename))
         {
-            String url = renderURL(ctx);
+            StringExpression s = compileExpression(ctx.getViewContext());
+            String url = null == s ? null : s.eval(ctx);
```
cnathe: I was confused by this change at first. However, looking at IntelliJ, it looks like you are effectively skipping over the renderURL override in DataColumn.java and reverting back to the one in DisplayColumn.java. Is that correct? If so, might be worth a comment here.

api/src/org/labkey/api/study/assay/FileLinkDisplayColumn.java
```diff
@@ -255,6 +255,9 @@ protected String getFileName(RenderContext ctx, Object value)
             {
                 result = f.getName();
             }
+
+            if (!f.exists())
+                result += " (unavailable)";
```
cnathe, line 265: it looks like I am getting double "(unavailable)" messages in the LKS side now with this change. This class already adds the suffix at line 328. is that not enough to detect on the client side for the app?
Screenshot 2024-05-23 at 1 58 16 PM

cnathe (changes requested):

cnathe (approved):

api/src/org/labkey/api/data/AbstractFileDisplayColumn.java
XingY: Yep, comment added

api/src/org/labkey/api/study/assay/FileLinkDisplayColumn.java
XingY: good catch. should be fixed now. The LKS side uses a different code path for display column to add the suffix.The client side needs getQuery api displayValues changed.

PR #5531, opened 2024-05-24
Issue 50380: LKSM: Explore adding a metric to provide number of unavailable files on a server

XingY:
#### Rationale
Issue [50380](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50380)
A new system maintenance task is added to query for missing files from file domain fields, which is then added to the metrics report. The metrics gives a summary stats of missing files per container per datatype. A new api action is also added to allow detailed query of which files are missing for a specific container. 

#### Related Pull Requests
- 

#### Changes
-


experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
```diff
@@ -0,0 +1,79 @@
...
+import org.labkey.api.usageMetrics.UsageMetricsService;
+import org.labkey.api.util.SystemMaintenance;
+import org.labkey.experiment.api.ExperimentServiceImpl;
+
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class FileLinkMetricsMaintenanceTask implements SystemMaintenance.MaintenanceTask
```
XingY, line 21: Should this system task be disabled by default?

experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
```diff
@@ -0,0 +1,79 @@
...
+                        String containerPath = containerId;
+                        Container container = ContainerManager.getForId(containerId);
+                        if (container != null)
+                            containerPath = container.getPath();
+                        Map<String, Object> containerMetrics = new HashMap<>();
+                        missingFilesMetrics.put(containerPath, containerMetrics);
+
+                        Map<String, Set<String>> missingFiles = results.get(containerId);
+                        for (String source : missingFiles.keySet())
+                            containerMetrics.put(source, missingFiles.get(source).size());
```
labkey-jeckels: Per our current policy, we shouldn't collect container-level metrics:

https://www.labkey.org/Documentation/Archive/24.3/wiki-page.view?name=reportUsageStatistics#exception

I suggest collapsing this to site-wide counts for metric reporting, but retaining the ability to report at a per-container level for admins who have access to the server itself, as it will be extremely helpful to track down the problematic records.

labkey-jeckels (changes requested):

experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
XingY: Good reminder. Updated to show a summary count instead.

api/src/org/labkey/api/exp/api/ExperimentService.java
```diff
@@ -1086,6 +1086,8 @@ List<? extends ExpProtocol> getExpProtocolsWithParameterValue(
 
     boolean useStrictCounter();
 
+    Map<String, Map<String, Set<String>>> doMissingFilesCheck(User user, Container container) throws SQLException;
```
labkey-jeckels: Do we need to expose this to other modules? It looks like all callers are within the experiment module so it probably isn't needed here.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -9729,6 +9733,63 @@ public boolean useStrictCounter()
...
+        SQLFragment selectSql;
+        if (containerIds == null || containerIds.isEmpty())
+            selectSql = unionSql;
+        else
+            selectSql = new SQLFragment("SELECT * FROM (").append(unionSql).append(") WHERE Container ").appendInClause(cf.getIds(), CoreSchema.getInstance().getSchema().getSqlDialect());
+        try (ResultSet rs = new SqlSelector(CoreSchema.getInstance().getSchema(), selectSql).getResultSet(false))
+        {
+            while (rs.next())
+            {
+                String filePath = rs.getString("FilePath");
```
labkey-jeckels, line 9723: Can we push this into a WHERE clause instead?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -9729,6 +9733,63 @@ public boolean useStrictCounter()
...
+                if (StringUtils.isEmpty(filePath))
+                    continue;
+
+                File file;
+                if (filePath.startsWith("file:"))
```
labkey-jeckels, line 9729: We may have S3 URLs in `exp.data`, which will start with `s3://`

Perhaps check if it's a URI (construct a `URI` and catch the exception falling back on the `File` constructor if not). If so, use `Files.exists(Path.of(uri))` instead.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -9729,6 +9733,63 @@ public boolean useStrictCounter()
...
+
+        Map<String, Map<String, Set<String>>> missingFiles = new HashMap<>();
+        SQLFragment unionSql = fileListener.listFilesQuery();
+        Collection<GUID> containerIds = cf.getIds();
+        SQLFragment selectSql;
+        if (containerIds == null || containerIds.isEmpty())
+            selectSql = unionSql;
+        else
+            selectSql = new SQLFragment("SELECT * FROM (").append(unionSql).append(") WHERE Container ").appendInClause(cf.getIds(), CoreSchema.getInstance().getSchema().getSqlDialect());
+        try (ResultSet rs = new SqlSelector(CoreSchema.getInstance().getSchema(), selectSql).getResultSet(false))
```
labkey-jeckels, line 9719: This has the potential to return a massive list. Maybe we should bail out if we hit 1,000 or 10,000 missing files for both perf and memory use?

experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
labkey-jeckels: That seems like a good idea for starters, but it might be a pain to enable where we want it. We could add a startup property to enable but should check with SysEng on preferences. Our initial focus is for LKSM customers.

@labkey-adam @labkey-stuartm @labkey-jony @labkey-willm

experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
```diff
@@ -0,0 +1,74 @@
...
+                {
+                    for (String containerId : results.keySet())
+                    {
+                        Map<String, Set<String>> missingFiles = results.get(containerId);
+                        for (String source : missingFiles.keySet())
+                            totalCount += missingFiles.get(source).size();
+                    }
+                }
+                missingFilesMetrics.put("Missing files count", totalCount);
+                svc.registerUsageMetrics(ExperimentServiceImpl.MODULE_NAME, () -> metrics);
```
labkey-jeckels: Won't this keep adding a usage metrics implementation every time it runs? Can we register it once and serve up the most recent data?

labkey-jeckels (changes requested):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: This task checks for files for File_Link field type only, so they would be from either exp.object or provisioned table, but not exp.data. But turns out we currently don't support File_Link fields with S3 storage (Issue 50538 , not sure if it ever worked). We should probably address 50538 and punt this to 24.7.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: The WHERE already checks for FilePath not null, I added the check here to be extra cautious, but can remove it.

experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
XingY: OK, I made this turned off by default and added a new startup prop to turn this on:
FileLinkMetrics.EnableFileLinkMetricsTask;startup = true

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: Changes made to cap at 1000 missing files

experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
XingY: updated to register only once

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: We might be able to punt the S3 part to 24.7. Since users aren't able to upload S3 files for FILE_LINK fields in the 1st place, the metrics is irrelevant at the moment.

experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
```diff
@@ -0,0 +1,132 @@
...
+                    for (String source : missingFiles.keySet())
+                        totalCount += missingFiles.get(source).size();
+                }
+            }
+            missingFilesMetrics.put("Missing files count", totalCount);
+            FileLinkMetricsProvider.getInstance().updateMetrics(metrics);
+        }
+        catch (Exception e)
+        {
+            log.error("Unable to run missing files check task. {}", e.getMessage());
```
labkey-jeckels: It's usually helpful to have the full stack trace logged.

```suggestion
            log.error("Unable to run missing files check task.", e);
```

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-jeckels: Deferring any S3 handling sounds fine. Can you add a comment in the code and/or issue to make sure we remember to make this metric compatible when we start supporting it?

experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
```diff
@@ -0,0 +1,132 @@
...
+        ModuleLoader.getInstance().handleStartupProperties(new StandardStartupPropertyHandler<>(STARTUP_SCOPE, StartupProperties.class)
+        {
+            @Override
+            public void handle(Map<StartupProperties, StartupPropertyEntry> map)
+            {
+                StartupPropertyEntry entry = map.get(StartupProperties.EnableFileLinkMetricsTask);
+                if (null != entry && Boolean.valueOf(entry.getValue()))
+                {
+                    PropertyManager.PropertyMap writableProps = PropertyManager.getWritableProperties(SystemMaintenance.SET_NAME, true);
+                    String disabled = writableProps.get(SystemMaintenance.DISABLED_TASKS_PROPERTY_NAME);
```
labkey-jeckels: Can we consolidate any of this with the code in `SystemMaintenance.setProperties()`? It's unfortunate to have more places that know the details of how these values are parsed and persisted.

labkey-jeckels (approved):

experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
XingY: yep, done

PR #5533, opened 2024-05-24
Assay import support for results domain file fields

cnathe:
#### Rationale
This story makes it easier to import an assay run to the app in the case where the assay results domain has File field types. Before this, you had to individually and manually add files to results after importing/creating the run. Now I can make the associations in bulk and upload the results files while importing a run through the app UI.

#### Related Pull Requests
- https://github.com/LabKey/labkey-api-js/pull/176
- https://github.com/LabKey/labkey-ui-components/pull/1501
- https://github.com/LabKey/labkey-ui-premium/pull/420
- https://github.com/LabKey/platform/pull/5533
- https://github.com/LabKey/limsModules/pull/302
- https://github.com/LabKey/commonAssays/pull/763

#### Changes
- AssayResultUpdateService updateRows to call convertTypes to handle saving files to run specific file location, also have bulk update of file prop reference the same file instead of creating N copies (1 for each selected row)
- AssayFileWriter.ensureUploadDirectoryPath to create parent directories if necessary
- AssayResultsFileWriter to save posted results files to run specific dir durinv saveExperimentRun
- Delete assay run results specific file dir on run delete
- PipelineService helper for getJobGUID from rowId
- AssayResultsFileWriter to save posted results files to job specific dir during saveExperimentRun of import in background
- AssayResultsFileWriter the trim multipartfile name to remove relative path, if it exists
- AssayFileWriter to only ensure target dir if a file is about to be written to it
- @assayfiles dir 
  - write run results files to dir
  - resolve file names to dir in ExpDataFileConverter and AssayUploadFileResolver
  - resolve download link to dir in FileLinkDisplayColumn and CoreController
  - AssayFilesResourceProvider to allow reading of @assayfiles dir in webdav controller and CrUD for admins


experiment/src/org/labkey/experiment/ScriptsResourceProvider.java
```diff
@@ -69,6 +70,11 @@ public WebdavResource resolve(@NotNull WebdavResource parent, @NotNull String na
             AttachmentDirectory dir = service.getMappedAttachmentDirectory(c, FileContentService.ContentType.scripts, false);
             if (dir != null)
             {
+                // don't create the dir here if it doesn't already exist, it will be created the first time an assay transform script is saved
+                java.nio.file.Path path = dir.getFileSystemDirectoryPath(c, false);
+                if (!Files.exists(path))
+                    return null;
+
```
labkey-klum: So we don't create @scripts through the _webdav action?

cnathe: Good point. Turns out we do need this. I had added this check for the @assayfiles dir since we only want the code path that writes files during an assay run import to be the one creating this dir. I had thought @scripts was similar, but it needs to create the dir in this ScriptsResourceProvider.

api/src/org/labkey/api/assay/AbstractAssayProvider.java
```diff
@@ -1925,7 +1925,7 @@ private void moveRunsBatch(List<ExpRun> runs, Container sourceContainer, Contain
                         ExpRun run = batchRun.get(experiment.getRowId());
                         Integer runId = run.getRowId();
                         movedFiles.putIfAbsent(runId, new ArrayList<>());
-                        movedFiles.get(runId).add(new AssayFileMoveData(run, fileProp.getName(), sourceFile, updatedFile));
+                        movedFiles.get(runId).add(new AssayFileMoveData(run, run.getContainer(), fileProp.getName(), sourceFile, updatedFile));
```
labkey-jeckels, line 1928: Do we need to pass the source container as a separate argument? Won't it always be `run.getContainer()`?

api/src/org/labkey/api/assay/AssayResultsFileWriter.java
```diff
@@ -0,0 +1,146 @@
...
+
+        // if this is a background import and the files have been stashed in the pipeline results dir, move them to the assay results dir
+        if (_run != null && _run.getJobId() != null)
+        {
+            String pipelineJobGUID = PipelineService.get().getJobGUID(context.getUser(), container, _run.getJobId());
+            Path pipelineResultsDir = pipelineJobGUID != null ? getAssayFilesDirectoryPath(container, getPipelineResultsFileDir(_protocol, pipelineJobGUID)) : null;
+            if (pipelineResultsDir != null && Files.exists(pipelineResultsDir))
+            {
+                Path targetDir = getAssayFilesDirectoryPath(container, getFileTargetDirName());
+                FileUtils.moveDirectory(pipelineResultsDir.toFile(), targetDir.toFile()); // TODO will .toFile() work here for cloud?
```
labkey-jeckels:
```suggestion
                FileUtils.moveDirectory(pipelineResultsDir.toFile(), targetDir.toFile()); // TODO refactor to support cloud-based storage which won't have a toFile()
```

api/src/org/labkey/api/assay/AssayResultsFileWriter.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    public void cleanupPostedFiles(Container container, boolean deleteIfEmptyOnly) throws ExperimentException
+    {
+        try
+        {
+            Path targetDir = getAssayFilesDirectoryPath(container, getFileTargetDirName());
+            if (targetDir != null && Files.exists(targetDir) && Files.isDirectory(targetDir))
+            {
+                if (!deleteIfEmptyOnly || FileUtils.isEmptyDirectory(targetDir.toFile())) // TODO will .toFile() work here for cloud?
```
labkey-jeckels:
```suggestion
                if (!deleteIfEmptyOnly || FileUtils.isEmptyDirectory(targetDir.toFile())) // TODO refactor to support cloud-based storage which won't have a toFile()
```

pipeline/src/org/labkey/pipeline/api/PipelineServiceImpl.java
```diff
@@ -970,6 +970,17 @@ public Integer getJobId(User u, Container c, String jobGUID)
         return statusFile.getRowId();
     }
 
+    @Override
+    public String getJobGUID(User u, Container c, Integer rowId)
```
labkey-jeckels: `int` instead of `Integer`

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -589,7 +589,7 @@ private boolean isGridFormat(List<Map<String, Object>> data)
             return true;
 
         // only the tabular formats will have the well location field
-        return !data.get(0).containsKey(AssayResultDomainKind.WELL_LOCATION_COLUMN_NAME);
+        return !data.get(0).containsKey(AssayResultDomainKind.WELL_LOCATION_COLUMN_NAME) && !data.get(0).containsKey("Well Location");
```
labkey-jeckels, line 592: Is this intentional? Seems unrelated.

api/src/org/labkey/api/assay/AssayResultsFileWriter.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public String getFileName(MultipartFile file)
+    {
+        // since results file upload can support uploading directories of files, some browsers retain path info
+        String filename = file.getOriginalFilename();
+        if (null == filename)
+            return null;
+        int slash = Math.max(filename.lastIndexOf("/"), filename.lastIndexOf("\\"));
```
labkey-jeckels: This is also good because it will strip out any `../../`, etc from the path

api/src/org/labkey/api/assay/AssayUploadFileResolver.java
```diff
@@ -65,7 +68,13 @@ else if (o instanceof String)
             if (fileToResolve != null)
             {
                 // For security reasons, make sure the user hasn't tried to reference a file that's not under
-                // the pipeline root. Otherwise, they could get access to any file on the server
+                // the pipeline root or @assayfiles root. Otherwise, they could get access to any file on the server
+
+                Path assayFilesRoot = FileContentService.get().getFileRootPath(container, FileContentService.ContentType.assayfiles);
+                if (assayFilesRoot != null && URIUtil.isDescendant(assayFilesRoot.toUri(), fileToResolve.toURI()))
```
labkey-jeckels, line 74: `URIUtil.isDescendant()` doesn't protected against `/../../` style paths. Let's make it more robust and make it return false if the descendant includes `"../"`. Otherwise, I'm worried that we'll think we're getting protection when we're not. I think other callers are doing supplemental checks but that doesn't protect new callers. @labkey-matthewb FYI

api/src/org/labkey/api/query/DefaultQueryUpdateService.java
```diff
@@ -716,7 +718,13 @@ final protected void convertTypes(User user, Container c, Map<String,Object> row
         convertTypes(user, c, row,  getDbTable(), null);
     }
 
-    protected void convertTypes(User user, Container c, Map<String,Object> row, TableInfo t, @Nullable String file_link_dir_name) throws ValidationException
+    final protected void convertTypes(User user, Container c, Map<String,Object> row, TableInfo t, @Nullable String file_link_dir_name) throws ValidationException
```
labkey-jeckels: `file_link_dir_name` isn't a normal argument naming convention

api/src/org/labkey/api/query/DefaultQueryUpdateService.java
```diff
@@ -716,7 +718,13 @@ final protected void convertTypes(User user, Container c, Map<String,Object> row
...
     }
 
-    protected void convertTypes(User user, Container c, Map<String,Object> row, TableInfo t, @Nullable String file_link_dir_name) throws ValidationException
+    final protected void convertTypes(User user, Container c, Map<String,Object> row, TableInfo t, @Nullable String file_link_dir_name) throws ValidationException
+    {
+        Path path = AssayFileWriter.getUploadDirectoryPath(c, file_link_dir_name);
+        convertTypes(user, c, row, t, path, false);
+    }
+
+    protected void convertTypes(User user, Container c, Map<String,Object> row, TableInfo t, @Nullable Path file_link_dir, boolean useExistingFile) throws ValidationException
```
labkey-jeckels: Same with `file_link_dir`

labkey-jeckels (approved):
Happy to re-review if any of my suggestions aren't clear or you want a double-check on any changes.

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
cnathe: Yes. In testing the upload of results files with the plate set assay case, I downloaded the assay results data file template and it had "Well Location" as the column header. That caused an issue with this line during the assay import.

api/src/org/labkey/api/query/DefaultQueryUpdateService.java
cnathe: updated to fileLinkDirName

api/src/org/labkey/api/query/DefaultQueryUpdateService.java
cnathe: updated to fileLinkDirPath

api/src/org/labkey/api/assay/AbstractAssayProvider.java
cnathe: Yes, the source container will always be `run.getContainer()` when this `AssayFileMoveData` is created. But the place I am using that sourceContainer is in the `ExperimentServiceImpl.moveAssayRuns()` after the run and files have been moved. In that method the run has already been moved and the container value updated by the time we get to the `cleanupPostedFiles` call so if I do `renameData.run().getContainer()` it gives me the new/updated container instead of the original source container. So I decided to just stash the source container with `AssayFileMoveData`

api/src/org/labkey/api/assay/AssayUploadFileResolver.java
cnathe: @labkey-jeckels Are you thinking a check like this directly in `URIUtil.isDescendant`?
```
    static public boolean isDescendant(URI base, URI descendant)
    {
        // to protect against paths that are trying to escape the base, return false if the descendant includes "../"
        if (descendant.getPath().contains(".."))
            return false;

        if (!descendant.getScheme().equalsIgnoreCase(base.getScheme()))
            return false;
        return descendant.getPath().startsWith(base.getPath());
    }
```

labkey-jeckels: Yes, but I guess we need to check explicitly for `"/../"`. Otherwise we'll erroneously reject paths like `/testing.../file.txt` which aren't a problem.

cnathe:
@labkey-jeckels There have been some changes on the server side based on manual testing issues found / feedback for this PR since you did your CR. I wanted to call them out here in case you want to take another quick pass over the code changes for review:
- Handle assay results file upload with only casing differences 
  - server side with uniquify names on save so we need to take that into account when resolving the file paths in AssayResultsFileConverter
- FileColumnValueMapper to handle mapping file column values to saved files for multiple row updates
  - if you Edit in Bulk and upload a file for N rows, that single file will be referenced by those N rows instead of creating N copies of the file
- Ensure exp.Data row for results files added during assay run import 
- Resolve exp.Data webdav urls for data files under @assayfiles
- Revert code so that we DO NOT clean-up / delete the assay run results files dir on run deletion or move
- AssayFilesResourceProvider to allow actions for admin

labkey-jeckels (approved):

api/src/org/labkey/api/assay/AbstractAssayProvider.java
```diff
@@ -2104,7 +2104,20 @@ private void updateResultFiles(FilteredTable assayResultTable, List<ExpRun> runs
                     movedFiles.putIfAbsent(resultRunId, new ArrayList<>());
                     movedFiles.get(resultRunId).add(new AssayFileMoveData(run, run.getContainer(), fileField, sourceFile, updatedFile));
 
-                    SQLFragment updateSql = new SQLFragment("UPDATE ").append(assayResultTable.getRealTable())
+                    // update the exp.Object row for the file results row
+                    TableInfo expDataTable = ExperimentService.get().getTinfoData();
```
XingY, line 2110: Don't we already have a ExpDataFileListener that handles updating exp.data when fileMoved? Curious why that didn't pick this up

labkey-jeckels: Good point - since we're passing in the source and target containers, maybe we should make the FileListeners responsible for reparenting the container too.

cnathe: So with my last change to fix this issue, I did add to the ExpDataFileListener so that it does `data.setContainer(targetContainer);` when the targetContainer is provided and different than the sourceContainer. So I'm not following the question. This update statement you are referencing is updating the related exp.Object table rows for those exp.Data rows being updated. Are you saying that we should have it so that the ExpDataFileListener also handles exp.Object updates?

labkey-jeckels: Yes, though I haven't looked at how feasible that might be.

XingY: I see what you are saying.. Here can we use exp.data.objectId instead of objecturi in the WHERE clause to identify exp.object? That would make future refactoring of incorporating exp.object update as part of ExpDataFileListener cleaner.

PR #5535, opened 2024-05-25
Plate: Source plates from plate templates

labkey-nicka:
#### Rationale
Introduces capability to source a plate from a plate template. These scenarios include:

- Sourcing of a set of plates via `plate-createPlateSet.api`. This includes re-plating, re-array, and custom plate construction.
- Sourcing of a plate via `plate-createPlate.api`.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1502
- https://github.com/LabKey/labkey-ui-premium/pull/421
- https://github.com/LabKey/platform/pull/5535
- https://github.com/LabKey/limsModules/pull/304

#### Changes
- Support specifying a "templateId" within a `CreatePlateSetPlate`
- Support replating a plate set as an explicit operation. Plates now copied via `copyPlate()`
- Re-array from with a selection of samples into a new plate set. These plates can be sourced from plate types and plate templates.
- Introduce `WellData` model to encapsulate all data within a well. Handles database deserialization and generation of query row data.
- Rename `Plate.Well.Group` column to `Plate.Well.WellGroup` to avoid conflicting with "GROUP" keyword in SQL (was making it so we had to do extra work to serialize).

### DIFF
--- a/assay/package-lock.json
+++ b/assay/package-lock.json
@@ -8,7 +8,7 @@
       "name": "assay",
       "version": "0.0.0",
       "dependencies": {
-        "@labkey/components": "3.46.0"
+        "@labkey/components": "3.46.6-fb-plate-template-deux.0"
       },
       "devDependencies": {
         "@labkey/build": "7.3.0",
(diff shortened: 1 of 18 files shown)


labkey-klum (approved):
I didn't have time to do an extensive review, but overall the approach looks good and is consistent with some of the recent conversations Nick and I have had regarding copying and replating.

PR #5537, opened 2024-05-27
Study navigator improvements

labkey-adam:
#### Rationale
[Issue 44350: Migrate from and remove legacy "visitRowId" parameter from DatasetAction](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=44350)

#### Important Implications
- Participant action caches (in session) the distinct participant list used for prev/next navigation. Previously, we extracted just dataset id, view name, cohort filter, and qc state filter from the URL and concatenated them to use as a cache key. And the action would aggressively clear the cache in dataset action, to work around potential filter/sort changes. I've switched to using the full query string as the key, which will include the above plus sort and other filter parameters. This seems like a simpler and more reliable approach.
- Any old saved favorites to the dataset page with visit filtering applied will no longer filter by visit.
- Switching to a standard filter parameter on visit label means the visit filter is clearly shown in the grid header as usual (see `Visit = 'Month 10'` below), so I removed it from the nav trail. We used to show "All Visits" or "Month 10" or whatever after the dataset label. (We're still "special casing" cohort and qcstate, extracting these values from the URL and displaying them above the grid, but I don't think this adds much value. I switched to cohort label filtering, which makes the grid filter much more useful. I would be in favor of removing these special case labels.)

image

#### Changes
- Study Navigator (overview.jsp) improvements:
  - Add a normal visit filter parameter to the visit-specific URLs. These are respected by summary statistics, charting, export script, etc.
  - Remove special handling for visitRowId in DatasetAction (special filtering, nav trail)
  - Fix broken cohort filter on most links (missing data region name)
  - Modernize HTML generation (use `HtmlStringBuilder` and `SelectBuilder`)
  - Fix nav trails using returnUrls
  - Cache participant lists using a Guava Cache that imposes a 10-minute expiration on entries and limits per-user entries to 10
  - Switch cohort filter to use label instead of rowId
- Miscellaneous study fixes (fix constant names, IntelliJ warnings)


study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -3257,27 +3242,20 @@ private static Map<String, Map<String, Integer>> getDatasetSortColumnMap(ViewCon
...
+        Map<String, List<String>> map = getParticipantMapFromSession(context);
+        map.remove(getParticipantListCacheKey(context));
     }
 
     @SuppressWarnings("unchecked")
-    private static Map<String, List<String>> getParticipantMapFromCache(ViewContext context)
+    private static Map<String, List<String>> getParticipantMapFromSession(ViewContext context)
     {
         HttpSession session = context.getRequest().getSession(true);
         Map<String, List<String>> map = (Map<String, List<String>>) session.getAttribute(PARTICIPANT_CACHE_PREFIX);
```
labkey-jeckels: Should we limit the size of this cache?

study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -3257,27 +3242,20 @@ private static Map<String, Map<String, Integer>> getDatasetSortColumnMap(ViewCon
...
-        // if there is also a view associated with the dataset, incorporate it into the key as well
-        if (viewName != null && !StringUtils.isEmpty(viewName))
-            key = key + viewName;
-        if (cohortFilter != null)
-            key = key + "cohort" + cohortFilter.getCacheKey();
-        if (encodedQCState != null)
-            key = key + "qcState" + encodedQCState;
-        return key;
+        // The query string includes all parameters that affect the participant list: dataset id, filters, sorts, etc.
+        return context.getActionURL().getQueryString();
```
labkey-jeckels: If a user adds/removes a filter from the saved view, so we invalidate the cache anywhere?

labkey-jeckels (approved):
See comment about possible (and likely preexisting) failure to invalidate the cached participant list if a view's filters are updated

study/src/org/labkey/study/controllers/StudyController.java
labkey-adam: I was thinking about time-limiting the entries instead. Maybe use a Guava `CacheBuilder` with `expireAfterAccess()` set to 10 minutes. Maybe use soft references as well.

study/src/org/labkey/study/controllers/StudyController.java
labkey-adam: They would likely do this by hitting the dataset action, which proactively clears the cache for the current query string

study/src/org/labkey/study/controllers/StudyController.java
labkey-adam: Switched to Guava CacheBuilder with a 10-minute expiration and max size of 10

PR #5538, opened 2024-05-27
Issue 50355: Add metrics for boolean field type

labkey-susanh:
#### Rationale
Issue [50355](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50355)

#### Related Pull Requests

- https://github.com/LabKey/limsModules/pull/305

#### Changes
- Metrics for assay run, assay result, sample type, data class domains with boolean field types


experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -696,6 +715,11 @@ WHERE op.propertyid IN (
                         "         JOIN exp.PropertyDomain PD ON D.propertyId = PD.propertyid\n" +
                         "         JOIN exp.DomainDescriptor DD on PD.domainID = DD.domainId\n" +
                         "WHERE DD.storageSchemaName = ? AND D.rangeURI = ?", DataClassDomainKind.PROVISIONED_SCHEMA_NAME, PropertyType.ATTACHMENT.getTypeUri()).getObject(Long.class));
+                results.put("dataClassWithBooleanColumnCount", new SqlSelector(ExperimentService.get().getSchema(), "SELECT COUNT(DISTINCT DD.DomainURI) FROM\n" +
+                        "     exp.PropertyDescriptor D \n" +
+                        "         JOIN exp.PropertyDomain PD ON D.propertyId = PD.propertyid\n" +
+                        "         JOIN exp.DomainDescriptor DD on PD.domainID = DD.domainId\n" +
+                        "WHERE DD.storageSchemaName = ? AND D.rangeURI = ?", DataClassDomainKind.PROVISIONED_SCHEMA_NAME, PropertyType.BOOLEAN.getTypeUri()).getObject(Long.class));
```
cnathe, line 722: It looks like we have about 12 of these similar metrics which are a select count from PropertyDescriptory with the same joins that just differ in the where clause. At some point we should factor out a helper method that builds the SQL and passes in the WHERE clause as a param.

cnathe (approved):

PR #5543, opened 2024-05-29
Tolerate bogus property validators

labkey-adam:
#### Rationale
We've seen deployments in the field that must have failed to run deferred upgrade code (from 2021) that cleared out obsolete 'urn:lsid:labkey.com:PropertyValidator:length' rows. In both cases, the server would no longer startup until the problematic rows were cleared. (I think the relatively new addition of audit indexes causing domains to resave surfaced these problems.) We can be more tolerant; by testing the type URI at cache load time we can ensure that all validators are able to resolve their `ValidatorKind`. If not, we log an error and skip the bogus validator. By never caching them, no other code path will see them, so no need for downstream null checks.

See:

* https://www.labkey.org/home/Support/LabKey%20Support%20Forum/announcements-thread.view?rowId=23503
* https://www.labkey.org/NCI/Support%20Tickets/issues-details.view?issueId=49542


experiment/src/org/labkey/experiment/api/property/DomainPropertyManager.java
```diff
@@ -169,7 +168,18 @@ public List<ConditionalFormatWithPropertyId> getConditionalFormats(Container con
 
         final MultiValuedMap<Integer, PropertyValidator> validators = new ArrayListValuedHashMap<>();
 
-        new SqlSelector(getExpSchema(), sql, container).forEach(PropertyValidator.class, pv -> validators.put(pv.getPropertyId(), pv));
+        new SqlSelector(getExpSchema(), sql, container).forEach(PropertyValidator.class, pv -> {
+            // Warn and skip property validators have unknown type URIs. Some instances in the field failed to run
+            // upgrade code that cleared out obsolete 'urn:lsid:labkey.com:PropertyValidator:length' rows.
```
labkey-jeckels, line 173: Clearly this is better than throwing NPEs.

Should we make another attempt to clean up the obsolete rows too? Is there any way for admins to manually recover, aside from deleting the properties completely or running delete SQL directly against the DB?

labkey-jeckels (approved):

experiment/src/org/labkey/experiment/api/property/DomainPropertyManager.java
labkey-adam: We could easily restore the `cleanupLengthTypePropertyValidators()` upgrade code, last seen in 22.11, and invoke it again (and make it available for manual invocation... at least until we remove it). I'm a bit meh on that idea... we've only seen two instances that skipped the deferred upgrade code. LMK if you feel strongly otherwise.

PR #5544, opened 2024-05-29
Fix bounds for one-dimensional plot brushing

labkey-tchad:
#### Rationale
Passing `null` as the selection extent for the opposing axis for one-dimensional brushing forces callers to add null handling to determine whether points are selected. The selection range is essentially infinite for these, so we could replace the `null`s with infinite values. (This will require a small update to CDS)
Our API examples ([demo_script.js](https://github.com/LabKey/platform/blob/develop/core/webapp/vis/demo/demo_script.js#L425), [setAesTest.html](https://github.com/LabKey/testAutomation/blob/develop/modules/chartingapi/resources/views/setAesTest.html#L291) and [interactivityTest.html](https://github.com/LabKey/testAutomation/blob/develop/modules/chartingapi/resources/views/interactivityTest.html#L101)) expect this sort of response.

#### Related Pull Requests
- 

#### Changes
- Fix bounds for one-dimensional plot brushing


cnathe (approved):
I pulled down the change and it looks to work as expected on Chrome and FF. Thanks.

labkey-tchad:
@cnathe Unfortunately, this update requires several CDS changes: https://github.com/LabKey/cds/compare/fb_enableAdvancedBrushTest
Since CDS is the primary use of this API and nobody is complaining, I'm inclined to leave the API unchanged rather than risk damaging CDS functionality.
Let me know if you disagree.

cnathe:
> @cnathe Unfortunately, this update requires several CDS changes: https://github.com/LabKey/cds/compare/fb_enableAdvancedBrushTest Since CDS is the primary use of this API and nobody is complaining, I'm inclined to leave the API unchanged rather than risk damaging CDS functionality. Let me know if you disagree.

Bummer, but yes I agree.

PR #5546, opened 2024-05-29
Sample Type & Source Type Designer Roles

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1505
- https://github.com/LabKey/labkey-ui-premium/pull/424
- https://github.com/LabKey/platform/pull/5546
- https://github.com/LabKey/limsModules/pull/319

#### Changes
- new designer roles for samples and dataclasses
- Change DeleteProtocolByRowIdsAPI to require DesignAssayPermission instead of delete perm
- check for existing data in datatype before deleting a datatype for non-admins
- jest integration test for designer roles


api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -2033,7 +2034,16 @@ else if (value.size() == 0)
...
+                        {
+                            try
+                            {
+                                rootValue = ConvertUtils.convert((String)rootValue, pkCol.getJavaClass());
+                            }
+                            catch (ConversionException x)
+                            {
+                                throw new IllegalArgumentException(x);
+                            }
+                        }
```
cnathe, line 2046: was this change for a different issue? if so, can you add the issue number? seems unrelated to this story

api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
```diff
@@ -342,6 +343,11 @@ public boolean canCreateDefinition(User user, Container container)
     @Override
     public boolean canDeleteDefinition(User user, Domain domain)
     {
+        if (!domain.getContainer().hasPermission(user, AdminPermission.class))
+        {
+            ExpSampleType st = getSampleType(domain);
+            return !st.hasData();
```
cnathe, line 351: this returns before checking DesignSampleTypePermission. so would this be true for an empty sample type if the user was and editor or reader?

core/src/org/labkey/core/CoreModule.java
```diff
@@ -1169,6 +1170,11 @@ public void moduleStartupComplete(ServletContext servletContext)
             results.put("uniqueUserCountThisMonth", UserManager.getUniqueUsersCount(cal.getTime()));
             results.put("scriptEngines", LabKeyScriptEngineManager.get().getScriptEngineMetrics());
             results.put("customLabels", CustomLabelService.get().getCustomLabelMetrics());
+            Map<String, Long> roleAssignments = new HashMap<>();
+            roleAssignments.put("assayDesignerCount", new SqlSelector(CoreSchema.getInstance().getSchema(), "SELECT COUNT(*) FROM core.RoleAssignments WHERE userid > 0 AND role = ?", "org.labkey.assay.security.AssayDesignerRole").getObject(Long.class));
+            roleAssignments.put("dataClassDesignerCount", new SqlSelector(CoreSchema.getInstance().getSchema(), "SELECT COUNT(*) FROM core.RoleAssignments WHERE userid > 0 AND role = ?", "org.labkey.experiment.security.DataClassDesignerRole").getObject(Long.class));
+            roleAssignments.put("sampleTypeDesignerCount", new SqlSelector(CoreSchema.getInstance().getSchema(), "SELECT COUNT(*) FROM core.RoleAssignments WHERE userid > 0 AND role = ?", "org.labkey.experiment.security.SampleTypeDesignerRole").getObject(Long.class));
```
cnathe: the SQL here is the same for all so could be factored out to a variable

experiment/src/org/labkey/experiment/api/DataClassDomainKind.java
```diff
@@ -293,6 +294,12 @@ public Set<String> getNonDisablebleFields()
     @Override
     public boolean canDeleteDefinition(User user, Domain domain)
     {
+        if (!domain.getContainer().hasPermission(user, AdminPermission.class))
+        {
+            ExpDataClass dataClass = getDataClass(domain);
+            if (dataClass != null)
+                return !dataClass.hasData();
```
cnathe, line 304: similar comment here, this could return true without checking for `DesignDataClassPermission`, right?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -4679,8 +4679,10 @@ public void deleteProtocolByRowIds(Container c, User user, String auditUserComme
                 {
                     for (Protocol protocol : protocols)
                     {
-                        ExpProtocol protocolToDelete = new ExpProtocolImpl(protocol);
+                        if (!c.hasPermission(user, AdminPermission.class) && runs.size() > 0)
+                            throw new UnauthorizedException("You do not have sufficient permissions to delete '" + protocol.getName() + "'.");
```
cnathe: this is all transacted, so should work fine but should this check be moved to the top of this method so that we don't have to go through the work of deleting the datasets and runs just to find out that we are throwing here?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -1639,7 +1639,7 @@ void onFirst()
 
         private NameGenerator.SampleNameExpressionSummary getSampleNameExpressionSummary(@NotNull NameGenerator nameGen, @NotNull NameGenerator aliquotNameGen)
         {
-            if (sampleType == null)
+            if (sampleType == null || nameGen.getExpressionSummary() == null)
```
cnathe, line 1645: this also seems unrelated to this story? issue number to add here maybe?

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -3423,6 +3423,9 @@ public ModelAndView getView(DeleteForm form, boolean reshow, BindException error
             {
                 for (ExpProtocol protocol : protocols)
                 {
+                    if (!protocol.getContainer().hasPermission(getUser(), DesignAssayPermission.class))
+                        throw new UnauthorizedException("You do not have sufficient permissions to delete this assay design.");
```
cnathe, line 3430: this check was added for DeleteProtocolByRowIdsAction but not for DeleteProtocolByRowIdsAPIAction above. Was that intentional?

cnathe (changes requested):

api/src/org/labkey/api/data/NameGenerator.java
XingY: Yep, this is from exception report, no issue number

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
XingY: same, this is from exception report, no issue number

api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
XingY: good catch, fixed

core/src/org/labkey/core/CoreModule.java
XingY: done

experiment/src/org/labkey/experiment/api/DataClassDomainKind.java
XingY: fixed

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: updated

experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
XingY: change made so it's checked for both actions

cnathe (approved):

PR #5548, opened 2024-05-30
Better handling of bogus filter column in linked schema XML

labkey-adam:
#### Rationale
`InClause.getLabKeySQLWhereClause()` will throw an NPE if passed a non-existent column. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50516

Key change is the null check at `SimpleFilter:706`. The other changes squelch warnings and consolidate duplicate code. I can't guarantee every other clause is perfect, but I no longer see any NPE warnings in this class. Adding a bogus filter with an equals clause gave a reasonable error message before my fix, although the message is now improved to: `Filter column 'Bogus' not found.` instead of previous `Column Bogus not found in column map.`

### DIFF
--- a/api/src/org/labkey/api/data/SimpleFilter.java
+++ b/api/src/org/labkey/api/data/SimpleFilter.java
@@ -18,7 +18,6 @@ package org.labkey.api.data;
 
 import org.apache.commons.beanutils.ConversionException;
 import org.apache.commons.beanutils.ConvertUtils;
-import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.junit.Assert;
@@ -49,6 +48,7 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import static org.labkey.api.data.CompareType.CONTAINS_NONE_OF;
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/data/SimpleFilter.java
```diff
@@ -700,41 +700,45 @@ public String getLabKeySQLWhereClause(Map<FieldKey, ? extends ColumnInfo> column
...
                 else if (!isNegated())
                     in.append(alias).append(" IN (NULL)");  // Empty list case; "WHERE column IN (NULL)" should always be false
-
-                return in.toString();
             }
+            else
+            {
+                if (null == col)
+                {
+                    throw new IllegalArgumentException("Filter column '" + _fieldKey.toDisplayString() + "' not found.");
```
labkey-jeckels, line 708: Does this result in a better user experience? Having a more specific error is great but doesn't it still break the whole schema in the query browser?

labkey-adam: Yes, it's much better and no it doesn't break the whole schema. LinkedSchema.createQueryDef() catches IllegalArgumentException and turns it into a parse exception. Only the single misconfigured query is affected. This makes InClause work the same as the equals case, which also throws IllegalArgumentException.

labkey-jeckels (approved):

api/src/org/labkey/api/data/SimpleFilter.java
labkey-jeckels: Ah, OK. Not my favorite pattern but better than catching an NPE in the same spot!

PR #5556, opened 2024-05-31
Plates: insert well data upon creation

labkey-nicka:
#### Rationale
 Improves performance of plate creation by eliminating second pass update of wells.

#### Changes
- Refactor `PlateManager.savePlateImpl()` to optionally process additional well data when inserting new wells.
- Refactor `PlateManager.saveWellData()` to process just the derivation of custom metadata columns from well data. Renamed to `PlateManager.deriveCustomFieldsFromWellData()`. Done for backwards compatibility. This may go away in the near future if we move to sourcing visible fields by Plate Set rather than by Plate.
- Factor out all usages of `PlateManager.deleteAllPlateData()` outside of container delete. This also stops us from duplicately calling delete all plate data on the same container from the `DilutionManager`. Remove from public service interface.

### DIFF
--- a/assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
+++ b/assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
@@ -8,7 +8,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONObject;
-import org.junit.After;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.labkey.api.assay.AssayProtocolSchema;
@@ -1033,11 +1032,6 @@ public class AssayPlateMetadataServiceImpl implements AssayPlateMetadataService
 
             container = JunitUtil.getTestContainer();
             user = TestContext.get().getUser();
-        }
-
-        @After
-        public void cleanup()
-        {
         }
 
         @Test
--- a/assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
+++ b/assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
@@ -80,6 +80,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.labkey.api.assay.AssayResultDomainKind.WELL_LSID_COLUMN_NAME;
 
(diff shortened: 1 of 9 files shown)


assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -1026,69 +1027,71 @@ public static final class TestCase
...
+
             container = JunitUtil.getTestContainer();
             user = TestContext.get().getUser();
-
-            PlateService.get().deleteAllPlateData(container);
         }
 
         @After
-        public void cleanupTest()
+        public void cleanup()
```
cnathe: minor: this method looks to now be empty, can it be removed?

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -796,7 +775,14 @@ private void setProperties(Container container, PropertySetImpl propertySet)
     public int save(Container container, User user, Plate plate) throws Exception
     {
         if (plate instanceof PlateImpl plateTemplate)
-            return savePlateImpl(container, user, plateTemplate);
+            return save(container, user, plateTemplate, null);
+        throw new IllegalArgumentException("Only plate instances created by the plate service can be saved.");
```
cnathe: minor: can this just call the save() method below instead of duplicating the if statement and error message here?
i.e. `return save(container, user, plate, null);`

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -328,39 +325,21 @@ private void saveWellData(
...
-                        PlateCustomField customField = new PlateCustomField(col.getPropertyURI());
-                        customFields.add(customField);
+                        customFields.add(new PlateCustomField(col.getPropertyURI()));
                     }
                 }
             }
-            else
-            {
-                // should we fail or just log?
-                LOG.error("Unable to add well data for plate: " + plate.getName() + " each data row must contain a wellLocation field.");
```
cnathe, line 355: unless I missed it, the updated code path doesn't have an error or log if the well location is missing from the row. Was that intentionally removed?

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -1087,6 +1096,58 @@ private int savePlateImpl(Container container, User user, PlateImpl plate, boole
...
+        Set<String> keywords = CaseInsensitiveHashSet.of(
+            WellTable.Column.Col.name(),
+            WellTable.Column.Container.name(),
+            WellTable.Column.Lsid.name(),
+            WellTable.Column.PlateId.name(),
+            WellTable.Column.Position.name(),
+            WellTable.Column.Properties.name(),
+            WellTable.Column.Row.name(),
+            WellTable.Column.RowId.name(),
+            "WellLocation"
```
cnathe: this "WellLocation" const looks to be used several times in this class, is it worth a variable?

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -1087,6 +1096,58 @@ private int savePlateImpl(Container container, User user, PlateImpl plate, boole
...
+                {
+                    if (StringUtils.trimToNull(entry.getKey()) == null || entry.getValue() == null)
+                        continue;
+
+                    var key = entry.getKey();
+                    var customFieldPrefix = WellTable.Column.Properties.name().toLowerCase() + "/";
+                    if (key.toLowerCase().startsWith(customFieldPrefix))
+                        key = key.substring(customFieldPrefix.length());
+
+                    if (StringUtils.trimToNull(key) != null && keywords.contains(key))
```
cnathe: there is a `trimToNull` of the key at line 1129 but then the check at line 1134 does not `trimToNull` and then it is done again here. thoughts on just having `var key = StringUtils.trimToNull(entry.getKey());` once at the beginning of the for block and then using `key` in call cases?

cnathe (approved):

RosalineP (approved):
Looks good, wasn't able to catch anything Cory hadn't already pointed out

assay/src/org/labkey/assay/plate/PlateManager.java
labkey-nicka: I'd intentionally removed it but have since reintroduced it as a `ValidationException`.

assay/src/org/labkey/assay/plate/PlateManager.java
labkey-nicka: These aren't the same since we've possibly modified the key since the first statement. The second trim is paranoid of someone sending something like "Properties/    ".

PR #5568, opened 2024-06-06
Issue 49830: DataLoader.inferColumnInfo fall back to the known column class based on the columnInfoMap instead of defaulting to String

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=49830

For a sample type with a file field type, the DataLoader.inferColumnInfo() will treat the incoming data as a String.class instead of File.class if it doesn't find any valid paths. Since we are passing in the known column types to the DataLoader in the sample type case, we can use that type instead of String (which will force File fields for sample types through the file converter).

#### Changes
- DataLoader.inferColumnInfo to use _columnInfoMap known column class if available

### DIFF
--- a/api/src/org/labkey/api/reader/DataLoader.java
+++ b/api/src/org/labkey/api/reader/DataLoader.java
@@ -322,6 +322,7 @@ public abstract class DataLoader implements Iterable<Map<String, Object>>, Loade
             for (int f = 0; f < nCols; f++)
             {
                 List<Class> classesToTest = new ArrayList<>(Arrays.asList(CONVERT_CLASSES));
+                Class knownColumnClass = null; // Issue 49830: if we know the column class based on the columnInfoMap, fall back to using that class instead of defaulting to String
 
                 int classIndex = -1;
                 //NOTE: this means we have a header row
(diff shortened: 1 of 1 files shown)


cnathe:
@labkey-jeckels @labkey-matthewb I pushed this change to a draft PR for you to take a look at and see what you think. It fixes the scenario I added yesterday to the issue 49830 details re: sample type file fields. However after making this change, I'm curious if we should instead just be using the known class types from the _columnInfoMap when it is provided instead of even trying to call the ConvertUtils.convert() on the CONVERT_CLASSES. Thoughts?

labkey-matthewb:
I thought inferColumns was for the "create from file" scenario.  Why would we even call inferColumns if we have a target TableInfo?

cnathe:
> I thought inferColumns was for the "create from file" scenario. Why would we even call inferColumns if we have a target TableInfo?

Good question. It looks like `DataLoader.getColumns()` calls through to `DataLoader.initialize()` which calls `inferColumnInfo()` unless the `setColumns()` was called, which looks to only happen in a few specific places. I assume maybe the `inferColumnInfo()` is used in the file import or bulk data import cases because the incoming data might contain a subset of the target table's columns or include extra columns. But either way, it seems like if we have the known target table columns and the incoming data column header matches one of those columns, then we should use that known column type.

labkey-matthewb:
> > I thought inferColumns was for the "create from file" scenario. Why would we even call inferColumns if we have a target TableInfo?
> 
> Good question. It looks like `DataLoader.getColumns()` calls through to `DataLoader.initialize()` which calls `inferColumnInfo()` unless the `setColumns()` was called, which looks to only happen in a few specific places. I assume maybe the `inferColumnInfo()` is used in the file import or bulk data import cases because the incoming data might contain a subset of the target table's columns or include extra columns. But either way, it seems like if we have the known target table columns and the incoming data column header matches one of those columns, then we should use that known column type.

Agreed

labkey-matthewb:
also inferColumns causes us to reread part of the file/stream.  Nice to avoid when possible.

labkey-matthewb (approved):

PR #5571, opened 2024-06-07
Fix code generation

labkey-adam:
#### Rationale
Use correct helpers


labkey-matthewb (changes requested):
else needs help too.

should we start returning JavaScriptFragment from methods like this?

labkey-adam:
> else needs help too.
> 
> should we start returning JavaScriptFragment from methods like this?

Good point... fixed else case. Funny thing is there are very, very few cases where we hit that code path... requires GET or POST (not LINK) plus selection not required.

Seems like a good direction (returning JavaScriptFragment). In this case, might want to reconcile Button vs. ActionButton first.

labkey-matthewb (approved):

PR #5574, opened 2024-06-08
Make other connections detection more reliable

labkey-adam:
#### Rationale
[Issue 50574: LabKey occasionally refuses to start when bootstrapping on SQL Server 2014](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50574)

I don't have a local repro, but best guess here is the create database connection doesn't disappear immediately, even though it's closed and not pooled. Let's try detecting other connections on the very first (unpooled) connection instead of waiting for the first pooled connection. There's no reason to wait until later and this also simplifies the state handling (no need to pass around application name or involve DbScopeLoader). And no need to check for other connections when we just created the database.

Also, query and show current application name on all existing connections to reinforce the conflicts.

### DIFF
--- a/api/src/org/labkey/api/data/DbScope.java
+++ b/api/src/org/labkey/api/data/DbScope.java
@@ -1448,7 +1448,7 @@ public class DbScope
         return getTableInfoCache(options.getSchema().getType()).get(options);
     }
 
-    /** This is expensive... use only for debugging and troubleshooting */
+    /** This is expensive... use only for local debugging and troubleshooting */
     public <OptionType extends SchemaTableOptions> boolean isCached(OptionType options)
     {
         return getTableInfoCache(options.getSchema().getType()).isCached(options);
@@ -1770,7 +1770,7 @@ public class DbScope
         }
         catch (SQLException e)
         {
-            throw new RuntimeException(e);
+            throw new RuntimeSQLException(e);
         }
     }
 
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/data/DbScope.java
```diff
@@ -1746,58 +1751,63 @@ public static String ensureDatabase(LabKeyDataSource ds) throws ServletException
...
             {
-                throw new RuntimeException(e);
+                detectUnexpectedConnections(conn, ds, applicationName);
             }
         }
-        else
+        catch (SQLException e)
         {
-            detectUnexpectedConnections(primaryScope);
+            throw new RuntimeException(e);
```
labkey-jeckels: RuntimeSQLException?

api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
```diff
@@ -1987,6 +1987,6 @@ public boolean shouldTest()
     @Override
     public @NotNull String getApplicationConnectionsSql()
     {
-        return "SELECT pid, usename, client_addr, client_hostname, xact_start, query_start, state, query FROM pg_stat_activity WHERE pid <> pg_backend_pid() AND datname = ? AND application_name = ?";
+        return "SELECT pid, usename, client_addr, client_hostname, xact_start, query_start, state, application_name, query FROM pg_stat_activity WHERE pid <> pg_backend_pid() AND datname = ? AND application_name = ?";
```
labkey-jeckels, line 1990: Might be useful to also select out the PID for the current connection for future troubleshooting if this doesn't fix it. Could add it to the list of values being selected, though of course it'll be the same for every row if there are multiple

labkey-jeckels (approved):

PR #5577, opened 2024-06-10
Pass and set SecurableResource instead of SecurityPolicy

labkey-adam:
#### Rationale
We want low-level permission checking to operate on `SecurableResource`, which means we need to pass them around instead of `SecurityPolicy`. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=45251

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5575

### DIFF
--- a/api/src/org/labkey/api/util/Path.java
+++ b/api/src/org/labkey/api/util/Path.java
@@ -32,20 +32,11 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
-import static org.apache.commons.lang3.CharSetUtils.containsAny;
-
 /**
- * User: matthewb
- * Date: Nov 20, 2009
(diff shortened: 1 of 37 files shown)


api/src/org/labkey/api/util/Path.java
```diff
@@ -179,19 +170,16 @@ public static Path decode(String path)
         return new Path(arr, arr.length, _abs(path), _dir(path));
     }
 
-
```
labkey-matthewb, line 182: Aww, I like double spacing between functions

labkey-matthewb (approved):
I think with this change there would be value in not having SecurableResource no longer implement HasPermission.    Given a stand-alone SecurableResource sr, it depends on context whether sr.hasPermission() or SecurityManager.hasPermission(sr) is correct.  Maybe HasPermission should just go away and have objects implement specific canVerb() method.

PR #5578, opened 2024-06-11
Fill PDFBox font cache proactively at startup to avoid test timeouts

labkey-adam:
#### Rationale
[PDF indexing is slow on first file after server startup on Windows](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50601)

### DIFF
--- a/core/src/org/labkey/core/CoreModule.java
+++ b/core/src/org/labkey/core/CoreModule.java
@@ -23,6 +23,8 @@ import jakarta.servlet.ServletRegistration;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.pdfbox.pdmodel.font.FontMapper;
+import org.apache.pdfbox.pdmodel.font.FontMappers;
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/CoreModule.java
```diff
@@ -1274,6 +1279,24 @@ public void startBackgroundThreads()
...
         TempTableTracker.init();
+
+        // Loading the PDFBox font cache can be very slow on some agents; fill it proactively. Issue 50601
+        JobRunner.getDefault().execute(() -> {
+            try
+            {
+                long start = System.currentTimeMillis();
+                FontMapper mapper = FontMappers.instance();
+                Method method = mapper.getClass().getMethod("getProvider");
+                method.setAccessible(true);
```
labkey-jeckels, line 1290: Should we turn set this to false afterwards?

labkey-jeckels (approved):

core/src/org/labkey/core/CoreModule.java
labkey-adam: No need. The call affects only this `Method` object (not the method itself), which nobody holds once we leave this block and should be garbage collected shortly.

PR #5580, opened 2024-06-11
Update FileLinkMetricsMaintenanceTask to include metric info for valid file paths broken down by type (i.e. schema name)

cnathe:
#### Rationale
The related PR added support for a better app user experience for assay imports that includes File fields at the results domain. With that, we want to track a metric counting the valid file paths for these fields. This PR adds that to the existing FileLinkMetricsMaintenanceTask which was already setup to check for missing file paths.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5533

#### Changes
- Update FileLinkMetricsMaintenanceTask to include metric info for valid file paths broken down by type (i.e. schema name)
- doMissingFilesCheck to only track paths for the per container action (otherwise just tracking counts)

### DIFF
--- a/experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
+++ b/experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
@@ -26,7 +26,7 @@ public class FileLinkMetricsMaintenanceTask implements SystemMaintenance.Mainten
     @Override
     public String getDescription()
     {
-        return "Task to calculate metrics for missing files for File fields.";
+        return "Task to calculate metrics for valid and missing files for File fields";
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


cnathe:
Screenshot 2024-06-11 at 8 12 17 AM

Screenshot 2024-06-11 at 8 31 29 AM

experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
```diff
@@ -55,22 +55,30 @@ public void run(Logger log)
...
                 {
-                    Map<String, Set<String>> missingFiles = results.get(containerId);
-                    for (String source : missingFiles.keySet())
-                        totalCount += missingFiles.get(source).size();
+                    Map<String, MissingFilesCheckInfo> info = results.get(containerId);
+                    for (String source : info.keySet())
+                    {
+                        missingFilesCount += info.get(source).getMissingFilesCount();
+
+                        String schemaName = source.substring(0, source.indexOf('.')); // e.g. 'assayresults.c9043290_test'
```
XingY: In the case of exp.object, this just says `exp`. Do we want to add a comment to clarify `exp` currently means assay batch or run fields?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -9730,24 +9732,27 @@ public Map<String, Map<String, Set<String>>> doMissingFilesCheck(User user, Cont
...
-                        missingFiles.get(containerId).put(sourceName, new HashSet<>());
-
-                    missingFiles.get(containerId).get(sourceName).add(filePath);
+                    fileResults.get(containerId).get(sourceName).addMissingFile(filePath, trackMissingFiles);
                 }
+                else
+                    fileResults.get(containerId).get(sourceName).incrementValidFilesCount();
 
-                if (missingCount >= MAX_MISSING_COUNT)
+                if (trackMissingFiles && missingCount >= MAX_MISSING_COUNT)
```
XingY, line 9763: If trackMissingFiles = false, or if the number of valid files is huge, will there be performance concern? By default ResultSet retrieves all records in memory, should we setFetchSize to a reasonable number?

XingY (approved):

experiment/src/org/labkey/experiment/FileLinkMetricsMaintenanceTask.java
cnathe: done

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
cnathe: Currently, the trackMissingFiles props is only set to false in the case where we are gathering the counts for the JSON metric. In that case, we don't hold the full set of missing file paths in memory so memory usage there shouldn't be a problem.

@labkey-jeckels Josh, with your comment in the last / related PR (https://github.com/LabKey/platform/pull/5531) were you talking about capping the query response to a max rows or were you talking about the perf for validation of the paths in the response? We can definitely use setMaxRows on the SqlSelector here so that we cap the query response to the first 100k or 1mil rows (or whatever N we are comfortable with).

labkey-jeckels: Since we're calling `SqlSelector.getResults(false)` here we shouldn't have a problem with the cached `ResultSet` taking lots of memory. We've got `MAX_MISSING_COUNT` so we shouldn't have a problem with too much memory on the list of missing files either.

Thus, the main concern is the time it takes to loop over the list of files and ask the file system if it exists. If we exceed the WAF timeout, we won't be able to see the results, for example. Thus, capping at 100,000 as currently implemented here makes sense to me, though we can adjust this based on real-world perf and file counts.

PR #5515, opened 2024-05-16
Issue 49975: Lower and Upper bound lines are displayed incorrectly on qc plots

ankurjuneja:
#### Rationale
The upper bound layer for fixed deviation from mean is creating two lines. This PR fixes the issue.
This PR also fixes the bound lines that were shown incorrectly for percent of mean and standard deviation y axis scales for fixed value cuttoff and fixed deviation from mean metrics.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5323

#### Changes
- fix bound lines for fixed cutoff and for percent of mean & SD y axis scales
-  fix bound lines for fixed deviation and for percent of mean & SD y axis scales
- hide bound lines for fixed cuttoff for all series shown in a plot

PR #5587, opened 2024-06-14
Issue 50520: Scale min/max y-axis based on metric values and outlier bounds

ankurjuneja:
#### Rationale
For all metric types and outlier definitions, we want the y axis to use a max of the higher of the bigger metric value and the upper bound. And use a min of the lower of the smallest metric value on the lower bound. If we don't have upper or lower bounds, we will use the biggest and smallest metric values within the range being plotted.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5515

#### Changes
-


labkey-jeckels (approved):
I did some testing on the FB and it was zooming as desired.

PR #5594, opened 2024-06-18
Issue 50495: Update sample timeline audit events table to populate missing sampleTypeId values

labkey-susanh:
#### Rationale
Issue [50495](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50495):

#### Changes
- Upgrade script that finds the proper sample type id value from either other audit entries or the exp.materials table.


experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -344,4 +345,70 @@ public static void populateDataAncestors(ModuleContext context)
...
+
+    // called from exp-24.003-24.004.sql
+    public static void addMissingSampleTypeIdsForSampleTimelineAudit(ModuleContext context)
+    {
+        if (context.isNewInstall())
+            return;
+
+        DbScope scope = ExperimentService.get().getSchema().getScope();
+        List<String> tableNames = new SqlSelector(scope, "SELECT StorageTableName FROM exp.domainDescriptor WHERE StorageSchemaName='audit' AND name='" + SampleTimelineAuditProvider.SampleTimelineAuditDomainKind.NAME + "'").getArrayList(String.class);
+        if (tableNames.size() > 1)
```
cnathe, line 357: minor: I assume this should never be zero, but is it worth warning in that case as well? or maybe it could be zero if no one on the given server has ever used sample types.

cnathe (approved):
tested locally with PG and SQL Sever DBs and both steps worked as expected to update audit rows

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
labkey-susanh: I think if its zero, we'll just let it be zero without comment.

PR #5601, opened 2024-06-18
Issues 50584 & 50461: Make ExpirationDate recognized for sample type and restrict field RunId for samples and data classes

labkey-susanh:
#### Rationale
- Issue [50584](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50584)
- Issue [50461](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50461)

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/378

#### Changes
- Add `RunId` to list of reserved fields for both sample types and data classes
- Add renaming mapping for "ExpirationDate" to "MaterialExpDate" ("Expiration Date" already gets mapped because it is the field label)

### DIFF
--- a/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
+++ b/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
@@ -182,6 +182,7 @@ public class SampleTypeUpdateServiceDI extends DefaultQueryUpdateService
         SAMPLE_ALT_IMPORT_NAME_COLS = new CaseInsensitiveHashMap<>();
         SAMPLE_ALT_IMPORT_NAME_COLS.put("SampleId", "Name");
         SAMPLE_ALT_IMPORT_NAME_COLS.put("Sample Id", "Name");
+        SAMPLE_ALT_IMPORT_NAME_COLS.put("ExpirationDate", "MaterialExpDate");
     }
 
     public enum Options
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -182,6 +182,7 @@ public class SampleTypeUpdateServiceDI extends DefaultQueryUpdateService
         SAMPLE_ALT_IMPORT_NAME_COLS = new CaseInsensitiveHashMap<>();
         SAMPLE_ALT_IMPORT_NAME_COLS.put("SampleId", "Name");
         SAMPLE_ALT_IMPORT_NAME_COLS.put("Sample Id", "Name");
+        SAMPLE_ALT_IMPORT_NAME_COLS.put("ExpirationDate", "MaterialExpDate");
```
XingY, line 185: So have we decided that we don't actually support "ExpirationDate" as a custom field, if we are doing this mapping? If so, we should add it to the list of reserved names. If not, I think we should be smart about when to do the mapping (when there is no custom field) vs not.

labkey-susanh: Somewhat surprisingly, it already is restricted. You can't create a field named either "Expiration Date" or "ExpirationDate". If such a field already existed before we introduced the system field, it can continue to exist. The change here is that during import (or update), we will now map the data into our system field. Chatted with @labkey-hannah about this and we agreed to go this route and help folks to get out of their problematic configuration if it existed instead of doing something more complicated.

XingY (approved):

PR #5611, opened 2024-06-20
Issue 49845: Manage Group - Member List is Not Sorted Alphabetically

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49845

#### Changes
- add sort on UserPrincipal name to group.jsp

### DIFF
--- a/api/src/org/labkey/api/security/UserPrincipal.java
+++ b/api/src/org/labkey/api/security/UserPrincipal.java
@@ -30,7 +30,7 @@ import java.util.Set;
 /**
  * A user-oriented principal within the system. This encompasses both {@link User}s and {@link Group}s (of groups and users).
  */
-public abstract class UserPrincipal implements Principal, Parameter.JdbcParameterValue, Serializable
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/security/UserPrincipal.java
```diff
@@ -157,4 +157,10 @@ public boolean hasPrivilegedRole()
         // Check for any privileged role assigned to this principal at the root
         return ContainerManager.getRoot().getPolicy().getRoles(getGroups()).stream().anyMatch(Role::isPrivileged);
     }
+
+    @Override
+    public int compareTo(@NotNull UserPrincipal o)
+    {
+        return getName().compareTo(o.getName());
```
labkey-jeckels: I recommend `compareToIgnoreCase()`

labkey-jeckels (approved):

api/src/org/labkey/api/security/UserPrincipal.java
cnathe: done

PR #5612, opened 2024-06-20
Issue 50701: Sample Manager: cycle of derivation dependencies error when importing samples

XingY:
#### Rationale
- Async import has crossFolder set to true always, even in the absence of 'Container' column. 
- When crossFolder import is true, the parent dependency check throws a false positive cycle detection error.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5612
- https://github.com/LabKey/limsModules/pull/392

#### Changes
- moves the 'Container' column check to a later point so it's done by in the sync and async code path. 
- avoid cycle detection error when sample type column is not provided
- use MapLoader instead of List so JUnit tests align closer with actual api action

### DIFF
--- a/api/src/org/labkey/api/query/DefaultQueryUpdateService.java
+++ b/api/src/org/labkey/api/query/DefaultQueryUpdateService.java
@@ -34,6 +34,8 @@ import org.labkey.api.data.TableSelector;
 import org.labkey.api.data.UpdateableTableInfo;
 import org.labkey.api.data.validator.ColumnValidator;
 import org.labkey.api.data.validator.ColumnValidators;
+import org.labkey.api.dataiterator.DataIteratorBuilder;
+import org.labkey.api.dataiterator.DataIteratorContext;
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/query/DefaultQueryUpdateService.java
```diff
@@ -804,4 +809,23 @@ protected boolean isAttachmentProperty(String name)
             return isAttachmentProperty(dp);
         return false;
     }
+
+    protected void configureCrossFolderImport(DataIteratorBuilder rows, DataIteratorContext context) throws IOException
+    {
+        if (!context.getInsertOption().updateOnly && context.isCrossFolderImport() && rows instanceof DataLoader dataLoader)
+        {
+            boolean hasContainerField = false;
+            for (ColumnDescriptor columnDescriptor : dataLoader.getColumns())
```
labkey-susanh, line 847: Could probably use a stream for this, but not necessary.

labkey-susanh (approved):

PR #5614, opened 2024-06-21
50059: Audit column name casing changes

labkey-klum:
[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50059)

#### Rationale
Because of the way we query for the provisioned storage tables in `AbstractAuditTypeProvider.udateIndices` we end up caching the `TableInfo` but the domain fields don't get fixed up they way they would ordinarily during typical construction. Maybe this is a bit heavy handed but one solution is to uncache the domain after initial startup in order for `StorageProvisionerImpl.fixupProvisionedDomain` to run normally.

### DIFF
--- a/api/src/org/labkey/api/audit/AbstractAuditTypeProvider.java
+++ b/api/src/org/labkey/api/audit/AbstractAuditTypeProvider.java
@@ -254,7 +254,9 @@ public abstract class AbstractAuditTypeProvider implements AuditTypeProvider
                 }
 
                 updateIndices(domain, domainKind);
-                transaction.commit();
+                // Issue 50059, don't cache the DB schema table queried by updateIndices in order for the provisioned
+                // domain fields to be properly fixed up.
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/audit/AbstractAuditTypeProvider.java
```diff
@@ -254,7 +254,9 @@ protected void ensureProperties(User user, Domain domain, AbstractAuditDomainKin
                 }
 
                 updateIndices(domain, domainKind);
-                transaction.commit();
+                // Issue 50059, don't cache the DB schema table queried by updateIndices in order for the provisioned
+                // domain fields to be properly fixed up.
+                transaction.addCommitTask(() -> domainKind.invalidate(domain), DbScope.CommitTaskOption.POSTCOMMIT);
```
labkey-nicka, line 259: Did you mean to drop the `transaction.commit()` call here?

labkey-klum: Heh, yeah that's important.

labkey-susanh (approved):

PR #5627, opened 2024-06-25
Misc 24.7 fixes (issues 50710, 50715, 50672)

cnathe:
#### Rationale
- Issue [50710](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50710): DataLoader converter to treat "Name" column as a string value for sample or data class import
- Issue [50715](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50715): Unique ID column population to quote column names with spaces for update statement
- Issue [50672](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50672): FileLinkDisplayColumn to include modified date in url so app thumbnails are updated

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1521

#### Changes
- query import configureLoader to treat "Name" similar to lineage columns for converter (i.e. force as String)
- FileLinkDisplayColumn to add modified date to URL to prevent browser cache issues on update
- DomainImpl to use getLegalSelectName() in unique ID update statement

### DIFF
--- a/api/src/org/labkey/api/query/AbstractQueryImportAction.java
+++ b/api/src/org/labkey/api/query/AbstractQueryImportAction.java
@@ -572,7 +572,7 @@ public abstract class AbstractQueryImportAction<FORM> extends FormApiAction<FORM
                                 .setAuditBehaviorType(behaviorType)
                                 .setAuditUserComment(_auditUserComment)
                                 .setOptionParamsMap(getOptionParamsMap())
-                                .setHasLineageColumns(hasLineageColumns())
+                                .setAllowLineageColumns(allowLineageColumns())
                                 .setJobDescription(getQueryImportDescription())
                                 .setJobNotificationProvider(getQueryImportJobNotificationProviderName());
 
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/query/AbstractQueryImportAction.java
```diff
@@ -673,7 +673,8 @@ public static void configureLoader(DataLoader loader, @Nullable TableInfo target
                 if (name.startsWith(ExpMaterial.MATERIAL_INPUT_PARENT.toLowerCase() + "/") ||
                     name.startsWith(ExpMaterial.MATERIAL_OUTPUT_CHILD.toLowerCase() + "/") ||
                     name.startsWith(ExpData.DATA_INPUT_PARENT.toLowerCase() + "/") ||
-                    name.startsWith(ExpData.DATA_OUTPUT_CHILD.toLowerCase() + "/"))
+                    name.startsWith(ExpData.DATA_OUTPUT_CHILD.toLowerCase() + "/") ||
+                    name.equalsIgnoreCase("Name") /* Issue 50710: Treat "Name" column as a string value for sample or data class import */)
```
labkey-susanh, line 677: Looks like this will only get hit when there are lineage columns included in the import. Don't we need to do this in all cases when we're splitting the file into multiple files to be read back in again?

cnathe: The `includeLineageColumns` and the method that populates it `hasLineageColumns()` are a bit of a misnomer. They aren't based on the contents of the data being imported but based on the import action itself and the target table. So `hasLineageColumns()` really means, from what I understand, that the data tables allow import of lineage columns (i.e. Sample Type and Data Class imports). 

I was testing this with an import that didn't have any lineage columns in the tsv data and it hit this code as expected.

labkey-susanh: OK, yeah. I saw the test that was added in limsModules for this, so I guess it's good.  Definitely a strange name for that method.

labkey-susanh (approved):

api/src/org/labkey/api/query/AbstractQueryImportAction.java
cnathe: I changed it from `hasLineageColumns` to `allowLineageColumns`, still not my favorite name but seems better.

PR #5508, opened 2024-05-14
Add Ancestors node to exp.materials grid

labkey-susanh:
#### Rationale
Clients have found the Ancestors data we track to be useful and also want to be able to add this data to the various All Samples grids we present in our applications.  To accomplish this, we move away from the use of the temporary closure tables, which we had been computing for each source type, and introduce two hard tables `exp.materialAncestors` and `exp.dataAncestors`.  

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/261

#### Changes
- Add upgrade script to create new ancestors tables
- Update `ClosureQueryHelper` to operate on the new hard table

PR #5628, opened 2024-06-25
Sync ancestor nodes for multiple runs all at once

labkey-susanh:
#### Rationale
When we create many derivatives or aliquots at once, it makes more sense to batch the ancestor updates all together rather than do each run individually.  We've seen some degradation in our UI performance tests after the Ancestor implementation was updated. Locally, things perform much better with these changes.

#### Related Pull Requests
- #5508 

#### Changes
- Add method to `ClosureQueryHelper` to recompute ancestors for a set of runs.

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
+++ b/experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
@@ -497,6 +497,26 @@ public class ClosureQueryHelper
         recomputeFromSeeds(selectSeedsSql, true);
     }
 
+    private static void recomputeMaterialAncestorsForRuns(Collection<Integer> runIds)
+    {
+        var tx = getScope().getCurrentTransaction();
+        if (null != tx)
+        {
+            tx.addCommitTask(() -> recomputeMaterialAncestorsForRuns(runIds), DbScope.CommitTaskOption.POSTCOMMIT);
+            return;
+        }
+
+        SQLFragment selectSeedsSql = new SQLFragment()
+                .append("SELECT m.RowId, m.ObjectId, 'm' AS ObjectType FROM exp.material m\n")
+                .append("INNER JOIN exp.MaterialInput mi ON m.rowId = mi.materialId\n")
+                .append("INNER JOIN exp.ProtocolApplication pa ON mi.TargetApplicationId = pa.RowId\n")
+                .append("WHERE pa.RunId ");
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
```diff
@@ -531,6 +551,32 @@ public static void recomputeDataAncestorsForRun(String sourceTypeLsid, int runId
...
+                .append("WHERE pa.RunId ");
+        getScope().getSqlDialect().appendInClauseSql(selectSeedsSql, runIds);
+        selectSeedsSql
+                .append(" AND pa.CpasType = ").appendValue(ExperimentRunOutput);
+        recomputeFromSeeds(selectSeedsSql, false);
+    }
+
+    public static void recomputeAncestorsForRuns(Collection<Integer> runIds)
+    {
+        recomputeDataAncestorsForRuns(runIds);
```
labkey-nicka, line 579: If `runIds` is empty could both of these calls be skipped?

labkey-susanh: Yes, sure. (Not likely given where it's currently called from, but good to check just the same.)

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
```diff
@@ -497,6 +497,26 @@ public static void recomputeMaterialAncestorsForRun(String sourceTypeLsid, int r
...
+            tx.addCommitTask(() -> recomputeMaterialAncestorsForRuns(runIds), DbScope.CommitTaskOption.POSTCOMMIT);
+            return;
+        }
+
+        SQLFragment selectSeedsSql = new SQLFragment()
+                .append("SELECT m.RowId, m.ObjectId, 'm' AS ObjectType FROM exp.material m\n")
+                .append("INNER JOIN exp.MaterialInput mi ON m.rowId = mi.materialId\n")
+                .append("INNER JOIN exp.ProtocolApplication pa ON mi.TargetApplicationId = pa.RowId\n")
+                .append("WHERE pa.RunId ");
+        getScope().getSqlDialect().appendInClauseSql(selectSeedsSql, runIds);
```
labkey-nicka, line 514: nit: Just a note the in clause can also be appended directly on the fragment:

```java
.append("WHERE pa.RunId ").appendInClause(runIds, getScope().getSqlDialect())
```

labkey-nicka (approved):

PR #5629, opened 2024-06-26
Issue 50115: Audit logging for module properties

labkey-susanh:
#### Rationale
[Issue 50115](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=50115)

#### Changes
- Add AuditProvider and all related classes for Module Properties
- Update module properties test to verify presence of audit logs

### DIFF
--- a/api/src/org/labkey/api/module/ModuleProperty.java
+++ b/api/src/org/labkey/api/module/ModuleProperty.java
@@ -19,6 +19,9 @@ import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.json.JSONObject;
+import org.labkey.api.audit.AbstractAuditTypeProvider;
+import org.labkey.api.audit.AuditLogService;
+import org.labkey.api.audit.provider.ModulePropertiesAuditProvider;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/module/ModuleProperty.java
```diff
@@ -293,12 +296,26 @@ public void saveValue(@Nullable User user, Container c, @Nullable String value)
         validate(user, c, value);
 
         PropertyManager.PropertyMap props = PropertyManager.getWritableProperties(PropertyManager.SHARED_USER, c, getCategory(), true);
-
+        Object oldValue = props.get(getName());
+        Map<String, Object> oldProp = Map.of(getName(), oldValue == null ? "" : oldValue );
```
labkey-jeckels: Seems like it would be more convenient to have columns for the old and new values instead of storing maps with a single value.

labkey-jeckels (approved):
One suggestion but OK if you want to use maps for planned future expansion. Thanks for the test coverage too.

api/src/org/labkey/api/module/ModuleProperty.java
labkey-susanh: Yeah, went this route when I was thinking we would log all property changes that happened with one save as a single entry, but since I don't think there will ever be lots of these and the wiring for that is less straightforward, this is probably a bit of overkill.

PR #5631, opened 2024-06-27
Fix for select query logging

labkey-matthewb:
#### Rationale
Select query logging (vs detailed logging) was working for sampletype and dataclass

#### Related Pull Requests
- 

#### Changes
-


labkey-matthewb:
This will probably get changed to 24.7 PR, but wanted to tag reviewers.

PR #5633, opened 2024-06-27
Clean up StudyCache and QueryHelper

labkey-adam:
#### Rationale
We want `QueryHelper` to use `BlockingCache` and improve generics use. We want to stop querying and caching redundant lists that differ only by filter/sort. We want to stop sorting and filtering unnecessarily. We want to cache and invalidate a single map of objects per container, indexed by key.

#### Changes
* Eliminate `StudyCache`, all QueryHelper.getList() variants, and `TableInfo.getNotificationKey()` implementations. Each QueryHelper now holds a BlockingCache that caches a single "collections" object per container. This object holds a pk -> object map that supports lookup by pk and returning a collection of all objects in that container. Subclasses (VisitHelper and DatasetHelper) can choose to cache additional collections and maps to support common sorting, filtering, and lookup patterns.
* Improve generics for `StudyCachable` and `QueryHelper`. These implementations now know the PK type and the cached collection type.
* There are a lot of changed files here, but most simply reflect the `List<T>` -> `Collection<T>` change and/or generics declarations.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5519


api/src/org/labkey/api/data/TableInfo.java
```diff
@@ -606,11 +606,11 @@ void fireRowTrigger(Container c, User user, TriggerType type, boolean before, in
...
 
-    /**
-     * TableInfos that can be associated with a QueryHelper need a reliable key other than a TableInfo instance.
-     * Return null if QueryHelper is not supported.
-     */
-    Path getNotificationKey();
+    @Deprecated // No longer used
+    default Path getNotificationKey()
+    {
+        return null;
```
labkey-jeckels, line 612: If we can't completely delete this method, should it throw `UnsupportedOperationException`?

study/src/org/labkey/study/model/TreatmentManager.java
```diff
@@ -530,10 +530,10 @@ public void deleteAssaySpecimen(Container container, User user, int rowId)
         filter.addCondition(FieldKey.fromParts("AssaySpecimenId"), rowId);
         Table.delete(StudySchema.getInstance().getTableInfoAssaySpecimenVisit(), filter);
 
-        // the delete the AssaySpecimen record by RowId
+        // delete the AssaySpecimen record by RowId
         filter = SimpleFilter.createContainerFilter(container);
         filter.addCondition(FieldKey.fromParts("RowId"), rowId);
-        Table.delete(StudySchema.getInstance().getTableInfoAssaySpecimen(), filter);
+        Table.delete(StudySchema.getInstance().getTableInfoAssaySpecimen(), filter); // TODO: Uncache! (StudyManager._assaySpecimenHelper)
```
labkey-jeckels: Reminder of this TODO

labkey-jeckels (approved):

api/src/org/labkey/api/data/TableInfo.java
labkey-adam: I plan to delete this method in a quick follow-up. I just didn't feel like adding the WNPRC and LIMS repos to this branch.

study/src/org/labkey/study/model/TreatmentManager.java
labkey-adam: Thanks

labkey-adam: That was a test to see if you were actually reviewing the code. Congratulations.

PR #5636, opened 2024-06-27
AbDisc: Worklist & Instrument Instructions, Improvements 

RosalineP:
#### Rationale
See spec [here](https://docs.google.com/document/d/1vbXcwRL8Dval9ZkkUm0txsFpikp77L29ynnYxGRfKAA/edit?pli=1).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5636
- https://github.com/LabKey/limsModules/pull/412
- https://github.com/LabKey/labkey-ui-components/pull/1524
- https://github.com/LabKey/labkey-ui-premium/pull/455

#### Changes
- Create TSVArrayWriter
- Update WorkListAction and InstrumentInstructionAction to support CSV/TSV generation
- Remove -well suffix
- Support empty destination wells for worklists
- Bump versions


api/src/org/labkey/api/data/TSVArrayWriter.java
```diff
@@ -0,0 +1,55 @@
...
+import org.labkey.api.reader.ColumnDescriptor;
+import org.labkey.api.util.FileUtil;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+// This class supports generating files with duplicate column names. Consider using TSVMapWriter if
+// multiple identical column names is not an implementation concern.
+public class TSVArrayWriter extends TSVWriter
```
RosalineP, line 12: For our case we only need this one constructor, specific as it is. Not sure if it'd be a good idea to add a more generic constructor or such, as this feels a little bit weird on its own?

api/src/org/labkey/api/data/TSVArrayWriter.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public TSVArrayWriter(String fileName, ColumnDescriptor[] columns, List<Object[]> rows)
+    {
+        _fileName = fileName;
+        _columns = Arrays.stream(columns)
+                .map(ColumnDescriptor::getColumnName)
+                .collect(Collectors.toList());
+        _rows = rows.stream()
+                .map(array -> Arrays.stream(array)
+                        .map(obj -> (obj == null) ? "" : (obj instanceof String) ? (String) obj : String.valueOf(obj))
```
labkey-klum: You probably don't need to make this distinction, `String.valueOf` should be appropriate for both string and non-string cases.

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -1390,6 +1400,16 @@ public void setDestinationPlateSetId(int destinationPlateSetId)
         {
             _destinationPlateSetId = destinationPlateSetId;
         }
+
+        public String getFileType()
```
labkey-klum: Consider binding to the `enum` type to avoid having to convert between string and enum values.

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -1457,6 +1493,16 @@ public void setPlateSetId(int plateSetId)
         {
             _plateSetId = plateSetId;
         }
+
+        public String getFileType()
+        {
+            return _fileType;
```
labkey-klum, line 1484: Similar suggestion as the `WorklistForm`. There are also some similarities between the two forms, you could also consider having one extend the other (or factor out the common bits).

assay/src/org/labkey/assay/PlateController.java
```diff
@@ -1481,9 +1527,23 @@ public Object execute(InstrumentInstructionForm form, BindException errors) thro
                 ColumnDescriptor[] xlCols = PlateSetExport.getColumnDescriptors("", includedMetadataCols);
                 List<Object[]> plateDataRows = PlateManager.get().getInstrumentInstructions(form.getPlateSetId(), includedMetadataCols, getContainer(), getUser());
 
-                ArrayExcelWriter xlWriter = new ArrayExcelWriter(plateDataRows, xlCols);
-                xlWriter.setFullFileName(plateSet.getName());
-                xlWriter.renderWorkbook(getViewContext().getResponse());
+                boolean isCSV = form.getFileType().equals(FileType.CSV.name());
```
labkey-klum: Nearly duplicate code to the other action, consider some ways to factor this code out.

labkey-klum (approved):

PR #5637, opened 2024-06-27
Move Well Metadata out from under "Properties"

labkey-klum:
#### Rationale
Well metadata was previously exposed in the virtual table under the `Properties` column. This change moves those fields out into the `Well` table directly.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5637
- https://github.com/LabKey/limsModules/pull/413
- https://github.com/LabKey/labkey-ui-premium/pull/456

#### Changes
- Join WellTable directly to fields in metadata table.
- Adjust PlateManagerTest to not reference the properties FK
- Remove the virtual table used in the properties FK.
- Update PlateSetDataGenerator to remove properties references
- Update PlateManager references to properties for metadata fields


labkey-klum:
This PR can hang out until we branch for 24.7, since we don't need this until 24.8.

assay/src/org/labkey/assay/plate/query/WellTable.java
```diff
@@ -299,18 +242,6 @@ public MutableColumnInfo wrapColumn(ColumnInfo col)
         return columnInfo;
     }
 
-    @Override
-    public List<FieldKey> getDefaultVisibleColumns()
```
labkey-nicka, line 303: I see here you completely removed the override of `getDefaultVisibleColumns` but shouldn't we still be returning `defaultVisibleColumns`? Or are those now handed off in some other way?

labkey-nicka (approved):

assay/src/org/labkey/assay/plate/query/WellTable.java
labkey-klum: It gets handled in the parent class, the default behavior is all non-hidden columns. I had to override it initially to get the metadata fields to show by default.

labkey-nicka: Alright, in that case, should we remove the `defaultVisibleColumns` declaration in this class since it is no longer used?

labkey-klum: On second thought, I think your initial comment was correct. We don't want the default behavior since that would display columns we don't want to show in the default view. For example : `RowId, LSID`. I've restored the previous behavior.

PR #5643, opened 2024-07-01
select query logging

labkey-matthewb:
#### Rationale
Fixes for log all queries

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/419
- https://github.com/LabKey/nlp/pull/242
- https://github.com/LabKey/compliance/pull/219

#### Changes
-


study/src/org/labkey/study/query/SampleDatasetTable.java
```diff
@@ -122,8 +134,27 @@ private TableInfo getSamplesTable()
         return _sampleTable;
     }
 
+    // NOTE we are going to wrap a
```
labkey-adam: ???

labkey-adam (changes requested):
Seems good and not too invasive. Need to fix test failures in compliance, et al. Please finish comment in SampleDatasetTable:137.

api/src/org/labkey/api/query/FilteredTable.java
```diff
@@ -142,8 +141,21 @@ public FilteredTable(@NotNull TableInfo table, @NotNull SchemaType userSchema, @
             else
                 applyContainerFilter(getDefaultContainerFilter());
         }
+    }
+
+    private @NotNull TableRules getTableRules()
+    {
+        if (_rules == null)
+            _rules = canApplyTableRules() ? findTableRules() : TableRules.NOOP_TABLE_RULES;
```
labkey-adam, line 149: Any risk of race conditions here? Should we assign to `_rules` in the constructor instead of lazily?

labkey-adam (approved):

api/src/org/labkey/api/query/FilteredTable.java
labkey-matthewb: Explicitly in the constructor might be better.   However, there's no race condition.  getTableRules() will get called before/by applyTableRules() which happens during construction anyway.

PR #5645, opened 2024-07-02
Plate: Reformat operations

labkey-nicka:
#### Rationale
This introduces reformat operations for plates. In general, reformatting is taking some set of source plates and transforming the well information into newly generated plates. This initial implementation supports stamping, quadrant stamping and reverse quadrant stamping.

#### Changes
- Add `plate-reformat.api` endpoint
- Introduce a `LayoutEngine` which drives the iteration/transformation of plate layout operations
- Supported operations are `quadrant`, `reverseQuadrant`, and `stamp`
- Rename the commonly used `CreatePlateSetPlate` to a more appropriate name of `PlateData`
- expand `PlateManagerTest` to test additional scenarios including reformatting (WIP)

### DIFF
--- a/assay/src/org/labkey/assay/plate/layout/QuadrantOperation.java
+++ b/assay/src/org/labkey/assay/plate/layout/QuadrantOperation.java
@@ -9,22 +9,12 @@ import org.labkey.assay.plate.model.ReformatOptions;
 import java.util.ArrayList;
 import java.util.List;
 
-public class CompressOperation implements LayoutOperation
+public class QuadrantOperation implements LayoutOperation
 {
     private PlateType _sourcePlateType;
 
     @Override
-    public List<WellLayout> execute(ReformatOptions.OperationOptions options, @NotNull List<Plate> sourcePlates, PlateType targetPlateType)
-    {
-        ReformatOptions.FillStrategy fillStrategy = options.getFillStrategy();
-
-        if (ReformatOptions.FillStrategy.quadrant == fillStrategy)
-            return quadrant(sourcePlates, targetPlateType);
-
-        throw new UnsupportedOperationException(String.format("CompressOperation does not support the \"%s\" fill strategy.", fillStrategy));
-    }
-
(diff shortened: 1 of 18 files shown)


assay/src/org/labkey/assay/plate/layout/WellLayout.java
```diff
@@ -0,0 +1,34 @@
+package org.labkey.assay.plate.layout;
+
+import org.jetbrains.annotations.NotNull;
+import org.labkey.api.assay.plate.PlateType;
+
+public class WellLayout
+{
+    public record Well(int destinationRowIdx, int destinationColIdx, int sourcePlateId, int sourceRowId, int sourceColIdx) {}
```
labkey-klum: Isn't parameter 4 a `sourceRowIdx`?

assay/src/org/labkey/assay/plate/layout/QuadrantOperation.java
```diff
@@ -0,0 +1,86 @@
...
+                colOffset = _sourcePlateType.getColumns();
+            }
+
+            for (int r = 0; r < _sourcePlateType.getRows(); r++)
+            {
+                for (int c = 0; c < _sourcePlateType.getColumns(); c++)
+                    target.setWell(r + rowOffset, c + colOffset, plateRowId, r, c);
+            }
+
+            if ((i + 1) % 4 == 0)
```
labkey-klum, line 50: Do we need this check, wouldn't there always be 4 source plates so we would fall out of the loop and invoke line 52?

labkey-klum (approved):

assay/src/org/labkey/assay/plate/layout/QuadrantOperation.java
labkey-nicka: We support any positive number of plates when quadrant stamping filling left to right.

assay/src/org/labkey/assay/plate/layout/WellLayout.java
labkey-nicka: Yup. Fixed.

PR #5654, opened 2024-07-05
Generalize experimental features to optional features

labkey-adam:
#### Rationale
Experimental feature flags are used for a variety of purposes: to introduce features that aren't yet ready for prime time, to deprecate features we wish to remove, to customize product behavior in client-specific ways, etc. This PR moves our code from "experimental" feature flags to "optional" feature flags that designate their type (Experimental vs. Deprecated vs. Optional). The main goal is to provide guidance to clients that's appropriate for each type of feature flag.

This is *mostly* an internal code rename (`ExperimentalFeature*` --> `OptionalFeature*`) exercise but there are a few externally facing functional changes:

- A new "Deprecated Features" admin page shows optional features that are designated as Deprecated with appropriate messaging
- The admin action that displays Experimental Features, Deprecated Features, etc. is now `admin-optionalFeatures.view`. It now takes a `type` parameter but defaults to `Experimental`.
- The API action that queries and sets optional features is now `admin-optionalFeature.api`
- Metrics now include all available feature flags with enabled true/false, grouping by type. For example:
```
  "optionalFeatures": {
    "deprecated": {
      "stinkyBadFeature": true
    },
    "optional": {
      "localMarketingUpdates": true
    },
    "experimental": {
      "disableEnforceCsp": false,
      "harvestCovidPayrollSupport": false,
      "sqlfragment-disable-strict-checks": false,
      "experimental-app-plate-support": false,
      "StudySubSchemas": false,
      "queryMetadataLastModified": false,
      "experimental-notificationmenu": true,
      "AllowCounterGap": false,
      "queryProductProjectDataListingScoped": false,
      "experimental-biologics-chart-builder": false,
      "rDockerSandbox": false,
      "javascriptMothership": false,
      "queryProductAllFolderLookups": false,
      "LinkedDatasetCheck": false,
      "experimental-tcr-region-detection": false,
      "skipRStudioAlbWorkaround": false,
      "blockMaliciousClients": false,
      "javascriptErrorServerLogging": true,
      "experimental-biologics-requests-menu": false,
      "experimental-eln-custom-fields": false,
      "mediaIngredientReadRole": false,
      "MergeWithManagedDatasetKeys": false,
      "generic-details-url": false,
      "experimental-use-queryselect-component": false,
      "experimental-eln-folder-export-import": false,
      "disableGuestAccount": false,
      "harvestPayrollDiscrepancyReport": false,
      "resolve-property-uri-columns": true,
      "useLegacyUpdateRows": false,
      "experimental-media-folder-export-import": false,
      "experimental-sample-aliquot-selector": false,
      "disable-login-xframe-options": false
    }
  }
```
What's **not** changing:
- The "Experimental Features" admin page looks and operates the same as before, although some of the features will migrate to other type-specific pages or go away completely
- Persistence of the flags and startup property scope/names remain the same. (The type is just a run-time registration property that's not persisted. Flags can switch types easily.)

Note that we haven't yet added or designated any Deprecated or Optional features. That will come shortly.


api/src/org/labkey/api/settings/OptionalFeatureService.java
```diff
@@ -0,0 +1,168 @@
...
+    {
+        Deprecated
+        {
+            @Override
+            public HtmlString getAdminGuidance()
+            {
+                return HtmlString.unsafe(
+                    """
+                    <strong>WARNING</strong>:
+                    Deprecated features will be removed very soon, most likely for the next major release.
```
labkey-jeckels, line 74: We could consider proactively reporting on this via admin banners or other places that people like to ignore.

api/src/org/labkey/api/settings/OptionalFeatureService.java
```diff
@@ -0,0 +1,168 @@
...
+        {
+            @Override
+            public HtmlString getAdminGuidance()
+            {
+                return HtmlString.unsafe(
+                    """
+                    <strong>WARNING</strong>:
+                    Experimental features may change, break, or disappear at any time.
+                    We make absolutely no guarantees about what may happen if you turn on these experimental
+                    features. Enabling or disabling some features will require a restart of the server.
```
labkey-jeckels, line 90: The restart caveat could apply to any of the three types, right?

core/src/org/labkey/core/CoreModule.java
```diff
@@ -1131,10 +1133,9 @@ public void moduleStartupComplete(ServletContext servletContext)
             javaInfo.put("java.vm.name", System.getProperty("java.vm.name"));
             results.put("javaRuntime", javaInfo);
             results.put("applicationMenuDisplayMode", LookAndFeelProperties.getInstance(ContainerManager.getRoot()).getApplicationMenuDisplayMode());
-            results.put("experimentalFeaturesEnabled", AdminConsole.getExperimentalFeatureFlags().stream()
-                .filter(AdminConsole.ExperimentalFeatureFlag::isEnabled)
-                .map(AdminConsole.ExperimentalFeatureFlag::getFlag)
-                .collect(Collectors.toList()));
+            results.put("optionalFeaturesEnabled", AdminConsole.getOptionalFeatureFlags().stream()
+                .filter(OptionalFeatureFlag::isEnabled)
+                .collect(Collectors.groupingBy(optionalFeatureFlag -> optionalFeatureFlag.getType().name().toLowerCase(), Collectors.mapping(OptionalFeatureFlag::getFlag, Collectors.toList()))));
```
labkey-jeckels: While reporting this as a list makes sense from a pure data structure standpoint, it also means that we can't easily report on them via the metric list UI.

Alternatively, we could collect them as a map with the name as the key and the value as `true`. Optionally including features that are available but not enabled as `false`

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -9140,7 +9142,7 @@ public void setEnabled(boolean enabled)
     }
 
     @RequiresPermission(AdminOperationsPermission.class)
-    public static class ExperimentalFeatureAction extends BaseApiAction<ExperimentalFeaturesForm>
+    public static class OptionalFeatureAction extends BaseApiAction<OptionalFeatureForm>
```
labkey-jeckels, line 9145: I imagine the only place beyond the webapp itself that might hit this endpoint is the test code, but is it worth adding an alias for the old action name?

labkey-jeckels (approved):

api/src/org/labkey/api/settings/OptionalFeatureService.java
labkey-adam: In theory, yes. In practice, there's only one experimental feature that requires this (the withCounter gap thing). Any feature that requires restart is clearly marked with a warning, so the top-level warning seems unnecessary to me. I'm tempted to remove it from the exp features description.

core/src/org/labkey/core/CoreModule.java
labkey-adam: Okay, I'm all for making this easy to query. I'll make this a map of enabled features. Let me know if you prefer to have a map of enabled & disabled instead.

api/src/org/labkey/api/settings/OptionalFeatureService.java
labkey-adam: Yep, will add this once we add some deprecated features

core/src/org/labkey/core/admin/AdminController.java
labkey-adam: It would be easy to add, but I switched the webapp and the test code to use the new endpoint. I seriously doubt third-party devs are using this API.

core/src/org/labkey/core/CoreModule.java
labkey-jeckels: I recommend not splitting by enabled/disabled. It'll be easiest if a given feature has a stable JSON path. Thinking about it a little more, I recommend always including a value for each feature, whether it's `true` or `false`.

labkey-adam: Okay, pushed a change and updated description with an example

PR #5658, opened 2024-07-09
ExperimentService: consolidate object fetch and deserialization logic

labkey-nicka:
#### Rationale
This consolidates logic in `ExperimentServiceImpl` for fetching a deserializing experiment objects from the database. There is a good amount of repeated and redundant logic/code in this service implementation and it makes it difficult to say how any one action performs its function. I've been able to consolidate logic for `Data`, `ExpData`, `ExpDataProtocolInput`,  `ExpExperiment`, `ExpMaterial`, `ExpProtocol`, `ExpProtocolInput`, `ExpRun`, and `ExperimentRun` and refactor the various incantations of fetching these objects to work via the same pattern for each.

#### Changes
- Add `SimpleFilter` parameter implementations of getters for `Data`, `ExpData`, `ExpExperiment`, `ExpMaterial`, and `ExpProtocol`
- Refactor usages around these implementations

### DIFF
--- a/api/src/org/labkey/api/exp/api/ExperimentService.java
+++ b/api/src/org/labkey/api/exp/api/ExperimentService.java
@@ -155,6 +155,7 @@ public interface ExperimentService extends ExperimentRunTypeSource
 
     List<? extends ExpRun> getExpRuns(Collection<Integer> rowIds);
 
+    @Nullable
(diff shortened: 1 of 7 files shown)


labkey-klum (approved):
thanks for cleaning this up

PR #5661, opened 2024-07-10
Eliminate getNotificationKey()

labkey-adam:
#### Rationale
The `getNotificationKey()` method is no longer used

Also add an alias for `experimentalFeature.api` action to make some tests happy

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5633


labkey-jeckels (commented):
This PR doesn't actually delete `getNotificationKey()`. Isn't that the intent?

labkey-adam:
> This PR doesn't actually delete `getNotificationKey()`. Isn't that the intent?

I suppose it is...

labkey-jeckels (approved):

PR #5664, opened 2024-07-11
fix delete performance bug

labkey-matthewb:
#### Rationale
rewrite DELETE to workaround performance bug

#### Related Pull Requests
- 

#### Changes
-


labkey-susanh (approved):
Way to go, Postgres.

PR #5675, opened 2024-07-13
Issue 50620: Slow file uploads to remote NFS file root

labkey-jeckels:
#### Rationale
Some remote file systems are incredibly slow unless we can use a FileChannel->FileChannel copy. After upgrading to Spring 6 and Tomcat 10 we're now getting a wrapped Channel that doesn't offer the same perf.

#### Changes
- Experimental feature (to be marked as Optional when it's merged forward) to copy to a staging file which we can then access via a FileChannel for the fast transfer

### DIFF
--- a/api/src/org/labkey/api/ApiModule.java
+++ b/api/src/org/labkey/api/ApiModule.java
@@ -135,6 +135,7 @@ import org.labkey.api.security.PasswordExpiration;
 import org.labkey.api.security.SecurityManager;
 import org.labkey.api.security.UserManager;
 import org.labkey.api.security.ValidEmail;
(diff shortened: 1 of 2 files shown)


labkey-adam (approved):
Some BVT tests are failing with "Read timed out" errors

Should we file an issue with Tomcat and/or Spring to reconsider this wrapping?

labkey-jeckels:
@labkey-adam tests passed after a rerun. The behavior should be exactly the same unless the experimental feature is enabled, so I'm guessing it was unrelated to my FB.

Unfortunately, making this better would involve both Tomcat and then Spring changes. I also debated whether it's worth campaigning for...

PR #5677, opened 2024-07-15
Support for calculated fields in Field Editor UI for certain data types

cnathe:
#### Rationale
Part 1 of allowing create/update of calculated fields in the Field Editor UI for certain data types: sample type, data class, list, dataset, and assay design. This PR allows admins to add a calculated field to the Fields panel of the field editor and set the expression (LabKey SQL) for that calculated field. The fields are then added to the default view of those data types and can be edited/removed from the field editor similar to other user defined fields. 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1532
- https://github.com/LabKey/labkey-ui-premium/pull/468
- https://github.com/LabKey/platform/pull/5677
- https://github.com/LabKey/limsModules/pull/459
- https://github.com/LabKey/provenance/pull/179

#### Changes
- DomainUtil to include calculated columns from XML in fields array for getDomainDescriptor
- Add conceptURI for calculated fields
- Support valueExpression in Query Metadata save
- move some metadata related classes to org.labkey.api.query
- GWTDomain constructor to use getAllFields() instead of just standard fields
- DomainKind.allowCalculatedFields with default to false
- Support saveCalculatedFieldsMetadata for sample type, data class, list, dataset, assay createDomain and updateDomain cases
- Add calculatedFieldKeys to the default visible columns on supported domains
- Add calculated field label/value to item details view
- Add ProductFeature.CalculatedFields
- DomainUtil.validateProperties to also check calculatedFields for reserved and duplicate names


api/gwtsrc/org/labkey/api/gwt/client/model/GWTDomain.java
```diff
@@ -252,14 +258,33 @@ public void setContainer(String container)
...
+    {
+        if (standardFields == null)
+            standardFields = fields.stream().filter(f -> f.getValueExpression() == null).toList();
+        return standardFields;
+    }
+
+    public List<FieldType> getCalculatedFields()
+    {
+        if (calculatedFields == null)
+            calculatedFields = fields.stream().filter(f -> f.getValueExpression() != null).toList();
```
labkey-susanh, line 280: Is it possible that this is a calculated field but does not currently have an expression? Would checking for a ConceptURI here be better?

api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -320,6 +327,10 @@ public static GWTDomain<GWTPropertyDescriptor> getDomainDescriptor(Container con
             list.add(p);
         }
 
+        // add calculated columns to the list of properties
+        if (!calculatedFields.isEmpty())
+            list.addAll(calculatedFields);
```
labkey-susanh: Could skip the `if` here since adding all from an empty list is a no-op.

api/src/org/labkey/api/view/DataView.java
```diff
@@ -143,7 +143,9 @@ public DataRegion getDataRegion()
         {
             DataRegion dr = new DataRegion();
             dr.setTable(form.getTable());
-            List<ColumnInfo> allCols = form.getTable().getUserEditableColumns();
+            List<ColumnInfo> allCols = form.getTable().getColumns().stream()
+                    .filter(col -> col.isUserEditable() || col.isValueExpressionColumn())
```
labkey-susanh, line 148: Does this mean a `valueExpressionColumn` may not be user-editable? When does that happen?

api/gwtsrc/org/labkey/api/gwt/client/model/GWTDomain.java
```diff
@@ -252,14 +258,33 @@ public void setContainer(String container)
...
     }
 
-    public List<FieldType> getFields()
+    @JsonIgnore
+    public List<FieldType> getAllFields()
     {
         return fields;
     }
 
+    public List<FieldType> getFields()
```
labkey-susanh, line 270: I fear there may be confusion about `getAllFields` meaning `fields` whereas `getFields` means `standardFields`. Will a switch of names be too much of potential for breaking things in lots of places/

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -707,6 +708,9 @@ protected ContainerFilter getLookupContainerFilter()
             setGridURL(new DetailsURL(gridUrl));
         }
 
+        List<FieldKey> calculatedFieldKeys = DomainUtil.getCalculatedFieldsForDefaultView(this);
+        defaultCols.addAll(calculatedFieldKeys);
```
labkey-susanh, line 712: I assume putting these in some order relative to the other default columns is coming in a later story, right?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -885,7 +895,7 @@ public ExpSampleTypeImpl createSampleType(Container c, User u, String name, Stri
                     transaction.addCommitTask(() -> clearMaterialSourceCache(c), DbScope.CommitTaskOption.IMMEDIATE, POSTCOMMIT, POSTROLLBACK);
                     return st;
                 }
-                catch (ExperimentException eex)
+                catch (ExperimentException | MetadataUnavailableException eex)
```
labkey-susanh, line 898: I know this isn't your doing here, but I wonder if it's intentional that the exception pattern is slightly different for data classes and sample types.

api/gwtsrc/org/labkey/api/gwt/client/model/GWTDomain.java
cnathe: Checking for a conceptURI would probably be better, but I believe it is possible that a user could have added a calculated field via query XML without setting the conceptURI (i.e. by only setting the valueExpression prop). We will set the conceptURI any time the user saves fields via the field editor.

Also, it is possible for calculated field to be saved with an empty valueExpression but that comes through as "" instead of null. So I think this check is valid.

api/src/org/labkey/api/exp/property/DomainUtil.java
cnathe: done

api/src/org/labkey/api/view/DataView.java
cnathe: This change was so that we would show calculated fields in the details view for things like List rows. I'll add a comment, but we want to include both user editable columns (i.e. non-system fields) and calculated fields in the display.

api/gwtsrc/org/labkey/api/gwt/client/model/GWTDomain.java
cnathe: Yeah....this naming wasn't my favorite either. There are a lot of usages of getFields() in the Java code. Those would be easy to refactor / rename. However, since the method names here also impact the domain design response object get getDomainDesign.api my fear with renaming this is that we might miss somewhere on the client side that silently starts ignoring the renamed fields in the response. We can discuss more and come up with a plan.

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
cnathe: Yes, that will be in part 2 of the client side dev.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
cnathe: Yeah, probably not intentional. I'm not sure what all of this `DbScope.RetryPassthroughException` is about

api/gwtsrc/org/labkey/api/gwt/client/model/GWTDomain.java
labkey-susanh: Looks like the conceptURI gets added as the calculated fields are being added to the TableInfo, which seems early enough that the rest of the code shouldn't need to worry about this distinction. Did the test plan cover this kind of setup as well?

labkey-susanh (approved):

api/gwtsrc/org/labkey/api/gwt/client/model/GWTDomain.java
cnathe: Sorry I didn't see this comment before starting the merge process. I will take a look and see if I can change this to a check for conceptURI.

PR #5678, opened 2024-07-15
Issue 50834: LKSM: Cross folder import of samples with multiline text

XingY:
#### Rationale
For cross folder/type import, we read data from the original file and shred them into txt files grouped by sample type and container. This PR add surrounding quotes to the original values contains special characters such as newline or tab. Additionally, special handling for values that contains both quotes and newline/tab characters are added. 

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5678
- https://github.com/LabKey/limsModules/pull/462

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/query/QueryUpdateService.java
+++ b/api/src/org/labkey/api/query/QueryUpdateService.java
@@ -112,8 +112,7 @@ public interface QueryUpdateService extends HasPermission
         SkipInsertOptionValidation,  // (Bool) Skip assert(supportsInsertOption(context.getInsertOption())) for special scenarios (e.g., folder import uses merge action that's otherwise not supported for a table),
         PreferPKOverObjectUriAsKey,    // (Bool) Prefer getPkColumnNames instead of getObjectURIColumnName to use as keys
         SkipReselectRows, // (Bool) If true, skip qus.getRows and use raw returned rows. Applicable for CommandType.insert/insertWithKeys/update/updateChangingKeys
-        TargetContainer,
-        EscapedMultiLineText // (Bool) If true, "\"" has previously been substituted with "&quot;" as part of cross folder/type import file partition
+        TargetContainer
     }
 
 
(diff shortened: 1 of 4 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2909,6 +2911,20 @@ private Object getSerializingObject(Object data)
...
+                if (!s.startsWith("\"") || !s.endsWith("\""))
+                {
+                    if (s.contains("\""))
+                    {
+                        s = s.replaceAll("\"", "&quot;");
+                        _context.putConfigParameter(QueryUpdateService.ConfigParameters.EscapedMultiLineText, true);
+                    }
+                    s = "\"" + s + "\"";
+                }
+                return s;
```
labkey-susanh: Usually when including quotes in CSV/TSV files, we require the use of "quoted quotes" ("""), which the parser knows how to convert to a single quote.  Does that not work here?

api/src/org/labkey/api/query/QueryUpdateService.java
```diff
@@ -112,7 +112,8 @@ enum ConfigParameters
         SkipInsertOptionValidation,  // (Bool) Skip assert(supportsInsertOption(context.getInsertOption())) for special scenarios (e.g., folder import uses merge action that's otherwise not supported for a table),
         PreferPKOverObjectUriAsKey,    // (Bool) Prefer getPkColumnNames instead of getObjectURIColumnName to use as keys
         SkipReselectRows, // (Bool) If true, skip qus.getRows and use raw returned rows. Applicable for CommandType.insert/insertWithKeys/update/updateChangingKeys
-        TargetContainer
+        TargetContainer,
+        EscapedMultiLineText // (Bool) If true, "\"" has previously been substituted with "&quot;" as part of cross folder/type import file partition
```
labkey-jeckels: It feels wrong to have TSV data that includes HTML encoding. Where is that coming from?

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: quoted quotes doesn't work, but quoted double (double) quotes worked!

api/src/org/labkey/api/query/QueryUpdateService.java
XingY: The html encoding has been reverted.

This was added as a work around when we need to enclose a multiple line string in quotes, but the multiple line string contains quote itself. Turns out we can escape the inner quote by converting it into a double quote.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-jeckels: This approach looks good. We should consider consolidating to use `TSVWriter.quoteValue()`. For example, it also checks for `\r` while this implementation doesn't.

XingY: super! I didn't know this util already exists

labkey-jeckels (approved):
I haven't tested the scenario but this seems like a nice, minimal change and uses time-tested code.

labkey-susanh (approved):

PR #5682, opened 2024-07-15
Improve delete module confirmation message and logging

labkey-adam:
#### Rationale
Previously, the "delete module" action would list all schemas owned (or previously owned) by that module as being deleted, even though still-in-use schemas would be skipped. The confirmation message is now consistent with the delete operation.

Example:

image

### DIFF
--- a/api/src/org/labkey/api/module/ModuleLoader.java
+++ b/api/src/org/labkey/api/module/ModuleLoader.java
@@ -79,6 +79,7 @@ import org.labkey.api.util.FileUtil;
 import org.labkey.api.util.HtmlString;
 import org.labkey.api.util.MemTracker;
 import org.labkey.api.util.MemTrackerListener;
+import org.labkey.api.util.Pair;
 import org.labkey.api.util.Path;
 import org.labkey.api.util.StartupListener;
 import org.labkey.api.util.StringUtilsLabKey;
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -1781,6 +1772,38 @@ public void removeModule(ModuleContext context, boolean deleteFiles)
         clearUnknownModuleCount();
     }
 
+    // Divide the schemas reported by the specified module context into two lists: schemas that should be deleted and
+    // schemas that shouldn't. If module is not-null (known) then all schemas will be in the delete list. If the module
+    // is null (unknown) then the should-be-deleted list contains the schemas that no known module claims and the
+    // do-not-delete list contains schemas that known modules still manage.
+    public Pair<List<String>, List<Pair<String, String>>> getSchemaDeleteActions(@Nullable Module module, ModuleContext context)
```
labkey-jeckels: This is a pretty complex return type. Consider making a simple Java class to help keep track of what's what.

labkey-jeckels (approved):

api/src/org/labkey/api/module/ModuleLoader.java
labkey-adam: Swapped in a couple records

labkey-jeckels: Thanks, that's much more self-documenting.

PR #5683, opened 2024-07-16
Provide better error message when API authentication fails

### LINKED ISSUE #5692: [MacCoss] Issue 50841: Include detail when basic auth password matches but does not meet complexity rules
[Issue 50841](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50841)

- [x] Dev work @labkey-adam 
- [x] Fix automated tests @labkey-adam 
- [x] Test Python client API @labkey-nicka 
- [x] Test R client API @labkey-klum 
  - Valid password which doesn't meet complexity requirements. Server log shows : `failed to login: password does not meet the complexity requirements`
  - However the client is just seeing this message `"exception" : "User does not have permission to perform this operation."`. Perhaps still throwing `UnauthorizedException`?

Test scenario:
- Valid credentials
- Bad credentials
- Credentials that match the stored password but are expired
-- Note the Database Authentication "expire every five seconds" option that makes testing this easy
- Credentials that match the stored password but don't meet current complexity requirements
-- Reduce complexity requirement
-- Create user and set minimal password
-- Raise complexity requirement
-- Make API call with minimal credentials
-- Verify response includes message about complexity rule and needing to change password

labkey-adam:
#### Rationale
We want API clients to receive more helpful guidance when authentication fails. For example, password expired vs. password doesn't meet complexity requirements vs. bad credentials. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50841

#### Changes
* If Basic authentication fails, return a JSON exception with user-appropriate details. Note: This is a behavior change; previously, no exception was returned, the server simply set up a guest session.
* `AuthenticationStatus` can now tailor error messages for web UI vs. API purposes. For example, returning a message to a piece of code that suggests checking the caps lock seems unhelpful.

### DIFF
--- a/api/src/org/labkey/api/security/AuthenticationManager.java
+++ b/api/src/org/labkey/api/security/AuthenticationManager.java
@@ -25,6 +25,7 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.junit.Assert;
 import org.junit.Test;
+import org.labkey.api.action.ApiResponseWriter;
 import org.labkey.api.action.LabKeyError;
 import org.labkey.api.action.LabKeyErrorWithHtml;
 import org.labkey.api.action.LabKeyErrorWithLink;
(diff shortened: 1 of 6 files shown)


labkey-jeckels (commented):
Changes look OK but I suspect the test failures in the BVTs are related.

api/src/org/labkey/api/security/AuthenticationManager.java
```diff
@@ -1207,7 +1242,16 @@ private static void addDefaultUserLoginDelay(HttpServletRequest request, String
             return handleAuthentication(request, ContainerManager.getRoot()).getUser();
         }
 
-        return null;
+        // Basic auth has failed so send failure response in a format that APIs can consume
+        // TODO: Shouldn't our client APIs set a default "respFormat" attribute? Java library doesn't...
```
labkey-jeckels: I added `respFormat` years ago for a niche use case that could only consume XML. I think it's fine to default to JSON and not bother explicitly requesting it.

labkey-adam: This was necessary to get ClientAPITest to pass, since it sets "respFormat" to XML in a few cases, and was unhappy with hard-coded JSON response. Happy to discuss and clean this up on develop or later 24.7.x... I don't love what feels like one-off handling here.

labkey-jeckels (approved):
Please hold off on merging until we have a decision on timing. See my question to the Release Management group in the chat.

PR #5686, opened 2024-07-16
Issue 50781: LKSM: File fields in Sample Types sometimes showing as Text Fields

XingY:
#### Rationale
The file metrics query iterate though provisioned tables containsing filelink columns. In doing so, a "provisioned" table tableinfo that lacks the full domain descriptor info for the datatype is generated and got put in the cache for the schema/table and later gets used. This PR switches to use StorageProvisioner.get().getSchemaTableInfo(domain) to get the fixed up tableinfo with the correct domain.

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/experiment/src/org/labkey/experiment/FileLinkFileListener.java
+++ b/experiment/src/org/labkey/experiment/FileLinkFileListener.java
@@ -30,6 +30,9 @@ import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.dialect.SqlDialect;
 import org.labkey.api.exp.OntologyManager;
 import org.labkey.api.exp.PropertyType;
+import org.labkey.api.exp.api.StorageProvisioner;
+import org.labkey.api.exp.property.Domain;
+import org.labkey.api.exp.property.PropertyService;
 import org.labkey.api.files.FileListener;
 import org.labkey.api.files.TableUpdaterFileListener;
 import org.labkey.api.security.User;
@@ -169,7 +172,7 @@ public class FileLinkFileListener implements FileListener
     private void hardTableFileLinkColumns(final ForEachFileLinkColumn block)
     {
         // Figure out all of the FileLink columns in hard tables managed by OntologyManager
-        SQLFragment sql = new SQLFragment("SELECT dd.Container, dd.StorageTableName, dd.StorageSchemaName, pd.Name FROM ");
+        SQLFragment sql = new SQLFragment("SELECT dd.Container, dd.DomainId, dd.StorageTableName, dd.StorageSchemaName, pd.Name FROM ");
         sql.append(OntologyManager.getTinfoDomainDescriptor(), "dd");
         sql.append(", ");
         sql.append(OntologyManager.getTinfoPropertyDescriptor(), "pd");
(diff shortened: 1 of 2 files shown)


labkey-susanh (approved):

experiment/src/org/labkey/experiment/FileLinkFileListener.java
```diff
@@ -191,6 +193,8 @@ private void hardTableFileLinkColumns(final ForEachFileLinkColumn block)
                 {
                     block.exec(schema, tableInfo, pathCol, containerId);
                 }
+                // Issue 50781: LKSM: File fields in Sample Types sometimes showing as Text Fields
+                schema.getScope().invalidateTable(storageSchemaName, storageTableName, DbSchemaType.Provisioned);
```
labkey-jeckels: If we're taking this approach, can we invalidate the cache earlier, right after the null check?

labkey-jeckels (commented):
Another approach that would (I think) completely eliminate the race condition if some other code happens to pull from the cache before invalidation:

Switch the query to pull back the domain's `Id`. Then use `PropertyService.getDomain()` to get the `Domain`. Then use `domain.getDomainKind().getTableInfo()` to get the full metadata.

That's a bigger change, of course, and means that we'd need to deal with a `User`, `Container`, and `ContainerFilter`

experiment/src/org/labkey/experiment/FileLinkFileListener.java
labkey-jeckels: And/or do it in a `finally` to make sure the cache doesn't get polluted if something goes wrong?

XingY: Moved invalidate call to finally.

cnathe (approved):

cnathe:
I pulled down and verified the fix locally on PG and MSSQL for the repro scenario described in the issue.

experiment/src/org/labkey/experiment/FileLinkFileListener.java
```diff
@@ -181,8 +184,10 @@ private void hardTableFileLinkColumns(final ForEachFileLinkColumn block)
 
         new SqlSelector(OntologyManager.getExpSchema(), sql).forEachMap(row -> {
             // Find the DbSchema/TableInfo/ColumnInfo for the FileLink column
-            DbSchema schema = DbSchema.get(row.get("StorageSchemaName").toString(), DbSchemaType.Provisioned);
-            TableInfo tableInfo = schema.getTable(row.get("StorageTableName").toString());
+            String storageSchemaName = row.get("StorageSchemaName").toString();
+            DbSchema schema = DbSchema.get(storageSchemaName, DbSchemaType.Provisioned);
+            Domain domain = PropertyService.get().getDomain((Integer) row.get("DomainId"));
+            TableInfo tableInfo = StorageProvisioner.get().getSchemaTableInfo(domain);
```
labkey-susanh, line 192: Let's add a reference to the issue here.

XingY:
> StorageProvisioner.get().getSchemaTableInfo(domain)

@cnathe suggested to use StorageProvisioner.get().getSchemaTableInfo(domain) instead so we don't have to do invalidate. This appears to be a safe change and checks out with some initial testing. Let me know if you have any concerns with this approach.

labkey-jeckels (approved):
Looks good to me and I like that we don't need the cache clearing now.

labkey-susanh (approved):
I believe the change in `SampleTypeServiceImpl` is no longer strictly necessary, but seems a good idea just the same.

PR #5690, opened 2024-07-17
Update mini-profiler to filter requests by session

labkey-tchad:
#### Rationale
The mini-profiler records perf data for requests and queries. In order to organize the data for different simulations, we need to include the sessionId in the `RequestInfo` objects we collect.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/1987

#### Changes
- Add sessionId to mini-profiler `RequestInfo`s
- Add mini-profiler actions to return session-filtered request perf data


labkey-jeckels (approved):

core/src/org/labkey/core/admin/miniprofiler/MiniProfilerController.java
```diff
@@ -249,31 +249,45 @@ public Object execute(ReportForm form, BindException errors)
             LinkedHashSet<Long> ids = new LinkedHashSet<>(MemTracker.get().getUnviewed(getUser()));
             getViewContext().getResponse().setHeader("X-MiniProfiler-Ids", ids.toString());
 
+            if (req != null && !getUser().equals(req.getUser()) && !getUser().hasApplicationAdminPermission())
```
labkey-danield, line 252: Is it ever valid for the request to be null? Should there be a separate check for that?

labkey-danield (approved):

core/src/org/labkey/core/admin/miniprofiler/MiniProfilerController.java
labkey-tchad: It would mean that the form specified an id that didn't return any requests for some reason (fallen out of the RequestInfo cache maybe). Whatever the reason, this part of the mini-profiler might be going away soon so we probably don't need to worry about it much.

PR #5691, opened 2024-07-17
Issue 50842: Fix upper and lower bounds on qc plots

ankurjuneja:
#### Rationale
* Hide upper and lower bounds on the plot with a linear, log, and percent of mean scales for Levey-Jennings multi-series plots.
* Show the upper and lower bound lines for fixed cutoff value multi-series plots for linear & log but for SD & percent of mean * Hide the legend translation for SD for fixed cutoff value multi-series plots

#### Related Pull Requests
https://github.com/LabKey/targetedms/pull/933

#### Changes
-


labkey-jeckels (approved):
It would be great to introduce constants for more of these string values as they continue to proliferate.

PR #5703, opened 2024-07-22
Switch assay import to use DataIterator

labkey-jeckels:
#### Rationale
Building on earlier work to make assay import stream data to handle large files, using `DataIterator` for the streaming aligns this code with other streaming approaches.

#### Changes
- New helpers for Map-oriented DataIterating
- Switch from the temporary `ValidatingDataRowIterator` to `DataIterator`
- Stronger enforcement of `DataIterator` assumptions


api/src/org/labkey/api/dataiterator/AbstractMapDataIterator.java
```diff
@@ -149,25 +142,40 @@ public ListOfMapsDataIterator(DataIteratorContext context, Set<String> colNames,
...
+                if (!(row instanceof CaseInsensitiveMapWrapper || row instanceof ArrayListMap))
                 {
-                    // assumes all ArrayListMaps are case insensitive
-                    assert row instanceof CaseInsensitiveMapWrapper || row instanceof ArrayListMap : "all rows must be either CaseInsensitiveMapWrapper or ArrayListMap";
-                    if (row instanceof ArrayListMap listMap)
-                        listMap.setReadOnly(true);
-                    else
-                        row = Collections.unmodifiableMap(row);
-                    copy.add(row);
+                    return copyRows(rows);
```
labkey-matthewb, line 150: nice

api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -343,11 +349,13 @@ public static void insertTabDelimited(Container c,
                                           @Nullable Integer ownerObjectId,
                                           ImportHelper helper,
                                           List<PropertyDescriptor> descriptors,
-                                          ValidatingDataRowIterator rows,
+                                          DataIterator rawRows,
```
labkey-matthewb, line 336: This is great.  We should leave a TODO (and backlog task) that there really shouldn't be usages of insertTabDelimited() for the tuple-store where we have a Domain and a wrapping TableInfo.  StandardDataIteratorBuilder and TableInsertDataIteratorBuilder should understand most of this insertTabDelimited() stuff.  There's still all the inputs etc and other higher level data-handling, such as inputs, to contend with (see ExpDataIterators).

api/src/org/labkey/api/qc/TsvDataSerializer.java
```diff
@@ -46,18 +48,35 @@
...
+        {
+            if (iter.next())
+            {
+                Map<String, Object> row = iter.getMap();
+                if (columns == null)
                 {
-                    Map<String, Object> row = iter.next();
                     // write the column header
-                    List<String> columns = new ArrayList<>(row.keySet());
+                    columns = new ArrayList<>(row.keySet());
```
labkey-matthewb: It should be the same, but it is slightly odd to not ask the ColumnInfo's for the column names. ( Come to think of it, there's really not much reason to use wrapMap() here, well except that writeRow() is done already, that's a reason).

api/src/org/labkey/api/dataiterator/MapDataIterator.java
```diff
@@ -65,7 +65,7 @@ public MapDataIteratorImpl(DataIterator in, boolean mutable, Set<String> skip)
             for (int i=0 ; i<=in.getColumnCount() ; i++)
             {
                 String name = in.getColumnInfo(i).getName();
-                if (skip.contains(name))
+                if (name != null && skip.contains(name))
```
labkey-matthewb: should this be || instead of && (or just Objects.requireNonNull())?

api/src/org/labkey/api/dataiterator/AbstractMapDataIterator.java
```diff
@@ -197,6 +205,16 @@ public void close()
         }
     }
 
+    public static DataIteratorBuilder builderOf(List<Map<String, Object>> maps)
```
labkey-matthewb: Why not the static methods on class MapDataIterator?  Either way, probably should be in both places.

labkey-matthewb (changes requested):
Some small comments to consider.  It would be good to consolidate the MapDataIterator.of() methods and AbstractMapDataIterator.builderOf() methods.

api/src/org/labkey/api/dataiterator/MapDataIterator.java
labkey-jeckels: Yes

api/src/org/labkey/api/dataiterator/AbstractMapDataIterator.java
labkey-jeckels: No good reason. I consolidated to the `MapDataIterator` version.

api/src/org/labkey/api/qc/TsvDataSerializer.java
labkey-jeckels: Switched to use ColumnInfos

api/src/org/labkey/api/exp/OntologyManager.java
labkey-jeckels: Comment added and I consolidated some of the overrides. Take a look and please update the comment if I didn't capture what you wanted.

labkey-matthewb (approved):

api/src/org/labkey/api/qc/TsvDataSerializer.java
labkey-jeckels: Had to push a follow-up fix so that it noticed when a map transformer added new fields to the map. Feel free to review the latest

api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
```diff
@@ -454,6 +474,17 @@ public boolean next() throws BatchValidationException
             {
                 throw new BatchValidationException(e.getValidationException());
             }
+
```
labkey-matthewb, line 472: This does not seem like the right fix to me.  Users of the DataIterator expect to be able to inspect the column list immediately after construction.  The constructor of the _MapTransformer class should know the columns it might add to the output map.

labkey-matthewb: For instance we will construct a INSERT/UPDATE statements based on the metadata of the dataiterator.  It won't work to add columns later.

labkey-matthewb (changes requested):

labkey-matthewb (approved):

api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
labkey-jeckels: Switched to having the map transforming step declare the possible fields up front.

PR #5710, opened 2024-07-25
Convert sample crud selenium tests to api tests

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1538
- https://github.com/LabKey/platform/pull/5710
- https://github.com/LabKey/limsModules/pull/498

#### Changes
- include 'rootmaterialrowid' and other missing fields in existing record query so it can be used for aliquot rollup in merge actions
- moved some integration/utils.ts to LabKey/test 
- migrated some aliquot import test scenarios from selenium tests to SampleTypeCrud.ispec.ts

### DIFF
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -382,6 +382,41 @@ public class ExpDataIterators
             }
         }
 
+        private Pair<Boolean, Integer> determineRecalcFromExistingRecord(int i, Map<String, Object> existingMap)
+        {
+            if (_storedAmountCol == null && _unitsCol == null && _sampleStateCol == null)
+                return null; // update/merge existing will only trigger recompute if stored amount or unit or status is updated)
+
+            if (i == _parentNameToRecomputeCol) // only return lsid if existing map not null
+                return null;
+
+            Integer rootAliquot = (Integer) existingMap.get(RootMaterialRowId.name());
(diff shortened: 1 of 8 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -382,6 +382,41 @@ protected AliquotRollupDataIterator(DataIterator di, DataIteratorContext context
...
+            boolean amountChanged = !(Objects.equals(existingAmount, newAmount) && Objects.equals(existingUnits, newUnits));
+            if (!amountChanged && (_storedAmountCol == null || _unitsCol == null))
+            {
+                if (_storedAmountCol == null && !Objects.equals(existingUnits, newUnits))
+                    amountChanged = true;
+                if (_unitsCol == null && !Objects.equals(existingAmount, newAmount))
+                    amountChanged = true;
+            }
+
+            return amountChanged ? new Pair<>(true, rootAliquot) : null;
```
cnathe, line 417: I am confused by this return type. It seems like the only returned values are `null` or a `Pair` with `true` as the first param. Couldn't the return type just be the Integer and this method either `return null` or `return rootAliquot`?

cnathe (approved):

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: we want to know if amountChanged is true but rootAliquot is null so we can get it elsewhere

PR #5712, opened 2024-07-25
Link to study using visit labels

### LINKED ISSUE #5699: [LabKey CRI] Link to study support for visit labels
[spec](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50224)

When importing data into visit based studies that have a visit map configured, specifying a visit label in the imported data will correctly map to the associated sequence number. However, this is not supported in the link to study scenarios and we always expect a sequence number for sample data.

This work adds the capability to use visit labels in place of sequence numbers when linking sample data to visit based studies.

**Tasks**
-  [x] New data type for sample result domains.
- [x] Client side changes
- [x] Publish confirm action
- [x] Sample type field key resolution
- [x] Derived sample support
- [x] Auto-link support
- [x] Feature verification @labkey-hannah @jryurkanin 
- [x] Manual test @labkey-sweta 
- [x] Automation test

> - [x] Create new CreateNewVisitPage
> - [x] testVisitLabelLinkToStudy() test case
> - [x] testVisitLabelAutoLinkToStudy() test case

- [x]  Code review for automation
- [x] Code review & merge

labkey-klum:
#### Rationale
[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50224)

This change will allow samples to use a visit label (instead of sequence numbers) to link to visit based studies. A `Visit Label` data type is introduced to the sample type domain editor. At link to study time, if the target study is visit based we will try to use the `Visit Label` value to match entries in the visit map of the target study. This matches the current behavior that we have during dataset import.

If both `Visit ID` and `Visit Label` data types are specified in the source sample type, we will try to resolve a sequence number before the visit label. The new data type will also work in the auto link scenarios as well as a derivation scenario (similar to the existing timepoint options).

#### Related Pull Requests
https://github.com/LabKey/labkey-ui-components/pull/1540
https://github.com/LabKey/testAutomation/pull/1997


study/src/org/labkey/study/controllers/publish/SampleTypePublishConfirmAction.java
```diff
@@ -74,55 +73,6 @@ public void setSampleType(ExpSampleType sampleType)
         }
     }
 
-    /**
```
labkey-klum, line 77: We don't need this for sample types, and can just rely on the field key map.

api/src/org/labkey/api/study/publish/StudyPublishService.java
```diff
@@ -72,18 +72,19 @@ public interface StudyPublishService
...
         SpecimenId,             // legacy specimen match columns
         SpecimenMatch,
         SpecimenPtid,
         SpecimenVisit,
-        SpecimenDate,
         TargetStudy,            // for assays the run level target study value
-        SampleId
+        VisitId,
+        VisitLabel              // used for translating visit label to sequence num for visit maps
     }
```
labkey-klum, line 88: only ordering changes here

api/src/org/labkey/api/study/query/PublishResultsQueryView.java
```diff
@@ -301,51 +305,27 @@ public static class ResolverHelper
...
                               Dataset.PublishSource publishSource,
-                              ColumnInfo sourceIdCol, ColumnInfo objectIdCol,
-                              ColumnInfo ptidCol, ColumnInfo visitIdCol, ColumnInfo dateCol,
-                              ColumnInfo specimenIDCol, ColumnInfo assayMatchCol,
-                              ColumnInfo specimenPTIDCol, ColumnInfo specimenVisitCol, ColumnInfo specimenDateCol,
-                              ColumnInfo targetStudyCol,
-                              ColumnInfo sampleIdCol)
+                              Map<LinkToStudyKeys, ColumnInfo> linkedColumnMap,
+                              ColumnInfo objectIdCol)
         {
```
labkey-klum, line 323: Most of the changes in this class were to transition away from passing in a sea of parameters in favor of a single key to `ColumnInfo` map.

study/src/org/labkey/study/assay/StudyPublishManager.java
```diff
@@ -1245,6 +1266,18 @@ public void autoLinkSamples(ExpSampleType sampleType, List<Map<FieldKey, Object>
...
                     LinkToStudyKeys timePointKey = visitBased ? LinkToStudyKeys.VisitId : LinkToStudyKeys.Date;
+                    Map<String, BigDecimal> translateMap = Collections.emptyMap();
+
+                    if (visitBased)
+                    {
+                        // try visit label if we don't have visit ID
+                        if (!publishKeys.containsKey(LinkToStudyKeys.VisitId) && publishKeys.containsKey(LinkToStudyKeys.VisitLabel))
+                        {
+                            timePointKey = LinkToStudyKeys.VisitLabel;
+                            translateMap = StudyService.get().getVisitImportMap(study, true);
```
XingY: If the set of samples to link is small, say only one, is it too costly to query for all visits just to get the visit that matches one label?

XingY (approved):

PR #5713, opened 2024-07-25
Stream runs from the ExpRuns table

Sigmonia:
#### Rationale
Attempts to address [Issue 50850](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=50850) by preventing caching and streaming runs from the ExpRuns table.

#### Related Pull Requests
- https://github.com/LabKey/commonAssays/pull/782

#### Changes
- Added a new overload of getExpRuns and modified existing one to use it

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -544,7 +544,7 @@ public class ExperimentServiceImpl implements ExperimentService, ObjectReference
                 + " WHERE ER.Container = ? ");
         sql.add(container.getId());
 
-        if (null != filterSQL)
+        if (null != filterSQL && !filterSQL.isEmpty())
             sql.append(" AND " ).append(filterSQL);
 
         sql.append(" ORDER BY ER.RowId ");
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -494,24 +494,51 @@ public class ExperimentServiceImpl implements ExperimentService, ObjectReference
     @Override
     public List<ExpRunImpl> getExpRuns(Container container, @Nullable ExpProtocol parentProtocol, @Nullable ExpProtocol childProtocol)
     {
-        SQLFragment sql = new SQLFragment(" SELECT ER.* "
-                + " FROM exp.ExperimentRun ER "
-                + " WHERE ER.Container = ? ");
-        sql.add(container.getId());
+        return getExpRuns(container, parentProtocol, childProtocol, run -> true);
+    }
+
+    @Override
+    public List<ExpRunImpl> getExpRuns(Container container, @Nullable ExpProtocol parentProtocol, @Nullable ExpProtocol childProtocol, @NotNull Predicate<ExpRun> filterFn)
+    {
+
+        SQLFragment sql = new SQLFragment();
         if (parentProtocol != null)
         {
-            sql.append("\nAND ER.ProtocolLSID = ?");
+            sql.append("\nER.ProtocolLSID = ?");
             sql.add(parentProtocol.getLSID());
         }
         if (childProtocol != null)
         {
-            sql.append("\nAND ER.RowId IN (SELECT PA.RunId "
+            if (parentProtocol != null)
+                sql.append(" AND ");
+
+            sql.append("\nER.RowId IN (SELECT PA.RunId "
                     + " FROM exp.ProtocolApplication PA "
                     + " WHERE PA.ProtocolLSID = ? ) ");
             sql.add(childProtocol.getLSID());
         }
+
+        return getExpRuns(container, sql, filterFn);
+    }
+
+    @Override
+    public List<ExpRunImpl> getExpRuns(@NotNull Container container, @Nullable SQLFragment filterSQL, @NotNull Predicate<ExpRun> filterFn)
+    {
+        SQLFragment sql = new SQLFragment(" SELECT ER.* "
+                + " FROM exp.ExperimentRun ER "
+                + " WHERE ER.Container = ? ");
+        sql.add(container.getId());
+
+        if (null != filterSQL)
+            sql.append(" AND " ).append(filterSQL);
+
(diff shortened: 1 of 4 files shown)


labkey-matthewb (approved):

experiment/src/org/labkey/experiment/xar/FolderXarWriterFactory.java
```diff
@@ -98,29 +98,36 @@ public boolean show(Container c)
...
             // Also don't include recipe protocols; there's a separate folder writer and importer for the recipe module.
             // if an additional context has been furnished, filter out runs not included in this export
 
-            List<ExpRun> allRuns = (List<ExpRun>) ExperimentService.get().getExpRuns(c, null, null,
-                        run -> !run.getProtocol().getLSID().equals(ExperimentService.SAMPLE_DERIVATION_PROTOCOL_LSID)
-                                && !run.getProtocol().getLSID().equals(ExperimentService.SAMPLE_ALIQUOT_PROTOCOL_LSID)
-                                && !"recipe".equalsIgnoreCase(run.getProtocol().getImplementationName())
-                                && !"recipe".equalsIgnoreCase(run.getProtocol().getLSIDNamespacePrefix())
-                    );
+            List<ExpRun> allRuns = (List<ExpRun>) ExperimentService.get().getExpRuns(c, null, null, this::runFilter);
```
labkey-jeckels: Since this is inside of the Experiment module, you can use ExperimentServiceImpl to avoid needing to cast. You could also get away without needing to add `hasRuns()` to the interface if that's appealing

```suggestion
            List<ExpRun> allRuns = ExperimentServiceImpl.get().getExpRuns(c, null, null, this::runFilter);
```

experiment/src/org/labkey/experiment/xar/FolderXarWriterFactory.java
```diff
@@ -98,29 +98,36 @@ public boolean show(Container c)
             return children;
         }
 
-        private List<ExpRun> getRuns(@Nullable FolderExportContext ctx, Container c)
+        private boolean hasRuns(Container c)
+        {
+            return ExperimentService.get().hasExpRuns(c, this::runFilter);
+        }
+
+        private boolean runFilter(ExpRun run) {
```
labkey-jeckels, line 106: This could also be expressed as part of the SQL by specifying as a WHERE clause. ExperimentServiceImpl has a private method that takes a SimpleFilter.

labkey-jeckels (approved):
This is clearly an improvement over the current behavior. However, unless many of the runs involved are derivation runs (maybe we're lucky and it is for this repro scenario), it will just delay the problem by one click until the user clicks to start the export.

experiment/src/org/labkey/experiment/xar/FolderXarWriterFactory.java
labkey-nicka: @labkey-jeckels the variant that takes a SimpleFilter is not available in 24.7. It was added with #5658.

Sigmonia: I had that originally, but it seemed like it was leaking SQL into the FolderXarWriter

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -494,24 +494,66 @@ public ExpRunImpl getExpRun(String lsid)
...
+    public boolean hasExpRuns(Container container, @NotNull Predicate<ExpRun> filterFn)
+    {
+        SQLFragment sql = new SQLFragment(" SELECT ER.* "
+                + " FROM exp.ExperimentRun ER "
+                + " WHERE ER.Container = ? ");
+        sql.add(container.getId());
+
+        return new SqlSelector(getSchema(), sql)
+                .setJdbcCaching(false)
+                .stream(ExperimentRun.class)
```
labkey-jeckels: `stream()` gets everything as a list and then streams the list. Suggest switching to `uncachedStream()` which needs to be closed.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -494,24 +494,66 @@ public ExpRunImpl getExpRun(String lsid)
...
+        sql.add(container.getId());
+
+        if (null != filterSQL && !filterSQL.isEmpty())
+            sql.append(" AND " ).append(filterSQL);
+
         sql.append(" ORDER BY ER.RowId ");
-        return ExpRunImpl.fromRuns(new SqlSelector(getSchema(), sql).getArrayList(ExperimentRun.class));
+        return new SqlSelector(getSchema(), sql)
+                .setJdbcCaching(false)
+                .stream(ExperimentRun.class)
```
labkey-jeckels: Same suggestion on `uncachedStream()`

labkey-jeckels (commented):
Pondering benefits/risks of 24.7 vs develop:

Benefit:
If switched to uncached stream, eliminate memory load and greatly reduce time to render for export confirmation page
Cut memory requirements for actual export in half (still holding all post-filtered runs in memory, but no longer holding filtered runs or the separate ResultSet copy)

Risk:
Potential merge conflicts due to other changes already made in develop
Export still holds a lot of data in memory
Any bugs lurking in the implementation

And of course any benefit is contingent on deploying 24.7.x on servers

I'm still leaning toward 24.7 but welcome pushback

Sigmonia:
Context for 24.7.x:  I chose 24.7 as the target branch, since that was where the ticket was targeted. It's a fairly new client, the issue was initially blocking, and at a minimum this should allow the export page to load for them... @jryurkanin any thoughts?

TBC: I am waiting until after the current branch freeze is lifted.

labkey-adam:
I don't have a strong preference for 24.7 vs. develop, though I also lean toward 24.7, especially given we should have two weeks+ for this to stabilize. (Unless there's another accelerated maintenance release...)

jryurkanin:
I would prefer 24.7 so that the client gets the fix faster.

labkey-jeckels:
@labkey-ians please go ahead and merge to 24.7 when the freeze is lifted and the build errors shown in the PR are resolved. @labkey-klum heads up that there will likely be conflicts that need to be resolved when merging to develop

PR #5714, opened 2024-07-25
Issue 50896: Fix population of ancestors when samples are in different containers

labkey-susanh:
#### Rationale
The code to populate the ancestors table was chunking up the updates by sample type and container but the logic is flawed and picks up only the samples in the container where the sample type is defined. 

This adds an upgrade script to rerun the populate task with repaired logic. Since there are currently no additional upgrade scripts on develop, this seems safe to do at this juncture.

#### Related Pull Requests
- #5508 

#### Changes
- Fix logic for chunking up the ancestor table inserts
- Add upgrade script to repopulate the ancestor tables

### DIFF
--- a/experiment/resources/schemas/dbscripts/postgresql/exp-24.002-24.003.sql
+++ b/experiment/resources/schemas/dbscripts/postgresql/exp-24.002-24.003.sql
@@ -7,8 +7,6 @@ CREATE TABLE exp.MaterialAncestors
     CONSTRAINT FK_MaterialAncestors_MaterialId FOREIGN KEY (RowId) REFERENCES exp.Material (RowId) ON DELETE CASCADE
 );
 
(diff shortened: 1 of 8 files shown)


XingY (approved):
Does that mean there is a test gap for cross folder scenarios? Thought I had done plenty of that in manual testing..

labkey-susanh:
> Does that mean there is a test gap for cross folder scenarios? Thought I had done plenty of that in manual testing..

This bug is specifically about the upgrade that is creating the initial entries in this table. Any edits to sample or source lineage will create the requisite ancestor table entries.  We don't have automated testing for upgrade scenarios, but this client scenario of having all data in subfolders is one we should keep in mind for manual testing (I've added it to the client scenarios doc).

labkey-matthewb (approved):

PR #5715, opened 2024-07-26
Issue 7527: Auto-detect missing sql views and attempt to recreate

labkey-jeckels:
#### Rationale
While many module schema upgrade problems can't be auto-recovered, we can automatically recreate database views due to our paired `drop` and `create` SQL scripts

#### Changes
- If there's a schema mismatch associated with a DB view (per schema XML metadata), automatically try to drop and recreate all DB views once
- Tolerate a missing `core.Users` and/or `core.ActiveUsers` during check for root admin


labkey-jeckels:
@labkey-adam no rush to review this - anytime next week is great.

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -1706,6 +1697,11 @@ private void ensureAtLeastOneRootAdminExists()
             if (UserManager.getActiveRealUserCount() > 0)
                 SecurityManager.ensureAtLeastOneRootAdminExists();
         }
+        catch (BadSqlGrammarException e)
+        {
+            // Issue 7527: Auto-detect missing sql views and attempt to recreate
+            _log.warn("core.Users and/or core.ActiveUsers DB views are likely missing. Skipping check for root admins. Schema validation will attempt to restore views");
```
labkey-nicka: It'd be cool if we could devise a way to have automated tests for upgrade/start-up scenarios like this.

labkey-jeckels: Agreed. I marked the issue as desiring automation and captured a possible test script.

labkey-adam (commented):
I'm not dead set on this, but I think my preference would be to check and recreate the views more proactively, before startup is complete. This would be fast: we can easily determine the schemas with create or drop scripts and check just those.

core/src/org/labkey/core/view/template/bootstrap/CoreWarningProvider.java
```diff
@@ -89,6 +94,16 @@ public void startSchemaCheck(int delaySeconds)
...
                 var schemaWarnings = TableXmlUtils.compareXmlToMetaData(schema, false, false, true);
+                if (schemaWarnings.hasViewProblem() && !_attemptedViewReload)
+                {
+                    // Issue 7527: Auto-detect missing sql views and attempt to recreate
+                    LOG.warn("At least one database view in the {} schema is not as expected. Attempting to recreate views automatically.", schema.getName());
+
+                    // Only attempt to recreate views once
+                    _attemptedViewReload = true;
+                    ModuleLoader.getInstance().recreateViews();
+                    return;
```
labkey-adam: If we stick with recreating views during this schema check, wouldn't we want to restart the full schema check at this point instead of just returning? Did the recreate fix all problems or not?

labkey-jeckels: Yes! This is happening already. See the call to `WarningService.get().rerunSchemaCheck()` at the end of `ModuleLoader.recreateViews()`

labkey-adam: Right, got it

labkey-adam:
Here's a "proactive" alternative check that would work as the first code in `CoreModule.startupAfterSpringConfig()`:
```
ModuleLoader.getInstance().getModules().stream()
    .map(FileSqlScriptProvider::new)
    .flatMap(p -> p.getSchemas().stream()
        .filter(schema-> SchemaUpdateType.Before.getScript(p, schema) != null || SchemaUpdateType.After.getScript(p, schema) != null)
    )
    .filter(schema -> TableXmlUtils.compareXmlToMetaData(schema, false, false, true).hasViewProblem())
    .findAny()
    .ifPresent(schema -> ModuleLoader.getInstance().recreateViews());
```
Need to make `SchemaUpdateType.getScript()` public as well.

labkey-jeckels:
> Here's a "proactive" alternative check that would work as the first code in `CoreModule.startupAfterSpringConfig()`:
> 
> ```
> ModuleLoader.getInstance().getModules().stream()
>     .map(FileSqlScriptProvider::new)
>     .flatMap(p -> p.getSchemas().stream()
>         .filter(schema-> SchemaUpdateType.Before.getScript(p, schema) != null || SchemaUpdateType.After.getScript(p, schema) != null)
>     )
>     .filter(schema -> TableXmlUtils.compareXmlToMetaData(schema, false, false, true).hasViewProblem())
>     .findAny()
>     .ifPresent(schema -> ModuleLoader.getInstance().recreateViews());
> ```
> 
> Need to make `SchemaUpdateType.getScript()` public as well.

Thanks. I was worried that this would be a drag on startup speed. In my testing, though, it's neutral. It also gets rid of the need to special-case the admin user check and prevents other error from being logged.

labkey-adam (approved):
How can I not approve this?

PR #5717, opened 2024-07-26
Add audit event for ELN custom label changes

XingY:
#### Rationale
We want to add audit events for when ELN labels are updated/reset. The events will be under "Site settings events", which is not currently exposed in the application. For 24.7, we will add the audit log. Exposing the event type in apps will be evaluated post 24.7.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5717
- https://github.com/LabKey/limsModules/pull/503

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/settings/AbstractCustomLabelProvider.java
+++ b/api/src/org/labkey/api/settings/AbstractCustomLabelProvider.java
@@ -2,10 +2,16 @@ package org.labkey.api.settings;
 
 import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.Nullable;
+import org.labkey.api.audit.AuditLogService;
+import org.labkey.api.audit.AuditTypeEvent;
+import org.labkey.api.audit.provider.SiteSettingsAuditProvider;
 import org.labkey.api.data.Container;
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/settings/AbstractCustomLabelProvider.java
```diff
@@ -88,18 +99,82 @@ public void saveLabels(HashMap<String, String> updatedLabels, @Nullable Containe
...
+        {
+            html.append("<tr><td class='labkey-form-label'>");
+            html.append(PageFlowUtil.filter(entry.getKey()));
+            html.append("</td><td>");
+            html.append(PageFlowUtil.filter(entry.getValue().first));
+            html.append("&nbsp;&raquo;&nbsp;");
+            html.append(PageFlowUtil.filter(entry.getValue().second));
+            html.append("</td></tr>");
+        }
+        html.append("</table>");
```
labkey-susanh, line 167: This is in keeping with how changes are stored for other events in this audit type, but it's certainly not my favorite way to do this. Usually we let display columns take care of the formatting and just store raw data, but now is not the time to change that.

labkey-susanh (approved):

PR #5731, opened 2024-08-02
Issue 50939: Mutation event used in DataRegion removed from browsers

labkey-jeckels:
#### Rationale
Chrome and Firefox have dropped support for a collection of DOM listeners. We rely on one of them to make sure that LKS-rendered grids lock their column headers correctly.

#### Changes
- Switch to a MutationObserver

### DIFF
--- a/api/webapp/clientapi/dom/DataRegion.js
+++ b/api/webapp/clientapi/dom/DataRegion.js
@@ -4115,7 +4115,8 @@ if (!LABKEY.DataRegions) {
             timeout,
             locked = false,
             lastLeft = 0,
-            pos = [ 0, 0, 0, 0 ];
+            pos = [ 0, 0, 0, 0 ],
+            domObserver = null;
 
         // init
         var floatRow = headerRow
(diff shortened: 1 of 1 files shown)


api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -4183,8 +4184,8 @@ if (!LABKEY.DataRegions) {
                     .unbind('load', domTask)
                     .unbind('resize', resizeTask)
                     .unbind('scroll', onScroll);
-            $(document)
-                    .unbind('DOMNodeInserted', domTask);
+            
+            domObserver.disconnect();
```
labkey-nicka: Should probably be defensive here.

```js
if (domObserver) {
    domObserver.disconnect();
    domObserver = null;
}

labkey-nicka (approved):

api/webapp/clientapi/dom/DataRegion.js
labkey-jeckels: Good suggestion. Committed

PR #5745, opened 2024-08-07
study import canUserAccessPhi

labkey-matthewb:
#### Rationale
Fix for import dataset with PHI columns and PHI not enabled.

#### Related Pull Requests
Same fix as for lists in https://github.com/LabKey/platform/pull/5689

#### Changes
-

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -87,7 +87,9 @@ import org.labkey.api.cache.CacheStats;
 import org.labkey.api.cache.TrackingCache;
 import org.labkey.api.cloud.CloudStoreService;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
+import org.labkey.api.compliance.ComplianceFolderSettings;
 import org.labkey.api.compliance.ComplianceService;
+import org.labkey.api.compliance.PhiColumnBehavior;
 import org.labkey.api.data.ButtonBar;
 import org.labkey.api.data.ColumnInfo;
 import org.labkey.api.data.ConnectionWrapper;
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -1626,11 +1628,12 @@ public void setFileName(String fileName)
     @RequiresPermission(AdminPermission.class)
     public class ViewValidationResultsAction extends SimpleViewAction<ViewValidationResultsForm>
     {
-        @Override
```
labkey-adam, line 1629: Why is `@Override` being removed?

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -4970,7 +4973,11 @@ protected HttpView<?> getTabView(ExportFolderForm form, boolean reshow, BindExce
         {
             form.setExportType(PageFlowUtil.filter(getViewContext().getActionURL().getParameter("exportType")));
 
-            form.setExportPhiLevel(ComplianceService.get().getMaxAllowedPhi(getContainer(), getUser()));
+            ComplianceFolderSettings settings = ComplianceService.get().getFolderSettings(getContainer(), User.getAdminServiceUser());
+            PhiColumnBehavior columnBehavior = null==settings ? PhiColumnBehavior.show : settings.getPhiColumnBehavior();
+            PHI maxAllowedPhiForExport = PhiColumnBehavior.show == columnBehavior ? PHI.Restricted : ComplianceService.get().getMaxAllowedPhi(getContainer(), getUser());
+            form.setExportPhiLevel(maxAllowedPhiForExport);
+
```
labkey-adam, line 4981: Should we push all this into a different method on ComplianceService? Maybe something for the future.

labkey-adam (approved):

core/src/org/labkey/core/admin/AdminController.java
labkey-matthewb: oops?

core/src/org/labkey/core/admin/AdminController.java
labkey-matthewb: yes, after the dust settles on the spec...

PR #5746, opened 2024-08-07
Naming pattern: Enable Source-based naming pattern regardless of ancestry

XingY:
#### Rationale
The existing ancestor syntax supported for naming expression is hard to use and requires exact path match that's too strict. We are building a new set of ancestor search syntax that allows flexible ancestor search with ease of use.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5746
- https://github.com/LabKey/limsModules/pull/549

#### Changes
- add metrics reporting for various naming pattern syntax usages
- add support for ~MaterialInputs and ~DataInputs ancestor search syntax built on top of existing exact ancestor path search


api/src/org/labkey/api/exp/api/ExpLineage.java
```diff
@@ -203,6 +204,15 @@ public Set<Identifiable> findAncestorObjects(ExpRunItem seed, List<Pair<ExpLinea
         return ExpLineageTree.getNodes(ancestorTree, ancestorPaths, user);
     }
 
+    public List<Identifiable> findAncestorByType(ExpRunItem seed, Pair<ExpLineageOptions.LineageExpType, String> ancestorType, User user)
+    {
+        if (!_seeds.contains(seed))
+            throw new UnsupportedOperationException();
+
+        ExpLineageTree ancestorTree = ExpLineageTree.getAncestorTree(this, seed);
+        return ExpLineageTree.getNodes(ancestorTree, ancestorType, user);
```
labkey-susanh, line 213: Is `ancestorType` nullable? If so, you could exit before getting the ancestor tree. If not, perhaps annotate the parameter as such.

api/src/org/labkey/api/exp/api/ExpLineage.java
```diff
@@ -624,5 +634,23 @@ public static Set<Identifiable> getNodes(ExpLineageTree tree, List<Pair<ExpLinea
 
             return targetNodes;
         }
+
+        public static List<Identifiable> getNodes(ExpLineageTree tree, Pair<ExpLineageOptions.LineageExpType, String> ancestorType, User user)
```
labkey-susanh: Annotate as `@Nullable`?

experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -225,6 +225,9 @@ protected void init()
 
         ExperimentService.get().registerExperimentDataHandler(new DefaultExperimentDataHandler());
         ExperimentService.get().registerProtocolInputCriteria(new FilterProtocolInputCriteria.Factory());
+        ExperimentService.get().registerNameExpressionType("sampletype", ExperimentService.get().getTinfoSampleType(), "nameexpression");
+        ExperimentService.get().registerNameExpressionType("aliquots", ExperimentService.get().getTinfoSampleType(), "aliquotnameexpression");
+        ExperimentService.get().registerNameExpressionType("dataclass", ExperimentService.get().getTinfoDataClass(), "nameexpression");
```
labkey-susanh, line 230: I guess it's OK to stash these TInfos since they are system tables that will need a server restart to be changed.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -763,20 +809,35 @@ public static boolean isLineageInputWithDataType(String[] parts, @Nullable Strin
...
+                isData = !isMaterial;
+                String dataTypeLsid = ancestorSearchType.second;
+                inputDataType = isMaterial ? sampleTypeNames.get(dataTypeLsid) : dataClassNames.get(dataTypeLsid);
+            }
+            else if (ancestorPaths != null && !ancestorPaths.isEmpty())
+            {
+                isAncestor = true;
+                Pair<ExpLineageOptions.LineageExpType, String> ancestorType = ancestorPaths.get(ancestorPaths.size() - 1);
+                isMaterial = ExpLineageOptions.LineageExpType.Material == ancestorType.first;
+                isData = ExpLineageOptions.LineageExpType.Data == ancestorType.first;
```
labkey-susanh, line 847: Is there a reason this is not also `!isMaterial` as above?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -873,35 +936,36 @@ else if (dataType instanceof ExpDataClass)
...
                 {
-                    _syntaxErrors.add("Invalid lineage lookup: " + fkTok.toString() + ".");
+                    _syntaxErrors.add("Invalid parent lookup: " + fieldKeyDisplay + ".");
                     return null;
                 }
             }
         }
 
-        _syntaxErrors.add("Lineage lookup field does not exist: " + fkTok.toString());
+        _syntaxErrors.add("Parent lookup field does not exist: " + fieldKeyDisplay);
```
labkey-susanh: This change from "lineage" to "parent" permeates this PR and will change the message the user sees. Was previous messaging seen as confusing? "Lineage" is actually more accurate than "parent", I think.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1105,7 +1182,7 @@ else if (fieldParts.size() > 2)
                     {
                         if (_parentTable == null && domainFields.isEmpty())
                         {
-                            String errorMsg = "Parent table is required for name expressions with lookups: " + fkTok + ".";
+                            String errorMsg = "Parent table is required for name expressions with lookups: " + fieldKeyDisplay + ".";
```
labkey-susanh:
```suggestion
                            String errorMsg = "Parent table is required for naming patterns with lookups: " + fieldKeyDisplay + ".";
```

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1235,56 +1312,104 @@ else if (pt != null)
...
+            boolean isMaterialAncestor = isMaterialStr.toLowerCase().contains("material");
+            if (!StringUtils.isEmpty(typeStr))
+            {
+                String dataTypeLsid = isMaterialAncestor ? sampleTypeLSIDs.get(typeStr) : dataClassLSIDs.get(typeStr);
+                Pair<ExpLineageOptions.LineageExpType, String> ancestorType = new Pair<>(isMaterialAncestor ? ExpLineageOptions.LineageExpType.Material : ExpLineageOptions.LineageExpType.Data, dataTypeLsid);
+                partAncestorOptions.put(fkTok.encode(), new NameExpressionAncestorPartOption(options, null, ancestorType, null, fieldParts.get(fieldParts.size() - 1)));
+            }
+            else
+            {
+                _syntaxErrors.add("Invalid substitution token, ancestor " + (isMaterialAncestor ? "sample type" : "dataclass") + " must be specified in lineage search syntax: ${" + fkTokDisplay + "}.");
```
labkey-susanh:
```suggestion
                _syntaxErrors.add("Invalid substitution token. Ancestor " + (isMaterialAncestor ? "sample type" : "dataclass") + " must be specified in lineage search syntax: ${" + fkTokDisplay + "}.");
```

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1235,56 +1312,104 @@ else if (pt != null)
...
                 {
-                    _syntaxErrors.add("Invalid substitution token, a max of 10 generations of ancestor lookup is supported: ${" + fkTokDisplay + "}.");
-                    return fieldParts;
+                    hasLookupColumn = false;
+                    ancestorLevel++;
+
+                    if (partInd == 0)
+                    {
+                        // Syntax should be ${MaterialInput/..[MaterialInputs]/name} where the first input is the direct parent, instead of ${..[MaterialInputs]/name}.
+                        _syntaxErrors.add("Invalid substitution token, parent input must be specified for ancestor lookup: ${" + fkTokDisplay + "}.");
```
labkey-susanh:
```suggestion
                        _syntaxErrors.add("Invalid substitution token. Parent input must be specified for ancestor lookup: ${" + fkTokDisplay + "}.");
```

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1235,56 +1312,104 @@ else if (pt != null)
...
+                        return fieldParts;
+                    }
+
+                    if (ancestorLevel > 9) // 1 generation of direct parent + 9 extra generations of ancestors
+                    {
+                        _syntaxErrors.add("Invalid substitution token, a max of 10 generations of ancestor lookup is supported: ${" + fkTokDisplay + "}.");
```
labkey-susanh:
```suggestion
                        _syntaxErrors.add("Invalid substitution token. A max of 10 generations of ancestor lookup is supported: ${" + fkTokDisplay + "}.");
```

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1804,69 +1931,98 @@ else if (parentObject instanceof ExpData)
...
+                    Identifiable seed = LsidManager.get().getObject(parentLsid);
+                    ExpLineage lineage = ExperimentService.get().getLineage(_container, _user, seed, options);
+
+                    if (ancestorOptions.ancestorSearchType() != null)
+                    {
+                        List<Identifiable> candidateAncestors = lineage.findAncestorByType(parentObject, ancestorOptions.ancestorSearchType(), _user);
+                        candidateAncestors.sort(Comparator.comparing(a -> a.getName()));
+                        for (Identifiable candidateAncestor : candidateAncestors)
+                        {
+                            if (candidateAncestor instanceof ExpRunItem candidate)
```
labkey-susanh: Doesn't `findAncestorsByType` always return `ExpRunItem`s?

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1804,69 +1931,98 @@ else if (parentObject instanceof ExpData)
...
+                            {
+                                if (candidate.getCpasType().equals(ancestorOptions.ancestorSearchType().second))
```
labkey-susanh: Is this check necessary? Seems like that's also taken care of by `findAncestorByType`.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1804,69 +1931,98 @@ else if (parentObject instanceof ExpData)
...
 
-                        for (Identifiable ancestorObject : ancestorObjects)
-                        {
-                            if (ancestorObject instanceof ExpMaterial || ancestorObject instanceof ExpData)
+                            for (Identifiable ancestorObject : ancestorObjects)
                             {
-                                Object lookupValue = getParentFieldValue((ExpObject) ancestorObject, fieldName);
-                                if (lookupValue != null)
-                                    ancestorLookupValues.add(lookupValue);
+                                if (ancestorObject instanceof ExpMaterial || ancestorObject instanceof ExpData)
```
labkey-susanh, line 1980: This check also seem like it could be taken care of by `findAncestorObjects`

api/src/org/labkey/api/exp/api/ExpLineage.java
XingY: done

api/src/org/labkey/api/data/NameGenerator.java
XingY: I reverted the message to show Lineage instead. I meant to update the variables in the code to distinguish between parents and ancestor, but agree lineage is generic enough in the msg.

api/src/org/labkey/api/data/NameGenerator.java
XingY: No it returns Identifiable

api/src/org/labkey/api/data/NameGenerator.java
XingY: `findAncestorObjects` returns identifiable, which could be other types, such as ExpRun or ExpObject

labkey-susanh: Not a big deal, but it seems like we're only ever using `findAncestorObjects` here where we don't care about these other types so we could specialize that method and reduce memory usage.

api/src/org/labkey/api/data/NameGenerator.java
labkey-susanh: Does anyone ever want the other types? Seems like we shouldn't return them if we'll immediately discard them.

api/src/org/labkey/api/data/NameGenerator.java
XingY: `findAncestorObjects` returning other types is only theoretical. If the specified path indicates material or data parents, they should only return those data types.

api/src/org/labkey/api/data/NameGenerator.java
XingY: good point. Updated

api/src/org/labkey/api/data/NameGenerator.java
labkey-susanh: Still curious about this...

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1036,45 +1107,52 @@ else if (pt != null)
...
+                                if (!isAncestorSearch)
+                                    parentLookupFields.computeIfAbsent(fieldParts.get(0) + "/" + fieldParts.get(1), (s) -> new ArrayList<>()).add(fieldParts.get(2));
+
+                                lookupValuePreview = getParentLookupTokenPreview(_currentDataTypeName, fkTok, fieldParts.get(0), fieldParts.get(1), ancestorPartOption, fieldParts.get(2), user, dataClassNames, sampleTypeNames);
                             }
                         }
                         else
                         {
-                            String errorMsg = "Only one level of lookup is supported for lineage input: " + fkTok + ".";
+                            String errorMsg = "Only one level of lookup is supported for parent input: " + fieldKeyDisplay + ".";
```
labkey-susanh:
```suggestion
                            String errorMsg = "Only one level of lookup is supported for lineage input: " + fieldKeyDisplay + ".";
```

api/src/org/labkey/api/data/NameGenerator.java
labkey-susanh: OK, probably still more general than we need, and we should probably consider at some point to have these methods return streams, but at least it's not being too greedy.

labkey-susanh (approved):

PR #5756, opened 2024-08-12
51040: AssayModule does not claim the "assaywell" schema

labkey-klum:
#### Rationale
[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51040)

I did some research and we still do need this provisioned schema (for assay well metadata). I ended up adding it to the set of  provisioned schema names.

### DIFF
--- a/assay/src/org/labkey/assay/AssayModule.java
+++ b/assay/src/org/labkey/assay/AssayModule.java
@@ -20,6 +20,7 @@ import jakarta.servlet.ServletContext;
 import org.apache.commons.collections4.Factory;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


assay/src/org/labkey/assay/AssayModule.java
```diff
@@ -306,7 +307,10 @@ public Set<String> getSchemaNames()
     @NotNull
     public Set<String> getProvisionedSchemaNames()
     {
-        return PageFlowUtil.set("assayresult");
+        return Set.of(
+            AbstractTsvAssayProvider.ASSAY_SCHEMA_NAME,
```
labkey-adam, line 311: 👍

labkey-adam (approved):

PR #5758, opened 2024-08-12
Issue 50964: Save Incomplete Disabled Authentication Configurations

RosalineP:
#### Rationale
See issue [here](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=50964).

#### Related Pull Requests
- https://github.com/LabKey/premiumModules/pull/32

#### Changes
- If auth config is disabled, bypass warning of empty required fields upon save attempt

### DIFF
--- a/api/src/org/labkey/api/security/SaveConfigurationForm.java
+++ b/api/src/org/labkey/api/security/SaveConfigurationForm.java
@@ -8,6 +8,7 @@ import org.labkey.api.settings.AppProps;
 import org.labkey.api.util.ConfigurationException;
 
 import java.util.Map;
+import java.util.Objects;
 
 public abstract class SaveConfigurationForm
 {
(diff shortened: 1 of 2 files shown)


core/src/client/components/DynamicConfigurationModal.tsx
```diff
@@ -91,6 +91,10 @@ export default class DynamicConfigurationModal extends PureComponent<Props, Part
     };
 
     areRequiredFieldsEmpty = (): boolean => {
+        if (!this.state.fieldValues.enabled) {
```
RosalineP, line 95: ToDo: Call out issue number and title

api/src/org/labkey/api/security/SaveConfigurationForm.java
```diff
@@ -78,7 +79,7 @@ public void setDomain(String domain)
     public final @Nullable String getProperties()
     {
         Map<String, Object> map = getPropertyMap();
-        assert map.containsKey("domain") ? map.get("domain").equals(getDomain()) : (map.containsKey("Domain") ? map.get("Domain").equals(getDomain()) : null == getDomain()) : "Inconsistency between getDomain() and map.get(\"domain\")";
+        assert Objects.equals(map.get("domain"), getDomain());
```
labkey-alan, line 82: Is it appropriate to use assert here? Or should this throw an exception? Asserts aren't run on production servers, so this would only run on dev servers (and even then, some dev servers have asserts disabled).

I'm not going to hold up the merge on this, because we're already using assert, but we should probably consider an alternative.

labkey-alan (approved):
Looks good, but

api/src/org/labkey/api/security/SaveConfigurationForm.java
labkey-adam: This is intended to be a development-only check, ensuring that any subclass that wants to use getDomain() also populates the property into the map. If the invariant check fails then we have a coding error, not a user mistake or data issue, hence the use of an assert.

PR #5763, opened 2024-08-15
Adjust permissions for Troubleshooters

labkey-adam:
#### Rationale
Troubleshooters should be able to view metrics information and shouldn't be able to invoke expensive tests on data sources. We also should check for read permissions in folders with external schemas before linking to them.

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -11477,7 +11477,7 @@ public class AdminController extends SpringActionController
         }
     }
 
-    @RequiresPermission(AdminPermission.class)
+    @RequiresPermission(TroubleshooterPermission.class)
     public class ViewUsageStatistics extends SimpleViewAction<Object>
     {
         @Override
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
This seems OK but it would be good to capture this in our definition of the Troubleshooter role so we can aim to be consistent going forward. Until now, I think troubleshooters could effectively do all read-only actions. Now they can do almost all read-only actions.

https://www.labkey.org/Documentation/wiki-page.view?name=permissionLevels

labkey-adam:
> This seems OK but it would be good to capture this in our definition of the Troubleshooter role so we can aim to be consistent going forward. Until now, I think troubleshooters could effectively do all read-only actions. Now they can do almost all read-only actions.
> 
> https://www.labkey.org/Documentation/wiki-page.view?name=permissionLevels

Yeah, bit of a judgement call here. Troubleshooters can still read the details of all data sources, they just can't invoke the potentially time-consuming test that's really intended for development & testing of dialects. (I debated making the test button devMode-only.) I added "(not) invoke resource-intensive processes" to the doc, which is consistent with how I've been tagging actions. We don't let Troubleshooters invoke system maintenance tasks or clear the query profiler either, and these aren't technically "changing settings."

PR #5767, opened 2024-08-15
Issue 51021: WebSocket to add window focus listener 

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51021

We previously added code in the apps so that on browser window focus we do a whoAmI check via WebSocket. checkForExpiredSession() so that we can reliably show the session timeout / logout modal to the user. This PR moves that approach to the WebSocket.js file directly (note that I will remove the app specific code in develop).

#### Changes
- add window focus listener to help identify session timeout and user logout more consistently session timeout and user logout more consistently

### DIFF
--- a/api/webapp/clientapi/dom/WebSocket.js
+++ b/api/webapp/clientapi/dom/WebSocket.js
@@ -35,6 +35,7 @@ LABKEY.WebSocket = new function ()
     var _websocket = null;
     var _callbacks = {};
     var _modalShowing = false;
+    var _shouldAddFocusListener = true;
 
     function openWebsocket() {
         if (_websocket !== null || !('WebSocket' in window)) return;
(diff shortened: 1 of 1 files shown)


labkey-nicka (approved):

labkey-jeckels:
Is the problem that the web socket connection gets cut off when the tab/window doesn't have focus, so we don't get the notification that the session is done?

cnathe:
> Is the problem that the web socket connection gets cut off when the tab/window doesn't have focus, so we don't get the notification that the session is done?

I'm not really sure. The previous iteration of this WebSocket.js code didn't seem to have a problem with getting notified of the session timeout event when the tab/browser was not in focus. Not sure if the change was related to embedded tomcat, browser versions, etc. but this approach of checking for whoAmI on tab focus is something we've been doing on the app side React code for awhile now.

PR #5768, opened 2024-08-15
Issue 51067: Add "StorageRowSort" and "StorageColSort" as reserved inventory columns

labkey-nicka:
#### Rationale
This addresses [Issue 51067](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51067) by renaming "Row" and "Col" to "StorageRowSort" and "StorageColSort" and adding them as reserved column names on sample type domains.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/573

#### Changes
- Introduce `enum InventoryStatusColumn` to support constant references to columns that are built-in to inventory status.


api/src/org/labkey/api/inventory/InventoryService.java
```diff
@@ -31,40 +31,63 @@
...
+    enum InventoryStatusColumn
+    {
+        CheckedOut,
+        CheckedOutBy,
+        Col,
+        FreezeThawCount("Freeze/Thaw Count"),
+        Row,
+        SampleTypeUnits,
+        StorageCol,
+        StorageComment, // TODO: For reviewer. Seems like this column may have been removed/renamed? It was only listed in the labels and not the names.
```
labkey-susanh: This field is reserved for adding an audit log comment related to storage actions during file import. We could reserve the label version (with the space) as well, but if we do that we should probably update our logic for file import to look for this column label as well. (See InventoryItemTable.PersistStorageItemDataIterator.)

api/src/org/labkey/api/inventory/InventoryService.java
```diff
@@ -31,40 +31,63 @@
...
+        {
+            this.label = label;
+        }
+
+        public String label()
+        {
+            return this.label;
+        }
+
+        public FieldKey fieldKey()
```
labkey-susanh, line 78: This does no harm, but doesn't seem to be used.

labkey-susanh (approved):

api/src/org/labkey/api/inventory/InventoryService.java
labkey-susanh: Ah, never mind. I see it now.

api/src/org/labkey/api/inventory/InventoryService.java
labkey-nicka: We were already reserving the label (with the space). What this does is makes it so we are reserving both "StorageComment" and "Storage Comment".

Thanks for the pointer. Updated `InventoryItemTable.PersistStorageItemDataIterator` to attempt to resolve this column by label as well.

PR #5569, opened 2024-06-07
Revert DROP * IF EXISTS on SQL Server change

labkey-adam:
#### Rationale
Turns out support for `DROP * IF EXISTS` was added for SQL Server 2016... and we still support 2014. :(

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5567

PR #5772, opened 2024-08-15
Reinstate support for DROP * IF EXISTS on MS SQL Server

labkey-adam:
#### Rationale
Now that we've removed support for SQL Server 2014, we can start using `DROP TABLE IF EXISTS ...`, `DROP VIEW IF EXISTS ...`, etc.

Allow the `DROP * IF EXISTS ...` syntax and use it in a few scripts.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5569

### DIFF
--- a/core/resources/schemas/dbscripts/sqlserver/core-24.003-24.004.sql
+++ b/core/resources/schemas/dbscripts/sqlserver/core-24.003-24.004.sql
@@ -1,1 +1,1 @@
-EXEC core.fn_dropifexists 'PrincipalRelations', 'core', 'TABLE', NULL
+DROP TABLE IF EXISTS core.PrincipalRelations;
--- a/bigiron/src/org/labkey/bigiron/mssql/BaseMicrosoftSqlServerDialect.java
+++ b/bigiron/src/org/labkey/bigiron/mssql/BaseMicrosoftSqlServerDialect.java
@@ -332,17 +332,11 @@ abstract class BaseMicrosoftSqlServerDialect extends SqlDialect
         }
     }
 
-    private static final Pattern DROP_IF_EXISTS = Pattern.compile(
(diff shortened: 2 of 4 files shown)


core/resources/schemas/dbscripts/sqlserver/core-24.003-24.004.sql
```diff
@@ -1 +1 @@
-EXEC core.fn_dropifexists 'PrincipalRelations', 'core', 'TABLE', NULL
+DROP TABLE IF EXISTS core.PrincipalRelations;
```
labkey-nicka, line 1: I see a lot more usages (in platform alone) that make use of `EXEC core.fn_dropifexists`. Is the plan to migrate those over and remove the `core.fn_dropifexists` or does this not completely logically replace all of those usages?

Is there support now for `DROP CONSTRAINT IF EXISTS`?

labkey-adam: There's fairly broad support for `IF EXISTS`; `ALTER TABLE <table> DROP CONSTRAINT ...` is an option. My primary goal was to enable `IF EXISTS` going forward, for future scripts. I don't plan to migrate all uses of `core.fn_dropifexists` to `DROP * IF EXISTS`, but you're welcome to give it a go. 😄 I haven't looked to see if the "new" syntax covers everything our function does... but, for example, I don't think it helps much with dropping a default value since it still requires the default constraint's name (which is a pain to get). So, obviously, no plan to drop the old function.

labkey-nicka: Thanks for the details.

labkey-nicka (approved):

labkey-adam:
Old Dependencies suite shows no complaints about `DROP * IF EXISTS`: https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_InternalSuites_OldDependenciesSqlserver/3150751?buildTab=tests&status=failed

PR #5773, opened 2024-08-16
getRemoteAddr() helper that uses X-Forwarded-For header

labkey-matthewb:
#### Rationale
Load balancers

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/util/HttpUtil.java
+++ b/api/src/org/labkey/api/util/HttpUtil.java
@@ -282,4 +282,20 @@ public class HttpUtil
     {
         return request.getHeader("User-Agent");
     }
+
+    public static @Nullable String getRemoteAddr(HttpServletRequest request)
+    {
+        String xff = request.getHeader("X-Forwarded-For");
+        if (!StringUtils.isBlank(xff))
+        {
+            var arr = xff.split(",");
+            if (arr.length > 0)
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
Looks good. Note that it swaps `getRemoteAddr()` where it previously used `getRemoteHost()`. Seems fine.

There are a handful of other places to call either method. Most also check for the header. Probably we should consolidate and even add an IntelliJ warning to steer people to the utility code.

labkey-matthewb:
> Looks good. Note that it swaps `getRemoteAddr()` where it previously used `getRemoteHost()`. Seems fine.
> 
> There are a handful of other places to call either method. Most also check for the header. Probably we should consolidate and even add an IntelliJ warning to steer people to the utility code.

I figured migrating other usages could  be done on develop.

labkey-matthewb:
PopZ

labkey-jeckels:
> > Looks good. Note that it swaps `getRemoteAddr()` where it previously used `getRemoteHost()`. Seems fine.
> > There are a handful of other places to call either method. Most also check for the header. Probably we should consolidate and even add an IntelliJ warning to steer people to the utility code.
> 
> I figured migrating other usages could be done on develop.

Yup. Sounds good!

PR #5779, opened 2024-08-16
Allow SQL dialects to enforce identifier name limitations

labkey-tchad:
#### Rationale
Snowflake only allows all-caps field and table identifiers. Need to add a way for SQL dialects to enforce identifier name limitations. This change fixed the errors I was having with PIVOT queries and expression columns in LabKey SQL.
It also makes LabKey SQL not be as strict about column alias casing when querying Snowflake.

#### Related Pull Requests
- https://github.com/LabKey/premiumModules/pull/30

#### Changes
- Allow SQL dialects to enforce identifier name limitations

### DIFF
--- a/api/src/org/labkey/api/query/AliasManager.java
+++ b/api/src/org/labkey/api/query/AliasManager.java
@@ -175,7 +175,9 @@ public class AliasManager
             ret = ret + "_";
         int length = ret.length();
         if (0 == length)
-            return "X_";
+            ret = "X_";
+        if (dialect != null)
+            ret = dialect.makeLegalIdentifierName(ret);
         // we use 28 here because Oracle has a limit or 30 characters, and that is likely the shortest restriction
         int maxLength = useLegacyMaxLength ? 40 : (dialect == null ? 28 : dialect.getIdentifierMaxLength());
         if (reserveCount > 0)
--- a/api/src/org/labkey/api/data/dialect/SimpleSqlDialect.java
+++ b/api/src/org/labkey/api/data/dialect/SimpleSqlDialect.java
@@ -441,6 +441,13 @@ public abstract class SimpleSqlDialect extends SqlDialect
     {
         return false;
     }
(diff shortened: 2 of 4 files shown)


api/src/org/labkey/api/query/AliasManager.java
```diff
@@ -175,7 +175,9 @@ private static String makeLegalName(String str, @Nullable SqlDialect dialect, bo
             ret = ret + "_";
         int length = ret.length();
         if (0 == length)
-            return "X_";
+            ret = "X_";
+        if (dialect != null)
+            ret = dialect.makeLegalIdentifierName(ret);
```
labkey-adam, line 180: Shouldn't we continue to short-circuit (skip the inapplicable `maxLength` stuff) in the `0 == length` case? Could do something like:

```
if (0 == length)
{
    ret = "X_";
}
else
{
   ... maxLength stuff
}
return dialect != null ? dialect.makeLegalIdentifierName(ret) : ret;
```

labkey-tchad: Out of preponderance of caution, I was guarding against the remote possibility that `makeLegalIdentifierName` might make an identifier that is too long or that "X_" isn't legal.
Looking at this, is there a reason to keep the `useLegacyMaxLength` flag around? It looks like its purpose for existence is pretty ancient:  https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=26311

labkey-adam: I think you'll have to check with @labkey-susanh about `useLegacyMaxLength`.

I don't think we need to worry about `makeLegalIdentifier("X_")` exceeding any length restrictions, but you're right that `makeLegal()` of other values might push them past the limit. Feel free to adjust or leave this as you had it... it's not a big deal.

labkey-adam (approved):

PR #5780, opened 2024-08-16
Return status code 200 when validation finds a problem with a query

labkey-jeckels:
#### Rationale
Finding a problem when validating a query isn't an error condition with the HTTP request/response. The schema browser client code that calls this API is equally happy with a 200 or 400 response that includes the validation details.

#### Changes
- Return a 200 response code where the JSON payload indicates the failure condition


labkey-jeckels:
@labkey-mohara this change will let you remove the caveat in the note on this page: https://www.labkey.org/Documentation/wiki-page.view?name=querySchemaBrowser#val

labkey-matthewb (approved):

PR #5781, opened 2024-08-17
Issue 50917: imported folder archives do not correct file paths for file fields

XingY:
#### Rationale
FileLink field type is supported for study dataset, assay batch/run/result fields and for sample types. Some of those datatypes support relative file path as input, some support alternative root (`assayfiles`), and all of them support absolute file path. This PR address 3 main issues around filelink fields folder export and import:
- for absolute file path values, substitute the source file root with token during export and replace with target file root during import. Some of the datatype exports using tsv writer, some as xar. For import, some uses xar reader, some tsv loader then QUS/FileColumn/FileDataIterator. 
-  for `@assayfiles` assay result files, the files were not previously exported. This PR started to export them as part of File export.
- We don't handle values with `\` (backslash) well. WIndows filepath are now exported with quotes and then stripped of the quotes during import so our TSVloader can parse them correctly.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5781/files
- https://github.com/LabKey/testAutomation/pull/2026

#### Changes
- Substitute file root with new ${FileRoot} substitution param for FileLink fields during folder export
- Substitute ${FileRoot} with current fileroot during folder import
- Include `@assayfiles` root as part of Files folder export
- Quote file path during export and unquote during import for Windows file paths with backslashes


api/src/org/labkey/api/assay/TsvDataHandler.java
```diff
@@ -184,4 +194,50 @@ public void exportFile(ExpData data, Path dataFile, User user, OutputStream out)
...
+            if (StringUtils.isEmpty(_fileRoot))
+                return o;
+            if (o instanceof String filePath)
+            {
+                if (!filePath.startsWith(_fileRoot))
+                    return filePath;
+
+                String formatted =  filePath.replace(_fileRoot, FILE_ROOT_SUBSTITUTION);
+                if (formatted.contains("\\"))
+                    return "\"" + formatted + "\"";
```
cnathe, line 235: this seems like something that could use a comment as to why the quotes are needed here

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -1739,22 +1742,60 @@ public Object get()
         }
     }
 
+    public static String getFileRootSubstitutedFilePath(@Nullable String filePath, @Nullable String fileRootPath)
+    {
+        if (!StringUtils.isEmpty(fileRootPath) && filePath != null)
+        {
+            if (filePath.startsWith("\"") && filePath.endsWith("\"") && filePath.contains("\\")) // fix windows exported path
+                filePath = filePath.substring(1, filePath.length() - 1);
```
cnathe, line 1750: would it be the case that we'd want to strip off the leading and trailing quote regardless of `filePath.contains("\\")`? Seems like if it starts with a quote and ends with one then we don't need the extra check for the `\\` but I could be wrong

cnathe (approved):

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
XingY: I believe it's legitimate for filename to be "inquote", same with "in\quotewithslash", though both scenarios are extremely unlikely. This change is not ideal and was aimed to handle those produced by the TsvDataHandler.ExportFileLinkColumn so I was thinking limiting scope is probably safer.

PR #5783, opened 2024-08-19
48429 : support shift click to select/deselect child nodes

labkey-klum:
#### Rationale
[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48429)

This change adds the ability to select/deselect all child nodes by shift + clicking on the parent node in the column list of the view customizer. Specifically, the new behavior is:

On shift click on the parent node checkbox:
- Will check or uncheck all child nodes (single level) of the parent node. If the parent node is expanded.
- If the parent node is initially collapsed, the node will be expanded and all children selected. The converse is not true (no collapsing of the parent node on shift+click)


core/webapp/internal/ViewDesigner/tab/BaseTab.js
```diff
@@ -291,7 +291,7 @@ Ext4.define('LABKEY.internal.ViewDesigner.tab.BaseTab', {
             index = fieldKeyOrIndex;
         }
         else {
-            index = list.store.find('fieldKey', fieldKeyOrIndex, 0, false, false);
+            index = list.store.find('fieldKey', fieldKeyOrIndex, 0, false, false, true);
```
labkey-klum, line 294: Take a close look at this, there was a subtle bug where if you expanded a FK (selected some child values) and positioned the parent node in the column list below any of the children. Selecting/deselecting the FK node would remove the wrong item (one of the children) on the right and get you into a state where you would have multiple parent nodes.

The sixth parameter in the `find` function specifies an exact or partial match on the search field value. I think we want the exact match otherwise `Parent` and `Parent/Child` would both match depending on what order they appeared in the store.

core/webapp/internal/ViewDesigner/Designer.js
```diff
@@ -1062,9 +1068,53 @@ Ext4.define('LABKEY.internal.ViewDesigner.Designer', {
...
+
+        var event = this.eventMap[node.internalId] || {};
+        if (event)
+            delete this.eventMap[node.internalId];
+        // issue 48429 : select all child nodes (and expand if necessary)
+        if (event.shiftKey)
+            this.toggleChildNodeSelection(node, checked);
+    },
+
+    toggleChildNodeSelection(node, checked) {
```
labkey-klum, line 1082: Obviously there is no UI hint for this behavior, I'm curious whether there are any patterns we have used in our apps to reveal functionality (maybe in the editable grid)?

cnathe: We do have shift click type events to allow quick selection in the app, but we don't have any visual hints that I know about in the UI to make users aware of this behavior. One thought would be to add an info tooltip icon next to the "Available Fields" header in the LKS customize view with some help text like "hold this SHIFT key when clicking to select all child fields for a lookup"

core/webapp/internal/ViewDesigner/Designer.js
```diff
@@ -1062,9 +1068,53 @@ Ext4.define('LABKEY.internal.ViewDesigner.Designer', {
...
+                    node.expand(false, function (children) {
+                        Ext4.each(children, function (child) {updateNode(child, checked, this);}, this);
+                    }, this);
+                }
+                else {
+                    Ext4.each(node.childNodes, function (child) {updateNode(child, checked, this);}, this);
+                }
+            }
+            else {
+                // unselect all child nodes if the parent node is expanded
```
cnathe: this code path only gets executed if the SHIFT key is used, so it would seem to me like we'd want to unselect all child nodes in this case regardless of if the node was expanded (i.e. similar to the check behavior above for lines 1102-1109.

core/webapp/internal/ViewDesigner/tab/BaseTab.js
cnathe: yes that sounds correct, i.e. we'd want an exact match. I wish true would be the Ext default here to default to exact match.

cnathe (approved):

core/webapp/internal/ViewDesigner/Designer.js
labkey-klum: Hmm, did you pull this change down to try it out? The reason I didn't unselect if the node wasn't expanded is because the user wouldn't necessarily see those nodes get cleared. I guess the right hand side would get cleared but those columns might be scrolled off so I was worried about doing something that wasn't apparent to the user.

It's a simple change to remove that condition if you think it's better to be consistent on the behavior regardless of expanded state.

cnathe: I did pull down the change and try it out. I agree it might not be apparent to the user how many items are being removed in the collapsed case, but I think I prefer the consistency. If the user knows about the "hold SHIFT and check/uncheck" trick, then I think they'd expect all to be remove/added regardless of expand/collapse state.

PR #5785, opened 2024-08-20
Plates: support replicate well groups

labkey-nicka:
#### Rationale
This enables replicate well groups for standard assay plates and introduces validation of replicate wells.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/495
- https://github.com/LabKey/platform/pull/5785
- https://github.com/LabKey/limsModules/pull/588

#### Changes
- Query for well types during validation to determine if a well change requires further validation.
- Validate replicate well groups across plate set ensuring that all the data is consistent between rows.

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateManager.java
+++ b/assay/src/org/labkey/assay/plate/PlateManager.java
@@ -3117,7 +3117,7 @@ public class PlateManager implements PlateService, AssayListener, ExperimentList
 
         for (WellData wellData : wellDataList)
         {
-            boolean isSampleWell = WellGroup.Type.SAMPLE.equals(wellData.getType());
+            boolean isSampleWell = WellGroup.Type.SAMPLE.equals(wellData.getType()) || WellGroup.Type.REPLICATE.equals(wellData.getType());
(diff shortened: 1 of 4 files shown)


assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3511,6 +3524,92 @@ private void validateWells(Plate plate) throws ValidationException
...
+
+        columnNames.removeAll(excludedColumns);
+
+        var sql = new SQLFragment("SELECT\n");
+        var separator = "";
+        for (String columnName : columnNames)
+        {
+            sql.append(separator).append(columnName).append("\n");
+            separator = ", ";
+        }
```
labkey-klum: you could probably save and reuse the column list in both places.

labkey-klum (approved):

assay/src/org/labkey/assay/plate/PlateManager.java
labkey-nicka: Done.

PR #5786, opened 2024-08-20
Ensure module properties are current on every startup

labkey-adam:
#### Rationale
A handful of unrelated improvements... see Changes section for details of each

#### Changes
- Previously, we updated the `core.Modules` table only when upgrades occurred, but these properties could change outside of an upgrade scenario (e.g., the schemas a module claims, https://github.com/LabKey/platform/pull/5760) . With this PR, we update module properties on every server startup, but only if they've changed vs. what's stored in the database. For consistency, we also sort the schema name list before saving and avoid `Set.of()` with multiple schema names since this method is aggressively non-deterministic.
- When encountering a display column formatting problem, provide more detailed logging so there's a chance we can determine which column is at fault
- Update MySQL reserved word list based on candidates added due to Snowflake
- Log the details when encountering an unsupported JDBC driver class names and product versions; logging "JDBC database driver is not supported" with no other details is unhelpful.
- Fix comments and warnings

### DIFF
--- a/bigiron/src/org/labkey/bigiron/mysql/MySql90Dialect.java
+++ b/bigiron/src/org/labkey/bigiron/mysql/MySql90Dialect.java
@@ -15,7 +15,7 @@ public class MySql90Dialect extends MySql80Dialect
         Set<String> words = super.getReservedWords();
 
         words.removeAll(new CsvSet("master_bind, master_ssl_verify_server_cert"));
-        words.addAll(new CsvSet("intersect, parallel, tablesample"));
+        words.addAll(new CsvSet("parallel, qualify, tablesample"));
 
         return words;
     }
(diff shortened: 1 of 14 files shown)


bigiron/src/org/labkey/bigiron/mysql/MySql90Dialect.java
```diff
@@ -15,7 +15,7 @@ protected Set<String> getReservedWords()
         Set<String> words = super.getReservedWords();
 
         words.removeAll(new CsvSet("master_bind, master_ssl_verify_server_cert"));
-        words.addAll(new CsvSet("intersect, parallel, tablesample"));
+        words.addAll(new CsvSet("parallel, qualify, tablesample"));
```
labkey-jeckels, line 18: FWIW (not much in this context), I've never really like CsvSet with a hardcoded string. Seems cleaner to use Set.of() or similar.

labkey-jeckels (approved):

bigiron/src/org/labkey/bigiron/mysql/MySql90Dialect.java
labkey-adam: So many double quotes 🙂

PR #5790, opened 2024-08-21
47542: Defer and consolidate audit domain changes.

labkey-klum:
#### Rationale
[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47542)

Perhaps this is just moving the problem around but there were potentially a bunch of audit log changes that happened very early (in the constructor) that were probably causing some of these test deadlocks. This change just limits the initialization in the constructor to domain creation but then defers (and batches) any potential domain updates until after server startup time.

This also addresses the fix for 50059 in a more graceful way, instead of having to invalidate the provisioned domain after querying.

### DIFF
--- a/api/schemas/view.xsd
+++ b/api/schemas/view.xsd
@@ -12,7 +12,11 @@
 
     <xsd:complexType name="viewType">
         <xsd:sequence>
-            <xsd:element name="permissions" type="vw:permissionsListType" minOccurs="0" maxOccurs="1"/>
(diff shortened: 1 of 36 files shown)


labkey-klum:
Targeting 24.7 mainly because this is where we are seeing the persistent [failure](https://teamcity.labkey.org/buildConfiguration/LabKey_247Release_Community_DailySuites_CustomModulesSqlserver?mode=branches).

labkey-adam (approved):

PR #5799, opened 2024-08-24
Plates: move custom fields from plate to plate set

labkey-nicka:
#### Rationale
This moves server-side implementation of custom fields to be declared on a plate set instead of on each plate individually. I have not changed the API so it still accepts a `plateId` from which is resolves the plate set. This leaves the current UX intact.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/605

#### Changes
- Create and hydrate a new `assay.PlateSetProperty` table from the pre-existing `assay.PlateProperty` table.
- Update queries mutating custom fields to resolve to/from the plate set.
- Drop the `assay.PlateProperty` table.
- Fix [Issue 50161](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50161) by deleting `assay.PlateSetProperty` rows via `ON DELETE CASCADE`.

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateCache.java
+++ b/assay/src/org/labkey/assay/plate/PlateCache.java
@@ -204,6 +204,7 @@ public class PlateCache
     public static void uncache(Container c, PlateSet plateSet)
     {
         // TODO: Feels like this could end up caching a bunch of a plates that were not cached just to uncache them.
+        // Consider introducing plate set caching and moving custom field modeling to plate sets
         getPlatesForPlateSet(c, plateSet.getRowId()).forEach(plate -> uncache(c, plate));
     }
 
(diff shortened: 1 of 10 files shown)


assay/src/org/labkey/assay/plate/PlateCache.java
```diff
@@ -200,6 +201,12 @@ public static void uncache(Container c, Plate plate)
             _loader._containerPlateMap.get(c).remove(plate);
     }
 
+    public static void uncache(Container c, PlateSet plateSet)
+    {
+        // TODO: Feels like this could end up caching a bunch of a plates that were not cached just to uncache them.
```
labkey-nicka, line 206: @labkey-klum I'd like to discuss this with you. Moving custom fields to the plate set could result in a lot of cache churning as we now need to uncache all plates in a plate set.

labkey-klum (approved):

PR #5801, opened 2024-08-26
AbDisc: Plate Controls

RosalineP:
#### Rationale
See spec [here](https://docs.google.com/document/d/1jXnyEJcPekCcMOi5yeQLL-4f6XsXtUzZjPC5O3rWHL4/edit#heading=h.4rg4h5vuek0).

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1560
- https://github.com/LabKey/labkey-ui-premium/pull/501
- https://github.com/LabKey/platform/pull/5801
- https://github.com/LabKey/limsModules/pull/606

#### Changes
- Add additional validation
- Bump version


assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3452,9 +3452,25 @@ public void computeWellGroups(
...
+    private Set<Integer> getValidControlSampleIds(Container container, User user, Integer plateRowId) throws ValidationException
+    {
+        UserSchema userSchema = QueryService.get().getUserSchema(user, container, PlateSchema.SCHEMA_NAME);
+        TableInfo controlSamplesTInfo = userSchema.getTable("ControlSamples");
+        TableSelector selector = new TableSelector(controlSamplesTInfo, Collections.singleton("RowId"));
+
+        // Feed plate set id into ControlSamples' query parameter
+        Plate p = getPlate(container, plateRowId);
+        if (p == null || p.getPlateSet() == null)
+            throw new ValidationException("Unable to resolve Plate or Plate Set.");
```
labkey-nicka: This should specify the `plateRowId` that was not able to be resolved.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3452,9 +3452,25 @@ public void computeWellGroups(
...
+
+        selector.setNamedParameters(Collections.singletonMap("PlateSetRowId", p.getPlateSet().getRowId()));
+        return new HashSet<>(selector.getCollection(Integer.class));
+    }
+
+    public void validateWellGroups(Container container, User user, Collection<Integer> plateRowIds) throws ValidationException
     {
         clearCache(plateRowIds);
+        Set<Integer> validControls = null;
```
labkey-nicka: Hmm, this set could be really really large as there is no cap on how many samples could fit our definition of a sample. Seems like we should consider coalescing the set of samples that are being asked for as samples and then verify them all at once with a query instead of the other way around.

labkey-nicka (changes requested):

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3461,11 +3461,33 @@ public void computeWellGroups(
         }
     }
 
+    private Set<Integer> getValidControlSampleIds(Container container, User user, Integer plateRowId) throws ValidationException // flag for deletion
```
labkey-nicka: This function is not used and can be removed.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3461,11 +3461,33 @@ public void computeWellGroups(
...
+        return new HashSet<>(selector.getCollection(Integer.class));
+    }
+
     public void validateWellGroups(Container container, User user, Collection<Integer> plateRowIds) throws ValidationException
     {
         clearCache(plateRowIds);
         Set<Integer> plateSetsWithReplicates = new HashSet<>();
 
+        Integer plateId = plateRowIds.iterator().next();
+        PlateSet ps = requirePlate(container, plateId, null).getPlateSet();
```
labkey-nicka: There is no stipulation that these results will be within a single plate set. Someone could make an update rows call across two plates in two different plate sets. I think what we may want to do here is to track which plate sets have controls, then outside of the `plateRowIds` loop validate the controls across the plate sets. See `plateSetsWithReplicates` and how it is used as example.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3613,6 +3642,55 @@ private void validatePlateSetReplicates(Container container, User user, @NotNull
...
+        }
+        catch (SQLException e)
+        {
+            throw UnexpectedException.wrap(e);
+        }
+
+        return invalidSampleName;
+    }
+
+    private void validateControlWellGroup(UserSchema schema, Plate plate, WellGroup wellGroup, Integer plateSetRowId) throws ValidationException
```
labkey-nicka: Let's add an integration test case to `PlateManagerTest` that runs through this validation through the front door (i.e. update rows).

labkey-nicka (approved):
Thanks for adding the test.

PR #5805, opened 2024-08-26
Fix SQL Server timestamp handling with non-US date formats

labkey-adam:
#### Rationale
On SQL Server with a non-US language set (which sets the date format to "dmy"), all Timestamp JDBC parameters are sent in a format that SQL Server parses incorrectly. This leads to incorrect data (when the day of the month is <= 12) or exceptions (when day of the month > 12). https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51129

### DIFF
--- a/bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServer2016Dialect.java
+++ b/bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServer2016Dialect.java
@@ -15,21 +15,54 @@
  */
 package org.labkey.bigiron.mssql;
 
+import org.apache.logging.log4j.Logger;
+import org.junit.Assert;
+import org.junit.Test;
 import org.labkey.api.data.ConnectionWrapper;
+import org.labkey.api.data.DbScope;
+import org.labkey.api.data.SqlSelector;
 import org.labkey.api.data.dialect.StatementWrapper;
+import org.labkey.api.module.ModuleLoader;
+import org.labkey.api.util.logging.LogHelper;
 
+import java.sql.CallableStatement;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.sql.Timestamp;
+import java.sql.Types;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 import java.util.Calendar;
+import java.util.Date;
+import java.util.Map;
 
-/**
- * User: adam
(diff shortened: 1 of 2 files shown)


bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServer2016Dialect.java
```diff
@@ -15,21 +15,55 @@
...
+    private volatile String _language = null;
+    private volatile String _dateFormat = null;
+    private volatile DateTimeFormatter _timestampFormatter = null;
+
+    @Override
+    public void prepare(DbScope scope)
+    {
+        super.prepare(scope);
+
+        Map<String, Object> map = new SqlSelector(scope, "SELECT language, date_format FROM sys.dm_exec_sessions WHERE session_id = @@spid").getMap();
```
labkey-jeckels, line 55: Not something for this PR, but it would be cool to support `record` objects for `SqlSelector`, etc. Maybe we do, but I'm guessing we don't.

bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServer2016Dialect.java
```diff
@@ -15,21 +15,55 @@
...
+        _language = (String) map.get("language");
+        _dateFormat = (String) map.get("date_format");
+
+        // This seems to be the only string format acceptable for sending Timestamps, but unfortunately it's ambiguous;
+        // SQL Server interprets the "MM-dd" portion based on the database's regional settings. So we must query the
+        // current date format and switch the formatter pattern based on what we find. See Issue 51129.
+        _timestampFormatter = DateTimeFormatter.ofPattern("yyyy-" + ("mdy".equals(_dateFormat) ? "MM-dd" : "dd-MM") + " HH:mm:ss.SSS");
```
labkey-jeckels: Worth explicitly checking for whether we support the format, instead of assuming it's `mdy` vs something else (presumably `dmy`)?

labkey-jeckels (approved):

bigiron/src/org/labkey/bigiron/mssql/MicrosoftSqlServer2016Dialect.java
labkey-adam: Yes it would...

labkey-adam: Implemented on develop!

PR #5807, opened 2024-08-27
Calculated fields remove experimental feature flag

cnathe:
#### Rationale
Recent PRs have added the last bit of changes for the calculated fields v1 feature set. This PR removes the experimental feature flag and makes a couple of final changes: catch an exception related to date value expressions, add a max length check, and updates the related package versions.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1561

#### Changes
- Issue 51135: Calc fields fix to catch exception for invalid parsed date interval value
- Add calculated field value expression max length check on domain save
- Remove experimental feature flag


api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -1343,6 +1343,11 @@ public static ValidationException validateProperties(@Nullable Domain domain, @N
             {
                 namePropertyIdMap.put(name, field.getPropertyId());
             }
+
+            if (field.getValueExpression() != null && field.getValueExpression().trim().length() > 4000)
+            {
+                exception.addFieldError(name, getDomainErrorMessage(updates,"The value expression for '" + name + "' is too long. Please limit to 4000 characters."));
```
labkey-susanh: Is it worth also echoing the length so folks know how much over the limit the string is? Something like:

"... is too long (4010 characters). Please ..."

labkey-susanh (approved):

api/src/org/labkey/api/exp/property/DomainUtil.java
cnathe: Yes good idea. Done

PR #5808, opened 2024-08-27
Exclude special "serverGuid" document from validation checks

labkey-adam:
#### Rationale
Related PR added a "special" document to the Lucene index that doesn't conform to our usual rules. `IndexInspector` didn't get the memo, resulting in highly truncated index content exports.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5446

### DIFF
--- a/search/src/org/labkey/search/model/IndexInspector.java
+++ b/search/src/org/labkey/search/model/IndexInspector.java
@@ -152,6 +152,10 @@ public class IndexInspector
 
                         if (titles.length != 1 || urls.length != 1 || uniqueIds.length != 1 || containerIds.length != 1)
                         {
+                            // Skip the special "serverGuid" doc
+                            if (doc.get("serverGuid") == null)
+                                continue;
+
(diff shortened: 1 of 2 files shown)


search/src/org/labkey/search/model/IndexInspector.java
```diff
@@ -152,6 +152,10 @@ protected int writeBody()
 
                         if (titles.length != 1 || urls.length != 1 || uniqueIds.length != 1 || containerIds.length != 1)
                         {
+                            // Skip the special "serverGuid" doc
+                            if (doc.get("serverGuid") == null)
```
labkey-jeckels: Seems like an excellent candidate for a constant.

labkey-jeckels (approved):

search/src/org/labkey/search/model/IndexInspector.java
labkey-adam: Yeah yeah

PR #5812, opened 2024-08-28
Misc 24.9 issue fixes

cnathe:
#### Rationale
Issue [51123](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51123): App can't Import Assay Results from Workflow Job
Issue [50935](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50935): Storage duplicate key violation happening when saving hierarchy
Issue [51025](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51025): App shouldn't allow saving of data in editable grid if there are know errors
Issue [50719](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50719): When a Sample field for an assay domain begins with a #, you can't create results in the grid
Issue [50546](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50546): Cross folder support for assay run "Update QC State"

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1563

#### Changes
- ImportRunApiAction fix for if the first column name starts with a # (the data loader will treat the header row as a comment so need to drop the #)


assay/src/org/labkey/assay/actions/ImportRunApiAction.java
```diff
@@ -257,9 +257,22 @@ else if (rawData != null && !rawData.isEmpty())
...
+                    String newFirstColumn = firstColumn.substring(1);
+                    columns.set(0, newFirstColumn);
+                    for (Map<String, Object> row : rawData)
+                    {
+                        if (row.containsKey(firstColumn))
+                            row.put(newFirstColumn, row.remove(firstColumn));
+                    }
+                }
+
+                try (TSVMapWriter tsvWriter = new TSVMapWriter(columns, rawData))
```
labkey-susanh, line 275: We're able to properly match the data if the column has been renamed?

labkey-susanh (approved):

assay/src/org/labkey/assay/actions/ImportRunApiAction.java
cnathe: yes, in the code above we are renaming the first column to drop the leading '#' but then also updating the data map with that same rename.

PR #5813, opened 2024-08-28
Reduce role/permission log warnings

labkey-adam:
#### Rationale
`RoleManager` and `RoleConverter` log warnings every time they encounter an unknown role. This can result in an excessive number of warnings on development machines where roles are often set in test folders, their corresponding modules often come & go, and the indexer & other processes reference security policies that hold these missing roles.

RoleConverter logging is completely redundant since it calls a RoleManager method that logs, so remove that. RoleManager logging is now throttled to one warning per day per each unique message.


labkey-susanh (approved):
Yay for this!

PR #5814, opened 2024-08-29
include module supplied query metadata when editing XML metadata

labkey-klum:
#### Rationale
[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50362)

It was thought that initially this was a regression but after some investigation it appears that the related issue may have always been a problem. For Biologics in particular there are a number of data classes that apply module scoped query metadata. This was not getting applied to the `rawTableInfo` for the table being edited so comparisons to the incoming XML metadata would be inaccurate resulting in the related issue.

Unfortunately there isn't a direct way to create a table with some (but not all) metadata applied. This means we need to look at the list of `QueryDefinition` objects returned by the `QueryServiceImpl`. This returns a mix of module resource and definitions saved to the database. The first entry without a `rowId` distinguishes a module supplied definition (if present).

@labkey-nicka I'm happy to target this for `24.9.0` or `24.10.0` depending on perceived need and level of risk. Adding @cnathe since he recently changed some of this code in support of calculated columns.

### DIFF
--- a/query/src/org/labkey/query/MetadataTableJSON.java
+++ b/query/src/org/labkey/query/MetadataTableJSON.java
@@ -39,6 +39,7 @@ import org.labkey.api.gwt.client.model.GWTDomain;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.MetadataColumnJSON;
 import org.labkey.api.query.MetadataUnavailableException;
+import org.labkey.api.query.QueryException;
 import org.labkey.api.query.QueryParseException;
 import org.labkey.api.query.QueryService;
 import org.labkey.api.query.UserSchema;
@@ -46,6 +47,7 @@ import org.labkey.api.security.User;
 import org.labkey.api.util.Pair;
 import org.labkey.api.util.StringExpression;
 import org.labkey.api.util.StringExpressionFactory;
+import org.labkey.api.util.UnexpectedException;
 import org.labkey.api.util.logging.LogHelper;
 import org.labkey.data.xml.ColumnType;
 import org.labkey.data.xml.DefaultScaleType;
@@ -60,6 +62,7 @@ import org.w3c.dom.Text;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


query/src/org/labkey/query/MetadataTableJSON.java
```diff
@@ -106,15 +109,53 @@ public void setDefinitionFolder(String definitionFolder)
         _definitionFolder = definitionFolder;
     }
 
+    /**
+     * Returns the table info for this query. The table does not apply any external metadata except module
+     * scoped metadata (Issue : 50362).
```
labkey-nicka: nit: Make this "Issue 50362" so it is linkable or don't refer to it here since it is referred to a few lines below with a more contextually sensitive comment.

query/src/org/labkey/query/MetadataTableJSON.java
```diff
@@ -106,15 +109,53 @@ public void setDefinitionFolder(String definitionFolder)
...
+     *
+     * @throws MetadataUnavailableException
+     */
+    private static TableInfo getRawTableInfo(UserSchema schema, String tableName) throws MetadataUnavailableException
```
labkey-nicka, line 118: Seems good. Is there a junit test where we could appropriately test this processing?

labkey-nicka (approved):

query/src/org/labkey/query/MetadataTableJSON.java
labkey-klum: Good point, I can take a look. It may be difficult to simulate module query metadata but if not either I or @labkey-chrisj can potentially add to the Biologics tests.

query/src/org/labkey/query/MetadataTableJSON.java
```diff
@@ -106,15 +109,53 @@ public void setDefinitionFolder(String definitionFolder)
...
+    {
+        TableInfo tableInfo = schema.getTable(tableName, null, false, true);
+        if (null == tableInfo)
+            throw new MetadataUnavailableException("No such table: " + tableName);
+
+        List<QueryDef> queryDefs = QueryServiceImpl.get().findMetadataOverrideImpl(schema, tableName, false, false, null);
```
cnathe: it looks like `findMetadataOverrideImpl` calls two helpers within it to get the queryDefs: 1)findMetadataOverrideInModules and 2) findMetadataOverrideInDatabase. Would this be worth making `findMetadataOverrideInModules` public and calling that here?

cnathe (approved):

query/src/org/labkey/query/MetadataTableJSON.java
labkey-klum: Sure, good idea. It looks like we would bypass some of the outer cacheing that is in place but it looks like the module resources has its own separate cache.

PR #5817, opened 2024-08-30
Stop building Radeox as a separate JAR

labkey-jeckels:
#### Rationale
We've been awkwardly patching Radeox by maintaining the code as part of the Wiki module, but separate from the `src` root. We've published updates to Artifactory, which were then pulled in as a dependency of the `Core` module. We can streamline the process and also clean up the Radeox codebase.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51190

#### Changes
- Move code from side location in `Wiki` to the regular source root in `Core`
- Stop referencing Radeox as if it was a third-party dependency
- Remove a bunch of macros that never worked (due to assumptions about config files in a `./conf` directory relative to the working directory) and were useless anyway (CD shopping comparison links, etc)
- Switch Radeox tests to JUnit 4 and register them as unit tests
- Simplify and flatten config
- Get rid of example code, docs, and other unneeded resources
- Assorted code updates (generics, useless returns, etc)


labkey-jeckels:
@labkey-mohara FYI, this will remove some of the useless macros that we currently list here:

https://www.labkey.org/Documentation/wiki-page.view?name=moreWikiSyntax

See `list-of-macros` for the latest:

anchor
code
comment
div
file-path
h1
image
labkey
link
list-of-macros
mailto
new-tab-link
quote
span
study
table
video

Let me know if you see any unexpected behavior in wikis after this is merged.

labkey-adam (changes requested):
Code review looks reasonable. I built and ran locally. Looks like `{span}`, `{comment}`, `{anchor}`, `{code}`, `{div}`, `{video}`, `{link}`, and `{table}` macros are used extensively in the documentation, so I believe these need to be restored, at a minimum. There may be others...

labkey-jeckels:
@labkey-adam what are you seeing? They appear to be working for me:

```
{code}
public static void main (String... args) {
    System.out.println("Howdy");
}
{code}

{table}
Header|Other
Value|2
{table}
```
Screenshot 2024-09-02 at 1 09 36 PM

You may need to clean the Core module. I certainly did. I don't think Gradle is smart enough to clear out everything for a change like this.

labkey-adam:
None of the macros are working. I wasn't sure from your comment which ones you removed. I'll do a clean build and try again after lunch.

labkey-adam (approved):
Confirmed that all key macros seem to be working after cleaning the core module. Looks like `list-of-macros` is now outputting strange text ("unexplained, lazy programmer, probably [funzel]") in the "Parameters" column where the previous value was blank (`link`, `list-of-macros`, `rfc`). I suggest correcting this, but no need for a re-review.

labkey-jeckels:
> Confirmed that all key macros seem to be working after cleaning the core module. Looks like `list-of-macros` is now outputting strange text ("unexplained, lazy programmer, probably [funzel]") in the "Parameters" column where the previous value was blank (`link`, `list-of-macros`, `rfc`). I suggest correcting this, but no need for a re-review.

I see the same thing in develop, but it's easy enough to fix. Also, in testing, I discovered the `rfc` macro sends you to a URL that doesn't work. Bye bye, `rfc` macro!

PR #5822, opened 2024-09-03
Initial Plate set cache

labkey-klum:
#### Rationale
Establishes an initial cache for plate sets and begins to use the cache for referencing (and uncaching) plate set objects. A follow on phase will be to move properties cached at the plate level into the plate sets. The only planned work at this time is for plate metadata to be moved (but not limited to).

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateCache.java
+++ b/assay/src/org/labkey/assay/plate/PlateCache.java
@@ -22,11 +22,12 @@ import org.labkey.assay.plate.model.PlateBean;
 import org.labkey.assay.plate.query.PlateTable;
 import org.labkey.assay.query.AssayDbSchema;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 public class PlateCache
 {
@@ -36,7 +37,7 @@ public class PlateCache
 
     private static class PlateLoader implements CacheLoader<String, Plate>
     {
-        private final Map<Container, List<Plate>> _containerPlateMap = new HashMap<>();            // internal collection to help un-cache all plates for a container
+        private final Map<Container, Set<Integer>> _containerPlateMap = new HashMap<>();            // internal collection to help un-cache all plates for a container
 
         @Override
         public Plate load(@NotNull String key, @Nullable Object argument)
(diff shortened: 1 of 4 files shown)


assay/src/org/labkey/assay/plate/query/PlateSetTable.java
```diff
@@ -218,9 +221,27 @@ protected Map<String, Object> deleteRow(
...
                 return returnMap;
             }
         }
+
+        @Override
+        protected Map<String, Object> updateRow(User user, Container container, Map<String, Object> row, @NotNull Map<String, Object> oldRow, @Nullable Map<Enum, Object> configParameters) throws InvalidKeyException, ValidationException, QueryUpdateServiceException, SQLException
+        {
+            Integer plateSetId = (Integer) oldRow.get("rowId");
+            PlateSet plateSet = PlateManager.get().getPlateSet(container, plateSetId);
+            if (plateSet == null)
```
labkey-nicka: Hmm, when would this be the case? Seems like we should throw here instead? Something to the effect of "Unable to resolve plate set by row id in folder path.".

That said, consider using `PlateManager.get().requirePlateSet()`. You'll need to expose package private.

labkey-nicka (approved):
Looks good. Please kickoff a run against the full biologics suite plus the endpoint integration tests so we can see how it goes.

assay/src/org/labkey/assay/plate/PlateCache.java
```diff
@@ -173,12 +174,25 @@ public static void uncache(Container c)
...
+            for (Integer rowId : rowIds)
             {
-                uncache(c, plate);
+                uncache(c, rowId);
             }
-            _loader._containerPlateMap.remove(c);
+        }
+    }
+
+    public static void uncache(Container c, int rowId)
```
labkey-nicka, line 189: With these changes can the TODO comment in `PlateCache.java` be removed?

labkey-nicka (approved):
Nice. Glad this was caught.

assay/src/org/labkey/assay/plate/query/PlateSetTable.java
labkey-klum: good point, thanks

assay/src/org/labkey/assay/plate/PlateCache.java
labkey-klum: Correct, as long as we use this method to uncache, we won't run the risk of caching just to uncache.

PR #5823, opened 2024-09-03
Issue 51164: Handle query name changes for query.QueryDef rows

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51164

This issue was exposed by testing for calculated fields but was present prior to the calculated fields UI dev changes. Several of our data types (sample type, dataset, data class, assay design) allow for name changes via the Field Editor. If that table (or set of tables in the assay design case) has custom user-defined query metadata XML, that metadata will get orphaned in the DB on a query name changes or a delete event. This PR fixes that by adding in a QueryDefQueryChangeListener for those two events to update the DB QueryDef object schema or query name.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5807

#### Changes
- MetadataTableJSON to update XML tableName if applicable
- add QueryDefQueryChangeListener to handle rename of query/schema and delete of queries to clean up QueryDef rows


cnathe:
Test plan: 
- [x] verify user defined XML metadata is retained on rename of the following data types:
  - [x] sample type
  - [x] data class
  - [x] study dataset
  - [x] list
      - [x] Issue: After a rename (which is successful), the calculated columns no longer appear in the designer.
  - [x] assay design (for batch, run, and results field domains)
     - [ ] Calculated batch and run fields are not shown in grids, but after renaming and then re-editing, the calculated fields are retained, so I think this bug fix is good. (Issue [51202](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51202))
- [x] verify calculated fields are retained and can be added to each data type during a rename event
- [x] verify user defined XML metadata is removed from query.QueryDef table on delete of data type

api/src/org/labkey/api/query/QueryService.java
```diff
@@ -226,7 +226,7 @@ UserSchema createLinkedSchema(User user, Container container, String name, Strin
 
     TableType parseMetadata(String metadataXML, Collection<QueryException> errors);
 
-    void saveCalculatedFieldsMetadata(String schemaName, String queryName, List<? extends GWTPropertyDescriptor> fields, boolean isUserDefinedQuery, User user, Container container) throws MetadataUnavailableException;
+    void saveCalculatedFieldsMetadata(String schemaName, String queryName, String updatedQueryName, List<? extends GWTPropertyDescriptor> fields, boolean isUserDefinedQuery, User user, Container container) throws MetadataUnavailableException;
```
labkey-klum: `@Nullable` annotation for `updatedQueryName`?

labkey-klum (approved):

api/src/org/labkey/api/query/QueryService.java
cnathe: done

PR #5825, opened 2024-09-04
RecordFactory: an ObjectFactory that handles Java records

labkey-adam:
#### Rationale
`Selector` methods can produce beans and simple classes, but can't produce records. `RecordFactory` enables production of records from Selector and other code paths that use `ObjectFactory.getFactory()`.

Remove `ObjectFactory.handleArray()` and all implementations. Arrays are lame and this method was called in exactly one place, and that code doesn't even care about arrays.

Clean up some warnings and an NPE waiting to happen.

### DIFF
--- a/api/src/org/labkey/api/data/RecordFactory.java
+++ b/api/src/org/labkey/api/data/RecordFactory.java
@@ -147,14 +147,15 @@ public class RecordFactory<K> implements ObjectFactory<K>
                 Assert.assertEquals(users.get(0), factory.handle(rs));
             }
             MiniUser randomUser = users.get((int)(Math.random() * users.size()));
-            MiniUser selectedUser = new TableSelector(CoreSchema.getInstance().getTableInfoUsers(), new SimpleFilter(FieldKey.fromString("UserId"), randomUser.userId), null).getObject(MiniUser.class);
+            MiniUser selectedUser = new TableSelector(CoreSchema.getInstance().getTableInfoUsers(), new SimpleFilter(FieldKey.fromString("UserId"), randomUser.userid), null).getObject(MiniUser.class);
             Assert.assertEquals(randomUser, selectedUser);
 
             // Test fromMap() variant (should ignore selectedUser)
             assertEquals(adHocUser, factory.fromMap(selectedUser, adHocMap));
         }
 
-        public record MiniUser(String firstName, String lastName, Date LastLogin, int userId)
+        // Simple test record. Weird casing is intentional to test case-insensitivity.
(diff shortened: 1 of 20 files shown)


api/src/org/labkey/api/data/RecordFactory.java
```diff
@@ -0,0 +1,160 @@
...
+            }
+            MiniUser randomUser = users.get((int)(Math.random() * users.size()));
+            MiniUser selectedUser = new TableSelector(CoreSchema.getInstance().getTableInfoUsers(), new SimpleFilter(FieldKey.fromString("UserId"), randomUser.userId), null).getObject(MiniUser.class);
+            Assert.assertEquals(randomUser, selectedUser);
+
+            // Test fromMap() variant (should ignore selectedUser)
+            assertEquals(adHocUser, factory.fromMap(selectedUser, adHocMap));
+        }
+
+        public record MiniUser(String firstName, String lastName, Date LastLogin, int userId)
```
labkey-jeckels:
```suggestion
        public record MiniUser(String firstName, String lastName, Date lastLogin, int userId)
```

api/src/org/labkey/api/data/RecordFactory.java
```diff
@@ -0,0 +1,160 @@
...
+ */
+public class RecordFactory<K> implements ObjectFactory<K>
+{
+    private final Constructor<K> _constructor;
+    private final Parameter[] _parameters;
+    private final List<Field> _fields;
+
+    public RecordFactory(Class<K> clazz)
+    {
+        assert clazz.isRecord() : clazz + " is not a record!";
```
labkey-jeckels: Per discussion this afternoon, switching to `IllegalArgumentException` seems microscopically better. This will be a very cheap thing to check, and why not match behavior between dev and prod?

labkey-jeckels (approved):

api/src/org/labkey/api/data/RecordFactory.java
labkey-adam: I'm purposefully using a mix of casing here and in the map to ensure proper handling

api/src/org/labkey/api/data/RecordFactory.java
labkey-adam: Why not, indeed?

api/src/org/labkey/api/data/RecordFactory.java
labkey-jeckels: Makes sense. Might be worth a comment to prevent someone with good intentions from destroying the coverage in the future.

PR #5827, opened 2024-09-04
Issue 51207: SampleType import of file-field files doesn't work with special characters

XingY:
#### Rationale
When a sample field file name contains special character such as %, importing it resulting in the file field not populated. This is due to the exception that arises from Path.decode an invalid webdav path. This PR applies FileUtil.encodeForURL for filename before Path.decode. 

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2037

#### Changes
-


labkey-jeckels (approved):
This change looks OK, but can you add some comments to `SimpleTranslator.getFileRootSubstitutedFilePath()` on what it does and/or add some unit tests for it? It's hard to track which places are passing around URLs and which expect local file paths.

XingY:
> This change looks OK, but can you add some comments to `SimpleTranslator.getFileRootSubstitutedFilePath()` on what it does and/or add some unit tests for it? It's hard to track which places are passing around URLs and which expect local file paths.

Good reminder about URL vs file path. My change actually broken the URL scenario since it resulted in double encoding. I've added a work around that should handle both. getFileRootSubstitutedFilePath is meant to be used for file path only, for fileLink fields. I've added comment and junit tests for it.

PR #5837, opened 2024-09-07
Rework categories option in advanced search

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51111

Add "Projects and Folders" as an additional search category for advanced search. More importantly, switch from a hard-coded list of categories (including categories that might not even exist in the deployment) to a dynamically generated list based on existing categories that specify "Show in Advanced Options". Update some descriptions to plurals and modern conventions.

Note: Advanced search is displaying the same categories as before with the addition of "Projects and Folders", but it's now trivial to add other categories to the list.

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpDataImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpDataImpl.java
@@ -134,14 +134,14 @@ public class ExpDataImpl extends AbstractRunItemImpl<Data> implements ExpData
         }
     }
 
-    public static final SearchService.SearchCategory expDataCategory = new SearchService.SearchCategory("data", "ExpData") {
+    public static final SearchService.SearchCategory expDataCategory = new SearchService.SearchCategory("data", "ExpDatas", false) {
         @Override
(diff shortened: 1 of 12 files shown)


experiment/src/org/labkey/experiment/api/ExpDataImpl.java
```diff
@@ -134,14 +134,14 @@ public Class<? extends Permission> getPermissionClass()
         }
     }
 
-    public static final SearchService.SearchCategory expDataCategory = new SearchService.SearchCategory("data", "ExpData") {
+    public static final SearchService.SearchCategory expDataCategory = new SearchService.SearchCategory("data", "ExpDatas", false) {
```
labkey-susanh: "data" is already plural. The extra "s" is grating (to some).

labkey-adam: Addressed

labkey-susanh (approved):

PR #5839, opened 2024-09-07
Support description and last-used tracking for API keys

labkey-adam:
#### Rationale
Server-side part of https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50569

- Add `LastUsed` and `Description` columns to `core.ApiKeys` table
- Update `LastUsed` column on successful login via an API key, throttled to max once every 10 minutes
- Update `CreateApiKeyAction` to take a description and save it on API key creation, if provided
- Display `LastUsed` and `Description` columns on "External Tool Access" page (free functionality!)


core/src/org/labkey/core/security/SecurityController.java
```diff
@@ -2429,7 +2443,7 @@ public Object execute(CreateApiKeyForm form, BindException errors)
                     if (!AppProps.getInstance().isAllowApiKeys())
                         throw new NotFoundException("Creation of API keys is disabled");
 
-                    apiKey = ApiKeyManager.get().createKey(getUser());
+                    apiKey = ApiKeyManager.get().createKey(getUser(), form.getDescription());
```
labkey-susanh, line 2446: Is it worth trimming the description and/or verifying its length will fit in the field (either here or in `ApiKeyManager`?

labkey-susanh (approved):

labkey-jeckels (approved):

core/src/org/labkey/core/security/SecurityController.java
labkey-adam: Sure

PR #5840, opened 2024-09-08
Issues 51215 & 51212: Metrics and query for tracking effects of Issue 51210 

labkey-susanh:
#### Rationale
Issue [51215](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51215)
Issue [51212](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51212)

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/667

#### Changes
- Add `LineageUpdateDiscrepancies` query
- Add metrics for number-name sample types
- Add metric for sites potentially affected by 51210

### DIFF
--- a/experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
+++ b/experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
@@ -1,0 +1,46 @@
+-- lineage update events for samples that had lineage and were involved in a transaction that
+-- updated lineage on some but not all samples.
+-- Includes samples that may have had their lineage intentionally removed before this transaction happened.
+WITH MixedTransactions AS
+         (SELECT a1.transactionId
+          FROM (SELECT DISTINCT transactionId
+                FROM auditlog.SampleTimelineEvent
+                -- created before the bug fix was introduced
+                WHERE islineageupdate = true
+                  AND comment = 'Sample was updated.'
+                  AND created < '2024-10-01') a1
+                   JOIN
+               -- but have associated entries that are not lineage updates
+                   (SELECT DISTINCT transactionid
+                    FROM auditlog.SampleTimelineEvent
+                    WHERE islineageupdate = false) a2
+               ON a1.transactionid = a2.transactionid
+          )
+SELECT
+    rowid,
+    container,
+   sampletype,
+   sampletypeid,
+   samplename,
+   d.sampleid,
+    latestLineageEventDate,
+   oldrecordmap,
+   newrecordmap
+   --transactionid,  TODO this should be the transactionid of the problematic transaction
+
+FROM (
+         SELECT a.sampleId, max(created) as latestLineageEventDate
+         FROM auditlog.SampleTimelineEvent a
+         WHERE sampleId IN (
+             -- samples involved in transactions where some, but not all, samples had lineage update events prior to the fix for Issue 51210
+             SELECT DISTINCT sampleId
+             FROM auditLog.SampleTimelineEvent sa
+             WHERE transactionId IN ( SELECT transactionId FROM MixedTransactions )
+         )
+         -- samples had an audit log creating lineage
+         AND (a.newrecordmap LIKE '%MaterialInputs%' OR a.newrecordmap LIKE '%DataInputs%')
+         GROUP BY sampleid
+     ) b
+
+         JOIN auditLog.SampleTimelineEvent d on d.sampleid = b.sampleid AND d.created = b.latestLineageEventDate
+ORDER BY sampleId
--- a/experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
+++ b/experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
@@ -5,8 +5,7 @@ WITH MixedTransactions AS
                  (
                      SELECT DISTINCT transactionId
                      FROM auditlog.SampleTimelineEvent
-                     -- created before the Issue 51210 fix was introduced
-                     WHERE islineageupdate = true AND comment = 'Sample was updated.' and created < '2024-10-01'
+                     WHERE islineageupdate = true AND comment = 'Sample was updated.'
                  ) a1
                      JOIN
                  -- but have associated entries that are not lineage updates
--- a/experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
+++ b/experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
@@ -1,46 +1,55 @@
--- lineage update events for samples that had lineage and were involved in a transaction that
--- updated lineage on some but not all samples.
--- Includes samples that may have had their lineage intentionally removed before this transaction happened.
+-- transaction ids where some, but not all, samples had lineage update events prior to the fix for Issue 51210
 WITH MixedTransactions AS
-         (SELECT a1.transactionId
-          FROM (SELECT DISTINCT transactionId
-                FROM auditlog.SampleTimelineEvent
-                -- created before the bug fix was introduced
-                WHERE islineageupdate = true
-                  AND comment = 'Sample was updated.'
-                  AND created < '2024-10-01') a1
-                   JOIN
-               -- but have associated entries that are not lineage updates
-                   (SELECT DISTINCT transactionid
-                    FROM auditlog.SampleTimelineEvent
-                    WHERE islineageupdate = false) a2
-               ON a1.transactionid = a2.transactionid
-          )
-SELECT
-    rowid,
-    container,
-   sampletype,
-   sampletypeid,
-   samplename,
-   d.sampleid,
-    latestLineageEventDate,
-   oldrecordmap,
-   newrecordmap
-   --transactionid,  TODO this should be the transactionid of the problematic transaction
+         (
+             SELECT a1.transactionId FROM
+                 (
+                     SELECT DISTINCT transactionId
+                     FROM auditlog.SampleTimelineEvent
+                     -- created before the Issue 51210 fix was introduced
+                     WHERE islineageupdate = true AND comment = 'Sample was updated.' and created < '2024-10-01'
+                 ) a1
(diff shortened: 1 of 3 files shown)


experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
```diff
@@ -0,0 +1,55 @@
+-- transaction ids where some, but not all, samples had lineage update events prior to the fix for Issue 51210
+WITH MixedTransactions AS
+         (
+             SELECT a1.transactionId FROM
+                 (
+                     SELECT DISTINCT transactionId
+                     FROM auditlog.SampleTimelineEvent
+                     -- created before the Issue 51210 fix was introduced
+                     WHERE islineageupdate = true AND comment = 'Sample was updated.' and created < '2024-10-01'
```
XingY: Do we know for sure that all clients will get the fix build deployed before 10/1?

experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
```diff
@@ -0,0 +1,55 @@
...
+         FROM auditlog.SampleTimelineEvent a
+         WHERE sampleId IN (
+             SELECT DISTINCT sampleId
+             FROM auditLog.SampleTimelineEvent sa
+             WHERE transactionId IN (
+                 SELECT transactionId FROM MixedTransactions
+             )
+         )
+           -- samples had an audit log creating lineage
+           AND (a.newrecordmap LIKE '%MaterialInputs%' OR a.newrecordmap LIKE '%DataInputs%')
```
XingY, line 46: seems like there would be false positive record returned, but we are ok with that?

XingY, line 46: Seems like this won't detect lineage events that use import alias.

experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -666,92 +671,136 @@ WHERE op.propertyid IN (
...
+                if (schema.getSqlDialect().isPostgreSQL()) // SQLServer does not support regular expression queries
+                {
+                    Collection<Map<String, Object>> numSampleCounts = new SqlSelector(schema, """
+                            SELECT totalCount, numberNameCount FROM
+                                        (SELECT cpastype, COUNT(*) AS totalCount from exp.material GROUP BY cpastype) t
+                                    JOIN
+                                        (SELECT cpastype, COUNT(*) AS numberNameCount FROM exp.material m WHERE m.name SIMILAR TO '[0-9.]*' GROUP BY cpastype) ns
+                                    ON t.cpastype = ns.cpastype""").getMapCollection();
+                    results.put("sampleSetWithNumberNamesCount", numSampleCounts.size());
+                    results.put("sampleSetWithOnlyNumberNamesCount", numSampleCounts.stream().filter(map -> map.get("totalCount") == map.get("numberNameCount")).count());
```
XingY: also check `map.get("totalCount") > 0`?

experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
labkey-susanh: Yes, we are OK with that.

experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
labkey-susanh: Cloud instances should get it before then as it's targeting 24.7 and 24.9.  It could be that on-prem clients will not deploy this until much later, but since this query is mainly for immediate use in discovering what samples may need follow-up after the release, I think this is OK.  Another option would be to leave off the date and potentially have more false positives (since after the fix is introduced the combination of `true` and `false` for `islineageupdate` is as expected). Users can also choose to create a new custom query based on this one that removes or modifies the date.  I'll check with @labkey-hannah to see what is preferred.

experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
labkey-susanh: In the audit log it looks like we record the changes using "MaterialInputs" and "DataInputs" even when an import alias is used.

experiment/src/org/labkey/experiment/ExperimentModule.java
labkey-susanh: Yes, good thought.

experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
```diff
@@ -0,0 +1,54 @@
...
+         GROUP BY sampleid
+     ) b
+         JOIN auditLog.SampleTimelineEvent d on d.sampleid = b.sampleid AND d.created = b.latestLineageEventDate
+         JOIN (
+    SELECT DISTINCT sampleId, transactionId, oldrecordmap, newrecordmap, rowId
+    FROM auditlog.SampleTimelineEvent WHERE transactionID IN (SELECT transactionId FROM MixedTransactions)
+) m ON m.sampleId = b.sampleId
+ORDER BY problemTransaction, sampleId
```
cnathe, line 54: was there going to be a where clause in here somewhere with a date filter? once we fix the issue, I believe the audit log events will still fit this criteria but would be all false positives (but maybe you adjusted the query in another way to account for that).

cnathe (approved):

experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
XingY: Hmm, strange, I'm seeing import alias

XingY: NVM, it seems both MaterialInputs/ as well as alias are recorded in audit log

experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
labkey-susanh: I had the date filter in there, but checked with Hannah who indicated it should not be necessary. There may be false positives, but as long as we know about it we should be able to deal with them.

experiment/resources/queries/auditLog/LineageUpdateDiscrepancies.sql
labkey-susanh: With a file import? I'm seeing MaterialInputs for both an insert and update.

XingY (approved):

experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -666,92 +671,138 @@ WHERE op.propertyid IN (
...
+                {
+                    Collection<Map<String, Object>> numSampleCounts = new SqlSelector(schema, """
+                            SELECT totalCount, numberNameCount FROM
+                                        (SELECT cpastype, COUNT(*) AS totalCount from exp.material GROUP BY cpastype) t
+                                    JOIN
+                                        (SELECT cpastype, COUNT(*) AS numberNameCount FROM exp.material m WHERE m.name SIMILAR TO '[0-9.]*' GROUP BY cpastype) ns
+                                    ON t.cpastype = ns.cpastype""").getMapCollection();
+                    results.put("sampleSetWithNumberNamesCount", numSampleCounts.size());
+                    results.put("sampleSetWithOnlyNumberNamesCount", numSampleCounts.stream().filter(
+                            map -> (Integer) map.get("totalCount") > 0 && map.get("totalCount") == map.get("numberNameCount")
```
cnathe: FYI, when I try to look at usage statistics locally, I get the following exception at this line:
```
ERROR ExceptionUtil            2024-09-09T11:50:18,439     http-nio-8080-exec-9 : Exception detected and logged to mothership with error code: RHMWSN
java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.lang.Integer (java.lang.Long and java.lang.Integer are in module java.base of loader 'bootstrap')
	at org.labkey.experiment.ExperimentModule.lambda$startupAfterSpringConfig$2(ExperimentModule.java:687) ~[experiment-24.7-SNAPSHOT.jar:?]
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:178) ~[?:?]
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:992) ~[?:?]
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509) ~[?:?]
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[?:?]
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921) ~[?:?]
	at java.base/java.util.stream.ReduceOps$5.evaluateSequential(ReduceOps.java:258) ~[?:?]
	at java.base/java.util.stream.ReduceOps$5.evaluateSequential(ReduceOps.java:248) ~[?:?]
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
	at java.base/java.util.stream.ReferencePipeline.count(ReferencePipeline.java:709) ~[?:?]
	at org.labkey.experiment.ExperimentModule.lambda$startupAfterSpringConfig$4(ExperimentModule.java:688) ~[experiment-24.7-SNAPSHOT.jar:?]
	at org.labkey.core.admin.usageMetrics.UsageMetricsServiceImpl.lambda$getModuleUsageMetrics$2(UsageMetricsServiceImpl.java:65) ~[core-24.7-SNAPSHOT.jar:?]
	at java.base/java.lang.Iterable.forEach(Iterable.java:75) ~[?:?]
```

cnathe: fix verified

PR #5842, opened 2024-09-09
Register the audit domain kinds earlier

labkey-klum:
#### Rationale
In this previous [PR](https://github.com/LabKey/platform/pull/5790) we deferred initialization and fixup of audit domains until after the server had completed startup. However, domain kinds still need to registered early. Otherwise existing domain descriptors may not get initialized properly in the cache.

I also added a note to clean up how domain kinds get associated with each audit provider but plan to do that on `develop`.

### DIFF
--- a/api/src/org/labkey/api/audit/AbstractAuditTypeProvider.java
+++ b/api/src/org/labkey/api/audit/AbstractAuditTypeProvider.java
@@ -15,6 +15,7 @@
  */
 package org.labkey.api.audit;
 
+import org.jetbrains.annotations.NotNull;
 import org.labkey.api.audit.data.DataMapColumn;
 import org.labkey.api.audit.data.DataMapDiffColumn;
 import org.labkey.api.audit.query.AbstractAuditDomainKind;
@@ -99,7 +100,7 @@ public abstract class AbstractAuditTypeProvider implements AuditTypeProvider
         this(null);
     }
 
-    public AbstractAuditTypeProvider(AbstractAuditDomainKind domainKind)
+    public AbstractAuditTypeProvider(@NotNull AbstractAuditDomainKind domainKind)
     {
         // TODO : consolidate domain kind initialization to either this constructor or to override
         // getDomainKind.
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/audit/AbstractAuditTypeProvider.java
```diff
@@ -101,21 +101,23 @@ public AbstractAuditTypeProvider()
...
+        _domainKind = domainKind;
+        // Register the DomainKind
+        PropertyService.get().registerDomainKind(getDomainKind());
     }
 
     protected AbstractAuditDomainKind getDomainKind()
     {
-        assert null != domainKind;
-        return domainKind;
+        assert null != _domainKind;
```
labkey-nicka: Seems like this assertion (or throw) should be done in the constructor since `_domainKind` is `final` rather than lazily if/when someone calls `getDomainKind()`.

Possibly a part of the subsequent clean up you've called out.

api/src/org/labkey/api/audit/AbstractAuditTypeProvider.java
```diff
@@ -101,21 +101,23 @@ public AbstractAuditTypeProvider()
 
     public AbstractAuditTypeProvider(AbstractAuditDomainKind domainKind)
```
labkey-nicka: nit: Consider marking `@NotNull AbstractAuditDomainKind domainKind`.

Possibly a part of the subsequent clean up you've called out.

labkey-nicka (approved):

api/src/org/labkey/api/audit/AbstractAuditTypeProvider.java
labkey-klum: agreed, done.

api/src/org/labkey/api/audit/AbstractAuditTypeProvider.java
labkey-klum: I'm going to throw for now, but I'm thinking of getting rid of the constructor initialization in the future.

PR #5845, opened 2024-09-10
Handle null values for Timestamp JDBC parameters

XingY:
#### Rationale
There are a couple setObject functions that each takes a different set of params. PR  https://github.com/LabKey/platform/pull/5805 modified all of those to call our own override version, which all end up calling the 2-parametere version of the util, if targetSqlType == Types.TIMESTAMP. This is not working well for null values (not sure why, but the old way uses the N param function for null values), see the following test failure. 
[ETLTest.testStoredProcTransforms](https://teamcity.labkey.org/viewLog.html?buildId=3175197&tab=buildResultsDiv&buildTypeId=LabKey_247Release_Premium_ModulesSuites_DataIntegrationSqlserver#testNameId-5800387439352806963)

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5805

#### Changes
- don't use the 2 parametered setObject when value is null


labkey-jeckels (approved):

labkey-adam (approved):

labkey-adam:
@XingY thanks for tracking this down and fixing!

PR #5854, opened 2024-09-12
Calculated fields issue fixes for 24.10

cnathe:
#### Rationale
see ui-components PR for rationale

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1576

#### Changes
- DomainUtil.getCalculatedFieldsForTableInfo to use last metadata object in collection since DB metadata is added last
- add null check before setting rangeURI for calculated field
- show calculated fields in default view for Assay Run and Batch domains
- saveCalculatedFieldsMetadata to account for removing all calculated fields from a domain


assay/api-src/org/labkey/api/assay/AssayResultTable.java
```diff
@@ -240,12 +240,26 @@ else if (baseColumn.isMvIndicatorColumn())
             if (!prop.isHidden())
                 visibleColumns.add(FieldKey.fromParts("Run", prop.getName()));
         }
+        TableInfo runTable = schema.getTable(AssayProtocolSchema.RUNS_TABLE_NAME);
+        if (runTable != null)
+        {
+            calculatedFieldKeys = DomainUtil.getCalculatedFieldsForDefaultView(runTable);
+            for (FieldKey calculatedFieldKey : calculatedFieldKeys)
+                visibleColumns.add(FieldKey.fromParts("Run", calculatedFieldKey.getName()));
+        }
```
labkey-matthewb, line 249: OK, seems different from what we do elsewhere?  Why are these special?

labkey-matthewb (approved):

assay/api-src/org/labkey/api/assay/AssayResultTable.java
cnathe: These are the calculated fields from the Run domain being added to the default assay Results view. I debated if we should add these here, but since we do add the custom fields from the Run domain to the Results view, I thought it would be best to keep that in sync with the calculated fields as well.

PR #5866, opened 2024-09-17
clean up use of Renderable with DOM.Attribute

labkey-matthewb:
#### Rationale
Html rendered into an DOM.Attribute needs to be double-encoded.

#### Related Pull Requests
- 

#### Changes
-


labkey-adam (approved):
Thanks for the fix... pretty sure I tested this as working during my refactoring 🤷

PR #5872, opened 2024-09-18
Lineage Relationships: Optionally Require during Sample Registration

XingY:
#### Rationale
The feature to start enforcing that a sample or data must have certain sample/data as parent input during insert and update actions. The configuration is done on designer page when user set a parent alias for a data type to 'required'. The apis and data interators have been updated to ensure lineage is present during insert/update/derive actions.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1585
- https://github.com/LabKey/labkey-ui-premium/pull/539
- https://github.com/LabKey/platform/pull/5872
- https://github.com/LabKey/limsModules/pull/730
- https://github.com/LabKey/testAutomation/pull/2072

#### Changes
- modified importalias column to support storing 'required' flag
- created ImportAliasesDeserializer to support parsing legacy and new alias json formats
- modified sample and data insert/update api to check for required lineage
- modified sample and data domain update actions to check data for required lineage
- modified folder import to defer required lineage check
- fixed query change listener for alias update


api/src/org/labkey/api/exp/api/DataClassDomainKindProperties.java
```diff
@@ -159,12 +167,26 @@ public void setStrictFieldValidation(boolean strictFieldValidation)
...
     {
         this.importAliases = importAliases;
     }
 
-    public Map<String, String> getImportAliases()
+    @JsonIgnore
+    public Map<String, String> getImportAliasesMap()
+    {
+        if (getImportAliases() == null)
+            return null;
```
labkey-susanh, line 181: Looks like `getImportAliasesMap` is annotated as `@NotNull` in other classes. Also, you could grab the import aliases just once at the top here.

api/src/org/labkey/api/exp/api/ExperimentJSONConverter.java
```diff
@@ -919,20 +925,50 @@ private static Map<String, String> sanitizeAliases(Map<String, String> importAli
...
                     trimmedVal.equalsIgnoreCase(MATERIAL_INPUTS_PREFIX + NEW_SAMPLE_TYPE_ALIAS_VALUE))
             {
-                trimmedVal = MATERIAL_INPUTS_PREFIX + currentAliasName;
+                trimmedVal = MATERIAL_INPUTS_ALIAS_PREFIX + currentAliasName;
             }
             else if (trimmedVal.equalsIgnoreCase(NEW_DATA_CLASS_ALIAS_VALUE) ||
                     trimmedVal.equalsIgnoreCase(DATA_INPUTS_PREFIX + NEW_DATA_CLASS_ALIAS_VALUE))
             {
-                trimmedVal = DATA_INPUTS_PREFIX + currentAliasName;
+                trimmedVal = DATA_INPUTS_ALIAS_PREFIX + "/" + currentAliasName;
```
labkey-susanh: Doesn't this add the "/" twice?

api/src/org/labkey/api/exp/api/ExperimentJSONConverter.java
```diff
@@ -107,6 +109,9 @@ public class ExperimentJSONConverter
     public static final String MATERIAL_OUTPUTS = "materialOutputs";
     public static final String STEPS = "steps";
 
+    public static final String MATERIAL_INPUTS_ALIAS_PREFIX = MATERIAL_INPUTS + "/";
+    public static final String DATA_INPUTS_ALIAS_PREFIX = DATA_INPUTS + "/";
```
labkey-susanh, line 113: Oh boooo for case differences here. `DATA_INPUTS_PREFIX` uses `DataInputs` but here we need `dataInputs`? Since we aren't changing how the aliases are stored, I guess there's no getting around this unless we can or want to do case-insensitive matching.

api/src/org/labkey/api/exp/api/SampleTypeDomainKindProperties.java
```diff
@@ -122,12 +130,26 @@ public int getRowId()
...
     {
         this.importAliases = importAliases;
     }
 
-    public Map<String, String> getImportAliases()
+    @JsonIgnore
+    public Map<String, String> getImportAliasesMap()
+    {
+        if (getImportAliases() == null)
+            return null;
```
labkey-susanh, line 145: Same thing here about `EmptyMap` vs. @`NotNull`

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -1022,13 +1019,36 @@ public String getImportAliasJson()
     }
 
     @Override
-    public @NotNull Map<String, String> getImportAliasMap() throws IOException
+    public @NotNull Map<String, String> getImportAliases() throws IOException
```
labkey-susanh, line 1022: Would be nice to share some of this logic for the import aliases between data classes and sample types.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8074,25 +8129,31 @@ public ValidationException updateDataClass(@NotNull Container c, @NotNull User u
...
             }
             dataClass.setDescription(options.getDescription());
             dataClass.setNameExpression(options.getNameExpression());
             dataClass.setSampleType(options.getSampleType());
             dataClass.setCategory(options.getCategory());
+            if (options.getImportAliases() != null && !options.getImportAliases().isEmpty())
+            {
+                try
+                {
+                    Map<String, Map<String, Object>> newAliases = options.getImportAliases();
```
labkey-susanh: Could do a small optimization of extracting the import aliases once and using it in all these various places

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -9169,7 +9230,69 @@ public void registerNameExpressionType(String dataType, String schemaName, Strin
...
+        }
+
+        return new Pair<>(requiredSampleParentCount, requiredDataParentCount);
+    }
+
+    private Map<String, Object> getParentAliasMetrics()
+    {
+        Map<String, Object> metrics = new HashMap<>();
+        Pair<Long, Long> samplesMetrics = getParentAliasMetrics(getTinfoSampleType(), "materialparentimportaliasmap");
+        if (samplesMetrics.first > 0)
```
labkey-susanh: Seems like we should report the values even if they are 0.

labkey-susanh (changes requested):

api/src/org/labkey/api/exp/api/DataClassDomainKindProperties.java
XingY: Since this is the domain options bean class, I think it's fine to have it return null to reflect what's provided by api. Updated annotation to `@Nullable`

api/src/org/labkey/api/exp/api/ExperimentJSONConverter.java
XingY: good catch. fixed

labkey-susanh (approved):

PR #5755, opened 2024-08-12
Issue 45594: Update lineage for run/result sample lookups

labkey-nicka:
#### Rationale
This addresses [Issue 45594](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45594) by updating the lineage when assay run/result properties are mutated.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5755
- https://github.com/LabKey/limsModules/pull/563
- https://github.com/LabKey/provenance/pull/187
- https://github.com/LabKey/commonAssays/pull/788

#### Changes
- Process and modify lineage in `AbstractAssayProvider.updatePropertyLineage()`
- Update `ExperimentService.findExpMaterial()` to resolve from `Object` sample identifier.
- Introduce `AssayService.getPropertyInputLineageRole()` to centralize logic for "role" processing of assay property inputs
- Move `getLookupSampleType` and `isLookupToMaterials` utilities to `ExperimentService` from `DefaultAssayRunCreator`.
- Consolidate `ExperimentService.createDataClass` interface.

PR #5874, opened 2024-09-19
Issue 51316: disable update of assay run/result sample lineage

labkey-nicka:
#### Rationale
This disables `AssayProvider.updatePropertyLineage()` processing in 24.9 and we'll revisit the implementation approach in a future release cycle. See [Issue 51316](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51316).

#### Related Pull Requests
- #5755

#### Changes
- ~~In addition to the `row` data pass in the `oldRow` data to `AssayProvider.updatePropertyLineage()` when updating assay run/result rows.~~
- ~~Inspect the old row value for the given lookup and only remove a sample as a material input if it was the original value.~~
- Disable `AssayProvider.updatePropertyLineage()` processing and `@Ignore` associated regression test coverage.

### DIFF
--- a/api/src/org/labkey/api/assay/AbstractAssayProvider.java
+++ b/api/src/org/labkey/api/assay/AbstractAssayProvider.java
@@ -2111,6 +2111,9 @@ public abstract class AbstractAssayProvider implements AssayProvider
         }
     }
 
+    // Issue 51316: While this issue is being addressed this feature has been disabled.
+    private final static boolean ENABLE_UPDATE_PROPERTY_LINEAGE = false;
+
     @Override
     public void updatePropertyLineage(
         Container container,
@@ -2124,6 +2127,9 @@ public abstract class AbstractAssayProvider implements AssayProvider
         @NotNull Map<Integer, ExpMaterial> materialsCache
     ) throws ValidationException
     {
+        if (!ENABLE_UPDATE_PROPERTY_LINEAGE)
+            return;
+
         Domain domain = table.getDomain();
         if (domain == null || newRow == null || oldRow == null)
             return;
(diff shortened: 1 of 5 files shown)


labkey-nicka:
I've kicked off a run against the LKSM suite for 24.9 in CI.

api/src/org/labkey/api/assay/AbstractAssayProvider.java
```diff
@@ -2147,14 +2153,28 @@ public void updatePropertyLineage(
...
+                            // Issue 51316: Resolve the original material input rowId for this column
+                            Integer originalRowId = null;
+                            if (oldValue instanceof Integer _originalRowId)
+                                originalRowId = _originalRowId;
+                            else if (oldValue != null)
+                            {
+                                ExpMaterial originalMaterial = ExperimentService.get().findExpMaterial(container, user, oldValue, sampleType, cache, materialsCache);
+                                if (originalMaterial != null)
+                                    originalRowId = originalMaterial.getRowId();
+                            }
```
cnathe, line 2176: this currently would happened each iteration through the run material inputs. could this instead be moved above the for loop so that it only has to happen once?

cnathe (changes requested):

labkey-nicka:
@cnathe Alright, I've disabled the update functionality, updated the associated regression test to cover further edge cases, and disabled the associated regression test.

api/src/org/labkey/api/assay/AbstractAssayProvider.java
labkey-nicka: As discussed, we're looking to take a different approach to this problem so this implementation is effectively defunct.

cnathe (approved):

PR #5883, opened 2024-09-23
Rename Project to Folder in apps

labkey-susanh:
#### Rationale
Using Project in the apps to label LKS Folders causes confusion internally and the word "Folder" resonates with clients, so we are switching our terminology in the applications. This brings about changes in class names, variable name, constants, etc. Most cases changed Project to Folder, but I've left Project when we are clearly referencing the LKS project and changed to container (sometimes) when it can be either a LKS Project or LKS Folder. Names of experimental feature flags have not been updated and some metric names have also intentionally not been updated.

#### Related Pull Requests
- 

#### Changes
-


api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
```diff
@@ -128,7 +128,6 @@ public class SampleTypeDomainKind extends AbstractDomainKind<SampleTypeDomainKin
         RESERVED_NAMES.add("RunId"); // Issue 50461
         RESERVED_NAMES.addAll(InventoryService.InventoryStatusColumn.names());
         RESERVED_NAMES.addAll(InventoryService.InventoryStatusColumn.labels());
-        RESERVED_NAMES.add("Project"); // Label for "Folder", remove when "Project" is renamed back to "Folder"
```
cnathe, line 131: I don't see them in this list but I assume that Container and Folder are just always considered reserved?

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -10912,6 +10912,7 @@ protected TYPE getType()
     }
 
     @RequiresPermission(AdminPermission.class)
+    // TODO Should this be renamed UpdateContainerSettingsAction? There's nothing that explicity checks for this being a project vs. a folder
```
cnathe, line 10915: yeah, this probably would make sense to rename but we would likely need to at least alias it as the current name if it was changed. leaving it as-is for this PR seems fine

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1859,11 +1859,11 @@ public Map<String, Integer> moveSamples(Collection<? extends ExpMaterial> sample
...
                     for (ExpMaterial sample : typeSamples)
                     {
-                        SampleTimelineAuditEvent event = createAuditRecord(targetContainer, "Sample project was updated.", userComment, sample, null);
+                        SampleTimelineAuditEvent event = createAuditRecord(targetContainer, "Sample folder was updated.", userComment, sample, null);
                         Map<String, Object> oldRecordMap = new HashMap<>();
-                        oldRecordMap.put("Project", sourceContainer.getName());
+                        oldRecordMap.put("ContainerName", sourceContainer.getName());
                         Map<String, Object> newRecordMap = new HashMap<>();
-                        newRecordMap.put("Project", targetContainer.getName());
+                        newRecordMap.put("ContainerName", targetContainer.getName());
```
cnathe: I would have expected this to be "Folder" instead of "ContainerName", or maybe just "Container"

cnathe (approved):

api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
labkey-susanh: Yes, "Container" comes from `BASE_PROPERTIES` and "Folder" from `ExpSampleTypeTable.Column`

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: I chose not to use just "Container" so as not to confuse with the Container field on the audit table. It gets remapped to "Folder" within `SampleTimelineEvent`, so, yeah, not sure why I didn't just use "Folder" here.

labkey-susanh: Ah, now I remember. It's because, by default "Folder" (the field in the SampleType table) is filtered out from the timeline events, so we use this other name to bypass that in the case of move, but remap it to be labeled as "Folder". I'll add a comment.

PR #5888, opened 2024-09-26
Issue 50774: Track processes a job launches, kill them when job is canceled

labkey-jeckels:
#### Rationale
Users want to be able to cancel pipeline jobs and have them kill any spawned external processes, and terminate the job immediately.

#### Changes
- Track the processes launched by jobs
- Kill the processes when a user cancels a job
- Minor code cleanup

### DIFF
--- a/api/src/org/labkey/api/pipeline/PipelineJob.java
+++ b/api/src/org/labkey/api/pipeline/PipelineJob.java
@@ -112,7 +112,7 @@ abstract public class PipelineJob extends Job implements Serializable
     // only write ERROR level events to the system log
     private static final Logger _logJobStopStart = LogManager.getLogger(Job.class);
 
-    public static Logger getJobLogger(Class clazz)
+    public static Logger getJobLogger(Class<?> clazz)
     {
         return LogManager.getLogger(PipelineJob.class.getName() + ".." + clazz.getName());
     }
@@ -420,7 +420,7 @@ abstract public class PipelineJob extends Job implements Serializable
         return updateStatusForTask();
     }
 
-    public TaskFactory getActiveTaskFactory()
+    public TaskFactory<?> getActiveTaskFactory()
     {
         if (getActiveTaskId() == null)
             return null;
@@ -443,7 +443,7 @@ abstract public class PipelineJob extends Job implements Serializable
     public void setLogFile(Path logFile)
     {
         // Set Log file path and clear/reset logger
-        _logFile = logFile.toAbsolutePath().normalize();;
+        _logFile = logFile.toAbsolutePath().normalize();
         _logger = null; //This should trigger getting the new Logger next time getLogger is called
     }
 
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/reports/ExternalScriptEngine.java
```diff
@@ -318,6 +321,9 @@ else if (value.startsWith("\""))
      */
     protected int runProcess(ScriptContext context, ProcessBuilder pb, StringBuffer output, long timeout, TimeUnit timeoutUnit)
     {
+        String jobGuid = (String)context.getBindings(ScriptContext.ENGINE_SCOPE).get(PIPELINE_JOB_GUID);
+        AutoCloseable canceler = null;
```
labkey-nicka: `canceler` is never used.

api/src/org/labkey/api/reports/ExternalScriptEngine.java
```diff
@@ -337,81 +343,84 @@ protected int runProcess(ScriptContext context, ProcessBuilder pb, StringBuffer
...
-            }
 
-            int code = proc.exitValue();
+                int code = proc.exitValue();
 
-            appendConsoleOutput(context, output);
+                appendConsoleOutput(context, output);
 
-            int count = out.get();
+                int count = out.get();
```
labkey-nicka: nit: (Not new code) `count` is not used but I guess theoretically `out.get()` could be side-effecting? Unclear if this line can be removed all together.

api/src/org/labkey/api/pipeline/PipelineJob.java
```diff
@@ -1331,87 +1337,86 @@ public void runSubProcess(ProcessBuilder pb, File dirWork, File outputFile, int
...
+                        proc.waitFor();
                     }
+
+                    int result = proc.exitValue();
+                    if (result != 0)
+                    {
+                        throw new ToolExecutionException("Failed running " + pb.command().get(0) + ", exit code " + result, result);
+                    }
+
+                    int count = output.get();
```
labkey-nicka, line 1397: nit: (Not new code) Slightly different incantation of what I pointed out in `ExternalScriptEngine`. This statement could theoretically be moved inside the subsequent `if (fileWriter != null)` so long as `out.get()` does not have side-effects.

labkey-nicka (approved):
Very cool.

api/src/org/labkey/api/reports/ExternalScriptEngine.java
labkey-jeckels: Removed.

api/src/org/labkey/api/reports/ExternalScriptEngine.java
labkey-jeckels: I simplified a bit but retained the `get()` call. It ensures that `Callable` linked to the `Future` is complete, and propagates exceptions that might have happened during the `Callable`.

PR #5889, opened 2024-09-26
AbDisc: Allow Type, Group, and Sample ID Column Inclusion/Exclusion From Grid

RosalineP:
#### Rationale
See story [here](https://github.com/LabKey/limsModules/issues/725).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5889
- https://github.com/LabKey/limsModules/pull/758
- https://github.com/LabKey/labkey-ui-premium/pull/546

#### Changes
- Upgrade script
- Add constructor(s) for PlateCustomField
- Add trigger EnsureSampleWellType
- Update removeFields(), addFields(), setFields()
- Updates for exports
- Update tests (more to come)

**ToDo:**
- [x] Write new cases in PlateManagerTest
- [x] Update file export selenium test constants
- [x] Investigate sqlserver-specific issue


assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3249,8 +3331,11 @@ private List<DisplayColumn> getPlateDisplayColumns(QueryView queryView)
         // columns from the default view of the Well table, which could be anything.
         DataRegion dataRegion = queryView.createDataView().getDataRegion();
 
-        // Filter on isQueryColumn so we don't get the details or update columns
-        return dataRegion.getDisplayColumns().stream().filter(DisplayColumn::isQueryColumn).toList();
+        // Filter on isQueryColumn, so we don't get the details or update columns
+        return dataRegion.getDisplayColumns().stream()
+                .filter(DisplayColumn::isQueryColumn)
+                .filter(col -> !col.getName().equals("sampleID")) // PR flag / comment flag
```
RosalineP: While we now never want the sampleID column returned from this function, based on the way it's used, this is not very nice. At worst I will add an explanatory comment, but hopefully someone has a better idea on what to do

RosalineP:
ToDo:
- [ ] Write new cases in PlateManagerTest
- [ ] Update file export selenium test constants
- [ ] Investigate sqlserver-specific issue

assay/src/org/labkey/assay/plate/data/PlateMapExcelWriter.java
```diff
@@ -184,27 +187,39 @@ protected void renderGrid(Sheet sheet, List<DisplayColumn> displayColumns) throw
...
 
+    // Removes fields explicitly excluded for Map export
+    protected List<DisplayColumn> getDisplayColumns()
+    {
+        return _displayColumns.stream().filter(col -> !excludedFields.contains(col.getName().toLowerCase())).toList();
+    }
+
+    protected List<PlateCustomField> getCustomFields()
+    {
+        return _plate.getCustomFields().stream().filter(field -> !new ArrayList<>(excludedFields).contains(field.getFieldKey() != null ? field.getFieldKey().toLowerCase() : null)).toList();
```
RosalineP: A bit clunky here...

RosalineP:
@labkey-nicka 
Ended up feeling as though the constants enum I wanted was not super needed. Let me know if you disagree after reviewing, though, and I can add it back in

assay/api-src/org/labkey/api/assay/plate/PlateCustomField.java
```diff
@@ -14,13 +16,26 @@ public class PlateCustomField
...
 
     public PlateCustomField()
     {
     }
 
+    public PlateCustomField(FieldKey fieldKey)
+    {
+        _fieldKey = fieldKey.getName();
+        _name = fieldKey.getName();
+    }
```
RosalineP: I would like to be able to get rid of this constructor. Given its usage though I am a little puzzled as to how.

assay/resources/schemas/dbscripts/postgresql/assay-24.012-24.013.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE assay.platesetproperty
+    ADD COLUMN FieldKey VARCHAR(256);
```
labkey-nicka: If you want the string length to be contained within a byte then declare it as `VARCHAR(255)`.

> With a maximum length of 255 characters, the DBMS can choose to use a single byte to indicate the length of the data in the field. If the limit were 256 or greater, two bytes would be needed.

https://stackoverflow.com/a/2340662

assay/src/org/labkey/assay/AssayUpgradeCode.java
```diff
@@ -674,6 +681,65 @@ public static void updateBarcodeSequence(ModuleContext ctx) throws Exception
...
+            SQLFragment sqlFragment = new SQLFragment("SELECT DISTINCT plateset FROM assay.Plate WHERE assaytype = '" + TsvPlateLayoutHandler.TYPE + "'");
+            ArrayList<Integer> plateSetIds = new SqlSelector(AssayDbSchema.getInstance().getSchema(), sqlFragment).getArrayList(Integer.class);
+
+            Set<Integer> assayPSes = new HashSet<>();
+            Set<Integer> primaryPSes = new HashSet<>();
+            Set<Integer> templatePSes = new HashSet<>();
+
+            for (Integer plateSetId : plateSetIds)
+            {
+                PlateSet PS = PlateService.get().getPlateSet(ContainerFilter.EVERYTHING, plateSetId);
```
labkey-nicka: nit: Perhaps `plateSet` rather than `PS`?

assay/src/org/labkey/assay/AssayUpgradeCode.java
```diff
@@ -674,6 +681,65 @@ public static void updateBarcodeSequence(ModuleContext ctx) throws Exception
...
+                if (PS == null)
+                    throw new IllegalStateException("updateBuiltInColumns: Plate Set with plate of id " + plateSetId + " not found.");
+
+                if (PS.isTemplate())
```
labkey-nicka: I think we should try and use SQL here rather than asking the model objects. The underlying implementation/meaning of these functions could change over time and cause this script to subsequently fail.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -304,7 +302,14 @@ public List<WellGroup.Type> getWellGroupTypes()
             if (plate == null)
                 throw new IllegalStateException("Unexpected failure. Failed to retrieve plate after save (pre-commit).");
 
-            deriveCustomFieldsFromWellData(container, user, plate, data);
+            List<PlateCustomField> builtInFields = getPlateBuiltInFields(container, user);
+            if (plateSetId == null)
+                deriveCustomFieldsFromWellData(container, user, plate, data, builtInFields.subList(1, 3));
```
RosalineP: ToDo: Add comments explaining all this subList business

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2236,39 +2312,68 @@ public Container getPlateMetadataDomainContainer(Container container)
...
+                else
+                    fields.add(new PlateCustomField(propertyURI));
+            }
+        });
+
+        // For now, we have a static sorting of the columns
+        return sortCustomFields(fields);
+    }
+
+    private List<PlateCustomField> sortCustomFields(List<PlateCustomField> fields)
```
labkey-nicka, line 2345: Having the server declare this sorting should be sufficient enough that we do not need to have this replicated on the client in `PlateEditor`.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -2236,39 +2312,68 @@ public Container getPlateMetadataDomainContainer(Container container)
...
+    {
+        Map<FieldKey, Integer> order = new HashMap<>();
+        order.put(WellTable.Column.Type.fieldKey(), 0);
```
labkey-nicka, line 2348: Note, I've retained the T/G/S column ordering as outlined in the specification.

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -324,39 +327,69 @@ public List<WellGroup.Type> getWellGroupTypes()
         }
     }
 
+    private List<PlateCustomField> getDefaultFieldsForPlateSet(@NotNull Plate plate, @Nullable PlateSet plateSet)
```
labkey-nicka, line 330: I've updated how we compute the default set of columns for a given plate (set) type. This is in lieu of using the `subList()` handling that was being done previously.

assay/src/org/labkey/assay/plate/data/WellTriggerFactory.java
```diff
@@ -31,11 +31,54 @@ public class WellTriggerFactory implements TriggerFactory
     public @NotNull Collection<Trigger> createTrigger(@Nullable Container c, TableInfo table, Map<String, Object> extraContext)
     {
         return List.of(
+            new EnsureSampleWellTypeTrigger(),
             new ValidatePrimaryPlateSetUniqueSamplesTrigger(),
             new ComputeWellGroupsTrigger()
         );
     }
 
+    protected class EnsureSampleWellTypeTrigger implements Trigger
```
labkey-nicka, line 43: I still need to review this trigger more thoroughly.

labkey-nicka (commented):
I've incorporated the feedback we discussed and added additional comments in this thread regarding ground we did not cover. Review and we can sync up tomorrow.

assay/src/org/labkey/assay/plate/PlateManager.java
RosalineP: I'm seeing that with the current client-side code, the cols order in: Type, Group, Sample ID, <metadata> whereas without the client side code for newly created PSes we get: Sample ID, Group, <metadata>, Type

labkey-nicka: The reason is we're losing the sorting when iterating over the `queryInfo.columns`. Instead, if we use:

```tsx
constructor(queryInfo: QueryInfo, fieldMap: ExtendedMap<string, PlateField>, containerPath: string) {
    this.columns = [queryInfo.getColumn(WellTableColumn.Position)];

    for (const [name] of fieldMap) {
        const col = queryInfo.getColumn(name);
        if (col) {
            this.columns.push(applyLookupPath(col, containerPath));
        }
    }

    this.mode = EditorMode.Update;
    this.queryInfo = queryInfo;
}
```

it respects the sort order of the columns as supplied by the `Plate`.

assay/src/org/labkey/assay/plate/data/WellTriggerFactory.java
```diff
@@ -31,11 +31,54 @@ public class WellTriggerFactory implements TriggerFactory
...
+    {
+        // When no type is given but sampleid is populated, provide 'Sample' as the type
+        private void addTypeSample(@Nullable Map<String, Object> newRow)
```
labkey-nicka: Bug. This steps on pre-existing `Type` information when the well is being updated and the call does not explicitly include the `Type` column. Repro:

1. Configure a well on a plate to be of type replicate. Specify a well group and a sample ID.
2. Save it.
3. Edit the Sample ID to be some other sample. Do not change any other values.
4. Save it.

Expected: The well type remains "Replicate".
Actual: The well type is changed to "Sample".

The non-trivial part of fixing this is we need to know the pre-existing type of the well. This information will not ever be furnished by `oldRow` since that column is an expression column. As such, we'll need to dynamically query to find out. I have a change set that fixes this.

assay/src/org/labkey/assay/plate/data/WellTriggerFactory.java
```diff
@@ -31,11 +31,54 @@ public class WellTriggerFactory implements TriggerFactory
...
+                    newRow != null &&
+                            newRow.containsKey(WellTable.Column.SampleId.name()) &&
+                            newRow.getOrDefault(WellTable.Column.SampleId.name(), null) != null &&
+                            !newRow.containsKey(WellTable.Column.Type.name())
+            )
+                newRow.put(WellTable.Column.Type.name(), WellGroup.Type.SAMPLE.name());
+        }
+
+        @Override
+        public void afterInsert(
```
labkey-nicka: I think we should use `beforeInsert` and `beforeUpdate` in this trigger.

labkey-nicka (approved):

PR #5892, opened 2024-09-27
Issue 50131: Kill DB queries when pipeline job is canceled

labkey-jeckels:
#### Rationale
A pipeline job may be stuck running a slow query. Users want to be able to cancel the job and abort the query to free up the pipeline queue.

#### Changes
- Track the thread running the job
- Kill the DB connections when the job is canceled
- Only supported on Postgres since SQL Server driver doesn't actually kill the queries when the connection is closed


api/src/org/labkey/api/data/ConnectionWrapper.java
```diff
@@ -176,6 +178,31 @@ public static boolean dumpOpenConnections(@NotNull LoggerWriter logWriter)
         return true;
     }
 
+    public static void closeConnections(Thread thread)
+    {
+        List<ConnectionWrapper> toClose = new ArrayList<>();
+
+        synchronized (_openConnections)
+        {
+            for (Map.Entry<ConnectionWrapper, Pair<Thread, Throwable>> entry : _openConnections.entrySet())
```
labkey-adam: Consider using a record... could be Map<ConnectionWrapper, ConnectionRecord(ConnectionWrapper, Thread, Throwable)>, since this and other callers just want to iterate the map's values.

labkey-adam:
@labkey-jeckels note the connection-related test failures

Unless there's urgency here, you may want to merge after 24.10 branching to give time to shake out any issues.

labkey-adam (approved):

api/src/org/labkey/api/data/ConnectionWrapper.java
labkey-jeckels: Good point. I ended up just folding them into `ConnectionWrapper` itself. No need for separate classes or objects.

PR #5897, opened 2024-09-30
LIMS: Charting Story #3 - audit logs, cross folder & rendering improvements

cnathe:
#### Rationale
This PR adds audit events for insert/update/create of charts/reports, allows for Generic Chart reports to be inherited to child folders, and improves some rendering issues. 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1597

#### Changes
- Charting updates to default tickOverlapRotation and bottom margin for long/overlapping tick labels
- Allow for Generic Charts to be inherited in child folders
- Save chart inherit checkbox to be hidden/shown based on user perm (to match R Report inherit checkbox perm check)
- Add ReportAuditProvider with events for insert, update and delete of reports and charts


query/src/org/labkey/query/reports/ReportServiceImpl.java
```diff
@@ -432,14 +436,17 @@ private ReportDB _saveDbReport(User user, Container c, String key, ReportDescrip
...
-        if (null != reportId && reportExists(reportId.getRowId()))
+        boolean reportExists = null != reportId && reportExists(reportId.getRowId());
+        if (reportExists)
             reportDB = Table.update(user, getTable(), reportDB, reportId.getRowId());
         else
             reportDB = Table.insert(user, getTable(), reportDB);
 
         DatabaseReportCache.uncache(c);
 
+        ReportAuditProvider.ReportAuditEvent event = new ReportAuditProvider.ReportAuditEvent(reportDB, descriptor, c, reportExists ? "Report updated" : "Report created");
```
XingY, line 447: Are we planning to audit for all report types, including, for example "SampleManager.SampleFinder" reports? (this comes free but just want to confirm)

XingY: Also, what about hidden report, or user's private report?

query/src/org/labkey/query/reports/ReportAuditProvider.java
```diff
@@ -0,0 +1,195 @@
...
+        public static String NAMESPACE_PREFIX = "Report-" + NAME;
+
+        private final Set<PropertyDescriptor> fields;
+
+        public ReportAuditDomainKind()
+        {
+            super(EVENT_NAME);
+
+            fields = new LinkedHashSet<>();
+            fields.add(createPropertyDescriptor(COLUMN_NAME_REPORT_ID, PropertyType.STRING, "Report Id", null, true));
```
XingY: Isn't rowId integer type?

query/src/org/labkey/query/reports/ReportServiceImpl.java
cnathe: Re: SampleFinder saved searches/reports - I guess I wasn't intending on auditing those saves but it seems like a fine side effect. I think auditing them might be nice unless you see a reason we shouldn't?

Re: hidden and private reports - yes, this will audit those as well. Since we are just tracking the report id, name, reportKey, and type this seems fine to me. Can you think of a reason we shouldn't audit these?

query/src/org/labkey/query/reports/ReportAuditProvider.java
cnathe: Yes, good catch. I've changed it. 

@XingY did you pull down this branch and start your server locally? If so, you'll have to do some manual database deletions to remove the domain so that it can change the reportId prop type. Let me know and I can give you my SQL script I used for the deletes locally.

XingY (approved):

PR #5898, opened 2024-10-01
Fix java.nio.channels.OverlappingFileLockException

XingY:
#### Rationale
[Exception](https://www.labkey.org/_mothership/mothership-showStackTraceDetail.view?exceptionStackTraceId=38415) report.

#### Related Pull Requests
- 

#### Changes
- call FileChannel.tryLock instead of lock to handle the case when a file was already locked by JVM.

### DIFF
--- a/api/src/org/labkey/api/util/FileUtil.java
+++ b/api/src/org/labkey/api/util/FileUtil.java
@@ -941,7 +941,7 @@ public class FileUtil
 
         try (FileOutputStream os = new FileOutputStream(dst);
              FileChannel out = os.getChannel();
-             FileLock lockOut = out.lock())
+             FileLock lockOut = out.tryLock())
         {
             do
             {


labkey-jeckels:
@labkey-matthewb can you review this? I touched that line last but I believe you introduced that `lock()` call many years ago. My concern is that this change will avoid the exception but potentially copy a partial or mangled file.

XingY:
> @labkey-matthewb can you review this? I touched that line last but I believe you introduced that `lock()` call many years ago. My concern is that this change will avoid the exception but potentially copy a partial or mangled file.



> @labkey-matthewb can you review this? I touched that line last but I believe you introduced that `lock()` call many years ago. My concern is that this change will avoid the exception but potentially copy a partial or mangled file.

Here is the doc for `tryLock`. From the description, it returns null if the file is already locked, but would still throw exception due to other reasons.

`Attempts to acquire a lock on the given region of this channel's file.
This method does not block. An invocation always returns immediately, either having acquired a lock on the requested region or having failed to do so. If it fails to acquire a lock because an overlapping lock is held by another program then it returns null. If it fails to acquire a lock for any other reason then an appropriate exception is thrown.`

labkey-matthewb:
Doesn't this mean that there is another thread that is trying to write to the same file?  That seems bad.  Maybe we should add error logging to see why we have two threads writing to the same file.  Or just have the caller handle this exception and present a message like "file is in use".  It doesn't seem right to continue.

I say this knowing that the majority of cases in the code we are not so careful and wouldn't notice that we have conflicts.

PR #5903, opened 2024-10-01
Sync Assay Sample Lookup Lineage

labkey-nicka:
#### Rationale
This reintroduces support for updating experiment material inputs when sample lookup values change in assay domains.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5903
- https://github.com/LabKey/limsModules/pull/779

#### Addresses Issues
- [Issue 45594](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45594): Inconsistent Data Displayed After Updating Assay Runs & Results
- [Issue 51126](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51126): Remapping assay results to a new sample disallows delete of old sample
- [Issue 51316](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51316): Updates of assay run/result row sample lookups incorrectly clear lineage from other rows

#### Changes

**Assay**
- Introduce `AssaySampleLookupContext` class to handle syncing of the material inputs for a collection of experiment runs.
- Resolve `SampleLookup` fields from assay domains and dynamically query for all current material values across these fields.
- Persist material input changes for a run when necessary.
- Wire up usage of `AssaySampleLookupContext` in updating of assay run and result domains.
- Expose a new `assay-syncRunLineage.api` endpoint to allow for users to manually sync a collection of runs.
- Refactor all "is this a sample lookup column" checks to use `AssaySampleLookupContext.checkSampleLookup`.
- Enable integration test case.

**Experiment**
- Introduce `ExpProtocolApplication.addMaterialInputs` which supports batch insertion of material inputs for higher throughput than pre-existing `addMaterialInput`.
- Introduce `ExpProtocolApplication.removeAllDataInputs` and `.removeAllMaterialInputs`.
- Rename `ExperimentListener.beforeRunCreated` to `ExperimentListener.beforeRunSaved` as this is more accurate and aligns with `afterRunSaved`.


api/src/org/labkey/api/assay/AbstractAssayProvider.java
```diff
@@ -17,6 +17,7 @@
 package org.labkey.api.assay;
 
 import jakarta.servlet.http.HttpServletRequest;
+import org.apache.commons.collections4.map.LRUMap;
```
labkey-klum: interesting, but it doesn't appear we are using this class.

labkey-nicka: I'll get these imports removed.

api/src/org/labkey/api/assay/AbstractAssayProvider.java
```diff
@@ -1202,8 +1206,13 @@ public static void addInputMaterials(ExpRun expRun, User user, Map<ExpMaterial,
                         ExpSampleType st = newInput.getSampleType();
                         role = st != null ? st.getName() : "Sample";
                     }
-                    protApp.addMaterialInput(user, newInput, role);
+
+                    inputGroups.putIfAbsent(role, new HashSet<>());
+                    inputGroups.get(role).add(newInput.getRowId());
```
labkey-jeckels: `computeIfAbsent()` is a little more convenient and doesn't create HashSets that are ignored. Untested code:

```suggestion
                    inputGroups.computeIfAbsent(role, r -> new HashSet<>()).add(newInput.getRowId());
```

api/src/org/labkey/api/assay/sample/AssaySampleLookupContext.java
```diff
@@ -0,0 +1,406 @@
...
+
+import static java.util.Collections.emptySet;
+
+public class AssaySampleLookupContext
+{
+    private final Set<Integer> _runIds;
+    private final Map<FieldKey, Boolean> _sampleLookups;
+
+    /** Structure representing a material and its designated role. */
+    public record MaterialInput(Integer materialRowId, String role) {}
```
labkey-jeckels: No need to support nulls - consider using `int` instead of `Integer`

api/src/org/labkey/api/assay/sample/AssaySampleLookupContext.java
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    public void syncLineage(Container container, User user, BatchValidationException errors)
+    {
+        if (_runIds.isEmpty() || errors.hasErrors())
+            return;
+
+        // Perform the sync using the service user to ensure that all pre-existing
+        // material inputs are retained regardless of the current user's permissions.
+        var serviceUser = User.getAdminServiceUser();
```
labkey-jeckels, line 180: Why is this needed? When a run (in a container where the user has update permission) has a lookup to a sample in a container where the user doesn't have access?

api/src/org/labkey/api/assay/sample/AssaySampleLookupContext.java
```diff
@@ -0,0 +1,406 @@
...
+            var assayProvider = AssayService.get().getProvider(run);
+            if (assayProvider == null)
+            {
+                errors.addRowError(new ValidationException("Failed to resolve assay provider for run with rowId " + expRunRowId));
+                return;
+            }
+
+            var sampleLookups = getSampleLookups(container, serviceUser, run, assayProvider);
+
+            // CONSIDER: Do not short-circuit if there are no sample lookups. A domain could have been modified
```
labkey-jeckels, line 201: Do we run this code when a domain is edited to drop a sample lookup?

api/src/org/labkey/api/assay/sample/AssaySampleLookupContext.java
```diff
@@ -0,0 +1,406 @@
...
+        sortedInputs.sort(new MaterialInputRoleComparator(inputLineageRoles));
+
+        for (var input : inputs)
+        {
+            // A sample can be marked as a material input at most once
+            if (seen.contains(input.materialRowId))
+                continue;
+
+            seen.add(input.materialRowId);
+            groups.putIfAbsent(input.role, new HashSet<>());
```
labkey-jeckels: Consider `computeIfAbsent()` here too

api/src/org/labkey/api/assay/sample/AssaySampleLookupContext.java
```diff
@@ -0,0 +1,406 @@
...
+                newInputs.removeIf(input -> current.materialRowId.equals(input.materialRowId));
+            }
+
+            newInputs.add(current);
+        }
+
+        return Pair.of(currentInputs.equals(newInputs), newInputs);
+    }
+
+    private Set<MaterialInput> getCurrentMaterialInputs(Integer expRunRowId)
```
labkey-jeckels:
```suggestion
    private Set<MaterialInput> getCurrentMaterialInputs(int expRunRowId)
```

api/src/org/labkey/api/assay/sample/AssaySampleLookupContext.java
```diff
@@ -0,0 +1,406 @@
...
+            SELECT MI.MaterialId AS MaterialRowId, MI.Role AS MaterialInputRole
+            FROM exp.MaterialInput MI
+            INNER JOIN exp.ProtocolApplication PA ON MI.TargetApplicationId = PA.RowId
+            WHERE PA.RunId = ?
+        """).add(expRunRowId);
+
+        return getMaterialInputs(new SqlSelector(ExperimentService.get().getSchema(), sql));
+    }
+
+    private Set<MaterialInput> getNewMaterialInputs(Map<TableInfoKey, List<SampleLookup>> sampleLookups, Integer expRunRowId)
```
labkey-jeckels:
```suggestion
    private Set<MaterialInput> getNewMaterialInputs(Map<TableInfoKey, List<SampleLookup>> sampleLookups, int expRunRowId)
```

api/src/org/labkey/api/assay/sample/AssaySampleLookupContext.java
```diff
@@ -0,0 +1,406 @@
...
+                .columns(Set.of(tableInfoKey.keyColumn, column))
+                .filter(tableFilter)
+                .buildSqlFragment();
+
+        var role = AssayService.get().getPropertyInputLineageRole(sampleLookup.domainProperty);
+        var sql = new SQLFragment();
+
+        if (column.getJdbcType().isInteger())
+        {
+            sql.append("SELECT DA.").append(columnName).append(" AS MaterialRowId");
```
labkey-jeckels, line 390: @labkey-matthewb `columnName` should be quoted already (if needed) from the call to `getColumnSelectName()`. Should it be added via `appendIdentifier()` instead?

experiment/src/org/labkey/experiment/api/ExpRunImpl.java
```diff
@@ -468,6 +467,12 @@ public List<ExpProtocolApplicationImpl> getProtocolApplications()
         return _protocolSteps;
     }
 
+    @Override
+    public @Nullable ExpProtocolApplication getCoreProtocolApplication()
```
labkey-jeckels: This isn't relevant for non-"simple" runs. Maybe not important enough to check first, but perhaps some JavaDoc on the interface method would be useful.

experiment/src/org/labkey/experiment/api/ExpProtocolApplicationImpl.java
```diff
@@ -469,11 +494,63 @@ public ExpMaterialRunInputImpl addMaterialInput(User user, ExpMaterial material,
...
+
+        if (StringUtils.trimToNull(role) == null)
+            role = ExpMaterialRunInput.DEFAULT_ROLE;
+
+        for (Integer materialRowId : materialRowIds)
+            params.add(Arrays.asList(materialRowId, getRowId(), role, protocolInputRowId));
+
+        String sql = "INSERT INTO " + ExperimentServiceImpl.get().getTinfoMaterialInput().getSelectName() +
+                " (MaterialId, TargetApplicationId, Role, ProtocolInputId)" +
+                " VALUES (?,?, CAST(? AS VARCHAR), CAST(? AS INTEGER))";
```
labkey-klum: should we use : `getSqlDialect().getSqlCastTypeName`?

experiment/src/org/labkey/experiment/api/ExpProtocolApplicationImpl.java
```diff
@@ -354,44 +359,62 @@ public void save(User user)
...
+        if (getRowId() == 0)
+            return 0;
 
-            Table.delete(ExperimentServiceImpl.get().getTinfoProtocolApplication(), getRowId());
-        }
+        final ExperimentServiceImpl svc = ExperimentServiceImpl.get();
+        final SqlDialect dialect = svc.getSchema().getSqlDialect();
+
+        OntologyManager.deleteOntologyObjects(svc.getSchema(), new SQLFragment("SELECT " +
+            dialect.concatenate("'" + DataInput.lsidPrefix() + "'",
```
labkey-jeckels, line 398: Presumably this needs to be kept in sync with other code that's generating URIs. Can you add a pointer in both spots?

labkey-klum (approved):

assay/src/org/labkey/assay/AssayExperimentListener.java
```diff
@@ -34,4 +35,10 @@ public void afterRunSaved(Container container, User user, ExpProtocol protocol,
     {
         AssayManager.get().indexAssayRun(run.getRowId());
     }
+
+    @Override
+    public void beforeRunSaved(Container container, User user, ExpProtocol protocol, ExpRun run) throws BatchValidationException
```
labkey-jeckels: Is this needed? Can we rely on the default implementation from the interface?

labkey-jeckels (approved):

api/src/org/labkey/api/assay/sample/AssaySampleLookupContext.java
labkey-nicka: No, but someone could call the endpoint with any run from any assay type where that assay type domain may have been mutated.

api/src/org/labkey/api/assay/sample/AssaySampleLookupContext.java
labkey-nicka: I thought about this and looked into it. `dialect.getColumnSelectName()` calls `dialect.makeLegalIdentifier()` which does the quoting if necessary.

labkey-nicka: I see, you're saying that it does but do we also need to do via `appendIdentifier`. I hadn't seen usages like that but seems plausible.

api/src/org/labkey/api/assay/sample/AssaySampleLookupContext.java
labkey-nicka: Exactly.

experiment/src/org/labkey/experiment/api/ExpRunImpl.java
labkey-nicka: I've changed this to `getProtocolApplication()` and added documentation.

assay/src/org/labkey/assay/AssayExperimentListener.java
labkey-nicka: Removed.

cnathe:
@labkey-nicka  Issue repro:
1. setup LKB with a sample type and an assay design that has a Sample field on the Run domain
2. import an assay run and select a sample for the run prop
3. verify the run Material Inputs has the sample
4. edit assay run prop to remove the sample
Result: run Material Input still lists the sample. Note that changing the sample does update the Inputs as expected.

experiment/src/org/labkey/experiment/api/ExpProtocolApplicationImpl.java
labkey-nicka: There are 5 different places we generate SQL like this. I'm going to leave as-is for this PR and can do a minor refactor on develop to shore these up.

labkey-nicka:
> @labkey-nicka Issue repro:
> 
> 1. setup LKB with a sample type and an assay design that has a Sample field on the Run domain

Alright, I've addressed this and added regression coverage for this scenario with 1d6c2cf878.

cnathe:
> > @labkey-nicka Issue repro:
> > 
> > 1. setup LKB with a sample type and an assay design that has a Sample field on the Run domain
> 
> Alright, I've addressed this and added regression coverage for this scenario with [1d6c2cf](https://github.com/LabKey/platform/commit/1d6c2cf87848569c123a197d89653d3efcac3f39).

Verified. Looks good.

PR #5906, opened 2024-10-02
Plate replicate rollups and AbDisc assay re-import

labkey-klum:
#### Rationale
Introduces support for the calculation of standard deviation and mean values for replicate well groups for protocols that support plate metadata in LKB folders. Replicate aggregates are computed for numerical measures in the results domain and are automatically joined to the results views.

In support of LKB AB discovery scenarios (and potentially others), this also introduces the notion of the addition of assay run data over time through the existing assay run re-import feature. Plate set plate data can be added to a run over multiple imports. The server will attempt to combine / merge prior data with new data on a plate boundary within a single plate set during each subsequent import / re-import.

#### Changes
- Create a protocol schema domain : `AssayPlateReplicateStats` to store the replicate standard deviation and mean values. One row per replicate well group per run will be created.
- When the assay results domain changes, create fields in the `AssayPlateReplicateStats` domain to track any numeric measures.
- On insert, compute and save replicate stat values to the protocol schema table.
- Join the assay results table to the replicate stats table.
- `AssayPlateTriggerFactory` to handle updates and deletes on the results table. Depending on whether the updates occur on replicate well group rows, recompute and update replicate stats. Unfortunately, assays handle inserts and updates via different code paths, the former happens in 
- `AssayRunUploadContext.ReImportOption` `enum` to control the server side re-import behavior. The parameter can be passed into the `ImportRunApiAction` and defaults to the current behavior of `REPLACE`.
- `AssayPlateMetadataServiceImpl.mergeReRunData` handles the details of merging current and previous results data. If data is merged, the uploaded data is replaced with the `tsv` tabular representation of the combined data and the data output `ExpData` is updated to reflect the change. This is always the case regardless of whether previous or current data is in the graphical format.
- Hit selections are now cleared from the previous run regardless of whether this is a straight re-import or the `ABDisc` re-import where plate set data can potentially be merged.
- Hit selections are propagated in the merge case on a plate by plate basis if the new data is not replacing any existing plate data. The implementation was a little disjointed so here are the details:
  - Hit re-selection needs to be done after new assay results have been persisted because we need the result row IDs. This happens in `PlateMetadataImportHelper.afterBatchInsert`. 
  - We only know about the incoming versus existing plate data during the parse phase of the new data file. This happens in `AssayPlateMetadataServiceImpl.mergeReRunData`. We end up removing all hit selections that we do not plan to carry forward based on our plate merge policy. Any remaining hits are carried forward in the `ImportHelper`.

#### Debug Notes
- Until the client dev is in place, the assay run re-import behavior can be controlled by changing the default `reImportOption` in the `ImportRunApiForm` around `ImportRunApiAction:356`

#### implementation Details
Statistical values should appear automatically in the default view of the results table. Stats fields should be aligned close to their source.
image

After the re-import case if data is combined with a previous run
image

#### Related Pull Requests
https://github.com/LabKey/commonAssays/pull/813


api/src/org/labkey/api/assay/AssayRunUploadContext.java
```diff
@@ -47,6 +47,13 @@
 */
 public interface AssayRunUploadContext<ProviderType extends AssayProvider> extends ContainerUser, HasHttpRequest
 {
+    // options for how re-imported data is handled, default is REPLACE
+    enum ReImportOption
+    {
+        REPLACE,
+        MERGE_DATA
```
cnathe: do we have a comment in the code somewhere as to what "merge" means in this context (i.e. regarding the "replace existing plates but append new plates" behavior)? just thinking it might be good to put some comment / description here in the enum otherwise.

assay/src/org/labkey/assay/TSVProtocolSchema.java
```diff
@@ -34,33 +35,55 @@
...
     {
         super(user, container, provider, protocol, targetStudy);
     }
 
+    @Override
+    public Set<String> getTableNames()
+    {
+        Set<String> names = super.getTableNames();
+
+        if (getProvider().isPlateMetadataEnabled(getProtocol()))
```
cnathe: should this also have the `isExperimentalAppPlateEnabled` or is that not needed here?

assay/src/org/labkey/assay/TSVProtocolSchema.java
```diff
@@ -71,7 +94,15 @@ public FilteredTable createDataTable(ContainerFilter cf, boolean includeLinkedTo
     public TableInfo createProviderTable(String name, ContainerFilter cf)
     {
         if (EXCLUSION_REPORT_TABLE_NAME.equalsIgnoreCase(name))
+        {
             return createExclusionReportTable(cf);
+        }
+        else if (name.equalsIgnoreCase(PLATE_REPLICATE_STATS_TABLE))
+        {
+            TableInfo replicateTable = createPlateReplicateStatsTable(cf);
+            if (replicateTable != null)
```
cnathe: if replicateTable is null here, it looks like the call to `super.createProviderTable(name, cf)` will also end up returning null, so probably can just drop this if statement and always return replicateTable here.

assay/src/org/labkey/assay/TSVProtocolSchema.java
```diff
@@ -164,11 +198,99 @@ private class _AssayResultTable extends AssayResultTable
...
+                        Map<String, FieldKey> replicateFields = new HashMap<>();
+                        for (DomainProperty prop : replicateDomain.getProperties())
+                            replicateFields.put(prop.getName(), FieldKey.fromParts("Replicate", prop.getName()));
+
+                        List<FieldKey> newDefaultColumns = new ArrayList<>();
+                        for (FieldKey fk : defaultColumns)
+                        {
+                            newDefaultColumns.add(fk);
+                            ColumnInfo col = getColumn(fk);
+                            if (col != null && col.isMeasure() && col.getJdbcType().isNumeric())
```
cnathe, line 218: this check for "isMeasure and isNumeric" happens in a couple of places (maybe only 2 that I see?). I was thinking it would be good to share in a helper method, but I don't think that will work since this one is ColumnInfo objects and the other is DomainProperty.

assay/src/org/labkey/assay/TSVProtocolSchema.java
```diff
@@ -164,11 +198,99 @@ private class _AssayResultTable extends AssayResultTable
...
+        @Override
+        public boolean hasPermission(@NotNull UserPrincipal user, @NotNull Class<? extends Permission> perm)
+        {
+            return _userSchema.getContainer().hasPermission(user, perm);
+        }
+
+        @Override
+        public @Nullable QueryUpdateService getUpdateService()
+        {
+            return new DefaultQueryUpdateService(this, getRealTable());
```
cnathe, line 286: Do we really want this PlateReplicateStats table to be editable? I was thinking it would be read only. We don't want users to be able to insert/update/delete rows into this table directly, do we?

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -208,24 +245,185 @@ public DataIteratorBuilder parsePlateData(
...
+                    {
+                        // incoming plate data has precedence over any previous plate data.
+                        Map<String, Plate> plateMap = plates.stream().collect(Collectors.toMap(Plate::getName, p -> p));
+                        Set<String> existingPlates = new HashSet<>();
+                        Set<Integer> prevPlates = new HashSet<>();
+                        rows.forEach(r -> existingPlates.add(r.get(AssayResultDomainKind.PLATE_COLUMN_NAME).toString()));
+                        List<Map<String, Object>> newRows = new ArrayList<>(rows);
+
+                        // parse the previous data and combine with any new data
+                        for (Map<String, Object> row : _parsePlateData(container, user, prevData, provider, protocol, plateSet, plates, prevFile, settings))
```
cnathe: I wasn't expecting that we'd reparse the previous file here. For example, if I imported plate1 data for a plate set with plate1 and plate2 and then came back to re-import and attach plate2. Then this works. But if I had edited results in the plate1 data prior to re-import, those edits would be lost as we'd be going back to the original file here. Thoughts?

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -520,32 +720,261 @@ public OntologyManager.UpdateableTableImportHelper getImportHelper(
...
+                    if (run != null)
+                        replicateRow.put(PlateReplicateStatsDomainKind.Column.Run.name(), run.getRowId());
+
+                    for (Map.Entry<String, List<Double>> measure : measures.entrySet())
+                    {
+                        MathStat stat = StatsService.get().getStats(measure.getValue());
+                        if (!Double.isNaN(stat.getMean()))
+                            replicateRow.put(measure.getKey() + PlateReplicateStatsDomainKind.REPLICATE_MEAN_SUFFIX, stat.getMean());
+                        if (!Double.isNaN(stat.getStdDev()))
+                            replicateRow.put(measure.getKey() + PlateReplicateStatsDomainKind.REPLICATE_STD_DEV_SUFFIX, stat.getStdDev());
```
cnathe: there is a case on update where we would need to remove a mean / sd value instead of just updating it. For example, if you had a replicate well group with values for measures1 which resulted in a mean / sd value and then edit the assay results rows to blank out those values for measure1, we'd want the mean / sd to also be removed.

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -520,32 +720,261 @@ public OntologyManager.UpdateableTableImportHelper getImportHelper(
...
+        }
+    }
+
+    @Nullable
+    private QueryUpdateService getReplicateStatsUpdateService(Container container, User user, AssayProvider provider, ExpProtocol protocol)
+    {
+        AssayProtocolSchema schema = provider.createProtocolSchema(user, container, protocol, null);
+        TableInfo tableInfo = schema.createTable(TSVProtocolSchema.PLATE_REPLICATE_STATS_TABLE, null);
+        if (tableInfo != null)
+            return tableInfo.getUpdateService();
```
cnathe: see previous comment. should this just return the default QUS here instead of having it on the table itself?

assay/src/org/labkey/assay/plate/AssayPlateTriggerFactory.java
```diff
@@ -0,0 +1,127 @@
...
+                // recompute the stats for the changed replicate rows
+                AssayProvider provider = AssayService.get().getProvider(_protocol);
+                if (provider == null)
+                    throw new IllegalStateException(String.format("Unable to find the provider for protocol : %s", _protocol.getName()));
+
+                AssayProtocolSchema schema = provider.createProtocolSchema(user, c, _protocol, null);
+                TableInfo dataTable = schema.createDataTable(null, false);
+                if (dataTable != null)
+                {
+                    try (DbScope.Transaction tx = ExperimentService.get().ensureTransaction())
```
cnathe: I'm surprised we'd have a new transaction here. isn't this trigger already happening in a transaction? or do we need a new transaction here because this is on complete?

cnathe (changes requested):

labkey-chrisj:
One thing we may want to do is apply a standard format for decimal/float-based columns; with the default format we end up not rounding numbers we probably ought to round in our mean/stddev calculation column: 
image

api/src/org/labkey/api/assay/AssayRunUploadContext.java
labkey-klum: I added some comments to this enum, the various options were meant to be interpreted by the data handlers but I described the plate based behavior.

assay/src/org/labkey/assay/TSVProtocolSchema.java
labkey-klum: no longer needed, this was leftover from when there were two variants of plate metadata

assay/src/org/labkey/assay/TSVProtocolSchema.java
labkey-klum: done

assay/src/org/labkey/assay/TSVProtocolSchema.java
labkey-klum: exactly 😢

assay/src/org/labkey/assay/plate/AssayPlateTriggerFactory.java
labkey-klum: that's true, removed.

assay/src/org/labkey/assay/TSVProtocolSchema.java
labkey-klum: The table is provisioned, so we want to use the update service to interact with the generated stats. I've removed the table from the visible names in the schema and added some code to prevent users from getting an updatable version of the table.

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
labkey-klum: agreed, fixed.

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
labkey-klum: After discussing with Cory and Nick, the plan is to not reparse the previous file, but to rely on the run results in the database for this re-import

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -208,24 +247,217 @@ public DataIteratorBuilder parsePlateData(
...
+            List<Map<String, Object>> newRows = new ArrayList<>();
+
+            // parse the existing run data and combine with any new data
+            AssayProtocolSchema schema = provider.createProtocolSchema(user, container, protocol, null);
+            TableInfo resultsTable = schema.createDataTable(null, false);
+            if (resultsTable != null)
+            {
+                // The plate value is either a string or integer on the incoming data, need to match that when merging
+                // existing data
+                Object plateObj = rows.get(0).get(AssayResultDomainKind.PLATE_COLUMN_NAME);
```
cnathe, line 359: it doesn't look like `plateObj` is used, I believe it can be removed.

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -208,24 +247,217 @@ public DataIteratorBuilder parsePlateData(
...
+                final FieldKey plateFieldKey;
+                //if (plateObj instanceof String)
+                    plateFieldKey = FieldKey.fromParts(AssayResultDomainKind.PLATE_COLUMN_NAME, "Name");
+                //else
+                //    plateFieldKey = FieldKey.fromParts(AssayResultDomainKind.PLATE_COLUMN_NAME);
```
cnathe: remove commented lines?

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -515,37 +749,275 @@ private Map<String, Object> getDataRowFromWell(String plateId, Well well, String
...
+                    if (run != null)
+                        replicateRow.put(PlateReplicateStatsDomainKind.Column.Run.name(), run.getRowId());
+
+                    for (Map.Entry<String, List<Double>> measure : measures.entrySet())
+                    {
+                        MathStat stat = StatsService.get().getStats(measure.getValue());
+                        if (!Double.isNaN(stat.getMean()))
+                            replicateRow.put(measure.getKey() + PlateReplicateStatsDomainKind.REPLICATE_MEAN_SUFFIX, stat.getMean());
+                        else
+                            replicateRow.put(measure.getKey() + PlateReplicateStatsDomainKind.REPLICATE_MEAN_SUFFIX, null);
```
cnathe, line 912: minor: could collapse to a single line:
`replicateRow.put(measure.getKey() + PlateReplicateStatsDomainKind.REPLICATE_MEAN_SUFFIX, !Double.isNaN(stat.getMean()) ? stat.getMean() : null);`

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -515,37 +749,275 @@ private Map<String, Object> getDataRowFromWell(String plateId, Well well, String
...
+
+                        if (!Double.isNaN(stat.getStdDev()))
+                            replicateRow.put(measure.getKey() + PlateReplicateStatsDomainKind.REPLICATE_STD_DEV_SUFFIX, stat.getStdDev());
+                        else
+                            replicateRow.put(measure.getKey() + PlateReplicateStatsDomainKind.REPLICATE_STD_DEV_SUFFIX, null);
```
cnathe, line 917: minor: could collapse to a single line:
`replicateRow.put(measure.getKey() + PlateReplicateStatsDomainKind.REPLICATE_STD_DEV_SUFFIX, !Double.isNaN(stat.getStdDev()) ? stat.getStdDev() : null);`

cnathe (approved):

PR #5907, opened 2024-10-02
attempt to fix intermittently-failing test in StudySimpleExportTest

labkey-chrisj:
#### Rationale
Intermittently, [StudySimpleExportTest.verifySuppressQueryValidation](https://teamcity.labkey.org/test/-903265045397405732?currentProjectId=LabkeyTrunk_DailySuites&expandTestHistoryInvestigationsSection=true&expandTestHistoryChartSection=true&expandedTest=build%3A%28id%3A3186563%29%2Cid%3A2000000022) fails while attempting to assert expected text to be present in order in a view of an import log file.

The test performs an import, navigates to the `FileContentBrowser `view, double-clicks the log-file object and attempts to assert that a few expected string values are present in the page that's opened in a separate browser tab- and when the test fails, it does so with error text like this: [Timed out getting page text. Page is probably too complex. Refactor test to look for specific element(s) instead.]

This change downloads the log file and uses a `TextSearcher `to assert the expected string values in order from it, rather than trying to scrape that much string value out of the browser.

#### Related Pull Requests
n/a

#### Changes

- [x] download the file and check its contents instead of getting the entire page text from the browser

### DIFF
--- a/study/test/src/org/labkey/test/tests/study/StudySimpleExportTest.java
+++ b/study/test/src/org/labkey/test/tests/study/StudySimpleExportTest.java
@@ -30,6 +30,7 @@ import org.labkey.test.components.DomainDesignerPage;
 import org.labkey.test.components.domain.DomainFormPanel;
 import org.labkey.test.components.ext4.Window;
 import org.labkey.test.pages.ImportDataPage;
+import org.labkey.test.pages.files.FileContentPage;
 import org.labkey.test.pages.study.DatasetDesignerPage;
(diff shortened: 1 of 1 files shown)


labkey-tchad:
I think that error message is probably misleading here. The text file in that tab is, almost certainly, less complicated than almost any other page we use `assertTextPresent` on. The test is probably trying to grab the tab contents before the new tab is ready (for some unknown measure of "ready", maybe some `document.readyState` value).
That being said, this is probably a fine fix since it's the log itself, not the tab, that we care about validating.

labkey-danield (approved):

PR #5908, opened 2024-10-02
Issue 51263: Deadlock while deleting materialIndexed values

XingY:
#### Rationale
From the [exception](https://www.labkey.org/_mothership/mothership-showStackTraceDetail.view?exceptionStackTraceId=38512) report, there are 2 transactions deadlocked waiting for ShareLock on exp.materialindexed table. One thread is attempting to delete a exp.materialindexed record (cascade deletion) as part of a sample type deletion.  The other thread is indexing samples in the sample type, which would insert/update the exp.materialindexed records. It's unknown which action triggered the search index to index or re-index the samples due to lack of server log. 

Sample type type delete action deletes samples in ascending rowId order. If the search indexer happened to be indexing a set of samples that's not in the same ascending order, deadlock might occur. This PR modifies sample/dataclass data so that they are indexed in rowId ascending order. 

#### Related Pull Requests
- 

#### Changes
- index sample/data in ascending rowId order for various CRUD actions to reduce deadlocks

### DIFF
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -2050,9 +2050,9 @@ public class ExpDataIterators
     public static class SearchIndexIteratorBuilder implements DataIteratorBuilder
     {
         final DataIteratorBuilder _pre;
-        final Function<Pair<List<String>, List<Integer>>, Runnable> _indexFunction;
+        final Function<SearchIndexDataKeys, Runnable> _indexFunction;
 
-        public SearchIndexIteratorBuilder(DataIteratorBuilder pre, Function<Pair<List<String>, List<Integer>>, Runnable> indexFunction)
+        public SearchIndexIteratorBuilder(DataIteratorBuilder pre, Function<SearchIndexDataKeys, Runnable> indexFunction)
         {
             _pre = pre;
             _indexFunction = indexFunction;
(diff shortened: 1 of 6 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2050,9 +2050,9 @@ private static String dataRole(ExpData data, User user)
     public static class SearchIndexIteratorBuilder implements DataIteratorBuilder
     {
         final DataIteratorBuilder _pre;
-        final Function<List<String>, Runnable> _indexFunction;
+        final Function<Pair<List<String>, List<Integer>>, Runnable> _indexFunction;
```
labkey-jeckels: Consider making a simple `record` class, which is more self-documenting (`lsids` and `rowIds` instead of `first` and `second`) than a `Pair`

labkey-jeckels (commented):
I really like the change to use RowIds when we have them available. They're smaller and therefore take less memory, it's the PK for the table, etc.

I'm less sold on the benefit of translating LSIDs to RowIds for the purpose of ordering them to avoid deadlocks. This approach actually adds a query against the target table, which seems like it would increase (or at least no decrease) the likelihood of deadlocks. Maybe I'm missing something about the benefits of RowId ordering?

XingY:
> I really like the change to use RowIds when we have them available. They're smaller and therefore take less memory, it's the PK for the table, etc.
> 
> I'm less sold on the benefit of translating LSIDs to RowIds for the purpose of ordering them to avoid deadlocks. This approach actually adds a query against the target table, which seems like it would increase (or at least no decrease) the likelihood of deadlocks. Maybe I'm missing something about the benefits of RowId ordering?

Changes made to only use rowIds when available, otherwise use lsid directly. The only time lsid might be needed is when there are new records created using MERGE option. So hopefully we can now index using ordered rowId for most scenarios, which would reduce deadlock by 2 threads insert/update/delete records in reversed order.

labkey-jeckels (approved):

PR #5918, opened 2024-10-04
Issue 51261: 'System Default Domain' tooltip

RosalineP:
#### Rationale
See issue [here](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51261).

#### Related Pull Requests
- N/A

#### Changes
- Add a ServerContextProvider


core/src/client/AuthenticationConfiguration/AuthenticationConfiguration.tsx
```diff
@@ -12,6 +12,16 @@ import { reorder, isEquivalent, addOrUpdateAnAuthConfig } from './utils';
 const UNSAVED_ALERT =
     'You have unsaved changes to your authentication configurations. Click "Save and Finish" to apply these changes.';
 
+export const App: FC = memo(() => {
```
labkey-alan: App should be defined below AuthenticationConfiguration, because it uses it. Yes this works because of JavaScript hoisting, but we should order things top to bottom when we can, and IIRC our linter complains about this as well.

labkey-alan (approved):
Looks good, left a minor piece of feedback

PR #5920, opened 2024-10-04
Issue 51403: Slow container renames end up timing out other threads

labkey-jeckels:
#### Rationale
It's a bad idea to allow multiple threads to concurrently rename, move, or delete the same container. We had delete protection in place and should expand it to other mutation events.

A future expansion would be to also lock all child containers.

#### Changes
- Track renames and moves alongside deletes
- Convert more errors to ApiUsageException as they're bad input, not buggy code
- Create a reusable AutoCloseable that doesn't throw exceptions
- Assorted code cleanup

### DIFF
--- a/api/src/org/labkey/api/data/generator/DataGenerator.java
+++ b/api/src/org/labkey/api/data/generator/DataGenerator.java
@@ -177,7 +177,7 @@ public class DataGenerator<T extends DataGenerator.Config>
             throw new CancelledException();
         }
 
-        if (ContainerManager.isMutating(c) == ContainerManager.MutatingOperation.delete)
+        if (!ContainerManager.exists(c))
         {
             job.warn("Container is being deleted: " + c.getPath());
             throw new CancelledException();
--- a/api/src/org/labkey/api/data/generator/DataGenerator.java
+++ b/api/src/org/labkey/api/data/generator/DataGenerator.java
@@ -15,8 +15,10 @@ import org.labkey.api.data.SqlSelector;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.TableSelector;
 import org.labkey.api.data.dialect.SqlDialect;
+import org.labkey.api.dataiterator.DataIteratorBuilder;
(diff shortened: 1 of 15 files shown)


api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -175,7 +177,7 @@ public static void checkAlive(PipelineJob job)
             throw new CancelledException();
         }
 
-        if (ContainerManager.isDeleting(c))
+        if (ContainerManager.isMutating(c) == ContainerManager.MutatingOperation.delete)
```
labkey-nicka: nit: Could also consider retaining `ContainerManager.isDeleting(c)` and have that internally do this mutation check. Then others don't have to know about the implementation detail of mutation operations if they don't want/need.

api/src/org/labkey/api/exp/LsidManager.java
```diff
@@ -375,7 +375,7 @@ public ActionURL getDisplayURL(Lsid lsid)
      */
     public boolean testResolveAll(Container c)
     {
-        if (ContainerManager.isDeleting(c))
+        if (ContainerManager.isMutating(c) != null)
```
labkey-nicka: nit: Consider having `ContainerManager.isMutating(c)` return a boolean and then a `ContainerManager.getMutation(c)` if that is necessary.

labkey-nicka (approved):
Looks good.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-jeckels: Thanks for the nudge. Turns out that `ContainerManager.exists()` is perfect for these callers, and lets me make `MutatingOperation` private within `ContainerManager`

api/src/org/labkey/api/exp/LsidManager.java
labkey-jeckels: Switched to `exists()`

PR #5921, opened 2024-10-04
Use separate caches for different Container caching needs

labkey-jeckels:
#### Rationale
We need to cache a few different things about containers. We can eliminate casting and awkward cache clearing by using separate caches.

#### Changes
- One cache per caching use case!

### DIFF
--- a/api/src/org/labkey/api/data/ContainerManager.java
+++ b/api/src/org/labkey/api/data/ContainerManager.java
@@ -152,15 +152,14 @@ public class ContainerManager
     private static final Logger LOG = LogHelper.getLogger(ContainerManager.class, "Container (projects, folders, and workbooks) retrieval and management");
     private static final CoreSchema CORE = CoreSchema.getInstance();
 
-    private static final String CONTAINER_PREFIX = ContainerManager.class.getName() + "/";
-    private static final String CONTAINER_CHILDREN_PREFIX = ContainerManager.class.getName() + "/children/";
-    private static final String CONTAINER_ALL_CHILDREN_PREFIX = ContainerManager.class.getName() + "/children/*/";
     private static final String PROJECT_LIST_ID = "Projects";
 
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -2034,37 +2033,34 @@ private static Container _addToCache(Container c)
...
-        CACHE.remove(CONTAINER_PREFIX + toString(c));
-        CACHE.remove(CONTAINER_PREFIX + c.getId());
-        CACHE.remove(CONTAINER_CHILDREN_PREFIX + c.getId());
-
-        if (null != parent)
-            CACHE.remove(CONTAINER_CHILDREN_PREFIX + parent.getId());
+        CACHE.remove(toString(c));
+        CACHE.remove(c.getId());
 
         // blow away the all children caches
```
labkey-adam:
```suggestion
        // blow away the children caches

```

api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -2034,37 +2033,34 @@ private static Container _addToCache(Container c)
     {
         assert DATABASE_QUERY_LOCK.isHeldByCurrentThread() : "Any cache modifications must be synchronized at a " +
                 "higher level so that we ensure that the container to be inserted still exists and hasn't been deleted";
-        CACHE.put(CONTAINER_PREFIX + toString(c), c);
-        CACHE.put(CONTAINER_PREFIX + c.getId(), c);
+        CACHE.put(toString(c), c);
+        CACHE.put(c.getId(), c);
```
labkey-adam, line 2038: If we're caching every container by ID and path, shouldn't the capacity be 2 * MAX?

labkey-adam (approved):
Love it, was thinking about doing this myself last time I looked at this class

PR #5841, opened 2024-09-08
Validator and metrics for non-standard date-time display formats

labkey-adam:
#### Rationale
First part of Date/Time formatting simplifications story: https://docs.google.com/document/d/1XcatQe9wRDw1ONLML9x4fFv_fIq9zhXPHoYANBuOykw/edit

Adds a validation provider for non-standard date display formats and a metrics provider that collects & reports all non-standard patterns.

#### Changes
- Define server-side lists of standard date and time format patterns. Add `isStandard*DisplayFormat()` methods. Push standard display formats into client-side `LABKEY` object.
- Rework site validation to register provider factories instead of providers. Creating a new provider on every run allows it to query and compute warnings before container iteration, which allows for a more efficient validation process in some cases.
- Add root container when validating "all projects & folders"
- Fix `ModuleReportResource` date parsing... don't use display formats to parse!
- Add validation for root/folder/project default date/time display formats
- Add validation for custom query and table override metadata (QueryDef XML)
- Add validation for provisioned table columns
- Add metrics provider implemented using the same code paths

PR #5923, opened 2024-10-07
Date/time formatting simplifications

labkey-adam:
#### Rationale
Server-side UI changes for https://docs.google.com/document/d/1XcatQe9wRDw1ONLML9x4fFv_fIq9zhXPHoYANBuOykw/edit#heading=h.4rg4h5vuek0

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5841

- https://github.com/LabKey/labkey-api-js/pull/184
- https://github.com/LabKey/platform/pull/5923
- https://github.com/LabKey/labkey-ui-components/pull/1607
- https://github.com/LabKey/labkey-ui-premium/pull/563
- https://github.com/LabKey/limsModules/pull/826
- https://github.com/LabKey/testAutomation/pull/2099
- https://github.com/LabKey/premiumModules/pull/76
- https://github.com/LabKey/ehrModules/pull/844
- https://github.com/LabKey/snprcEHRModules/pull/762

#### Changes
- Select date-time formats via drop-down lists holding standard formats (and non-standard, if already set)
- Add "Inherit" checkboxes to allow inheritance/override of individual properties (allowed previously for a few properties via hidden UI)
- Fix broken `LookAndFeelProperties` methods that ostensibly retrieved the stored value but actually retrieved inherited values. (Calling the parent class lookup method with a enum inadvertently walked the hierarchy.) Export and inheritance UI should be correct for all settings now.
- Simplify `SettingsForm` / `FolderSettingsForm` / `ProjectSettings` hierarchy: remove pointless SettingsForm interface and have ProjectSettings extend FolderSettings to reduce boilerplate
- Update `saveProjectSettings()` to take a form instead of JSON


api/src/org/labkey/api/data/Container.java
```diff
@@ -1384,10 +1385,27 @@ public Map<String, Object> toJSON(User user, boolean includePermissions, boolean
 
             LookAndFeelProperties props = LookAndFeelProperties.getInstance(this);
             Map<String, Object> formats = new HashMap<>();
-            formats.put("dateFormat", DateUtil.getDateFormatString(this));
+            formats.put("dateFormat", props.getDefaultDateFormat()); // when to use DateUtil.getDateFormatString vs laf.getDefaultDateFormat?
```
cnathe: is this comment a TODO for this PR or for a future refactor / change?

api/src/org/labkey/api/data/Container.java
```diff
@@ -1384,10 +1385,27 @@ public Map<String, Object> toJSON(User user, boolean includePermissions, boolean
...
+            if (includeParentFormatProps)
+            {
+                formats.put("dateFormatInherited", StringUtils.isEmpty(props.getDefaultDateFormatStored()));
+                formats.put("dateTimeFormatInherited", StringUtils.isEmpty(props.getDefaultDateTimeFormatStored()));
+                formats.put("numberFormatInherited", StringUtils.isEmpty(props.getDefaultNumberFormatStored()));
+                formats.put("timeFormatInherited", StringUtils.isEmpty(props.getDefaultTimeFormatStored()));
+
+                if (!isRoot())
+                {
+                    Container parentContainer = getParent();  // when to use DateUtil.getDateFormatString vs laf.getDefaultDateFormat?
```
cnathe: same question, is this comment meant for a TODO in this PR?

api/src/org/labkey/api/settings/LookAndFeelProperties.java
```diff
@@ -154,33 +158,60 @@ public static WriteableFolderLookAndFeelProperties getWriteableFolderInstance(Co
...
+    // Site-only properties
+
+    public String getCustomWelcome()
+    {
+        return lookupStringValue(customWelcome, null);
+    }
+
+    public DateParsingMode getDateParsingMode()
+    {
+        assert _settingsContainer.isRoot();
```
cnathe: looks like this assert is only in this one getter method. should it be in `getCustomWelcome` as well (and maybe others)?

api/src/org/labkey/api/settings/LookAndFeelProperties.java
```diff
@@ -192,46 +223,113 @@ public String getThemeName()
...
         String enabled = lookupStringValue(discussionEnabled.name(), null);
         return enabled != null ? "TRUE".equalsIgnoreCase(enabled) : lookupBooleanValue("dicussionEnabled", true);
     }
 
+    public Boolean isDiscussionEnabledStored()
+    {
+        // Prefer correctly spelled property name, but fall-back to the old, misspelled one
+        String enabled = getStoredValue(discussionEnabled);
+        if (enabled == null)
+            enabled = getStoredValue(_settingsContainer, "dicussionEnabled");
```
cnathe, line 277: bummer to have to keep the fallback to the misspelled property name. is that something we could just fix in an upgrade script and then stop supporting here?

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -5504,25 +5556,69 @@ public void validateCommand(FolderSettingsForm form, Errors errors)
...
+        validateAndSaveFormat(form.getDefaultDateTimeFormat(), form.isDefaultDateTimeFormatInherited(), props::clearDefaultDateTimeFormat, props::setDefaultDateTimeFormat, errors, "date-time");
+        validateAndSaveFormat(form.getDefaultTimeFormat(), form.isDefaultTimeFormatInherited(), props::clearDefaultTimeFormat, props::setDefaultTimeFormat, errors, "time");
+        validateAndSaveFormat(form.getDefaultNumberFormat(), form.isDefaultNumberFormatInherited(), props::clearDefaultNumberFormat, props::setDefaultNumberFormat, errors, "number");
+        validateAndSaveFormat(form.getExtraDateParsingPattern(), form.isExtraDateParsingPatternInherited(), props::clearExtraDateParsingPattern, props::setExtraDateParsingPattern, errors, "date");
+        validateAndSaveFormat(form.getExtraDateTimeParsingPattern(), form.isExtraDateTimeParsingPatternInherited(), props::clearExtraDateTimeParsingPattern, props::setExtraDateTimeParsingPattern, errors, "date-time");
+        validateAndSaveFormat(form.getExtraTimeParsingPattern(), form.isExtraTimeParsingPatternInherited(), props::clearExtraTimeParsingPattern, props::setExtraTimeParsingPattern, errors, "time");
```
cnathe: do we want the "what" last parameter to be distinquishable for these "extra parsing" options? right now these match the lines 5566 through 5569 above (but maybe we don't care since this is just used in the error message).

core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
```diff
@@ -43,36 +54,50 @@
...
+    {
+        Container parent = c.getParent();
+        parentName = parent.isRoot() ? "the site root" : (parent.isProject() ? "project" : "folder") + " " + parent.getPath();
+        String helpText = "Settings where the \"Inherited\" box is checked inherit their values from " + parentName +
+                ". Settings where \"Inherited\" is unchecked override their values in this " + (c.isProject() ? "project": "folder") + ".";
+        inheritHelp = helpPopup("Inherited", helpText, false);
+    }
+    HtmlString clearMessage = HtmlString.of(
+        c.isRoot() ? "all look & feel properties to their default values" : (
+        c.isProject() ? "all project settings to inherit from the site look & feel properties" :
```
cnathe, line 70: the `&` here ends up showing up as `&amp;` in the FF and Chrome modal:
Screenshot 2024-10-15 at 9 54 38 AM

core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
```diff
@@ -95,107 +120,133 @@
...
 <tr>
-    <td colspan=2>Customize the look and feel of <%=h(c.isRoot() ? "your LabKey Server installation" : "the '" + c.getProject().getName() + "' project")%> (<%=bean.helpLink%>)</td>
+    <td>Customize the look and feel of <%=h(c.isRoot() ? "your LabKey Server installation" : "the '" + c.getProject().getName() + "' project")%> (<%=bean.helpLink%>)</td>
+    <%
+        if (c.isProject())
+        {
+    %>
+        <td style="padding-left: 5px; padding-right: 5px;">Inherited<%=inheritHelp%></td>
+    <%
+        }
```
cnathe, line 131: should this have an else that writes out an empty `<td>` so that the colspan of 3 matches? or should the colspan be conditional on `c.isProject()`?

core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
```diff
@@ -43,36 +54,50 @@
...
     {
 %>
 <tr>
-    <td colspan=2>Security defaults</td>
+    <td colspan="3">Security defaults</td>
 </tr>
 <tr>
     <td class="labkey-form-label"><label for="shouldInherit">New folders should inherit permissions by default</label></td>
-    <td><input type="checkbox" id="shouldInherit" name="shouldInherit" size="50"<%=checked(SecurityManager.shouldNewSubfoldersInheritPermissions(c))%>></td>
+    <td><input type="checkbox" id="shouldInherit" name="shouldInherit" size="<%=standardInputWidth%>"<%=checked(SecurityManager.shouldNewSubfoldersInheritPermissions(c))%>></td>
```
cnathe: for consistency in the UI, should we add an empty `<td>` here so that the checkbox lines up with the 3rd table column (since this property doesn't have the inherited checkox).

core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
```diff
@@ -408,4 +499,109 @@
...
+            set.add(current);
+            set.addAll(options);
+            options = set;
+        }
+
+        // Add a non-standard and an invalid format to every drop-down for testing purposes
+        // TODO: Delete this block - just for testing
+        Set<String> set = new LinkedHashSet<>(options);
+        set.add(type == DateDisplayFormatType.Date ? "MM-dd-yyyy" : "kk:mm"); // non-standard
+        set.add("xyz"); // invalid
```
cnathe: just adding a note here to make sure this is removed after testing / before merge.

cnathe (approved):

api/src/org/labkey/api/settings/LookAndFeelProperties.java
labkey-susanh: Should this be an explicit check that throws an exception (or the like) so we don't do the wrong thing if assertions are not enabled?

api/src/org/labkey/api/data/Container.java
XingY: I added this comment and hope someone with more knowledge could provide an answer. Historically the usage have been inconsistent.

api/src/org/labkey/api/settings/LookAndFeelProperties.java
XingY: I added this assert originally. Agree it should be updated to exception.

api/src/org/labkey/api/data/Container.java
labkey-adam: Interesting... I would have thought that `props.getDefaultDateFormat()` would be the preferred method, but it always walks the container hierarchy whereas the `DateUtil` method uses the cache. I'll take a closer look at this... we may want to adjust LookAndFeelProperties to reference the cache since it's used extensively.

api/src/org/labkey/api/data/Container.java
labkey-adam: I'll take a look

api/src/org/labkey/api/data/Container.java
```diff
@@ -1384,10 +1385,27 @@ public Map<String, Object> toJSON(User user, boolean includePermissions, boolean
...
-            formats.put("dateFormat", DateUtil.getDateFormatString(this));
+            formats.put("dateFormat", props.getDefaultDateFormat()); // when to use DateUtil.getDateFormatString vs laf.getDefaultDateFormat?
             formats.put("dateTimeFormat", props.getDefaultDateTimeFormat());
             formats.put("numberFormat", props.getDefaultNumberFormat());
-            formats.put("timeFormat", DateUtil.getTimeFormatString(this));
+            formats.put("timeFormat", props.getDefaultTimeFormat());
+
+            if (includeParentFormatProps)
+            {
+                formats.put("dateFormatInherited", StringUtils.isEmpty(props.getDefaultDateFormatStored()));
```
labkey-adam: I'll update these, but I think they should be checking for null, not empty. Number format is often set to "", which doesn't mean inherit.

api/src/org/labkey/api/settings/LookAndFeelProperties.java
labkey-adam: I would do that, but discussions (and this setting) are deprecated and likely going away soon, so not worth the bother.

core/src/org/labkey/core/admin/AdminController.java
labkey-adam: `what` is now more specific in all cases

api/src/org/labkey/api/data/Container.java
labkey-adam: These methods returned identical results, but one used the cache and one didn't. I did a mini-refactor on `LookAndFeelFolderProperties` so it delegates to the cache as well, except for new methods (e.g., `calculateDefaultDateFormat()`) called only by the cache loader. `DateUtil.getDateFormatString()` and `props.getDefaultDateFormat()` are now identical; callers can choose one or the other based solely on convenience (code needing just a single format would likely call `DateUtil`; code needing multiple properties would likely use LAFFP).

core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
labkey-adam: Double encoding fixed

core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
labkey-adam: Sure

core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
labkey-adam: This <td> didn't have a colspan, so I added it conditionally (site case)

core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
labkey-adam: Added

cnathe (approved):

core/src/org/labkey/core/admin/lookAndFeelProperties.jsp
labkey-adam: Removed

PR #5924, opened 2024-10-07
Misc 24.11 fixes

cnathe:
#### Rationale
Issue [51373](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51373): Simplify description of Author role to match other roles
Issue [51321](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51321): Sample Type and Data Class to consider "All" and "First" as reserved names
Issue [50376](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=50376): DataClassFolderWriter refactor to get ExpRuns ids per data class instead of all at once

#### Changes
- simplify author permissions role description
- create and update domain for SampleType and DataClass to check for reserved type names
- DataClassFolderWriter refactor to get ExpRuns ids per data class instead of all at once

### DIFF
--- a/experiment/src/org/labkey/experiment/samples/DataClassFolderWriter.java
+++ b/experiment/src/org/labkey/experiment/samples/DataClassFolderWriter.java
@@ -50,7 +50,6 @@ public abstract class DataClassFolderWriter extends AbstractExpFolderWriter
         XarExportSelection typesSelection = new XarExportSelection();
         XarExportSelection runsSelection = new XarExportSelection();
         Set<ExpDataClass> dataClasses = new HashSet<>();
-        List<ExpData> datasToExport = new ArrayList<>();
         _exportPhiLevel = ctx.getPhiLevel();
         boolean exportTypes = false;
         boolean exportRuns = false;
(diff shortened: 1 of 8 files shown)


experiment/src/org/labkey/experiment/samples/DataClassFolderWriter.java
```diff
@@ -75,25 +77,28 @@ public void write(Container c, FolderExportContext ctx, VirtualFile vf) throws E
...
+                    return lsid.equals(ExperimentService.SAMPLE_DERIVATION_PROTOCOL_LSID) && isValidRunType(ctx, run);
+                })
+                    .collect(Collectors.toSet())
+                    .stream().map(ExpObject::getRowId).toList();
+
+                if (!exportedRunIds.isEmpty())
+                {
+                    runsSelection.addRunIds(exportedRunIds);
+                    exportRuns = true;
+                }
```
labkey-susanh, line 100: This will certainly be an improvement. At some point we should also consider using streams instead of lists, but that's a bigger change, of course.

labkey-susanh (approved):

PR #5930, opened 2024-10-08
Experiment: stream results from the lineage API

labkey-nicka:
#### Rationale
As a part of my investigation into experiment lineage API performance it was clear that we could move towards a streaming approach for lineage results as opposed to loading everything into memory before sending down the wire. These changes stream the lineage API payload after the initial lineage query is performed allowing for batched resolution (and discard) of experiment objects (i.e. `ExpData`, `ExpMaterial`, etc.).

#### Changes
- Refactor `ExpLineage.toJSON()` into a new `ExpLineageServiceImpl.streamLineage()` which streams the lineage results without building up the entire JSON response in memory.
- Convert `ExpLineage.Edge` from a class to a record. Inline the least useful "no role" role that all of these edges have.
- Introduce `ExpLineage.Edges` (note the trailing "s") to make it more conceptually clear the parents and children edges.
- Update `ApiJsonWriter` interface to allow for starting and ending JSON objects.
- Introduce a new `ExpLineageService` to house experiment lineage related services. 
  - The old standard of `ExpServiceImpl` has eclipsed 10k LOC and just feels like it needs to be broken out. That said, I didn't fully refactor out everything that is "lineage related" but more just the primary methods for querying and streaming lineage results.


api/src/org/labkey/api/action/ApiJsonWriter.java
```diff
@@ -458,12 +465,12 @@ public void testExceptionNotCommitted() throws IOException
...
-            assertEquals( "throwing up", json.getString("exception") );
-            assertTrue( json.has("stackTrace") );
-            assertFalse( json.has("schemaName"));
-            System.out.println(json);
+            assertEquals("throwing up", json.getString("exception"));
+            assertTrue(json.has("stackTrace"));
+            assertFalse(json.has("schemaName"));
         }
 
+        @Test
```
labkey-nicka: @labkey-matthewb This is a test case that you introduced a long time ago that wasn't annotated as such. I've enabled it but it fails. Might you have thoughts on the expectations here?

api/src/org/labkey/api/data/generator/DataGenerator.java
```diff
@@ -502,7 +502,9 @@ private int generateAliquotsForParents(List<Map<String, Object>> parents, QueryU
         // for some of the aliquots, possibly generate further aliquot generations
         if (generatedCount < quantity && generation < maxGenerations)
         {
-            generatedCount += generateAliquotsForParents(aliquots.subList(randomInt(0, aliquots.size() / 2), randomInt(aliquots.size() / 2, aliquots.size())), svc, quantity - generatedCount, numGenerated + generatedCount, generation + 1, maxGenerations, sampleStatuses);
+            List<Map<String, Object>> nextParents = aliquots.isEmpty() ? parents : aliquots;
```
labkey-nicka, line 505: @labkey-susanh I was utilizing the data generator and had set the `percentAliquots=1`. This ended up causing an infinite recursion. I've updated the logic here to attempt to avoid this. This is not necessary to include but thought you might be curious.

api/src/org/labkey/api/exp/api/ExperimentJSONConverter.java
```diff
@@ -526,8 +526,9 @@ private static JSONObject serializeIdentifiableBean(@NotNull Identifiable obj, @
         if (url != null)
             json.put(URL, url);
 
-        json.put(CONTAINER, obj.getContainer().getId());
-        json.put(CONTAINER_PATH, obj.getContainer().getId());
+        final String containerId = obj.getContainer().getId();
+        json.put(CONTAINER, containerId);
+        json.put(CONTAINER_PATH, containerId); // TODO: Should probably add a comment here about why not getPath()
```
labkey-nicka: @labkey-susanh Before I add a comment here I just wanted to check in with you that this was indeed intentional. This comes from your changes in #4597.

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
```diff
@@ -0,0 +1,404 @@
...
+public class ExpLineageServiceImpl implements ExpLineageService
+{
+    private static final Logger LOG = LogHelper.getLogger(ExpLineageServiceImpl.class, "Experiment lineage service");
+
+    public static ExpLineageServiceImpl get()
+    {
+        return (ExpLineageServiceImpl) ExpLineageService.get();
+    }
+
+    private static void forEachBatchExpData(Collection<Integer> rowIds, int batchSize, final Selector.ForEachBatchBlock<ExpDataImpl> block)
```
labkey-nicka: nit: Someone might know a prettier way to do this where the generics actually work. I tinkered for awhile but could not get it to be as generic as I'd expect.

labkey-jeckels: What were you wanting to have? A consolidated `forEachBatchExpObject()` or similar? I doubt you do better than the two separate data and material variants.

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
```diff
@@ -0,0 +1,404 @@
...
+            }
+            else
+            {
+                edges.add(new ExpLineage.Edge(parentLSID, childLSID));
+
+                // Don't include the seed in the lineage collections
+                if (!seedIdentifiers.lsids().contains(parentLSID))
+                {
+                    // process parents
+                    if ("Data".equals(parentExpType))
```
labkey-jeckels: Can we use constants for these string comparisons? Ideally existing constants, but new ones if required.

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    private static void forEachBatchExpData(Collection<Integer> rowIds, int batchSize, final Selector.ForEachBatchBlock<ExpDataImpl> block)
+    {
+        if (rowIds.isEmpty())
+            return;
+
+        SimpleFilter filter = new SimpleFilter(FieldKey.fromParts(ExpDataTable.Column.RowId.name()), rowIds, IN);
+        new TableSelector(ExperimentServiceImpl.get().getTinfoData(), filter, null)
+                .forEachBatch(Data.class, batchSize, (batch) -> block.exec(ExpDataImpl.fromDatas(batch)));
```
labkey-jeckels: Is there a benefit to doing these as batches instead of individual `forEach()` handlers?

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
```diff
@@ -0,0 +1,404 @@
...
+
+        json.put("parents", edges.parents().stream().map(ExpLineage.Edge::toParentJSON).toList());
+        json.put("children", edges.children().stream().map(ExpLineage.Edge::toChildJSON).toList());
+
+        return json;
+    }
+
+    private static void writeSeed(LineageResult lineage, StreamContext context, ExpLineageOptions options) throws IOException
+    {
+        // If the request was made with a single 'seed' property, use single 'seed' property in the response
```
labkey-jeckels, line 358: This seems odd but I see it's the old behavior

labkey-jeckels (approved):
Some small suggestions and a question or two. No need for me to approve again but happy to review anything else.

api/src/org/labkey/api/data/generator/DataGenerator.java
labkey-susanh: Thanks for fixing that.

api/src/org/labkey/api/exp/api/ExperimentJSONConverter.java
labkey-susanh: This looks like a cut-and-paste error. It may work as is since we often can resolve via the guid or the path, but I don't remember explicitly setting this to be `containerId`

api/src/org/labkey/api/exp/api/ExpLineageService.java
```diff
@@ -0,0 +1,27 @@
...
+        return ServiceRegistry.get().getService(ExpLineageService.class);
+    }
+
+    static void setInstance(ExpLineageService impl)
+    {
+        ServiceRegistry.get().registerService(ExpLineageService.class, impl);
+    }
+
+    /**
+     * Get the lineage for the seed Identifiable object. Typically, the seed object is a ExpMaterial,
```
labkey-susanh:
```suggestion
     * Get the lineage for the seed Identifiable object. Typically, the seed object is an ExpMaterial,
```

api/src/org/labkey/api/exp/api/ExpLineageService.java
```diff
@@ -0,0 +1,27 @@
...
+     * a ExpData (in a DataClass), or an ExpRun.
```
labkey-susanh:
```suggestion
     * an ExpData (in a DataClass), or an ExpRun.
```

api/src/org/labkey/api/exp/api/ExpLineage.java
```diff
@@ -413,125 +396,32 @@ else if ((clazz == ExpMaterial.class && target instanceof ExpMaterial) ||
...
-
-            Edge edge = (Edge) o;
-
-            if (!parent.equals(edge.parent)) return false;
-            if (!child.equals(edge.child)) return false;
-            return _role.equals(edge._role);
-
+            JSONObject json = new JSONObject();
+            json.put("lsid", target);
+            json.put("role", NO_ROLE);
```
labkey-susanh: If this is always `NO_ROLE` is it actually useful?

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
```diff
@@ -0,0 +1,404 @@
...
+            return;
+
+        SimpleFilter filter = new SimpleFilter(FieldKey.fromParts(ExpMaterialTable.Column.RowId.name()), rowIds, IN);
+        new TableSelector(ExperimentServiceImpl.get().getTinfoMaterial(), filter, null)
+                .forEachBatch(Material.class, batchSize, (batch) -> block.exec(ExpMaterialImpl.fromMaterials(batch)));
+    }
+
+    private static void forEachBatchExpRun(Collection<Integer> rowIds, int batchSize, final Selector.ForEachBatchBlock<ExpRunImpl> block)
+    {
+        if (rowIds == null || rowIds.isEmpty())
```
labkey-susanh: Why the null check here but not elsewhere? Should there be different annotations in the different methods?

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
```diff
@@ -0,0 +1,404 @@
...
+            {
+                LOG.warn("Lineage not available for unknown material: " + seed.getLSID());
+                continue;
+            }
+
+            // ensure the user has read permission in the seed container
+            if (c != null && !c.equals(seed.getContainer()))
+            {
+                if (!seed.getContainer().hasPermission(user, ReadPermission.class))
+                    throw new UnauthorizedException("Lineage not available. User does not have permission to view " + seed.getName() + ".");
```
labkey-susanh: Seems to be a chance of leaking a name the non-permissioned user is not meant to see here.

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
```diff
@@ -0,0 +1,404 @@
...
+                throw new RuntimeException(e);
+            }
+        };
+    }
+
+    private static void writeBatchList(List<? extends Identifiable> batch, StreamContext context) throws IOException
+    {
+        for (var node : batch)
+        {
+            if (context.hasPermission(node.getContainer(), context.user))
```
labkey-susanh: Are we bypassing permission checks when getting the objects that are being streamed so we have to check that permission here? If so, it seems like we should be checking for `DataClassReadPermission` and `MediaReadPermission` somewhere. If not is this redundant or does it save processing by cutting checks off at the pass?

labkey-susanh (approved):

api/src/org/labkey/api/exp/api/ExpLineage.java
labkey-nicka: I was looking to be fully backwards compatible but I've switched and just removed this superfluous property. I've left a note on the `record Edge` regarding this change.

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
labkey-nicka: Removed.

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
labkey-nicka: Updated error to reflect the seed LSID which is what was provided to the API by the user.

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
labkey-nicka: I agree.

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
labkey-nicka: I've switch to using `forEach()`.

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
labkey-nicka: Updated constants all over the place.

experiment/src/org/labkey/experiment/lineage/ExpLineageServiceImpl.java
labkey-nicka: Switched to using `forEach()` and was able to get better mileage out of the generics.

api/src/org/labkey/api/action/ApiJsonWriter.java
labkey-nicka: I've unmarked this test case and left a TODO comment.

PR #5933, opened 2024-10-09
Issue 51440: Safari no longer needs special date formatting

labkey-nicka:
#### Rationale
This addresses [Issue 51440](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51440) by removing Safari-specific date format handling. See ticket for further details.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/2457

#### Changes
- Update `DateUtil.formatJsonDateTime()` to no longer analyze request for browser context
- Remove `DateUtil.getSafariJsonDateTimeFormatString()`

### DIFF
--- a/api/src/org/labkey/api/util/DateUtil.java
+++ b/api/src/org/labkey/api/util/DateUtil.java
@@ -1330,18 +1330,17 @@ validNum:       {
 
     public static String formatJsonDateTime(Date date)
     {
-        if (HttpView.hasCurrentView())
+        if (date instanceof Time)
         {
-            ViewContext context = HttpView.currentContext();
-            if (context != null && context.getRequest() != null)
+            if (HttpView.hasCurrentView())
             {
-                if (date instanceof Time && context.getContainer() != null)
+                ViewContext context = HttpView.currentContext();
+                if (context != null && context.getContainer() != null)
                     return FastDateFormat.getInstance(FolderSettingsCache.getDefaultTimeFormat(context.getContainer())).format(date);
             }
-        }
 
-        if (date instanceof Time)
             return jsonTimeFormat.format(date);
+        }
 
         return jsonDateFormat.format(date);
     }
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
It would be great to get rid of this special casing for Safari. Any guesses of what version of Safari enabled parsing of our default format though? I'm worried that there may be versions of Safari still within their supported lifespan that still need this workaround.

labkey-nicka:
> It would be great to get rid of this special casing for Safari. Any guesses of what version of Safari enabled parsing of our default format though? I'm worried that there may be versions of Safari still within their supported lifespan that still need this workaround.

I believe it was patched almost 3 years ago. https://github.com/WebKit/WebKit/commit/09c28ab505d2306a60c363c817eb97ec196760f2. I do not know what the first version of Safari is that contains this patch but I think we are safe.

api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -1332,31 +1326,23 @@ public static boolean isSpecialNamedFormat(String dateFormat)
...
-        // Issue 43557 - Safari needs a 'T' between the date and time. Instead of needing to rev all client APIs to parse
-        // the new variant, conditionally send Safari its preferred format (which works fine in other browsers but not
-        // in all other parsing code)
-        boolean isSafari = false;
-        if (HttpView.hasCurrentView())
+        if (date instanceof Time)
         {
-            ViewContext context = HttpView.currentContext();
-            if (context != null && context.getRequest() != null)
+            if (HttpView.hasCurrentView())
```
labkey-jeckels: We can eliminate this check. `HttpView.currentContext()` will return null if there's no current view.

labkey-jeckels (approved):

api/src/org/labkey/api/util/DateUtil.java
labkey-nicka: Done.

PR #5935, opened 2024-10-10
PlateGridInfo parsing to support measure annotations only when plate set only has a single plate

cnathe:
#### Rationale
We have an example file of graphical plate assay data with multiple measures where the plate annotations are the measure names/labels but the header annotations do not have the plate identifier. We can support this scenario with a small code change so that if the plate set selected only has a single plate, we use that as the plate for the import and assume that the single annotation is the measure name. 

#### Changes
- PlateGridInfo check for single plate in plate set for the single annotation case


labkey-klum (approved):
Seems fine, just curious was this found by a client (or internal testing), it seems like such a corner case.

cnathe:
> Seems fine, just curious was this found by a client (or internal testing), it seems like such a corner case.

This was based on an example data file provided by a sales prospect.

PR #5938, opened 2024-10-10
Calculated fields error handling

cnathe:
#### Rationale
The initial implementation for calculated fields in the field editor UI did some validation of the expression after the user enters it into the domain row textarea but there were still plenty of cases where the user didn't see an error until they saved the domain and tried to view the grid. In this case, sometimes the error would actually prevent the grid from loading and just showed an error message (for example for a calculated field expression that results in a "bad SQL" error). This PR improves on the validation check that is done within the field editor UI before save and also attempts to recognize the case where the grid getQuery call has an error and retries the query while dropping the calculated fields.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1582

#### Changes
- add view name to NotFoundException
- AbstractTableInfo._allowCalculatedColumns to prevent inclusion of calc fields in QueryUpdateService case
- remove experimental feature flag for calculated fields
- parseCalculatedColumn to include PHI check in validation (Issue 51235)
- query metadata editor fix for calc field locked status and onChangeHandler


api/src/org/labkey/api/query/UserSchema.java
```diff
@@ -164,21 +164,28 @@ final public TableInfo getTable(String name, boolean includeExtraMetadata)
...
+    final public TableInfo getTable(String name, @Nullable ContainerFilter cf, boolean allowCalculatedColumns)
+    {
+        return getTable(name, cf, true, false, true, allowCalculatedColumns);
     }
 
     @Nullable
     public TableInfo getTable(String name, @Nullable ContainerFilter cf, boolean includeExtraMetadata, boolean forWrite)
     {
-        return getTable(name, cf, includeExtraMetadata, forWrite, false);
+        return getTable(name, cf, includeExtraMetadata, forWrite, true, true);
```
labkey-susanh: This seems to be changing the value for `skipSuggestedColumns`. Is that intentional?

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -3455,7 +3455,7 @@ public ApiResponse execute(APIQueryForm form, BindException errors)
             if (null != StringUtils.trimToNull(form.getViewName()) &&
                     null == view.getQueryDef().getCustomView(getUser(), getViewContext().getRequest(), form.getViewName()))
             {
-                throw new NotFoundException("The requested view does not exist for this user!");
+                throw new NotFoundException("The requested view does not exist for this user: " + form.getViewName() + ".");
```
labkey-susanh: I might phrase this as "The requested view '" + form.getViewName() + "' does not exist for this user." just to make it more friendly.

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -8092,11 +8093,23 @@ public void setExpression(String expression)
...
+            this.phiColumns = phiColumns;
+        }
+
         @Override
         public void bindJson(JSONObject json)
         {
             if (json.has("expression"))
                 setExpression(json.getString("expression"));
+            if (json.has("phiColumns"))
+                setPhiColumns(json.getJSONArray("phiColumns").toList().stream().map(s -> FieldKey.fromString(s.toString())).collect(Collectors.toList()));
```
labkey-susanh, line 8112: Doesn't the server already know which columns are PHI columns?

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -8169,6 +8182,8 @@ public Object execute(ParseForm form, BindException errors) throws Exception
                 for (var entry : form.getColumnMap().entrySet())
                 {
                     BaseColumnInfo entryCol = new BaseColumnInfo(entry.getKey(), entry.getValue());
+                    if (form.getPhiColumns().contains(entry.getKey()))
+                        entryCol.setPHI(PHI.PHI); // the specific PHI level to set doesn't matter as long as it is > NotPHI
```
labkey-susanh: Might want to comment why this is being set here at all.

labkey-susanh (approved):

api/src/org/labkey/api/query/UserSchema.java
cnathe: good catch. definitely not intentional. fixed

query/src/org/labkey/query/controllers/QueryController.java
cnathe: done

query/src/org/labkey/query/controllers/QueryController.java
cnathe: Not always for this case. This is used only for the ParseCalculatedColumnAction which can be used for a new or an existing domain. So we need to pass over the field editor's current field state / info (which can include existing fields, new fields, edited fields).

query/src/org/labkey/query/controllers/QueryController.java
cnathe: done

PR #5939, opened 2024-10-10
Updates to CPUTimer

labkey-danield:
#### Rationale
The CPUTimer maintains an internal list of timers, this adds a method to the class that allows the caller to remove a timer from the collection.
Updated the clear method to reset all the internal timing variables. This can be used if a caller wants the timer to remain but doesn't want the current information. Example: throwing away the first iteration in a run.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/812

#### Changes
- Add a remove method.
- Update clear method to reset all local variable.

### DIFF
--- a/api/src/org/labkey/api/util/CPUTimer.java
+++ b/api/src/org/labkey/api/util/CPUTimer.java
@@ -117,15 +117,18 @@ public class CPUTimer
 
     protected void _update(long elapsed)
     {
-        _cumulative += elapsed;
-        _min = Math.min(_min, elapsed);
-        _max = Math.max(_max, elapsed);
-        if (_first == 0)
+        synchronized(timers)
         {
-            _first = elapsed;
+            _cumulative += elapsed;
+            _min = Math.min(_min, elapsed);
+            _max = Math.max(_max, elapsed);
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/util/CPUTimer.java
```diff
@@ -143,13 +143,41 @@ public boolean saveTo(CPUTimer accumulator)
...
+        _first = 0;
+        _last = 0;
+        _start = 0;
+        _stop = 0;
+        _calls = 0;
+        return true;
+    }
+
+    /**
+     * Remove this time from the internal collection of timers.
```
labkey-susanh: Since this doesn't destroy the object, but does call clear, I would include something about that side effect in the JavaDoc. Or don't call clear here and require the caller to do that separately (which seems better since it doesn't have a side effect).

labkey-susanh (approved):

api/src/org/labkey/api/util/CPUTimer.java
labkey-danield: Changed.
I've removed the call to clear.

PR #5941, opened 2024-10-11
Assay transform script fix (for re-import)

labkey-klum:
#### Rationale
This fixes the error with the null value for the `ExpData sourceApplication` during re-import when there was a transform script configured. The problem was that during a transform we need to parse and write the results file for the transform script. This happens around `TsvDataExchangeHandler:291`. The `ExpData` object is fake and is just there to pass in the data file for parsing but it doesn't represent the actual uploaded data. If this is the case we want to ignore that object and instead find the actual data object which has been fully initialized.

I considered not creating the temporary `ExpData` object during the transform step but couldn't be sure if there was a reason we didn't want to use the actual object.


cnathe (approved):
Change looks good. I also pulled it down locally and verified it with the assay re-import case that appends plate data and a re-import that replaces plate data. Both look to work as expected and the data written to the transform script runDataFile.tsv matched what I expected.

PR #5949, opened 2024-10-16
Issue 51305: insertRows does not trim before attempting to parse ints

labkey-jeckels:
#### Rationale
We've been happy to convert padded strings to floats and doubles, but not ints, longs, or booleans. We can handle them all!

#### Changes
- Trim before attempting conversions
- Misc code cleanup

### DIFF
--- a/api/src/org/labkey/api/data/ConvertHelper.java
+++ b/api/src/org/labkey/api/data/ConvertHelper.java
@@ -244,13 +244,13 @@ public class ConvertHelper implements PropertyEditorRegistrar
             if (value instanceof Boolean)
                 return value;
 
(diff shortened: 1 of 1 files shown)


labkey-martyp (approved):
Looks fine to me. All of these look like they would have thrown exceptions before so I don't think there's any risk of mismatched data.

labkey-jeckels:
> Looks fine to me. All of these look like they would have thrown exceptions before so I don't think there's any risk of mismatched data.

Correct. We used to reject values like ` 10` or `false ` but will now trim and accept them. No changes to how we treat values that are going into string/varchar/text fields.

PR #5952, opened 2024-10-16
Compute and show file root sizes

labkey-adam:
#### Rationale
Administrators want to know how much space each file root consumes in the file system. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=50393

#### Related Pull Requests
* https://github.com/LabKey/premiumModules/pull/75

### DIFF
--- a/api/src/org/labkey/api/util/FileUtil.java
+++ b/api/src/org/labkey/api/util/FileUtil.java
@@ -71,6 +71,7 @@ import java.nio.file.attribute.PosixFilePermissions;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashSet;
@@ -1971,6 +1972,20 @@ quickScan:
         }
     }
 
+    private static final String[] UNITS = new String[] { "bytes", "KB", "MB", "GB", "TB", "PB", "EB" };
+
+    /**
+     * Formats a (potentially large) number of bytes in a human-readable way: in bytes, KB, MB, GB, etc. as appropriate
+     */
+    public static String formatFileSize(long size)
+    {
+        if (size <= 0)
+            return "0 bytes";
+
+        int digitGroups = (int) (Math.log10(size) / Math.log10(1024));
+
+        return new DecimalFormat("#,##0.#").format(size / Math.pow(1024, digitGroups)) + " " + UNITS[digitGroups];
+    }
 
     @SuppressWarnings("SSBasedInspection")
     public static class TestCase extends Assert
@@ -2178,5 +2193,19 @@ quickScan:
             assertNotNull(isAllowedFileName("a`b"));
         }
 
+        @Test
+        public void testFormatFileSize()
+        {
+            assertEquals("0 bytes",   formatFileSize(           -22));
+            assertEquals("0 bytes",   formatFileSize(             0));
+            assertEquals("100 bytes", formatFileSize(           100));
+            assertEquals("2 KB",      formatFileSize(         2_000));
+            assertEquals("195.3 KB",  formatFileSize(       200_000));
+            assertEquals("1.9 MB",    formatFileSize(     2_000_000));
+            assertEquals("19.1 MB",   formatFileSize(    20_000_000));
+            assertEquals("190.7 MB",  formatFileSize(   200_000_000));
+            assertEquals("1.9 GB",    formatFileSize( 2_000_000_000));
+            assertEquals("18.6 GB",   formatFileSize(20_000_000_000L));
+        }
     }
 }
--- a/api/src/org/labkey/api/util/FileUtil.java
+++ b/api/src/org/labkey/api/util/FileUtil.java
@@ -71,7 +71,6 @@ import java.nio.file.attribute.PosixFilePermissions;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashSet;
@@ -82,15 +81,9 @@ import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
-
-/**
- * User: jeckels
- * Date: Dec 5, 2005
- */
 public class FileUtil
 {
     public static final String FILE_SCHEME = "file";    // url scheme for local file system
-
 
     private static final Logger LOG = LogHelper.getLogger(FileUtil.class, "FileUtil.java logger");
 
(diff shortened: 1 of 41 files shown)


filecontent/src/org/labkey/filecontent/FileContentModule.java
```diff
@@ -221,8 +228,21 @@ public void doStartup(ModuleContext moduleContext)
                 results.put("fileRootCrawlSucceeded", succeeded);
                 results.put("fileRootMillisecondsToCalculateSize", System.currentTimeMillis() - startTime);
             }
```
labkey-adam: @labkey-jeckels I'd like to get rid of the entire block above, but wanted to get your feedback

api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -1971,6 +1972,20 @@ public void close() throws IOException
         }
     }
 
+    private static final String[] UNITS = new String[] { "bytes", "KB", "MB", "GB", "TB", "PB", "EB" };
+
+    /**
+     * Formats a (potentially large) number of bytes in a human-readable way: in bytes, KB, MB, GB, etc. as appropriate
+     */
+    public static String formatFileSize(long size)
```
labkey-matthewb: FileUtils.byteCountToDisplaySize()

labkey-matthewb (approved):

core/src/org/labkey/core/admin/view/filesProjectSettings.jsp
```diff
@@ -98,43 +100,51 @@
...
             files. By default, LabKey stores your files in a standard directory structure. Site administrators can override this location for each
             folder if you wish.
         </td></tr>
-        <tr><td></td></tr>
+        <%
+            if (FileRootProp.siteDefault.name().equals(currentOption) || FileRootProp.folderOverride.name().equals(currentOption))
+            {
+                Long fileRootSize = getContainer().getFileRootSize();
+        %>
+        <tr><td><p>Total size of the current file root: <%=h(fileRootSize != null ? FileUtil.formatFileSize(getContainer().getFileRootSize()) : "Unknown")%></p></td></tr>
```
labkey-jeckels: Might be useful to show when this was last refreshed.

core/resources/schemas/core.xml
```diff
@@ -35,6 +35,8 @@
       <column columnName="Searchable"/>
       <column columnName="LockState"/>
       <column columnName="ExpirationDate"/>
+      <column columnName="FileRootSize"/>
+      <column columnName="LastCrawled"/>
```
labkey-jeckels: This could be confused with crawling for full-text search indexing. Worth a comment or rename?

filecontent/src/org/labkey/filecontent/FileContentModule.java
```diff
@@ -221,8 +228,21 @@ public void doStartup(ModuleContext moduleContext)
                 results.put("fileRootCrawlSucceeded", succeeded);
                 results.put("fileRootMillisecondsToCalculateSize", System.currentTimeMillis() - startTime);
             }
+
+            // During system maintenance, FileRootMaintenanceTask populates FileRootSize and LastCrawled for every container (subject to a timeout)
+            TableInfo containers = CoreSchema.getInstance().getTableInfoContainers();
+            long total = new SqlSelector(containers.getSchema(), "SELECT SUM(FileRootSize) FROM " + containers.getSelectName())
```
labkey-jeckels: These could easily be combined into a single query. Not sure if it's worth consolidating, but on some servers we have tens of thousands of containers.

filecontent/src/org/labkey/filecontent/FileRootMaintenanceTask.java
```diff
@@ -0,0 +1,98 @@
...
+    {
+        FileContentService service = FileContentService.get();
+
+        if (service != null)
+        {
+            long deadline = HeartBeat.currentTimeMillis() + DateUtils.MILLIS_PER_MINUTE * MAX_MINUTES;
+            MutableInt rootCount = new MutableInt();
+            MutableBoolean finished = new MutableBoolean(true);
+
+            new TableSelector(CoreSchema.getInstance().getTableInfoContainers(), Set.of("RowId", "EntityId", "FileRootSize"), null, new Sort("LastCrawled"))
```
labkey-jeckels: Postgres and SQLServer sort nulls differently. Probably OK, but I suppose on SQLServer we might end up never visiting some containers.

labkey-jeckels: Here's an idea for an ORDER BY that makes it predictable:

https://stackoverflow.com/questions/7621205/sort-null-values-to-the-end-of-a-table

labkey-jeckels (approved):

filecontent/src/org/labkey/filecontent/FileContentModule.java
labkey-jeckels: Yes, I think we can remove. I'll add a comment to the metric list to say these are removed in 24.11 to make it less likely to confuse people with the switch.

api/src/org/labkey/api/util/FileUtil.java
labkey-adam: Thanks... I thought I remembered seeing that somewhere. Switched.

core/src/org/labkey/core/admin/view/filesProjectSettings.jsp
labkey-adam: Pushed `FileRootLastCrawled` into `Container` and displayed it here. Also provides more detail if crawl hasn't happened or has but wasn't successful.

core/resources/schemas/core.xml
labkey-adam: Now `FileRootLastCrawled`

filecontent/src/org/labkey/filecontent/FileContentModule.java
labkey-adam: I combined total root size and the earliest crawl time into one query. I don't think I can add EXISTS(nulls) as well in an efficient way, but open to suggestions.

filecontent/src/org/labkey/filecontent/FileRootMaintenanceTask.java
labkey-adam: Thanks... rewrote using a `SqlSelector` that adds `NULLS FIRST` in the PostgreSQL case.

filecontent/src/org/labkey/filecontent/FileContentModule.java
labkey-jeckels: Something like this should work:

`SUM(CASE WHEN FileRootLastCrawled IS NULL THEN 1 ELSE 0 END) AS NotYetCrawledCount`

labkey-adam: I took a slightly different approach to capture # crawled and # not crawled, but it's one query now.

PR #5955, opened 2024-10-17
Cancel jobs on server shutdown

labkey-tchad:
#### Rationale
[Issue](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51483)
Cancel jobs on server shutdown.
Maybe we need a more global query termination on server shutdown?

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5892

#### Changes
- Cancel jobs on server shutdown


labkey-tchad:
This would be a change in behavior since we typically want to restart pipeline jobs that were running when the server shut down.

PR #5959, opened 2024-10-18
Issue 51213: Add temp tokens to prevent temp tables from disappearing before use

labkey-susanh:
#### Rationale
Issue [51213](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51213) - we are seeing a good number of exception reports referencing temp tables used during computation of ancestor data. The complaint in the exception reports is that the tables needed don't exist.  The theory is that these tables are getting cleaned up prematurely because the reference object used to create the `TempTableTracker` isn't used elsewhere and is thus seen as a candidate for garbage collection, which then will cause the temp tables to be deleted before use. We do see a lot of GC activity in our production instances, which lends credence to this theory.

This PR attaches the `ref` `Objects` used for the tracker to the relevant SQL fragments so they should stick around.

#### Related Pull Requests
- 

#### Changes
- Add some logging (may want to reduce to debug level, though)
- Exit methods earlier if we don't have any work to do
- Update `SQLFragments` with a temp token that should keep the relevant temp tables alive.

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
+++ b/experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
@@ -301,13 +301,12 @@ public class ClosureQueryHelper
         try
         {
             Object ref = new Object();
-            String tempTableName = "closinc_"+temptableNumber.incrementAndGet();
+            String tempTableName = "closinc_" + temptableNumber.incrementAndGet();
             ttt = TempTableTracker.track(tempTableName, ref);
             SQLFragment from = new SQLFragment("FROM temp." + familyTempTable).append(" WHERE ObjectType = ").appendValue(isSampleType ? "m" : "d").append(" ");
             SQLFragment selectInto = selectIntoTempTableSql(getScope().getSqlDialect(), from, tempTableName);
             int count = new SqlExecutor(getScope()).execute(selectInto);
-            if (count == 0)
-                return;
+            logger.debug("Selected {} rows into {} for recompute.", count, tempTableName);
 
             SQLFragment upsert;
             TableInfo tInfo = isSampleType ? ExperimentServiceImpl.get().getTinfoMaterialAncestors() : ExperimentServiceImpl.get().getTinfoDataAncestors();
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
```diff
@@ -301,11 +301,13 @@ private static void incrementalRecomputeFromTempTable(String familyTempTable, bo
         try
         {
             Object ref = new Object();
-            String tempTableName = "closinc_"+temptableNumber.incrementAndGet();
+            String tempTableName = "closinc_" + temptableNumber.incrementAndGet();
```
labkey-matthewb, line 304: an equivalent, but maybe slightly more elegant idea would be to change tempTableName into a SQLfragment() and call addTempToken() to that.  Then everyone who uses that name will autmatically "inherit" the token as well.

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
```diff
@@ -301,11 +301,13 @@ private static void incrementalRecomputeFromTempTable(String familyTempTable, bo
...
             ttt = TempTableTracker.track(tempTableName, ref);
             SQLFragment from = new SQLFragment("FROM temp." + familyTempTable).append(" WHERE ObjectType = ").appendValue(isSampleType ? "m" : "d").append(" ");
             SQLFragment selectInto = selectIntoTempTableSql(getScope().getSqlDialect(), from, tempTableName);
-            new SqlExecutor(getScope()).execute(selectInto);
+            selectInto.addTempToken(ref);
+            int count = new SqlExecutor(getScope()).execute(selectInto);
+            logger.info("Selected {} rows into temp.{} for recompute.", count, tempTableName);
```
labkey-matthewb: .info() logging seems aggressive to me (here and above) unless this is intended as a temporary change?

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
labkey-susanh: Done.

labkey-matthewb (approved):

experiment/src/org/labkey/experiment/api/ClosureQueryHelper.java
labkey-susanh: Yeah, I figured it would be too verbose. Now that we have a good working theory for why this happens, I'll change back to debug.

PR #5960, opened 2024-10-18
Less verbose logging

labkey-matthewb:
#### Rationale
reduce logging due to duplicate name validation
enforce unique names in class _MapTransformer

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/dataiterator/MapDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/MapDataIterator.java
@@ -86,7 +86,7 @@ public interface MapDataIterator extends DataIterator
                 _findMap.put(in.getColumnInfo(i).getName(),i);
             }
             if (null != duplicates)
-                LOGGER.warn("Data has duplicates columns: '" + StringUtils.join(duplicates.toArray(), ", ") + "'");
+                LOGGER.warn("Data has duplicate columns: '" + StringUtils.join(duplicates.toArray(), ", ") + "'");
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/dataiterator/MapDataIterator.java
```diff
@@ -74,11 +78,15 @@ public MapDataIteratorImpl(DataIterator in, boolean mutable, Set<String> skip)
...
-                    LOGGER.warn("Map already has column named '" + name + "'");
+                    if (null == duplicates)
+                        duplicates = new CaseInsensitiveTreeSet();
+                    duplicates.add(name);
                     continue;
                 }
                 _findMap.put(in.getColumnInfo(i).getName(),i);
             }
+            if (null != duplicates)
+                LOGGER.warn("Data has duplicates columns: '" + StringUtils.join(duplicates.toArray(), ", ") + "'");
```
labkey-jeckels: This change is great. Can we log anything else to help track down the target table or other context? Would the toString() in the input DataIterator have anything useful, for example?

labkey-jeckels (approved):

api/src/org/labkey/api/dataiterator/MapDataIterator.java
labkey-matthewb: That's an interesting question.  Context can be difficult here.  We have very good logging of a _running_ data iterator.  But while setting it up, the pieces don't necessarily know what's going on.  I wonder if we could augment the DataIteratorContext to provide useful info (e.g. what constructed dataiterators it knows about).

labkey-jeckels:
```suggestion
                LOGGER.warn("Data has duplicate columns: '" + StringUtils.join(duplicates.toArray(), ", ") + "'");
```

PR #5961, opened 2024-10-18
Issue 50861: Pass around properly encoded WebDav URLs

labkey-jeckels:
#### Rationale
We are inconsistently encoding URIs in Strings and passing them around. Many codepaths end up double-encoding and then selectively decoding specific characters.

We can do better by using `URI` on the server and properly encoding throughout.

#### Changes
- Switch to `URI` to make it clear what's being passed around
- Return encoded values in JSON and other responses
- Adjust file UI to expect properly encoding URLs and skip the need for most encoding/decoding operations


labkey-jeckels:
@labkey-nicka The failure in FileContent looks related to the decision to not encode single-quotes. Do we need to make the test code know the same convention?

labkey-nicka:
> @labkey-nicka The failure in FileContent looks related to the decision to not encode single-quotes. Do we need to make the test code know the same convention?

I'll take a look. We retained encoding single-quotes so I would hope they're being encoded.

**Edit:** Alright, should be fixed with https://github.com/LabKey/testAutomation/pull/2107/commits/9d963b7 which corresponds with https://github.com/LabKey/platform/pull/5961/commits/4908fc0d.

labkey-nicka (approved):

PR #5962, opened 2024-10-19
Issue 51480: Pipeline OutputLogger fails to log parameterized messages

labkey-jeckels:
#### Rationale
IntelliJ is encouraging us to switch to parameterized Log4J logging, so it's time for the pipeline job logger to catch up.

#### Changes
- Hook in at a lower level to support all variants of the various logging methods


labkey-adam (approved):
Nice!

PR #5966, opened 2024-10-21
core.Containers should have only one key field

labkey-adam:
#### Rationale
Adding `RowId` as the `core.Containers` primary key ended up giving us too many key columns (e.g., columns that wrap RowId), causing errors when attempting to delete a workbook.

Also, some recently added Container fields were appearing on the "edit workbook" page. They should not.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5952

### DIFF
--- a/api/src/org/labkey/api/data/ContainerTable.java
+++ b/api/src/org/labkey/api/data/ContainerTable.java
@@ -31,7 +31,6 @@ import org.labkey.api.query.ExprColumn;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.FilteredTable;
 import org.labkey.api.query.UserSchema;
-import org.labkey.api.query.column.BuiltInColumnTypes;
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/data/ContainerTable.java
```diff
@@ -100,6 +94,7 @@ private void init(ActionURL url)
         MutableColumnInfo col = this.wrapColumn("ID", getRealTable().getColumn("RowId"));
         col.setReadOnly(true);
         col.setURL(detailsURL);
+        col.setKeyField(false); // RowId column is a key. We don't want to double post this value on delete, etc.
```
labkey-matthewb, line 97: We do this way too often.  Another pattern is to use AliasedColumn.  That doesn't setKeyField(false) by default either.

labkey-jeckels (approved):

labkey-matthewb (approved):

PR #5969, opened 2024-10-22
Misc 24.11 Fixes

cnathe:
#### Rationale
Issue [51136](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51136): LKSM: Calculated Columns pushed to the bottom in the customize grid view
Issue [48684](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48684): Cutoff word in error message modal
Issue [51321](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51321): App sample parent issues when sample type is named "First"

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/850
- https://github.com/LabKey/premiumModules/pull/80

#### Changes
- Reorder columns in table info when calculated fields are added via `loadAllButCustomizerFromXML`
- Check for reserved names on assay design create/update: "First" and "All"
- use `Ext4.Msg.show` instead of `Ext4.Msg.alert` to prevent message clipping

### DIFF
--- a/assay/src/org/labkey/assay/actions/SaveProtocolAction.java
+++ b/assay/src/org/labkey/assay/actions/SaveProtocolAction.java
@@ -57,15 +57,7 @@ public class SaveProtocolAction extends MutatingApiAction<GWTProtocol>
         }
 
         if (protocol.getName() != null)
-        {
-            String protocolName = protocol.getName().trim();
-
-            // Issue 51321: check reserved sample type name: First
-            if ("First".equalsIgnoreCase(protocolName) || "All".equalsIgnoreCase(protocolName))
-                throw new IllegalArgumentException("Assay design name '" + protocolName + "' is reserved.");
-
-            protocol.setName(protocolName);
-        }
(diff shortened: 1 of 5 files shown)


assay/src/org/labkey/assay/actions/SaveProtocolAction.java
```diff
@@ -57,7 +57,15 @@ public Object execute(GWTProtocol protocol, BindException errors) throws Excepti
         }
 
         if (protocol.getName() != null)
-            protocol.setName(protocol.getName().trim());
+        {
+            String protocolName = protocol.getName().trim();
+
```
XingY: Does it make sense to add this check in `AssayDomainServiceImpl.saveChanges` instead?

XingY (approved):

assay/src/org/labkey/assay/actions/SaveProtocolAction.java
cnathe: yes, good call. moved

PR #5974, opened 2024-10-22
More plate metrics

labkey-alan:
#### Rationale
There are some additional metrics that we want to track

#### Related Pull Requests
- n/a

#### Changes
- PlateSetMetricService
  - Don't count archived or templates in existing metrics
  - Add archived plate sets count
  - Add archived plates count
  - Add plate sets with no plates count
  - Add plate sets with one plate count
  - Add maximum plates count
  - Add plate template count
  - Add archived plate template count
  - Add plate sets with hits count
- PlateController
  - Add metrics for plate export
  - Add metric for worklist export

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateMetricsProvider.java
+++ b/assay/src/org/labkey/assay/plate/PlateMetricsProvider.java
@@ -1,6 +1,5 @@
 package org.labkey.assay.plate;
 
-import org.checkerframework.checker.units.qual.A;
 import org.labkey.api.assay.AssayProtocolSchema;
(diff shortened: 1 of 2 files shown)


assay/src/org/labkey/assay/plate/PlateMetricsProvider.java
```diff
@@ -1,5 +1,6 @@
 package org.labkey.assay.plate;
 
+import org.checkerframework.checker.units.qual.A;
```
labkey-nicka: Remove this import.

labkey-nicka (approved):

PR #5979, opened 2024-10-23
Issue 51488: wiki "link" macro does extra HTML encoding

labkey-jeckels:
#### Rationale
The Radeox refactor got too aggressive at HTML encoding.

#### Changes
- Revert the newly introduced calls to `PageFlowUtil.filter()`

### DIFF
--- a/core/src/org/radeox/macro/LinkMacro.java
+++ b/core/src/org/radeox/macro/LinkMacro.java
@@ -74,15 +74,15 @@ public class LinkMacro extends BaseLocaleMacro {
         if (url != null && text != null) {
             writer.write("<span class=\"nobr\"");
             if(style != null)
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-chrisj @labkey-tchad @labkey-jony @labkey-susanh This fix will be helpful for labkey.org. That's currently (and usually) running a snapshot build. My thought is to merge this to the 24.10 branch and get a fresh snapshot build deployed (no need for a maintenance release). I'll then manually merge it to develop for 24.11 and beyond.

If you don't like this plan for any reason, please let me know.

labkey-adam (approved):
Would feel better if these params were `HtmlString`s, but yeah, I know. Maybe a comment here that `MacroFilter` (I think) ensures all params are HTML filtered?

PR #5980, opened 2024-10-23
Revert PK add to core.Containers

labkey-adam:
#### Rationale
My attempt to designate `core.Containers.RowId` as this table's PK failed. Too many code paths assume `EntityId` is the (pseudo) PK... for example, lookups to the Containers table. Best to revert it for now. Better would be to designate `EntityId` as an actual PK, but it's a nullable column so this is not a trivial change.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5952


labkey-jeckels (approved):
:(

labkey-matthewb (approved):

PR #5982, opened 2024-10-24
Folders: Archive V1 (Hide from view/show hidden)

XingY:
#### Rationale
We now support marking a child folder as "archived" in the apps. Archived folders in app will be excluded from certain actions in app and also will have special visual indicators. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/185
* https://github.com/LabKey/labkey-ui-components/pull/1618
* https://github.com/LabKey/labkey-ui-premium/pull/573
* https://github.com/LabKey/platform/pull/5982
* https://github.com/LabKey/limsModules/pull/858

#### Changes
- Add new Container.LockState.Archived
- util to setting a container to archived / no archived
- add metrics for archived containers


XingY:
Tagging @labkey-adam for review in case there is a concern with re-purposing container.lockstate for marking a container as archived.

labkey-adam:
> Tagging @labkey-adam for review in case there is a concern with re-purposing container.lockstate for marking a container as archived.

The existing "lock states" apply only to projects. Does "archived" apply only to folders? If so, then I don't see a conflict or concern.

XingY:
> > Tagging @labkey-adam for review in case there is a concern with re-purposing container.lockstate for marking a container as archived.
> 
> The existing "lock states" apply only to projects. Does "archived" apply only to folders? If so, then I don't see a conflict or concern.

Yes, currently we only support archiving a child folder. Thanks for the feedback.

api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -881,6 +881,33 @@ public static void setExcludedProjects(Collection<GUID> ids, @NotNull Runnable a
         auditRunnable.run();
     }
 
+    public static void archiveContainer(User user, Container container, boolean archive)
+    {
+        if (container.isRoot() || container.isAppHomeFolder())
```
labkey-susanh: Should this also check `container.isProject()` so as not to interfere with the `locked` state?

XingY: good point. updated

api/src/org/labkey/api/data/Container.java
```diff
@@ -206,19 +206,22 @@ public void propertyChange(PropertyChangeEvent evt)
     // Might add others in the future (e.g., ReadOnly)
     public enum LockState
     {
-        Unlocked("unlocked", false, false),
-        Inaccessible("locked, making it inaccessible to everyone except administrators", true, false),
-        Excluded("excluded from project locking and review", false, true);
+        Unlocked("unlocked", false, false, false),
+        Archived("archived", false, false, true),
+        Inaccessible("locked, making it inaccessible to everyone except administrators", true, false, false),
+        Excluded("excluded from project locking and review", false, true, false);
```
labkey-adam: Might be helpful to add comments to indicate whether each state applies to folders or projects

labkey-susanh (approved):

PR #5987, opened 2024-10-24
Fix "days active" query on French SQL Server

labkey-adam:
#### Rationale
[Issue 51472: SQL exception when calculating days active on French SQL Server](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51472)

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5805


labkey-jeckels (approved):
We should convert `toWhereClause()` to return a `SQLFragment`. Perhaps not today.

PR #5989, opened 2024-10-25
Issue 51521: Add switch to disable Server HTTP header

labkey-jeckels:
#### Rationale
Some organizations don't like having the product and version identified in HTTP headers. We can make it optional.

#### Changes
- New site setting that lets admins opt-in or -out of the `Server` HTTP response header
- Move the setting of the header into `AuthFilter`, alongside other header setting, to be consistent across all responses
- Also fix SQL for new plate metrics


labkey-jeckels:
@labkey-alan FYI on the plate metrics SQL. I was getting SQLExceptions due to the lack of an alias on the subquery

labkey-adam (approved):

PR #5994, opened 2024-10-26
Invalidate sessions on password change

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51523

### DIFF
--- a/api/src/org/labkey/api/security/UserManager.java
+++ b/api/src/org/labkey/api/security/UserManager.java
@@ -590,7 +590,8 @@ public class UserManager
     {
         // Return true to increment the count passed to complete() after handleSession() has been invoked for each active session
         boolean handleSession(HttpSession session);
-        // Called after all sessions have been passed to handleSession(), only if user is non-null and was logged in
+        // Called after all sessions have been passed to handleSession(), only if user is non-null and was logged in.
+        // Useful for coalescing logging, for example.
         void complete(int count);
     }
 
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/security/UserManager.java
```diff
@@ -545,25 +552,72 @@ public void sessionCreated(HttpSessionEvent event)
...
+            {
+                //noinspection DataFlowIssue
+                LOG.debug("Invalidated {} for {}.", StringUtilsLabKey.pluralize(count, "session"), user.getEmail());
+            }
+        });
+    }
+
+    public interface SessionHandler
+    {
+        // Return true to increment the count passed to complete()
```
labkey-jeckels:
```suggestion
        // Return true to increment the count passed to complete() after handleSession() has been invoked for each active session
```

api/src/org/labkey/api/security/UserManager.java
```diff
@@ -545,25 +552,72 @@ public void sessionCreated(HttpSessionEvent event)
...
+        boolean handleSession(HttpSession session);
+        // Called after all sessions have been passed to handleSession(), only if user is non-null and was logged in
```
labkey-jeckels:
```suggestion
        // Called after all sessions have been passed to handleSession(), only if user is non-null and was logged in.
        // Useful for coalescing logging, for example.
```

labkey-jeckels (approved):

labkey-jeckels:
It took me a little while to understand the `SessionHandler` interface so I suggested some additional comments. See if you think they're helpful or just clutter.

labkey-adam:
> It took me a little while to understand the `SessionHandler` interface so I suggested some additional comments. See if you think they're helpful or just clutter.

Yeah, it's not the most obvious interface ever. More docs don't hurt, IMO.

PR #6000, opened 2024-10-28
Allow adding alternate keys to dataiterator

labkey-martyp:
#### Rationale
Add adder and update getter for alt keys.

#### Related Pull Requests
- https://github.com/LabKey/dataintegration/pull/280
- https://github.com/LabKey/ehrModules/pull/854
- https://github.com/LabKey/snd/pull/242

#### Changes
- Change getter to return unmodifiable set.
- Add addAlternateKeys

### DIFF
--- a/api/src/org/labkey/api/dataiterator/DataIteratorContext.java
+++ b/api/src/org/labkey/api/dataiterator/DataIteratorContext.java
@@ -22,6 +22,7 @@ import org.labkey.api.collections.CaseInsensitiveHashSet;
 import org.labkey.api.query.BatchValidationException;
 import org.labkey.api.query.QueryUpdateService;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/dataiterator/DataIteratorContext.java
```diff
@@ -227,7 +228,12 @@ public Set<String> getDontUpdateColumnNames()
     @NotNull
     public Set<String> getAlternateKeys()
     {
-        return _alternateKeys;
+         return  Collections.unmodifiableSet(_alternateKeys);
+    }
+
+    public void addAlternateKeys(Set<String> alternateKeys)
```
labkey-matthewb, line 234: I might have gone with set() instead of add(), but this is much better.

labkey-matthewb (approved):

api/src/org/labkey/api/dataiterator/DataIteratorContext.java
labkey-martyp: yeah I thought about that, just noticed all the usages were just adding to the existing set so just gave that level of access

PR #6004, opened 2024-10-29
avoid logging exception in ensureNotClosed()

labkey-matthewb:
#### Rationale
This should avoid the exception in the client abort case.

#### Related Pull Requests
- 

#### Changes
-


labkey-nicka (approved):
Clever.

PR #6007, opened 2024-10-30
Issue 51305: insertRows for samples doesn't trim before parsing integers

labkey-jeckels:
#### Rationale
I made an earlier fix in ConvertHelper which works for most table types. But not for samples, and perhaps others.

#### Changes
- Trim in `JdbcType` which does the the conversion for some sample type fields


labkey-alan (approved):
Looks good to me, and works when I manually tested it

PR #6013, opened 2024-11-01
Don't fail when attempting to load sample data for non-local modules

labkey-tchad:
#### Rationale
More server-side tests are now running on TeamCity without the module source present. `JUnitUtil.getSampleData` should be more tolerant of this and not fail tests when module source is not present.

#### Related Pull Requests
- N/A

#### Changes
- Skip tests with missing sampledata when module source isn't present


labkey-tchad:
Will update this PR to target 24.11 once the branches are ready

labkey-sweta (approved):

labkey-tchad:
Verifying on TC: https://teamcity.labkey.org/buildConfiguration/LabKey_2411_Community_DrtSqlserver/3254635

PR #6014, opened 2024-11-01
Issue 51563: LKSM: Add additional date and date/time formats

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1629
- https://github.com/LabKey/labkey-ui-premium/pull/585
- https://github.com/LabKey/limsModules/pull/879
- https://github.com/LabKey/platform/pull/6014

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/util/DateUtil.java
+++ b/api/src/org/labkey/api/util/DateUtil.java
@@ -118,32 +118,27 @@ public class DateUtil
 
     // Splits the date and time portions of a standard date-time format, where date portion is required and time portion
     // is optional. Returns null if the format is non-standard, which means we can't split it.
-    public static DateTimeFormat splitDateTimeFormat(String dateTimeFormat)
+    public static DateTimeFormat splitDateTimeFormat(@NotNull String dateTimeFormat)
     {
         dateTimeFormat = dateTimeFormat.trim();
-        String datePortion = null;
+        String timePortion = null;
 
         // We can't just split on whitespace because non-standard formats could: have any amount of whitespace between
         // characters, put the time portion before the date portion, or even intermingle date and time characters.
-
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -125,8 +130,11 @@ public static DateTimeFormat splitDateTimeFormat(String dateTimeFormat)
         {
             if (dateTimeFormat.startsWith(format))
             {
-                datePortion = format;
-                break;
+                if (dateTimeFormat.equals(format) || dateTimeFormat.charAt(format.length()) == ' ')
```
labkey-susanh: I'm not seeing why this extra check is needed. Is there an example where we don't split correctly without it?

XingY: This check is added so "ddMMMyyyy" won't be parsed as ""ddMMMyy" and "yy"

labkey-susanh: Ah, I see. Might be worth a comment?

labkey-susanh (approved):

api/src/org/labkey/api/util/DateUtil.java
labkey-adam: Should we change the logic to look for dateTimeFormat.endsWith(<a standard time format>), clip off the standard time format if it exists, then trim the remainder & check STANDARD_DATE_FORMATS.contains(remainder)? There are far fewer time formats now and this may be cleaner.

labkey-adam (approved):

api/src/org/labkey/api/util/DateUtil.java
XingY: Changes made to check time format first.

PR #6016, opened 2024-11-04
CloudWebDavProvider

labkey-matthewb:
#### Rationale
Reimplement cloud webdav provider over new FileSystemBlobStore (vfs), which in turn is implemented more directly over the jclouds.Blobstore.  Reducing dependencies on S3FS and implement new caching scheme.

#### Related Pull Requests
- 

#### Changes
copy
rename file
disable rename collection (for now at least)
Path case sensitive/insensitive support

### DIFF
--- a/api/src/org/labkey/api/data/TempTableTracker.java
+++ b/api/src/org/labkey/api/data/TempTableTracker.java
@@ -163,11 +163,13 @@ public class TempTableTracker extends WeakReference<Object>
 
         synchronized(createdTableNames)
         {
-            createdTableNames.remove(qualifiedName);
+            var ttt = createdTableNames.remove(qualifiedName);
             appendToLog("-" + schemaName + "\t" + tableName + "\n");
 
             if (createdTableNames.isEmpty() || System.currentTimeMillis() > lastSync + CacheManager.DAY)
                 synchronizeLog(false);
+
+            ttt.clear();
         }
     }
 
--- a/api/src/org/labkey/api/webdav/FileSystemResource.java
+++ b/api/src/org/labkey/api/webdav/FileSystemResource.java
@@ -60,6 +60,7 @@ import org.labkey.api.writer.ContainerUser;
 import org.labkey.api.writer.DefaultContainerUser;
 import org.labkey.vfs.FileLike;
 import org.labkey.vfs.FileSystemLike;
+import org.springframework.util.StreamUtils;
 
 import java.io.File;
 import java.io.FileInputStream;
@@ -125,7 +126,7 @@ public class FileSystemResource extends AbstractWebdavResource
         setSecurableResource(folder.getSecurableResource());
 
         _files = folder._files.stream()
-            .map(file -> file.resolveChild(name.toString()))
+            .map(file -> file.resolveChild(name))
             .toList();
     }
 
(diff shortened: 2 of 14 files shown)


api/src/org/labkey/api/webdav/FileSystemResource.java
```diff
@@ -259,14 +260,32 @@ public String getAbsolutePath(User user)
...
     @Override
     public long copyFrom(User user, FileStream is) throws IOException
     {
-        File file = getFile();
+        FileLike file = getFileLike();
         boolean created = false;
         if (!file.exists())
         {
-            FileUtil.mkdirs(file.getParentFile(), AppProps.getInstance().isInvalidFilenameUploadBlocked());
+            var parent = file.getParent();
```
Sigmonia: should this be inlined?

api/src/org/labkey/api/data/TempTableTracker.java
```diff
@@ -163,11 +163,13 @@ private void untrack()
...
         synchronized(createdTableNames)
         {
-            createdTableNames.remove(qualifiedName);
+            var ttt = createdTableNames.remove(qualifiedName);
             appendToLog("-" + schemaName + "\t" + tableName + "\n");
 
             if (createdTableNames.isEmpty() || System.currentTimeMillis() > lastSync + CacheManager.DAY)
                 synchronizeLog(false);
+
+            ttt.clear();
```
Sigmonia, line 172: is this needed?

api/src/org/labkey/api/webdav/FileSystemResource.java
```diff
@@ -277,18 +296,26 @@ public long copyFrom(User user, FileStream is) throws IOException
...
         }
 
         try
         {
-            is.transferTo(file);
-            if (is.getLastModified() != null)
+            File ioFile = getFile();
+            if (null != ioFile)
+            {
+                is.transferTo(getFile());
```
Sigmonia: should this be `ioFile`?

api/src/org/labkey/vfs/AbstractFileSystemLike.java
```diff
@@ -1,23 +1,26 @@
...
 import java.net.URI;
 import java.nio.file.Path;
 
 abstract public class AbstractFileSystemLike implements FileSystemLike
 {
     final URI uri;
     final String scheme;
-    final String strUri;    // no trailing '/'
+    final boolean caseSensitive;
+    final String strUri; // no trailing '/'
```
Sigmonia: Is the comment still accurate? re: line 23

api/src/org/labkey/vfs/FileSystemLike.java
```diff
@@ -52,6 +52,13 @@
  */
 public interface FileSystemLike
 {
+    // NOTE: a full webdav path consist of case-sensitive and case-sensitive parts
```
Sigmonia: is one of those supposed to be `case-insensitive`?

Sigmonia (approved):
Again, mostly minor nits

api/src/org/labkey/api/data/TempTableTracker.java
labkey-matthewb: OK, this is unrelated.  If we don't clear the weak reference, we will get called later at GC time to delete a table we've already taken care of.

api/src/org/labkey/api/webdav/FileSystemResource.java
labkey-matthewb: yes

PR #6021, opened 2024-11-05
Stop passing around email addresses

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51520

#### Changes
* Add `UserId` column to `core.Logins` and populate it from the `core.Principals` table. Make this the primary key.
* Consolidate all core.Logins-handling methods from `UserManager` & `SecurityManager` into `LoginManager`. Migrate these methods and their callers to use UserId instead of email address.
* Update actions to pass and use `userId` parameter
* Update verification URLs to remove email parameter and not add userId. Verification is now a lookup based solely on the verification token.
* Make `Email` column nullable and stop populating it. It'll get removed in 24.12.

### DIFF
--- a/api/src/org/labkey/api/security/LoginManager.java
+++ b/api/src/org/labkey/api/security/LoginManager.java
@@ -1,0 +1,297 @@
+package org.labkey.api.security;
+
+import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.Nullable;
+import org.labkey.api.data.Container;
+import org.labkey.api.data.CoreSchema;
+import org.labkey.api.data.DbScope;
+import org.labkey.api.data.SQLFragment;
+import org.labkey.api.data.Selector;
+import org.labkey.api.data.SqlExecutor;
+import org.labkey.api.data.SqlSelector;
+import org.labkey.api.util.PageFlowUtil;
+import org.labkey.api.util.Pair;
+import org.labkey.api.view.ActionURL;
+import org.springframework.dao.DataIntegrityViolationException;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+
+// Handles all CRUD operations on the core.Logins table
+public class LoginManager
+{
+    public static final int TEMP_PASSWORD_LENGTH = 32;
+    private static final int MAX_HISTORY = 10;
+    private static final CoreSchema CORE = CoreSchema.getInstance();
+    private static final String PASSWORD_CHARS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+
+    // Create record for database login, saving UserId and hashed password. Return verification token.
+    public static String createLogin(int userId, ValidEmail email /* Just for logging errors */) throws SecurityManager.UserManagementException
+    {
+        // Create a placeholder password hash and a separate email verification key that will get emailed to the new user
+        String tempPassword = createTempPassword();
+        String verification = createTempPassword();
+
+        String crypt = Crypt.MD5.digestWithPrefix(tempPassword);
+
+        try
+        {
+            // TODO: Stop inserting Email after removing that column
+            // Don't need to set LastChanged -- it defaults to current date/time.
+            int rowCount = new SqlExecutor(CORE.getSchema()).execute("INSERT INTO " + CORE.getTableInfoLogins() +
+                    " (UserId, Email, Crypt, LastChanged, Verification, PreviousCrypts) VALUES (?, ?, ?, ?, ?, ?)",
+                    userId, email.getEmailAddress(), crypt, new Date(), verification, crypt);
+            if (1 != rowCount)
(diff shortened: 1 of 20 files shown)


core/resources/schemas/dbscripts/postgresql/core-24.007-24.008.sql
```diff
@@ -0,0 +1,9 @@
+-- Shift the core.Logins PK from Email to UserId. Add UserId column as NULLABLE, populate it from core.Principals,
+-- delete rows that didn't join (RowId IS NULL), make RowId NOT NULL, drop the old PK, and add the new PK.
```
labkey-jeckels:
```suggestion
-- delete rows that didn't join (UserId IS NULL), make UserId NOT NULL, drop the old PK, and add the new PK.
```

core/resources/schemas/dbscripts/sqlserver/core-24.007-24.008.sql
```diff
@@ -0,0 +1,13 @@
+-- Shift the core.Logins PK from Email to UserId. Add UserId column as NULLABLE, populate it from core.Principals,
+-- delete rows that didn't join (RowId IS NULL), make RowId NOT NULL, drop the old PK, and add the new PK.
```
labkey-jeckels:
```suggestion
-- delete rows that didn't join (UserId IS NULL), make UserId NOT NULL, drop the old PK, and add the new PK.
```

api/src/org/labkey/api/security/LoginManager.java
```diff
@@ -0,0 +1,257 @@
...
+    {
+        // Create a placeholder password hash and a separate email verification key that will get emailed to the new user
+        String tempPassword = createTempPassword();
+        String verification = createTempPassword();
+
+        String crypt = Crypt.MD5.digestWithPrefix(tempPassword);
+
+        try
+        {
+            // TODO: Stop inserting Email after removing that column
```
labkey-jeckels: Are you planning to drop the column in this PR?

api/src/org/labkey/api/security/LoginManager.java
```diff
@@ -0,0 +1,257 @@
...
+        else
+            return Crypt.MD5.matches(password, hash);
+    }
+
+    public static String createTempPassword()
+    {
+        StringBuilder tempPassword = new StringBuilder(TEMP_PASSWORD_LENGTH);
+
+        for (int i = 0; i < TEMP_PASSWORD_LENGTH; i++)
+            tempPassword.append(PASSWORD_CHARS.charAt((int) Math.floor((Math.random() * PASSWORD_CHARS.length()))));
```
labkey-jeckels: We should be using `SecureRandom`.

api/src/org/labkey/api/security/LoginManager.java
```diff
@@ -0,0 +1,257 @@
...
+        verificationUrl.addParameter("verification", verification);
+        verificationUrl.addParameter("userId", user.getUserId());
+
+        if (null != extraParameters)
+            verificationUrl.addParameters(extraParameters);
+
+        return verificationUrl;
+    }
+
+    // Test if user has been verified for database authentication. Use only when email address could be invalid ().
```
labkey-jeckels:
```suggestion
    // Test if user has been verified for database authentication. Use only when email address could be invalid.
```

core/src/org/labkey/core/login/LoginController.java
```diff
@@ -1769,47 +1757,34 @@ public ModelAndView getSuccessView(SetPasswordForm form)
         }
     }
 
-    private static boolean attemptVerification(SetPasswordForm form, ValidEmail email, Errors errors)
+    //
```
labkey-jeckels: Did you intend to add a comment here?

labkey-jeckels (commented):
Looks like a handful of test failures.

core/src/org/labkey/core/login/LoginController.java
labkey-adam: Probably...

api/src/org/labkey/api/security/LoginManager.java
labkey-adam: I have a follow-on commit that makes the column nullable and stops inserting email addresses into new rows, but I'm keeping the column for now, just in case. I'll drop it in 24.12.

labkey-adam:
> Looks like a handful of test failures.

Yep, working on those

api/src/org/labkey/api/security/LoginManager.java
labkey-adam: Second half of comment no longer applies, so I removed it

core/src/org/labkey/core/login/LoginController.java
labkey-adam: Comment added

api/src/org/labkey/api/security/LoginManager.java
labkey-adam: Replaced this with `RandomStringUtils.secureStrong().nextAlphanumeric(TEMP_PASSWORD_LENGTH);`

labkey-jeckels (approved):

PR #6034, opened 2024-11-07
remote static files

labkey-matthewb:
#### Rationale
re-enable remote serving of static files (eval) given CORS and CSP

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/settings/AppPropsImpl.java
+++ b/api/src/org/labkey/api/settings/AppPropsImpl.java
@@ -564,14 +564,17 @@ class AppPropsImpl extends AbstractWriteableSettingsGroup implements AppProps
         return lookupStringValue(XFrameOption, "SAMEORIGIN");
     }
 
-    private String staticFilesPrefix = "";
+
+    private static final String not_init = "";
+    private String staticFilesPrefix = not_init;
 
     @Override
     public String getStaticFilesPrefix()
     {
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/settings/AppPropsImpl.java
```diff
@@ -563,13 +564,32 @@ public String getXFrameOption()
...
+                try
+                {
+                    var url = new URLHelper(s).setPath("");
+                    if (StringUtils.isNotEmpty(url.getHost()))
+                        ContentSecurityPolicyFilter.registerAllowedConnectionSource("static.files.prefix", url.toString());
+                    staticFilesPrefix = s;
+                }
+                catch (URISyntaxException e)
+                {
+                    staticFilesPrefix = null;
```
labkey-jeckels, line 591: Won't this doom us to NPEs every time the method is called in the future? Some sort of logging about the bogus value would be useful too.

labkey-jeckels (changes requested):

api/src/org/labkey/api/settings/AppPropsImpl.java
labkey-matthewb: It's fine.  I cleverly made sure that this method is only called exactly once.

labkey-jeckels (approved):

PR #6041, opened 2024-11-10
Improve session tracking and invalidating

labkey-adam:
#### Rationale
* https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51638
* https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51637

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5994

#### Changes
* Fix session tracking to respect impersonation. Sessions are always tracked to the original user, not an impersonated user.
* On API key authentication, stash the API key's RowId in session. Use that to skip invalidation of API key sessions on password change and invalidate associated API key sessions on API key delete.
* Simplify creation of `AuthenticationResponse`

### DIFF
--- a/api/src/org/labkey/api/security/SecurityManager.java
+++ b/api/src/org/labkey/api/security/SecurityManager.java
@@ -180,7 +180,8 @@ public class SecurityManager
     private static final String AUTHENTICATION_METHOD = "SecurityManager.authenticationMethod";
 
     public static final String PRIMARY_AUTHENTICATION_CONFIGURATION = PrimaryAuthenticationConfiguration.class.getName();
-    public static final String AUTHENTICATION_ATTRIBUTES_KEY = User.class.getName() + "$AuthenticationAttributes";
+    public static final String USER_ATTRIBUTES_KEY = User.class.getName() + "$UserAttributes";
+    public static final String AUTHENTICATION_PROPERTIES = AuthenticationManager.class.getName() + "$AuthenticationProperties";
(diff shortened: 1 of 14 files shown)


labkey-jeckels:
@labkey-adam there are test failures related to invalid sessions. Should I review now or after they're passing?

labkey-adam:
> @labkey-adam there are test failures related to invalid sessions. Should I review now or after they're passing?

Short answer: feel free to wait. I haven't been able to reproduce these failures locally and plan to investigate (e.g., bumping up the logging temporarily). Fairly sure these are race conditions where the tests rapidly delete multiple api keys associated with the same user and/or log out / invalidate a session in parallel with deletes, but I want to verify this because previous failures like this pointed out issues with the fundamental bookkeeping. The answer is likely tolerating an invalidate() that throws, checking isValid() (which is not a public method, but seems to be on the Tomcat impl), or adding synchronization.

api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -521,6 +532,7 @@ public static User getSessionUser(HttpServletRequest request)
         return sessionUser;
     }
 
+    // If currently impersonating, returns the impersonated user with impersonation context set
```
labkey-jeckels, line 535: If you make these JavaDoc style comments they'll show up in more places in the IDE

labkey-jeckels (approved):

PR #6042, opened 2024-11-11
51629: On a windows system, transform scripts with backslashes should be handled better

labkey-klum:
[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51629)

#### Rationale
Transformed result data values that contain backslashes are `unescaped` during parsing on the subsequent import. It doesn't seem like this was a regression and it is possible that this scenario never worked. There was an addition to the test as part of this change:

https://github.com/LabKey/platform/pull/5781

That may have exposed the problem and didn't start failing until we recently started running that test on a windows agent.

### DIFF
--- a/api/src/org/labkey/api/qc/TsvDataSerializer.java
+++ b/api/src/org/labkey/api/qc/TsvDataSerializer.java
@@ -122,7 +122,14 @@ public class TsvDataSerializer implements DataExchangeHandler.DataSerializer
                 else if (Object[].class.isAssignableFrom(o.getClass()))
                     pw.append(StringUtils.join((Object[]) o, ","));
                 else
-                    pw.append(String.valueOf(o));
+                {
+                    String val = String.valueOf(o);
+                    // double quote the value if it contains backslashes to avoid tab loader mangling
+                    // on import
+                    if (val.contains("\\"))
+                        val = "\"" + val + "\"";
+                    pw.append(val);
+                }
             }
             sep = "\t";
         }
--- a/api/src/org/labkey/api/qc/TsvDataSerializer.java
+++ b/api/src/org/labkey/api/qc/TsvDataSerializer.java
@@ -127,7 +127,7 @@ public class TsvDataSerializer implements DataExchangeHandler.DataSerializer
                     // Issue 51629 double quote the value if it contains backslashes to avoid tab loader mangling
                     // on import
                     if (!val.startsWith("\"") && !val.endsWith("\"") && val.contains("\\"))
-                        val = "\"" + val + "\"";
+                        val = "\"" + StringUtils.replace(val,"\"", "\"\"") + "\"";
                     pw.append(val);
                 }
             }
--- a/api/src/org/labkey/api/qc/TsvDataSerializer.java
+++ b/api/src/org/labkey/api/qc/TsvDataSerializer.java
@@ -124,9 +124,9 @@ public class TsvDataSerializer implements DataExchangeHandler.DataSerializer
                 else
                 {
                     String val = String.valueOf(o);
-                    // double quote the value if it contains backslashes to avoid tab loader mangling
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/qc/TsvDataSerializer.java
```diff
@@ -122,7 +122,14 @@ else if (Collection.class.isAssignableFrom(o.getClass()))
                 else if (Object[].class.isAssignableFrom(o.getClass()))
                     pw.append(StringUtils.join((Object[]) o, ","));
                 else
-                    pw.append(String.valueOf(o));
+                {
+                    String val = String.valueOf(o);
+                    // double quote the value if it contains backslashes to avoid tab loader mangling
+                    // on import
+                    if (val.contains("\\"))
```
XingY: Maybe check `if (val.startsWith("\"") && val.endsWith("\"") && val.contains("\\"))` to avoid double quoting?

XingY (approved):

api/src/org/labkey/api/qc/TsvDataSerializer.java
labkey-matthewb: TSVWriter.quoteValue()

labkey-matthewb (changes requested):

api/src/org/labkey/api/qc/TsvDataSerializer.java
XingY: Does `TSVWriter.quoteValue` quote backslashes?

labkey-matthewb: Good question.  It does handle double-quoting " which is necessary if you're going to append " at the beginning and end.  I guess the bigger observation, is that we know that the format generated by TsvWriter and TabLoader should be compatible and as compatible as possible with Excel, R, etc, etc.  

I'm honestly not sure why we have the option setUnescapeBackslashes().  I suspect it was for a particular assay import scenario.  

That's why this is more of a spec issue than a code bug.  We just need to decide what format we want to round-trip and be consistent.

labkey-klum: `TSVWriter.quoteValue()` for some reason doesn't quote backslashes. 

I've updated my small change to incorporate Xing's feedback. I'm thinking that for 24.11 I'll check in this narrow change and then reopen the tracking issue to make some of these larger changes in `develop`:
- Consolidate quoting code into `TSVWriter`
- Default `TabLoader._unescapeBackslashes` to false
- Add unit test coverage to `TSVWriter` to verify uniform serialization/deserialization

labkey-matthewb: I'm fine with this as long as we double-check that we haven't traded breaking values with \ for breaking values with ".  I think the parser expected " within quoted strings to be doubled?

api/src/org/labkey/api/qc/TsvDataSerializer.java
```diff
@@ -122,7 +122,14 @@ else if (Collection.class.isAssignableFrom(o.getClass()))
                 else if (Object[].class.isAssignableFrom(o.getClass()))
                     pw.append(StringUtils.join((Object[]) o, ","));
                 else
-                    pw.append(String.valueOf(o));
+                {
+                    String val = String.valueOf(o);
+                    // Issue 51629 double quote the value if it contains backslashes to avoid tab loader mangling
+                    // on import
+                    if (!val.startsWith("\"") && !val.endsWith("\"") && val.contains("\\"))
+                        val = "\"" + val + "\"";
```
labkey-matthewb: "\\"" + StringUtils.replace(val,"\\"", "\\"\\""), + "\\""

labkey-matthewb (approved):
Consider replacing single quote with double quote to be compatible with TabLoader#481

PR #6043, opened 2024-11-11
Add the Ability to filter file uploads based on file extensions

Sigmonia:
#### Rationale
[Secure Issue 51524: Allowlist for acceptable types of file upload](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51524)

#### Related PRs
* https://github.com/LabKey/premiumModules/pull/100

#### Changes
- updated shared jsps
- Add allow list
- and extension checker to file uploads

### DIFF
--- a/api/src/org/labkey/api/util/FileUtil.java
+++ b/api/src/org/labkey/api/util/FileUtil.java
@@ -1535,7 +1535,7 @@ quickScan:
 
         String result = new String(ret);
 
-        assert isAllowedFileName(result) == null :
+        assert !AppProps.getWriteableInstance().isInvalidFilenameBlocked() || isAllowedFileName(result) == null :
                 "Failed to make filename safe. Original: " + name + ", transformed: " + result + ", error: " + isAllowedFileName(result);
 
         return new String(ret);
--- a/api/src/org/labkey/api/util/FileUtil.java
+++ b/api/src/org/labkey/api/util/FileUtil.java
@@ -311,20 +311,16 @@ public class FileUtil
 
     public static String isAllowedFileName(String s, boolean checkFileExtension)
     {
-        if (AppProps.getInstance().isInvalidFilenameBlocked())
+        return isAllowedFileName(s, checkFileExtension, AppProps.getInstance());
+    }
+
+    static String isAllowedFileName(String s, boolean checkFileExtension, AppProps appProps)
+    {
+        if (appProps.isInvalidFilenameBlocked())
         {
-            if (StringUtils.isBlank(s))
-                return "Filename must not be blank";
-            if (!ViewServlet.validChars(s))
-                return "Filename must contain only valid unicode characters.";
-            if (StringUtils.containsAny(s, restrictedPrintable))
-                return "Filename may not contain any of these characters: " + restrictedPrintable;
-            if (StringUtils.containsAny(s, "\t\n\r"))
-                return "Filename may not contain 'tab', 'new line', or 'return' characters.";
-            if (StringUtils.contains("-$", s.charAt(0)))
-                return "Filename may not begin with any of these characters: -$";
-            if (Pattern.matches("(.*\\s--[^ ].*)|(.*\\s-[^- ].*)", s))
-                return "Filename may not contain space followed by dash.";
+            String msg = validateFileName(s);
+            if (msg != null)
+                return msg;
         }
 
         if (checkFileExtension)
@@ -333,6 +329,24 @@ public class FileUtil
             if (badExtension != null)
                 return "This file type [" + badExtension + "] is not allowed.";
         }
+        return null;
+    }
+
+    private static @Nullable String validateFileName(String s)
+    {
+        if (StringUtils.isBlank(s))
+            return "Filename must not be blank";
+        if (!ViewServlet.validChars(s))
+            return "Filename must contain only valid unicode characters.";
+        if (StringUtils.containsAny(s, restrictedPrintable))
+            return "Filename may not contain any of these characters: " + restrictedPrintable;
+        if (StringUtils.containsAny(s, "\t\n\r"))
+            return "Filename may not contain 'tab', 'new line', or 'return' characters.";
+        if (StringUtils.contains("-$", s.charAt(0)))
+            return "Filename may not begin with any of these characters: -$";
+        if (Pattern.matches("(.*\\s--[^ ].*)|(.*\\s-[^- ].*)", s))
+            return "Filename may not contain space followed by dash.";
+
         return null;
     }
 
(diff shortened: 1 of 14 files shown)


api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -316,9 +318,33 @@ public static String isAllowedFileName(String s)
...
+            setExtensionChecker();
+
+        String extension = FilenameUtils.getExtension(filename);
+        return extensionChecker.matcher(filename).matches() ? null : extension;
+    }
+
+    public static void setExtensionChecker()
+    {
+        // Regex encode the allowed extensions (escape periods and add '|' optional matcher)
+        String allowedExtensions = String.join("|", AppProps.getInstance().getAllowedExtensions()).replace(".", "\\.");
```
labkey-jeckels: What about other special regex characters, like `-` or `(`?

This seems like a good candidate for a small unit test.

core/src/org/labkey/core/admin/ExternalServerType.java
```diff
@@ -83,8 +94,60 @@ public void setHosts(Collection<String> hosts)
...
+        }
+    },
+    FileExtension {
+        @Override
+        public HtmlString getDescription()
+        {
+            return HtmlString.unsafe("""
+                <div style="width: 700px">
+                    <div>
+                        This list is the set of file extensions that LabKey will allow through. Any extension that is not in this list will be rejected, this includes mulitple extensions. If the list is blank, then this check is ignored.
```
labkey-jeckels:
```suggestion
                        This list is the set of file extensions that LabKey will accept for uploads. Any extension that is not in this list will be rejected, this includes multiple extensions. For example, .gz is not sufficient to allow .tar.gz; you must specify .tar.gz. If the list is blank, then this check is ignored.
```

api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -316,9 +318,33 @@ public static String isAllowedFileName(String s)
             return "Filename may not begin with any of these characters: -$";
         if (Pattern.matches("(.*\\s--[^ ].*)|(.*\\s-[^- ].*)",s))
             return "Filename may not contain space followed by dash.";
+
+        String badExtension = checkExtension(s);
```
labkey-jeckels: This method is only invoked if the server has been configured to block these patterns, via a separate property. It's also only consulted for WebDav uploads.

The customer who raised this concern would want it to apply to all uploads, including attachments to issues, wikis, etc.

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -449,6 +449,8 @@ public static void registerAdminConsoleLinks()
         AdminConsole.addLink(Configuration, "system maintenance", new ActionURL(ConfigureSystemMaintenanceAction.class, root));
         AdminConsole.addLink(Configuration, "external redirect hosts", new ActionURL(ExternalHostsAdminAction.class, root).addParameter("type", ExternalServerType.Redirect.name()), TroubleshooterPermission.class);
         AdminConsole.addLink(Configuration, "external allowed sources", new ActionURL(ExternalHostsAdminAction.class, root).addParameter("type", ExternalServerType.Source.name()), TroubleshooterPermission.class);
+        if (ModuleLoader.getInstance().hasModule("Premium"))
```
labkey-jeckels: The combo of checking for the premium module and the class/enum names on the next line is too much for me.

We could clean it up by making this a non-premium feature. I'm fine with that given the implementation approach.

api/src/org/labkey/api/util/FileUtil.java
labkey-jeckels: I think that means also checking in AntiVirusMultipartResolver.

labkey-jeckels (changes requested):

core/src/org/labkey/core/admin/ExternalServerType.java
Sigmonia: updated

api/src/org/labkey/api/util/FileUtil.java
Sigmonia: test added, using Pattern.quote()

api/src/org/labkey/api/util/FileUtil.java
Sigmonia: Added it to the AntiVirusMultipartResolver. Its unclear to me how to best add this call to the Spring `StandardServletMultipartResolver`

api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -2221,5 +2253,62 @@ public void testAllowedFileName()
...
+            assertNull("Extension should be allowed", checkExtension("test.txt", mockProps));
+            assertNull("Multiple extension should be allowed", checkExtension("archive.tar.gz", mockProps));
+            assertNull("Case-insensitive extension should be allowed", checkExtension("archive.TaR.Gz", mockProps));
+            assertNull("Special characters aren't escaped properly", checkExtension("my test.l-()[]{}1☃", mockProps));
+            assertNull("Numeric extension should be allowed", checkExtension("test.1", mockProps));
+            assertNotNull("Multiple extension matched when it shouldn't", checkExtension("tar.gz", mockProps));
+            assertNotNull("Matched unlist extension", checkExtension("my test.notListed", mockProps));
+            assertNotNull("Combined multiple extension matched incorrectly", checkExtension("multi.a_v.tar", mockProps));
+            assertNotNull("Multi-multi extension matched unexpectedly", checkExtension("multi.not.tar.gz", mockProps));
+            assertNotNull("No extension matched unexpectedly", checkExtension("No extension", mockProps));
```
labkey-jeckels, line 2316: I hadn't considered files with no extension. I guess it's OK to reject them all at this point.

core/src/org/labkey/core/admin/ExternalServerType.java
```diff
@@ -83,8 +100,72 @@ public void setHosts(Collection<String> hosts)
...
+        }
+    },
+    FileExtension {
+        @Override
+        public HtmlString getDescription()
+        {
+            return HtmlString.unsafe("""
+                <div style="width: 700px">
+                    <div>
+                        This list is the set of file extensions that LabKey will accept for uploads. Any extension that is not in this list will be rejected, this includes multiple extensions. For example, .gz is not sufficient to all .tar.gz; you must specify .tar.gz. If the list is empty, then this check is ignored.
```
labkey-jeckels: Typo in my suggestion: "is not sufficient to all" should be "is not sufficient to allow"

api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -316,9 +324,33 @@ public static String isAllowedFileName(String s)
             return "Filename may not begin with any of these characters: -$";
```
labkey-jeckels, line 346: We should pull the check of `AppProps.getInstance().isInvalidFilenameUploadBlocked()` out of `DavController.checkAllowedFileName()` and put it here around the prexisting set of checks. Otherwise, we're ignoring that setting.

core/src/org/labkey/core/webdav/DavController.java
```diff
@@ -3074,7 +3074,10 @@ public void closeInputStream() throws IOException
                     return lastModified;
                 }
             };
-
+            if (FileUtil.isAllowedFileName(name) != null)
+            {
+                throw new IOException("The file extension is not allowed.");
```
labkey-jeckels: This may reject the name unrelated to the exception. Let's use the message this call returns instead of hard-coding the error message

api/src/org/labkey/api/util/FileUtil.java
labkey-jeckels: Looks like we'll need another subclass of `StandardServletMultipartResolver` that lives in the API module. The default implementation in AntiVirusProviderRegistry should return that new subclass, which looks at the file name. We can refactor `AntiVirusMultipartResolver` to inherit from the new subclass and reduce duplicated code.

labkey-jeckels (changes requested):

labkey-jeckels (approved):

api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -304,21 +312,48 @@ public static void copyDirectory(Path srcPath, Path destPath) throws IOException
...
         return null;
     }
 
+    private static String checkExtension(String filename, AppProps appProps)
+    {
+        // If the allow list is empty, allow any extension
+        if (appProps.getAllowedExtensions().isEmpty())
+            return null;
+
+        if (extensionChecker == null)
```
labkey-matthewb, line 359: When does this get cleared?  It seems very sticky.

core/src/org/labkey/core/webdav/DavController.java
```diff
@@ -3075,6 +3075,11 @@ public void closeInputStream() throws IOException
                 }
             };
 
+            String notAllowedMsg = FileUtil.isAllowedFileName(name);
+            if (StringUtils.isNotBlank(notAllowedMsg))
+            {
+                throw new IOException(notAllowedMsg);
```
labkey-matthewb: throw DavException in  the dav controller to get the file browser to display an expected UI.  Maybe use 409 SC_CONFLICT?  or 406 (SC_NOT_ACCEPTABLE)

labkey-matthewb:
Do we want to allow folder names to have "." in them?  Disallowing doesn't seem to be related to the original goal of the feature.

labkey-matthewb (changes requested):

api/src/org/labkey/api/util/FileUtil.java
Sigmonia: It only gets changed in setExtensionChecker which is called when the WritableAppProps.setAllowedFileExtensions is called when the new Allow List prop is saved or when the server is starting up

labkey-matthewb: I see.  The pattern confused me.  I would have expected a getExtensionCheck() that wrapped the null check.

labkey-matthewb (approved):

PR #6052, opened 2024-11-13
renderIconAndFilename() expects actual path

labkey-matthewb:
#### Rationale
renderIconAndFilename() expects actual path
belt and suspenders fix for !f.exists()

#### Related Pull Requests
- 

#### Changes
-


labkey-jeckels (approved):
Code change looks OK but there appear to be tests expecting different results.

PR #6053, opened 2024-11-13
CalculateCurveFitAction and support for 3PL curve fit

cnathe:
#### Rationale
We will be adding support for including a curve fit line in the chart wizard. This set of PRs are the first step: adding an API to request a curve fit for a set of data points. The API uses our existing curve fit types along with the new 3PL fit type. The response includes the curve fit parameters for the requested fit type, the stats for the fit, and generated data points to use for plotting (optional).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6053
- https://github.com/LabKey/premiumModules/pull/103

#### Changes
- ParameterCurveFit to support constants for min and max asymptote
- Add Sigmoidal Logistic 3 Parameter curve fit
- StatsServiceImpl.TestCase update for 3PL curve fit and 4PL simplex fit
- Add curve fit stats for new API: residual sum of squares, root mean square error, R2, adjusted R2

### DIFF
--- a/api/src/org/labkey/api/data/statistics/CurveFit.java
+++ b/api/src/org/labkey/api/data/statistics/CurveFit.java
@@ -146,4 +146,54 @@ public interface CurveFit<P extends CurveFit.Parameters>
      * @return The integrated area under the curve.
      */
     double calculateAUC(StatsService.AUCType type, double startX, double endX) throws FitFailedException;
+
+    default double residualSumSquares(P parameters)
+    {
+        double sumSq = 0;
+        for (DoublePoint point : getData())
+        {
+            double expectedValue = point.getY();
+            double foundValue = fitCurve(point.getX(), parameters);
+
+            sumSq += Math.pow(foundValue - expectedValue, 2);
+        }
+        return sumSq;
+    }
+
+    default double rootMeanSquareError(P parameters)
+    {
+        return Math.sqrt(residualSumSquares(parameters) / getData().length);
+    }
+
+    default double totalSumSquares()
+    {
+        double sumSq = 0;
+        double mean = 0;
+        for (DoublePoint point : getData())
+            mean += point.getY();
+        mean /= getData().length;
+
+        for (DoublePoint point : getData())
+        {
+            double expectedValue = point.getY();
+            sumSq += Math.pow(expectedValue - mean, 2);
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/data/statistics/CurveFit.java
```diff
@@ -146,4 +146,54 @@ default JSONObject toJSON()
      * @return The integrated area under the curve.
      */
     double calculateAUC(StatsService.AUCType type, double startX, double endX) throws FitFailedException;
+
```
labkey-klum, line 149: A description of what some of these methods are might be helpful as the titles alone aren't entirely clear.

core/src/org/labkey/core/statistics/StatsServiceImpl.java
```diff
@@ -56,14 +58,24 @@ public MathStat getStats(Collection<Double> data)
...
     public CurveFit getCurveFit(CurveFitType type, DoublePoint[] data)
+    {
+        return getCurveFit(type, data, null, null);
+    }
+
+    @Override
+    public CurveFit getCurveFit(CurveFitType type, DoublePoint[] data, @Nullable Double asymptoteMin, @Nullable Double asymptoteMax)
     {
         switch (type)
         {
```
labkey-klum, line 69: Is the plan to keep both 3 Parameter variants, or are we waiting for feedback from the client? Unless there is a good reason to keep both, it feels like having a single fit function might be less confusing.

labkey-klum (approved):

api/src/org/labkey/api/data/statistics/CurveFit.java
cnathe: done

core/src/org/labkey/core/statistics/StatsServiceImpl.java
cnathe: I'm not entirely sure yet. I was thinking of keeping them both around for now (similar to how we currently have 2 variants of 4PL curve fits). Then when we get to the actually plotting story to implement the selection of the curve fit type, we can make the final call at that time and either: 1) only show a subset of the options in the plotting UI or 2) remove the 3PL option we don't want.

PR #6057, opened 2024-11-15
Assay: resolve a single provider/protocol before creating protocol schema

labkey-nicka:
#### Rationale
When attempting to resolve a `TableInfo` from a `Domain` for the `AssayDomainKind` we currently iterate over all the assay definitions in a container and for each provider we call `provider.createProtocolSchema()` and then hydrate full `TableInfo` instances until we find a match.

This refactors the approach to first resolve the singularly desired provider and only induce the `provider.createProtocolSchema()` call for that provider to then resolve the `TableInfo`.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1645
- https://github.com/LabKey/labkey-ui-premium/pull/599
- https://github.com/LabKey/platform/pull/6057
- https://github.com/LabKey/limsModules/pull/919
- https://github.com/LabKey/commonAssays/pull/822

#### Changes
- Refactor away `AssayManager.getTableInfoForDomainId()` by directly resolving the assay protocol/provider in `AssayDomainKind.getTableInfo()` and then instantiating a new schema for only that assay.
- Separate `AbstractAssayProvider.getDomains()` into `getDomains()` and `getDomainsAndDefaultValues()` to make `getDomains()` more clear in its role and what it will return (i.e. a `List<Domain>` rather than a `List<Pair<Domain, Map<DomainProperty, Object>>>`)
- Update `assay-assayList.api` matching on protocol "name" and provider "type" to be case-insensitive.
- Add null checks

### DIFF
--- a/assay/api-src/org/labkey/api/assay/AssayDomainKind.java
+++ b/assay/api-src/org/labkey/api/assay/AssayDomainKind.java
@@ -33,7 +33,6 @@ import org.labkey.api.exp.api.ExpProtocol;
 import org.labkey.api.exp.api.ExperimentService;
 import org.labkey.api.exp.property.BaseAbstractDomainKind;
 import org.labkey.api.exp.property.Domain;
-import org.labkey.api.exp.property.DomainProperty;
 import org.labkey.api.exp.property.PropertyService;
 import org.labkey.api.gwt.client.DefaultValueType;
 import org.labkey.api.gwt.client.model.GWTDomain;
@@ -50,14 +49,8 @@ import org.labkey.api.view.ActionURL;
 import org.labkey.api.writer.ContainerUser;
 
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 
-/**
- * User: brittp
- * Date: June 25, 2007
- * Time: 1:01:43 PM
- */
 public abstract class AssayDomainKind extends BaseAbstractDomainKind
 {
     private final String _namespacePrefix;
@@ -142,7 +135,7 @@ public abstract class AssayDomainKind extends BaseAbstractDomainKind
         return new SQLFragment("NULL");
     }
 
-    protected ExpProtocol findProtocol(Domain domain)
+    protected @Nullable ExpProtocol findProtocol(Domain domain)
     {
         Pair<AssayProvider, ExpProtocol> pair = findProviderAndProtocol(domain);
         if (pair == null)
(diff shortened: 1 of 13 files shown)


assay/api-src/org/labkey/api/assay/AssayDomainKind.java
```diff
@@ -243,7 +234,33 @@ public boolean showDefaultValueSettings()
...
+            return null;
+
+        AssayProvider provider = providerAndProtocol.first;
+        ExpProtocol protocol = providerAndProtocol.second;
+
+        AssayProtocolSchema schema = provider.createProtocolSchema(user, container, protocol, null);
+        for (var tableName : schema.getTableNames())
+        {
+            // First, fetch a table that is more likely to be cached
+            var table = schema.getTable(tableName);
```
cnathe, line 251: is there any chance that this TableInfo object here would be "different" in what metadata or cf is applies vs the one below from line 258? I recall an issue a few month ago where the issue was that at one point in the call stack we were getting a TableInfo without the metadata when it wanted it applied.

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
```diff
@@ -571,16 +566,9 @@ public GWTProtocol saveChanges(GWTProtocol assay, boolean replaceIfExisting) thr
 
                 for (GWTDomain<GWTPropertyDescriptor> domain : assay.getDomains())
                 {
-                    GWTDomain<GWTPropertyDescriptor> previous = DomainUtil.getDomainDescriptor(getUser(), domain.getDomainURI(), protocol.getContainer());
+                    GWTDomain<GWTPropertyDescriptor> previous = updateDomainDescriptor(domain, protocol, provider, hasNameChange);
```
cnathe: `previous` is used in the line below to check for existing calculated fields. This line now sets previous to the new/updated domain with changes, right? Seems like we'd still want the call to getDomainDescriptor for previous.

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
```diff
@@ -623,7 +611,11 @@ private ValidationException updateDomainDescriptor(GWTDomain<GWTPropertyDescript
             auditComment = "The name of the assay domain '" + previous.getName() + "' was changed to '" + domain.getName() + "'.";
         }
 
-        return DomainUtil.updateDomainDescriptor(previous, domain, getContainer(), getUser(), hasNameChange, auditComment);
+        ValidationException domainErrors = DomainUtil.updateDomainDescriptor(previous, domain, getContainer(), getUser(), hasNameChange, auditComment);
+        if (domainErrors.hasErrors())
+            throw domainErrors;
+
+        return previous;
```
cnathe: oh, I see. regarding my last comment. this is actually returning the previous GWTDomain object. This was confusing to me since I would expect that the updateDomainDescriptor would return the updated version if anything.

cnathe (approved):

assay/api-src/org/labkey/api/assay/AssayDomainKind.java
labkey-nicka: Good question. It will almost certainly be "different" in each case. What I'm trying to take advantage of here is the underlying tableInfo caching which is why I request the more expensive version of the table only once I've resolved it.

cnathe: ah I see, thanks for clarifying. I missed the part that this wasn't the returned table object

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
labkey-nicka: Yea, this was a silly thing to do. I've updated the pattern so the function now takes the `previous` descriptor as an argument and returns `void`.

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
labkey-nicka: See https://github.com/LabKey/platform/pull/6057#discussion_r1843926757.

PR #6058, opened 2024-11-15
remove call to FileUtil.stringToPath()

labkey-matthewb:
#### Rationale
FileUtil.stringToPath() is returning null on s3: urls.  I'm not sure why (because they don't have account name?).
This breaks extended properties in the file browser.

I'm not sure why we don't pass the dataFileUrl from exp.data as is? 

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/filecontent/src/org/labkey/filecontent/FileContentController.java
+++ b/filecontent/src/org/labkey/filecontent/FileContentController.java
@@ -1260,8 +1260,7 @@ public class FileContentController extends SpringActionController
                 Map<FieldKey, Object> data = result.getFieldKeyRowMap();
                 Map<String, Object> row = new HashMap<>();
 
-                java.nio.file.Path dataFilePath = FileUtil.stringToPath(getContainer(), (String) data.get(dataFileURLFieldKey));
-                row.put("dataFileUrl", null != dataFilePath ? FileUtil.pathToString(dataFilePath) : null);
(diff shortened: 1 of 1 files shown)


labkey-tchad:
Verifying with S3: https://teamcity.labkey.org/buildConfiguration/LabKey_2411_Premium_ModulesSuites_CloudPostgres/3263074?hideTestsFromDependencies=false&hideProblemsFromDependencies=false

filecontent/src/org/labkey/filecontent/FileContentController.java
```diff
@@ -1260,8 +1260,7 @@ public ApiResponse execute(CustomPropertiesForm form, BindException errors)
                 Map<FieldKey, Object> data = result.getFieldKeyRowMap();
                 Map<String, Object> row = new HashMap<>();
 
-                java.nio.file.Path dataFilePath = FileUtil.stringToPath(getContainer(), (String) data.get(dataFileURLFieldKey));
```
Sigmonia, line 1263: So there were a few reasons for this. Mainly was as a workaround for mangled S3 urls that had been incorrectly relativized, which resulted in the protocol incorrectly substituted in the middle of the URI. I think if thats been dealt with upstream (which I believe to be the case) it should be safe to remove this. 

The other reason was to remove the access token from the path so it wouldn't get leaked into the log.

Sigmonia (approved):

PR #6059, opened 2024-11-15
Support long database names in PostgreSQL 17.x

labkey-adam:
#### Rationale
PostgreSQL used to accept database names longer than 63 characters, truncating them for all operations. As of 17.x, it allows creation but fails to connect when the database name is too long. Given the (intentional) inconsistent behavior and misleading exception on connection attempt, LabKey now throws a configuration exception if a too-long database name is detected. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51676

Also, change `getIdentifierMaxLength()` to report the database's actual limit. Previously, the dialect reported two or three characters less than the actual limit (it was inconsistent) in an attempt to help callers that might be adding suffixes. Now the dialect just reports the facts; callers are responsible for respecting the database-imposed limit.

And avoid an unnecessary 10-second sleep between creating the database and first connection. We need the timeout only if the initial connection failed.

### DIFF
--- a/api/src/org/labkey/api/query/AliasManager.java
+++ b/api/src/org/labkey/api/query/AliasManager.java
@@ -208,7 +208,8 @@ public class AliasManager
 
     private static int getMaxLength(@Nullable SqlDialect dialect, boolean useLegacyMaxLength)
     {
-        // we use 28 here because Oracle has a limit or 30 characters, and that is likely the shortest restriction
+        // we use 28 here because Oracle has a limit of 30 characters, and that is likely the shortest restriction
+
         // But note: Oracle 12c raised the limit to 128 characters, so perhaps increase the fall-back length now?
         return useLegacyMaxLength ? 40 : (dialect == null ? 28 : dialect.getIdentifierMaxLength() - 3 /* leave room for possible suffixes */);
     }
--- a/api/src/org/labkey/api/query/AliasManager.java
+++ b/api/src/org/labkey/api/query/AliasManager.java
@@ -178,8 +178,7 @@ public class AliasManager
             ret = "X_";
         if (dialect != null)
             ret = dialect.makeLegalIdentifierName(ret);
-        // we use 28 here because Oracle has a limit or 30 characters, and that is likely the shortest restriction
-        int maxLength = useLegacyMaxLength ? 40 : (dialect == null ? 28 : dialect.getIdentifierMaxLength());
+        int maxLength = getMaxLength(dialect, useLegacyMaxLength);
         if (reserveCount > 0)
             maxLength -= reserveCount;
         if (maxLength < 5)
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/query/AliasManager.java
```diff
@@ -202,12 +201,17 @@ public static String makeLegalName(FieldKey key, @Nullable SqlDialect dialect, b
...
-        // we use 28 here because Oracle has a limit or 30 characters, and that is likely the shortest restriction
-        var ret = truncate(sb.toString(), useLegacyMaxLength ? 40 : (dialect == null ? 28 : dialect.getIdentifierMaxLength()));
+        var ret = truncate(sb.toString(), getMaxLength(dialect, useLegacyMaxLength));
         assert isLegalName(ret);
         return ret;
     }
 
+    private static int getMaxLength(@Nullable SqlDialect dialect, boolean useLegacyMaxLength)
+    {
+        // we use 28 here because Oracle has a limit or 30 characters, and that is likely the shortest restriction
```
labkey-jeckels:
```suggestion
        // we use 28 here because Oracle has a limit of 30 characters, and that is likely the shortest restriction

```

labkey-jeckels (commented):
I'm OK with this change, though it seems wiser to stop generating ones that are so long.

labkey-jeckels (approved):

labkey-adam:
> I'm OK with this change, though it seems wiser to stop generating ones that are so long.

@labkey-tchad ?

labkey-tchad:
> > I'm OK with this change, though it seems wiser to stop generating ones that are so long.
> 
> @labkey-tchad ?

Agreed; it seems bad to just silently truncate to something other than what the configuration specifies. I presume the Postgres folks made the same conclusion.
I can configure TeamCity to override the database name provided by the gradle plugin. It isn't really necessary to use the build ID, that was just something easy and relatively unique.

labkey-adam:
> > > I'm OK with this change, though it seems wiser to stop generating ones that are so long.
> > 
> > 
> > @labkey-tchad ?
> 
> Agreed; it seems bad to just silently truncate to something other than what the configuration specifies. I presume the Postgres folks made the same conclusion. I can configure TeamCity to override the database name provided by the gradle plugin. It isn't really necessary to use the build ID, that was just something easy and relatively unique.

I filed an issue with the PostgreSQL team given the long-standing truncation behavior, the now inconsistent behavior (CREATE and ALTER DATABASE work fine, but connecting fails), and the lack of any mention in the release notes... while fully acknowledging that we were relying on undocumented behavior and we can easily work around this. Once TeamCity is using shorter database names I can remove the code that truncates and futzes with the URL in the DataSource. The rest of this PR (having the dialect report the actual max length instead of guessing what callers might do) seems worthwhile, so I'll merge it eventually.

@labkey-tchad The current truncation code should allow us to start testing on PostgreSQL 17.x without adjusting those names. Can you enable PG 17 on this branch?

labkey-tchad:
> @labkey-tchad The current truncation code should allow us to start testing on PostgreSQL 17.x without adjusting those names. Can you enable PG 17 on this branch?

I've updated the database name on 24.11 and develop and switched them both to postgres 17

labkey-tchad:
> > @labkey-tchad The current truncation code should allow us to start testing on PostgreSQL 17.x without adjusting those names. Can you enable PG 17 on this branch?
> 
> I've updated the database name on 24.11 and develop and switched them both to postgres 17

Correction. On Friday, I updated the names and switched 24.11 to postgres 17.
Develop is still on 16, pending review of the weekend's 24.11 runs on pg 17.

labkey-tchad (approved):

PR #6060, opened 2024-11-16
Result level exclusions for plate based assays

labkey-klum:
#### Rationale
Adds initial support to exclude wells from certain operations for plate enabled assays. Assay results will include a `state` field which can be set to a value in the `core.DataState` table. The data state will determine what operations are allowed for values in that row. Initially only a single `excluded` data state will be supported but others can be added later. Also, currently the excluded state will prevent the well from participating in any replicate statistical calculations.

Data states will be scoped to the project level. An upgrade script was added to update the result domain for any plate enabled assays in a LKB folder.

#### Related Pull Requests
https://github.com/LabKey/limsModules/pull/925

### DIFF
--- a/api/src/org/labkey/api/assay/plate/PlateDataStateManager.java
+++ b/api/src/org/labkey/api/assay/plate/PlateDataStateManager.java
@@ -1,0 +1,154 @@
+package org.labkey.api.assay.plate;
+
+import org.jetbrains.annotations.Nullable;
+import org.labkey.api.data.Container;
+import org.labkey.api.qc.AbstractManageDataStatesForm;
+import org.labkey.api.qc.DataState;
+import org.labkey.api.qc.DataStateHandler;
+import org.labkey.api.qc.DataStateManager;
+import org.labkey.api.security.User;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class PlateDataStateManager implements DataStateHandler
+{
+    public static final String HANDLER_NAME = "PlateDataStateHandler";
+    private static PlateDataStateManager _instance = new PlateDataStateManager();
+
+    /**
+     * Maps to DataStates for plate related operations. Initially, there is only a single data state
+     * supported, but plans are to add more over time.
+     */
+    public enum StateType
+    {
+        Excluded("Excludes data rows from certain operations.",
+                Collections.emptySet());
+
+        final Set<DataOperation> _permittedOps;
+        final String _description;
+
+        StateType(String description, Set<DataOperation> permittedOps)
+        {
+            _description = description;
+            _permittedOps = permittedOps;
+        }
+
+        public Set<DataOperation> getPermittedOps()
+        {
+            return _permittedOps;
+        }
+
+        @Nullable
+        static StateType getType(String typeName)
+        {
+            for (StateType type : StateType.values())
+            {
+                if (type.name().equals(typeName))
+                    return type;
+            }
+            return null;
+        }
+    }
+
+    /**
+     * The operation that might be performed for an associated DataState
+     */
+    public enum DataOperation
+    {
+        analysis("included for any data or statistical analysis, curve fitting etc");
+
+        final String _description;
+
+        DataOperation(String description)
+        {
+            _description = description;
+        }
+    }
+
+    private PlateDataStateManager(){}
+
+    public static PlateDataStateManager get()
+    {
+        return _instance;
+    }
+
+    public boolean isOperationPermitted(@Nullable DataState state, DataOperation operation)
+    {
+        if (state == null)
+            return true;
+
+        StateType stateType = StateType.getType(state.getStateType());
+        return stateType != null && stateType.getPermittedOps().contains(operation);
+    }
+
+    public void ensureDefaultStates(Container container, User user)
+    {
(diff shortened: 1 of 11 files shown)


api/src/org/labkey/api/assay/plate/PlateDataStateManager.java
```diff
@@ -0,0 +1,169 @@
...
+            return null;
+        }
+    }
+
+    /**
+     * The operation that might be performed for an associated DataState
+     */
+    public enum DataOperation
+    {
+        analysis("included for any data or statistical analysis, curve fitting etc");
```
cnathe, line 62: nice, I like that this is a more general DataOperation that can be checked rather than just having the AssayPlateMetadataServiceImpl check for the "Excluded" state directly.

api/src/org/labkey/api/assay/plate/PlateDataStateManager.java
```diff
@@ -0,0 +1,169 @@
...
+        return DataStateManager.getInstance().getStates(container).stream()
+                .filter(state -> StateType.getType(state.getStateType()) != null)
+                .collect(Collectors.toList());
+    }
+
+    @Override
+    public boolean isStateInUse(Container container, DataState state)
+    {
+        // for now, we won't allow removal of the default plate data states
+        return true;
```
cnathe: this works to prevent delete of this state. the error message is a little odd in that I didn't actually have any usages of the state yet, but I think that is fine.

api/src/org/labkey/api/assay/plate/PlateDataStateManager.java
```diff
@@ -0,0 +1,169 @@
...
+        return _instance;
+    }
+
+    public boolean isOperationPermitted(@Nullable DataState state, DataOperation operation)
+    {
+        if (state == null)
+            return true;
+
+        StateType stateType = StateType.getType(state.getStateType());
+        return stateType != null && stateType.getPermittedOps().contains(operation);
```
cnathe, line 85: as mentioned in the manual testing "issues found", this looks to be excluding rows from replicate calculations if they have a non-Excluded state (i.e. Available, Locked). Is that expected / wanted behavior?

cnathe (approved):

api/src/org/labkey/api/assay/plate/PlateDataStateManager.java
cnathe: see manual test plan. looks like this is preventing any core.DataState from being deleted (even those from sample statuses for example).

cnathe (changes requested):

api/src/org/labkey/api/assay/plate/PlateDataStateManager.java
labkey-klum: I think for now since we will be limiting what data state can be added to the assay row for the near term. This maybe speaks to adding some validation on insert / update regarding what is a valid state?

api/src/org/labkey/api/assay/plate/PlateDataStateManager.java
labkey-klum: I'll update, it probably just needs to be conditionalized to be one of the states we care about.

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -1045,6 +1054,64 @@ private void insertOrUpdateReplicateStats(
...
+    public static void validateRowDataStates(Container container, Map<String, Object> row, DomainProperty stateProp) throws ValidationException
+    {
+        try
+        {
+            DataState state = getStateFromRow(container, row, stateProp);
+            if (state != null)
+            {
+                if (PlateDataStateManager.StateType.getType(state.getStateType()) == null)
+                {
+                    throw new ValidationException(String.format("The data state : %s is not valid for this assay", state.getStateType()));
```
cnathe: I think we want to use `state.getLabel()` in the error message instead of `state.getStateType()`.

Also, the format of the error message seems odd to me with the colon in it. I'd suggest: 
```suggestion
                    throw new ValidationException(String.format("The data state '%s' is not valid for this assay.", state.getLabel()));
```

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -1045,6 +1054,64 @@ private void insertOrUpdateReplicateStats(
...
+            importAlias.addAll(stateProp.getImportAliasSet());
+            for (Map.Entry<String, Object> entry : row.entrySet())
+            {
+                if (importAlias.contains(entry.getKey()))
+                {
+                    if (entry.getValue() instanceof Integer stateRowId)
+                    {
+                        DataState state = PlateDataStateManager.get().getStateForRowId(container, stateRowId);
+                        if (state == null)
+                            throw new ExperimentException(String.format("No data states for the rowID : %d was found", stateRowId));
```
cnathe: as mentioned in the other comment, I'd suggest removing the colon from the error message and adding a period.
```suggestion
                            throw new ExperimentException(String.format("No data states for the rowID %d was found.", stateRowId));
```

cnathe (approved):
A couple of minor comments, but otherwise the fixes for the issues mentioned in the test plan check out. Looks good.

PR #6061, opened 2024-11-16
Timeout for long-running read-only HTTP requests

labkey-jeckels:
#### Rationale
It's useless to keep cranking away on read-only HTTP requests if the client has disconnected, as they'll never be able to see the results. We can't easily detect the connection dropping in the general case, but we can do our best to kill the DB connections and spawned processes after a timeout. We can configure the timeout to be just slightly longer than the ALB or other upstream timeouts.

#### Changes
- New Site Setting to configure the timeout
- Timer thread to kill DB connections and processes
- If the thread isn't actively using a DB connection when the timeout fires, prevent its next attempt to get a DB connection
- Reshuffle responsibilities between `ViewServlet`, `ContextListener`, and `TransactionFilter` to be a little more consistent
- Lots of minor code cleanup, including better use of generics


api/src/org/labkey/api/util/AbortedRequestException.java
```diff
@@ -0,0 +1,13 @@
+package org.labkey.api.util;
+
+/**
+ * Signals that we're trying to abort processing. Examples include a HTTP request that's exceeded a timeout or a
```
labkey-adam:
```suggestion
 * Signals that we're trying to abort processing. Examples include an HTTP request that's exceeded a timeout or a
```

api/src/org/labkey/api/view/ViewServlet.java
```diff
@@ -333,13 +289,13 @@ private static void initializeAllSpringControllers()
...
         {
-            for (Class controllerClass : module.getControllerClassToName().keySet())
+            for (Class<?> controllerClass : module.getControllerClassToName().keySet())
             {
                 if (Controller.class.isAssignableFrom(controllerClass))
                 {
                     try
                     {
-                        getController(module, controllerClass);
+                        getController(module, controllerClass.asSubclass(Controller.class));
```
labkey-adam, line 298: I didn't know `asSubclass()` was a thing. Seems handy...

labkey-adam (approved):

PR #6068, opened 2024-11-19
Upgrade to Angus Mail 2.0.3

### LINKED ISSUE #6072: [LabKey CI] Upgrade Jakarta Mail --> Angus Mail
Angus Mail is the direct successor to Jakarta Mail (which is the successor to JavaMail). We want to switch to Angus Mail because that's where all maintenance and improvements are happening going forward.

- [x] Update all dependency declarations from Jakarta Mail implementation to Angus Mail
- [x] Switch to a non-JNDI approach for initializing the mail Session
- [x] Fix problems with EHR notifications (notification emails not rendering as HTML)
- [x] Code review @labkey-susanh @labkey-jeckels 
- [x] TC and merge
- [x] Notify Systems Engineering team about this change

labkey-adam:
#### Rationale
Angus Mail is the active development project, successor to Jakarta Mail

#### Related Pull Requests
- https://github.com/LabKey/server/pull/929

### DIFF
--- a/api/src/org/labkey/api/util/MailHelper.java
+++ b/api/src/org/labkey/api/util/MailHelper.java
@@ -379,7 +379,7 @@ public class MailHelper
             for (int i = 0; i < multipart.getCount(); i++)
             {
                 BodyPart part = multipart.getBodyPart(i);
-                multipartHandler.handle(StringUtils.substringBefore(part.getContentType(), ";"), part);
+                multipartHandler.handle(StringUtils.substringBefore(part.getDataHandler().getContentType(), ";"), part);
             }
         }
         else
--- a/api/src/org/labkey/api/util/MailHelper.java
+++ b/api/src/org/labkey/api/util/MailHelper.java
@@ -31,6 +31,7 @@ import jakarta.mail.internet.InternetAddress;
 import jakarta.mail.internet.MimeBodyPart;
 import jakarta.mail.internet.MimeMessage;
 import jakarta.mail.internet.MimeMultipart;
+import jakarta.servlet.ServletContext;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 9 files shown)


api/src/org/labkey/api/util/MailHelper.java
```diff
@@ -114,31 +115,47 @@ public void handle(Collection<StartupPropertyEntry> entries)
...
+                        @Override
+                        protected PasswordAuthentication getPasswordAuthentication()
+                        {
+                            return new PasswordAuthentication(username, password);
+                        }
+                    });
+                }
             }
+// I couldn't get this to work once I switched to Angus Mail; kept getting this exception:
+// java.util.ServiceConfigurationError: jakarta.mail.util.StreamProvider: org.eclipse.angus.mail.util.MailStreamProvider not a subtype
```
labkey-susanh: Is it not finding `MailStreamProvider` or is it, in fact, not a subtype?  If it seems like it is a subtype, perhaps there's some kind of classpath problem, but is this needed? Will it be checking something different from the `if (properties.isEmpty())` block above?

labkey-adam: It seems to be finding `MailStreamProvider` which is a subtype. So yes, I agree it's likely a classpath problem. I tried quite a few combinations of Angus Mail + Jakarta Mail API + Angus Activation + Jakarta Activation API in both embedded and our API (none in embedded, all in both, etc.) to no avail. The new approach avoids JNDI and Tomcat's involvement, which seems to be an important difference and works as expected. As a non-trivial upgrade, it seemed best to get a working combination committed and tested. We can certainly revisit the JNDI approach in this or a follow-on PR.

labkey-jeckels: I struggled with similar problems when migrating to Tomcat 10. They stemmed from having copies of the classes loaded by different class loaders, as you're both suspecting here. The error messages were frustrating because all the talk about is class names, which always looked correct.

If we're not relying on JNDI to wire things up, don't we need to handle all of the properties we propagate here?

https://github.com/LabKey/server/blob/93839d04e95a8f377f034b7a7094bf9e5ad9d437/server/embedded/src/org/labkey/embedded/LabKeyTomcatServletWebServerFactory.java#L379

labkey-adam: I believe the new code populates all the same properties that the old code did

labkey-jeckels: > I believe the new code populates all the same properties that the old code did

Looks like you're right. I hadn't read enough of the adjacent, untouched code.

labkey-susanh (approved):

api/src/org/labkey/api/util/MailHelper.java
labkey-adam: There's a first time for everything, I guess...

PR #6065, opened 2024-11-18
Issue 51557: Fix folder deletion summary stats type casting

labkey-nicka:
#### Rationale
This fixes a type casting issue that can appear on MSSQL instances that was introduced with #6005 as part of a fix for [Issue 51557](https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=51557).

#### Related Pull Requests
- #6005

#### Changes
- Use `Integer`/`int` instead of `Long`/`long`

PR #6070, opened 2024-11-19
Issue 51557: Avoid folder deletion summary stats type casting

labkey-nicka:
#### Rationale
This is a followup to #6065 where the fix to casting switched from `Long` to `Integer`. That worked great on SQL Server, however, caused the same problem on Postgres instances. This fix avoids calling `TableSelector.getValueMap()` all together as it provides inconsistent type instances depending on the database type.

#### Related Pull Requests
- #6005 
- #6065

#### Changes
- Refactor folder summary stats for data class and sample type data counts to use `getResultSet()` instead of `getValueMap()`.

### DIFF
--- a/experiment/src/org/labkey/experiment/ExperimentModule.java
+++ b/experiment/src/org/labkey/experiment/ExperimentModule.java
@@ -32,8 +32,6 @@ import org.labkey.api.data.CoreSchema;
 import org.labkey.api.data.DbSchema;
 import org.labkey.api.data.JdbcType;
 import org.labkey.api.data.NameGenerator;
-import org.labkey.api.data.Results;
-import org.labkey.api.data.RuntimeSQLException;
 import org.labkey.api.data.SQLFragment;
 import org.labkey.api.data.SqlSelector;
 import org.labkey.api.data.TableInfo;
@@ -152,7 +150,6 @@ import org.labkey.experiment.types.TypesController;
 import org.labkey.experiment.xar.FolderXarImporterFactory;
 import org.labkey.experiment.xar.FolderXarWriterFactory;
 
-import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


labkey-nicka:
Test suites:

- [BiologicsFolderAdminTest](https://teamcity.labkey.org/buildConfiguration/LabKey_2411_Premium_ProductSuites_Biologics_BiologicsAPostgres/3272710)
- [SMProFolderAdminFromChildFolderTest](https://teamcity.labkey.org/buildConfiguration/LabKey_2411_Premium_ProductSuites_SampleManager_SampleManagerBPostgres/3272724) (PG)
- [SMProFolderAdminFromChildFolderTest](https://teamcity.labkey.org/buildConfiguration/LabKey_2411_Premium_ProductSuites_SampleManager_SampleManagerBSqlserver/3272712) (MSSQL)

experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -889,12 +892,21 @@ public Collection<String> getSummary(Container c)
             columns.add(ExpDataClassTable.Column.Name.name());
             columns.add(ExpDataClassTable.Column.DataCount.name());
 
-            Map<String, Integer> results = new TableSelector(table, columns).getValueMap();
```
labkey-jeckels, line 892: This refactor seems fine. As an alternative approach, I've used `Map<String, Number>` in similar scenarios.

labkey-jeckels (approved):

experiment/src/org/labkey/experiment/ExperimentModule.java
labkey-nicka: That seems reasonable, however, doesn't that just push the problem to the next statement?

labkey-jeckels: You can call `Number.intValue()` or `Number.longValue()`

labkey-susanh (approved):

experiment/src/org/labkey/experiment/ExperimentModule.java
labkey-nicka: I've tested locally and revised to this approach. Thanks for the tip.

PR #6075, opened 2024-11-20
AbDisc: Plate Set ELN Reference

RosalineP:
#### Rationale
See spec [here](https://docs.google.com/document/d/11Mtrt2Gguf7EWrHAzGZNBnR9GtUWzhpyWiKah9t8wN8/edit?tab=t.0#heading=h.4rg4h5vuek0).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6075
- https://github.com/LabKey/limsModules/pull/940
- https://github.com/LabKey/labkey-ui-components/pull/1648
- https://github.com/LabKey/labkey-ui-premium/pull/603

#### Changes
- Add plate sets to search engine
- Add LSIDs to plate sets
- Bump package version


assay/src/org/labkey/assay/plate/PlateSetDocumentProvider.java
```diff
@@ -0,0 +1,178 @@
...
+        props.put(SearchService.PROPERTY.title.toString(), "Plate Set - " + plateSet.getName());
+
+        ActionURL url = plateSet.detailsURL();
+        if (url != null)
+            url.setExtraPath(plateSet.getContainer().getId());
+
+        StringBuilder body = new StringBuilder();
+
+        if (plateSet.getDescription() != null)
+            append(body, plateSet.getDescription()); // PR Flag: what should the summary display?
```
RosalineP: Flagging. Happy to pop whatever makes sense in there. Some things that could make sense that we could nab are the plate count (or the plate names), the type, and the id

assay/src/org/labkey/assay/plate/PlateSetImpl.java
```diff
@@ -173,6 +188,15 @@ public String getDescription()
         return _description;
     }
 
+    @JsonIgnore
+    @Override
+    public @Nullable ActionURL detailsURL()
```
RosalineP, line 196: Currently the plate details url leads to the LKS plate editor. Do we always want this to be a LKS link? Happy to update this to whatever url makes sense

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -473,6 +481,20 @@ public List<Plate> getPlateTemplates(Container container)
         return PlateSetCache.getPlateSet(cf, plateSetId);
     }
 
+    @Override
+    public @Nullable PlateSet getPlateSet(String plateSetID)
```
RosalineP, line 485: Oh darn, just noticed that I left the id strung in here rather than the lsid. Will fix.

assay/api-src/org/labkey/api/assay/plate/PlateService.java
```diff
@@ -167,6 +167,7 @@ static PlateService get()
 
     @Nullable PlateSet getPlateSet(ContainerFilter cf, int plateSetId);
 
+    @Nullable PlateSet getPlateSet(String plateSetID);
```
labkey-nicka: Lets remove this from the service interface so there are no external usages of this call (the only call I see is using `PlateManager` directly).

assay/api-src/org/labkey/api/assay/plate/PlateSet.java
```diff
@@ -1,25 +1,36 @@
...
-public interface PlateSet
+public interface PlateSet extends Identifiable
 {
     int MAX_PLATES = 60;
     int MAX_PLATE_WELL_SIZE = 384;
     int MAX_PLATE_SET_WELLS = MAX_PLATES * MAX_PLATE_WELL_SIZE;
 
     Integer getRowId();
 
+    @Override
```
labkey-nicka: Now that this class extends `Identifiable` we can remove the `Container getContainer()` from this interface rather than marking it as `@Override`.

assay/api-src/org/labkey/api/assay/plate/PlateSet.java
```diff
@@ -1,25 +1,36 @@
...
     Container getContainer();
 
     String getDescription();
 
+    @Override
```
labkey-nicka: All of the `@Override` declared in this change set for this interface are duplicative and can be removed. The resulting definition is:

```java
public interface PlateSet extends Identifiable
{
    int MAX_PLATES = 60;
    int MAX_PLATE_WELL_SIZE = 384;
    int MAX_PLATE_SET_WELLS = MAX_PLATES * MAX_PLATE_WELL_SIZE;

    Integer getRowId();

    String getDescription();

    String getPlateSetId();

    boolean isArchived();

    boolean isAssay();

    boolean isPrimary();

    boolean isStandalone();

    boolean isTemplate();

    List<Plate> getPlates();

    PlateSetType getType();

    Integer getRootPlateSetId();
}
```

assay/src/org/labkey/assay/AssayUpgradeCode.java
```diff
@@ -681,6 +681,38 @@ public static void updateBarcodeSequence(ModuleContext ctx) throws Exception
...
+
+        DbScope scope = AssayDbSchema.getInstance().getSchema().getScope();
+        try (DbScope.Transaction tx = scope.ensureTransaction())
+        {
+            try (Results rs = new TableSelector(AssayDbSchema.getInstance().getTableInfoPlateSet()).getResults())
+            {
+                while (rs.next())
+                {
+                    Map<String, Object> row = rs.getRowMap();
+                    Container container = ContainerManager.getForId((String) row.get("container"));
```
labkey-nicka: `Result` provides type-strong getters for some basic types. Use `rs.getString("Container")` instead.

```java
Container container = ContainerManager.getForId(rs.getString("Container"));
```

assay/src/org/labkey/assay/AssayUpgradeCode.java
```diff
@@ -681,6 +681,38 @@ public static void updateBarcodeSequence(ModuleContext ctx) throws Exception
...
+                    Lsid lsid = PlateManager.get().getLsid(PlateSet.class, container);
+
+                    SQLFragment sql = new SQLFragment("UPDATE ").append(AssayDbSchema.getInstance().getTableInfoPlateSet(), "")
```
labkey-nicka: Get the `TableInfo` and `DbSchema` instances once. The `tx.commit()` should be called from the outer `try` scope.

```java
DbSchema schema = AssayDbSchema.getInstance().getSchema();
try (DbScope.Transaction tx = schema.getScope().ensureTransaction())
{
    TableInfo plateSetTable = AssayDbSchema.getInstance().getTableInfoPlateSet();
    try (Results rs = new TableSelector(plateSetTable).getResults())
    {
        while (rs.next())
        {
            Map<String, Object> row = rs.getRowMap();
            Container container = ContainerManager.getForId(rs.getString("Container"));
            Lsid lsid = PlateManager.get().getLsid(PlateSet.class, container);

            SQLFragment sql = new SQLFragment("UPDATE ").append(plateSetTable, "")
                    .append(" SET LSID = ?")
                    .add(lsid)
                    .append(" WHERE RowId = ?")
                    .add(row.get("rowId"));
            new SqlExecutor(schema).execute(sql);
        }
    }

    tx.commit();
}
```

assay/src/org/labkey/assay/plate/PlateSetDocumentProvider.java
```diff
@@ -0,0 +1,178 @@
...
+    public static WebdavResource createDocument(@NotNull PlateSet plateSet)
+    {
+        Map<String, Object> props = new HashMap<>();
+        Set<String> identifiersHi = new HashSet<>();
+        Set<String> identifiersMed = new HashSet<>();
+        final String documentId = getDocumentId(plateSet);
+
+        identifiersHi.add(plateSet.getName());
+        identifiersHi.add(plateSet.getPlateSetId());
+        identifiersMed.add(plateSet.getLSID());
```
labkey-nicka: Let's skip putting the LSID in the search document. We don't need to enabling searching by the LSID.

assay/src/org/labkey/assay/plate/PlateSetDocumentProvider.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    public static String getDocumentId(@NotNull Lsid plateSetLsid)
+    {
+        return getDocumentIdPrefix() + plateSetLsid;
+    }
+
+    public static String getDocumentId(@NotNull PlateSet plateSet)
+    {
+        return getDocumentId(new Lsid(plateSet.getLSID()));
```
labkey-nicka: I'm going to say we should not use LSID for the document identifier here because it has down stream effects and is not really a column we even want but are forced to have it since notebook writer for references requires identifiable objects. Rather than propagating LSID on plate sets further I recommend we use a combination of the container/rowId so that lookups into the plate set caching are more easily resolvable.

```java
// On PlateSetDocumentProvider
private static String getDocumentIdPrefix()
{
    return getSearchCategory().getName() + ":";
}

public static String getDocumentId(@NotNull PlateSet plateSet)
{
    return getDocumentIdPrefix() + plateSet.getContainer().getId() + ":" + plateSet.getRowId();
}

// On the SearchService.ResourceResolver instance
private @Nullable PlateSet getPlateSet(@NotNull String resourceIdentifier)
{
    final String prefix = getDocumentIdPrefix();

    if (resourceIdentifier.startsWith(prefix))
        resourceIdentifier = resourceIdentifier.substring(prefix.length());

    String[] parts = resourceIdentifier.split(":");
    if (parts.length != 2)
        return null;

    int rowId;
    try
    {
        rowId = Integer.parseInt(parts[1]);
    }
    catch (NumberFormatException e)
    {
        // skip it
        return null;
    }

    Container container = ContainerManager.getForId(parts[0]);
    if (container == null)
        return null;

    return PlateManager.get().getPlateSet(container, rowId);
}
```

This removes the need for downstream caching/querying by LSID which is debt we don't need.

assay/src/org/labkey/assay/plate/PlateSetDocumentProvider.java
```diff
@@ -0,0 +1,178 @@
...
+                        }
+                    }
+                }
+
+                return results;
+            }
+
+            private Map<String, Object> serialize(@NotNull PlateSet plateSet) throws JsonProcessingException
+            {
+                return new JSONObject(JsonUtil.DEFAULT_MAPPER.writeValueAsString(plateSet)).toMap();
```
labkey-nicka: This is currently serializing too much information. I don't think we should be serializing all related plates in this payload so I recommend stripping it out here in post-processing.

```java
private Map<String, Object> serialize(@NotNull PlateSet plateSet) throws JsonProcessingException
{
    JSONObject json = new JSONObject(JsonUtil.DEFAULT_MAPPER.writeValueAsString(plateSet));
    // Skip serializing plates into search results
    json.remove("plates");

    return json.toMap();
}
```

assay/src/org/labkey/assay/plate/query/PlateSetTable.java
```diff
@@ -55,17 +55,36 @@ public class PlateSetTable extends SimpleUserSchema.SimpleTable<UserSchema>
     private static final List<FieldKey> defaultVisibleColumns = new ArrayList<>();
     private final boolean _allowInsert;
 
+    public enum Column
+    {
+        Name,
```
labkey-nicka: Thanks for adding this column enumeration. This will be used widely. Order these columns by name so it is easier to find. Also, add `RowId`:

```java
public enum Column
{
    Created,
    CreatedBy,
    Description,
    Folder,
    Lsid,
    Modified,
    ModifiedBy,
    Name,
    PlateCount,
    RowId,
    Type;

    public FieldKey fieldKey()
    {
        return FieldKey.fromParts(name());
    }
}
```

assay/src/org/labkey/assay/plate/query/PlateSetTable.java
```diff
@@ -55,17 +55,36 @@ public class PlateSetTable extends SimpleUserSchema.SimpleTable<UserSchema>
...
-        defaultVisibleColumns.add(FieldKey.fromParts("Name"));
-        defaultVisibleColumns.add(FieldKey.fromParts("Folder"));
-        defaultVisibleColumns.add(FieldKey.fromParts("Type"));
-        defaultVisibleColumns.add(FieldKey.fromParts("Description"));
-        defaultVisibleColumns.add(FieldKey.fromParts("PlateCount"));
-        defaultVisibleColumns.add(FieldKey.fromParts("Created"));
-        defaultVisibleColumns.add(FieldKey.fromParts("CreatedBy"));
-        defaultVisibleColumns.add(FieldKey.fromParts("Modified"));
-        defaultVisibleColumns.add(FieldKey.fromParts("ModifiedBy"));
+        defaultVisibleColumns.add(FieldKey.fromParts(PlateSetTable.Column.Name.name()));
```
labkey-nicka: Use the `Column.fieldKey()` utility you declared above for these:

```java
static
{
    defaultVisibleColumns.add(PlateSetTable.Column.Name.fieldKey());
    defaultVisibleColumns.add(PlateSetTable.Column.Folder.fieldKey());
    defaultVisibleColumns.add(PlateSetTable.Column.Type.fieldKey());
    defaultVisibleColumns.add(PlateSetTable.Column.Description.fieldKey());
    defaultVisibleColumns.add(PlateSetTable.Column.PlateCount.fieldKey());
    defaultVisibleColumns.add(PlateSetTable.Column.Created.fieldKey());
    defaultVisibleColumns.add(PlateSetTable.Column.CreatedBy.fieldKey());
    defaultVisibleColumns.add(PlateSetTable.Column.Modified.fieldKey());
    defaultVisibleColumns.add(PlateSetTable.Column.ModifiedBy.fieldKey());
}
```

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -1153,6 +1182,8 @@ private int savePlateImpl(
             transaction.addCommitTask(() -> {
                 clearCache(container, plate);
                 indexPlate(container, plateRowId, false);
+                if (plate.getPlateSet() != null && SearchService.get() != null)
```
labkey-nicka, line 1171: We don't want to index the plate set each time a plate is saved. Instead, we should be more precise about the scenarios in which a plate set needs to be indexed and ensure all of those cases are hit:

1. Index when a plate set is created.
2. Index when a plate set is renamed.
3. Deindex when a plate set is deleted.
4. Deindex when a plate sets container is deleted.

assay/resources/schemas/dbscripts/postgresql/assay-24.013-24.014.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE assay.plateset ADD COLUMN LSID VARCHAR(200);
+
+SELECT core.executeJavaUpgradeCode('addLsidToPlateSets');
+
+ALTER TABLE assay.plateset ALTER COLUMN LSID SET NOT NULL;
```
labkey-nicka: In general, I'd say we should add an index on LSID, however, I'd like to have this LSID addition have as small of a footprint as possible in the hope that we can remove this someday soon. Let's stick with no index on this column.

assay/resources/schemas/dbscripts/postgresql/assay-24.013-24.014.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE assay.plateset ADD COLUMN LSID VARCHAR(200);
```
labkey-nicka: Use `LSIDtype` instead.

```sql
ALTER TABLE assay.plateset ADD COLUMN LSID LSIDtype;

**Review by labkey-nicka — CHANGES_REQUESTED** [review]

File: assay/src/org/labkey/assay/plate/PlateSetImpl.java (hunk 2 above)
Comment (labkey-nicka) [reply assay/src/org/labkey/assay/plate/PlateSetImpl.java:196]: Good to call this out. We do not want this to be the same URL as plates nor do we want to add a new route that is specific to plate sets. I recommend we resolve the query row details and then update the client-side to handle this URL mapping.

```java
@JsonIgnore
@Override
public @Nullable ActionURL detailsURL()
{
    // Plate sets do not currently have their own page in LKS. Link to the default query row details.
    ActionURL url = QueryService.get().urlDefault(getContainer(), QueryAction.detailsQueryRow, PlateSchema.SCHEMA_NAME, PlateSetTable.NAME);
    url.addParameter(PlateSetTable.Column.RowId.name(), getRowId());
    return url;
}
```

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -1350,6 +1368,18 @@ private void beforePlateSetsDelete(Collection<Integer> plateSetIds)
...
 
+        // De-index plate sets
+        try (DbScope.Transaction transaction = AssayDbSchema.getInstance().getScope().ensureTransaction())
+        {
+            transaction.addCommitTask(() -> {
+                for (Integer plateSetId : plateSetIds)
+                {
+                    deindexPlateSet(getPlateSet(container, plateSetId));
+                }
+            }, DbScope.CommitTaskOption.POSTCOMMIT);
```
RosalineP: Realized this isn't right. Updating

labkey-nicka (approved):

PR #6076, opened 2024-11-20
Issue 51702: Creating new workbooks/experiments is slow

labkey-jeckels:
#### Rationale
Creating containers is taking > 5 seconds on servers that have many sibling workbooks/

#### Changes
- Don't update the max size statistic on caches on every put. Puts can happen many, many times for big caches, and getting the size is expensive enough to add up. We only show the max size in the Admin Console, so it's OK to be a little less granular
- Avoid blowing away the container child cache when the hierarchy hasn't changed

### DIFF
--- a/api/src/org/labkey/api/data/ContainerManager.java
+++ b/api/src/org/labkey/api/data/ContainerManager.java
@@ -360,7 +360,7 @@ public class ContainerManager
                 SecurityManager.setAdminOnlyPermissions(c, savePolicyUser);
             }
 
-            _removeFromCache(c); // seems odd, but it removes c.getProject() which clears other things from the cache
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -920,7 +920,7 @@ public static void updateExpirationDate(Container container, LocalDate expiratio
         // Note: jTDS doesn't support LocalDate, so convert to java.sql.Date
```
labkey-adam, line 920: Just FYI: SQL Server JDBC supports LocalDate. That's for another day, of course.

labkey-adam (approved):

PR #6078, opened 2024-11-20
Misc exception report fixes

labkey-jeckels:
#### Rationale
Small fixes for issues without concrete repros

#### Changes
- Tolerate the Container disappearing in the midst of indexing materials/samples
- Disable both levels of ResultSet caching when generating schema TSVs
- Minor code cleanup

### DIFF
--- a/query/src/org/labkey/query/controllers/QueryController.java
+++ b/query/src/org/labkey/query/controllers/QueryController.java
@@ -1026,7 +1026,7 @@ public class QueryController extends SpringActionController
 
             try
             {
-                if (form.ff_baseTableName == null || "".equals(form.ff_baseTableName))
+                if (form.ff_baseTableName == null || form.ff_baseTableName.isEmpty())
                 {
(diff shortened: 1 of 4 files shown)


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -1026,7 +1026,7 @@ public boolean handlePost(NewQueryForm form, BindException errors)
 
             try
             {
-                if (form.ff_baseTableName == null || "".equals(form.ff_baseTableName))
+                if (form.ff_baseTableName == null || form.ff_baseTableName.isEmpty())
```
labkey-nicka: nit: Could use `StringUtils.isEmpty(form.ff_baseTableName)`

labkey-nicka (approved):

labkey-adam (approved):

query/src/org/labkey/query/controllers/QueryController.java
labkey-jeckels: Updated

PR #6091, opened 2024-11-22
Issue 51727: Metrics to count the usage of webparts

labkey-jeckels:
#### Rationale
We'd like insights into which web parts are being used and which can be deleted.

#### Changes
- New `webPartCounts` metric, with each web part name as a child and its usage count as its value


api/src/org/labkey/api/view/Portal.java
```diff
@@ -199,7 +204,24 @@ public static void clearWebPartProperties(String nameSearchText, String properti
...
+        {
+            SQLFragment sql = new SQLFragment("SELECT Name, COUNT(*) AS C FROM ").
+                    append(getTableInfoPortalWebParts(), "pwp").
+                    append(" GROUP BY Name");
+
+            Map<String, Object> counts = new HashMap<>();
+            new SqlSelector(getSchema(),
+                    sql).forEach(rs -> {
+                        counts.put(rs.getString("Name"), rs.getInt("C"));
+                    }
```
labkey-adam: Could use `getValueMap()`, but whatever

labkey-adam (approved):

PR #6093, opened 2024-11-22
two ideas for taming robots

labkey-matthewb:
#### Rationale
what do you think?

#### Related Pull Requests
- 

#### Changes
-


wiki/src/org/labkey/wiki/WikiController.java
```diff
@@ -1122,6 +1124,14 @@ public PageAction(ViewContext ctx, Wiki wiki, WikiVersion wikiversion)
         @Override
         public ModelAndView getView(WikiNameForm form, BindException errors)
         {
+            // Don't index page with non default parameters (e.g. targeting webparts in the page)
+            for (var e = getViewContext().getRequest().getParameterNames() ; e.hasMoreElements() ; )
+            {
+                String p = e.nextElement();
+                if (p.contains(".") && !LAST_FILTER_PARAM.equals(p))
+                    getPageConfig().setNoIndex();
```
labkey-matthewb, line 1132: maybe getPageConfig().setNoFollow() as well?

api/src/org/labkey/api/view/PopupMenuView.java
```diff
@@ -184,14 +185,31 @@ protected static void renderLink(NavTree item, String cls, Writer out) throws IO
...
         out.write("<a id='" + id + "'");
         if (null != cls)
             out.write(" class=\"" + cls + "\"");
-        if (null != item.getHref() && !item.isPost())
-            out.write(" href=\"" + PageFlowUtil.filter(item.getHref()) + "\"");
+        if (null != href && !item.isPost())
+            out.write(" href=\"" + PageFlowUtil.filter(href) + "\"");
         else
             out.write(" href=\"#\"");
+        if (null != dataHref)
```
labkey-matthewb: BTW I don't actually believe this is sufficient to hide the URL, but I think it could be with a very little more effort.

api/src/org/labkey/api/view/PopupMenuView.java
```diff
@@ -184,14 +187,45 @@ protected static void renderLink(NavTree item, String cls, Writer out) throws IO
...
 
+        // NOTE: nofollow is not recommended as way to avoid crawling internal links
+        // instead let's use an onclick handler to "hide" the link
+        String dataQuery = null;
+        String href = item.getHref();
+        if (null != href && null == item.getScript() && !item.isPost())
+        {
+            try
+            {
+                var context = HttpView.currentContext();
```
labkey-jeckels, line 198: It looks like this might help the Container Filter menu items? Looks like they're `href` and not JS handlers. They already have `rel="nofollow"` but that's not enough?

wiki/src/org/labkey/wiki/WikiController.java
labkey-jeckels: What's the use case here? An HTML wiki that uses `QueryWebPart` via a `<script>` tag? Something that uses the server-side webpart include syntax?

labkey-jeckels (commented):
I'm not following

api/src/org/labkey/api/view/PopupMenuView.java
labkey-matthewb: From my reading.  nofollow doesn't actually mean "pretend this target link or page doesn't exist".  It just means "don't give this link any weight in your magic SEO algorithm, I don't vouch for it".  Google is going to crawl every link it can find.

wiki/src/org/labkey/wiki/WikiController.java
labkey-matthewb: Exactly.  This certainly isn't the only place this could be relevant (portal pages are probably more important).  But letting the crawler go nuts on the pages with multiple data regions seems to exacerbate the combinatorics.

labkey-matthewb: Open to other ideas, just trying to keep up with the bots.

api/src/org/labkey/api/view/PopupMenuView.java
labkey-jeckels: I see.

Why the split between the `href` and the `data-query`?

labkey-matthewb: Yes, this is an important question.  The interwebs seems to think google is very good at finding links in attributes and javascript.  I am not sure that this change is sufficient, but it seemed worth trying to separate the parts so that the varying part does not look like a URL.

wiki/src/org/labkey/wiki/WikiController.java
labkey-jeckels: This seems like it's worth trying. It shouldn't hurt and might help/.

api/src/org/labkey/api/security/AuthFilter.java
```diff
@@ -228,21 +230,9 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 
         QueryService.get().setEnvironment(QueryService.Environment.USER, user);
 
-        if (AppProps.getInstance().isOptionalFeatureEnabled("experimental-unsafe-inline"))
```
labkey-jeckels, line 240: Did you mean to delete this? Maybe since I don't see it being registered as an optional feature anywhere?

api/src/org/labkey/api/security/AuthFilter.java
```diff
@@ -313,4 +303,26 @@ private void ensureFirstRequestHandled(HttpServletRequest request)
...
+    {
+        var userAgent = req.getHeader("User-Agent");
+        /* NOTE: the health checker is currently considered a robot.  Don't 429 the healthchecker */
+        if (StringUtils.contains(userAgent, "healthchecker"))
+            return false;
+        if (!req.isRobot())
+            return false;
+        var count = robotLimiter.getCount();
+        var delay = robotLimiter.getDelay();
+        if (count >= 10 && delay > 0)
```
labkey-jeckels, line 328: This overall approach seems great to me, but I'll admit that I can't figure out what the effective limits will be. I'm confused on how `10` and `0` connect with the arguments to `Rate` above.

Regardless, to move this forward, we might want config to control this. Either a simple on/off switch or a way to control the allowed rate.

labkey-matthewb: yes.  Needs startup properties/configuration before being released into the wild.  I'll add comments.

PR #6094, opened 2024-11-22
Address NPE in set password workflows

labkey-adam:
#### Rationale
Users are clicking their emailed verification links long after they've verified. A recent refactor caused this case to NPE.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6021

#### Changes
- Ensure that no attempt is made to generate a success message when there's an unrecoverable error (e.g., verification failed, so no user)
- In the verification failure case, inspect the token to provide slightly more specific guidance
- Also, recommend PostgreSQL 17.x

### DIFF
--- a/core/src/org/labkey/core/login/LoginController.java
+++ b/core/src/org/labkey/core/login/LoginController.java
@@ -133,6 +133,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Supplier;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.labkey.api.security.AuthenticationManager.AUTO_CREATE_ACCOUNTS_KEY;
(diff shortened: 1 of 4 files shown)


core/src/org/labkey/core/login/LoginController.java
```diff
@@ -1709,9 +1710,9 @@ protected void verify(SetPasswordForm form, Errors errors)
         }
 
         @Override
-        protected String getMessage(SetPasswordForm form)
+        protected Supplier<String> getSuccessMessageSupplier(SetPasswordForm form)
         {
-            return "Your email address (" + _user.getEmail() + ") has been verified! Create an account password below.";
+            return () -> "Your email address (" + _user.getEmail() + ") has been verified! Create an account password below.";
```
labkey-jeckels, line 1715: Would a null check still be useful here?

labkey-jeckels (approved):

core/src/org/labkey/core/login/LoginController.java
labkey-adam: If `_user` is null then our logic is broken. I'd rather throw so we know, especially for a security-sensitive area.

I was tempted to rewrite this whole thing (we're clearly trying to cram too many disparate workflows into a single JSP, bean, form, and base action), but I restrained myself.

PR #6095, opened 2024-11-22
Issue 51727: Metrics to count the usage of webparts

labkey-jeckels:
#### Rationale
We'd like insights into which web parts are being used and which can be deleted.

#### Changes
- New `webPartCounts` metric, with each web part name as a child and its usage count as its value


labkey-jeckels:
@labkey-adam I accidentally merged develop into my first FB. Was easier for me to create a new FB and PR. I switched to getValueMap() to make you happy. (Note: I'd probably use getValueMap() more often if it had an overloaded version that accepted the column names for the key and value)

labkey-adam (approved):

PR #6096, opened 2024-11-23
Issue 51749: Site validator fails on some wikis

labkey-jeckels:
#### Rationale
Some Wikis need the current URL to render. We have to supply a fake one for site validation that's happening in a pipeline job.

#### Changes
- Supply a dummy URL


labkey-adam (approved):
👍

PR #6100, opened 2024-11-25
Chart builder support for Trendline option in line chart

cnathe:
#### Rationale
Adding a chart builder feature to allow selection of a line chart Trendline option: Linear Regression, 3PL, 4PL, etc. This will be in addition to the default line chart trendline option of "Point-to-Point". Note that this new option will only show in the UI for LIMS products. Also note that the nonlinear curve fit options are only available for assay data. Other data types will see the Linear Regression and Polynomial types.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1650

#### Changes
- GenericChartHelper support for querying and rendering line chart config layers based on trendline config
- query for per-series trendline data using new premium-calculateCurveFit API
- render new Path layer for each trendline series data
- queryTrendlineData() to be called during queryChartData()
- Add trendline hover text to show curve fit params and stats
- Add Trendline Option to Line Charts for LKS Chart Wizard
- Filter trendline options based on schema prefix (i.e. linear vs non-linear options)
- LKS chart wizard trendline option to be LIMS only (check feature flag in moduleContext)
- null checks for vis sort fn string checks


visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -1179,6 +1210,134 @@ LABKEY.vis.GenericChartHelper = new function(){
...
+        'Linear': { label: 'Linear Regression', value: 'Linear', equation: 'y = x * slope + intercept' },
+        'Polynomial': { label: 'Polynomial', value: 'Polynomial', equation: 'y = a0 + a1*x + a2*x^2' },
+        'Three Parameter': { label: 'Nonlinear 3PL', value: 'Three Parameter', showMax: true, schemaPrefix: 'assay', equation: 'y = max / [1 + (inflection - x) * slope]' },
+        '3 Parameter': { label: 'Nonlinear 3PL (Alternate)', value: '3 Parameter', showMax: true, schemaPrefix: 'assay', equation: 'y = max * abs(x/inflection)^abs(slope) / [1 + abs(x/inflection)^abs(slope)]' },
+        'Four Parameter': { label: 'Nonlinear 4PL', value: 'Four Parameter', showMin: true, showMax: true, schemaPrefix: 'assay', equation: 'y = min + (max - min) / [1 + (inflection - x) * slope]' },
+        '4 Parameter': { label: 'Nonlinear 4PL (Simplex)', value: '4 Parameter', schemaPrefix: 'assay', equation: 'y = max + (min - max) / [1 + (x/inflection)^slope]' },
+        'Five Parameter': { label: 'Nonlinear 5PL', value: 'Five Parameter', showMin: true, showMax: true, schemaPrefix: 'assay', equation: 'y = min + (max - min) / [[1 + (inflection - x) * slope]^asymmetry]' },
+    }
```
cnathe: @labkey-klum can you verify these equations against the server side ParameterCurveFit (and related classes) fitCurve() method?

labkey-klum: Yup, those match the server side code.

core/webapp/vis/src/utils.js
```diff
@@ -331,8 +331,8 @@ LABKEY.vis.arrayObjectIndexOf = function(myArray, searchTerm, property) {
 
 LABKEY.vis.discreteSortFn = function(a,b) {
     // Issue 23015: sort categorical x-axis alphabetically with special case for "Not in X" and "[Blank]"
-    var aIsEmptyCategory = a && (a.indexOf("Not in ") == 0 || a == '[Blank]'),
-        bIsEmptyCategory = b && (b.indexOf("Not in ") == 0 || b == '[Blank]');
+    var aIsEmptyCategory = a && a.indexOf !== undefined && (a.indexOf("Not in ") == 0 || a == '[Blank]'),
```
labkey-alan: would it maybe be more correct to do something like:

```suggestion
    var aIsEmptyCategory = a && typeof a === 'string' && (a.indexOf("Not in ") == 0 || a == '[Blank]'),
```

core/webapp/vis/src/utils.js
```diff
@@ -349,6 +349,8 @@ LABKEY.vis.naturalSortFn = function(aso, bso) {
     var a, b, a1, b1, i= 0, n, L,
         rx=/(\.\d+)|(\d+(\.\d+)?)|([^\d.]+)|(\.\D+)|(\.$)/g;
     if (aso === bso) return 0;
+    if (aso.toLowerCase === undefined || bso.toLowerCase === undefined) return 0;
```
labkey-alan:
```suggestion
    if (typeof aso !== 'string' || typeof bso !== 'string') return 0;
```

visualization/resources/web/vis/chartWizard/chartTypeDialog/trendlineField.js
```diff
@@ -0,0 +1,81 @@
...
+        this.callParent();
+    },
+
+    getTrendlineTypeStore : function() {
+        if (!this.trendlineTypeStore) {
+            var data = [];
+            for (var value in this.options) {
+                var option = LABKEY.vis.GenericChartHelper.TRENDLINE_OPTIONS[this.options[value]];
+                data.push([option.value, option.label, option.showMin, option.showMax, option.schemaPrefix]);
+            }
```
labkey-alan:
```suggestion
            const data = this.options.map((key) => {
                const option = LABKEY.vis.GenericChartHelper.TRENDLINE_OPTIONS[key];
                return [option.value, option.label, option.showMin, option.showMax, option.schemaPrefix];
            });
```

labkey-alan: I'm fairly certain we can use let/const in our normal JS files now. All browsers support these keywords and IIRC we updated our minifier to support the newer syntax.

visualization/resources/web/vis/chartWizard/chartTypeDialog/trendlineField.js
```diff
@@ -0,0 +1,81 @@
...
+        return this.trendlineTypeCombo;
+    },
+
+    getValue : function()
+    {
+        var values = {
+            trendlineType: this.getTrendlineTypeCombo().getValue()
+        };
+
+        return values;
```
labkey-alan:
```suggestion
        return {
            trendlineType: this.getTrendlineTypeCombo().getValue()
        };
```

visualization/resources/web/vis/chartWizard/chartTypeDialog/trendlineField.js
```diff
@@ -0,0 +1,81 @@
...
+    bodyStyle: 'background-color: transparent;',
+    height: 34,
+
+    baseQueryKey: null,
+    initData: null,
+
+    options: ['', 'Linear', 'Polynomial', 'Three Parameter', '3 Parameter', 'Four Parameter', '4 Parameter', 'Five Parameter'],
+
+    initComponent : function()
+    {
```
labkey-alan: 1. I'm assuming it's just our Ext "standard" to use `initComponent : function ()` instead of `initComponent: function()`, is that correct?
2. Nit: The opening paren is on a new line here, but elsewhere below it's on the same line as `function()`, so we should probably put it on the same line as function. However, it looks like in other files we have it on the next line. I'm not sure it matters too much, as long as we're consistent within the same file.

visualization/resources/web/vis/chartWizard/chartTypePanel.js
```diff
@@ -898,10 +898,20 @@ Ext4.define('LABKEY.vis.ChartTypeFieldSelectionsPanel', {
                 fieldSelection = undefined;
             }
 
+            // special case for trendline field, only visible for LIMS+ when premium module is available
+            var hidden = false;
+            if (field.name === 'trendline') {
+                var productFeatures = LABKEY.getModuleContext('core').productFeatures ?? [];
+                hidden = LABKEY.getModuleContext('api').moduleNames.indexOf('premium') === -1
+                    || productFeatures.indexOf('ChartBuilding') === -1;
+            }
```
labkey-alan:
```suggestion
            let hidden = false;
            if (field.name === 'trendline') {
                const hasChartBuilding = LABKEY.getModuleContext('core').productFeatures?.indexOf('ChartBuilding') !== -1;
                const hasPremium = LABKEY.getModuleContext('api').moduleNames.indexOf('premium') !== -1;
                hidden = !hasPremium || !hasChartBuilding;
            }
```

visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -1103,17 +1109,42 @@ LABKEY.vis.GenericChartHelper = new function(){
...
-                        geom: new LABKEY.vis.Geom.Path({
-                            color: '#' + chartConfig.geomOptions.pointFillColor,
-                            size: chartConfig.geomOptions.lineWidth?chartConfig.geomOptions.lineWidth:3,
-                            opacity:chartConfig.geomOptions.opacity
-                        }),
-                        aes: pathAes
-                    })
-                );
+                if (trendlineData) {
+                    $.each(trendlineData, async function (idx, trendline) {
```
labkey-alan: Two things:

1. This doesn't need to be async, since we're not awaiting anything
2. We should be able to just do `trendlineData.forEach((trendline, idx) => {});`, I know we use jQuery elsewhere in this file for historical reasons, but it's not really needed anymore, so I am not sure we should add more jQuery code.

visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -1179,6 +1210,134 @@ LABKEY.vis.GenericChartHelper = new function(){
...
+        'Four Parameter': { label: 'Nonlinear 4PL', value: 'Four Parameter', showMin: true, showMax: true, schemaPrefix: 'assay', equation: 'y = min + (max - min) / [1 + (inflection - x) * slope]' },
+        '4 Parameter': { label: 'Nonlinear 4PL (Simplex)', value: '4 Parameter', schemaPrefix: 'assay', equation: 'y = max + (min - max) / [1 + (x/inflection)^slope]' },
+        'Five Parameter': { label: 'Nonlinear 5PL', value: 'Five Parameter', showMin: true, showMax: true, schemaPrefix: 'assay', equation: 'y = min + (max - min) / [[1 + (inflection - x) * slope]^asymmetry]' },
+    }
+
+    var generateTrendlinePathHover = function(trendline)
+    {
+        var hoverText = trendline.name + '\n';
+        hoverText += '\n' + TRENDLINE_OPTIONS[trendline.data.curveFit.type].label + ':\n';
+        $.each(trendline.data.curveFit, function (key, value) {
```
labkey-alan: Same as my previous comment, and for all below, I don't think we should use `$.each()`, as it's no longer necessary, and it's more verbose. I don't think we need to be going out of our way to remove existing usages, but we shouldn't add new ones.

visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -1179,6 +1210,134 @@ LABKEY.vis.GenericChartHelper = new function(){
         return plotConfig;
     };
 
+    var hasPremiumModule = function() {
```
labkey-alan: For this definition, and all other new ones in this PR I think we should use `let` and `const` where appropriate, and avoid `var`, as `var` has several downsides, and we're no longer limited by our minifier.

visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -1179,6 +1210,134 @@ LABKEY.vis.GenericChartHelper = new function(){
...
+    };
+
+    // support for y-axis trendline data when a single y-axis measure is selected
+    var queryTrendlineData = function(chartConfig, data, callback) {
+        var chartType = getChartType(chartConfig);
+        var yMeasures = ensureMeasuresAsArray(chartConfig.measures.y);
+        if (chartType === 'line_plot' && chartConfig.geomOptions?.trendlineType && chartConfig.geomOptions.trendlineType !== '' && yMeasures.length === 1) {
+            var xName = chartConfig.measures.x.name;
+            var trendlineConfig = getTrendlineConfig(chartConfig, data);
+            _queryTrendlineData(trendlineConfig, xName, yMeasures[0].name).then(function() {
```
labkey-alan: It's a little odd that `_queryTrendLineData` is `async`, but this method isn't, and is using promises.

visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -1179,6 +1210,134 @@ LABKEY.vis.GenericChartHelper = new function(){
...
+                if (series.rawData.length > 1) {
+                    series.data = await _querySeriesTrendlineData(trendlineConfig, series, xName, yName);
+                }
+            } catch (e) {
+                console.error(e);
+            }
+        }
+    };
+
+    var _querySeriesTrendlineData = async function(trendlineConfig, seriesData, xName, yName) {
```
labkey-alan: This doesn't need to be `async function`, because it's using Promise to wrap `Ajax.request`

visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -1658,21 +1817,22 @@ LABKEY.vis.GenericChartHelper = new function(){
         }
 
         queryConfig.success = function(measureStore) {
-            callback.call(this, measureStore);
+            queryTrendlineData(chartConfig, measureStore.records(), function(trendlineData) {
```
labkey-alan: I think we should make `queryTrendlineData` an async function that returns the data, and make this success callback an async function. Then we can skip having to pass a callback to `queryTrendlineData`

labkey-alan (changes requested):
Overall this looks good, but I do think there are some changes we should make

core/webapp/vis/src/utils.js
cnathe: yup, done

core/webapp/vis/src/utils.js
cnathe: done

visualization/resources/web/vis/chartWizard/chartTypeDialog/trendlineField.js
cnathe: yes, I verified by running `./gradlew compressClientLibs` locally

visualization/resources/web/vis/chartWizard/chartTypeDialog/trendlineField.js
cnathe: done

visualization/resources/web/vis/chartWizard/chartTypeDialog/trendlineField.js
cnathe: I don't know that we were ever very good about using "standards" or linting back in the Ext4 days. I've updated this file to remove the space before the colon and to move the opening paren to the same line.

visualization/resources/web/vis/chartWizard/chartTypePanel.js
cnathe: done

visualization/resources/web/vis/genericChart/genericChartHelper.js
cnathe: done and done

visualization/resources/web/vis/genericChart/genericChartHelper.js
cnathe: done

visualization/resources/web/vis/genericChart/genericChartHelper.js
cnathe: done

visualization/resources/web/vis/genericChart/genericChartHelper.js
cnathe: done

visualization/resources/web/vis/genericChart/genericChartHelper.js
cnathe: I switched this to async / await to match

visualization/resources/web/vis/genericChart/genericChartHelper.js
cnathe: done

labkey-alan (approved):
Looks good

PR #6104, opened 2024-11-26
Issue 51752: LKSM: Using a period in entity name and making it a required parent errors

XingY:
#### Rationale
When a parent data type is supplied as a column input, its query encoded since the data type name is a query name. The required field check currently checks for the decoded form only. Changes made to accept encoded parent input during validation.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1585
- https://github.com/LabKey/labkey-ui-premium/pull/539
- https://github.com/LabKey/platform/pull/5872
- https://github.com/LabKey/limsModules/pull/730
- https://github.com/LabKey/testAutomation/pull/2072

- https://github.com/LabKey/limsModules/pull/952

#### Changes
- check encoded parent fields during required field validation

### DIFF
--- a/api/src/org/labkey/api/dataiterator/StandardDataIteratorBuilder.java
+++ b/api/src/org/labkey/api/dataiterator/StandardDataIteratorBuilder.java
@@ -38,6 +38,7 @@ import org.labkey.api.query.ValidationException;
 import org.labkey.api.security.User;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/dataiterator/StandardDataIteratorBuilder.java
```diff
@@ -256,15 +257,13 @@ public DataIterator getDataIterator(DataIteratorContext context)
...
                 {
                     Set<String> allColumns = new CaseInsensitiveHashSet(convert.getColumnNameMap().keySet());
-                    for (Map.Entry<String, String> entry : additionalRequiredColumns.entrySet())
+                    for (Set<String> requiredColumnSet : additionalRequiredInsertColumns)
                     {
-                        if (!allColumns.contains(entry.getKey()) && !allColumns.contains(entry.getValue()))
-                        {
-                            setupError.addGlobalError("Data does not contain required field: " + entry.getValue());
-                        }
+                        if (Collections.disjoint(allColumns, new CaseInsensitiveHashSet(requiredColumnSet)))
```
labkey-susanh, line 265: I suppose we don't have a chance that there's a same case-insensitive column in the required set for two fields because we prevent that elsewhere.

I wonder if this will help us to walk back the change we made when we introduced the "Required" option that now permits only one alias per type.

labkey-susanh (approved):

labkey-susanh:
Let's please add a test for this somewhere.

XingY:
> Let's please add a test for this somewhere.

I planned to but realized the selenium test story for the original dev is still in the backlog. So instead I tagged the issue as need automated test.

PR #6110, opened 2024-12-02
Deprecate assay progress report

labkey-adam:
#### Rationale
Report is not used

### DIFF
--- a/study/src/org/labkey/study/StudyModule.java
+++ b/study/src/org/labkey/study/StudyModule.java
@@ -200,6 +200,7 @@ import java.util.stream.Collectors;
 public class StudyModule extends SpringModule implements SearchService.DocumentProvider
 {
     public static final String MODULE_NAME = "Study";
+    public static final String ASSAY_PROGRESS_REPORT_FLAG = "assayProgressReportFlag";
 
     public static final BaseWebPartFactory reportsPartFactory = new ReportsWebPartFactory();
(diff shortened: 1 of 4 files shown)


study/src/org/labkey/study/StudyModule.java
```diff
@@ -411,6 +412,11 @@ protected void startupAfterSpringConfig(ModuleContext moduleContext)
             "Merging of dataset that uses server-managed third key (such as GUID or auto RowId) is not officially supported. Unexpected outcome might be experienced when merge is performed.",
             false);
 
+        AdminConsole.addOptionalFeatureFlag(new OptionalFeatureFlag(ASSAY_PROGRESS_REPORT_FLAG,
```
labkey-nicka, line 415: I would expect this to cause the `ProgressReportTest` to start failing so there should likely be necessary test code updates as well.

labkey-nicka (approved):

study/src/org/labkey/study/StudyModule.java
labkey-adam: Yes, there will be

labkey-adam: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyEPostgres/3292356?buildTab=tests

PR #6113, opened 2024-12-02
Handle data state name collisions better

labkey-klum:
[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51781)

#### Rationale
In both the upgrade and folder creation code we try to ensure that the plate related data states exist. If other data states exist that may have been created by other QC workflows we need to handle potential name collisions better:
- If a data state with the desired name exists but doesn't match the expected state type then we will attempt to create a new data state with a unique name and has the required state type (which is the more important property). For now only LKSM and LKB create data states with a non-null state type and there is no overlap between the set of state types the applications employ.


labkey-alan (approved):
Looks good 👍

PR #6118, opened 2024-12-04
Issue 51618: Start tracking queries before execution

labkey-jeckels:
#### Rationale
When a query is very slow, it's helpful to capture and expose its SQL in the query tracker before it completes

#### Changes
- Start tracking queries before they're executed
- Update their runtime stats when they're complete
- Show whether we've dropped any executions based on throttling the number of queries in the queue
- Minor code cleanup

### DIFF
--- a/api/src/org/labkey/api/data/dialect/StatementWrapper.java
+++ b/api/src/org/labkey/api/data/dialect/StatementWrapper.java
@@ -2909,7 +2909,8 @@ public class StatementWrapper implements Statement, PreparedStatement, CallableS
 
     private @Nullable List<Object> translateParametersForQueryTracking()
     {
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/data/dialect/StatementWrapper.java
```diff
@@ -2941,6 +2907,42 @@ else if (o instanceof String)
         }
     }
 
+    private @Nullable List<Object> translateParametersForQueryTracking()
+    {
+        // Make a copy of the parameters list (it gets modified below) and switch to zero-based list (_parameters is a one-based list)
```
labkey-adam:
```suggestion
        // Make a copy of the parameters list (it gets modified by callers) and switch to zero-based list (_parameters is a one-based list)

```

labkey-adam (approved):

PR #6125, opened 2024-12-04
Issue 51432: special character cause aliquot rollup and assay transform script to work unexpectedly

XingY:
#### Rationale
The app quotes parent names that contains comma, including any data/material parents or aliquot parent names. The AliquotRollupDataIterator is fetching parent specified by aliquotedFrom as is, which might contain enclosing quotes, which resulted in parent sample not found and rollup skipped. 

Separately, PlateSamplesTest.testAssayImportWithTransformScript fails to transform and import data when well sample name contains double quote. The PR fixed the intermediate tsv file that LKS writes so the transfom script can work on the correct file. The other part of the problem is transcript script itself (for example, R) might not know how to property quote values when writing files back to LKS, which this PR doesn't address. 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1658
- https://github.com/LabKey/platform/pull/6125
- https://github.com/LabKey/limsModules/pull/975
- https://github.com/LabKey/testAutomation/pull/2172
- https://github.com/LabKey/provenance/pull/193
- https://github.com/LabKey/premiumModules/pull/127

#### Changes
- refactor the parentName extraction util from DerivationDataIterator and re-use for AliquotRollupDataIterator
- quoteValue when writing tsv file for transform scripts


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -419,6 +428,23 @@ private Pair<Boolean, Integer> determineRecalcFromExistingRecord(int i, Map<Stri
...
+            Object parentObj = get(i);
+            Collection<String> parentNames = getParentNames(parentObj, _tsvWriter, "AliquotedFrom", null);
+            if (parentNames != null)
+            {
+                List<String> parents = parentNames.stream()
+                        .map(String::trim)
+                        .filter(s -> !StringUtils.isEmpty(s))
+                        .toList();
+                if (!parents.isEmpty())
+                    return parents.get(0);
```
labkey-susanh, line 447: If we have more than one parent after all this work, should that throw or log an error or warning?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -971,69 +1061,18 @@ protected Set<Pair<String, String>> _getParentParts()
...
-                        continue;
-                    }
+                    Collection<String> parentNames = getParentNames(o, _tsvWriter, _parentCols.get(parentCol), getErrors());
 
-                    String parentColName = _parentCols.get(parentCol);
-                    Set<Pair<String, String>> parts = parentNames.stream()
-                        .map(String::trim)
-                        .map(s -> Pair.of(parentColName, s))
-                        .collect(Collectors.toSet());
+                    if (parentNames != null)
```
labkey-susanh, line 1072: Looks like `parentNames` can't actually be `null` here.

labkey-susanh (approved):

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: good idea. updated

PR #6127, opened 2024-12-05
Issue 51800: Add a metric for uses of the "Flag" field type

labkey-jeckels:
#### Rationale
We want to know how often different field types are used so we can invest in the right ones and discontinue others. Report the total number of properties by range (raw data type) and concept (specialized use of the data type). Flags use string as the range and flag as the concept.

#### Changes
- modules.Experiment.propertyCountsByRange.* holds the raw data type, such as: attachment, boolean, dateTime, double, fileLink, int, long, multiLine, string
- modules.Experiment.propertyCountsByConcept.* holds the "concept" data type, such as: auditRecordDataMap, autoInt, flag, ParticipantId

### DIFF
--- a/experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
@@ -680,7 +680,11 @@ public class PropertyServiceImpl implements PropertyService, UsageMetricsProvide
         );
     }
 
-    /** @return a map where the URI keys have been stripped of everything before the hash. http://www.labkey.org/exp/xml#attachment -> attachment */
+    /**
+     * @return a map where the URI keys have been stripped of everything before the hash or xsd: prefix.
+     * http://www.labkey.org/exp/xml#attachment -> attachment
+     * xsd:int -> int
+     */
     private Map<String, Integer> stripUriPrefixes(@NotNull Map<String, Object> valueMap)
     {
         Map<String, Integer> result = new HashMap<>();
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
```diff
@@ -670,10 +670,33 @@ public Map<String, Object> getUsageMetrics()
...
 
+    /** @return a map where the URI keys have been stripped of everything before the hash. http://www.labkey.org/exp/xml#attachment -> attachment */
+    private Map<String, Integer> stripUriPrefixes(@NotNull Map<String, Object> valueMap)
+    {
+        Map<String, Integer> result = new HashMap<>();
+        for (Map.Entry<String, Object> entry : valueMap.entrySet())
+        {
+            String key = entry.getKey();
+            int value = ((Number)entry.getValue()).intValue();
+            if (key.contains("#"))
```
XingY, line 695: Do we not care about types that are not of standard formats? I've seen rangeURIs like`xsd:int`, `xsd:string`. Those are probably mostly used by our selenium test apis. But our domain apis do accept those type of formats.

XingY (approved):

experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
labkey-jeckels: Thanks. My 24.11 DB was too minimal and clean to be a good test case. I tried on my develop DB and it has a mix. Updated to code to handle them. We should switch to storing a canonical URI but that's for another day.

PR #6128, opened 2024-12-05
Avoid returning FK metadata for a malformed TableInfo

labkey-jeckels:
#### Rationale
https://www.labkey.org/_mothership/mothership-showStackTraceDetail.view?exceptionStackTraceId=39156

```
Stack Trace	
org.json.JSONException: JSONObject["queryName"] not found.
	at org.json.JSONObject.get(JSONObject.java:596)
	at org.json.JSONObject.getString(JSONObject.java:883)
	at org.labkey.query.persist.QueryManager.validateFk(QueryManager.java:826)
	at org.labkey.query.persist.QueryManager.validateColumn(QueryManager.java:768)
	at org.labkey.query.persist.QueryManager.validateQueryMetadata(QueryManager.java:749)
	at org.labkey.query.controllers.QueryController$ValidateQueryMetadataAction.execute(QueryController.java:7338)
	at org.labkey.query.controllers.QueryController$ValidateQueryMetadataAction.execute(QueryController.java:7299)
	at org.labkey.api.action.BaseApiAction.handlePost(BaseApiAction.java:239)
	at org.labkey.api.action.ReadOnlyApiAction.handleGet(ReadOnlyApiAction.java:34)
```

#### Changes
- Bail out if we have a null query or schema name

### DIFF
--- a/api/src/org/labkey/api/data/JsonWriter.java
+++ b/api/src/org/labkey/api/data/JsonWriter.java
@@ -329,7 +329,7 @@ public class JsonWriter
         TableDescription lookupTable = columnInfo.getFkTableDescription();
         if (null != fk
                 && null != lookupTable
-                && (!(fk instanceof RowIdForeignKey) || !(((RowIdForeignKey)fk).getOriginalColumn().equals(columnInfo))))
+                && (!(fk instanceof RowIdForeignKey rfk) || !(rfk.getOriginalColumn().equals(columnInfo))))
         {
             JSONObject lookupInfo = new JSONObject();
             if (null != fk.getLookupContainer())
@@ -353,6 +353,12 @@ public class JsonWriter
                 queryName = lookupTable.getPublicName();
             else
                 queryName = lookupTable.getName();
+
+            if (queryName == null || schemaName == null)
+            {
+                // Bail out if something went wrong with creating the lookup target
+                return null;
+            }
 
             // Duplicate info with different property names for backwards compatibility
             lookupInfo.put("queryName", queryName);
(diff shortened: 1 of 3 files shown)


labkey-jeckels:
@labkey-matthewb do you think we should handle this differently?

api/src/org/labkey/api/data/JsonWriter.java
```diff
@@ -354,6 +354,12 @@ public static JSONObject getLookupInfo(ColumnInfo columnInfo, boolean includeDom
             else
                 queryName = lookupTable.getName();
 
+            if (queryName == null || schemaName == null)
```
labkey-matthewb: This isn't an error.  There are many foreign keys that don't point at tables in the visible user schemas.  The FK is only required to support getLookupTableInfo().  These two properties are meant to help the UI recognize that it might already have the meta-data of the lookup table cached.

labkey-matthewb (changes requested):
There are many cases where we want to show that FKs exist that don't point to public tables (e.g. lineage lookups)

api/src/org/labkey/api/data/JsonWriter.java
labkey-jeckels: OK. What do you suggest in we do with a null `queryName` in `QueryManager.validateFk()`? Show a warning? Treat it as healthy and skip trying to resolve the `TableInfo` around line 855 in `QueryManager`?

We've been getting this exception for a while now. https://www.labkey.org/_mothership/project-begin.view?ExceptionSummary.StackTrace~contains=org.json.JSONException%3A%20JSONObject%5B%22queryName%22%5D%20not%20found.

labkey-matthewb: I'm pretty sure the schema browser/customize view use this information.  The fix might be for the validateQuery to not try to validate queries that are not marked "public".  And maybe use JSONObject.optString() to avoid throwing an unhandled exception.

labkey-jeckels: It's already warning about non-public lookup targets. I made it not do any validation if there's no `queryName`

labkey-jeckels: @labkey-matthewb can you take a look at the latest?

labkey-matthewb (approved):

PR #6129, opened 2024-12-05
retry function should throw if it fails

labkey-matthewb:
#### Rationale
Fix for reported exception

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
@@ -192,7 +192,6 @@ public class ExpDataClassDataTableImpl extends ExpRunItemTableImpl<ExpDataClassD
     }
 
     @Override
-    @NotNull
     public Domain getDomain()
     {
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -412,73 +410,78 @@ protected void populateColumns()
         Supplier<Map<DomainProperty, Object>> defaultsSupplier = null;
 
         // Add the domain columns
-        Set<String> skipCols = CaseInsensitiveHashSet.of("lsid", "rowid", "name", "classid");
-        for (ColumnInfo col : extTable.getColumns())
+        // It is not true that getDomain() does not return null
+        if (null != getDomain())
```
labkey-jeckels, line 413: If that's the expected behavior, seems better to change the annotation on `getDomain()` instead of adding a comment to one if it's callers.

labkey-jeckels (approved):

PR #6132, opened 2024-12-06
Issue 51815: FileRootMaintenanceTask throws exception for S3 folders

labkey-jeckels:
#### Rationale
We should be able to calculate sizes for local and S3-backed storage.

#### Changes
- Switch from `File` to `Path` to be S3-compatible
- Add more detail to exception message to make finding offenders easier


labkey-adam (approved):
Thanks!

PR #6137, opened 2024-12-09
Handle unusual query encoding in link href

labkey-tchad:
#### Rationale
The link this test is checking doesn't encode a `~`. Need to adjust the expected `href` to account for this.
Also, fixing some invalid Locators that are used for validation here.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2169

#### Changes
- Handle unusual query encoding in href
- Fix invalid Locators

### DIFF
--- a/study/test/src/org/labkey/test/tests/study/StudyMergeParticipantsTest.java
+++ b/study/test/src/org/labkey/test/tests/study/StudyMergeParticipantsTest.java
@@ -135,8 +135,8 @@ public class StudyMergeParticipantsTest extends StudyBaseTest
 
         log("Check url's to are correctly constructed");
         final String url = WebTestHelper.buildRelativeUrl("study", PROJECT_NAME + "/" + FOLDER_NAME, "dataset",
-            Maps.of("datasetId", 5018,
-                "Dataset.ParticipantId~in", PTID_NO_ALIAS + ";" + PTID_NEW_2));
+            Maps.of("datasetId", 5018, "Dataset.ParticipantId~in", PTID_NO_ALIAS + ";" + PTID_NEW_2))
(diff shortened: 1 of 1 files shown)


study/test/src/org/labkey/test/tests/study/StudyMergeParticipantsTest.java
```diff
@@ -135,8 +135,8 @@ protected void doVerifySteps()
 
         log("Check url's to are correctly constructed");
         final String url = WebTestHelper.buildRelativeUrl("study", PROJECT_NAME + "/" + FOLDER_NAME, "dataset",
-            Maps.of("datasetId", 5018,
-                "Dataset.ParticipantId~in", PTID_NO_ALIAS + ";" + PTID_NEW_2));
+            Maps.of("datasetId", 5018, "Dataset.ParticipantId~in", PTID_NO_ALIAS + ";" + PTID_NEW_2))
+            .replace("%7E", "~"); // '~' isn't encoded in href
```
labkey-chrisj, line 139: is this substitution generalizable for other use cases?  Asking to prompt the question whether this should be done in a helper vs. in test code

labkey-chrisj (approved):

study/test/src/org/labkey/test/tests/study/StudyMergeParticipantsTest.java
labkey-tchad: I wouldn't count on it. There are a lot of "valid" encodings and I doubt we're consistent on the product side. We should probably avoid finding elements by `href` attribute like this. It would be preferable to grab the link by some other means and decode it.

PR #6144, opened 2024-12-11
Clear hit selections for excluded rows

labkey-klum:
#### Rationale
This adds validation to hit selections that will prevent marking rows as hits that are currently QC excluded. Additionally, any rows that are excluded will have any hit selections removed.

### DIFF
--- a/assay/src/org/labkey/assay/plate/AssayPlateTriggerFactory.java
+++ b/assay/src/org/labkey/assay/plate/AssayPlateTriggerFactory.java
@@ -6,6 +6,7 @@ import org.labkey.api.assay.AssayProvider;
 import org.labkey.api.assay.AssayResultDomainKind;
 import org.labkey.api.assay.AssayService;
 import org.labkey.api.assay.plate.AssayPlateMetadataService;
+import org.labkey.api.assay.plate.PlateDataStateManager;
 import org.labkey.api.data.CompareType;
 import org.labkey.api.data.Container;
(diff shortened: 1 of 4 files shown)


assay/src/org/labkey/assay/plate/AssayPlateTriggerFactory.java
```diff
@@ -139,11 +143,29 @@ public void complete(TableInfo table, Container c, User user, TableInfo.TriggerT
...
     {
+        Set<Integer> _excludedRows = new HashSet<>();
+
         @Override
         public void beforeUpdate(TableInfo table, Container c, User user, @Nullable Map<String, Object> newRow, @Nullable Map<String, Object> oldRow, ValidationException errors, Map<String, Object> extraContext) throws ValidationException
         {
             if (newRow != null && _qcStateProp != null)
-                AssayPlateMetadataServiceImpl.validateRowDataStates(c, newRow, _qcStateProp);
+            {
+                DataState state = AssayPlateMetadataServiceImpl.validateRowDataStates(c, newRow, _qcStateProp);
```
labkey-nicka, line 153: nit: Is there anything we could cache locally that would prevent replicate processing of permitted operations when handling lots of assay result rows?

labkey-nicka (approved):
Looks good. Seems like this could be covered well in the `PlateManagerTest` or perhaps a new `AssayPlateMetadataServiceTest` which could be useful for testing the boundaries between plate and assay.

PR #6151, opened 2024-12-14
Minor fixes to various study features

labkey-adam:
#### Rationale
Fixing confusing UI, bad wording & formatting, incomplete navtrails, clipped buttons, double print buttons, etc.

#### Changes
- Manage Participants
  - Rename `ManageAlternateIdsAction` --> `ManageParticipantsAction`, since it's evolved significantly over time. Same with its JSP.
  - Reference the participant noun (not hard-coded "participant") consistently in the "manage participants" link description
  - Clarify that an "Alternate ID" is used for masking
  - Move "Change and Merge <Participant> ID" button out of the "Alternate IDs" section into its own section since it has nothing to do with Alternate IDs
  - Use subject noun more consistently
- Fix vertically clipped Ext4 buttons on a couple pages
- Improve merge participants page's navtrail and wording
- [Issue 51865: Data Region on Manage Location page for a Study has two print buttons](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51865)

### DIFF
--- a/api/src/org/labkey/api/util/StringUtilsLabKey.java
+++ b/api/src/org/labkey/api/util/StringUtilsLabKey.java
@@ -467,6 +467,23 @@ public class StringUtilsLabKey
         return combined;
     }
 
+    /**
+     * @return the indefinite article to use with the passed in noun. Returns "a" for words starting with most
+     * consonants, "an" for words starting with a vowel, and "a(n)" for words starting with "h". This isn't perfect
+     * since it's not distinguishing voiced vs. unvoiced "h", acronyms that are spelled out, and other pronunciation
+     * quirks.
+     */
+    public static String getArticleForNoun(String noun)
+    {
+        char firstChar = noun.charAt(0);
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/util/StringUtilsLabKey.java
```diff
@@ -467,6 +467,23 @@ public static String joinWithConjunction(List<String> list, String conjunction)
...
+     */
+    public static String getArticleForNoun(String noun)
+    {
+        char firstChar = noun.charAt(0);
+        return switch (firstChar)
+        {
+            case 'h', 'H' -> "a(n)";
+            case 'a', 'e', 'i', 'o', 'u' -> "an";
+            default -> "a";
+        };
```
labkey-bpatel, line 484: Nice!

labkey-bpatel (approved):
Looks great! 
Also, verified locally.

PR #6153, opened 2024-12-16
set query environment before calling getDependencies()

labkey-matthewb:
#### Rationale
Need to set container environment sooner
51872

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/study/src/org/labkey/study/dataset/DatasetSnapshotProvider.java
+++ b/study/src/org/labkey/study/dataset/DatasetSnapshotProvider.java
@@ -864,6 +864,9 @@ public class DatasetSnapshotProvider extends AbstractSnapshotProvider implements
                 Map<Container, List<QuerySnapshotDefinition>> snapshotMap = new HashMap<>();
                 for (SnapshotDependency.SourceDataType sourceData : _sourceDataTypes)
                 {
+                    // getDependencies() can execute LabKey SQL.  Make sure environment is set up here.
+                    // Also see 51200.
+                    QueryService.get().setEnvironment(QueryService.Environment.CONTAINER, sourceData.getContainer());
                     for (QuerySnapshotDefinition snapshotDef : getDependencies(sourceData))
                     {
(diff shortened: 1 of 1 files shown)


study/src/org/labkey/study/dataset/DatasetSnapshotProvider.java
```diff
@@ -864,6 +864,9 @@ public void run()
                 Map<Container, List<QuerySnapshotDefinition>> snapshotMap = new HashMap<>();
                 for (SnapshotDependency.SourceDataType sourceData : _sourceDataTypes)
                 {
+                    // getDependencies() can execute LabKey SQL.  Make sure environment is set up here.
+                    // Also see 51200.
+                    QueryService.get().setEnvironment(QueryService.Environment.CONTAINER, sourceData.getContainer());
                     for (QuerySnapshotDefinition snapshotDef : getDependencies(sourceData))
                     {
                         QueryService.get().setEnvironment(QueryService.Environment.CONTAINER, snapshotDef.getContainer());
```
labkey-jeckels, line 872: Is this still useful to retain? I assume it'll be more accurate than the new call?

labkey-jeckels (approved):

study/src/org/labkey/study/dataset/DatasetSnapshotProvider.java
labkey-matthewb: I thought about removing it.  In the original intention of this it should be the container on the URL, in the compliance usage what should it be: the source query container?, the definition container?  We should probably be disabling compliance logging for these metadata only usages.

labkey-matthewb: I think we need to address 51200 and the new PHI spec to really clean this up. 
https://docs.google.com/document/d/16tEyFQp98JjoJQwkxeAJsFV78aQjYFh0aM_fCIpfRL8/edit?tab=t.0#heading=h.4rg4h5vuek0

PR #6156, opened 2024-12-17
Fix notebook data reference

XingY:
#### Rationale
The documentId for dataclass data include the dataclass name. When the dataclass name contains special characters such as "/", the documentId cannot be parsed correctly. 

[SMProFolderExclusionTest.testExistingReferencesWithUpdatedExclusion](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SampleManagerBPostgres/3311567?buildTab=tests&status=failed&name=SMProFolderExclusionTest.testExistingReferencesWithUpdatedExclusion)

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6156
- https://github.com/LabKey/limsModules/pull/1007
- https://github.com/LabKey/testAutomation/pull/2200

#### Changes
- encode datatype name when generating documentid for search results

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpDataImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpDataImpl.java
@@ -633,7 +633,8 @@ public class ExpDataImpl extends AbstractRunItemImpl<Data> implements ExpData
         ExpDataClass dc = getDataClass(null);
         if (dc != null)
             dataClassName = dc.getName();
-        return "data:" + new Path(getContainer().getId(), dataClassName, Integer.toString(getRowId()));
+        // why not just data:rowId?
+        return "data:" + new Path(getContainer().getId(), dataClassName, Integer.toString(getRowId())).encode();
     }
 
     @Override


experiment/src/org/labkey/experiment/api/ExpDataImpl.java
```diff
@@ -633,7 +633,8 @@ public String getDocumentId()
         ExpDataClass dc = getDataClass(null);
         if (dc != null)
             dataClassName = dc.getName();
-        return "data:" + new Path(getContainer().getId(), dataClassName, Integer.toString(getRowId()));
+        // why not just data:rowId?
```
labkey-susanh, line 636: good question. I'm not sure why the dataClassName is needed. We don't use it for materials, for example. Probably done to make the id more human-readable in some way, but we don't make any concessions to that in other places, so I would guess data:rowId would work just fine. 

The last revision of this method had this comment:

```
        // CONSIDER: use lsid so we can crack it later?  Include dataClassId?
        return "data:" + new Path(getContainer().getId(), Integer.toString(getRowId()));
```
But for materials, we don't even add the container id (certainly not needed for uniqueness.

labkey-susanh (approved):

PR #6162, opened 2024-12-19
Change 'sessionKey' query parameter name

labkey-klum:
[tracking issue](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51715)

#### Rationale
While not a security concern, rename the `sessionKey` parameter and consolidate the code which uses this method to stash and retrieve files from the session.

### DIFF
--- a/api/src/org/labkey/api/reports/report/r/view/ImageOutput.java
+++ b/api/src/org/labkey/api/reports/report/r/view/ImageOutput.java
@@ -18,16 +18,15 @@ package org.labkey.api.reports.report.r.view;
 
 import org.apache.commons.lang3.BooleanUtils;
 import org.labkey.api.reports.Report;
-import org.labkey.api.reports.report.r.RReport;
 import org.labkey.api.reports.report.ReportDescriptor;
 import org.labkey.api.reports.report.ReportUrls;
 import org.labkey.api.reports.report.ScriptOutput;
 import org.labkey.api.reports.report.ScriptReportDescriptor;
 import org.labkey.api.reports.report.r.AbstractParamReplacement;
 import org.labkey.api.reports.report.r.ParamReplacement;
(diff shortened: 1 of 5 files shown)


labkey-jeckels (approved):

api/src/org/labkey/api/reports/report/r/view/ImageOutput.java
```diff
@@ -133,10 +132,9 @@ protected String renderInternalAsString(File file)
 
                 if (imgFile != null)
                 {
-                    String key = "temp:" + GUID.makeGUID();
-                    getViewContext().getRequest().getSession(true).setAttribute(key, imgFile);
+                    String key = ImageUtil.setFileInSession(getViewContext().getRequest(), imgFile);
```
labkey-adam, line 135: `setFileInSession()` could return `null`... is that okay for this code path?

api/src/org/labkey/api/util/ImageUtil.java
```diff
@@ -53,6 +54,10 @@
 public class ImageUtil
 {
     private static Logger LOG = LogManager.getLogger(ImageUtil.class);
+    public static final String FILE_SESSION_PARAM = "fileCacheKey";
+    public static final String DELETE_FILE_PARAM = "deleteFile";
+    public static final String CACHE_FILE_PARAM = "cacheFile";
+    public static final String ATTACHMENT_PARAM = "attachment";
```
labkey-adam, line 60: Yay constants!

labkey-adam (approved):

api/src/org/labkey/api/reports/report/r/view/ImageOutput.java
labkey-klum: yes, it should result in a noop

PR #6165, opened 2024-12-20
Move StatusCake sync code into mothership module system maintenance task

labkey-jeckels:
#### Rationale
Instead of using a standalone script that has to be run manually, it'll be more convenient to have a nightly sync of uptime data from StatusCake to labkey.org. As a system maintenance task, it can also be run on demand.

#### Changes
- New maintenance task that extracts data from StatusCake and inserts into three lists
- API key and target container via the existing Mothership config page

### DIFF
--- a/mothership/src/org/labkey/mothership/statuscake/StatusCakeMaintenanceTask.java
+++ b/mothership/src/org/labkey/mothership/statuscake/StatusCakeMaintenanceTask.java
@@ -1,0 +1,287 @@
+package org.labkey.mothership.statuscake;
+
+import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
+import org.apache.hc.client5.http.impl.classic.HttpClients;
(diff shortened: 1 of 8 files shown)


mothership/src/org/labkey/mothership/statuscake/StatusCakeMaintenanceTask.java
```diff
@@ -0,0 +1,287 @@
...
+        }
+    }
+
+    private static List<History> getUptime(Server server, CloseableHttpClient client, LocalDateTime since, String statusCakeAPIKey) throws IOException
+    {
+        Instant instant = since.toInstant(ZoneOffset.UTC);
+
+        int limit = 100;
+
+        // Limit to data from 2022 and later
```
labkey-adam: Maybe move comment up to where `since` is defined?

labkey-adam (approved):

PR #6167, opened 2024-12-20
Customizable File Templates for Sources, Sample Types & Assay Designs

XingY:
#### Rationale
Custom import templates were previously supported by overriding the the datatype's query metadata xml by providing the webdav location of an uploaded file. This set of PRs allows users to manage custom templates from app UI.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1671
- https://github.com/LabKey/labkey-ui-premium/pull/625
- https://github.com/LabKey/platform/pull/6167
- https://github.com/LabKey/limsModules/pull/1021

#### Changes
- Created `UpdateDomainImportTemplateAction` api to upload custom template files and save domain template settings to query xml at query.querydef.metadata
- Modified `getQueryDetailsAction` to include `available` status of import templates
- Added `CustomImportTemplates` product feature


api/src/org/labkey/api/exp/property/DomainKind.java
```diff
@@ -415,4 +415,9 @@ public boolean supportsNamingPattern()
     {
         return false;
     }
+
+    public String getDomainFileDirectory()
+    {
+        return getKindName();
```
cnathe, line 421: do we want toLowerCase() for these? maybe it doesn't matter but the other dir names seem to all be lowercase.

api/src/org/labkey/api/util/FileUtil.java
```diff
@@ -1282,9 +1282,17 @@ public static String encodeForURL(String str, boolean checkEncoded)
         return res.toString();
     }
 
+    private static final String[] ENCODED = {"%25", "%23", "%26", "%40", "%7E", "%7B", "%7D", "%5B", "%5D", "%2B", "%20"};
+    private static final String[] DECODED = {"%", "%#", "&", "@", "~", "{", "}", "[", "]", "+", " "};
```
cnathe: can the if/elseif block in the `encodeForURL` for loop above be refactored to use these String[] objects? basically to a single if statement like "if character exists in DECODED, replace with the same index in ENCODED".

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -501,6 +519,229 @@ public Object execute(DomainApiForm form, BindException errors)
...
+    {
+        @Override
+        protected ObjectMapper createResponseObjectMapper()
+        {
+            return this.createRequestObjectMapper();
+        }
+
+        @Override
+        public void validateForm(DomainTemplateForm form, Errors errors)
+        {
```
cnathe: there is a lot of validation that is happening in the `getUploadedTemplates()` method and the 'execute()` method. Can / should any of that be moved to here?

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -501,6 +519,229 @@ public Object execute(DomainApiForm form, BindException errors)
...
+                }
+            }
+            return rowFiles;
+        }
+
+        private List<Pair<String, String>> getUploadedTemplates(DomainTemplateForm form, DomainKind kind) throws ValidationException, QueryUpdateServiceException, ExperimentException
+        {
+            FileContentService fcs = FileContentService.get();
+            if (fcs == null)
+                throw new IllegalStateException("Unable to load file service");
```
cnathe:
```suggestion
                throw new IllegalStateException("Unable to load file service.");

```

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -501,6 +519,229 @@ public Object execute(DomainApiForm form, BindException errors)
...
+            Container container = getContainer();
+            String domainURI = PropertyService.get().getDomainURI(form.getSchemaName(), form.getQueryName(), container, user);
+            DomainKind kind = PropertyService.get().getDomainKind(domainURI);
+            Domain domain = PropertyService.get().getDomain(container, domainURI);
+            if (domain == null)
+                throw new IllegalArgumentException("Domain '" + domainURI + "' not found");
+            if (!kind.canEditDefinition(user, domain))
+                throw new UnauthorizedException("You don't have permission to update import templates for this domain.");
+
+            List<Pair<String, String>> updatedTemplates = getUploadedTemplates(form, kind);
```
cnathe: we call `getUploadedTemplates` here, which will end up saving the files to the filesystem. And then we do a few more validation checks that aren't related to the files. If one of those throws, the files will stay on the filesystem right? Should we move this call after the validation checks (down towards line 703 where it is used)?

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -501,6 +519,229 @@ public Object execute(DomainApiForm form, BindException errors)
...
+        @Override
+        public Object execute(DomainTemplateForm form, BindException errors) throws ValidationException, QueryUpdateServiceException, SQLException, ExperimentException
+        {
+            User user = getUser();
+            Container container = getContainer();
+            String domainURI = PropertyService.get().getDomainURI(form.getSchemaName(), form.getQueryName(), container, user);
+            DomainKind kind = PropertyService.get().getDomainKind(domainURI);
+            Domain domain = PropertyService.get().getDomain(container, domainURI);
+            if (domain == null)
+                throw new IllegalArgumentException("Domain '" + domainURI + "' not found");
```
cnathe:
```suggestion
                throw new IllegalArgumentException("Domain '" + domainURI + "' not found.");

```

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -501,6 +519,229 @@ public Object execute(DomainApiForm form, BindException errors)
...
+            if (!kind.canEditDefinition(user, domain))
+                throw new UnauthorizedException("You don't have permission to update import templates for this domain.");
+
+            List<Pair<String, String>> updatedTemplates = getUploadedTemplates(form, kind);
+
+            QuerySchema querySchema = DefaultSchema.get(user, container, form.getSchemaName());
+            if (!(querySchema instanceof UserSchema schema))
+                throw new NotFoundException("Could not find the specified schema in the folder '" + container.getPath() + "'");
```
cnathe:
```suggestion
                throw new NotFoundException("Could not find the specified schema in the folder '" + container.getPath() + "'.");

```

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -501,6 +519,229 @@ public Object execute(DomainApiForm form, BindException errors)
...
+            QuerySettings settings = schema.getSettings(getViewContext(), QueryView.DATAREGIONNAME_DEFAULT, form.getQueryName());
+            QueryDefinition queryDef = settings.getQueryDef(schema);
+            if (null == queryDef)
+                throw new NotFoundException("Could not find the specified query in the schema '" + form.getSchemaName() + "'");
```
cnathe:
```suggestion
                throw new NotFoundException("Could not find the specified query in the schema '" + form.getSchemaName() + "'.");

```

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -501,6 +519,229 @@ public Object execute(DomainApiForm form, BindException errors)
...
+            if (!queryDef.isMetadataEditable())
+                throw new UnsupportedOperationException("Query metadata is not editable.");
+            TableInfo tInfo = queryDef.getTable(schema, new ArrayList<>(), true, true);
+            if (tInfo == null)
+                throw new NotFoundException("Could not find the specified query in the schema '" + form.getSchemaName() + "'");
```
cnathe:
```suggestion
                throw new NotFoundException("Could not find the specified query in the schema '" + form.getSchemaName() + "'.");

```

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -501,6 +519,229 @@ public Object execute(DomainApiForm form, BindException errors)
...
+
+            List<Pair<String, String>> existingTemplates = tInfo.getImportTemplates(getViewContext());
+            List<Pair<String, String>> existingCustomTemplates = new ArrayList<>();
+            for (Pair<String, String> template_ : existingTemplates)
+            {
+                if (!template_.second.toLowerCase().contains("exportexceltemplate"))
+                    existingCustomTemplates.add(template_);
+            }
+            if (!updatedTemplates.equals(existingCustomTemplates))
```
cnathe: since these vars are both ArrayLists, will they ever be considered equal? From the UI, I opened the manage templates modal for a sample type that already had 2 custom tempaltes > edited the label for one and then put the label back to what it was > clicked Save. The code still went into this if statement.

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -501,6 +519,229 @@ public Object execute(DomainApiForm form, BindException errors)
...
+                        newTemplateBuilder.append("\" url=\"");
+                        newTemplateBuilder.append(template_.second);
+                        newTemplateBuilder.append("\">\n\t\t\t</template>\n");
+                    }
+                    newTemplateBuilder.append("\t\t</importTemplates>\n");
+
+                    String newTemplateStr = newTemplateBuilder.toString();
+                    final String insertLocator = "tableDbType=\"NOT_IN_DB\">";
+                    metadata = metadata.replace(insertLocator, insertLocator + newTemplateStr);
+                }
```
cnathe: I don't love this string replacement in the XML. Does it not work to do something like what MetadataTableJSON.java does where it uses `parseDocument(queryDef.getMetaData())` to get the current document > call something like `xmlTable.setImportTemplates(...)` to set the new values > then do the `queryDef.setMetadataXml(metadata)` and save below?

cnathe (changes requested):

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
XingY: Good point. moved this action to query controller and updated to use xml parser/setting instead.

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
XingY: Hmm.. for this scenario described, it doesn't go inside the if block for me.

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
XingY: updated

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
XingY: moved to validate

api/src/org/labkey/api/util/FileUtil.java
XingY: good point. updated

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -8216,6 +8168,294 @@ public Object execute(ParseForm form, BindException errors) throws Exception
...
+            for (Pair<String, String> template_ : existingTemplates)
+            {
+                if (!template_.second.toLowerCase().contains("exportexceltemplate"))
+                    existingCustomTemplates.add(template_);
+            }
+            if (!updatedTemplates.equals(existingCustomTemplates))
+            {
+                QueryDef queryDef = null;
+                if (queryDef == null)
+                    queryDef = QueryManager.get().getQueryDef(container, schemaName, queryName, false);
```
cnathe: queryDef is always null here so not sure that if statement is needed. 
```suggestion
                QueryDef queryDef = QueryManager.get().getQueryDef(container, schemaName, queryName, false);

```

cnathe (approved):

PR #6168, opened 2024-12-21
Issue 51904: Strict-Transport-Security header when HTTPS is required

labkey-jeckels:
#### Rationale
We have many protections in place already to ensure that traffic is always over HTTPS. We can set this header to catch an edge case for an initial http:// request on a third-party link.

#### Changes
- Set a long-lived `Strict-Transport-Security` header when the server is already requiring HTTPS connections

### DIFF
--- a/api/src/org/labkey/api/security/AuthFilter.java
+++ b/api/src/org/labkey/api/security/AuthFilter.java
@@ -125,49 +125,51 @@ public class AuthFilter implements Filter
             }
         }
 
-        // No startup failure, so check for SSL redirection
-        boolean sslRequired = AppProps.getInstance().isSSLRequired();
-        if (!req.getScheme().equalsIgnoreCase("https") && sslRequired)
+        if (AppProps.getInstance().isSSLRequired())
         {
-            // We can't redirect posts (we'll lose the post body), so return an error code
-            if ("post".equalsIgnoreCase(req.getMethod()))
+            // No startup failure, so check for SSL redirection
+            if (!req.getScheme().equalsIgnoreCase("https"))
             {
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-adam @labkey-matthewb no need to rush to adopt this but it was easy so I figured I'd open the PR before the holidays

api/src/org/labkey/api/security/AuthFilter.java
```diff
@@ -164,6 +165,11 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             return;
         }
 
+        if (sslRequired)
```
labkey-matthewb: Consider adding a devMode check.  I believe the reason we don't already have this is because it is incredibly sticky.  If you do this on a dev machine and test SSL, you won't be able to use http: until next year or until you figure out how to make your browser forget this setting.

labkey-matthewb: Ask me how I know...

labkey-jeckels: That's a good idea. Update pushed.

labkey-matthewb (approved):
should we set the header even if the current request is http? (along with the redirect?)

labkey-jeckels:
> should we set the header even if the current request is http? (along with the redirect?)

We could, but I saw commentary that it's only supposed to be set for HTTPS responses. Apparently browsers ignore it for plain HTTP

labkey-matthewb:
> > should we set the header even if the current request is http? (along with the redirect?)
> 
> We could, but I saw commentary that it's only supposed to be set for HTTPS responses. Apparently browsers ignore it for plain HTTP

That makes sense actually.

PR #6174, opened 2024-12-26
Issue 51300: on assay run import, don't keep the primary file if the new run failed to save

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51300

When an assay import fails, we leave around the uploaded primary data file in the assaydata dir. We have code that cleans up any posted assay results files. This PR adds to that so that we clean up the primary file as well. Note that if the run is a re-run we do not do this cleanup as we don't want to remove a file that was already used in an existing run.

#### Changes
- saveExperimentRun catch update to delete primary file

### DIFF
--- a/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
+++ b/api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
@@ -448,12 +448,30 @@ public class DefaultAssayRunCreator<ProviderType extends AbstractAssayProvider> 
             AssayResultsFileWriter<?> resultsFileWriter = new AssayResultsFileWriter<>(context.getProtocol(), run, null);
             resultsFileWriter.cleanupPostedFiles(context.getContainer(), false);
 
+            cleanPrimaryFile(context);
+
             if (e instanceof ExperimentException)
                 throw (ExperimentException)e;
             else
                 throw new ExperimentException(e);
         }
         catch (BatchValidationException e)
+        {
+            throw new ExperimentException(e);
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
```diff
@@ -459,6 +461,22 @@ public ExpExperiment saveExperimentRun(
         }
     }
 
+    private void cleanPrimaryFile(AssayRunUploadContext<ProviderType> context) throws ExperimentException
+    {
+        try
+        {
+            // Issue 51300: don't keep the primary file if the new run failed to save
+            boolean isReRun = context.getReRunId() != null;
+            FileLike primaryFile = context.getUploadedData().get(AssayDataCollector.PRIMARY_FILE);
```
labkey-jeckels, line 470: Deleting this file seems good. Should we also delete other uploaded files? Luminex may have multiple. I'm not sure if assay run properties of type file would be in the map too.

labkey-jeckels (approved):

labkey-klum (approved):

api/src/org/labkey/api/assay/DefaultAssayRunCreator.java
cnathe: looks like run property files are not in this `getUploadedData()` map, but it does look like the multiple files in the luminex case are (i.e. key `__primaryFile__`, `__primaryFile__2`, etc.). I'll open another PR to handle this case.

PR #6175, opened 2024-12-27
Issue 51776: LKSM: Updating a sample type name and only changing casing errors

XingY:
#### Rationale
Issue [51776](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51776)

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6175
- https://github.com/LabKey/limsModules/pull/1032

#### Changes
- By pass existing dataclass/sample type check during renaming when only changing casing


labkey-susanh (approved):
Assay designs seem to already allow changing to a name with different casing. Interesting.

PR #6176, opened 2024-12-28
Issue 51770: Enable "Missing files count" metric by default

XingY:
#### Rationale
"Missing files count" metric was implemented as a system maintenance task that's disabled by default, that can be enabled from UI or as a startup property. This PR enables the task by default and drops the related startup property.

#### Related Pull Requests
[platform#5531](https://github.com/LabKey/platform/pull/5531)
[Dockerfile#100](https://github.com/LabKey/Dockerfile/pull/100)

#### Changes
-

### DIFF
--- a/experiment/src/org/labkey/experiment/ExperimentModule.java
+++ b/experiment/src/org/labkey/experiment/ExperimentModule.java
@@ -578,8 +578,6 @@ public class ExperimentModule extends SpringModule
 
         SystemMaintenance.addTask(new FileLinkMetricsMaintenanceTask());
 
-        FileLinkMetricsMaintenanceTask.populateStartupProperties();
-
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
This all looks fine, but it also makes me think that we should have startup properties for all maintenance tasks. I couldn't find any in either the code or the Admin Console page. Fine to wait until we have a need for them more generally, but hopefully at that point we can use a mechanism that covers them all instead of individual ones of interest. @labkey-adam

labkey-adam:
> This all looks fine, but it also makes me think that we should have startup properties for all maintenance tasks. I couldn't find any in either the code or the Admin Console page. Fine to wait until we have a need for them more generally, but hopefully at that point we can use a mechanism that covers them all instead of individual ones of interest. @labkey-adam

Agreed

PR #6179, opened 2024-12-30
Issues 51894 and 51895: Treat import alias columns as strings

labkey-susanh:
#### Rationale
Issues [51894](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51894) and [51895](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51895): When names of entities look like numbers, our type inference logic will convert them to numbers during a file import, which will strip off any leading zeros (and could convert names that look like scientific notation to integers or floats). We have logic to protect against this when the input columns begin with MaterialInputs or DataInputs and more obviously look like ancestor columns. This PR adds logic to check for lineage import alias names as well so we will treat these input columns as strings

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1034

#### Changes
- Add optional `lineageAliasNames` parameter to `configureLoader` and override the call to this method for data classes and sample types so they can provide their import aliases.

### DIFF
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -2602,8 +2602,10 @@ public class ExpDataIterators
             {
                 try (DataLoader loader = DataLoader.get().createLoader(typeData.dataFile, "text/plain", true, null, null))
                 {
+                    ExpSampleTypeImpl sampleType = SampleTypeServiceImpl.get().getSampleType(typeData.container, _user, typeData.tableInfo.getName());
+                    Set<String> aliasNames = new CaseInsensitiveHashSet(sampleType.getImportAliases().keySet());
                     // We do not need to configure the loader for renamed columns as that has been taken care of when writing the file.
-                    configureLoader(loader, typeData.tableInfo, null, true);
+                    configureLoader(loader, typeData.tableInfo, null, true, aliasNames);
                     updateService.loadRows(_user, typeData.container, loader, _context, null);
                 }
                 catch (SQLException | IOException e)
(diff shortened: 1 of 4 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2602,8 +2602,10 @@ private void _importPartition(TypeData typeData)
             {
                 try (DataLoader loader = DataLoader.get().createLoader(typeData.dataFile, "text/plain", true, null, null))
                 {
+                    ExpSampleTypeImpl sampleType = SampleTypeServiceImpl.get().getSampleType(typeData.container, _user, typeData.tableInfo.getName());
+                    Set<String> aliasNames = new CaseInsensitiveHashSet(sampleType.getImportAliases().keySet());
                     // We do not need to configure the loader for renamed columns as that has been taken care of when writing the file.
-                    configureLoader(loader, typeData.tableInfo, null, true);
+                    configureLoader(loader, typeData.tableInfo, null, true, aliasNames);
```
XingY, line 2611: Instead of querying and passing in aliasNames to configureLoader, what about adding a method to TableInfo instead. We already do that for required lineage, see `getAdditionalRequiredInsertColumns`. I think that would be much cleaner and future proof.

labkey-susanh: Hmm.  I don't know that it's much cleaner and is definitely a tradeoff. We need this only for lineage columns (won't even be checked if "allowLineageColumns" is false), and that's really only a thing for a few number of TableInfos. Is it really worth adding this somewhat specific method to the `TableInfo` class? If we did that we might look at adding a method that encapsulates all of the name checks done in `configureLoader`. I'm inclined to leave this as is (but I am going to change `configureLoader` to use the existing `ExperimentService.isInputOutputColumn` method).

XingY (approved):

PR #6182, opened 2024-12-31
Query Datasets

labkey-martyp:
#### Rationale
Live query snapshot datasets. Essentially an unprovisioned dataset backed by a live query.

#### Changes
- Create new dataset type that uses a LK query as it's data.
- Update dataset definition to save the required source query fields.
- Add virtual table backing the query dataset
- Update snapshot UI and add experimental flag.
- Update create snapshot API
- Add Query change listener
- Prevent any attempts to mutate this dataset
- Opt out of export

### DIFF
--- a/study/src/org/labkey/study/model/ContinuousDatasetDomainKind.java
+++ b/study/src/org/labkey/study/model/ContinuousDatasetDomainKind.java
@@ -15,7 +15,6 @@
  */
 package org.labkey.study.model;
 
-import org.labkey.api.data.Container;
 import org.labkey.api.exp.property.Domain;
 import org.labkey.api.security.User;
 import org.labkey.api.study.TimepointType;
@@ -62,11 +61,5 @@ public class ContinuousDatasetDomainKind extends DatasetDomainKind
         fields.addAll(DatasetDefinition.DEFAULT_ABSOLUTE_DATE_FIELDS);
 
         return Collections.unmodifiableSet(fields);
-    }
-
-    @Override
-    public boolean isProvisioned(Container container, String name)
-    {
-        return !isQueryDataset(container, name);
     }
 }
--- a/study/src/org/labkey/study/model/ContinuousDatasetDomainKind.java
+++ b/study/src/org/labkey/study/model/ContinuousDatasetDomainKind.java
@@ -15,6 +15,7 @@
  */
 package org.labkey.study.model;
 
+import org.labkey.api.data.Container;
 import org.labkey.api.exp.property.Domain;
 import org.labkey.api.security.User;
 import org.labkey.api.study.TimepointType;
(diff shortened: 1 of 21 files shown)


study/src/org/labkey/study/model/ContinuousDatasetDomainKind.java
```diff
@@ -62,4 +63,10 @@ public Set<String> getReservedPropertyNames(Domain domain, User user)
 
         return Collections.unmodifiableSet(fields);
     }
+
+    @Override
+    public boolean isProvisioned(Container container, String name)
```
labkey-matthewb: Hmm, this is interesting.  The ContinuousDomainKind _is_ provisioned, but query dataset does not have a domain do do domainkind...  Hmm.  Maybe make this static and add a comment?   or change calling code to ask the dataset, not the domainkind?  (I haven't look to see where this is called yet.)

study/src/org/labkey/study/query/QueryDatasetQueryChangeListener.java
```diff
@@ -0,0 +1,90 @@
...
+    {
+
+    }
+
+    @Override
+    public void queryChanged(User user, Container container, ContainerFilter scope, SchemaKey schema, @NotNull QueryProperty property, @NotNull Collection<QueryPropertyChange> changes)
+    {
+        if (property.equals(QueryProperty.Name))
+        {
+            for (QueryPropertyChange change : changes)
```
labkey-matthewb, line 32: nice

study/src/org/labkey/study/query/QueryDatasetTable.java
```diff
@@ -0,0 +1,126 @@
...
+        return Collections.emptyMap();
+    }
+
+    @Override
+    public QueryUpdateService getUpdateService()
+    {
+        User user = _userSchema.getUser();
+        if (!user.hasRootAdminPermission() && !hasPermission(user, ReadPermission.class))
+            return null;
+        return new QueryDatasetUpdateService(this);
```
labkey-matthewb: I think you can just return null here.

labkey-matthewb (approved):

study/src/org/labkey/study/model/ContinuousDatasetDomainKind.java
labkey-martyp: I moved this to the abstract DatasetDomainKind class. It does ultimately ask the dataset if it's a query dataset or not. It's called outside the study module so it doesn't really have any idea what kind of domain it is at that point, it's up to the domain kind to figure out if it's provisioned. I added a comment.

study/src/org/labkey/study/query/QueryDatasetTable.java
labkey-martyp: Ah yes, that is more concise. Done.

PR #6194, opened 2025-01-07
[LabKey CRI] Issue 51636: Update Duo to support Universal Prompt

labkey-bpatel:
#### Rationale
Update getRedirectURL() to return a URLHelper to support a redirect to the Duo-hosted Universal Prompt -- a change required for the changes in the related PR.

#### Related Pull Requests
* https://github.com/LabKey/premiumModules/pull/130

#### Changes
- Update getRedirectURL() to return a URLHelper.


devtools/src/org/labkey/devtools/authentication/TestSecondaryConfiguration.java
```diff
@@ -15,8 +16,15 @@ public TestSecondaryConfiguration(TestSecondaryProvider provider, Map<String, Ob
...
     {
-        return TestSecondaryController.getTestSecondaryURL(c, getRowId());
+        try
+        {
+            return new URLHelper(TestSecondaryController.getTestSecondaryURL(c, getRowId()).getURIString());
+        }
+        catch (URISyntaxException e)
+        {
+            throw new IllegalArgumentException(e.getMessage(), e);
+        }
```
labkey-adam: Why was this changed? ActionURL extends URLHelper, so it should have been fine to return an ActionURL.

labkey-bpatel: Ah good point, thanks! Change reverted.

labkey-adam: Not critical from a functional standpoint, of course, but why not revert completely? The method can return ActionURL as before and unused imports should be removed. A clean revert means the file doesn't appear in the PR. Same comment for `premiumModules` `TotpConfiguration`.

labkey-adam (approved):

labkey-bpatel (commented):
Works as expected - tested with SAML and Duo.

PR #6197, opened 2025-01-08
Improve schema upgrade logging and version enforcement

labkey-jeckels:
#### Rationale
A recent failed upgrade of an older-than-supported DB pointed out some shortcomings

#### Changes
- Log the starting version for schemas needing upgrade
- Check for the way that GitHub VCS URLs are being captured as module properties

### DIFF
--- a/api/src/org/labkey/api/module/ModuleLoader.java
+++ b/api/src/org/labkey/api/module/ModuleLoader.java
@@ -475,6 +475,15 @@ public class ModuleLoader implements MemTrackerListener
         }
     }
 
+    private record UpgradeInfo(String moduleName, double installedVersion)
+    {
+        @Override
+        public String toString()
+        {
+            return moduleName + " (from schema " + ModuleContext.formatVersion(installedVersion) + ")";
+        }
+    }
+
     /** Full web-server initialization */
     private void doInit(Execution execution) throws ServletException
     {
@@ -646,7 +655,7 @@ public class ModuleLoader implements MemTrackerListener
         upgradeLabKeySchemaInExternalDataSources();
 
         ModuleContext coreCtx;
-        List<String> modulesRequiringUpgrade = new LinkedList<>();
+        List<UpgradeInfo> modulesRequiringUpgrade = new LinkedList<>();
         List<String> additionalSchemasRequiringUpgrade = new LinkedList<>();
         List<String> downgradedModules = new LinkedList<>();
 
@@ -675,7 +684,7 @@ public class ModuleLoader implements MemTrackerListener
                     if (context.needsUpgrade(module.getSchemaVersion()))
                     {
                         context.setModuleState(ModuleState.InstallRequired);
-                        modulesRequiringUpgrade.add(context.getName());
+                        modulesRequiringUpgrade.add(new UpgradeInfo(context.getName(), context.getInstalledVersion()));
                         addModuleToLockFile(context.getName(), lockFile);
                     }
                     else
--- a/api/src/org/labkey/api/module/ModuleLoader.java
+++ b/api/src/org/labkey/api/module/ModuleLoader.java
@@ -780,11 +780,11 @@ public class ModuleLoader implements MemTrackerListener
     /**
      * Does this module live in a repository that's managed by LabKey Corporation?
      * @param module a Module
-     * @return true if the module's VCS URL is non-null and starts with one of the GitHub organizations that LabKey manages
+     * @return true if the module's VCS URL is non-null and includes "github.com:LabKey/"
      */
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -475,6 +475,15 @@ public void updateModuleDirectory(File dir, File archive)
         }
     }
 
+    private record UpgradeInfo(String moduleName, double installedVersion)
+    {
+        @Override
+        public String toString()
+        {
+            return moduleName + " (from schema " + ModuleContext.formatVersion(installedVersion) + ")";
```
```suggestion
            return moduleName + " (from schema version " + ModuleContext.formatVersion(installedVersion) + ")";

```

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -675,7 +684,7 @@ public void addStaticWarnings(@NotNull Warnings warnings, boolean showAllWarning
                     if (context.needsUpgrade(module.getSchemaVersion()))
                     {
                         context.setModuleState(ModuleState.InstallRequired);
-                        modulesRequiringUpgrade.add(context.getName());
+                        modulesRequiringUpgrade.add(new UpgradeInfo(context.getName(), context.getInstalledVersion()));
```
labkey-adam, line 688: FYI: installed version could be `null`... did you test bootstrap?

labkey-adam (approved):

api/src/org/labkey/api/module/ModuleLoader.java
labkey-jeckels: I didn't but TeamCity will.

`ModuleContext.getInstalledVersion()` returns a `double` so I'm not sure how I would handle nulls.

labkey-adam: Oh right... that method maps `null` to 0.000. `getSchemaVersion()` returns `null`.

api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -771,11 +781,11 @@ private void warnAboutDuplicateSchemas(Collection<ModuleContext> values)
     /**
      * Does this module live in a repository that's managed by LabKey Corporation?
      * @param module a Module
-     * @return true if the module's VCS URL is non-null and starts with one of the GitHub organizations that LabKey manages
+     * @return true if the module's VCS URL is non-null and includes "github.com:LabKey/"
      */
     private boolean isFromLabKeyRepository(Module module)
     {
-        return StringUtils.startsWithAny(module.getVcsUrl(), "https://github.com/LabKey/");
+        return StringUtils.containsAny(module.getVcsUrl(), "github.com:LabKey/");
```
labkey-susanh, line 788: Hmm.  This "url" is odd to me. Seems to be what is actually being spit out, so probably ok for current purposes, but do people actually use "urls" in this form to clone from github?

labkey-susanh (approved):

api/src/org/labkey/api/module/ModuleLoader.java
labkey-nicka: I know I do. My clone command looks like `git clone git@github.com:LabKey/platform.git`.

labkey-adam: Looks like this is the standard format used by TeamCity and therefore all the distributions it produces. ITN (running on Windows) shows this as well.

That said, I think we should stop relying on VCS URL entirely, getting rid of this method and the corresponding check. Dev mode deployments don't have a VCS URL and therefore don't check for old versions. This PR is fine for now (in fact, should this change be backported to 24.11?), but we should update to rely solely on the `ManageVersion` property going forward (on `develop`).

labkey-jeckels: I see `ManageVersion` is explicitly set to false in some client repos, but otherwise unset. If customers have custom modules that aren't following our year-based numbering scheme, wouldn't that prevent them from upgrading their own schemas?

labkey-adam: We need to add `ManageVersion: true` in all of "our" modules and change the default value to false, as I think you pointed out in the old issue. Not a trivial effort, but not that hard. I just did something similar for `SupportedDatabases`.

We probably want to set the property to true in the module template.

labkey-jeckels: I think I'll go ahead and merge this to develop as-is. It's an improvement over current behavior. And I am skeptical that it'll actually help anyone in the 24.11 branch. They'd have to be using new maintenance release and upgrading from something older than 23.1.

PR #6206, opened 2025-01-10
Issue 51572: Bug with creating a new list by uploading a csv file in "UTF-8 with BOM" format

labkey-bpatel:
#### Rationale
Issue [51572](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51572): Bug with creating a new list by uploading a csv file in "UTF-8 with BOM" format

#### Related Pull Requests
- 

#### Changes
- Remove the BOM character from column names in DataLoader when inferring columns from the uploaded file.

### DIFF
--- a/api/src/org/labkey/api/reader/DataLoader.java
+++ b/api/src/org/labkey/api/reader/DataLoader.java
@@ -425,7 +425,14 @@ public abstract class DataLoader implements Iterable<Map<String, Object>>, Loade
         {
             String[] headers = lineFields[_skipLines - 1];
             for (int f = 0; f < nCols; f++)
-                colDescs[f].name = (f >= headers.length || StringUtils.isBlank(headers[f])) ? getDefaultColumnName(f) : headers[f].trim().replaceAll("[\\t\\n\\r]+"," ");
+            {
+                colDescs[f].name = (f >= headers.length || StringUtils.isBlank(headers[f])) ? getDefaultColumnName(f) : headers[f].trim().replaceAll("[\\t\\n\\r]+", " ");
+
+                //Issue 51572: Bug with creating a new list by uploading a csv file in "UTF-8 with BOM" format
+                //If the first character of the column name is the BOM character, remove it
+                if (colDescs[f].name.startsWith("\uFEFF"))
+                    colDescs[f].name = colDescs[f].name.replace("\uFEFF", "");
+            }
         }
         else
         {


labkey-adam (changes requested):
Aren't we using a `BOMInputStream` in this code path? i.e., getting our `Reader` from `Readers.getBOMDetectingReader()`? Ignoring the BOM (and, for that matter, only handling a UTF-8 BOM) seems wrong. I haven't debugged the failure case, but I can if you want.

labkey-bpatel:
> Aren't we using a `BOMInputStream` in this code path? i.e., getting our `Reader` from `Readers.getBOMDetectingReader()`? Ignoring the BOM (and, for that matter, only handling a UTF-8 BOM) seems wrong. I haven't debugged the failure case, but I can if you want.

Indeed, we are using BOMInputStream(), however, it seems like BOMInputStream is not stripping the UTF BOM from the input stream as expected in the use case provided in the ticket. It does seem to strip the others. What would be the right course of action here? Any insight is appreciated, thanks!

labkey-matthewb (changes requested):
I'm pretty sure this is not the proper way to handle this.  The BOM sequences [ 0xff, 0xfe ] and [ 0xfe, 0xff ] should be handled when scanning the byte steam not in the decoded char stream.  

The problem might be with the file encoding?

labkey-bpatel:
> > Aren't we using a `BOMInputStream` in this code path? i.e., getting our `Reader` from `Readers.getBOMDetectingReader()`? Ignoring the BOM (and, for that matter, only handling a UTF-8 BOM) seems wrong. I haven't debugged the failure case, but I can if you want.
> 
> Indeed, we are using BOMInputStream(), however, it seems like BOMInputStream is not stripping the UTF BOM from the input stream as expected in the use case provided in the ticket. It does seem to strip the others. What would be the right course of action here? Any insight is appreciated, thanks!



> I'm pretty sure this is not the proper way to handle this. The BOM sequences [ 0xff, 0xfe ] and [ 0xfe, 0xff ] should be handled when scanning the byte steam not in the decoded char stream.
> 
> The problem might be with the file encoding?

Is it possible that the source file already contains a Unicode BOM character (\uFEFF) in addition to the byte-level BOM? it seems that way since the \uFEFF remains in the decoded text even after the BOM bytes are stripped. Yeah could be the problem with the file encoding.

labkey-adam:
> > > Aren't we using a `BOMInputStream` in this code path? i.e., getting our `Reader` from `Readers.getBOMDetectingReader()`? Ignoring the BOM (and, for that matter, only handling a UTF-8 BOM) seems wrong. I haven't debugged the failure case, but I can if you want.
> > 
> > 
> > Indeed, we are using BOMInputStream(), however, it seems like BOMInputStream is not stripping the UTF BOM from the input stream as expected in the use case provided in the ticket. It does seem to strip the others. What would be the right course of action here? Any insight is appreciated, thanks!
> 
> > I'm pretty sure this is not the proper way to handle this. The BOM sequences [ 0xff, 0xfe ] and [ 0xfe, 0xff ] should be handled when scanning the byte steam not in the decoded char stream.
> > The problem might be with the file encoding?
> 
> Is it possible that the source file already contains a Unicode BOM character (\uFEFF) in addition to the byte-level BOM? it seems that way since the \uFEFF remains in the decoded text even after the BOM bytes are stripped. Yeah could be the problem with the file encoding.

I think the problem is we're not using `Readers.getBOMDetectingReader()` in all code paths. We use it for a file input (see `TabLoader(final File inputFile...)`) but not for an `InputStream` (see `TsvFactory` and `CsvFactory`).

PR #6211, opened 2025-01-10
Check all managed modules for upgrade suitability

labkey-adam:
#### Rationale
Previously, we excluded from the upgrade check modules with a VCS URL that was missing or in a format we didn't understand. The URL format wasn't consistent across servers and VCS URLs weren't present on dev machines by default, making the check unreliable (or maybe worthless). We required the VCS URL check because the `ManageVersion` property defaulted to true... so, absent that check, an external module with a non-standard schema version that didn't specify ManageVersion would block the upgrade. We've now switched the default value to false, so ManageVersion solely determines whether we do the upgrade suitability check. All deployments should now consistently fail to startup if the installed version of any managed module is less than the earliest upgrade version (currently 24.000). https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=47369

Also, improve the information included in the upgrade blocked exception message.

#### Related Pull Requests
- https://github.com/LabKey/server/pull/964

### DIFF
--- a/api/src/org/labkey/api/module/ModuleLoader.java
+++ b/api/src/org/labkey/api/module/ModuleLoader.java
@@ -582,35 +582,31 @@ public class ModuleLoader implements MemTrackerListener
         {
             // Refuse to start up if any LabKey-managed module has a schema version that's too old. Issue 46922.
 
-            // Modules that are designated as "managed" and reside in LabKey-managed repositories
+            // Collect all modules that are designated as "managed"
             var labkeyModules = _modules.stream()
                 .filter(Module::shouldManageVersion)
-                .filter(this::isFromLabKeyRepository) // Do the check only for modules in LabKey repositories, Issue 47369
                 .toList();
 
-            // Likely empty if running in dev mode... no need to log or do other work
-            if (!labkeyModules.isEmpty())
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/module/ModuleLoader.java
```diff
@@ -582,35 +582,31 @@ private void doInit(Execution execution) throws ServletException
...
+                .filter(ctx -> ctx.getSchemaVersion() != null)
+                .collect(Collectors.toMap(ModuleContext::getName, ctx->ctx));
+
+            // List of "<name> (<installedSchemaVersion>)" of LabKey-managed modules with schemas where the installed
+            // version is less than "earliest upgrade version"
+            var tooOld = labkeyModules.stream()
+                .map(m -> moduleContextMap.get(m.getName()))
+                .filter(Objects::nonNull)
+                .filter(ctx -> ctx.getInstalledVersion() < Constants.getEarliestUpgradeVersion())
+                .map(ctx -> ctx.getName() + " (" + ModuleContext.formatVersion(ctx.getInstalledVersion()) + ")")
```
labkey-jeckels: Consider using the `UpgradeInfo` record I added

labkey-jeckels (approved):

api/src/org/labkey/api/module/ModuleLoader.java
labkey-adam: Right, lovely

PR #6213, opened 2025-01-13
Throw if MapDataIterator finds duplicate columns

labkey-matthewb:
#### Rationale
Avoid ambiguous condition.
DataIterator does not prevent duplicate columns, but in the case of MapDataIterator this would lead to ambiguous/undefined behavior.

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/dataiterator/MapDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/MapDataIterator.java
@@ -25,7 +25,9 @@ import org.labkey.api.collections.CaseInsensitiveHashSet;
 import org.labkey.api.collections.CaseInsensitiveTreeSet;
 import org.labkey.api.data.ColumnInfo;
 import org.labkey.api.query.BatchValidationException;
+import org.labkey.api.util.ExceptionUtil;
 import org.labkey.api.util.logging.LogHelper;
+import org.labkey.api.view.HttpView;
 
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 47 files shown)


api/src/org/labkey/api/dataiterator/MapDataIterator.java
```diff
@@ -86,7 +86,7 @@ public MapDataIteratorImpl(DataIterator in, boolean mutable, Set<String> skip)
                 _findMap.put(in.getColumnInfo(i).getName(),i);
             }
             if (null != duplicates)
-                LOGGER.warn("Data has duplicate columns: '" + StringUtils.join(duplicates.toArray(), ", ") + "'");
+                throw new IllegalStateException("Data has duplicate columns: '" + StringUtils.join(duplicates.toArray(), ", ") + "'");
```
labkey-matthewb: alternately change to LOGGER.error() if we think test coverage might miss some of these?

labkey-jeckels: I like changing this to an `ERROR` level log for now. That will still cause tests to fail when they check the log at the end. We can set a Datadog alert so we get notified of any examples on servers that are streaming logs.

If you want to be more aggressive, we could make it throw `IllegalStateException` but have an experimental feature that knocks it down to `ERROR` logging. That gives customers an escape hatch if we missed use cases, but makes it more likely that they'd get reported to us even if they're on-premise.

labkey-jeckels: Either way, after we go through a ESR version we can bump it to always throw the exception

labkey-jeckels (approved):
Looks good

PR #6217, opened 2025-01-14
Issue 51933: Assure we remove all invisible characters from header text

labkey-susanh:
#### Rationale
Issue [51933](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51933)  the `trim()` method does not remove non-breaking spaces and some other invisible characters, but these characters are not (or should not be) used as characters in the names of fields. We replace a larger set of invisible characters with a space, and then trim the resulting string.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2229

#### Changes
- Update `DataLoader`'s handling of whitespace characters.

### DIFF
--- a/api/src/org/labkey/api/reader/DataLoader.java
+++ b/api/src/org/labkey/api/reader/DataLoader.java
@@ -430,7 +430,8 @@ public abstract class DataLoader implements Iterable<Map<String, Object>>, Loade
         {
             String[] headers = lineFields[_skipLines - 1];
             for (int f = 0; f < nCols; f++)
-                colDescs[f].name = (f >= headers.length || StringUtils.isBlank(headers[f])) ? getDefaultColumnName(f) : headers[f].trim().replaceAll("[\\t\\n\\r]+"," ");
+                // Issue 51933: the trim() method does not trim off any non-breaking spaces, so we'll do a replacement using the larger regex class of invisible characters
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/reader/DataLoader.java
```diff
@@ -430,7 +430,8 @@ else if (renamedColumns.containsKey(name) && _columnInfoMap.containsKey(renamedC
         {
             String[] headers = lineFields[_skipLines - 1];
             for (int f = 0; f < nCols; f++)
-                colDescs[f].name = (f >= headers.length || StringUtils.isBlank(headers[f])) ? getDefaultColumnName(f) : headers[f].trim().replaceAll("[\\t\\n\\r]+"," ");
+                // Issue 51933: the trim() method does not trim off any non-breaking spaces, so we'll do a replacement using the larger regex class of invisible characters
+                colDescs[f].name = (f >= headers.length || StringUtils.isBlank(headers[f])) ? getDefaultColumnName(f) : headers[f].replaceAll("[\\p{Z}\\t\\n\\r]+", " ").trim();
```
XingY, line 434: Is the same treatment needed for `inferColumnInfo`?

XingY (approved):

api/src/org/labkey/api/reader/DataLoader.java
labkey-susanh: Good call. Thank you. We seem to also not be trimming the names over there, so I've added that as well.

PR #6219, opened 2025-01-14
Issue 51908: Increase precision for display and storage of stored amounts to nanoliters and nanograms

labkey-susanh:
#### Rationale
Issue [51908](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51908): The request here is to allow values to be stored with a precision of nanoliter or nanogram, but not to add these as explicit unit types (yet).

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1690
- https://github.com/LabKey/labkey-ui-premium/pull/644
- https://github.com/LabKey/limsModules/pull/1076

#### Changes
- Add a precision scale for the different unit kinds
- Use precision scale from unit kind when available

### DIFF
--- a/api/src/org/labkey/api/data/measurement/Measurement.java
+++ b/api/src/org/labkey/api/data/measurement/Measurement.java
@@ -13,6 +13,8 @@ import java.util.regex.Pattern;
 
 public class Measurement
 {
+    public static final int DEFAULT_PRECISION_SCALE = 6;
+
     private static final String NUMBER_REGEX = "[+\\-]?\\d+(?:\\.\\d+)?(?:[Ee][+\\-]\\d+)?";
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/data/measurement/Measurement.java
```diff
@@ -21,9 +23,21 @@ public class Measurement
 
     public enum Kind
     {
-        Mass,
-        Volume,
-        Count
+        Mass(12),
```
XingY: Should precision be associated with `Unit` instead of `Kind`?

labkey-susanh: Yeah, that mimics what we do for client-side as well.

XingY (approved):

PR #6221, opened 2025-01-15
Teach TabLoader to always respect InputStream BOMs

labkey-adam:
#### Rationale
Text files with BOMs were failing infer fields and therefore import: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51572

#### Changes
- Introduce new `TabLoader` constructor that takes an `InputStream` and always wraps it with a BOM-detecting InputStream
- Use that constructor from `TsvFactory` and `CsvFactory`
- Use commons.io builders instead of deprecated constructors
- Clean up a few other IntelliJ warnings

### DIFF
--- a/api/src/org/labkey/api/reader/TabLoader.java
+++ b/api/src/org/labkey/api/reader/TabLoader.java
@@ -81,7 +81,7 @@ public class TabLoader extends DataLoader
             return new TabLoader(file, hasColumnHeaders, mvIndicatorContainer);
         }
 
-        /** A DataLoader created with this constructor does NOT close the InputStream */
+        /** This constructor does NOT close the InputStream. Call close(), in a try-with-resources, if appropriate, to ensure resources are released.  */
         @NotNull @Override
         public DataLoader createLoader(InputStream is, boolean hasColumnHeaders, Container mvIndicatorContainer) throws IOException
         {
--- a/api/src/org/labkey/api/reader/TabLoader.java
+++ b/api/src/org/labkey/api/reader/TabLoader.java
@@ -19,7 +19,6 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.io.input.CharSequenceReader;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.text.StringEscapeUtils;
-import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


labkey-adam:
FWIW: Couldn't help but listen to this while coding up the fix: https://www.youtube.com/watch?v=MUaL1FnotRQ&ab_channel=RATMVEVO

api/src/org/labkey/api/reader/TabLoader.java
```diff
@@ -82,11 +81,11 @@ public DataLoader createLoader(File file, boolean hasColumnHeaders, Container mv
             return new TabLoader(file, hasColumnHeaders, mvIndicatorContainer);
         }
 
-        /** A DataLoader created with this constructor does NOT close the reader */
+        /** A DataLoader created with this constructor does NOT close the InputStream */
```
labkey-jeckels: I've always found this guidance a little confusing. Does this seem accurate and clearer?
```suggestion
        /** This constructor does NOT close the InputStream. Call close(), in a try-with-resources, if appropriate, to ensure resources are released.  */
```

api/src/org/labkey/api/reader/TabLoader.java
```diff
@@ -1186,7 +1189,7 @@ public void testMySql() throws IOException
 
             final List<Map<String, Object>> rows;
 
-            try (TabLoader loader = (TabLoader)new TabLoader.MysqlFactory().createLoader(new StringBufferInputStream(mysqlData), false, null))
+            try (TabLoader loader = (TabLoader)new TabLoader.MysqlFactory().createLoader(new ByteArrayInputStream(mysqlData.getBytes(StringUtilsLabKey.DEFAULT_CHARSET)), false, null))
```
labkey-jeckels: Can we add a test that parses a BOM file?

labkey-jeckels (approved):
Two suggestions

api/src/org/labkey/api/reader/TabLoader.java
labkey-adam: Yes, working on it

api/src/org/labkey/api/reader/TabLoader.java
labkey-adam: Sure. The guidance is a bit self-evident at this point, since `TabLoader` is `AutoCloseable`. But doesn't hurt to be explicit.

labkey-bpatel:
Are we also planning to make other callers of TabLoader(Reader...) also use BOMInputStream as part of this change or that will be handled separately? Like the ones we had looked at in SqlScriptExecutor.java among others.

labkey-bpatel:
> FWIW: Couldn't help but listen to this while coding up the fix: https://www.youtube.com/watch?v=MUaL1FnotRQ&ab_channel=RATMVEVO

🎧 🎸 definitely a serious coding tune

labkey-adam:
> Are we also planning to make other callers of TabLoader(Reader...) also use BOMInputStream as part of this change or that will be handled separately? Like the ones we had looked at in SqlScriptExecutor.java among others.

I reviewed all callers of the `Reader` constructors. There are certainly some callers that turn text InputStreams into Readers without supporting BOMs, but they all seemed to me to be very specific formats that are unlikely to include a BOM. What I saw were standard machine-generated formats and human-generated formats like LabKey SQL scripts, neither of which have seen BOMs in the past. This contrasts with our generic TSV and CSV import code paths which can see files generated from a wide variety of apps and systems.

labkey-bpatel (approved):

api/src/org/labkey/api/reader/TabLoader.java
labkey-adam: I've added a beautiful new junit test that exercises all the encodings + BOMs for both CSV and TSV formats. Adding a selenium test case to exercise infer + import through the domain designer would be a good addition. I'll make sure the issue notes this.

PR #6226, opened 2025-01-15
Replace GraalJS-based Markdown renderer with Commonmark Java

labkey-adam:
#### Rationale
Using GraalJS to execute a JavaScript implementation of Markdown is an unnecessarily unwieldy rendering approach. GraalJS adds 60MB of code to our distributions and presents problems since it includes some native code. [commonmark-java](https://github.com/commonmark/commonmark-java) provides a native Java implementation of Markdown (actually, of the [Commonmark standard specification](https://github.com/commonmark/commonmark-java/blob/main/commonmark-test-util/src/main/resources/spec.txt)) that is highly compatible with our GraalJS implementation, much simpler to integrate (no pooling and no exceptions at webapp reload), and 1/300 the size. It also includes an extension mechanism allowing us to extend and customize the implementation.

This initial integration includes the base commonmark-java implementation plus three extensions (Autolink, Strikethrough, and Tables) that bring us close to parity with the previous markdown-it.js implementation. I've also added the extensions for Heading Anchor (automatically adds an anchor to every heading element) and Image Attributes (allows arbitrary attributes on images, e.g., width and height), requested by the LabKey docs team.


labkey-jeckels (approved):
Code looks good but we should get some more manual testing on wikis and message forum examples.

labkey-adam:
> Code looks good but we should get some more manual testing on wikis and message forum examples.

Molly has volunteered; I'll merge to develop so she can access the new impl.

PR #6227, opened 2025-01-16
Make ExcelLoader.getSheet public

labkey-martyp:
#### Rationale
Make getSheet() available for handling of .xls files beyond what getFirstNLinesXLS offers.

#### Changes
- Make ExcelLoader.getSheet public

### DIFF
--- a/api/src/org/labkey/api/reader/ExcelLoader.java
+++ b/api/src/org/labkey/api/reader/ExcelLoader.java
@@ -303,7 +303,7 @@ public class ExcelLoader extends DataLoader
     }
 
 
-    private Sheet getSheet() throws IOException
+    public Sheet getSheet() throws IOException
     {
         try
         {


labkey-jeckels:
I'm not opposed to this change, but what's the use case? If the caller needs access to the `Sheet` are they getting value out of using `ExcelLoader` instead of just getting the `Workbook` from `ExcelFactory.create()`?

labkey-martyp:
> I'm not opposed to this change, but what's the use case? If the caller needs access to the `Sheet` are they getting value out of using `ExcelLoader` instead of just getting the `Workbook` from `ExcelFactory.create()`?

I don't think they'll be using the iterator or anything but it prevents them from having to copy the getSheet function to their local code and has the option to specify sheet name or sheet index.

labkey-jeckels (approved):

PR #6231, opened 2025-01-17
Better consistency between ApiKeys and UserApiKeys tables

labkey-adam:
#### Rationale
Testing pointed out that the API Keys table wasn't wired up to use the custom query update service that the User API Keys table uses. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52004

The primary difference between these two tables is a filter on the current user, so combine the two classes for consistency.


core/src/org/labkey/core/query/ApiKeysTableInfo.java
```diff
@@ -17,27 +17,41 @@
...
         addWrapColumn(getRealTable().getColumn("Created"));
         addWrapColumn(getRealTable().getColumn("Expiration"));
         addWrapColumn(getRealTable().getColumn("LastUsed"));
         addWrapColumn(getRealTable().getColumn("Description"));
+
+        if (filterToCurrentUser)
+        {
+            setTitle("User API Keys");
+            createdBy.setHidden(true);
+            addCondition(new SimpleFilter(FieldKey.fromParts("CreatedBy"), schema.getUser().getUserId(), CompareType.EQUAL));
```
labkey-jeckels:
```suggestion
            addCondition(new SimpleFilter(createdBy.getFieldKey(), schema.getUser().getUserId(), CompareType.EQUAL));

```

labkey-jeckels (approved):

labkey-susanh (approved):

PR #6234, opened 2025-01-20
Add support for conditional formatting in LIMS and SDMS products

labkey-susanh:
#### Rationale
In LKS, users have the ability to configure fields with conditional formatting so they will show in different fonts or colors in grids. We want to extend this ability to our applications so they will show the formatting in grids but also in detail pages.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1694
- https://github.com/LabKey/limsModules/pull/1086
- https://github.com/LabKey/labkey-ui-premium/pull/648

#### Changes
- Add ConditionalFormatting product feature


api/src/org/labkey/api/action/ExtendedApiQueryResponse.java
```diff
@@ -55,6 +55,7 @@ public enum ColMapEntry
         value,
         displayValue,
         formattedValue,
+        hoverContent,
```
cnathe: I think this can be removed, since the decision was to not include the hover details at this time.

cnathe (approved):

api/src/org/labkey/api/action/ExtendedApiQueryResponse.java
labkey-susanh: Done.

PR #6238, opened 2025-01-22
Fix several publish study issues

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52025

#### Changes
- Create a helper method that determines the StartDate dataset & column. Use that method in update participant and publish study "extra dataset" code paths.
- DatasetDefinition.isQuerySnapshot()
- Address query logging exceptions during publish study
- Don't fail publish / import if security policy refers to an unknown dataset: best effort and just log a warning


study/src/org/labkey/study/visitmanager/VisitManager.java
```diff
@@ -603,39 +604,57 @@ private static SQLFragment studyDataPtids(Collection<DatasetDefinition> defs)
...
+    public record StartDateInfo(DatasetDefinition dataset, TableInfo storageTableInfo, ColumnInfo startDateColumn){}
+
+    // The first non-query and non-query-snapshot demographics dataset with a "StartDate" column is the start-date dataset
+    //TODO: Use Property URI & Make sure "StartDate" is set properly
+    public static @Nullable StartDateInfo getStartDateInfo(StudyImpl study)
+    {
+        return study.getDatasets().stream()
+            .filter(DatasetDefinition::isDemographicData)
+            .filter(dataset -> !dataset.isQueryDataset())
+            .filter(dataset -> !dataset.isQuerySnapshot())
```
labkey-martyp: You could probably combine a couple of these filters to reduce cycles, 

`.filter(dataset -> !dataset.isQueryDataset() && !dataset.isQuerySnapshot())`

study/src/org/labkey/study/visitmanager/VisitManager.java
```diff
@@ -603,39 +604,57 @@ private static SQLFragment studyDataPtids(Collection<DatasetDefinition> defs)
...
+            .map(dataset -> {
+                TableInfo ti = dataset.getStorageTableInfo();
```
labkey-martyp: This would essentially exclude the query datasets since this would return null, but doesn't hurt to filter them out earlier.

labkey-martyp (approved):
Looks good to me

study/src/org/labkey/study/visitmanager/VisitManager.java
labkey-matthewb: Consider dataset.getDomain().getPropertyByName("StartDate")
The domain is already sitting in a cache somewhere and doesn't need to be constructed.

labkey-matthewb: I see that  this is not new code.

study/src/org/labkey/study/pipeline/DatasetImportRunnable.java
```diff
@@ -333,16 +333,20 @@ public void run()
...
+            for (ValidationException err : batchErrors.getRowErrors())
+                _logger.error(_fileName + " -- " + err.getMessage());
+
+            if (batchErrors.hasErrors())
+                throw batchErrors;
         }
         catch (Exception x)
         {
-            _logger.error("Exception while importing dataset " + _datasetDefinition.getName() + " from " + _fileName, x);
+            throw new RuntimeException("Exception while importing dataset " + _datasetDefinition.getName() + " from " + _fileName, x);
```
labkey-matthewb: This looks cleaner, but  I'm not sure that qus.importRows() always returns normally when it encounters "expected" errors.  I think it sometimes it may throw the batchErrors object.  I need to look at some other code.

labkey-matthewb: See the comments for QueryUpdateService.insertRows() and this DataIteratorContext.checkShouldCancel()

labkey-adam: @labkey-matthewb please let me know what you find

labkey-matthewb: Yup see my follow on comment  You have to also catch BatchValidationException (I realize this is not the best pattern).  The easiest thing to do is probably to check batchErrors.hasErrors() and throw if true.  Then a single catch (BatchValidationException) will cosolidate the error handling.

study/src/org/labkey/study/visitmanager/VisitManager.java
labkey-adam: Revised to inspect domain. The new caller doesn't care about the provisioned table... and the old one only cares about it for the single dataset.

study/src/org/labkey/study/pipeline/DatasetImportRunnable.java
labkey-adam: @labkey-matthewb any thoughts?

labkey-matthewb: Just that it would be really nice if we fixed QUS to always throw or always not throw BVE

labkey-matthewb (approved):

PR #6239, opened 2025-01-22
Add numbered cells for terminal storage

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1704
- https://github.com/LabKey/labkey-ui-premium/pull/650
- https://github.com/LabKey/platform/pull/6239
- https://github.com/LabKey/limsModules/pull/1089

#### Changes
- add InventoryStatusColumn.StoragePositionNumber
- exclude hashCode method from object.toMap


api/src/org/labkey/api/data/BuilderObjectFactory.java
```diff
@@ -112,7 +112,7 @@ public BuilderObjectFactory(@NotNull Class<K> classReturn, @NotNull Class<? exte
                 name = name.substring(3);
             else if (name.startsWith("is"))
                 name = name.substring(2);
-            else if (name.startsWith("has"))
+            else if (name.startsWith("has") && !name.equals("hashCode"))
```
labkey-susanh, line 115: What prompted this change? Just curious.

labkey-susanh (approved):

api/src/org/labkey/api/data/BuilderObjectFactory.java
XingY: I was seeing hashCode in the detailed audit log record maps. Not sure if my changes caused that to appear, but I figure it's a safe assumption we never want this.

PR #6241, opened 2025-01-22
don't duplicate rowNumber column in ListOfMapsDataIterator constructor

labkey-matthewb:
#### Rationale
don't duplicate rowNumber column in ListOfMapsDataIterator constructor
use ROWNUMBER_COLUMNNAME in more places
fix JHU_Test

#### Related Pull Requests
- 

#### Changes
-


labkey-jeckels (approved):
Change looks fine but needs adjustment to avoid the immutable collection exception in the BVTs.

PR #6243, opened 2025-01-23
Details page doesn't need hidden PK fields

labkey-adam:
#### Rationale
Hidden fields on the DataRegion details page are unused; might as well stop rendering them. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=52071

### DIFF
--- a/api/src/org/labkey/api/data/DataRegion.java
+++ b/api/src/org/labkey/api/data/DataRegion.java
@@ -193,7 +193,7 @@ public class DataRegion extends DisplayElement
      * Messages that are displayed to the user and included in Query API responses.
      * These messages' content should NOT be HTML encoded as the responsibility
      * for encoding is left to the caller.
-     * See https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42017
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/data/DataRegion.java
```diff
@@ -446,19 +446,6 @@ public void addHiddenFormField(String name, String value)
             _hiddenFormFields.add(Pair.of(name, value));
     }
 
-    public String getHiddenFormFieldValue(String name)
```
labkey-nicka, line 449: nit: Just curious, do you know when this was rendered useless? It'd be interesting to see where this was missed in a previous refactor / code change.

labkey-nicka (approved):

api/src/org/labkey/api/data/DataRegion.java
labkey-adam: I do not. 22.11 is the oldest version I have locally and the method is unused there. (As is `setInputPrefix()`.)

PR #6245, opened 2025-01-23
Configure Markdown renderer to escape and sanitize

labkey-adam:
#### Rationale
We don't want HTML tags in Markdown

### DIFF
--- a/announcements/src/org/labkey/announcements/AnnouncementsController.java
+++ b/announcements/src/org/labkey/announcements/AnnouncementsController.java
@@ -802,7 +802,7 @@ public class AnnouncementsController extends SpringActionController
             List<AttachmentFile> files = getAttachmentFileList();
 
             AnnouncementModel insert = form.getBean();
-            if (null == insert.getParent() || 0 == insert.getParent().length())
+            if (null == insert.getParent() || insert.getParent().isEmpty())
                 insert.setParent(form.getParentId());
 
             try
--- a/announcements/src/org/labkey/announcements/model/AnnouncementManager.java
+++ b/announcements/src/org/labkey/announcements/model/AnnouncementManager.java
@@ -522,7 +522,7 @@ public class AnnouncementManager
 
                 body = PageFlowUtil.validateHtml(body, validateErrors, hasDeveloperPermission);
 
-                if (validateErrors.size() > 0)
+                if (!validateErrors.isEmpty())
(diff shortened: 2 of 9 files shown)


labkey-klum (approved):

labkey-tchad (changes requested):
`MarkdownTestCase` needs an update. It's expecting the HTML to be unescaped

labkey-adam:
> `MarkdownTestCase` needs an update. It's expecting the HTML to be unescaped

Surprising... this suggests that the previous implementation allowed script & html tags. Those test cases are unchanged.

labkey-tchad:
> > `MarkdownTestCase` needs an update. It's expecting the HTML to be unescaped
> 
> Surprising... this suggests that the previous implementation allowed script & html tags. Those test cases are unchanged.

Looks like the previous implementation of `MarkdownService.toHtml` allowed callers to pass in a list of options, including a flag to enable html. The default wiki and announcements didn't use it. The only places that did were that test and `IpynbOutput`

labkey-adam:
> > > `MarkdownTestCase` needs an update. It's expecting the HTML to be unescaped
> > 
> > 
> > Surprising... this suggests that the previous implementation allowed script & html tags. Those test cases are unchanged.
> 
> Looks like the previous implementation of `MarkdownService.toHtml` allowed callers to pass in a list of options, including a flag to enable html. The default wiki and announcements didn't use it. The only places that did were that test and `IpynbOutput`

Yep, I saw the same, so that explains that mystery. Current implementation offers no option to "pass through HTML tags"... do we need to add that option for `IpynbOutput`?

labkey-tchad:
> Yep, I saw the same, so that explains that mystery. Current implementation offers no option to "pass through HTML tags"... do we need to add that option for `IpynbOutput`?

It depends on how folks are using Jupyter reports. I compared our test report from `JupyterOutputTest` and it looks identical in 24.11 but that's because our markdown example doesn't have any HTML in it.

labkey-tchad (approved):

labkey-adam:
> > Yep, I saw the same, so that explains that mystery. Current implementation offers no option to "pass through HTML tags"... do we need to add that option for `IpynbOutput`?
> 
> It depends on how folks are using Jupyter reports. I compared our test report from `JupyterOutputTest` and it looks identical in 24.11 but that's because our markdown example doesn't have any HTML in it.

Okay, let's move forward with no HTML tags in Markdown across the board. If a client complains, it should be easy to enable for Jupyter, as we had before.

PR #6247, opened 2025-01-23
Allow Java-style comments in CSPs

labkey-adam:
#### Rationale
We want the ability to document our CSP directives, but CSP syntax doesn't allow comments. If Java-style comments are present in a policy, strip them.

### DIFF
--- a/api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
+++ b/api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
@@ -10,7 +10,7 @@ import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
 import org.labkey.api.settings.AppProps;
 import org.labkey.api.settings.OptionalFeatureService;
-import org.labkey.api.util.JavaScanner;
+import org.labkey.api.util.JavaBlockCommentScanner;
 import org.labkey.api.util.PageFlowUtil;
 import org.labkey.api.util.StringExpression;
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
```diff
@@ -148,7 +149,9 @@ public void init(FilterConfig filterConfig) throws ServletException
             String paramValue = filterConfig.getInitParameter(paramName);
             if ("policy".equalsIgnoreCase(paramName))
             {
-                String s = paramValue.trim();
+                // We use Java-style block comments to document our policies; strip them since CSP syntax doesn't allow them.
+                JavaScanner scanner = new JavaScanner(paramValue.trim());
+                String s = scanner.stripComments().toString();
```
labkey-matthewb: Won't this also strip // line comments?  as in https://www.labkey.org

labkey-matthewb: I like that we have this code handy!

labkey-matthewb (changes requested):

api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
labkey-adam: > Won't this also strip // line comments? as in https://www.labkey.org

Yeah, good point. I'll work up something that just handles block comments.

labkey-jeckels (approved):
A few lines of unit testing would be great

labkey-matthewb (approved):

PR #6249, opened 2025-01-24
Fix performance of filter(encodeLinks=true)

labkey-matthewb:
#### Rationale
Filter performance can be very bad when encodeLinks=true (e.g. large report results)

#### Related Pull Requests
- 

#### Changes
-


api/src/org/labkey/api/util/PageFlowUtil.java
```diff
@@ -276,13 +278,12 @@ else if (newline)
                 case 'm':
                     if (encodeLinks)
                     {
-                        CharSequence sub = s.subSequence(i, s.length());
-                        if (StringUtilsLabKey.startsWithURL(sub))
+                        if (StringUtilsLabKey.startsWithURL(s.subSequence(i, Math.min(s.length(),i+10))))
                         {
-                            Matcher m = urlPatternStart.matcher(sub);
-                            if (m.find())
+                            urlMatcher.region(i, s.length());
```
labkey-jeckels, line 283: Any chance this gets faster if you don't give it the full length of the potentially very long string? Something like:

urlMatcher.region(i, Math.min(s.length(),i+10000));

Probably not, but I saw your chat note about this being slow when there are lots of URLs

labkey-jeckels (approved):

api/src/org/labkey/api/util/PageFlowUtil.java
labkey-matthewb: Because we're using lookingAt(), I don't think the end matters at all.

labkey-matthewb: hmmm, though looking at the pattern, I wonder if we need that .* at the end

labkey-matthewb: And the pattern is so simple, we really don't need it

labkey-matthewb: remove .* which is not needed.

PR #6250, opened 2025-01-24
Query dataset error handling updates

labkey-martyp:
#### Rationale
Need to make creating query based snapshot a little easier. First check for required columns on creation and return error if not all there. Second, throw a type of exception if there are issues that doesn't freeze up the schema browser.

#### Changes
- Check required columns on creation
- Use defined dataset col names
- Use MetadataParseExceptions

### DIFF
--- a/study/src/org/labkey/study/controllers/StudyController.java
+++ b/study/src/org/labkey/study/controllers/StudyController.java
@@ -64,7 +64,7 @@ import org.labkey.api.attachments.AttachmentService;
 import org.labkey.api.attachments.BaseDownloadAction;
 import org.labkey.api.audit.TransactionAuditProvider;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
-import org.labkey.api.collections.CaseInsensitiveHashSet;
+import org.labkey.api.collections.LabKeyCollectors;
 import org.labkey.api.compliance.ComplianceService;
 import org.labkey.api.data.ActionButton;
 import org.labkey.api.data.ButtonBar;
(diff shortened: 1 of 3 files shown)


study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -4859,8 +4862,25 @@ public void validateCommand(StudySnapshotForm form, Errors errors)
...
+            if (form.getQueryDataset() != null)
+            {
+                if (study.getTimepointType() != TimepointType.CONTINUOUS)
+                {
+                    errors.reject("snapshotQuery.error", "Query based snapshot is only available for continuous studies");
+                }
+                else
+                {
+                    TableInfo ti = QueryService.get().getUserSchema(getUser(), getContainer(), form.getSchemaName()).getTable(form.getQueryName());
+                    Set<String> colNames = ti.getColumns().stream().map(column -> column.getName().toLowerCase()).collect(Collectors.toSet());
```
labkey-matthewb: use CaseInsensitiveHashSet or use FieldKeys which always use case insensitive compare

labkey-martyp: Updated

labkey-matthewb (approved):

study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -4859,8 +4862,29 @@ public void validateCommand(StudySnapshotForm form, Errors errors)
...
+                {
+                    TableInfo ti = QueryService.get().getUserSchema(getUser(), getContainer(), form.getSchemaName()).getTable(form.getQueryName());
+                    var colNames = new CaseInsensitiveHashSet();
+                    for (ColumnInfo column : ti.getColumns())
+                    {
+                        colNames.add(column.getName());
+                    }
+
+                    List<String> notFound = Arrays.stream(QueryDatasetTable.REQUIRED_COLUMNS)
+                            .filter(value -> !colNames.contains(value.toLowerCase()))
```
labkey-adam: `colNames` is case-insensitive... no need for `toLowerCase()` here

study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -4859,8 +4862,29 @@ public void validateCommand(StudySnapshotForm form, Errors errors)
...
+                    errors.reject("snapshotQuery.error", "Query based snapshot is only available for continuous studies");
+                }
+                else
+                {
+                    TableInfo ti = QueryService.get().getUserSchema(getUser(), getContainer(), form.getSchemaName()).getTable(form.getQueryName());
+                    var colNames = new CaseInsensitiveHashSet();
+                    for (ColumnInfo column : ti.getColumns())
+                    {
+                        colNames.add(column.getName());
+                    }
```
labkey-adam: Of course this works, no problem. But if you want to use a stream (as you did before), we have `LabKeyCollectors.toCaseInsensitiveHashSet()`.

labkey-adam (changes requested):
Remove unnecessary `toLowerCase()` call

study/src/org/labkey/study/controllers/StudyController.java
labkey-martyp: got it, updated

study/src/org/labkey/study/controllers/StudyController.java
labkey-martyp: removed

labkey-adam (approved):

PR #6258, opened 2025-01-28
Issue 52119: assay batch and run properties need to use column name instead of fieldKey for importRun.api

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52119

Server-side code expects column names to be sent when calling assay importRuns.api for the batchProperties and properties (i.e. run properties). We were passing column encoded fieldKeys instead of the names. This PR updates ImportRunApiAction bindParameters() to account for the single quotes around the properties key and to account for the url encoding of double quotes.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/658

#### Changes
- ImportRunApiAction bindParameters() to account for the single quotes around the properties key and to account for the url encoding of double quotes

### DIFF
--- a/assay/src/org/labkey/assay/actions/ImportRunApiAction.java
+++ b/assay/src/org/labkey/assay/actions/ImportRunApiAction.java
@@ -612,21 +612,30 @@ public class ImportRunApiAction extends MutatingApiAction<ImportRunApiAction.Imp
                 {
                     if (name.startsWith("properties["))
                     {
-                        String key = name.substring("properties[".length(), name.length() - 1);
-                        if (key.startsWith("'") && key.endsWith("'"))
-                            key = key.substring(1, key.length()-1);
+                        String key = parsePropertiesKey(name.substring("properties[".length(), name.length() - 1));
                         getProperties().put(key, pv.getValue());
                     }
                     else if (name.startsWith("batchProperties["))
                     {
-                        String key = name.substring("batchProperties[".length(), name.length()-1);
-                        if (key.startsWith("'") && key.endsWith("'"))
-                            key = key.substring(1, key.length()-1);
+                        String key = parsePropertiesKey(name.substring("batchProperties[".length(), name.length()-1));
                         getBatchProperties().put(key, pv.getValue());
                     }
                 }
             }
             return springBindParameters(this, "form", m);
+        }
+
+        private String parsePropertiesKey(String key)
+        {
+            if (key == null || key.isEmpty())
+                return key;
+
+            // Issue 52119: account for leading/trailing single quotes and decode double quotes
+            if (key.startsWith("'") && key.endsWith("'"))
+                key = key.substring(1, key.length()-1);
+            key = key.replaceAll("%22", "\"");
+
+            return key;
         }
     }
 }
--- a/assay/src/org/labkey/assay/actions/ImportRunApiAction.java
+++ b/assay/src/org/labkey/assay/actions/ImportRunApiAction.java
@@ -611,9 +611,19 @@ public class ImportRunApiAction extends MutatingApiAction<ImportRunApiAction.Imp
                 if (name.endsWith("]"))
                 {
                     if (name.startsWith("properties["))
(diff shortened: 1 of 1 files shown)


assay/src/org/labkey/assay/actions/ImportRunApiAction.java
```diff
@@ -611,12 +611,31 @@ public void setAuditUserComment(String auditUserComment)
                 if (name.endsWith("]"))
                 {
                     if (name.startsWith("properties["))
-                        getProperties().put(name.substring("properties[".length(), name.length()-1), pv.getValue());
+                    {
+                        String key = parsePropertiesKey(name.substring("properties[".length(), name.length() - 1));
+                        getProperties().put(key, pv.getValue());
```
labkey-nicka: Given the implementation of `parsePropertiesKey` it is possible for empty string and null to be the value of `key`. Should we avoid putting these into the properties? Certainly seems like we should for null values. Unclear on empty string although seems like it should not go in either.

cnathe: Good point. I made a small change so that parsePropertiesKey will return null in this case and then added a check for `key != null` before putting the key/value in the properties map. I don't think null or empty string will actually get this far, but leaving the check just incase.

assay/src/org/labkey/assay/actions/ImportRunApiAction.java
```diff
@@ -611,12 +611,31 @@ public void setAuditUserComment(String auditUserComment)
...
             return springBindParameters(this, "form", m);
         }
+
+        private String parsePropertiesKey(String key)
+        {
+            if (key == null || key.isEmpty())
+                return key;
+
+            // Issue 52119: account for leading/trailing single quotes and decode double quotes
+            if (key.startsWith("'") && key.endsWith("'"))
```
labkey-susanh, line 636: These never start and end with double quotes?

cnathe: They can, but we add single quotes around these keys on the client side when building up the JSON for the properties when passed to importRun.api so this is accounting for that on the server side.

assay/src/org/labkey/assay/actions/ImportRunApiAction.java
```diff
@@ -611,12 +611,33 @@ public void setAuditUserComment(String auditUserComment)
...
+                        String key = parsePropertiesKey(name.substring("properties[".length(), name.length() - 1));
+                        if (key != null)
+                            getProperties().put(key, pv.getValue());
+                    }
                     else if (name.startsWith("batchProperties["))
-                        getBatchProperties().put(name.substring("batchProperties[".length(), name.length()-1), pv.getValue());
+                    {
+                        String key = parsePropertiesKey(name.substring("batchProperties[".length(), name.length()-1));
+                        if (key != null)
+                        getBatchProperties().put(key, pv.getValue());
```
labkey-nicka: nit: Missing indentation. If the indentation is necessary to respect then `if`, then, not a nit and it is a required change 🙃.

cnathe: huh that's weird, it was indented on my machine when I committed it

assay/src/org/labkey/assay/actions/ImportRunApiAction.java
labkey-susanh: Not knowing much about this, it makes me wonder if there's something better we could do to avoid the string parsing and manipulation, but definitely not something for this PR.

labkey-susanh (approved):

assay/src/org/labkey/assay/actions/ImportRunApiAction.java
cnathe: I totally agree. It has been this way for quite awhile but we should look into if we can just post / include these batch and run properties in a way that binds better on the server side.

labkey-klum (approved):

PR #6259, opened 2025-01-28
Issue 52037: Update TSVWriter to make escaping of backslashes optional

labkey-susanh:
#### Rationale
Issue [52037](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=52037) The initial fix for this issue introduced a problem with a usage of TSVWriter that is writing out file paths on Windows (in `TsvDataExchangeHandler` specifically, but perhaps elsewhere as well). These paths have backslashes in them but the values are not expected to be quoted. 

#### Related Pull Requests
- #6246 

#### Changes
- Add new field and constant in `TSVWriter` to allow quoting of backslashes in particular but possibly other troubling characters in the future.

### DIFF
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -955,7 +955,7 @@ public class ExpDataIterators
                     throw new UnsupportedOperationException();
                 }
             };
-            _tsvWriter.setAdditionalEscapeChars(TSVWriter.BACKSLASH_CHAR_STRING);
+            _tsvWriter.setAdditionalQuotedChars(TSVWriter.BACKSLASH_CHAR_STRING);
 
         }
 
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -955,6 +955,7 @@ protected int write()
                     throw new UnsupportedOperationException();
                 }
             };
+            _tsvWriter.setAdditionalQuotedChars(TSVWriter.BACKSLASH_CHAR_STRING);
```
XingY: Is this added here (DerivationDataIteratorBase) to address known issue, or just to be safe? I tried testing with the change using parent type / data that contains backslash and it seems to be fine. I don't think the parent name part is broken prior to this either though? The reason I ask is because we have several other tsvWriter usages in the app that does`tabLoader.setUnescapeBackslashes(false)`  when parsing, but lacks this line when quoting, `SampleTimelineEvent` for example. It would be good to get those consistent.

labkey-susanh: It was added because of a failure I saw locally. I can test again, though.

labkey-susanh: OK, I tested again and am now not seeing the problems. I'm guessing I was testing before against a different build than I thought I was. Change reverted. Thanks, Xing.

PR #6262, opened 2025-01-29
Issue 51710: Remove replaced assay runs from the search index

RosalineP:
#### Rationale
See issue [here](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51710).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6262
- https://github.com/LabKey/limsModules/pull/1118

#### Changes
- Upon successful run replacement de-index the run
- When a run is deleted that has replaced a former run, then index the former run to ensure it appears in search results
- Ensure replaced runs are not indexed in the future (skip indexing them)

### DIFF
--- a/assay/src/org/labkey/assay/AssayManager.java
+++ b/assay/src/org/labkey/assay/AssayManager.java
@@ -708,7 +708,7 @@ public class AssayManager implements AssayService
      */
     private boolean shouldIndexRun(ExpRun expRun)
     {
-        return getProvider(expRun) != null && expRun.getReplacedByRun() == null;
+        return expRun != null && getProvider(expRun) != null && expRun.getReplacedByRun() == null;
     }
 
     public void indexAssayBatch(int expRunRowId)
--- a/assay/src/org/labkey/assay/AssayManager.java
+++ b/assay/src/org/labkey/assay/AssayManager.java
@@ -704,11 +704,11 @@ public class AssayManager implements AssayService
     }
 
     /**
-     * Only index runs that have an associated assay provider
+     * Only index runs that have an associated assay provider and that have not been replaced
      */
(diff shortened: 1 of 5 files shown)


assay/src/org/labkey/assay/AssayManager.java
```diff
@@ -704,11 +704,11 @@ private boolean shouldIndexProtocolBatches(ExpProtocol protocol)
     }
 
     /**
-     * Only index runs that have an associated assay provider
+     * Only index runs that have an associated assay provider and that have not been replaced
      */
     private boolean shouldIndexRun(ExpRun expRun)
     {
-        return getProvider(expRun) != null;
+        return getProvider(expRun) != null && expRun.getReplacedByRun() == null;
```
labkey-nicka: Consider adding a `expRun != null` to the beginning of this statement.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -4347,7 +4347,13 @@ public void deleteExperimentRuns(Container container, final User user, @Nullable
                 // Archive all data files prior to deleting
                 //  ideally this would be transacted as a commit task but we decided against it due to complications
                 run.archiveDataFiles(user);
-
+                // Re-index replaced run if replacing run is deleted
+                if (run.getReplacesRuns() != null)
```
labkey-nicka, line 4351: I'm skeptical of not doing this in a transaction post-commit. I realize there may be complications due to which runs are deleted, etc. but if the transaction fails this will still index the replaced runs which leaves us in an incongruent state/behavior from what we want.

labkey-nicka (approved):

PR #6263, opened 2025-01-30
Prevent import form from submitting twice

labkey-tchad:
#### Rationale
Rarely on TeamCity, this form submits data twice. An even less common subset of times, that double-submission triggers an "Ajax communication failed" error. Skipping additional submissions while the Ext mask is in place should prevent this problem.

#### Related Pull Requests
- N/A

#### Changes
- Prevent import form from submitting twice

### DIFF
--- a/api/src/org/labkey/api/query/import.jsp
+++ b/api/src/org/labkey/api/query/import.jsp
@@ -184,6 +184,12 @@
         if (!form)
             return;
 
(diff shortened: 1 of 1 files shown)


labkey-nicka (approved):
Looks good. There is probably a way to ask the form if it is already being submitted but this logic does the job. Thanks for looking/fixing.

PR #6265, opened 2025-01-31
Issue 52154: LKSM/LKB: Unexpected "AliquotCount" error on Sample Import 

XingY:
#### Rationale
Aliquot rollup columns are added to the data iterator during sample insert/import/merge so parent samples can have those values initialized to 0 while aliquot to null. When the import file also contains such columns, the import fails with "duplicate columns:" error.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6265
- https://github.com/LabKey/limsModules/pull/1120

#### Changes
- ignore provided aliquot rollup columns

### DIFF
--- a/api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
+++ b/api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
@@ -92,6 +92,13 @@ public class SampleTypeDomainKind extends AbstractDomainKind<SampleTypeDomainKin
     private static final Set<PropertyStorageSpec.ForeignKey> FOREIGN_KEYS;
     private static final Set<String> FORCE_ENABLED_SYSTEM_FIELDS;
 
+    public static final String ALIQUOT_COUNT_LABEL = "Aliquots Created Count";
+    public static final String ALIQUOT_VOLUME_LABEL = "Aliquot Total Amount";
+    public static final String AVAILABLE_ALIQUOT_COUNT_LABEL = "Available Aliquot Count";
+    public static final String AVAILABLE_ALIQUOT_VOLUME_LABEL = "Available Aliquot Amount";
+
+    public static final Set<String> ALIQUOT_ROLLUP_FIELD_LABELS = CaseInsensitiveHashSet.of(ALIQUOT_COUNT_LABEL, ALIQUOT_VOLUME_LABEL, AVAILABLE_ALIQUOT_VOLUME_LABEL, AVAILABLE_ALIQUOT_COUNT_LABEL);
+
     static
     {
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -1591,6 +1597,22 @@ public static boolean isUnitsHeader(String name)
         {
             return isExpMaterialColumn(ExpMaterialTable.Column.Units, name);
         }
+
+        private static boolean isAliquotRollupHeader(String name)
+        {
+            for (String rollupField : ALIQUOT_ROLLUP_FIELDS.keySet())
```
labkey-matthewb: Use CaseInsensitiveSet.of() instead of Set.of() so this code can use Map.containsKey()?

labkey-matthewb (approved):

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -456,26 +466,26 @@ public StringExpression getURL(ColumnInfo parent)
             }
             case AliquotCount ->
             {
-                var ret = wrapColumn(alias, _rootTable.getColumn("AliquotCount"));
-                ret.setLabel("Aliquots Created Count");
+                var ret = wrapColumn(alias, _rootTable.getColumn(AliquotCount.name()));
+                ret.setLabel(ALIQUOT_COUNT_LABEL);
```
labkey-susanh, line 468: I think we also want to make sure these fields and the labels are reserved.

labkey-susanh (changes requested):

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
XingY: Yep, they were already excluded, updated to use constants instead.

labkey-susanh (approved):

api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
```diff
@@ -92,6 +92,13 @@ public class SampleTypeDomainKind extends AbstractDomainKind<SampleTypeDomainKin
     private static final Set<PropertyStorageSpec.ForeignKey> FOREIGN_KEYS;
     private static final Set<String> FORCE_ENABLED_SYSTEM_FIELDS;
 
+    public static final String ALIQUOT_COUNT_LABEL = "Aliquots Created Count";
```
labkey-nicka, line 95: Consider using the same pattern as `InventoryService.InventoryStatusColumn.label()` so that these constants can be colocated with the column enumeration.

PR #6266, opened 2025-01-31
Issue 52145: Trim spaces around aliquot parent names

labkey-susanh:
#### Rationale
Issue 52145: When importing data from Excel, spaces around values in cells are not automatically trimmed. We've accounted for this when looking up entities for lineage, but not when looking up parents for aliquots.  A broader change to the ExcelLoader to trim out spaces will wait for a different branch.

#### Related Pull Requests
- 

#### Changes
- Trim the names of aliquot parents when finding the parent samples and constructing the names of aliquots.


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -1272,7 +1272,7 @@ else if (o instanceof Number)
                         }
 
                         if (aliquotParentName != null)
-                            _aliquotParents.put(lsid, aliquotParentName);
+                            _aliquotParents.put(lsid, aliquotParentName.trim());
```
XingY: I suggest to refactor and reuse `AliquotRollupDataIterator.getAliquotParent` for getting aliquoted from value here. And if it turns out `AliquotRollupDataIterator.getAliquotParent` is inadequate, then we need to fix there as well.

labkey-susanh: Done.

XingY (approved):

PR #6269, opened 2025-02-02
Fix creation of short URLs on SQL Server

labkey-adam:
#### Rationale
Casing mismatch on SQL Server causes an IllegalStateException

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52162


XingY (approved):

labkey-jeckels:
This inspires me to switch from String to GUID for these variables.

PR #6271, opened 2025-02-03
Update ExcelLoader to trim spaces surrounding cell values

labkey-susanh:
#### Rationale
Extra spaces around data being imported can cause problems (e.g., lookups won't be found if we try to match using the extra spaces and the lookup value doesn't have a space. See [Issue 51245](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52145), for example). We trim these spaces when loading data from TSVs and CSVs, but don't currently do that when loading data from Excel. This PR changes that so we trim off leading and trailing spaces from cell values when loading Excel files

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1147

#### Changes
- Update `ExcelLoader` to trim strings


api/src/org/labkey/api/reader/ExcelLoader.java
```diff
@@ -723,11 +726,13 @@ protected Object[] readFields() throws IOException
...
                             else if (useColumnFormats && column.clazz.equals(String.class))
                             {
-                                contents = ExcelFactory.getCellStringValue(cell);
+                                contents = ExcelFactory.getCellStringValue(cell).trim();
                             }
                             else
                             {
                                 contents = PropertyType.getFromExcelCell(cell);
+                                if (contents instanceof String)
+                                    contents = ((String) contents).trim();
```
labkey-jeckels:
```suggestion
                                if (contents instanceof String s)
                                    contents = s.trim();
```

api/src/org/labkey/api/reader/ExcelLoader.java
```diff
@@ -385,7 +385,7 @@ private String[][] getFirstNLinesXLSX(int n) throws IOException, InvalidFormatEx
                 String data = (v != null && !(v instanceof String)) ? String.valueOf(v) : (String) v;
                 if (!StringUtils.isEmpty(data))
                     foundData = true;
-                rowData.add(data != null ? data : "");
+                rowData.add(data != null ? data.trim() : "");
```
labkey-matthewb: Should we share the implementation of TabLoader.parseValue(String)?  Or if not maybe there should be ExcelLoader.parseValue(String) to consolidate any handling that happens after String.valueOf()?

labkey-matthewb: The only interesting part of parseValue() is _unescapeBackslashes which doesn't apply.   But consolidationg null, "", trim() behavior seems worthwhile.

labkey-matthewb (approved):

api/src/org/labkey/api/reader/ExcelLoader.java
labkey-susanh: Good idea to centralize. I also just want to throw out there that I recently made [changes to DataLoader](https://github.com/LabKey/platform/pull/6217)'s parsing of headers that cause it to also strip out nonbreaking spaces, which `trim()` does not handle. Should we do the same thing here?

labkey-matthewb: One thought, trimming starts with form entry because it's way too easy to have dangling space.  Then we add for other types of text input.  And now distinguishing form input vs API is not clear... 

I tend to think we should trim non-breaking spaces, but maybe we should just write it down what we think is best practice for: whitespace and non-printable character handling vs form/import/api).

labkey-susanh: I'm not sure I follow this comment about starting with form entry. Are you saying we also want to consider (perhaps in a separate PR) adding more trimming logic somewhere to deal with form inputs? Is there a centralized place for that?

labkey-matthewb: Sorry, I'm just musing about the historical reason for altering the provided character data (why we trim at all).  Which clearly starts with it being too easy for humans to enter unintended space in forms.  We should probably write down when/how we alter incoming data before we save it.  And it makes me curious what happens with with non-printing characters, but that is another discussion.

labkey-jeckels:
We want to trim in the `iterator()` use case too. I suggest trimming in `SheetLoader.endElement()` right before we set the value in `_currentRow`, around line 1322 in `ExcelLoader`. You could then remove the trim in `getFirstNLinesXLSX()` since it uses the results of the iterator.

labkey-susanh:
> We want to trim in the `iterator()` use case too. I suggest trimming in `SheetLoader.endElement()` right before we set the value in `_currentRow`, around line 1322 in `ExcelLoader`. You could then remove the trim in `getFirstNLinesXLSX()` since it uses the results of the iterator.

Good thought. My testing suggested this updated is indeed needed, but the trim in `getFirstNLines` is also needed (for inferring columns, for example).

labkey-susanh:
> > We want to trim in the `iterator()` use case too. I suggest trimming in `SheetLoader.endElement()` right before we set the value in `_currentRow`, around line 1322 in `ExcelLoader`. You could then remove the trim in `getFirstNLinesXLSX()` since it uses the results of the iterator.
> 
> Good thought. My testing suggested this updated is indeed needed, but the trim in `getFirstNLines` is also needed (for inferring columns, for example).

My bad. Missing an all-important `else` in there.  Still going to leave the second trim in though, I think as it looks a little cleaner.

PR #6274, opened 2025-02-03
Issue 51973: LIMS: Only one person's updates to templates are saved if two admins are working concurrently

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1708
- https://github.com/LabKey/labkey-ui-premium/pull/666
- https://github.com/LabKey/platform/pull/6274
- https://github.com/LabKey/limsModules/pull/1132

#### Changes
- add `modified` to `getQueryDetails.api` response
- check concurrent edit of import template by comparing `modified`
- add audit log for import template updates


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -8357,6 +8370,13 @@ public Object execute(QueryImportTemplateForm form, BindException errors) throws
             Container container = getContainer();
             String schemaName = form.getSchemaName();
             String queryName = form.getQueryName();
+            QueryDef queryDef = QueryManager.get().getQueryDef(container, schemaName, queryName, false);
+            if (queryDef != null && queryDef.getQueryDefId() != 0)
+            {
+                Long lastKnownModified = form.getLastKnownModified();
+                if (lastKnownModified == null || lastKnownModified != queryDef.getModified().getTime())
+                    throw new ApiUsageException("Unable to save import templates. The templates appears out of date, reload the page and try again.");
```
labkey-alan: This error message doesn't seem correct to me. At a minimum we should make the plural `appears` be `appear`, however, I think it should maybe be something like:

`Unable to save import templates. The Query Definition appears to be out of date, reload the page and try again.`

But I'm also not sure that saying "The Query Definition" is the correct term either. The reason I don't think we should be specific about saying the "templates appear out of date" is that you can trigger this error without updating the templates by:

1. Create a sample type
2. Add a template to the sample type
3. In a separate tab rename the sample type to something different, then change the name back
4. In the first tab try to change the name of the template that you previously uploaded.

Result: an error because I have an out of date query def, but it's telling me that the templates are out of date which is not true.

Side note: What other ways can we trigger the modified timestamp to change? I tried changing the description of a sample type and adding fields, but that didn't change the modified timestamp which I would have expected it to. This seems like a bug, but maybe it's intended behavior?

labkey-alan (changes requested):
Overall this looks good, and I was able to verify the functionality locally, however I think we should improve the error messaging, and I think there is potentially a bug (not 100% sure, and it definitely would predate this PR).

query/src/org/labkey/query/controllers/QueryController.java
labkey-susanh: I would argue that it's accurate to say the templates are out of date since that's literally what we are seeing (modified time is different), even if it doesn't appear that they are different.  If we update the messaging as Alan suggests, let's make sure we aren't showing the user anything about a Query Definition being out of date. They won't know what that means.

XingY: I left the error msg as is since this API is for import template and it would confuse user as to what Query Definition is, as Susan said. 

Regarding what actions updates QueryDef, there not much... Updating schemaName/queryName triggers it, as Alan pointed out. Otherwise, you would need to modify the table query source or XML. Changing sample type description or adding a regular field won't update queryDef, but updating a calculated column would.

labkey-alan: Ok, that all seems reasonable to me then. We should still make the message say:

```
The templates appear out of date, reload the page and try again.
```

Instead of:
```
The templates appears out of date, reload the page and try again.
```

labkey-alan (approved):
Looks good, but we should fix the minor issue with the error message.

PR #6275, opened 2025-02-04
Issue 51620: Remove the UI for Object-level discussions

RosalineP:
#### Rationale
See issue [here](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51620).

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/667
- https://github.com/LabKey/labkey-ui-components/pull/1709
- https://github.com/LabKey/limsModules/pull/1135
- https://github.com/LabKey/platform/pull/6275
- https://github.com/LabKey/testAutomation/pull/2260

#### Changes
- Check DEPRECATED_OBJECT_LEVEL_DISCUSSIONS before showing relevant UI
- Add deprecation flag for DEPRECATED_OBJECT_LEVEL_DISCUSSIONS


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -9478,6 +9478,33 @@ public void addNavTrail(NavTree root)
...
+        }
+    }
+
+    @RequiresPermission(ReadPermission.class)
+    public static class DeprecatedFlagAction extends ReadOnlyApiAction<DeprecatedFlagForm>
+    {
+        @Override
+        public ApiResponse execute(DeprecatedFlagForm form, BindException errors)
+        {
+            ApiSimpleResponse ret = new ApiSimpleResponse();
```
RosalineP: To support client-side update.

RosalineP:
Pending selenium test updates

core/src/org/labkey/core/admin/AdminController.java
cnathe: I don't think we need to add this new action. The approach we have taken for these experimental feature flags is to have the relevant module add it to its page context JSON object so that the client side code can just get the boolean value from there instead of needing to make an API call. 

See example in the biologics module for `EXPERIMENTAL_REQUESTS_MENU`. I'd suggest adding this `DEPRECATED_OBJECT_LEVEL_DISCUSSIONS` to the CoreModule `getPageContextJson` and then in ui-components we can add a helper in `app/utils.ts` that is similar to the `isSampleAliquotSelectorEnabled` or `isAssayRequestsEnabled`.

cnathe (changes requested):

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -9478,6 +9478,33 @@ public void addNavTrail(NavTree root)
...
+        }
+
+        public void setFeatureName(String featureName)
+        {
+            this.featureName = featureName;
+        }
+    }
+
+    @RequiresPermission(ReadPermission.class)
+    public static class DeprecatedFlagAction extends ReadOnlyApiAction<DeprecatedFlagForm>
```
cnathe: This action and related form can be removed right? it is no longer used from what I see.

cnathe (changes requested):

core/src/org/labkey/core/admin/AdminController.java
RosalineP: 🤦‍♀️  My bad! Thanks for the nudge Cory

cnathe (approved):

PR #6276, opened 2025-02-04
Issue 51887: Log scale extends much lower than needed for some Panorama QC plots

ankurjuneja:
#### Rationale
This PR fixes the y axis log scale values for the qc plots when the converted y axis domain falls below zero for closer to zero values.

#### Related Pull Requests
- 

#### Changes
-


labkey-jeckels (approved):
Tested in the FB and it looked good.

PR #6291, opened 2025-02-05
Add exp.FieldNames and exp.PhiFields tables

labkey-adam:
#### Rationale
Table to help locate and analyze fields with special characters: https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=52114

Also add a table that highlights fields with PHI Level set

### DIFF
--- a/experiment/src/org/labkey/experiment/api/BaseFieldNamesTable.java
+++ b/experiment/src/org/labkey/experiment/api/BaseFieldNamesTable.java
@@ -1,0 +1,72 @@
+package org.labkey.experiment.api;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.labkey.api.data.BaseColumnInfo;
+import org.labkey.api.data.ContainerFilter;
+import org.labkey.api.data.JdbcType;
+import org.labkey.api.data.MutableColumnInfo;
+import org.labkey.api.data.SQLFragment;
+import org.labkey.api.data.SimpleFilter;
+import org.labkey.api.exp.OntologyManager;
+import org.labkey.api.exp.query.ExpSchema;
+import org.labkey.api.query.FieldKey;
+import org.labkey.api.query.FilteredTable;
+import org.labkey.api.security.permissions.AdminPermission;
+
+public abstract class BaseFieldNamesTable extends FilteredTable<ExpSchema>
+{
+    private final String _tableName;
+
+    public BaseFieldNamesTable(String tableName, @NotNull ExpSchema userSchema, @Nullable ContainerFilter containerFilter)
+    {
+        super(OntologyManager.getTinfoPropertyDescriptor(), userSchema, containerFilter);
+        _tableName = tableName;
+
+        setName(tableName);
+        addWrapColumn(_rootTable.getColumn("Container"));
+        addColumn("DomainUri", JdbcType.VARCHAR);
+        addColumn("DomainName", JdbcType.VARCHAR);
+        addWrapColumn(_rootTable.getColumn("Name"));
+        addWrapColumn(_rootTable.getColumn("Label"));
+        addWrapColumn(_rootTable.getColumn("Description"));
+    }
+
+    protected MutableColumnInfo addColumn(String name, JdbcType type)
+    {
+        return addColumn(new BaseColumnInfo(FieldKey.fromParts(name), this, type));
+    }
+
+    @Override
+    protected SimpleFilter.FilterClause getContainerFilterClause(ContainerFilter filter, FieldKey fieldKey)
+    {
+        // Show only folders where current user has admin permissions
+        return filter.createFilterClause(getSchema(), fieldKey, AdminPermission.class, null);
+    }
+
+    @Override
+    @NotNull
+    public SQLFragment getFromSQL(String alias)
+    {
+        SQLFragment result = new SQLFragment();
+        result.appendComment("<" + _tableName + ">", getSqlDialect());
+        result.append("(SELECT dd.Name AS DomainName, dd.DomainUri, pd.*");
+
+        addColumnSQL(result);
+
+        result.append(" FROM ")
+            .append(super.getFromSQL("pd"))
+            .append("""
+
+                JOIN exp.PropertyDomain AS pdom ON pdom.PropertyId = pd.PropertyId
+                JOIN exp.DomainDescriptor AS dd ON dd.DomainId = pdom.DomainId""");
+
+        result.append(")\n")
+            .append(alias)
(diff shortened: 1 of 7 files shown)


experiment/src/org/labkey/experiment/api/FieldNamesTable.java
```diff
@@ -0,0 +1,45 @@
...
+        super("FieldNames", userSchema, containerFilter);
+
+        addColumn("AlphaNumeric", JdbcType.BOOLEAN).setDescription("Name contains only alphabetic and numeric characters");
+        addColumn("SpecialCharacters", JdbcType.BOOLEAN).setDescription("Name contains any of these specific punctuation characters: / \"&\\$}~,.");
+    }
+
+    @Override
+    protected void addColumnSQL(SQLFragment sql)
+    {
+        addBooleanPatternColumn(sql, "pd.Name NOT", "%[^A-Za-z0-9]%", "AlphaNumeric");
```
labkey-jeckels: Maybe lump in underscores too?

experiment/src/org/labkey/experiment/api/FieldNamesTable.java
```diff
@@ -0,0 +1,45 @@
...
+import org.labkey.api.data.ContainerFilter;
+import org.labkey.api.data.JdbcType;
+import org.labkey.api.data.SQLFragment;
+import org.labkey.api.exp.query.ExpSchema;
+
+public class FieldNamesTable extends BaseFieldNamesTable
+{
+    public FieldNamesTable(@NotNull ExpSchema userSchema, @Nullable ContainerFilter containerFilter)
+    {
+        super("FieldNames", userSchema, containerFilter);
```
labkey-jeckels, line 14: Add a description that talks about non-admins not seeing any rows.

labkey-jeckels (approved):

labkey-matthewb (approved):
This is cool.  I can't help but want to see datatype (perhaps a future thing).  Is treating space as special useful? To my eye it creates a bunch of false positives (at least hits that I don't find interesting)  Quick test of AdminPermission seems to work as expected.

experiment/src/org/labkey/experiment/api/FieldNamesTable.java
labkey-matthewb: Is regular space special?

labkey-adam: Now treating underscores as AlphaNumeric

Keith asked for space to considered special. We do warn about space in the domain designer. Feel free to comment in the issue or follow up with Keith. @keith-labkey

experiment/src/org/labkey/experiment/api/FieldNamesTable.java
labkey-adam: Done

labkey-adam:
> This is cool. I can't help but want to see datatype (perhaps a future thing). Is treating space as special useful? To my eye it creates a bunch of false positives (at least hits that I don't find interesting) Quick test of AdminPermission seems to work as expected.

I added RangeURI. I didn't add a display column to turn this into an actual type. See previous comment about space.

experiment/src/org/labkey/experiment/api/FieldNamesTable.java
labkey-susanh: Also, as being discussed in a different thread and related to [Issue 52103](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52103), having multiple spaces in names is somewhat troublesome (for humans at least).

labkey-keith: I think we leave space in as special to be consistent with the warnings we give in the browser.

labkey-matthewb: Ok.  good point.  That was me, but I've given up that fight a long time ago.  Maybe we should stop warning on space in the designer.

experiment/src/org/labkey/experiment/api/FieldNamesTable.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    private void addBooleanPatternColumn(SQLFragment sql, String expression, String pattern, String name)
+    {
+        // Need the operator that supports character classes
+        String operator = getSqlDialect().getCharClassLikeOperator();
+        sql.append(", ");
+
+        // Can't expose a boolean expression as a boolean column on SQL Server, so wrap with a CASE statement
+        if (getSqlDialect().isSqlServer())
```
labkey-nicka: Looks like `SqlDialect.wrapExistsExpression()` does this.

labkey-jeckels (approved):
This looks fine. FWIW, I probably would have exposed the PHI field in `exp.FieldNames` instead of making this a separate table, but maybe there are good reasons to have it split out

labkey-adam:
> This looks fine. FWIW, I probably would have exposed the PHI field in `exp.FieldNames` instead of making this a separate table, but maybe there are good reasons to have it split out

Reasons: hide the table when compliance isn't present, apply the "not NotPHI" base filter, different set of columns of interest. Arguably, views on a single table could have handled these differences, but it's nice to navigate to or link to the query.

experiment/src/org/labkey/experiment/api/FieldNamesTable.java
labkey-adam: Thanks, I switched the impl to use this. I also updated the comments and plan to rename the method on develop. This doesn't scream "I wrap boolean expressions" to me. I didn't find it because I was searching for "boolean"...

experiment/src/org/labkey/experiment/api/BaseFieldNamesTable.java
```diff
@@ -38,10 +40,11 @@ protected MutableColumnInfo addColumn(String name, JdbcType type)
         return addColumn(new BaseColumnInfo(FieldKey.fromParts(name), this, type));
     }
 
+    // This is a hack, required because ContainerFilter.Type.Current doesn't fill in the user
     @Override
-    protected ContainerFilter getDefaultContainerFilter()
+    public boolean hasPermission(@NotNull UserPrincipal user, @NotNull Class<? extends Permission> perm)
```
labkey-matthewb, line 45: I find it hard to consider this a hack, unless we're going to start using WHERE 0=1 as the general way to enforce permissions on tables.

labkey-adam: I want the same behavior as many of our other tables: any user can view the table; whether they see rows or not is dictated solely by folder filter permissions. The only reason I'm forbidding non-admins from viewing this table is because `Type.Current` is inconsistent with the other ContainerFilter types.

PR #6300, opened 2025-02-08
Avoid audit table deadlock during bootstrapping

labkey-jeckels:
#### Rationale
TeamCity hits intermittent deadlocks during bootstrapping related to queuing audit events and provisioning the audit tables.

Example failure:

https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_BvtBSqlserver/3372146?buildTab=artifacts#%2FembeddedTomcatLogs.tar.gz;%2FembeddedTomcatLogs.tar.gz

#### Changes
- Shrink the critical section for queued audit events


labkey-susanh (approved):
This seems reasonable enough, but I do wonder why it's started failing in this way only now.

labkey-jeckels:
> This seems reasonable enough, but I do wonder why it's started failing in this way only now.

Me too. I couldn't find anything, but since this is so intermittent and distributed across suites it's hard to really know when it started or got worse.

PR #6313, opened 2025-02-12
Limit special substitutions to HTML wikis only

labkey-adam:
#### Rationale
We want to support webpart and dependency substitutions in HTML wikis, but not announcements or Markdown wikis. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52214

Pass a flag through `WikiRenderingService` methods to a new configuration parameter on `HtmlRenderer()`.

### DIFF
--- a/api/src/org/labkey/api/wiki/WikiRenderingService.java
+++ b/api/src/org/labkey/api/wiki/WikiRenderingService.java
@@ -41,12 +41,12 @@ public interface WikiRenderingService
      * @param sourceDescription info on where the text came from for debugging purposes. For example: Announcement 6654 in /MyContainer
      */
     HtmlString getFormattedHtml(WikiRendererType rendererType, String source, @Nullable String sourceDescription,
-                                String attachPrefix, Collection<? extends Attachment> attachments);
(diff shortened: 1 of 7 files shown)


api/src/org/labkey/api/wiki/WikiRenderingService.java
```diff
@@ -41,12 +41,12 @@ static void setInstance(WikiRenderingService impl)
      * @param sourceDescription info on where the text came from for debugging purposes. For example: Announcement 6654 in /MyContainer
      */
     HtmlString getFormattedHtml(WikiRendererType rendererType, String source, @Nullable String sourceDescription,
-                                String attachPrefix, Collection<? extends Attachment> attachments);
+                                boolean handleSubstitutions, String attachPrefix, Collection<? extends Attachment> attachments);
```
labkey-jeckels, line 45: Consider a note about what "substitutions" means, similar to the comment on line 243 of `HtmlRenderer`

labkey-jeckels (approved):

api/src/org/labkey/api/wiki/WikiRenderingService.java
labkey-adam: Done

PR #6314, opened 2025-02-12
Restore and improve ext dependency loading shortcut

labkey-tchad:
#### Rationale
Removing the dependency loading shortcut for ext revealed a test page that was relying on the old behavior. Rather than waiting to see if this has any adverse impact in non-test areas, I'm restoring the basic functionality while ensuring that patches have bene loaded.

#### Related Pull Requests
- #6297 

#### Changes
- Restore and improve ext dependency loading shortcut


core/webapp/labkey.js
```diff
@@ -474,7 +474,14 @@ if (typeof LABKEY == "undefined")
 
             checkCallback('requiresExt3', callback);
 
-            requiresLib('Ext3', callback, scope);
+            if (window.Ext?.labkeyPatches)
```
labkey-nicka: Lets make these `window.Ext && window.Ext.labkeyPatches` as I'm not sure of support in some environments (e.g. server-side JS) for the conditional operator.

labkey-nicka (approved):

PR #6315, opened 2025-02-13
Issue 52254: Folder admins cannot choose a template folder

labkey-jeckels:
#### Rationale
Folder admins don't have access to the root container, so `getContainers` sends neither an `effectivePermissions` array nor a `path`.

The old behavior was to call `LABKEY.Security.hasPermission()` which did a bitwise AND, so therefore tolerated an undefined value

https://github.com/LabKey/platform/pull/5718

#### Changes
- Check if we have permissions to review

### DIFF
--- a/core/src/org/labkey/core/admin/createFolder.jsp
+++ b/core/src/org/labkey/core/admin/createFolder.jsp
@@ -622,7 +622,7 @@
         const getTemplateFolders = function(data) {
             // add the container itself to the templateFolder object if it is not the root and the user has admin perm to it
             // and if it is not a workbook or container tab folder
-            if (data.path !== "/" && LABKEY.Security.hasEffectivePermission(data.effectivePermissions, LABKEY.Security.effectivePermissions.admin)
+            if (data.path !== "/" && data.effectivePermissions && LABKEY.Security.hasEffectivePermission(data.effectivePermissions, LABKEY.Security.effectivePermissions.admin)
                     && !data.isWorkbook && !data.isContainerTab)
             {
                 templateFolders.push([data.id, data.path]);


core/src/org/labkey/core/admin/createFolder.jsp
```diff
@@ -622,7 +622,7 @@
         const getTemplateFolders = function(data) {
             // add the container itself to the templateFolder object if it is not the root and the user has admin perm to it
             // and if it is not a workbook or container tab folder
-            if (data.path !== "/" && LABKEY.Security.hasEffectivePermission(data.effectivePermissions, LABKEY.Security.effectivePermissions.admin)
+            if (data.path !== "/" && data.effectivePermissions && LABKEY.Security.hasEffectivePermission(data.effectivePermissions, LABKEY.Security.effectivePermissions.admin)
```
labkey-nicka, line 625: We could/should consider also making `LABKEY.Security.hasEffectivePermission()` defensive of falsey inputs.

labkey-jeckels: I briefly considered that and seems like it would be a fine approach too. Would that be here?

https://github.com/LabKey/platform/blob/98df4829620cdf56a23c1e629a7f91503397b4cf/core/resources/scripts/labkey/Security.js#L204

I'm confused about how the TS version is or isn't involved.

labkey-adam (approved):

core/src/org/labkey/core/admin/createFolder.jsp
labkey-nicka: I think its OK to not do this in 24.11 and just put the change in as you have it.

Additionally, I've tried to [describe how the client API is served](https://gist.github.com/labkey-nicka/3cc3ea4202ea358ae0a3a2d227c41f71). Let me know if this is helpful.

labkey-jeckels: > Additionally, I've tried to [describe how the client API is served](https://gist.github.com/labkey-nicka/3cc3ea4202ea358ae0a3a2d227c41f71). Let me know if this is helpful.

Yes, definitely. I hope to be able to find it in the future.

PR #6317, opened 2025-02-13
Try draining the indexer's queue before deleting to avoid deadlocks

labkey-jeckels:
#### Rationale
We get intermittent deadlocks in tests like `ExpSampleTypeTestCase.jsp` where the post-test container delete conflicts with the indexing work initiated by the test's import.

https://teamcity.labkey.org/buildConfiguration/LabKey_2411_External_Ehr_BvtEhrSqlserver/3377344

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52257

#### Changes
- Try letting the indexer finish the queued work before deleting the container

### DIFF
--- a/search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
+++ b/search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
@@ -143,6 +143,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Predicate;
 import java.util.regex.Matcher;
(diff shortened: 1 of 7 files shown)


search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
```diff
@@ -1104,6 +1105,25 @@ private void parse(WebdavResource r, FileStream fs, InputStream is, ContentHandl
         }
     }
 
+    @Override
+    public boolean drainQueue(PRIORITY priority, long timeout, TimeUnit unit) throws InterruptedException
```
labkey-matthewb: I think this could be moved to AbstractSearchService (or even SearchService default).  Not important, but this doesn't look Impl specific.

labkey-matthewb (approved):

search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
labkey-jeckels: I moved it. I'm not quick clear on the value of AbstractSearchService but you're correct. Nothing in the method relied on the Impl

XingY (approved):

PR #6318, opened 2025-02-13
Issue 52247: DataRegion.renderForm updates to move _groupTables table generation to after form table generation

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=52247

When the DataRegion.renderForm() is dealing with a "groupTable" case, like in the Luminex 3rd step of the upload wizard that renders the table version of the form inputs for the analyte properties, the generated HTML can include a `<table>` within a `<table>` in a way that is not properly closing the other table related HTML elements. This PR moves that groupTable HTML generation to after the single form table generation is properly closing the related elements.

#### Changes
- move _groupTables table generation to after form table generation
- encode the name prop of the hidden input for the PK column

### DIFF
--- a/api/src/org/labkey/api/data/DataRegion.java
+++ b/api/src/org/labkey/api/data/DataRegion.java
@@ -2187,6 +2187,53 @@ public class DataRegion extends DisplayElement
                 renderedColumns.add(renderer.getColumnInfo().getName());
         }
 
+        //Make sure all pks are included
+        if (action == MODE_UPDATE)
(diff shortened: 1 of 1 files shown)


labkey-nicka (approved):

labkey-adam (approved):

cnathe:
FYI, I'll merge this to develop here and then backport to 24.11 and add an automated test there.

PR #6319, opened 2025-02-13
Issue 52060: Do trimming of domain names earlier in the process so validation catches reserved words

labkey-susanh:
#### Rationale
Issue 52060 - [LKSM: Spaces at the end of source type names are problematic](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52060)

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1165
- https://github.com/LabKey/labkey-ui-components/pull/1719

#### Changes
- Trim domain names before validation happens
- Update error messaging about reserved domain names to be consistent in the different code paths.


api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -704,11 +704,11 @@ public static Domain createDomain(
 
 
         if (domainName != null)
-            domain.setName(domainName);
+            domain.setName(StringUtils.trimToNull(domainName));
 
         if (domain.getName() != null)
         {
-            String domainNameError = validateDomainName(domain.getName(), kind.getKindName(), kind.supportsNamingPattern());
+            String domainNameError = validateDomainName(StringUtils.trimToEmpty(domain.getName()), kind.getKindName(), kind.supportsNamingPattern());
```
XingY: I always find this `createDomain` interface odd, in that the `domain` can already have `name`, but then an explicit but optional `domainName` is passed in. If the domain itself's name contains leading/trailing whitespace, but `domainName` is null, then the space is saved. Perhaps we should be another `domain.setName(StringUtils.trimToNull(domain.getName))` here

api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -767,7 +767,7 @@ public static ValidationException updateDomainDescriptor(GWTDomain<? extends GWT
             return validationException;
         }
 
-        String updatedName = update.getName();
+        String updatedName = StringUtils.trimToEmpty(update.getName());
```
XingY: This change could be a little problematic. For legacy domains with names that contains illegal characters or patterns, we do still allow them to be updated as long as there is no domain name change. Now if the legacy domain name happened to have leading/trialing white space and also illegal pattern, then this domain can no longer be updated without having to be renamed.

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -1489,10 +1489,10 @@ private static ValidationException updateDomain(GWTDomain<? extends GWTPropertyD
 
         if (!kind.canEditDefinition(user, domain))
             throw new UnauthorizedException("You don't have permission to edit this domain.");
-
-        if (original.getName() != null && update.getName() != null && !original.getName().equals(update.getName()))
+        String trimmedName = StringUtils.trimToNull(update.getName());
+        if (original.getName() != null && trimmedName != null && !original.getName().equals(trimmedName))
```
XingY: Same comment as above regarding legacy domain name support.

XingY (approved):

api/src/org/labkey/api/exp/property/DomainUtil.java
labkey-susanh: Updated in both cases to use `trimToEmpty`.

api/src/org/labkey/api/exp/property/DomainUtil.java
labkey-susanh: Yeah, I can see that. Since it seems unlikely that the untrimmed name will exist in any case and there is a workaround, I'm leaning toward leaving in this change, though so we'll have less chance of having untrimmed names saved. If anyone sees this as overly problematic, though, I can probably be convinced otherwise. 

@cnathe Thoughts on this?

XingY:
Should the component version be bumped to bring the space warning to lks?

labkey-susanh:
> Should the component version be bumped to bring the space warning to lks?
good thought. I'll do that.

api/src/org/labkey/api/exp/property/DomainUtil.java
labkey-susanh: After further testing, I've moved this trim inside the if (in both cases).

PR #6306, opened 2025-02-11
Combine ContainerFilter.Types Current and CurrentWithUser

labkey-adam:
#### Rationale
Provide user context in more cases to support folder filter changes

https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=52229

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6291

PR #6322, opened 2025-02-13
Restore previous input role retrieval behavior

labkey-adam:
#### Rationale
After my `ContainerFilter.Type.Current` change (related PR), the sample derivation page failed to locate input roles. The service impl was explicitly passing a `null` user into `Current.create()` which now causes the container filter permission check to fail. Changing to `ContainerFilter.current()` restores this to a "no permission check" container filter.

Note that this was the only caller of `ContainerFilter.Type.Current.create(Container, User)` passing an explicit `null` value, although it's possible that a `null` user gets passed in to calling methods. For now, use an assert to flag any such cases.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6306

### DIFF
--- a/api/src/org/labkey/api/data/ContainerFilter.java
+++ b/api/src/org/labkey/api/data/ContainerFilter.java
@@ -554,6 +554,7 @@ public abstract class ContainerFilter
         public ContainerFilterWithPermission(Container c, User user)
         {
             super(c, user);
+            assert user != null : "User is required for permissions check!";
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/data/ContainerFilter.java
```diff
@@ -555,6 +554,7 @@ public static class ContainerFilterWithPermission extends ContainerFilter
         public ContainerFilterWithPermission(Container c, User user)
         {
             super(c, user);
+            assert user != null : "User is required for permissions check!";
```
labkey-jeckels: Any reason to not make this an `IllegalArgumentException` or `Objects.requireNonNull()` instead?

labkey-jeckels (approved):

api/src/org/labkey/api/data/ContainerFilter.java
labkey-adam: I'm not ready for this to throw on production deployments. Once we get test suite coverage and have more confidence that we're not passing in null we can amp it up.

labkey-adam:
SampleTypeLineageTest passed here: https://teamcity.labkey.org/buildConfiguration/LabKey_2411_Community_DailySuites_DailyBPostgres/3382234?buildTab=tests

PR #6323, opened 2025-02-14
Issue 52124, 52016 & 50768

XingY:
#### Rationale
Issue [52124](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52124): Sample Manager: after sample type is renamed, assay design that reference this sample type no longer show up as an Import Assay option for this sample
Issue [52016](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52016): Round tripping template from sample type design doesn't catch EnteredStorage field as a reserved field.
Issue [50768](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=50768): Need a better error message if date value is not in the expected format.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/684
- https://github.com/LabKey/platform/pull/6323
- https://github.com/LabKey/limsModules/pull/1167
- https://github.com/LabKey/testAutomation/pull/2280

#### Changes
-

### DIFF
--- a/experiment/src/org/labkey/experiment/PropertyQueryChangeListener.java
+++ b/experiment/src/org/labkey/experiment/PropertyQueryChangeListener.java
@@ -1,0 +1,99 @@
+package org.labkey.experiment;
+
+import org.jetbrains.annotations.NotNull;
+import org.labkey.api.collections.CaseInsensitiveHashMap;
+import org.labkey.api.data.Container;
+import org.labkey.api.data.ContainerFilter;
+import org.labkey.api.data.SQLFragment;
+import org.labkey.api.data.SqlExecutor;
+import org.labkey.api.data.TableInfo;
+import org.labkey.api.exp.OntologyManager;
+import org.labkey.api.query.QueryChangeListener;
+import org.labkey.api.query.SchemaKey;
+import org.labkey.api.security.User;
+
+import java.util.Collection;
+import java.util.List;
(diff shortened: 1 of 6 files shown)


experiment/src/org/labkey/experiment/PropertyQueryChangeListener.java
```diff
@@ -0,0 +1,99 @@
...
+
+        new SqlExecutor(pdTable.getSchema()).execute(updateSql);
+
+    }
+
+    @Override
+    public void queryChanged(User user, Container container, ContainerFilter scope, SchemaKey schema, @NotNull QueryProperty property, @NotNull Collection<QueryPropertyChange> changes)
+    {
+        // is there any other schema change other than assay renaming?
+        boolean isSchemaChange = schema.toString().toLowerCase().startsWith("assay.general.");
```
labkey-susanh, line 67: Anything need to be done for non-standard assays?

labkey-susanh (approved):

experiment/src/org/labkey/experiment/PropertyQueryChangeListener.java
XingY: I don't think we support renaming non-gpat assay designs.

labkey-susanh (approved):

PR #6325, opened 2025-02-14
Fix many issues with lists and special characters

labkey-adam:
#### Rationale
Lists with PK names containing special characters were having a rough time. They could not handle insert, update, or delete in many cases. They also showed mangled column captions when hovering over the grid headers.

Only one of the fixes is list-specific; the rest affect other table types. 

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52070
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52069

#### Changes
- `ListQueryUpdateService.getRow()`: ask the TableInfo for aliases instead of trying guess them. (This enables update.)
- Use `appendIdentifier()` in `PostgreSql91Dialect.addReselect()`. (This enables insert.)
- Refactor `CompareType.toWhereClause()` to return a `SQLFragment`, allowing the use of `appendIdentifier()`. Use it. (This enables delete... and probably resolves other issues.)
- Use `FieldKey.toDisplayString()` when displaying in the grid headers


api/src/org/labkey/api/data/CompareType.java
```diff
@@ -1202,9 +1202,9 @@ public CompareClause(@NotNull FieldKey fieldKey, CompareType comparison, Object
                 _paramVals = new Object[]{value};
         }
 
-        String toWhereClause(SqlDialect dialect, String alias)
+        SQLFragment toWhereClause(SqlDialect dialect, String alias)
```
labkey-matthewb, line 1205: happiness

api/src/org/labkey/api/data/CompareType.java
```diff
@@ -1697,11 +1699,13 @@ static class DateEqCompareClause extends DateCompareClause
...
         @Override
-        String toWhereClause(SqlDialect dialect, String alias)
+        SQLFragment toWhereClause(SqlDialect dialect, String alias)
         {
             String selectName = dialect.getColumnSelectName(alias);
             String q = dialect.isSqlServer() ? SS_CAST : "?";
-            return selectName + " >= " + q + " AND " + selectName + " < " + q;
+            return new SQLFragment()
+                .appendIdentifier(selectName)
+                .append(" >= ").append(q).append(" AND ").append(selectName).append(" < ").append(q);
```
labkey-matthewb: another appendIdentifier(selectName)

api/src/org/labkey/api/data/CompareType.java
```diff
@@ -1738,11 +1742,13 @@ static class DateNeqCompareClause extends DateCompareClause
...
         @Override
-        String toWhereClause(SqlDialect dialect, String alias)
+        SQLFragment toWhereClause(SqlDialect dialect, String alias)
         {
             String selectName = dialect.getColumnSelectName(alias);
             String q = dialect.isSqlServer() ? SS_CAST : "?";
-            return selectName + " < " + q + " OR " + selectName + " >= " + q;
+            return new SQLFragment()
+                .appendIdentifier(selectName)
+                .append(" < ").append(q).append(" OR ").append(selectName).append(" >= ").append(q);
```
labkey-matthewb: and here

api/src/org/labkey/api/data/CompareType.java
labkey-matthewb: consider making the oddly named getColumnSelectName() return SQLFragment

list/src/org/labkey/list/model/ListQueryUpdateService.java
```diff
@@ -124,10 +125,22 @@ protected Map<String, Object> getRow(User user, Container container, Map<String,
 
                     for (DomainProperty prop : _list.getDomain().getProperties())
                     {
-                        Object value = getField(raw, prop.getName());
-
-                        if (null != value)
-                            ret.put(prop.getName(), value);
+                        String propName = prop.getName();
```
labkey-matthewb, line 128: The ColumnInfo's in the TableInfo know the names of the columns in the result.  E.g.
for (var col : queryTable.getColumns())
    value = col.getValue(raw)

api/src/org/labkey/api/data/CompareType.java
labkey-adam: Thanks.
I added a comment to change this since this will touch a five different repos.

api/src/org/labkey/api/data/CompareType.java
labkey-adam: Thanks... got it.

list/src/org/labkey/list/model/ListQueryUpdateService.java
labkey-adam: I like it... so it's okay to key the map with `col.getName()` instead of `prop.getName()`?

labkey-matthewb: Unless someone is getting tricky, I think the map returned from TableSelector will be keyed with ColumnInfo.getAlias() and ColumnInfo.get() will use that.

labkey-adam: I agree that lookups into the `raw` map via the ColumnInfo should work fine. My question is really about the return map. That is currently keyed with `prop.getName()` and has far fewer entries vs. one constructed based on getColumns(). I'm concerned about making major changes to the return map.

labkey-matthewb: I don't know what is expected there.

labkey-adam: Neither do I... so I left it largely as it was before. But with ColumnInfo getting the value.

labkey-matthewb (approved):

labkey-adam:
BVT and BVT-EHR suites all green (above)

Daily postgres is all green, except for known, existing failures: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/3385122

PR #6330, opened 2025-02-17
Add StringUtilsLabKey.truncateToUtf8ByteLimit()

labkey-adam:
#### Rationale
This may come in handy when addressing [Issue 52271: List PK name gets truncated on table create](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52271)

Includes an extensive junit test


api/src/org/labkey/api/util/StringUtilsLabKey.java
```diff
@@ -953,5 +988,75 @@ public void testIsValidIdentifier()
...
+            for (int i = 0; i < 10; i++)
+            {
+                Collections.shuffle(uniqueChars);
+                String test = String.join("", uniqueChars);
+                testTruncateSingleString(test);
+            }
+        }
+
+        // Test truncating this string at every possible byte length from 0 to byte length + 1
+        private void testTruncateSingleString(String s)
```
labkey-jeckels, line 1052: I think this would pass if we always truncated down to zero characters. I suggest adding a few hard-coded strings with expected truncations. Maybe something approximately like:

assertEquals("abc", truncateToUtf8ByteLimit("abc", 3));
assertEquals("ab", truncateToUtf8ByteLimit("abc", 2));
assertEquals("☃☃", truncateToUtf8ByteLimit("☃☃", 10));
assertEquals("☃", truncateToUtf8ByteLimit("☃☃", 2));
"☃".equals(truncateToUtf8ByteLimit("☃☃", 3));

labkey-jeckels (approved):

api/src/org/labkey/api/util/StringUtilsLabKey.java
labkey-adam: Done

PR #6331, opened 2025-02-17
Issue 52280: Replicate stats not generated for preexisting measures

labkey-nicka:
#### Rationale
This addresses [Issue 52280](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52280) by distinguishing and handling when a) the replicate stat domain is new and b) the domain is not new but a measure exists that is not yet in the replicate stat domain.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1170

#### Changes
- Check for new domain
- Check for pre-existing measure not yet in the replicate stat domain
- Integration tests added in associated PR.


labkey-klum (approved):
good catch

PR #6332, opened 2025-02-17
Delegate folder name validation to the server. 

labkey-klum:
#### Related issues
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52288
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52087

#### Rationale
- Delegate folder name validation to the server (for more consistency)
- Simplify the error message to match the style used in our applications

I believe this is the direction we have decided on (see attached issue), let me know if that isn't the recommended approach.


api/src/org/labkey/api/data/Container.java
```diff
@@ -806,7 +806,7 @@ public static boolean isLegalName(String name, boolean isProject, StringBuilder
 
         if (!FileUtil.isLegalName(name))
         {
-            error.append("Folder name must be a legal filename and not contain one of '/', '\\', ':', '?', '<', '>', '*', '|', '\"', '^'");
+            error.append("Folder name must be a legal filename and not contain any of these characters: /\\:?<>*|\"^'");
```
cnathe: I didn't try it, but did the issue also mention that semicolon should be in this list of illegal chars?

cnathe (approved):

api/src/org/labkey/api/data/Container.java
labkey-klum: It's a little strange since semicolon is tested separately and has it's own specific error message. I didn't know why this was the case or if it should all be consolidated into the regex.

```     if (!FileUtil.isLegalName(name))
        {
            error.append("Folder name must be a legal filename and not contain one of '/', '\\', ':', '?', '<', '>', '*', '|', '\"', '^'");
            return false;
        }

        if (name.contains(";"))
        {
            error.append("Semicolons are not allowed in folder names.");
            return false;
        }

        if (name.startsWith("@"))
        {
            error.append("Folder name may not begin with '@'.");
            return false;
        }
```

labkey-nicka: Rather than hardcoding the error message characters can it refer to `FileUtils.ILLEGAL_CHARS_STRING`?

labkey-klum: sure, that's a good idea.

labkey-danield (approved):

PR #6336, opened 2025-02-18
Issue 52099: Don't allow sample statuses that differ only by case

labkey-susanh:
#### Rationale
Issue [52099](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52099)

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1174

#### Changes
- Update `DataStatesTableInfo` with a conditional `isDuplicate` check.


api/src/org/labkey/api/qc/DataStateHandler.java
```diff
@@ -28,6 +28,10 @@ public interface DataStateHandler<FORM extends AbstractManageDataStatesForm>
     boolean isStateInUse(Container container, DataState state);
     boolean isBlankStatePublic(Container container);
     boolean isRequireCommentOnQCStateChange(Container container);
+    default boolean caseInsensitiveDuplicateAllowed()
+    {
+        return true;
```
cnathe, line 33: This default makes sense as it makes sure the behavior doesn't change for the non-sample status related states, but it seems odd to allow for QC states that just differ in casing for those cases too. For another day I suppose.

core/src/org/labkey/core/query/DataStatesTableInfo.java
```diff
@@ -148,6 +171,8 @@ protected Map<String, Object> insertRow(User user, Container container, Map<Stri
         {
             if (!validateLabel(row, false))
                 throw new QueryUpdateServiceException("Label cannot be blank.");
+            if (isDuplicate(String.valueOf(row.get("label")), container))
```
cnathe, line 174: verify minor, but you could just pass row to `isDuplicate` and then have `String.valueOf(row.get("label"))` within that method so that that part is not duplicated at line 152 and 174

cnathe (approved):

core/src/org/labkey/core/query/DataStatesTableInfo.java
labkey-susanh: I think passing in the actual value is slightly more general, so even though not used in a general way, I think I'll leave this.

api/src/org/labkey/api/qc/DataStateHandler.java
labkey-susanh: Yeah, I didn't want to have too much unintended impact with this change. It's easy enough to update the default (or even take out this method altogether) at a later date.

PR #6343, opened 2025-02-19
Include websocket "ws:" for reactjs.hot.reload in allowed connections for dev mode

cnathe:
#### Rationale
The related PR added localhost to the allowed connections in dev mode to support the client side application dev for hot module reloading. We need to also include websocket "ws:" in that connect-src.

Screenshot 2025-02-18 at 11 27 04 AM

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6295/

#### Changes
- Include websocket "ws:" for reactjs.hot.reload in allowed connections for dev mode

PR #6345, opened 2025-02-19
Support adding module-specific resource hosts to more CSP directives

labkey-adam:
#### Rationale
Modules need the ability to allow specific font, style, frame, etc. hosts that they require. This is the programmatic part; admin UI will come in a separate PR.

[Spec](https://docs.google.com/document/d/1aUreAxpjOi75Iqx0ZfdRJLrL-pUGFWAPYLdIsJTcO2A/edit?tab=t.0#heading=h.4rg4h5vuek0)

#### Related Pull Requests
- https://github.com/LabKey/server/pull/996

### DIFF
--- a/api/src/org/labkey/api/security/Directive.java
+++ b/api/src/org/labkey/api/security/Directive.java
@@ -1,0 +1,30 @@
+package org.labkey.api.security;
+
+/**
+ * All CSP directives that support substitutions. These constant names are persisted to the database, so be careful with
+ * any changes. If adding a Directive, make sure to add the corresponding substitutions to application.properties.
+ */
+public enum Directive
+{
+    Connection("connect-src"),
+    Font("font-src"),
+    Frame("frame-src"),
+    Style("style-src");
+
+    private final String _cspDirective;
+
+    Directive(String cspDirective)
+    {
+        _cspDirective = cspDirective;
+    }
+
+    public String getCspDirective()
+    {
+        return _cspDirective;
+    }
+
+    public String getSubstitutionKey()
+    {
+        return "LABKEY.ALLOWED." + name().toUpperCase() + ".SOURCES";
+    }
+}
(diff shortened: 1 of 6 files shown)


api/src/org/labkey/api/security/Directive.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public String getCspDirective()
+    {
+        return _cspDirective;
+    }
+
+    public String getSubstitutionKey()
+    {
+        return "LABKEY.ALLOWED." + name().toUpperCase() + ".SOURCES";
```
labkey-matthewb: Do we still need "LABKEY.ALLOWED"?  The filter is not in the least generic any more.

labkey-adam: I'm fine removing it. Next commit.

api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
```diff
@@ -71,9 +71,13 @@ public String getHeaderName()
 
     static
     {
-        // ReactJS hot reload uses localhost port 3001. If in dev mode, allow browser to access that port.
+        // ReactJS hot reload uses localhost port 3001. If in dev mode, allow browser to access that port for fonts
+        // and connections.
         if (AppProps.getInstance().isDevMode())
-            registerAllowedConnectionSource("reactjs.hot.reload", "localhost:3001");
+        {
+            registerAllowedSources(Directive.Connection, "reactjs.hot.reload", "localhost:3001");
```
labkey-nicka: Is the fix in #6343 appropriate? Should we just do that here?

labkey-adam: Yes, we're doing that here. Cory has an edit to add `ws:`.

labkey-klum (approved):

PR #6350, opened 2025-02-19
refactor ParseDateTimeEN

labkey-matthewb:
#### Rationale
Part I of DateUtil refactor in preparation for more sophisticated handing of TimeZones and validation rules

* Break out parseDateTimeEN() as new class that implements org.apache.commons.lang3.time.DateParser
* Separate scanning and converting

I left the old code in place but commented out.  This is for comparison by reviewers if that's useful.
I'll delete before merge.

#### Related Pull Requests
- 

#### Changes
-


api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -405,8 +437,34 @@ private static long parseDateTimeUS(String s, DateTimeOption option, boolean str
...
+            var lenient = !strict;
+            DateParser parser = ParseDateTimeEN.getInstance(_localeDefault, _timezoneDefault, option, md, lenient);
+            Date date = parser.parse(s);
+            return date.getTime();
+        }
+        catch (IllegalArgumentException|ParseException ex)
+        {
+            throw new ConversionException("Could not parse date '" + s + "'", ex);
+        }
+/*
```
labkey-matthewb: Note to self: delete before merge

api/src/org/labkey/api/util/time/ParseDateTimeEN.java
```diff
@@ -0,0 +1,629 @@
...
+        @Override
+        public int getField()
+        {
+            return DAY_OF_WEEK;
+        }
+
+        @Override
+        public int getValue()
+        {
+            return ordinal()+1;
```
XingY, line 187: Would be good to use `Calendar.SUNDAY` to avoid having to do this confusing `+1`

api/src/org/labkey/api/util/time/ParseDateTimeEN.java
```diff
@@ -0,0 +1,629 @@
...
+     */
+
+    private static final int currentYear = new GregorianCalendar().get(YEAR);
+    private static final int twoDigitCutoff = (currentYear - 80) % 100;
+    private static final int defaultCentury = (currentYear - 80) - twoDigitCutoff;
+
+    interface CalendarConstant
+    {
+        int getField();
+        int getValue();
```
XingY, line 153: Better to use Calendar constants
`    enum Month implements CalendarConstant
    {
        january(Calendar.JANUARY),
        february(Calendar.FEBRUARY);

        private final int _value;
        Month(int value)
        {
            _value = value;
        }
        @Override
        public int getField()
        {
            return MONTH;
        }

        @Override
        public int getValue()
        {
            return _value;
        }
    }`

api/src/org/labkey/api/util/time/ParseDateTimeEN.java
```diff
@@ -0,0 +1,629 @@
...
+
+        @Override
+        public int getValue()
+        {
+            return ordinal();
+        }
+    }
+
+    @SuppressWarnings("PointlessArithmeticExpression")
+    enum TZ
```
XingY, line 209: I'm not understanding this enum... Is this just a subset that we chose to start with?

api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -304,118 +304,120 @@ public static String toISO(java.util.Date d)
...
+                                cal.get(DAY_OF_MONTH) == parts.getDayOfMonth());
+                        return true;
+                    }
+                },
+        TimeOnly
+                {
+                    @Override
+                    public boolean toCalendar(CalendarParts parts, Calendar cal)
+                    {
+                        if (!cal.isLenient())
```
XingY: I think perhaps we shouldn't be so lenient with time parsing. 
The current parsing doesn't work right with military time, for example, "1830" returns "00:00". This is because 1830 is parsed to be the year, and the hour/minutes are all not set. For timeOnly, we should probably check that at least Hours is set.
Also "19999 pm" returns "12:00", since "19999" is parsed into year and pm is parsed to 12 hours. We should probably check that if Year/Month/Day is set, then all of Year/Month/Day needs to be set.

api/src/org/labkey/api/util/time/ParseDateTimeEN.java
labkey-matthewb: This is basically still from the original javascript parser.  I'm surprised this has worked as long as it has.  I planned to make this more general with the "Phase II"  (there is some commented code that hints at this work).

api/src/org/labkey/api/util/DateUtil.java
labkey-matthewb: I was a little surprised at how loose parseTime() is too.  For now, I was trying to get through the refactor without changing how things work.  I really like the idea of requiring year/month/day all be set together.

api/src/org/labkey/api/util/time/ParseDateTimeEN.java
labkey-matthewb: I thought about doing that, and decided it was just a bunch of extra typing.

api/src/org/labkey/api/util/DateUtil.java
labkey-matthewb: Also, the scanner knows it's parsing time-only.  We could use that information to prefer interpreting 4 digits as a time... I'm happy to just error for now, unless this is a customer scenario.

XingY: I think error out rather than guessing the wrong value should be the preferred approach given our latest priority. I'm also fine with merging this as is. I can work on the TimeOnly part to be more strict as a follow up task.

api/src/org/labkey/api/util/time/ParseDateTimeEN.java
XingY: If not, should add a comment why +1 is added so it doesn't get dropped unintentionally in the future.

XingY (approved):

api/src/org/labkey/api/util/DateUtil.java
labkey-matthewb: Added these examples to the tests

PR #6352, opened 2025-02-20
Issue 51877: Allow calculated fields for list title column

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51877
We were previously filtering out the calculated fields from the options for the title column property of a list. It was still possible to set a calculated field as the title column via the XML metadata and it worked as expected, so this PR makes that possible in the list designer UI as well.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1723
- https://github.com/LabKey/platform/pull/6352

#### Changes
- ListTable.findTitleColumn to look for a list titleColumn via calculated field

### DIFF
--- a/list/src/org/labkey/list/model/ListTable.java
+++ b/list/src/org/labkey/list/model/ListTable.java
@@ -447,9 +447,18 @@ public class ListTable extends FilteredTable<ListQuerySchema> implements Updatea
         if (listDef.getTitleColumn() != null)
         {
             ColumnInfo titleColumn = getColumn(listDef.getTitleColumn());
-
             if (titleColumn != null)
                 return titleColumn.getName();
+            else
+            {
+                // Issue 51877: a list titleColumn might be a calculated field
+                List<FieldKey> calcFieldKeys = DomainUtil.getCalculatedFieldsForDefaultView(this);
+                for (FieldKey calcFieldKey : calcFieldKeys)
+                {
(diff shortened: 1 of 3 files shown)


list/src/org/labkey/list/model/ListTable.java
```diff
@@ -447,9 +447,18 @@ private String findTitleColumn(ListDefinition listDef, ColumnInfo colKey)
...
-
             if (titleColumn != null)
                 return titleColumn.getName();
+            else
+            {
+                // Issue 51877: a list titleColumn might be a calculated field
+                List<FieldKey> calcFieldKeys = DomainUtil.getCalculatedFieldsForDefaultView(this);
+                for (FieldKey calcFieldKey : calcFieldKeys)
+                {
+                    if (listDef.getTitleColumn().equalsIgnoreCase(calcFieldKey.getName()))
```
labkey-susanh, line 458: Since this seems to be comparing a name and a fieldKey, I just had to check, but it seems to be working ok.

labkey-susanh (approved):

list/src/org/labkey/list/model/ListTable.java
cnathe: I doubled checked in the DB table exp.list that the titleColumn property is storing the unencoded column name

PR #6354, opened 2025-02-20
Tell truncation to use UTF-8 instead of platform default

labkey-jeckels:
#### Rationale
https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_InternalSuites_OldDependenciesSqlserver/3389654

We need to specify our preferred character encoding, and not rely on the platform default

#### Changes
- UTF-8 forever


labkey-adam (approved):
Right, thanks, Josh!

labkey-tchad (approved):

PR #6355, opened 2025-02-20
Add support for image-src external sources

labkey-adam:
#### Rationale
At least one module needs to pull images from an external source

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6345

### DIFF
--- a/api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
+++ b/api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
@@ -240,15 +240,15 @@ public class ContentSecurityPolicyFilter implements Filter
             // Multi-line for readability, but notice that newlines are replaced before assignment
             String expected = """
                 default-src 'self' https: http: ;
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
```diff
@@ -303,6 +303,10 @@ public void testSubstitutionMap()
             assertEquals(4, ALLOWED_SOURCES.size());
             assertEquals(4, _substitutionMap.size());
 
+            registerAllowedSources(Directive.Image, "image", "ImageSource", "BetterImageStore");
+            assertEquals(5, ALLOWED_SOURCES.size());
+            assertEquals(5, _substitutionMap.size());
+
             // Restore the previous ALLOWED_SOURCES
             synchronized (ALLOWED_SOURCES_LOCK)
```
labkey-jeckels: This should be in a `finally` block.

labkey-jeckels (approved):

PR #6357, opened 2025-02-21
Issue 52355: Assure we always have non-blank rows for rows provided for cross-type import

labkey-susanh:
#### Rationale
Issue [52355](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52355): When writing out split files for cross-type import, we will include the sample type as well so we are assured that there is always data for each row in the original file. Without this, the blank rows are ignored.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1190

#### Changes
- Include SampleType in the data fields written to the split files during cross-type import

### DIFF
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -2961,7 +2961,7 @@ public class ExpDataIterators
                 ColumnInfo colInfo = getColumnInfo(i);
                 String name = colInfo.getName();
                 String lcName = name.toLowerCase();
-                if (i == _typeColIndex) // Issue 52355: assure we have some data in the row by including the type
+                if (_typeColIndex != null && _typeColIndex == i) // Issue 52355: assure we have some data in the row by including the type
                 {
                     fieldIndexes.add(i);
                     header.add(_typeColName);
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -2961,7 +2961,12 @@ public class ExpDataIterators
                 ColumnInfo colInfo = getColumnInfo(i);
                 String name = colInfo.getName();
                 String lcName = name.toLowerCase();
-                if (validFields.contains(name))
+                if (i == _typeColIndex)
+                {
+                    fieldIndexes.add(i);
+                    header.add("SampleType");
+                }
+                else if (validFields.contains(name))
                 {
                     fieldIndexes.add(i);
                     header.add(_tsvWriter.quoteValue(name));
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2961,7 +2961,12 @@ private TypeData createSampleHeaderRow(ExpSampleTypeImpl sampleType, Container c
                 ColumnInfo colInfo = getColumnInfo(i);
                 String name = colInfo.getName();
                 String lcName = name.toLowerCase();
-                if (validFields.contains(name))
+                if (i == _typeColIndex)
+                {
+                    fieldIndexes.add(i);
```
cnathe, line 2966: can we add the issue number in a comment here?

cnathe (approved):

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2961,7 +2961,12 @@ private TypeData createSampleHeaderRow(ExpSampleTypeImpl sampleType, Container c
                 ColumnInfo colInfo = getColumnInfo(i);
                 String name = colInfo.getName();
                 String lcName = name.toLowerCase();
-                if (validFields.contains(name))
+                if (_typeColIndex != null && _typeColIndex == i) // Issue 52355: assure we have some data in the row by including the type
+                {
```
XingY, line 2965: With this, cross folder import now behaves differently from single type import when it comes to no data import. Single type import will skip empty rows, but cross type creates one. I'm wondering if the change should be fixing the success msg and continue to skip those rows? 

Also, "Sample Type" is now included in the audit detail record, not sure if it matters.

labkey-susanh: The reason this is even seen as data loss is because we were ignoring the lines in the file that contain only the type name. To the user providing the file, those lines are not blank. It is, of course, seemingly very unlikely that anyone would provide a file that has only the type name and nothing else, but that is the case we are addressing.

Separately, in develop, Cory has a change that will update the messaging to reflect the number of samples actually created, should there be some other case that we aren't seeing currently that would cause some lines to be ignored. 

Including "Sample Type" in the audit record seems fine.

cnathe: I agree that the cross type import "blank rows" case here is a little different than the single type import because the row isn't technically blank as it has a valid sample type defined.

Here is the PR for the client side update related to the success message mismatch part of this issue: https://github.com/LabKey/labkey-ui-premium/pull/691. As Susan mentioned, this fix she just merged will make this part irrelevant for this scenario but will hopefully help if we encounter another issue in the crosstype import case.

PR #6363, opened 2025-02-23
Issue 52191: LKSM/LKB: Sample File Import accepts invalid time integers

XingY:
#### Rationale
LenientTimeConverter can yield unexpected results. For example, "0030" is parsed to "06:00", with "30" being the hours of the day, which is then converted to "6" by 30%24, which is almost never what one expects. This PR switch our Time converter from being lenient to being strict, aligning with our general approaches to be more strict about data validation. 

#### Related Pull Requests
- 

#### Changes
- Use strict mode when parsing time

### DIFF
--- a/api/src/org/labkey/api/data/ConvertHelper.java
+++ b/api/src/org/labkey/api/data/ConvertHelper.java
@@ -159,7 +159,7 @@ public class ConvertHelper implements PropertyEditorRegistrar
         _register(new NullSafeConverter(new _IntegerConverter()), Integer.class);
         _register(new _IntegerConverter(), Integer.TYPE);
         _register(new NullSafeConverter(new LenientSqlDateConverter()), java.sql.Date.class);
-        _register(new NullSafeConverter(new TimeConverter()), java.sql.Time.class);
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/data/ConvertHelper.java
```diff
@@ -306,7 +306,7 @@ public Object convert(Class clss, Object o)
             if (o instanceof java.sql.Time)
                 return o;
 
-            return DateUtil.fromTimeString(o.toString(), false);
+            return DateUtil.fromTimeString(o.toString(), true);
```
labkey-matthewb, line 309: Yes

api/src/org/labkey/api/data/ConvertHelper.java
```diff
@@ -295,7 +295,7 @@ public Object convert(Class clss, Object o)
         }
     }
 
-    public static class LenientTimeConverter implements Converter
```
labkey-matthewb: In this context, I think Lenient just means accepts null without blowing up.  So the old name seems consistent.

labkey-matthewb (approved):

PR #6365, opened 2025-02-24
Make session key use act more like API key use

labkey-adam:
#### Rationale
Current behavior: session key use authenticates the current request but doesn't alter the session. This works for clients that proactively send the session key on every request, but works inconsistently on clients that require a challenge/response (e.g., Java client API). Actions that require login work fine (although every request requires two round trips), but those that don't (e.g., whoAmI, getContainers, etc.) never get the session key and therefore are always considered "guest". https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=52302

New behavior: when a valid session key is presented, set a `JESSIONID` cookie in the response with the value of the impersonated session ID. The new connection now uses that session directly, as if it's another browser tab.

#### Changes
- Also log the URI when initiating the authentication process

### DIFF
--- a/api/src/org/labkey/api/security/SecurityManager.java
+++ b/api/src/org/labkey/api/security/SecurityManager.java
@@ -19,7 +19,9 @@ package org.labkey.api.security;
 import jakarta.mail.Message;
 import jakarta.mail.MessagingException;
 import jakarta.mail.internet.MimeMessage;
+import jakarta.servlet.http.Cookie;
 import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
 import jakarta.servlet.http.HttpSession;
 import jakarta.websocket.server.HandshakeRequest;
 import org.apache.commons.codec.binary.Base64;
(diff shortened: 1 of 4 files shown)


labkey-tchad (approved):
Works well for me

api/src/org/labkey/api/security/SecurityManager.java
```diff
@@ -549,8 +551,11 @@ public static Pair<User, HttpServletRequest> attemptAuthentication(HttpServletRe
 
                 if (null != session)
```
labkey-matthewb, line 554: Is there now a scenario where request.getSession() == session where we can just return?

labkey-matthewb (approved):

api/src/org/labkey/api/security/SecurityManager.java
labkey-adam: Yes, if the client sends both the session cookie and a session key. Probably not likely, but certainly possible.

labkey-adam: Updated to check and no-op if current session matches what we would have set

PR #6368, opened 2025-02-24
fix FormHandlerAction for action that sends response in handlePost()

labkey-matthewb:
#### Rationale
Issue 51755

#### Related Pull Requests
- 

#### Changes
-


labkey-jeckels (approved):
Is this worth a comment somewhere in `FormHandlerAction` on how subclasses can support this pattern?

labkey-matthewb:
> Is this worth a comment somewhere in `FormHandlerAction` on how subclasses can support this pattern?

Maybe?  But there are other classes that seem a better fit.  Like ExportAction

PR #6359, opened 2025-02-21
Admin UI for adding external hosts to a variety of directives

labkey-adam:
#### Rationale
Admins need a way to allow browsers to retrieve resources from trusted external hosts as CSP exceptions.

https://docs.google.com/document/d/1aUreAxpjOi75Iqx0ZfdRJLrL-pUGFWAPYLdIsJTcO2A/edit?tab=t.0#heading=h.4rg4h5vuek0

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6345

#### Changes
- Provide new admin page, action, persistence, and registration of admin-provided external hosts. Currently supports `connect-src`, `font-src`, `frame-src`, `image-src`, and `style-src`, though others can be added easily (see `Directive`).
- Add new startup properties to configure same
- Reduce CSP filter's per-request work to a single substitution
- Metrics for all configured CSPs (type, CSP substituted, CSP not substituted)

PR #6369, opened 2025-02-25
New metrics to track report and enforce CSPs

labkey-jeckels:
#### Rationale
Let's keep track of the Content Security Policies that are in use

#### Changes
- Two new metrics, tracking the templated CSPs deployed: `modules.API.contentSecurityPolicy.enforce` and `modules.API.contentSecurityPolicy.report`

### DIFF
--- a/api/src/org/labkey/api/ApiModule.java
+++ b/api/src/org/labkey/api/ApiModule.java
@@ -149,6 +149,7 @@ import org.labkey.api.settings.LookAndFeelProperties;
 import org.labkey.api.settings.OptionalFeatureService.FeatureType;
 import org.labkey.api.settings.OptionalFeatureStartupListener;
 import org.labkey.api.settings.WriteableLookAndFeelProperties;
+import org.labkey.api.usageMetrics.UsageMetricsService;
 import org.labkey.api.util.ChecksumUtil;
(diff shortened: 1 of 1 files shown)


labkey-adam:
I've added similar metrics in my PR: https://github.com/LabKey/platform/pull/6359

I'm including only CSP version, but I could easily add the full CSP. At a minimum, let's get my PR merged and iterate on that.

labkey-jeckels:
> I've added similar metrics in my PR: #6359
> 
> I'm including only CSP version, but I could easily add the full CSP. At a minimum, let's get my PR merged and iterate on that.

Great, thanks for the note. I won't do anything with this PR for the moment. Ultimately, I'm hoping for an approach that helps us understand which on-premise deployments have a CSP too.

labkey-adam:
> > I've added similar metrics in my PR: #6359
> > I'm including only CSP version, but I could easily add the full CSP. At a minimum, let's get my PR merged and iterate on that.
> 
> Great, thanks for the note. I won't do anything with this PR for the moment. Ultimately, I'm hoping for an approach that helps us understand which on-premise deployments have a CSP too.

Yep, me too. I'll look at adding the full CSP to my thing and we can go from there. Feel free to review, if you want.

PR #6381, opened 2025-02-26
Issue 52394: Move indexing out of invalidate method and into more post-commit task more explicitly.

labkey-susanh:
#### Rationale
[Issue 52394](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52394) Data class and sample type domain kinds were overriding the `invalidate` method in order to do an indexing of the associated data class or sample type. The invalidate method is called in a few places, including upon saving the table before having updated the domain. This can cause a bit of a race condition where the search indexer brings the old domain into the cache and then subsequent retrieval uses that old domain, even after the new domain has been saved. We really don't need the indexing to happen until after the transaction has committed, so this PR adds explicit post-commit actions to do the indexing after the caches have been cleared. 

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1196

#### Changes
- Remove override of `invalidate` for `DataClassDomainKind` and `SampleTypeDomainKind`
- Add explicit post-commit tasks for indexing updated domain classes and sample types

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpDataClassImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpDataClassImpl.java
@@ -15,11 +15,9 @@
  */
 package org.labkey.experiment.api;
 
-import com.fasterxml.jackson.core.type.TypeReference;
 import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
-import org.labkey.api.collections.CaseInsensitiveHashMap;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerFilter;
 import org.labkey.api.data.DbSequence;
@@ -295,7 +293,6 @@ public class ExpDataClassImpl extends ExpIdentifiableEntityImpl<DataClass> imple
         }
 
         ExperimentServiceImpl.get().clearDataClassCache(getContainer());
-        ExperimentServiceImpl.get().indexDataClass(this);
     }
 
     @Override
--- a/experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
@@ -827,11 +827,6 @@ public class ExpSampleTypeImpl extends ExpIdentifiableEntityImpl<MaterialSource>
     @Override
     public void save(User user)
     {
-        save(user, false);
-    }
-
-    public void save(User user, boolean skipCleanUpTasks /* index and cache might have been called explicitly in a postcommit task*/)
-    {
         boolean isNew = _object.getRowId() == 0;
 
         //Issue 51024: When materialLSIDprefix is set via XAR, naming collisions can happen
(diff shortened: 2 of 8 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -8013,6 +8016,7 @@ public ValidationException updateDataClass(@NotNull Container c, @NotNull User u
             if (!errors.hasErrors())
             {
                 transaction.addCommitTask(() -> clearDataClassCache(c), DbScope.CommitTaskOption.IMMEDIATE, POSTCOMMIT, POSTROLLBACK);
+                transaction.addCommitTask(() -> ExperimentServiceImpl.get().indexDataClass(getDataClass(c, dataClass.getName())), POSTCOMMIT, POSTROLLBACK);
```
labkey-adam: If we rollback, is there really a need to reindex?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -890,6 +893,10 @@ public ExpSampleTypeImpl createSampleType(Container c, User u, String name, Stri
                     else
                         ExperimentService.get().ensureDataTypeContainerExclusionsNonAdmin(ExperimentService.DataTypeForExclusion.DashboardSampleType, st.getRowId(), c, u);
                     transaction.addCommitTask(() -> clearMaterialSourceCache(c), DbScope.CommitTaskOption.IMMEDIATE, POSTCOMMIT, POSTROLLBACK);
+                    transaction.addCommitTask(() -> {
+                        SampleTypeService.get().indexSampleType(SampleTypeService.get().getSampleType(domain.getTypeURI()));
+                    }, POSTCOMMIT, POSTROLLBACK);
```
labkey-adam: Same question

labkey-adam (approved):
Just a question about whether indexing after rollback is needed in two places

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-nicka: 1. Why index after rolling back? Seems like we only want to index after a change has successfully been made. Same feedback for similar change in `SampleTypeServiceImpl`.
2. nit: This is already in `ExperimentServiceImpl` so can just call `indexDataClass()` directly.

experiment/src/org/labkey/experiment/api/ExpDataClassImpl.java
```diff
@@ -295,7 +293,6 @@ public void save(User user)
         }
 
         ExperimentServiceImpl.get().clearDataClassCache(getContainer());
-        ExperimentServiceImpl.get().indexDataClass(this);
```
labkey-nicka, line 298: With this I believe we'll no longer index upon data class creation. I'd recommend adding a similar commit task for indexing the data class in `ExperimentServiceImpl.createDataClass()`.

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
```diff
@@ -865,8 +865,6 @@ public void save(User user, boolean skipCleanUpTasks /* index and cache might ha
             // NOTE cacheMaterialSource() of course calls transactioncache.put(), which does not alter the shared cache! (BUG?)
             // Just call uncache(), and let normal cache loading do its thing
             SampleTypeServiceImpl.get().clearMaterialSourceCache(getContainer());
-
```
labkey-nicka, line 868: nit: With these changes I think the `skipCleanUpTasks` bit can be removed and just always call clear cache. This is the same as what `ExpDataClassImpl.save()` does.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: good point. removed.

labkey-susanh: updated

experiment/src/org/labkey/experiment/api/ExpSampleTypeImpl.java
labkey-susanh: done.

experiment/src/org/labkey/experiment/api/ExpDataClassImpl.java
labkey-susanh: Added

labkey-nicka (approved):

PR #6382, opened 2025-02-26
Demote logging about invalid password reset to WARN

labkey-jeckels:
#### Rationale
Attempting to reset a password for an account that doesn't exist isn't a server error. It's useful to log, but we can treat it as a warning.

#### Changes
- Demote logging from `ERROR` to `WARN`
- Misc code cleanup


labkey-willm:
makes sense to me, but can't speak to the code changes.

labkey-adam (approved):

PR #6384, opened 2025-02-26
Put study protocol designer tools under an optional flag

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51726

This is the first phase to put some of the study design tools behind an optional feature flag. In this initial phase we will hide the study manage links as well as web parts for these features in order to get a smaller change in place for the upcoming ESR release. In a subsequent release we plan to merge the larger change to hide the tables related to these features under the same flag.

#### Related PR
https://github.com/LabKey/testAutomation/pull/2303

### DIFF
--- a/study/src/org/labkey/study/StudyModule.java
+++ b/study/src/org/labkey/study/StudyModule.java
@@ -83,6 +83,7 @@ import org.labkey.api.study.Study;
 import org.labkey.api.study.StudyInternalService;
 import org.labkey.api.study.StudyService;
 import org.labkey.api.study.StudyUrls;
+import org.labkey.api.study.StudyUtils;
 import org.labkey.api.study.TimepointType;
 import org.labkey.api.study.importer.ImportHelperService;
 import org.labkey.api.study.model.CohortService;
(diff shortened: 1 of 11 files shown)


study/src/org/labkey/study/StudyModule.java
```diff
@@ -405,6 +425,11 @@ protected void startupAfterSpringConfig(ModuleContext moduleContext)
                 "Allow unprovisioned, query-based dataset snapshots to be created.",
                 false);
 
+        AdminConsole.addOptionalFeatureFlag(new OptionalFeatureFlag(StudyUtils.STUDY_DESIGN_FEATURE_FLAG,
+                "Study Protocol Design Tools",
+                "This option adds support for the study protocol and vaccine design tools.",
+                false, false, FeatureType.Optional));
```
labkey-klum: I could use extra input to decide on the name and description.

study/test/src/org/labkey/test/tests/study/StudyProtocolDesignerTest.java
```diff
@@ -105,13 +107,21 @@ public static void doSetup()
     {
         StudyProtocolDesignerTest initTest = (StudyProtocolDesignerTest)getCurrentTest();
 
+        OptionalFeatureHelper.enableOptionalFeature(initTest.createDefaultConnection(), "studyDesignFlag");
```
labkey-adam: For extra credit, tests could capture the previous setting value and restore it in `doCleanup()`. Though it's unlikely that folks will have this enabled.

study/src/org/labkey/study/StudyModule.java
labkey-adam: The wording seems fine to me. You could consider marking this as a `Deprecated` feature for now. While we aren't removing this from Atlas, we are removing it for everyone else... and the flag itself will disappear once we move the code to a separate module.

labkey-adam (approved):

cnathe (approved):
I pulled down the fb and verified that the manage study links and webparts are no longer shown when the optional flag is disabled. 

Do we want to also use this flag to optionally show/hide the related study design tables in the study schema? Or will that just wait until the module split.

study/src/org/labkey/study/StudyModule.java
labkey-klum: Yep, good idea

PR #6386, opened 2025-02-27
Issue 52417: StackOverflowError analyzing queries with linked core schema

labkey-jeckels:
#### Rationale
We can break a recursion cycle when checking linked schema FKs.

#### Changes
- Use the SchemaKey from the schema without creating the full TableInfo
- Scope the dependency error info to the dependency analysis

### DIFF
--- a/api/src/org/labkey/api/query/QuerySchema.java
+++ b/api/src/org/labkey/api/query/QuerySchema.java
@@ -138,6 +138,11 @@ public interface QuerySchema extends SchemaTreeNode, ContainerUser
     @NotNull
     String getSchemaName();
 
+    default SchemaKey getSchemaPath()
+    {
+        return SchemaKey.decode(getSchemaName());
+    }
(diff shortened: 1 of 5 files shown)


api/src/org/labkey/api/query/QuerySchema.java
```diff
@@ -138,6 +138,11 @@ default ContainerFilter getDefaultLookupContainerFilter()
     @NotNull
     String getSchemaName();
 
+    default SchemaKey getSchemaPath()
```
labkey-matthewb, line 141: UserSchema already has getPath(), we should probably use that.

labkey-matthewb: My bad it has both getPath() and getSchemaPath().  That doesn't seem like a good idea.

labkey-matthewb (approved):

PR #6387, opened 2025-02-27
Issue 52180: Naming Pattern cannot deal with fields that contain {} or /

XingY:
#### Rationale
- Supporting using backslash to escape special characters in field or datatype used in naming expressions.
- Issue 52305: NPE when field name contains special characters

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1730
- https://github.com/LabKey/platform/pull/6387
- https://github.com/LabKey/limsModules/pull/1206

#### Changes
- NameGenerator updates:
  - remove "$S" syntax support for forward slash in data type names
  - updated parse() method to account for backslash escaped parts
  - update validateSyntax to account for backslash escaped tokens
- refactored NameGenerator.getParentNames to be used by name expression, derivation as well as sample timeline detail api for parsing parent names from a string value
- Support QueryKey.decode with backslash escaped fields
- Fix Issue 52305 NPE check


api/src/org/labkey/api/util/StringExpressionFactory.java
```diff
@@ -1080,6 +1081,8 @@ public String getValue(Map map)
                     lookupKey = _key.toString();
                 if (!map.containsKey(lookupKey))
                     lookupKey = PageFlowUtil.decode(_key.getParent() == null ? _key.getName() : _key.encode());
+                if (!map.containsKey(lookupKey))
+                    lookupKey = QueryKey.decodePart(_key.getParent() == null ? _key.getName() : _key.toString());
```
labkey-susanh: Duplicated?

api/src/org/labkey/api/util/SubstitutionFormat.java
```diff
@@ -595,7 +596,7 @@ else if (endParen > -1 &&  nameExpression.charAt(endParen-1) != '\'')
     public static List<String> validateNonFunctionalSyntax(String formatName, String nameExpression, int start, String noun, boolean allowLookup)
     {
         List<String> messages = new ArrayList<>();
-        if (start < 2 || !nameExpression.startsWith("${", start-2))
+        if (start < 2 || !(nameExpression.startsWith("${", start-2) || nameExpression.startsWith("${\\", start-3)))
```
labkey-susanh, line 599: Will produce an index out of bounds exception if start == 2

api/src/org/labkey/api/util/SubstitutionFormat.java
```diff
@@ -606,14 +607,14 @@ public static List<String> validateNonFunctionalSyntax(String formatName, String
             if (NameGenerator.SubstitutionValue.DataInputs.name().equals(formatName) || NameGenerator.SubstitutionValue.MaterialInputs.name().equals(formatName))
             {
                 // check for ancestor lookup
-                if (nameExpression.startsWith("..[", start-3))
+                if (nameExpression.startsWith("..[", start-3) || nameExpression.startsWith("..[\\", start-4))
                     return messages;
                 // check for ancestor search
-                if (nameExpression.startsWith("${~", start-3))
+                if (nameExpression.startsWith("${~", start-3) || nameExpression.startsWith("${~\\", start-4))
```
labkey-susanh, line 613: Do these need to check for a valid value for start?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -913,17 +913,18 @@ static Collection<String> getParentNames(Object parentObj, TSVWriter tsvWriter,
...
                 {
-                    parentNames = Arrays.asList(((String) parentObj).trim());
+                    parentNames = Arrays.asList(parentStr.trim());
                 }
                 else
                 {
                     // Issue 44841: The names of the parents may include commas, so we parse the set of parent names
                     // using TabLoader instead of just splitting on the comma.
-                    String quotedStr = ((String) parentObj).contains(",") ? (String) parentObj : tsvWriter.quoteValue((String) parentObj); // if value contains comma, no need to quote again
+                    boolean likelyAlreadyQuoted = parentStr.contains(",") || parentStr.contains("\n") || parentStr.contains("\r");
```
labkey-susanh: The "likely" qualifier here does not instill confidence...

api/src/org/labkey/api/exp/api/ExperimentService.java
```diff
@@ -483,7 +483,10 @@ static boolean isInputOutputColumn(String columnName)
         if (StringUtils.isBlank(columnName))
             return false;
 
-        String prefix = columnName.split("[./]")[0];
+        String[] parts = columnName.split("[./]");
+        if (parts.length == 0) // if columnName consist of only '.'
```
labkey-susanh: Is there an issue number that could be referenced here?

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: This goes hand in hand with the client side encode util...

api/src/org/labkey/api/util/StringExpressionFactory.java
XingY: Not duplicate, one is PageFlowUtil.decode, the other is QueryKey.decodePart

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -706,7 +707,8 @@ protected int write()
 
             // Issue 44841: The names of the parents may include commas, so we parse the set of parent names
             // using TabLoader instead of just splitting on the comma.
-            String quotedStr = (valueStr).contains(",") ? valueStr : tsvWriter.quoteValue(valueStr); // if value contains comma, no need to quote again
+            boolean likelyAlreadyQuoted = valueStr.contains(",") || valueStr.contains("\n") || valueStr.contains("\r");
```
cnathe: I didn't mentioned this on the client side check, but shouldn't we also be checking if the valueStr starts and ends with a quote to see if it is "already quoted"?

api/src/org/labkey/api/query/QueryKey.java
```diff
@@ -40,14 +40,46 @@ protected interface Factory<T extends QueryKey<T>>
...
+            String part = "";
+            for (String encodedPart : encodedParts)
+            {
+                part = part + (StringUtils.isEmpty(part) ? "" : divider) + encodedPart;
+                if (!part.endsWith("\\"))
+                {
+                    parts.add(part);
+                    part = "";
+                }
+            }
```
cnathe, line 78: Can we add some comments and issue number to this new section of code?

Also, what if the last part does end with "\\"? then it wouldn't get added to the `parts` list, would it?

Finally, it would be great to get junit test coverage for this decode method. Could those be added (or maybe that is already coming)?

api/src/org/labkey/api/util/SubstitutionFormat.java
```diff
@@ -595,7 +596,7 @@ else if (endParen > -1 &&  nameExpression.charAt(endParen-1) != '\'')
     public static List<String> validateNonFunctionalSyntax(String formatName, String nameExpression, int start, String noun, boolean allowLookup)
```
cnathe, line 596: should there be junit tests for these public methods?

experiment/src/org/labkey/experiment/ExpDataIterators.java
cnathe: I am seeing 3 places in the server side code that do this `likelyAlreadyQuoted` check. Could these be refactored into a helper/util somewhere? Also, as mentioned in a previous comment, should there be a "starts and ends with quote check" too?

cnathe (commented):
I've added some comments to the code changes here but I don't think I'm the right person to do a final review of these changes.

experiment/package.json
```diff
@@ -17,7 +17,7 @@
   },
   "devDependencies": {
     "@labkey/build": "8.4.0",
-    "@labkey/test": "1.7.1",
+    "@labkey/test": "1.7.2-fb-nameExpressionEscape.7",
```
cnathe: was this change just to keep this package's `@labkey/test` version in sync or is there plans to add usages of the new random name generators here in experiment?

api/src/org/labkey/api/data/NameGenerator.java
XingY: This is different from the client side usage. The client is quoting a single value, here the valueStr could contain multiple values, part of it quoted, but some value might not need quoting.

api/src/org/labkey/api/query/QueryKey.java
XingY: Comment/issue added.
Good catch with the off by 1 error, fixed the last part that ends with "\" case.
Junit is added in NameGenerator that should cover this util

api/src/org/labkey/api/util/SubstitutionFormat.java
XingY: It's actually fine. If offset < 0, it would just return false

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: I refactored the 3 usages so we use a single util for parsing parent names. As for should we check for "starts and ends with quote", the answer here is no. The client side util is for handling a single parent, but this value here is the multiple parent string. We don't know if this multi-value comes from api, tsv/csv or excel.

api/src/org/labkey/api/util/SubstitutionFormat.java
XingY: There should be coverage for this in NameGenerator, in an explicit way.

experiment/package.json
XingY: Yes, just to update it for future use.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -2929,37 +2942,37 @@ public void testNumberFormats()
         public void testStringFormats()
         {
             Map<Object, Object> m = new HashMap<>();
-            m.put("a", "A");
-            m.put("b", " B ");
-            m.put("empty", "");
+            m.put("a/\\", "A");
```
labkey-matthewb, line 3062: NameGenerator extends FieldKeyExpression but none of the test maps are keyed using FieldKey objects.

Looking at the code it looks like _GenerateNamesDataIterator uses MapDataIterator.getMap() so that sort of makes sense.  I'm wondering if it makes more sense to avoid that short cut and use an actual Map<FieldKey> instead?

labkey-matthewb (changes requested):
Since we've done this work to parse FieldKey paths now.  I think it might be worth investigating using Map<FieldKey> for .eval() in _GenerateNamesDataIterator.  That would reduce the need for relying on the fall-back matching code FieldKeyStringExpresison.

XingY:
> Since we've done this work to parse FieldKey paths now. I think it might be worth investigating using Map for .eval() in _GenerateNamesDataIterator. That would reduce the need for relying on the fall-back matching code FieldKeyStringExpresison.

This has now been implemented.

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1010,6 +1014,24 @@ private PropertyType getPropertyType(ColumnInfo col)
...
+    {
+        if (parentImportAliases == null)
+            return null;
+
+        Map<String, FieldKey> parentImportAliasFieldKeys = new CaseInsensitiveHashMap<>();
+        for (Map.Entry<String, String> aliasField : parentImportAliases.entrySet())
+        {
+            String alias = aliasField.getKey();
+            String dataType = aliasField.getValue();
+            boolean isParentSamples = dataType.toLowerCase().startsWith(MATERIAL_INPUTS_ALIAS_PREFIX.toLowerCase());
```
labkey-matthewb, line 1027: It looks odd to not just create the FieldKey here (maybe FieldKey.fromParts(split(dataType)), but I can imagine there might be a reason for this (comment?)

api/src/org/labkey/api/data/NameGenerator.java
```diff
@@ -1165,36 +1188,35 @@ else if (pt != null)
...
+                                    parentLookupFields.computeIfAbsent(FieldKey.fromParts(fieldParts.get(0)), (s) -> new ArrayList<>()).add(fieldParts.get(1));
 
                                 lookupValuePreview = getParentLookupTokenPreview(_currentDataTypeName, fkTok, fieldParts.get(0), null, ancestorPartOption, fieldParts.get(1), user, dataClassNames, sampleTypeNames);
                             }
                             else if (fieldParts.size() == 3)
                             {
                                 // MaterialInputs/SampleType/lookup, DataInputs/DataClass/lookup
                                 if (!isAncestorSearch)
-                                    parentLookupFields.computeIfAbsent(fieldParts.get(0) + "/" + fieldParts.get(1), (s) -> new ArrayList<>()).add(fieldParts.get(2));
+                                    parentLookupFields.computeIfAbsent(FieldKey.fromParts(fieldParts.get(0), fieldParts.get(1)), (s) -> new ArrayList<>()).add(fieldParts.get(2));
```
labkey-matthewb, line 1219: Hmm, util.Path has Path.subpath(), maybe we need a method like that on FieldKey?

labkey-matthewb (approved):

PR #6389, opened 2025-02-27
Issue 52268: Data Views Webpart behaves badly across time zones

labkey-jeckels:
#### Rationale
My previous fix actually made things worse.

#### Changes
- Switch to treating the date as a string in the client to avoid timezone parsing

### DIFF
--- a/query/webapp/dataview/DataViewsPanel.js
+++ b/query/webapp/dataview/DataViewsPanel.js
@@ -466,7 +466,8 @@ Ext4.define('LABKEY.ext4.DataViewsPanel', {
         }
         this.dateFormat = LABKEY.extDefaultDateFormat;
         this.dateRenderer = function(value) {
-            return LABKEY.Utils.encodeHtml(value);
+            var formattedVal = Ext4.util.Format.date(value, LABKEY.extDefaultDateFormat);
+            return LABKEY.Utils.encodeHtml(formattedVal);
         };
         this.editInfo = json.editInfo;
         this.sortOrder = json.sortOrder;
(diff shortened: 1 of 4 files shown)


labkey-jeckels:
This feels hacky but seems to work. Open to better approaches.

query/webapp/dataview/DataViewsPanel.js
```diff
@@ -466,8 +466,7 @@ Ext4.define('LABKEY.ext4.DataViewsPanel', {
         }
         this.dateFormat = LABKEY.extDefaultDateFormat;
         this.dateRenderer = function(value) {
-            var formattedVal = Ext4.util.Format.date(value, LABKEY.extDefaultDateFormat);
-            return LABKEY.Utils.encodeHtml(formattedVal);
+            return LABKEY.Utils.encodeHtml(value);
```
labkey-nicka: I would refer to the issue at hand here so that it is known why this isn't being date formatted.

labkey-nicka (approved):
I think this is fine. I scanned for other usages and all the others seem to be actual timestamps (created, modified, etc.).

query/webapp/dataview/DataViewsPanel.js
cnathe: A couple of comments:
1. this change means that we no longer will try to render this date value in the project/server date format (i.e. it will always display as YYYY-mm-dd. I suppose ideally the reports-browseDataTree.api response object would include the date value and formatted display value so we could just use that here.

2. since `dateRenderer` how just does the htmlEncoding, line 543 in this field could be changed to just do `fm.htmlEncode(values.data.refreshDate)`, I believe. And then remove the `renderDate` function at line 559.

cnathe (approved):

query/webapp/dataview/DataViewsPanel.js
labkey-jeckels: Added comment. Simplified `dateRenderer ` and removed `renderDate`. Also switched `browseDataTree` to use the configured date format and the client to match. I was thrown by the "default" in `LABKEY.extDefaultDateFormat` - I didn't realize it was the configured format.

cnathe (approved):
Changes look good. I verified the Data Cut Date for one of my study reports locally.

PR #6392, opened 2025-02-28
Deprecate ancillary study feature

labkey-adam:
#### Rationale
Redundant with Publish Study. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51882

Also, address do-nothing cancel button on Manage QC States page. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52261

### DIFF
--- a/study/src/org/labkey/study/StudyModule.java
+++ b/study/src/org/labkey/study/StudyModule.java
@@ -26,9 +26,11 @@ import org.labkey.api.admin.FolderSerializationRegistry;
 import org.labkey.api.admin.notification.NotificationService;
 import org.labkey.api.attachments.AttachmentService;
 import org.labkey.api.audit.AuditLogService;
+import org.labkey.api.collections.LabKeyCollectors;
 import org.labkey.api.data.Container;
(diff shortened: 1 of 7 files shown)


study/src/org/labkey/study/StudyModule.java
```diff
@@ -504,6 +542,40 @@ protected void startupAfterSpringConfig(ModuleContext moduleContext)
...
+                        "Support for ancillary studies will be removed in the next release of LabKey Server. This " +
+                        "server has " + StringUtilsLabKey.pluralize(count.getValue(), "ancillary study",
+                        "ancillary studies") + ": [")
+                        .append(links)
+                        .append("]. Contact your LabKey Account Manager if you have any concerns about this change.");
+                    warnings.add(builder);
+                }
+            }
+
+            private HtmlString getAncillaryStudies(MutableInt count)
```
labkey-jeckels, line 565: This is fine but I probably would have return a List<HtmlString> and do the joining in the caller. That avoids the need for the MutableInt. But this short-lived code.

labkey-jeckels (approved):

PR #6398, opened 2025-03-01
Clean up ContainerFilter

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=52404

### DIFF
--- a/api/src/org/labkey/api/data/ContainerFilter.java
+++ b/api/src/org/labkey/api/data/ContainerFilter.java
@@ -517,7 +517,7 @@ public abstract class ContainerFilter
     }
 
     // Does not validate permissions!
-    @Deprecated // Use current(Container, User) or current(ContainerUser) instead
+    @Deprecated // Use current(Container, User) or current(ContainerUser) instead. TODO: Remove
     public static @NotNull ContainerFilter current(Container c)
     {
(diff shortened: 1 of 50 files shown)


api/src/org/labkey/api/data/ContainerFilter.java
```diff
@@ -556,19 +556,15 @@ public Type getType()
         }
     };
 
-    /* TODO ContainerFilter -- Consolidate with InternalNoContainerFilter
-    /** Use this with extreme caution - it doesn't check permissions */
+    /** Use this with extreme caution - this doesn't check permissions */
     public static final ContainerFilter EVERYTHING = new InternalNoContainerFilter();
```
labkey-jeckels: We could consider renaming this `EVERYTHING_WITHOUT_CHECKING_PERMISSIONS` or `UNSAFE_EVERYTHING` to help reinforce the danger, like we do for `HtmlString` and other classes

labkey-jeckels (approved):

api/src/org/labkey/api/data/ContainerFilter.java
labkey-adam: Done

PR #6401, opened 2025-03-03
Update expected result count for "Sample" search

labkey-tchad:
#### Rationale
Previously, a search for "Sample" would include the specimen sample type ("Sample Type - Study Specimens") in its results. After a recent change (I suspect #6381), that is no longer the case.
I investigated the previous behavior (on 24.11) and found that we were indexing the sample type but none of the individual samples. Since we don't index the samples, it makes sense to not index the sample type either. Let me know if you disagree and I will open an issue.

#### Related Pull Requests
- #6381 

#### Changes
- Update expected result count for "Sample" search


labkey-susanh (approved):
Makes sense to me. @labkey-adam does it make sense to you?

labkey-adam:
> Makes sense to me. @labkey-adam does it make sense to you?

Yes, agreed. Thanks, @labkey-tchad & @labkey-susanh !

PR #6406, opened 2025-03-03
Deprecate study protocol designer tools (phase 2)

labkey-klum:
#### Rationale
This is the follow on work for this [PR](https://github.com/LabKey/platform/pull/6384). For this phase of the work, the tables used by the study protocol designer will not be shown in the `study` schema. Additionally export and import options for this feature will not be available.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6406
- https://github.com/LabKey/clientModules/pull/4

#### Changes
- Move study design table classes into the study API for now to make it easier to move into a module later.
- Introduce `StudyDesignQuerySchema` and `StudyDesignSchema` to manage the study design tables.
- Conditionalize folder writers and importers to respect the deprecated flag.
- Fix automated tests


study/src/org/labkey/study/importer/CreateChildStudyPipelineJob.java
```diff
@@ -460,7 +462,7 @@ private StudyImportContext importToDestinationStudy(BindException errors, StudyI
      */
     private void importStudyDesignData(BindException errors, VirtualFile studyDir, StudyImportContext importContext) throws Exception
     {
-        if (importContext != null)
+        if (importContext != null && OptionalFeatureService.get().isFeatureEnabled(StudyUtils.STUDY_DESIGN_FEATURE_FLAG))
```
cnathe, line 465: I believe that the `importTreatmentVisitMapData` method below should have this same check

study/src/org/labkey/study/model/StudyManager.java
```diff
@@ -2783,10 +2784,6 @@ public void deleteAllStudyData(Container c, User user)
 
             deleteStudyDesignData(c, user, studyDesignTables);
```
cnathe, line 2785: should this `deleteStudyDesignData` be moved as well?

study/src/org/labkey/study/query/StudyQuerySchema.java
```diff
@@ -222,6 +183,7 @@ protected StudyQuerySchema(SchemaKey path, String description, @Nullable StudyIm
         if (null != contextualRole && ReaderRole.class != contextualRole.getClass())
             throw new IllegalStateException("Only ReaderRole is currently supported");
         _contextualRole = contextualRole;
+        _designQuerySchema = StudyDesignQuerySchema.get(this, c, study, contextualRole);
```
cnathe, line 186: the check for `isFeatureEnabled` happens in StudyDesignQuerySchema right? so it isn't needed here

study/api-src/org/labkey/api/studydesign/query/StudyDesignBaseTable.java
```diff
@@ -0,0 +1,89 @@
...
+
+        return addColumn(containerCol);
+    }
+
+    @Override
+    public boolean hasPermission(@NotNull UserPrincipal user, @NotNull Class<? extends Permission> perm)
+    {
+        checkedPermissions.add(perm);
+        // Most tables should not be editable in Dataspace
+        if (!perm.equals(ReadPermission.class) && getContainer().isDataspace())
```
cnathe, line 59: if Atlas is the only server using study design, we probably don't really need these Dataspace related checks anymore right? probably not something that needs to change in this PR though

study/api-src/org/labkey/api/studydesign/query/StudyDesignLookupBaseTable.java
```diff
@@ -1,162 +1,185 @@
...
+
+        for (ColumnInfo col : getRealTable().getColumns())
+        {
+            BuiltInColumnTypes type = BuiltInColumnTypes.findBuiltInType(col);
+            if (type == BuiltInColumnTypes.Container)
+            {
+                BaseColumnInfo containerCol = new AliasedColumn(this, "Container", _rootTable.getColumn("Container"));
+                containerCol.setConceptURI(BuiltInColumnTypes.CONTAINERID_CONCEPT_URI);
+
+                addColumn(containerCol);
```
cnathe: it looks like `StudyDesignBaseTable. addContainerColumn` matches these lines, can that be used here?

study/api-src/org/labkey/api/studydesign/query/StudyDesignLookupBaseTable.java
```diff
@@ -1,162 +1,185 @@
...
+
+    protected Set<Role> getContextualRoles()
+    {
+        return getUserSchema().getContextualRoles();
+    }
+
+    protected boolean checkContainerPermission(UserPrincipal user, Class<? extends Permission> perm)
+    {
+        return _userSchema.getContainer().hasPermission(user, perm, getContextualRoles());
+    }*/
```
cnathe: can these commented out methods be removed? I think these are in the parent `StudyDesignBaseTable` right?

cnathe (changes requested):

study/src/org/labkey/study/model/StudyManager.java
labkey-klum: I'll probably save that for the next phase when we split it out to a separate module

study/src/org/labkey/study/query/StudyQuerySchema.java
labkey-klum: Yes, it will return `null` in that case and there should be `null` checks throughout this class.

study/api-src/org/labkey/api/studydesign/query/StudyDesignBaseTable.java
labkey-klum: Yeah, that's a good question, I found these Dataspace checks confusing so it will be nice to clean these up in the final phase.

study/api-src/org/labkey/api/studydesign/query/StudyDesignLookupBaseTable.java
labkey-klum: yes, good call.

study/api-src/org/labkey/api/studydesign/query/StudyDesignLookupBaseTable.java
labkey-klum: correct

cnathe (approved):
I pulled down the PR and did some minor manual testing to verify the study design tables show in the study schema when the flag is enabled (and do not show when disabled).

PR #6412, opened 2025-03-04
Issues 52095 and 52365: Add more tests for TabLoader; fix indexing issue for sorting in query response

labkey-susanh:
#### Rationale
- Issue [52095](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52095): It can happen that a data field in a TSV or CSV starts with a quote, but the quote is just part of the data, and there is no ending quote at the end of the field. This is difficult to distinguish from a quote that is one of a pair that surrounds the entire field value. We are choosing to treat this as a somewhat invalid construct but we won't fail the parsing because of it. Instead, the companion quote, if it exists, will be stripped out. This parsing logic caused issue 52095 to be raised. After a good amount of back and forth on this (at least in my head), we're going to leave the parsing logic as is and require that when you have quotes at the beginning of your field, you have to surround the field with quotes and then double up the quotes inside. This PR therefore only serves to fix a couple of small issues about spaces in various places and to add a few tests. 

- Issue [52365](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52365): ascending sort character, if present, will be the first character, not the second

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1217

#### Changes
- Fix indexing problem in `ApiQueryResponse`
- Account for spaces in other places when parsing tabbed data


api/src/org/labkey/api/reader/TabLoader.java
```diff
@@ -461,17 +460,16 @@ else if (ch == chQuote)
...
                         continue;
                     }
-
                     if (end == buf.length() - 1 || buf.charAt(end + 1) != chQuote)
                     {
                         // Issue 51056: pooling sample parents with single quote doesn't work
                         // " a, " b should be parsed as [" a, " b], not [a,  b]
-                        if (_parseEnclosedQuotes && end != buf.length() - 1 && buf.charAt(end + 1) != _chDelimiter)
+                        int fieldEnd = buf.indexOf(_strDelimiter, end);
+                        if (_parseEnclosedQuotes && end != buf.length() - 1 && (fieldEnd == -1 || !buf.substring(end+1, fieldEnd).matches("\\s*")))
```
XingY: Is there a new test case that matches the added `!buf.substring(end+1, fieldEnd).matches("\\s*"))` check?

XingY (approved):

api/src/org/labkey/api/reader/TabLoader.java
labkey-matthewb: I had to read this a few times, but it seem that this code path (isDelimiterOrQuote=false) and the and the 9727 fix below are in disagreement about what to do about dangling single quote in the middle of a quote-escaped string.  9727 seems to treat the remainder of a the string as a literal non-quoted string concatenated to the end of a quote-escaped string.  While this says, didn't mean it, pretend this was not an escaped string in the first place.  So using [] as superquotes.

["AB""C"xyz]
In 9727 fix the first part is treated as quote encoded string [ABC"C] concatenated with unquoted remainder [xyz]
While isDelimiterOrQuote=false strategy means that the undoubled quote undoes the quoted string handling and turns this into the literal ["AB"""xyz]

labkey-susanh: Yes, this logic was updated because `testWithQuotes` failed initially.

labkey-matthewb: I guess that's fine (GIGO).  Interestingly, it still changes tab handling within.  Given ["A\tB"xy\tz]  the first tab is not a delimiter while the second is even though we we treat it like a non-quoted string.  Can we remove the code tagged  9727?

If there are "" is it better to handle the double "" correctly, or the single " ?   Just asking this question makes me wish we had chosen to make this an error when 9727 was "fixed"

labkey-matthewb: Also,consider a static final Pattern and using Matcher (see https://github.com/LabKey/platform/pull/6249/files)

labkey-matthewb: This if does need an english translation.

labkey-matthewb (changes requested):
I would consider avoiding the hidden Pattern.compile().  I don't care that much about the decision for best-effort parsing change (disappearing single quote vs not decoding double quote).  But I think we don't need both?

labkey-susanh:
> I would consider avoiding the hidden Pattern.compile(). I don't care that much about the decision for best-effort parsing change (disappearing single quote vs not decoding double quote). But I think we don't need both?

Sorry @labkey-matthewb, but I don't know what you are suggesting here. Do you mean factor out the `Pattern.compile` on line 442 into a constant? I did wonder why `chQuote`, `_strQuoteQuote` and `_replaceDoubleQuotes` aren't all constants.  Perhaps @labkey-adam remembers from 15 years ago?

Based on our previous discussion, I'm not intending to change the rules about how users can provide quoted values, so the single quotes at the beginning of a string will continue to disappear along with whatever quotes they are paired with.

labkey-matthewb:
> > I would consider avoiding the hidden Pattern.compile(). I don't care that much about the decision for best-effort parsing change (disappearing single quote vs not decoding double quote). But I think we don't need both?
> 
> Sorry @labkey-matthewb, but I don't know what you are suggesting here. Do you mean factor out the `Pattern.compile` on line 442 into a constant? I did wonder why `chQuote`, `_strQuoteQuote` and `_replaceDoubleQuotes` aren't all constants. Perhaps @labkey-adam remembers from 15 years ago?
> 
> Based on our previous discussion, I'm not intending to change the rules about how users can provide quoted values, so the single quotes at the beginning of a string will continue to disappear along with whatever quotes they are paired with.

Sorry.  I meant buf.substring(end+1, fieldEnd).matches("\\s*"), it compiles an expression for each call.  Probably not a hot code path, but the change in that other PR made a huge difference in that case.

labkey-susanh:
> > > I would consider avoiding the hidden Pattern.compile(). I don't care that much about the decision for best-effort parsing change (disappearing single quote vs not decoding double quote). But I think we don't need both?
> > 
> > 
> > Sorry @labkey-matthewb, but I don't know what you are suggesting here. Do you mean factor out the `Pattern.compile` on line 442 into a constant? I did wonder why `chQuote`, `_strQuoteQuote` and `_replaceDoubleQuotes` aren't all constants. Perhaps @labkey-adam remembers from 15 years ago?
> > Based on our previous discussion, I'm not intending to change the rules about how users can provide quoted values, so the single quotes at the beginning of a string will continue to disappear along with whatever quotes they are paired with.
> 
> Sorry. I meant buf.substring(end+1, fieldEnd).matches("\s*"), it compiles an expression for each call. Probably not a hot code path, but the change in that other PR made a huge difference in that case.

OK, that I've done and I also made the other variables here into constants.

api/src/org/labkey/api/reader/TabLoader.java
labkey-susanh: Done

labkey-susanh: Removing the code tagged for the 9272 fix will definitely change our parsing logic in the default case (when `_parseEnclosedQuotes` is false), which I'm no longer intending to do in this PR.  If we think the logic here needs additional surgery, I would like to defer that to a different PR.

labkey-matthewb: OK, got it.  I didn't read far enough past the end of the diffs.  I see now that this just says if this quoted string isn't property terminated, just start over and treat it like it started with any regular character.  May a comment like is the English I needed.

Though I'm still having a hard time seeing how you'd get into the 9727 code path now with non-whitespace characters to append.  Why would the loop above have stopped early?

labkey-matthewb (approved):

PR #6413, opened 2025-03-04
Require non-null user and container in UserSchemas

labkey-adam:
#### Rationale
A `UserSchema` with a `null` `User` or `Container` seems less useful. Throw if either are null at construction time.

#### Related Pull Requests
- https://github.com/LabKey/commonAssays/pull/852

### DIFF
--- a/api/src/org/labkey/api/assay/AssaySchema.java
+++ b/api/src/org/labkey/api/assay/AssaySchema.java
@@ -41,12 +41,12 @@ public abstract class AssaySchema extends UserSchema
     @Nullable
     protected Container _targetStudy;
 
-    public AssaySchema(String name, User user, Container container, DbSchema dbSchema, @Nullable Container targetStudy)
+    public AssaySchema(String name, @NotNull User user, @NotNull Container container, DbSchema dbSchema, @Nullable Container targetStudy)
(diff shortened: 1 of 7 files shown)


labkey-adam:
BVTs, etc. all green except for known, unrelated failures (see above).

Daily postgres all green except for known, unrelated failures: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyPostgres_2/3407932

labkey-jeckels (approved):
Can we also annotate the variables and getters as @NotNull?

labkey-matthewb (approved):

labkey-adam:
> Can we also annotate the variables and getters as @NotNull?

I think I have the ability to do that. Of course, there are 140 user schemas, so hard to know where to stop. I chose to annotate 6 - 7 "top-level" user schemas.

labkey-jeckels:
> > Can we also annotate the variables and getters as @NotNull?
> 
> I think I have the ability to do that. Of course, there are 140 user schemas, so hard to know where to stop. I chose to annotate 6 - 7 "top-level" user schemas.

I was just thinking within this class, but nice to begin to propagate out to subclasses

PR #6414, opened 2025-03-04
Rename exp.FieldNames -> exp.Fields

labkey-adam:
#### Rationale
"Fields" is a more appropriate name and better matches the other table names in this schema. Not bothering to alias the old name since Keith doesn't think it's necessary.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6291

### DIFF
--- a/experiment/src/org/labkey/experiment/api/FieldsTable.java
+++ b/experiment/src/org/labkey/experiment/api/FieldsTable.java
@@ -11,7 +11,7 @@ public class FieldsTable extends BaseFieldsTable
 {
     public FieldsTable(@NotNull ExpSchema userSchema, @Nullable ContainerFilter containerFilter)
     {
-        super("Fields", userSchema, containerFilter);
+        super(ExpSchema.TableType.Fields.name(), userSchema, containerFilter);
         setDescription("Shows one row for each administrator-defined field in the selected folder(s). Rows are shown in " +
             "a folder or project only if the user has administrator permissions in that folder.");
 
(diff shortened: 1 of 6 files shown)


experiment/src/org/labkey/experiment/api/FieldsTable.java
```diff
@@ -7,11 +7,11 @@
...
 import org.labkey.api.exp.query.ExpSchema;
 
-public class FieldNamesTable extends BaseFieldNamesTable
+public class FieldsTable extends BaseFieldsTable
 {
-    public FieldNamesTable(@NotNull ExpSchema userSchema, @Nullable ContainerFilter containerFilter)
+    public FieldsTable(@NotNull ExpSchema userSchema, @Nullable ContainerFilter containerFilter)
     {
-        super("FieldNames", userSchema, containerFilter);
+        super("Fields", userSchema, containerFilter);
```
labkey-jeckels: Haven't tested this, but I think it should compile and work as-is.

```suggestion
        super(ExpSchema.TableType.Fields.name(), userSchema, containerFilter);
```

experiment/src/org/labkey/experiment/api/PhiFieldsTable.java
```diff
@@ -10,7 +10,7 @@
 import org.labkey.api.data.SimpleFilter;
 import org.labkey.api.exp.query.ExpSchema;
 
-public class PhiFieldsTable extends BaseFieldNamesTable
+public class PhiFieldsTable extends BaseFieldsTable
 {
     public PhiFieldsTable(@NotNull ExpSchema userSchema, @Nullable ContainerFilter containerFilter)
```
labkey-jeckels, line 15: Could do the enum trick here but it's just outside the region where GitHub lets me add a line-level comment

labkey-jeckels (approved):

PR #6418, opened 2025-03-05
Additional test coverage for LKS client exception reporting

labkey-chrisj:
#### Rationale
This change adds test coverage for client-side application exception reporting to mothership
Also adds a page class to wrap the new `ClientExceptionPage`

[spec with test plan](https://docs.google.com/document/d/1VSMNqr8On108Ial2Twt0T7uoy0_xg8GQML-WhA_Bz2U/edit?tab=t.0#heading=h.7u5qro28q0oc)
[Issue](https://github.com/LabKey/kanban/issues/532)

#### Related Pull Requests
tests in limsModules add coverage in the apps
https://github.com/LabKey/limsModules/pull/1219

#### Changes
Add method to `MothershipHelper`
New page class for `ClientExceptionPage`
Added test methods to `MothershipTest`

### DIFF
--- a/mothership/test/src/org/labkey/test/pages/mothership/ClientExceptionPage.java
+++ b/mothership/test/src/org/labkey/test/pages/mothership/ClientExceptionPage.java
@@ -1,5 +1,6 @@
 package org.labkey.test.pages.mothership;
 
+import org.labkey.remoteapi.CommandException;
 import org.labkey.test.Locator;
 import org.labkey.test.WebDriverWrapper;
 import org.labkey.test.WebTestHelper;
@@ -8,9 +9,12 @@ import org.labkey.test.util.mothership.MothershipHelper;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 
+import java.io.IOException;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Date;
+import java.util.List;
+import java.util.Map;
 
 public class ClientExceptionPage extends LabKeyPage<ClientExceptionPage.ElementCache>
 {
@@ -40,43 +44,59 @@ public class ClientExceptionPage extends LabKeyPage<ClientExceptionPage.ElementC
                 "the page did not become enabled", WAIT_FOR_JAVASCRIPT);
     }
 
-    public ClientExceptionPage clickInlineScriptError() throws Exception
+    public ClientExceptionPage clickInlineScriptError(boolean expectError) throws Exception
     {
-        waitForNewTimestamp();
+        var initialState = mothershipHelper.getOrderedStackTraces();
         elementCache().inlineScriptErrBtn.click();
+        if (expectError)
+            waitForNewTimestamp(initialState);
         return this;
     }
 
-    public ClientExceptionPage clickResourceScriptError() throws Exception
+    public ClientExceptionPage clickResourceScriptError(boolean expectError) throws Exception
     {
-        waitForNewTimestamp();
+        var initialState = mothershipHelper.getOrderedStackTraces();
         elementCache().resourceScriptErrBtn.click();
+        if (expectError)
+            waitForNewTimestamp(initialState);
         return this;
     }
 
-    public ClientExceptionPage clickNestedScriptError() throws Exception
+    public ClientExceptionPage clickNestedScriptError(boolean expectError) throws Exception
     {
-        waitForNewTimestamp();
+        var initialState = mothershipHelper.getOrderedStackTraces();
         elementCache().nestedScriptErrBtn.click();
+        if (expectError)
+            waitForNewTimestamp(initialState);
         return this;
     }
 
-    public ClientExceptionPage clickAsyncScriptError() throws Exception
+    public ClientExceptionPage clickAsyncScriptError(boolean expectError) throws Exception
     {
-        waitForNewTimestamp();
+        var initialState = mothershipHelper.getOrderedStackTraces();
         elementCache().asyncScriptErrBtn.click();
+        if (expectError)
+            waitForNewTimestamp(initialState);
         return this;
     }
 
-
-    private void waitForNewTimestamp() throws Exception
+    private void waitForNewTimestamp(List<Map<String, Object>> initialState) throws IOException, CommandException
     {
-        var stackTraces = mothershipHelper.getOrderedStackTraces();
-        if (stackTraces.isEmpty())
-            return;
-
-        var lastTimestamp = (Date)stackTraces.get(0).get("LastReport");
-        waitFor(()-> Instant.now().minus(Duration.ofSeconds(1)).isAfter(lastTimestamp.toInstant()), 2000);
+        waitFor(()-> {
+            try
+            {
+                sleep(250);
+                var latestTraces =  mothershipHelper.getOrderedStackTraces();
+                var lastTimestamp = (Date)initialState.get(0).get("LastReport");
+                if (initialState.isEmpty())
+                    return !latestTraces.isEmpty();
+                else
+                    return  !latestTraces.isEmpty() && ((Date)latestTraces.get(0).get("LastReport")).after(lastTimestamp);
+            } catch (Exception e)
+            {
+                return false;
+            }
+        }, "No new report appeared in Mothership", 2000);
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


mothership/test/src/org/labkey/test/tests/mothership/MothershipTest.java
```diff
@@ -281,6 +285,143 @@ public void testNoErrorCodeWithReportingDisabled()
...
+        checker().verifyEquals("Unexpected exception message",
+                "TypeError: x is undefined", rowMap.get("ExceptionMessage"));
+        checker().verifyEquals("Expect confirmation that the user is correct",
+                getCurrentUser(), rowMap.get("Username"));
+        checker().verifyEquals("expext the referring URL to be the page from which we caused this",
+                url.toString(), rowMap.get("ReferrerURL"));
+        checker().verifyEquals("Expect instance count to be the same in the page",
+                testException.get("Instances"), stackTraceDetailPage.getExceptionReports().getDataRowCount());
+        checker().screenShotIfNewError("unexpected_record_data");
+        resetErrors();
```
labkey-nicka: I'd recommend calling `resetErrors()` after each test case in an `@After` method so they get reset regardless of the outcome of the test.

mothership/test/src/org/labkey/test/tests/mothership/MothershipTest.java
```diff
@@ -281,6 +285,143 @@ public void testNoErrorCodeWithReportingDisabled()
...
+    }
+
+    @Test
+    public void testAsyncScriptClientError() throws Exception
+    {
+        var initialStackTraces = _mothershipHelper.getOrderedStackTraces();
+
+        var exceptionPage = ClientExceptionPage.beginAt(this);
+        exceptionPage.clickAsyncScriptError(true);
+        sleep(1000); // give it a chance to happen
```
labkey-nicka: What are you giving a chance to happen? The reporting of the error to the server? Perhaps this should be codified into the taking of these actions lke `clickAsyncScriptError()`?

mothership/test/src/org/labkey/test/util/mothership/MothershipHelper.java
```diff
@@ -166,6 +166,22 @@ public int getLatestStackTraceId()
         }
     }
 
+    public List<Map<String, Object>> getOrderedStackTraces()
```
labkey-nicka: Rather than returning an unbounded collection of rows it seems like this should do the same thing as `getLatestStackTraceId()` but return the whole row instead of just the value for the `ID_COLUMN`.

I'd recommend replacing this function with a new `getLatestStackTrace()`, use the implementation of `getLatestStackTraceId()` including the maxRows setting and then have `getLatestStackTraceId()` use it.

```java
public @Nullable Map<String, Object> getLatestStackTrace()
{
    Connection connection = createDefaultConnection();
    SelectRowsCommand command = new SelectRowsCommand("mothership", "ExceptionStackTrace");
    command.addSort("LastReport", Sort.Direction.DESCENDING);
    command.setMaxRows(1);
    try
    {
        SelectRowsResponse response = command.execute(connection, MOTHERSHIP_PROJECT);
        if (response.getRows().isEmpty())
            return null;
        return response.getRows().get(0);
    }
    catch (IOException|CommandException e)
    {
        throw new RuntimeException(e);
    }
}

// Replaces current implementation of getLatestStackTraceId()
public int getLatestStackTraceId()
{
    Map<String, Object> lastStackTrace = getLatestStackTrace();
    if (lastStackTrace == null)
        return 0;

    return (Integer) lastStackTrace.get(ID_COLUMN);
}
```

mothership/test/src/org/labkey/test/tests/mothership/MothershipTest.java
labkey-chrisj: Whoops, left that sleep in from before refactoring `clickAsyncScriptError `to wait for `mothershipHelper `to register a new error instance.  Thanks for calling it out, I'll remove

mothership/test/src/org/labkey/test/util/mothership/MothershipHelper.java
labkey-chrisj: Ok, doing that now.
This will require some test refactoring

mothership/test/src/org/labkey/test/pages/mothership/ClientExceptionPage.java
```diff
@@ -0,0 +1,123 @@
...
+    {
+        return new ElementCache();
+    }
+
+    @Override
+    protected ElementCache elementCache()
+    {
+        return (ElementCache) super.elementCache();
+    }
+
```
labkey-tchad: This shouldn't be necessary. It's only necessary for classes that don't directly extend a class with an ElementCache type parameter (e.g. `extends LabKeyPage<ClientExceptionPage.ElementCache>`).
If you're unsure, you can check whether it's necessary by deleting the override; if references to things inside the cache resolve you're good (e.g. `elementCache().asyncScriptErrBtn`).

```suggestion
```

mothership/test/src/org/labkey/test/tests/mothership/MothershipTest.java
labkey-tchad: It's preferable to verify that the expected error occurred, rather than resetting blindly.
```
checkExpectedErrors(1);
```

mothership/test/src/org/labkey/test/tests/mothership/MothershipTest.java
```diff
@@ -281,6 +285,142 @@ public void testNoErrorCodeWithReportingDisabled()
...
+        checker().verifyEquals("Unexpected exception message",
+                "TypeError: x is undefined", rowMap.get("ExceptionMessage"));
+        checker().verifyEquals("Expect confirmation that the user is correct",
+                getCurrentUser(), rowMap.get("Username"));
+        checker().verifyEquals("expext the referring URL to be the page from which we caused this",
+                url.toString(), rowMap.get("ReferrerURL"));
+        checker().verifyEquals("Expect instance count to be the same in the page",
+                testException.get("Instances"), stackTraceDetailPage.getExceptionReports().getDataRowCount());
+        checker().screenShotIfNewError("unexpected_record_data");
+        resetErrors();
```
labkey-tchad: I don't think this is necessary. The `checkExpectedErrors` above will have already reset the error mark.

mothership/test/src/org/labkey/test/tests/mothership/MothershipTest.java
```diff
@@ -281,6 +285,142 @@ public void testNoErrorCodeWithReportingDisabled()
...
+                getCurrentUser(), rowMap.get("Username"));
+        checker().verifyEquals("expect the referring URL to be the page from which we caused this",
+                url.toString(), rowMap.get("ReferrerURL"));
+        checker().wrapAssertion(()-> Assertions.assertThat(rowMap.get("URL"))
+                .as("expect url to be some line in deepException.js")
+                .contains("mothership/deepException.js?"));
+        checker().verifyEquals("Expect instance count to be the same in the page",
+                testException.get("Instances"), stackTraceDetailPage.getExceptionReports().getDataRowCount());
+        checker().screenShotIfNewError("unexpected_record_data");
+        resetErrors();
```
labkey-tchad:
```suggestion
        checkExpectedErrors(1);
```

mothership/test/src/org/labkey/test/tests/mothership/MothershipTest.java
```diff
@@ -281,6 +285,142 @@ public void testNoErrorCodeWithReportingDisabled()
...
+        checker().verifyEquals("Unexpected exception message",
+                "TypeError: x is undefined", rowMap.get("ExceptionMessage"));
+        checker().verifyEquals("Expect confirmation that the user is correct",
+                getCurrentUser(), rowMap.get("Username"));
+        checker().verifyEquals("expext the referring URL to be the page from which we caused this",
+                url.toString(), rowMap.get("URL"));
+        checker().verifyEquals("Expect instance count to be the same in the page",
+                testException.get("Instances"), stackTraceDetailPage.getExceptionReports().getDataRowCount());
+        checker().screenShotIfNewError("unexpected_record_data");
+        resetErrors();
```
labkey-tchad:
```suggestion
        checkExpectedErrors(1);
```

labkey-tchad (approved):
Just some nits

PR #6393, opened 2025-02-28
Remove the ability to select specific objects to import and apply import to multiple folders under the advanced import options

labkey-bpatel:
#### Rationale
Issue [49544](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49544): Deprecate the ability to select specific objects to import and apply import to multiple folders under the advanced import options

#### Related Pull Requests
- https://github.com/LabKey/premiumModules/pull/183
- https://github.com/LabKey/testAutomation/pull/2306

#### Changes
- Remove 'Advanced Import Options' and 'Import to multiple folders' option during Folder import

PR #6419, opened 2025-03-05
Issue 49544: Deprecate the ability to select specific objects to import and apply import to multiple folders under the advanced import options

labkey-bpatel:
#### Rationale
Issue [49544](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49544): Deprecate the ability to select specific objects to import and apply import to multiple folders under the advanced import options

Keep 'Advanced Import Options' behind deprecated flag until 25.7, then we will remove it. I had already opened a [PR](https://github.com/LabKey/platform/pull/6393) where I had removed this option, can use it for reference.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2326

#### Changes
- Move 'Advanced Import Options' from admin-importFolder.view behind 'Deprecated' flag.
- Move 'Show advanced import options' from pipeline-startFolderImport.view behind 'Deprecated' flag.
- Display 'Create shared datasets' and 'Fail import for undefined visits' with tooltips on admin-importFolder.view page.


labkey-klum (changes requested):
These changes look good but there is also the check box for advanced options that is available from the folder settings import tab that should be put behind the flag:

image

labkey-bpatel:
Hi @labkey-klum - The reason I did not put that checkbox behind the flag since checking it displays below page, where there are two more options here 'Validate Queries' and 'Study Import Options' - should those be put behind the flag as well? Also, Validate Queries seems redundant here, since its already present in the previous step.

image

core/src/org/labkey/core/admin/importFolder.jsp
```diff
@@ -106,6 +107,44 @@
     .lk-import-folder label {
         font-weight: normal;
     }
+    span.labkey-help-pop-up {
```
cnathe: it seems odd to have to include these styles for the help tip. I noticed that the "compliance" tab has some help tips. Take a look at the `folderSettings.jsp` usages in the premiumModules/compliance module. Maybe you could add the help tips here in that same way for consistency.

core/src/org/labkey/core/admin/importFolder.jsp
```diff
@@ -141,24 +180,54 @@
...
+            </label>
         </td>
     </tr>
 <%
     }
 %>
     <tr>
         <td style="padding-left: 15px; padding-top: 5px;">
-            <label><input type="checkbox" name="validateQueries" <%=h(form.isValidateQueries() ? "checked" : "")%> value="true"> Validate all queries after <%=h(action.toLowerCase())%></label>
+            <label><input type="checkbox" data-title="test" name="validateQueries" <%=h(form.isValidateQueries() ? "checked" : "")%> value="true"> Validate all queries after <%=h(action.toLowerCase())%>
```
cnathe: what is this `data-title="test"` for? was that just a placeholder for something?

pipeline/webapp/pipeline/ImportAdvancedOptions.js
```diff
@@ -316,19 +293,19 @@ Ext4.define('LABKEY.import.SpecificImportOptions', {
...
 
-            // don't update checked state of the disabled checkbox inputs
-            if (!cb.disabled)
-                cb.checked = checked;
-        }, this);
+                // don't update checked state of the disabled checkbox inputs
+                if (!cb.disabled)
+                    cb.checked = checked;
+            }, this);
+        }
```
cnathe: not sure which change caused this, but when I check the "Select specific objects to import" option on the advanced import page, the section for the individual options doesn't show anymore. do you see that as well locally?

cnathe (changes requested):

core/src/org/labkey/core/admin/importFolder.jsp
labkey-klum: I agree, we don't want to override these themed classes. @labkey-bpatel is there a reason you can't use the help popup helpers in `JspBase`?

labkey-bpatel: Searching for examples without Ext4 did not result in what I was looking for (search criteria/keywords were off it looks like), so ended up creating my own. But this is good to know, thank you!

labkey-klum: I think for non-Ext `JspBase.helpPopup` is the normal way, for Ext you use the `helpText` attribute and we have overrides to inject the popup.

pipeline/webapp/pipeline/ImportAdvancedOptions.js
labkey-bpatel: This is now fixed, it was a regression given my latest changes.

core/src/org/labkey/core/admin/importFolder.jsp
labkey-bpatel: now removed

core/src/org/labkey/core/admin/importFolder.jsp
labkey-bpatel: Now using JspBase.helpPopup - I had to use some styling otherwise the long text gets truncated, so ended up using HtmlString.unsafe() like I noticed in other places.

core/src/org/labkey/core/admin/importFolder.jsp
```diff
@@ -42,10 +44,26 @@
...
+            "cause the import job to raise an error. To suppress this validation step, uncheck this box") + "";
+    StringBuilder validateQueriesTooltipStrBldr = new StringBuilder();
+    validateQueriesTooltipStrBldr.append("<span style=\"display: inline-block;width:300px;\">").append(validateQueriesHelpText).append("</span>");
+
+    String failForUndefinedVisitsHelpText = h("By default, new visit rows will be created in the study during import for any dataset or specimen rows " +
+            "which have a new, undefined visit. If, instead, you would like for the import of the folder archive to fail when it encounters a visit that " +
+            "is not already defined in the study or as part of the incoming visit map, check this box.") + "";
+    StringBuilder failForUndefinedVisitsTooltipStrBldr = new StringBuilder();
+    failForUndefinedVisitsTooltipStrBldr.append("<span style=\"display: inline-block;width:300px;\">").append(failForUndefinedVisitsHelpText).append("</span>");
+
```
labkey-klum, line 64: This seems fine but using the `DOM` builder classes is a more efficient way to generate server side HTML and ensures proper encoding. This way you don't need to use `unsafe` anywhere an `HtmlString` is expected.
```suggestion

    final String sharedDatasetsHelpText = "By default, datasets will be created in this container. For Dataspace projects, shared " +
            "datasets are created at the project level so that they can be used by each of the study folders in the project.";
    HtmlString sharedDatasetsTooltip = DOM.createHtml(SPAN(DOM.at(style, "display: inline-block;width:300px;"), sharedDatasetsHelpText));

    final String validateQueriesHelpText = "By default, queries will be validated upon import of a folder archive and any failure to validate will " +
            "cause the import job to raise an error. To suppress this validation step, uncheck this box";
    HtmlString validateQueriesTooltipStr = DOM.createHtml(SPAN(DOM.at(style, "display: inline-block;width:300px;"), validateQueriesHelpText));

    final String failForUndefinedVisitsHelpText = "By default, new visit rows will be created in the study during import for any dataset or specimen rows " +
            "which have a new, undefined visit. If, instead, you would like for the import of the folder archive to fail when it encounters a visit that " +
            "is not already defined in the study or as part of the incoming visit map, check this box.";
    HtmlString failForUndefinedVisitsTooltipStr = DOM.createHtml(SPAN(DOM.at(style, "display: inline-block;width:300px;"), failForUndefinedVisitsHelpText));


```

labkey-klum: Also, don't commit this suggestion without the corresponding `helpPopup` changes.

labkey-klum (approved):

core/src/org/labkey/core/admin/importFolder.jsp
labkey-adam: Yes, please no more String-based HTML generation.

Same idea, but even simpler: don't use DOM.createHtml() or HtmlString... just hold onto `Renderable` or `SPAN`:
```
Renderable failForUndefinedVisitsTooltipStr = SPAN(DOM.at(style, "display: inline-block;width:300px;"), failForUndefinedVisitsHelpText);
```
...and pass Renderables to the help popup. You'll need to adjust the JspBase.helpPopup() variants that take HtmlString to take Renderable instead, but that's a fine change. Or just use PageFlowUtil.popupHelp() if you're not comfortable with that change.

labkey-bpatel: This is great to know! Thank you both. Now using the DOM Builder to generate server side html, and also have added (and using) a variant of JspBase.helpPopup() that takes Renderable.

core/src/org/labkey/core/admin/importFolder.jsp
```diff
@@ -42,10 +46,23 @@
     Container c = getViewContext().getContainerNoTab();
     Container project = c.getProject();
     String requestOrigin = (request.getParameter("origin") != null) ? request.getParameter("origin") : "here";
-    boolean isStudyRequest = requestOrigin.equals("Study") || requestOrigin.equals("Reload");
     boolean canCreateSharedDatasets = false;
+    boolean isAdvancedImportOptionEnabled = OptionalFeatureService.get().isFeatureEnabled("advancedImportFlag");
 
-    if (isStudyRequest && !c.isProject() && null != project && project != c)
+    String sharedDatasetsHelpText = h("By default, datasets will be created in this container. For Dataspace projects, shared " +
+            "datasets are created at the project level so that they can be used by each of the study folders in the project.") + "";
```
labkey-adam: No need to use `h()` and then `toString()` it... DOM will HTML encode all Strings, so you're now double encoding. You could keep these all `HtmlString`s, but there's no need... just pass them in as Strings and DOM will do the right thing.

api/src/org/labkey/api/jsp/JspBase.java
```diff
@@ -518,6 +518,11 @@ public void include(ModelAndView view, Writer writer) throws Exception
         HttpView.currentView().include(view, writer);
     }
 
+    public HelpPopupBuilder helpPopup(Renderable helpText, String titleText)
+    {
+        return PageFlowUtil.popupHelp(helpText, titleText);
+    }
+
```
labkey-adam: Let's not add a new method... just change the two method variants that take HtmlString to take Renderable instead. HtmlString implements Renderable, so this is a compatible change.

core/src/org/labkey/core/admin/importFolder.jsp
labkey-adam: (You shouldn't ever need to call HtmlString.toString()...)

labkey-adam (changes requested):
See comments about not adding a JspBase method and not double encoding the help text

core/src/org/labkey/core/admin/importFolder.jsp
labkey-bpatel: Yup, I noticed DOM indeed HTML encodes it by eventually using PageFlowUtil's filter(), but missed to remove it. Now done!

api/src/org/labkey/api/jsp/JspBase.java
labkey-bpatel: The two method variants are now updated

labkey-adam (approved):

cnathe (approved):

pipeline/webapp/pipeline/ImportAdvancedOptions.js
cnathe: Verified. Thanks.

PR #6429, opened 2025-03-11
Issue 52467: query for fewer columns when indexing exp.data rows

labkey-nicka:
#### Rationale
We are querying for a large number of columns and in particular complex columns (MVFKs, ancestors, etc.) when indexing individual exp.data rows. Subsequently, we're dropping many of these columns results and not actually including them in the index and/or we don't need to be. See [Issue 52467](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52467).

#### Changes
- Skip `lsidtype` and `entityid` type columns in the query as their results are already ignored.
- Change how we skip columns to avoid try/catch.
- remove redundant indexing when comments are set on exp.data. Move to controller action.

### DIFF
--- a/api/src/org/labkey/api/exp/api/ExpObject.java
+++ b/api/src/org/labkey/api/exp/api/ExpObject.java
@@ -63,6 +63,10 @@ public interface ExpObject extends Identifiable, Comparable<ExpObject>
     String getComment();
     /** Stored in ontology manager */
     void setComment(User user, String comment) throws ValidationException;
+    default void setComment(User user, String comment, boolean index) throws ValidationException
+    {
+        setComment(user, comment);
+    }
     
     User getCreatedBy();
     Date getCreated();
--- a/api/src/org/labkey/api/exp/api/ExpObject.java
+++ b/api/src/org/labkey/api/exp/api/ExpObject.java
@@ -63,7 +63,9 @@ public interface ExpObject extends Identifiable, Comparable<ExpObject>
     String getComment();
     /** Stored in ontology manager */
     void setComment(User user, String comment) throws ValidationException;
+    /** @param index only a suggestion as not all implementations handle indexing - primarily used to suppress duplicate indexing with a false value */
     default void setComment(User user, String comment, boolean index) throws ValidationException
+
     {
         setComment(user, comment);
     }
--- a/experiment/src/org/labkey/experiment/api/ExpDataImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpDataImpl.java
@@ -58,6 +58,7 @@ import org.labkey.api.pipeline.PipelineService;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.QueryRowReference;
 import org.labkey.api.query.QueryService;
+import org.labkey.api.query.ValidationException;
 import org.labkey.api.search.SearchResultTemplate;
 import org.labkey.api.search.SearchScope;
 import org.labkey.api.search.SearchService;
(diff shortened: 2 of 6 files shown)


experiment/src/org/labkey/experiment/api/ExpDataImpl.java
```diff
@@ -498,31 +491,34 @@ private void getIndexValues(
         JSONObject jsonData
     )
     {
+        CaseInsensitiveHashSet skipColumns = new CaseInsensitiveHashSet();
+        for (ExpDataClassDataTable.Column column : ExpDataClassDataTable.Column.values())
+            skipColumns.add(column.name());
+        skipColumns.add("Ancestors");
+        skipColumns.add("Container");
```
labkey-jeckels, line 514: This is likely redundant with the check for `entityid` below.

experiment/src/org/labkey/experiment/api/ExpDataImpl.java
```diff
@@ -498,31 +491,34 @@ private void getIndexValues(
...
-            {
-                // ok
-            }
 
             // skip non-text columns or columns that aren't lookups
             if (!(col.getJdbcType().isText() || col.getFk() != null))
                 return false;
 
-            if (name.equalsIgnoreCase("container") || name.equalsIgnoreCase("folder"))
+            if ("lsidtype".equalsIgnoreCase(col.getSqlTypeName()) || "entityid".equalsIgnoreCase(col.getSqlTypeName()))
```
labkey-jeckels, line 527: Aspirational: Maybe someday we can have constants for these data types, as they're referenced in a number of other places in Java code

labkey-jeckels, line 527:
```suggestion
            // skip indexing both the raw columns like LSID as well as the wrapped versions of those columns that are lookups to other data (Issue 52467)
            if ("lsidtype".equalsIgnoreCase(col.getSqlTypeName()) || "entityid".equalsIgnoreCase(col.getSqlTypeName()))

```

experiment/src/org/labkey/experiment/api/ExpDataImpl.java
```diff
@@ -176,13 +176,6 @@ public ExpDataImpl(Data data)
         super(data);
     }
 
-    @Override
-    public void setComment(User user, String comment) throws ValidationException
```
labkey-jeckels, line 180: There are a bunch of other places that call `ExpData.setComment()` and I'm not sure we want them to skip updating the index. Maybe we should overload the method and let callers opt in or out?

labkey-jeckels (commented):
See comment about `setComment()`

experiment/src/org/labkey/experiment/api/ExpDataImpl.java
labkey-nicka: Sure, I can do that. I considered it at first but decided against it since exp.data are the only experiment object that indexes when you modify the comment. Seems like this should be something that is done or not done uniformly. I'll put down my aspirations now.

experiment/src/org/labkey/experiment/api/ExpDataImpl.java
labkey-nicka: Agreed. Not going to take action on this in this PR.

experiment/src/org/labkey/experiment/api/ExpDataImpl.java
labkey-jeckels: Oh, interesting. I didn't check the other implementations but it makes sense that exp.data was the oddball given the override.

experiment/src/org/labkey/experiment/api/ExpDataImpl.java
labkey-nicka: Done.

api/src/org/labkey/api/exp/api/ExpObject.java
```diff
@@ -63,6 +63,10 @@ public interface ExpObject extends Identifiable, Comparable<ExpObject>
     String getComment();
     /** Stored in ontology manager */
     void setComment(User user, String comment) throws ValidationException;
+    default void setComment(User user, String comment, boolean index) throws ValidationException
```
labkey-jeckels, line 67:
```suggestion
    /** @param index only a suggestion as not all implementations handle indexing - primarily used to suppress duplicate indexing with a false value */
    default void setComment(User user, String comment, boolean index) throws ValidationException

```

labkey-jeckels (approved):

PR #6432, opened 2025-03-12
Better support for spaces and other characters in replicate measure names

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52486

The handling of assay plate replicate wells in the `update` case will issue a query to get the replicate well rows for any changed replicate `LSIDs`. The referenced issue will occur if the column alias differs from the name, in this case we want to make sure there is an entry in the row map for the measure column name since this is what is expected in the code which handles inserting or updating the replicate table.

In addition to the fix here, there is also new test code to validate this behavior.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6432
- https://github.com/LabKey/limsModules/pull/1233


cnathe (approved):
Change looks good. FB pulled down locally and verified.

PR #6438, opened 2025-03-13
Issue 52520: Prevent guests from submitting client-side mothership reports

labkey-nicka:
#### Rationale
This addresses [Issue 52520](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52520) by taking action to restrict guest user reporting on both the client and server.

#### Changes
- Mark AdminController.LogClientExceptionAction as `@RequiresLogin`
- Prevent guest users from submitting reports via mothership.js.

### DIFF
--- a/core/webapp/mothership.js
+++ b/core/webapp/mothership.js
@@ -268,27 +268,22 @@ LABKEY.Mothership = (function () {
         }
 
         // CONSIDER: Remove the _mothership flag and use a site-setting to suppress mothership reporting of client-side exceptions.
-        var url;
-        var o;
-        if (_mothership || (LABKEY.user && LABKEY.user.isGuest)) {
-            url = LABKEY.contextPath + '/admin/logClientException.api';
-            o = {
-                username: LABKEY.user ? LABKEY.user.email : "Unknown",
-                //site: window.location.host,
-                //version: LABKEY.versionString || "Unknown",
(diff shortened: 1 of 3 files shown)


core/webapp/mothership.js
```diff
@@ -268,27 +268,22 @@ LABKEY.Mothership = (function () {
...
-                exceptionMessage: msg,
-                stackTrace: stackTrace || msg,
-                file: file,
-                line: line,
-                // column: column,
-                browser: navigator && navigator.userAgent || "Unknown",
-                platform:  navigator && navigator.platform  || "Unknown"
-            };
-
+        if (_mothership) {
```
labkey-jeckels, line 271: Fine to leave for now, but will anything set this to false at this point?

labkey-jeckels (approved):

core/webapp/mothership.js
labkey-nicka: No, I don't believe so. This can be set externally by a call to `LABKEY.Mothership.setReportErrors()` which is why I didn't yank it. That said, I don't think anyone calls that method. I can go further and remove it but there is a kind of a lot of cruft in this file in general.

labkey-jeckels: Definitely. Like I said, fine to leave as-is for 25.3 purposes.

PR #6442, opened 2025-03-14
Delete actions fail due to the IndexWriter closing

Sigmonia:
#### Rationale
[Secure Issue 51952](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51952): Delete actions fail to the indexWriter closing

#### Changes
- Move delete event listeners to Commit task
- Adjust AlreadyClosedException handling

### DIFF
--- a/api/src/org/labkey/api/exp/api/ExperimentListener.java
+++ b/api/src/org/labkey/api/exp/api/ExperimentListener.java
@@ -16,7 +16,6 @@
 package org.labkey.api.exp.api;
 
 import org.labkey.api.data.Container;
-import org.labkey.api.data.DbScope.CommitTaskOption;
 import org.labkey.api.query.BatchValidationException;
 import org.labkey.api.security.User;
 
@@ -24,14 +23,8 @@ import java.util.List;
 
 public interface ExperimentListener
 {
-    /** Called after an experiment is deleted */
+    /** Called after an experiment is deleted (in-transaction). */
     default void afterExperimentDeleted(Container c, User user, ExpExperiment experiment) { }
-
-    /** Allows the listener to determine when the action should be executed relative to the transaction */
-    default CommitTaskOption afterExperimentDeleteCommitOption()
-    {
-        return CommitTaskOption.IMMEDIATE;
-    }
 
     /** Called after an experiment is saved (post-transaction). */
     default void afterExperimentSaved(Container c, User user, ExpExperiment experiment) { }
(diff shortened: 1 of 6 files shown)


Sigmonia:
Should this be pulled to 25.3 instead of develop?

search/src/org/labkey/search/model/WritableIndexManagerImpl.java
```diff
@@ -194,7 +196,10 @@ public void deleteDocuments(Collection<String> ids)
                 throw new ConfigurationException("Unable to write to search index, Disk is full", e);
             }
             else
-                throw e;  //TODO Should we just ignore this --  it is typically caused by either the server shutting down or the index being deleted.
+            {
+                _log.error("Indexing error deleting {} indexer is already closed", StringUtils.trimToEmpty(currentId), e);
+                SimpleMetricsService.get().increment(SearchModule.NAME,"Delete", "IndexAlreadyClosed");
```
labkey-jeckels, line 201: Good idea for the metric

api/src/org/labkey/api/exp/api/ExperimentListener.java
```diff
@@ -16,16 +16,23 @@
...
+import org.labkey.api.data.DbScope.CommitTaskOption;
 import org.labkey.api.query.BatchValidationException;
 import org.labkey.api.security.User;
 
 import java.util.List;
 
 public interface ExperimentListener
 {
-    /** Called after an experiment is deleted (in-transaction). */
+    /** Called after an experiment is deleted */
```
labkey-jeckels: This approach looks it would work, and I'm only seeing the one implementation/override for this. That's `AssayExperimentListener`, which you updated accordingly.

However, it seems inconsistent to make these two methods work differently from all the others in this interface. I'd recommend keeping them as-is, but making `AssayExperimentListener` queue up a `CommitTask` on its own. What do you think?

labkey-jeckels (commented):
See comment about transaction handling for ExperimentListener. And yes, I think a 25.3 patch would be good.

api/src/org/labkey/api/exp/api/ExperimentListener.java
Sigmonia: That's how I approached it initially, but it was a little weird to me to create the needed transaction in the listener and we don't generally pass them around, so this seemed less "smelly."

labkey-jeckels: Why would the listener need to create a transaction? Couldn't it (or AssayManager) just poke in a CommitTask, similar to https://github.com/LabKey/platform/blob/81a698b556a5eda39d30a9409dc6645c5ce01d77/experiment/src/org/labkey/experiment/ExpDataIterators.java#L2179

Sigmonia:
Closing in favor of https://github.com/LabKey/platform/pull/6450 targeting 25.3

api/src/org/labkey/api/exp/api/ExperimentListener.java
Sigmonia: Made the change.

PR #6446, opened 2025-03-17
Fb file constructor mitigation

labkey-klum:
#### Rationale
Removes the `File` constructor usage from the `ReportsController.GetBackgroundReportResultsAction` endpoint. 

Added a little more extensive refactor to `ScriptEngineReport.getReportDir` to create a variant that returns a `FileLike` object and convert most (but not all) prior usages to use the new version.

### DIFF
--- a/api/src/org/labkey/api/reports/report/ScriptEngineReport.java
+++ b/api/src/org/labkey/api/reports/report/ScriptEngineReport.java
@@ -58,11 +58,14 @@ import org.labkey.api.reports.report.r.view.TextOutput;
 import org.labkey.api.reports.report.r.view.TsvOutput;
 import org.labkey.api.thumbnail.Thumbnail;
 import org.labkey.api.util.FileUtil;
+import org.labkey.api.util.Path;
 import org.labkey.api.util.UnexpectedException;
 import org.labkey.api.view.HttpView;
 import org.labkey.api.view.VBox;
 import org.labkey.api.view.ViewContext;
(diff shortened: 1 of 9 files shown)


api/src/org/labkey/api/reports/report/ScriptEngineReport.java
```diff
@@ -247,21 +253,66 @@ protected File getReportDir(@NotNull String executingContainerId, boolean isPipe
...
+                            Path.parse(executingContainerId + File.separator + "Report_" + reportId + File.separator + identifier));
+                else
+                    tempFolder = FileUtil.appendPath(tempRoot,
+                            Path.parse(executingContainerId + File.separator + "Report_" + reportId));
+            }
+            else
+                tempFolder = FileUtil.appendPath(tempRoot,
+                        Path.parse(executingContainerId + File.separator + "Report_" + reportId + File.separator + Thread.currentThread().getId()));
+
+            FileUtil.mkdirs(tempFolder);
```
labkey-matthewb, line 291: We should also migrate to using FileUtil.createTempDirectory() (or we can add a random value to this).  It's now considered risky to put things in the shared temp directory with somewhat predictable names.

labkey-klum: I think that would be good for a follow on release, the reporting framework relies on knowing a predictable working directory from just the report ID (and request or pipeline job).

labkey-matthewb (approved):

PR #6451, opened 2025-03-17
Add a long running action for testing purposes.

Sigmonia:
#### Rationale
During our Cloud meeting a while ago we discussed adding an api Action that would allow testing long running requests 
[#51619](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51619)

#### Changes
- Added actions to the DevModule/TestController

### DIFF
--- a/devtools/src/org/labkey/devtools/TestController.java
+++ b/devtools/src/org/labkey/devtools/TestController.java
@@ -213,7 +213,6 @@ public class TestController extends SpringActionController
         }
     }
 
-
     public static class LongRunningJobForm
     {
         private int maxTimeMS;
@@ -267,6 +266,8 @@ public class TestController extends SpringActionController
         {
             if (form.stepTimeMS <= 0)
                 errors.rejectValue("stepTimeMS", "Step time must be greater than zero");
+            if (form.stepTimeMS > form.maxTimeMS)
+                errors.rejectValue("stepTimeMS", "Step time must be less than or equal to Max Time.");
             if (form.maxTimeMS <= 0)
                 errors.rejectValue("maxTimeMS", "Max time must be greater than zero");
         }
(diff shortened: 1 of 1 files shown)


devtools/src/org/labkey/devtools/TestController.java
```diff
@@ -178,6 +179,153 @@ public void addNavTrail(NavTree root)
...
+
+        @Override
+        public void run() {
+            try {
+                int loops = 0;
+                int remaining = maxTime;
+                logger.info("Long runner starting...");
+                while (!Thread.currentThread().isInterrupted() && remaining > 0) {
+                    logger.debug("Long runner working... loop [{}], remaining [{}s]", loops, remaining/1000);
+                    Thread.sleep(stepTime); // Let the thread run for a while
```
labkey-jeckels, line 205: Per the comment, sleeping == running?

labkey-jeckels, line 205: May want to check for `stepTime > 0` or the crawler might cause exceptions here.

devtools/src/org/labkey/devtools/TestController.java
```diff
@@ -178,6 +179,153 @@ public void addNavTrail(NavTree root)
...
+        }
+
+        public int getStepTimeMS()
+        {
+            return stepTimeMS;
+        }
+    }
+
+    @RequiresSiteAdmin
+    public class CancelLongRunningAction extends FormViewAction<Object>
```
labkey-jeckels: Could both of these actions subclass `ApiAction` or something a little more tailored to this use case? As-is, they accept GET requests and return a blank page.

labkey-jeckels (approved):
Some suggestions but no need for another review

devtools/src/org/labkey/devtools/TestController.java
Sigmonia: yes, for the scenario we are trying to imitate we just need the response from the server to take a long time, not actually do anything.

Sigmonia: fixed

devtools/src/org/labkey/devtools/TestController.java
Sigmonia: fixed

PR #6457, opened 2025-03-18
Clean up after DisplayColumn refactor

labkey-adam:
#### Rationale
Tests are no longer querying the obsolete "lk-\*" attributes; they've all been converted to use "data-\*" attributes instead

Extract common code out of `ArrayDisplayColumn` and `MapArrayDisplayColumn` into `AbstractExcelDisplayColumn`

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6434x

### DIFF
--- a/api/src/org/labkey/api/writer/HtmlWriter.java
+++ b/api/src/org/labkey/api/writer/HtmlWriter.java
@@ -116,12 +116,14 @@ public class HtmlWriter implements Appendable
             return new AttributeValue(attribute.name(), value);
         }
 
-        public static AttributeValue of(String name, String value)
+        public static AttributeValue cl(String className)
         {
-            if (!name.equals("class") && !name.startsWith("data-"))
-                throw new IllegalStateException("Illegal attribute name: " + name);
+            return new AttributeValue("class", className);
+        }
 
-            return new AttributeValue(name, value);
+        public static AttributeValue data(String name, String value)
+        {
+            return new AttributeValue("data-" + name, value);
         }
 
         private String name()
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/writer/HtmlWriter.java
```diff
@@ -118,8 +118,7 @@ public static AttributeValue of(DOM.Attribute attribute, String value)
 
         public static AttributeValue of(String name, String value)
         {
-            // TODO: eliminate "lk-" option after tests are migrated to use "data-" only
-            if (!name.equals("class") && !name.startsWith("data-") && !name.startsWith("lk-"))
+            if (!name.equals("class") && !name.startsWith("data-"))
```
labkey-tchad: Maybe we just remove this method and have separate methods for data and class attributes? That would make this more of a compile-time check. It would also match the `DOM` builder methods for these attributes.
```
public static AttributeValue cl(String... classNames)
public static AttributeValue data(String nameSuffix, String value)
```

labkey-tchad (approved):
One suggestion

RosalineP (approved):
Thanks Adam!

api/src/org/labkey/api/writer/HtmlWriter.java
labkey-adam: I introduced cl() and data()

labkey-adam:
PlateSetWorklistTest (which exercises `ArrayExcelWriter` and `ArrayDisplayColumn`) passed here: https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_LabkeyPremiumTrunk_GitModules_BiologicsPostgres95/3429832

PR #6461, opened 2025-03-19
Issue 52325: False error on calculated column referencing another field with a slash in its name

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52325
ParseCalculatedColumnAction is provided a columnMap of just the fields in the given domain so use FieldKey.fromParts(key), if we use FieldKey.fromString(key) then the "/" characters of the field name are not encoded.

#### Related Pull Requests
- https://github.com/LabKey/premiumModules/pull/197

#### Changes
- FieldKey.fromParts(key) for columnMap key


labkey-matthewb (approved):
This would have to be changed in the future if we ever support lookups in CC.  That's fine, it's not a backward compatibility problem.

PR #6434, opened 2025-03-12
DisplayColumn refactor

labkey-adam:
#### Rationale
Eliminate many unused/unnecessary DisplayColumn methods. Migrate to using `HtmlWriter` and `DOM` for HTML generation.

PR #6464, opened 2025-03-20
Don't hide the first column in a display column group

labkey-adam:
#### Rationale
A bad refactor of DisplayColumnGroup ended up hiding every column when the "same" checkbox was checked. Need to leave the first column alone.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6434


labkey-adam:
Failing test passed on this FB: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySqlserverb/3429492?buildTab=tests

labkey-matthewb (approved):

PR #6465, opened 2025-03-20
Issue 52036: Add new ColumnHeaderType to allow for more human-readable headers for template files

labkey-susanh:
#### Rationale
Issue [52036](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52036) When a field names contain slashes, the template file that is created using the `DisplayFieldKey` header type replaces slashes with dots. The resulting file will work for importing data, but we generate a warning about the dotted field not being recognized, and a human reading the file will not know that the replacement is a valid one.  The dot replacement is mostly useful for saving query views where the dot syntax is useful in constructing the query. There isn't another existing header type that produces field names that function for importing but doesn't include unexpected substitutions, so a new one (`ImportField`) has been added, most specifically for the purpose of these template files.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/716

#### Changes
- Add `ColumnHeaderType.ImportField`
- Add `QueryKey.toImportDisplayString`


bbimber:
@labkey-adam and @labkey-susanh , yes Adam is exactly right about my 24.11 commit targeting the same problem. I think the other pr includes more detail, but one can basically address the 'null' template name problem by either doing what I added or what's done here. I don't think it matters very much which, but you don't need both. 

The one minor fringe case (which could be impossible to encounter) is a query name that is an invalid or suboptimal file name. No code checks for that in this code, I think. My change (which I am not all that attached to) defaulted to the noun 'Import' when prefix is null, rather than the query name. The latter is more like LK behavior before this bug was introduced.

Anyway, I suggest you guys pick one direction and go with that. I have no personal preference.

labkey-susanh:
> @labkey-adam and @labkey-susanh , yes Adam is exactly right about my 24.11 commit targeting the same problem. I think the other pr includes more detail, but one can basically address the 'null' template name problem by either doing what I added or what's done here. I don't think it matters very much which, but you don't need both.
> 
> The one minor fringe case (which could be impossible to encounter) is a query name that is an invalid or suboptimal file name. No code checks for that in this code, I think. My change (which I am not all that attached to) defaulted to the noun 'Import' when prefix is null, rather than the query name. The latter is more like LK behavior before this bug was introduced.
> 
> Anyway, I suggest you guys pick one direction and go with that. I have no personal preference.

Though this code doesn't check for legal file name, we do make a legal file name out of the prefix that's provided further down in the chain, so the case you mention (which is not such a fringe case) should be handled properly.

I've made the `@NotNull` annotation Adam suggested, and I think we can let Ben's change merge forward and eventually remove that as redundant.

bbimber:
OK. And fair enough on other code doing the checking. I am unsure if labkey will allow me to put illegal characters into a query name anyway, which was why I called it fringe. That plan seems fine.

api/src/org/labkey/api/data/SchemaTableInfo.java
```diff
@@ -1083,7 +1083,7 @@ else if (pair.second instanceof StringExpressionFactory.URLStringExpression)
         if (templates.size() == 0)
         {
             URLHelper url = PageFlowUtil.urlProvider(QueryUrls.class).urlCreateExcelTemplate(ctx.getContainer(), getPublicSchemaName(), getName());
-            url.addParameter("headerType", ColumnHeaderType.Name.name()); // CONSIDER: Use DisplayFieldKey instead
+            url.addParameter("headerType", ColumnHeaderType.ImportField.name()); // CONSIDER: Use DisplayFieldKey instead
```
XingY: stale comment?

api/src/org/labkey/api/data/ColumnHeaderType.java
```diff
@@ -58,6 +58,30 @@ public String getText(DisplayColumn dc)
...
+    ImportField("Import Field", "The official name of the field, without escaping or replacements") {
+        @Override
+        public String getText(DisplayColumn dc)
+        {
+            ColumnInfo columnInfo = dc.getColumnInfo();
+            String name;
+            if (columnInfo != null)
+            {
+                name = columnInfo.getName();
+                org.labkey.api.query.FieldKey fieldKey = org.labkey.api.query.FieldKey.fromString(name);
```
XingY, line 72: any possibility fieldKey is null?

XingY (approved):

api/src/org/labkey/api/data/ColumnHeaderType.java
labkey-susanh: Only if `name` were null, which doesn't seem possible.

api/src/org/labkey/api/data/SchemaTableInfo.java
labkey-susanh: Removed

PR #6469, opened 2025-03-21
Issue 52614: AssertionError deleting site group from its details page

labkey-jeckels:
#### Rationale
We were sometimes failing to properly do audit logging for group events.

#### Changes
- Ensure audit logging happens in the DB transaction
- Ensure a container is set
- Move group audit events to the main API instead of being handled by the caller
- Propagate User object to be used for logging
- Switch to passing Container objects instead of Strings representing the container's EntityId
- Pass key values into GroupAuditEvent constructor, avoiding the need to refetch them later
- Consistently set projectId in AuditTypeEvent constructor


api/src/org/labkey/api/audit/AuditTypeEvent.java
```diff
@@ -56,16 +55,16 @@ public class AuditTypeEvent
...
-    {
-        this(eventType, container.getId(), comment);
-    }
-
-    public AuditTypeEvent(String eventType, String container, String comment)
+    public AuditTypeEvent(String eventType, @NotNull Container container, String comment)
     {
         _eventType = eventType;
-        _container = container;
+        _container = container.getId();
```
labkey-adam: Should we take this a step further and keep `_container` and `_project` as `Container`? Looks like LogManager and AuditLogImpl turn these IDs back into Container objects.

labkey-adam (approved):

api/src/org/labkey/api/audit/AuditTypeEvent.java
labkey-jeckels: Sure. I thought about this but didn't go all of the way initially. I pushed edits to make the switch.

PR #6479, opened 2025-03-25
Issue 52537: WebDav request for S3 file list can stall indefinitely

labkey-jeckels:
#### Rationale
Some operations, like S3 API calls, may be blocked on a response in a way that can be interrupted.

#### Changes
- Treat `HEAD`, `PROPFIND`, and `OPTIONS` requests as read-only, like `GET` requests
- Interrupt the thread running the long-running read-only requests

### DIFF
--- a/api/src/org/labkey/api/data/TransactionFilter.java
+++ b/api/src/org/labkey/api/data/TransactionFilter.java
@@ -104,16 +104,19 @@ public class TransactionFilter implements Filter
                             long cutoff = HeartBeat.currentTimeMillis() - timeout;
                             for (Map.Entry<Thread, RequestTracker> entry : pending.entrySet())
                             {
-                                Object readOnly = entry.getValue().request.getAttribute(READ_ONLY_ATTRIBUTE_NAME);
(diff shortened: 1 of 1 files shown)


labkey-matthewb (approved):

labkey-matthewb (commented):
adding Thread.interrupt() looks appropriate here.  I'm not sure it will address the original issue.  I think most synchronization operations throw InterruptedException,  but this may not be the case for most IO operations.

labkey-jeckels:
> adding Thread.interrupt() looks appropriate here. I'm not sure it will address the original issue. I think most synchronization operations throw InterruptedException, but this may not be the case for most IO operations.

I'm hoping the interrupt() will help here since the thread includes:

java.base@17.0.14/java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1047)
java.base@17.0.14/java.util.concurrent.CountDownLatch.await(CountDownLatch.java:230) 

but there are plenty of cases where this won't help, and we're only attempting to interrupt for read-only ops so mutating requests will continue to block regardless

PR #6482, opened 2025-03-25
Cache reload task limit

labkey-martyp:
#### Rationale
In large imports database cache can queue up unlimited post commit cache reload tasks. This PR limits the cache reload tasks to the max size of the cache.

Part of the fix for this [ticket](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49926). 

#### Changes
- Count cache reload commit tasks to not exceed cache size
- Don't count tasks that have already been added
- Add unit test to cover


api/src/org/labkey/api/data/DatabaseCache.java
```diff
@@ -384,6 +393,50 @@ protected Cache<String, String> createSharedCache(int maxSize, long defaultTimeT
             }
         }
 
+        @Test
+        public void testDatabaseCacheTransactionLimit()
+        {
+            MyScope scope = new MyScope();
+
+            BlockingCache<String, Integer> cache = DatabaseCache.get(scope, 10, "Test Cache", new TestCacheLoader());
```
labkey-adam: You might want to test this, but I think this will leak a non-temporary cache on every invocation. See the test case above (line 290 or so) that documents this and offers a workaround. You could extract a class to use in both places (or not).

labkey-adam (changes requested):

api/src/org/labkey/api/data/DatabaseCache.java
labkey-martyp: Ok I wasn't able to prove if it was leaking or not, but I refactored to do the same method as you had done. I factored out a common class for the two places.

labkey-adam (approved):

labkey-adam:
@labkey-martyp search tests are hitting your assert

labkey-adam (approved):

PR #6487, opened 2025-03-25
Fix Issue 52636

labkey-alan:
#### Rationale
This PR alters the behavior of InClause to ignore values that cannot be parsed, which fixes Issues [52332](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52332) and [52636](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52636)

#### Related Pull Requests
- 

#### Changes
- SimpleFilter InClause - Ignore values that cannot be parsed


api/src/org/labkey/api/data/SimpleFilter.java
```diff
@@ -762,18 +762,32 @@ else if (!isNegated())
...
-                convertedParams = new Object[params.length];
+                convertedParams = new ArrayList<>();
 
-                for (int i = 0; i < params.length; i++)
-                    convertedParams[i] = CompareType.convertParamValue(colInfo, params[i]);
+                for (Object param : params)
+                {
+                    try
+                    {
+                        Object convertedValue = CompareType.convertParamValue(colInfo, param);
```
labkey-nicka, line 783: There is now the potential that none of the values are valid in which case I think we should induce the same logic that currently resides above these changes.

```java
if (params.length == 0)
{
    if (isIncludeNull())
        in.append(alias).append(" IS ").append(isNegated() ? " NOT " : "").append("NULL");
    else if (!isNegated())
        in.append(alias).append(" IN (NULL)");  // Empty list case; "WHERE column IN (NULL)" should always be false
    else
        in.append("1=1");

    return in;
}
```

api/src/org/labkey/api/data/SimpleFilter.java
```diff
@@ -762,18 +762,32 @@ else if (!isNegated())
...
+                        convertedParams.add(convertedValue);
+                    }
+                    catch (RuntimeSQLException e)
+                    {
```
labkey-nicka, line 787: Should be coverable by adding test cases to `InClauseTestCase`.

api/src/org/labkey/api/data/SimpleFilter.java
labkey-jeckels: Agreed

api/src/org/labkey/api/data/SimpleFilter.java
labkey-alan: Done.

api/src/org/labkey/api/data/SimpleFilter.java
labkey-alan: Done

labkey-nicka (approved):

labkey-jeckels (approved):
Nice test coverage.

PR #6491, opened 2025-03-26
Switch links over pipeline grid back to unstyled text

labkey-jeckels:
#### Rationale
Improvements to HTML generation switched the links over the pipeline job grid to styled text links instead of plain links

#### Changes
- Strip the class so the links render as they did before
- Comments about the need for no-arg audit event constructors


labkey-adam:
Nothing to do with this PR, but looks like `EHRSecurityEscalatorAuditProvider` is unhappy with today's constructor change to `SecurityEscalationEvent`

labkey-adam (approved):

PR #6492, opened 2025-03-26
Fix client dependency handling for wiki version and print actions

labkey-adam:
#### Rationale
The version, print, printAll, and printBranch actions were failing to propagate client dependencies into the view before rendering content. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52641

### DIFF
--- a/wiki/src/org/labkey/wiki/model/WikiVersion.java
+++ b/wiki/src/org/labkey/wiki/model/WikiVersion.java
@@ -17,13 +17,13 @@
 package org.labkey.wiki.model;
 
 import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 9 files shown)


wiki/src/org/labkey/wiki/model/WikiVersion.java
```diff
@@ -35,15 +35,8 @@
 import java.util.Collection;
 import java.util.Date;
 import java.util.Map;
-import java.util.Set;
 import java.util.regex.Pattern;
 
-
-/**
- * User: Tamra Myers
- * Date: Feb 14, 2006
```
labkey-klum, line 44: Bye Tamra!

labkey-klum (approved):

PR #6494, opened 2025-03-26
Input forms need to format Time values with milliseconds when present

labkey-adam:
#### Rationale
Milliseconds were truncated from Time values on update and reshow. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52390

### DIFF
--- a/api/src/org/labkey/api/util/DateUtil.java
+++ b/api/src/org/labkey/api/util/DateUtil.java
@@ -1532,6 +1532,11 @@ Parse:
             return DateUtil.parseDateTime(ContainerManager.getRoot(), s);
         }
 
+        private Time parseTime(String s)
+        {
+            return new Time(DateUtil.parseDateTimeUS(s, DateTimeOption.TimeOnly, false));
+        }
+
         void assertIllegalDate(String s)
         {
             try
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -1953,7 +2016,35 @@ public void testTimezone()
...
+            Calendar c = newCalendar(l);
+            c.clear(HOUR);
+            c.clear(AM_PM);
+            c.set(HOUR_OF_DAY,0);
+            l = c.getTimeInMillis();
+            assertEquals(toISO(l, false).length(), "1999-12-31".length());
+        }
+
+        @Test
+        public void testIsoTimeeFormat()
```
labkey-matthewb: Did you mean to update this to test toISO(Time)?

labkey-matthewb (changes requested):
looks good, but test needs update?

api/src/org/labkey/api/util/DateUtil.java
labkey-adam: Yes, working on that test now

XingY (commented):
Why re-show the converted values instead of the original values?

labkey-adam:
> Why re-show the converted values instead of the original values?

DataRegion-generated forms (insert, update) have always worked this way, using typed-value maps. These maps typically come directly from the database (update case) and of course need to be converted to String values for display. TableViewForm does have a String-valued map as well in the reshow case, but we haven't implemented an alternative code path that renders the form with String values in only that case. We definitely need to fix the conversion for the update case and that handles the reshow case as well, so I'm not inclined to rewrite `DataRegion.renderUpdateForm()` and `DataRegion.renderInputForm()`.

labkey-matthewb (approved):

XingY (approved):

PR #6496, opened 2025-03-27
Issue 52616: LKSM: When two lineage alias columns are in a file we silently ignore one

XingY:
#### Rationale
A change was introduced in #6213 to report exception when duplicate columns are detected and we have seen a handful exceptions since. Those exceptions helped uncover issues when we are not handling duplicate columns well, but it also adds noise to the reporting since most of the time it's due to user error that's handled as expected by the app (failed import attempt). This PR revert the change to log exceptions for duplicate columns. It also adds some additional checks for previously missed duplicate detections. 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1760
- https://github.com/LabKey/labkey-ui-premium/pull/723
- https://github.com/LabKey/platform/pull/6496
- https://github.com/LabKey/limsModules/pull/1282

#### Changes
- don't report duplicate column warnings to mothership
- add check for duplicate columns during SimpleTranslator.selectAll within skipColumns, since those won't be detected later due to being skipped here
- check for duplicate parent columns from user input for samples and dataclass

### DIFF
--- a/api/src/org/labkey/api/dataiterator/MapDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/MapDataIterator.java
@@ -88,10 +88,7 @@ public interface MapDataIterator extends DataIterator
                 _findMap.put(in.getColumnInfo(i).getName(),i);
             }
             if (null != duplicates)
-            {
-                var ex = new IllegalStateException("Data has duplicate columns: '" + StringUtils.join(duplicates.toArray(), ", ") + "'");
-                ExceptionUtil.logExceptionToMothership(HttpView.currentRequest(), ex, true);
-            }
+                LOGGER.warn("Data has duplicate columns: '" + StringUtils.join(duplicates.toArray(), ", ") + "'");
         }
 
         @Override
--- a/api/src/org/labkey/api/query/AbstractQueryUpdateService.java
+++ b/api/src/org/labkey/api/query/AbstractQueryUpdateService.java
@@ -327,6 +327,11 @@ public abstract class AbstractQueryUpdateService implements QueryUpdateService
         return hasPermission(user, context.getInsertOption().updateOnly ? UpdatePermission.class : InsertPermission.class);
     }
 
+    // override this
+    protected void preImportDIBValidation(DataIteratorBuilder in)
+    {
+    }
+
     protected int _importRowsUsingDIB(User user, Container container, DataIteratorBuilder in, @Nullable final ArrayList<Map<String, Object>> outputRows, DataIteratorContext context, @Nullable Map<String, Object> extraScriptContext)
     {
         if (!hasImportRowsPermission(user, container, context))
@@ -340,6 +345,8 @@ public abstract class AbstractQueryUpdateService implements QueryUpdateService
         {
             context.setDataSource((String) extraScriptContext.get(DataIteratorUtil.DATA_SOURCE));
         }
+
+        preImportDIBValidation(in);
 
         boolean skipTriggers = context.getConfigParameterBoolean(ConfigParameters.SkipTriggers) || context.isCrossTypeImport() || context.isCrossFolderImport();
         boolean hasTableScript = hasTableScript(container);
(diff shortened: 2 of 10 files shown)


api/src/org/labkey/api/dataiterator/MapDataIterator.java
```diff
@@ -88,10 +88,7 @@ public MapDataIteratorImpl(DataIterator in, boolean mutable, Set<String> skip)
                 _findMap.put(in.getColumnInfo(i).getName(),i);
             }
             if (null != duplicates)
-            {
-                var ex = new IllegalStateException("Data has duplicate columns: '" + StringUtils.join(duplicates.toArray(), ", ") + "'");
-                ExceptionUtil.logExceptionToMothership(HttpView.currentRequest(), ex, true);
-            }
```
labkey-matthewb, line 94: This was a .warn() originally and changed to mothership logging with the plan to make it a thrown exception.

Consult with Josh about this.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -327,6 +327,11 @@ protected boolean hasImportRowsPermission(User user, Container container, DataIt
         return hasPermission(user, context.getInsertOption().updateOnly ? UpdatePermission.class : InsertPermission.class);
     }
 
+    // override this
+    protected void preImportDIBValidation(DataIteratorBuilder in)
```
labkey-matthewb: Nice

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -10067,6 +10071,39 @@ public Map<String, Map<String, MissingFilesCheckInfo>> doMissingFilesCheck(User
...
+        Set<String> seenColumns = new CaseInsensitiveHashSet();
+        for (ColumnDescriptor columnDescriptor : allColumns)
+        {
+            String columnName = columnDescriptor.name;
+            if (parentAliasColumnMap.containsKey(columnName))
+                columnName = parentAliasColumnMap.get(columnName);
+            if (ExperimentService.isInputOutputColumn(columnName) || "parent".equalsIgnoreCase(columnName))
+            {
+                if (seenColumns.contains(columnName))
+                    throw new ApiUsageException(String.format(DUPLICATE_COLUMN_IN_DATA_ERROR, columnName));
```
labkey-matthewb, line 10101: Throwing her might be fine.  If there is a DataIteratorContext we can pass in we could also add to the errors collection there which is the typical way to handle import errors.

labkey-matthewb (approved):

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -1527,6 +1528,13 @@ public void configureDataIteratorContext(DataIteratorContext context)
             }
         }
 
+        @Override
+        protected void preImportDIBValidation(DataIteratorBuilder in)
+        {
+            if (in instanceof DataLoader dataLoader)
```
labkey-matthewb: I was thinking about this over the weekend.  It's hard to inspect DataITeratorBuilder in general.  Should this take DataIterator instead?  The check could also be implemented as a validation-only DataIterator.  There are several cases where we ignore checks if there are no rows.  E.g. we validate on the first call to next().

XingY: I changed the param of this function to take a collection of columnNames and also wired up `_toDataIteratorBuilder` (which the api uses) to call this validation. This function is for checking duplicate import alias vs xxInputs/Type syntax only. 

Because the rows param for query actions is an array of map, duplicate columns (case-insensitive) are lost way before it reaches the data iterator. I've added a TODO comment in QueryController to check for case-insensitive column names ('Name' vs 'name). To tackle exact duplicate ('Name' vs 'Name'), that'd be a more challenging task.

PR #6497, opened 2025-03-27
Issue 52537: Interrupt long-running read-only HTTP request threads

labkey-jeckels:
#### Rationale
An improved version of interrupting HTTP request threads when they exceed a timeout, addressing problems encountered with the original approach.

#### Changes
- Be sure to clear interrupted flag
- Explicitly cancel the async query if we're interrupted
- Suppress logging for InterruptedExceptions during AsyncQueryRequests
- Ensure that we close async Results if they're returned after the requesting thread has given up


labkey-jeckels:
I plan to merge this after we branch for 25.4, but wanted to get the PR opened since I have the changes ready to go.

api/src/org/labkey/api/data/TransactionFilter.java
```diff
@@ -171,6 +172,12 @@ public void doFilter(ServletRequest req, ServletResponse resp, FilterChain chain
             else
                 _pendingRequests.put(t, previousSummary);
             DbScope.finishedWithThread();
+
+            // Clear the interrupted flag before we wrap up
+            if (Thread.interrupted())
```
labkey-adam, line 177: I've never liked the name of this method, but I can't blame you for that...

labkey-adam:
> I plan to merge this after we branch for 25.4, but wanted to get the PR opened since I have the changes ready to go.

Great, I was going to suggest the same.

Do you understand why the test suites showed no problems with the first attempt? Probably good to run additional suites on this.

labkey-adam (approved):

labkey-matthewb (approved):

labkey-jeckels:
> > I plan to merge this after we branch for 25.4, but wanted to get the PR opened since I have the changes ready to go.
> 
> Great, I was going to suggest the same.
> 
> Do you understand why the test suites showed no problems with the first attempt? Probably good to run additional suites on this.

Yes. TeamCity doesn't run with a timeout set, so the code wasn't active enough to get exercised. For now, I changed the timeout to 10 seconds in my branch and will run a variety of suites on it.

labkey-jeckels:
I confirmed that TeamCity complained loudly when I enabled a timeout and reinstated the bug in my original attempt.

I'd like to get TeamCity running with a timeout but don't want to make it the default for on-premise. Merging as-is for now.

PR #6500, opened 2025-03-27
Show the length of the field name in the exp.Fields query

labkey-jeckels:
#### Rationale
Add some reporting to help identify fields that may have a storage mismatch

#### Changes
- Add a FieldNameLength column and other calculated values to the exp.Fields query
- Metric to report the number of fields with a storage column name mismatch
- Rudimentary report for lost values in sample types

### DIFF
--- a/experiment/src/org/labkey/experiment/api/BaseFieldsTable.java
+++ b/experiment/src/org/labkey/experiment/api/BaseFieldsTable.java
@@ -31,6 +31,7 @@ public abstract class BaseFieldsTable extends FilteredTable<ExpSchema>
         addWrapColumn(_rootTable.getColumn("Label"));
         addWrapColumn(_rootTable.getColumn("Description"));
         addWrapColumn(_rootTable.getColumn("RangeURI"));
+        addWrapColumn(_rootTable.getColumn("StorageColumnName"));
     }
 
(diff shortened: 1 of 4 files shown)


labkey-adam (approved):

labkey-tchad:
`ReportLostFieldValuesAction` doesn't seem to include sample types. I created mostly identical list, source type, and sample type domains with long field names. Only the list and source columns show up in the report.

labkey-jeckels:
> `ReportLostFieldValuesAction` doesn't seem to include sample types. I created mostly identical list, source type, and sample type domains with long field names. Only the list and source columns show up in the report.

The report looks at sample types in enough detail to show only rows that actually lost data. Other data types (lists, etc) show all fields that are long enough to potentially lose data.

labkey-tchad:
> > `ReportLostFieldValuesAction` doesn't seem to include sample types. I created mostly identical list, source type, and sample type domains with long field names. Only the list and source columns show up in the report.
> 
> The report looks at sample types in enough detail to show only rows that actually lost data. Other data types (lists, etc) show all fields that are long enough to potentially lose data.

I have samples that lost data that aren't showing up in the report. My sources that lost data show up in the lower list. (Haven't been able to make a list lose data)

PR #6501, opened 2025-03-27
Allow recovery from bad grid view filters

labkey-adam:
#### Rationale
An invalid parameter in a grid view filter blocks editing and deleting of that view. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52661

This PR catches view filter conversion exceptions and adds them as errors that the grid renders, which matches the behavior of URL filters.

Before:

image

After:

image

### DIFF
--- a/api/src/org/labkey/api/data/DataRegion.java
+++ b/api/src/org/labkey/api/data/DataRegion.java
@@ -69,6 +69,7 @@ import org.labkey.api.view.UnauthorizedException;
 import org.labkey.api.view.ViewContext;
 import org.labkey.api.view.template.PageConfig;
 import org.labkey.api.visualization.VisualizationUrls;
+import org.springframework.validation.Errors;
 
 import java.io.IOException;
 import java.io.StringWriter;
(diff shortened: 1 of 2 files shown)


labkey-adam:
@labkey-matthewb let me know if there's a better way to wire through errors like this. I'm surprised we haven't ever handled exceptions from `QueryView.setupDataView()` before.

labkey-matthewb:
Update is this an old bug or new?  If new could it be related to https://github.com/LabKey/platform/pull/6487 ?

labkey-matthewb (approved):
reusing the mechanism for handling SQL errors seems good.  Of course, not throwing in addURLFilters() also seems good.   Do filters need a validate(),  so we can control when the error is deteted?

labkey-adam:
> Update is this an old bug or new? If new could it be related to #6487 ?

I'm sure it's an old issue and not related to that PR. All my repro scenarios are on 25.3.

labkey-adam:
> reusing the mechanism for handling SQL errors seems good. Of course, not throwing in addURLFilters() also seems good. Do filters need a validate(), so we can control when the error is deteted?

I thought about using `_parseErrors`, but these get rendered as "Query <blah> has errors: " which isn't right. I like that the error can be very specific ("Invalid view filter") with this approach. Somehow *URL* filter problems are getting rendered as errors, so maybe I look at that.

labkey-adam:
> > reusing the mechanism for handling SQL errors seems good. Of course, not throwing in addURLFilters() also seems good. Do filters need a validate(), so we can control when the error is deteted?
> 
> I thought about using `_parseErrors`, but these get rendered as "Query has errors: " which isn't right. I like that the error can be very specific ("Invalid view filter") with this approach. Somehow _URL_ filter problems are getting rendered as errors, so maybe I look at that.

URL filters aren't applied until `DataRegion.getResults()`, which is surrounded by `try-catch(ConversionException)`, among other exceptions. The grid view error handling now shares the same rendering code path as the URL filters.

PR #6504, opened 2025-03-28
Remove Site Admins and Site Developers menu items

labkey-adam:
#### Rationale
- We've decided to stop creating the "Site Admins" and "Site Developers groups". For now, just remove the menu items, to get the test infrastructure acclimated to this change. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=52166
- Render PopupMenuView using LinkBuilder instead of hand-coded HTML
- Suppress most links when server is starting up, since they all just redirect back to the upgrade page

### DIFF
--- a/api/src/org/labkey/api/view/PopupMenuView.java
+++ b/api/src/org/labkey/api/view/PopupMenuView.java
@@ -16,25 +16,22 @@
 
 package org.labkey.api.view;
 
-import lombok.SneakyThrows;
-import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.Nullable;
 import org.labkey.api.util.DOM;
 import org.labkey.api.util.HtmlStringBuilder;
 import org.labkey.api.util.Link.LinkBuilder;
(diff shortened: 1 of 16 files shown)


api/src/org/labkey/api/view/PopupMenuView.java
```diff
@@ -85,194 +91,127 @@ public boolean hasChildren()
...
+                        lastIsSeparator = false;
+                        renderTreeDivider(out);
+                    }
+
+                    hasNonSeparatorItem = true;
+
+                    LI(
+                        cl("dropdown-submenu"),
+                        new LinkBuilder(HtmlStringBuilder.of(child.getText()).append(DOM.createHtmlFragment(I(cl("fa fa-chevron-right")))))
+                            .clearClasses()
```
labkey-jeckels, line 140: Doesn't need to be this PR, but maybe we should create a `replaceWithClass(String...)` or similar method, and change `addClass()` to take `String...`

labkey-jeckels (approved):

PR #6506, opened 2025-03-30
Search more test files for ControllerActionIds

labkey-adam:
#### Rationale
More tests are registering ControllerActionIds. Also, fix NPE when controller no longer exists.

### DIFF
--- a/devtools/src/org/labkey/devtools/ToolsController.java
+++ b/devtools/src/org/labkey/devtools/ToolsController.java
@@ -30,7 +30,6 @@ import org.labkey.api.writer.PrintWriters;
 import org.springframework.validation.BindException;
 import org.springframework.validation.Errors;
 import org.springframework.web.servlet.ModelAndView;
-import org.springframework.web.servlet.mvc.Controller;
 
 import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


devtools/src/org/labkey/devtools/ToolsController.java
```diff
@@ -526,9 +525,13 @@ public ModelAndView getView(Object o, BindException errors) throws IOException
 
             // As of now, Crawler.java and the study tests are the only classes that specify crawler actions
             for (String path : List.of(
-                sourcePath + "/../../../testAutomation/src/org/labkey/test/util/Crawler.java",
-                sourcePath + "/../study/test/src/org/labkey/test/tests/study")
-            )
+                sourcePath + "/../../clientModules/adjudication/test/src/org/labkey/test/tests/adjudication/AdjudicationAbstractBaseTest.java",
+                sourcePath + "/../../ehrModules/ehr/test/src/org/labkey/test/tests/ehr/ComplianceTrainingTest.java",
+                sourcePath + "/../../limsModules/biologics/test/src/org/labkey/test/tests/biologics/BiologicsReportTest.java",
```
labkey-tchad, line 530: Probably don't need to include `BiologicsReportTest`. The test will fail if the action ever goes away.

labkey-tchad (approved):

devtools/src/org/labkey/devtools/ToolsController.java
labkey-adam: > Probably don't need to include `BiologicsReportTest`. The test will fail if the action ever goes away.

Agreed it's not really necessary, but I like the simple rule that every test class that uses `ControllerActionId` is referenced here.

PR #6511, opened 2025-03-31
Migrate more classes and methods to specimen module

labkey-adam:
#### Rationale
Migrate settings, type, and location handling. Migrate pivot queries. Eliminate some unused migration and service methods.


study/api-src/org/labkey/api/specimen/SpecimenManagerNew.java
labkey-klum, line 1: hooray!

labkey-klum (approved):

PR #6519, opened 2025-04-01
Fix various column mapping related errors

XingY:
#### Rationale
The change introduced in #6496 calls dataIterator.getColumns(), which would try to inferColumns based on the current state of the of data iterator if columns are not already initialized, which is problematic for certain study dataset import actions. This PR revert part of the changes so that the added check is done for sampletype and dataclass data only. 

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6519
- https://github.com/LabKey/labkey-ui-components/pull/1768

- https://github.com/LabKey/platform/pull/6496

#### Changes
- limit preImportDIBValidation for only samples and dataclasses
- allow duplicate built-in columns [ListTest.testIgnoreReservedFieldNames](https://teamcity.labkey.org/viewLog.html?buildId=3447805&tab=buildResultsDiv&buildTypeId=LabKey_253Release_Premium_CommunitySqlserver_DailyDSqlserver#testNameId-6594495420183568951)

### DIFF
--- a/api/src/org/labkey/api/dataiterator/SimpleTranslator.java
+++ b/api/src/org/labkey/api/dataiterator/SimpleTranslator.java
@@ -1445,7 +1445,7 @@ public class SimpleTranslator extends AbstractDataIterator implements DataIterat
             skipColumns.addAll(added);
         }
 
-        t.selectAll(skipColumns);
+        t.selectAll(skipColumns, Collections.emptyMap(), false);
 
         return t;
     }
--- a/api/src/org/labkey/api/query/AbstractQueryUpdateService.java
+++ b/api/src/org/labkey/api/query/AbstractQueryUpdateService.java
@@ -77,8 +77,6 @@ import org.labkey.api.gwt.client.AuditBehaviorType;
 import org.labkey.api.ontology.OntologyService;
 import org.labkey.api.pipeline.PipeRoot;
 import org.labkey.api.pipeline.PipelineService;
-import org.labkey.api.reader.ColumnDescriptor;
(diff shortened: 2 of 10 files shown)


labkey-jeckels:
@XingY this change looks good. Are you confident that the problem we saw with datasets can't happen in other forms for samples and sources? I know we only saw test failures for datasets, but do we know why?

labkey-klum (approved):

XingY:
> @XingY this change looks good. Are you confident that the problem we saw with datasets can't happen in other forms for samples and sources? I know we only saw test failures for datasets, but do we know why?

Good question. There are components about study dataset (foreign key remapping?) that I don't have much knowledge about, but I do think samples and dataclass won't have similar problem. The reason is sample and dataclass overrides loadRows() and both had already called `dataLoader.getColumns` prior to calling `super.loadRow`. That said, I pushed another defensive change to never initialize columns during this new pre-validation step.

labkey-klum:
Agreed, my theory on why this is failing for the dataset tables (even though datasets don't override that method) is that asking the `DataLoader` to load rows from the incoming tsv doesn't infer the FK's correctly. That state is retained in the object so that when we try to remap FK row IDs to the new imported row IDs in the treatment or product tables we fail and just take the (incorrect) exported row IDs. The test fails when it see's broken lookups.

@XingY @labkey-jeckels There is this dreadful method that is probably involved in the error (but not directly causing it) : `DefaultStudyDesignWriter.isColumnNumericForeignKeyToDataspaceTable`

labkey-jeckels (approved):

PR #6522, opened 2025-04-01
Add a helper that relativizes and resolves child.

labkey-bpatel:
#### Related Pull Requests
- 

#### Changes
-


api/src/org/labkey/vfs/FileSystemLike.java
```diff
@@ -222,6 +222,38 @@ static File toFile(FileLike f)
...
+    // root "/a/b/c/", file "/x/y/z.txt" -> "/x/y/z.txt"
+    // root "/a/b/c/", file "/a/b/c/d.txt" -> "d.txt"
+    static URI getRelativeURI(URI rootURI, URI fileURI)
+    {
+        return rootURI.relativize(fileURI);
+    }
+
+    static FileLike resolveChildWithRelativeURI(URI rootURI, URI fileURI)
+    {
+        URI relativeURI = getRelativeURI(rootURI, fileURI);
```
labkey-matthewb: I think we can use URIUtil.relativize() here, rather than introduce a new method.

api/src/org/labkey/vfs/FileSystemLike.java
```diff
@@ -222,6 +222,38 @@ static File toFile(FileLike f)
...
+    // If fileURI is outside rootURI, the result remains fileURI/unchanged.
+    // Examples:
+    // root "/a/b/c/", file "/x/y/z.txt" -> "/x/y/z.txt"
+    // root "/a/b/c/", file "/a/b/c/d.txt" -> "d.txt"
+    static URI getRelativeURI(URI rootURI, URI fileURI)
+    {
+        return rootURI.relativize(fileURI);
+    }
+
+    static FileLike resolveChildWithRelativeURI(URI rootURI, URI fileURI)
```
labkey-matthewb: What is the use case for having a rootURI w/o already having a FileSystemLike?  e.g. where did root come from?  (e.g. PipeRoot can return FileSystemLike objects)

api/src/org/labkey/vfs/FileSystemLike.java
labkey-bpatel: good point, now removed

api/src/org/labkey/vfs/FileSystemLike.java
labkey-bpatel: Good questions! Intent was to pass rootURI created using FileSystemLike, now realized it is redundant. Please see my updates, let me know what you think! Thanks!

api/src/org/labkey/vfs/FileSystemLike.java
```diff
@@ -222,6 +222,33 @@ static File toFile(FileLike f)
         return p.toFile();
     }
 
+    static FileLike resolveChildWithRelativeURI(File file)
+    {
+        URI fileURI = file.toURI();
+        FileLike allowedRoot = new FileSystemLike.Builder(fileURI).root();
```
labkey-matthewb, line 228: allowedRoot==file???
If this were not static then allowed root would be the getRoot()

labkey-bpatel: ah, thought that the root() was getting the root directory from a file. I will change that line to
FileLike allowedRoot = new FileSystemLike.Builder(file.getParentFile()).root();

labkey-matthewb: If we already "trust" file we can simply use FileSystemLike.wrapFile().  This method helper is for when we don't trust the file yet, and want to verify it is within a known directory and then "adopt" it (e.g. turn a URI/File/Path into a FileLike).

PR #6526, opened 2025-04-02
Issue 52517: Add metric for row counts from file import, update, and merge

labkey-susanh:
#### Rationale
Issue [52517](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52517)

#### Changes
- Add metric after `AbstractQueryImportAction` for getting sense of file sizes


XingY (approved):
There is a AbstractBiologicsImportAction. Depending on scope, there are other import actions that needs wiring up.

cnathe (approved):

labkey-susanh:
> There is a AbstractBiologicsImportAction. Depending on scope, there are other import actions that needs wiring up.

Thanks, Xing. It looks like that action is for custom import files (e.g, FASTA files), which are not really of interest for the purposes of this metric.

PR #6527, opened 2025-04-03
Wiki webpart should continue to resolve renamed wiki with alias

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52729

### DIFF
--- a/wiki/src/org/labkey/wiki/model/BaseWikiView.java
+++ b/wiki/src/org/labkey/wiki/model/BaseWikiView.java
@@ -97,7 +97,8 @@ public abstract class BaseWikiView extends JspView<Object>
             {
                 String aliasedName = WikiSelectManager.getNameForAlias(c, name);
                 if (aliasedName != null)
-                    wiki = WikiSelectManager.getWiki(c, name);
+                    wiki = WikiSelectManager.getWiki(c, aliasedName);
+
             }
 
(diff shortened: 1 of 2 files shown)


Sigmonia (approved):

wiki/src/org/labkey/wiki/model/BaseWikiView.java
```diff
@@ -93,6 +93,13 @@ else if (folderHasWikis)
             if (null == wiki)
                 wiki = WikiSelectManager.getWiki(c, name);
 
+            if (null == wiki)
+            {
+                String aliasedName = WikiSelectManager.getNameForAlias(c, name);
+                if (aliasedName != null)
+                    wiki = WikiSelectManager.getWiki(c, name);
```
Sigmonia:
```suggestion
                    wiki = WikiSelectManager.getWiki(c, aliasedName);

```

Sigmonia (changes requested):

wiki/src/org/labkey/wiki/model/BaseWikiView.java
labkey-adam: Good catch, @labkey-ians !

Sigmonia (approved):

PR #6531, opened 2025-04-04
Stop supporting dataset schema.tsv import format

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52619

Remove the long-deprecated dataset schema TSV import format. In its place, I've added a bulk import option via the dataset XML format (manifest and metadata).

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2395


labkey-adam:
Study PostgreSQL suite looks good: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_ModuleSuites_StudyPostgres/3452942?buildTab=tests

labkey-klum (approved):
Changes look fine, I pulled down your branch as well. The layout of the new import page is a bit odd since we tend to avoid table layouts in favor of div based. Did you try any of the labkey taglibs or builders for this form, maybe something like:

```
       <%= new TextArea.TextAreaBuilder().name("manifest").label("Dataset manifest XML")
                .required(true)
                .formGroup(true)
                .value(h(bean.getManifest()))
                .columns(160)
                .rows(20)
        %>
```

labkey-adam:
> Changes look fine, I pulled down your branch as well. The layout of the new import page is a bit odd since we tend to avoid table layouts in favor of div based. Did you try any of the labkey taglibs or builders for this form, maybe something like:
> 
> ```
>        <%= new TextArea.TextAreaBuilder().name("manifest").label("Dataset manifest XML")
>                 .required(true)
>                 .formGroup(true)
>                 .value(h(bean.getManifest()))
>                 .columns(160)
>                 .rows(20)
>         %>
> ```

I mostly just copied over the previous "schema.tsv" JSP without thinking about it much. I expect only tests will use this page. But I can look at modernizing it a bit.

PR #6532, opened 2025-04-04
Issue 52748: NPE from AdminController$LogClientExceptionAction

labkey-jeckels:
#### Rationale
IntelliJ's warning was correct - `errorCode` can be null, and that makes `Map.of()` upset.

#### Changes
- Use a `HashMap` which is null-tolerant

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -10539,7 +10539,11 @@ public class AdminController extends SpringActionController
                 form.getUsername()
             );
 
-            return success(Map.of("errorCode", errorCode, "loggedToMothership", errorCode != null));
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -10539,7 +10539,11 @@ public Object execute(ExceptionForm form, BindException errors)
                 form.getUsername()
             );
 
-            return success(Map.of("errorCode", errorCode, "loggedToMothership", errorCode != null));
+            Map<String, Object> results = new HashMap<>();
```
labkey-nicka, line 10542: Aye, sorry I missed the warning. Thanks for fixing.

labkey-nicka (approved):

PR #6537, opened 2025-04-07
Issue 52716: Reserve SampleType and variants as a field name

labkey-susanh:
#### Rationale
Issue 52716 - [LKSM: Can create a field called "SampleType" and then we incorrectly message that we ignore it](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52716)

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1318

#### Changes
- Update `SampleTypeDomainKind` with more reserved field names
- Add `Data Class` as reserved field name for `DataClassDomainKind` (already have `DataClass` reserved)


api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
```diff
@@ -112,6 +112,9 @@ public class SampleTypeDomainKind extends AbstractDomainKind<SampleTypeDomainKin
         RESERVED_NAMES.add("Created By");
         RESERVED_NAMES.addAll(Arrays.stream(ExpSampleTypeTable.Column.values()).map(ExpSampleTypeTable.Column::name).toList());
         RESERVED_NAMES.addAll(Arrays.stream(ExpMaterialTable.Column.values()).map(ExpMaterialTable.Column::name).toList());
+        RESERVED_NAMES.add("Sample Set");
```
XingY: While at it, also add "Parent", "Modified By", "Aliquoted From"... Well, wish we have a smarter way to handle all these...

labkey-susanh: Done, perhaps in a smarter way, but at least with less typing.

XingY (commented):
Looks better. "Aliquoted From" is still allowed.

labkey-susanh:
> Looks better. "Aliquoted From" is still allowed.

Change pushed.

XingY (approved):

PR #6542, opened 2025-04-07
Issue 52404: Additional ContainerFilter cleanup

labkey-jeckels:
#### Rationale
A little more cleanup as we continue to make ContainerFilter APIs more robust.

#### Changes
- Remove now-unused variable
- Consolidate duplicate code
- Stronger Nullable/NotNull

### DIFF
--- a/api/src/org/labkey/api/study/DataspaceContainerFilter.java
+++ b/api/src/org/labkey/api/study/DataspaceContainerFilter.java
@@ -16,9 +16,7 @@
 package org.labkey.api.study;
 
 import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 import org.labkey.api.cache.Cache;
-import org.labkey.api.cache.CacheLoader;
 import org.labkey.api.cache.CacheManager;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerFilter;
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/study/DataspaceContainerFilter.java
```diff
@@ -170,7 +168,7 @@ public Collection<GUID> generateIds(Container currentContainer, Class<? extends
             allowedContainers.retainAll(studyContainers);
 
             if (allowedContainers.containsAll(studyContainers))
-                return null;
+                return Collections.emptySet();
```
labkey-adam: Are you sure this is correct? This used to return `null`, meaning all containers, but now returns empty set, meaning no containers.

labkey-jeckels:
@labkey-adam reminder to take a look whenever you have a chance. Next week is great.

labkey-adam:
> @labkey-adam reminder to take a look whenever you have a chance. Next week is great.

I left a comment yesterday, but I guess I didn't submit the "review".

api/src/org/labkey/api/study/DataspaceContainerFilter.java
labkey-jeckels: You're right. I got mixed up because `AllInProject` will never return null. `DataspaceContainerFilter` subclasses that, but can return null.

labkey-adam (approved):

PR #6546, opened 2025-04-09
Make StatusCake data sync more robust in the face of bad JSON responses

labkey-jeckels:
#### Rationale
Our nightly job that pulls uptime data from StatusCake fails every once in a while:

```
org.json.JSONException: A JSONObject text must begin with '{' at 0 [character 1 line 1]
	at org.json.JSONTokener.syntaxError(JSONTokener.java:536)
	at org.json.JSONObject.<init>(JSONObject.java:233)
	at org.json.JSONObject.<init>(JSONObject.java:204)
	at org.labkey.mothership.statuscake.StatusCakeMaintenanceTask.lambda$requestUptime$1(StatusCakeMaintenanceTask.java:312)
	at org.apache.hc.client5.http.impl.classic.CloseableHttpClient.execute(CloseableHttpClient.java:247)
	at org.apache.hc.client5.http.impl.classic.CloseableHttpClient.execute(CloseableHttpClient.java:188)
```

#### Changes
- Capture the string that failed to parse in the hopes of insights
- Retry if it happens

### DIFF
--- a/mothership/src/org/labkey/mothership/statuscake/StatusCakeMaintenanceTask.java
+++ b/mothership/src/org/labkey/mothership/statuscake/StatusCakeMaintenanceTask.java
@@ -43,7 +43,6 @@ import java.time.temporal.ChronoField;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Random;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.labkey.api.query.QueryUpdateService.ConfigParameters.BulkLoad;
--- a/mothership/src/org/labkey/mothership/statuscake/StatusCakeMaintenanceTask.java
+++ b/mothership/src/org/labkey/mothership/statuscake/StatusCakeMaintenanceTask.java
@@ -1,5 +1,6 @@
 package org.labkey.mothership.statuscake;
 
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


mothership/src/org/labkey/mothership/statuscake/StatusCakeMaintenanceTask.java
```diff
@@ -307,9 +314,12 @@ private void requestUptime(Server server, CloseableHttpClient client, ClassicHtt
                 throw new FailedRequestException("Bad response " + response.getCode() + " for " + httpGet.getPath());
             }
 
-            try (Reader reader = new InputStreamReader(entity1.getContent(), StandardCharsets.UTF_8))
+            // Responses are small so they fit comfortably in memory. Capture as a string so that we can report on
```
labkey-nicka, line 317: nit: Is there an associated issue? May be worth mentioning/referring to here.

labkey-nicka (approved):

mothership/src/org/labkey/mothership/statuscake/StatusCakeMaintenanceTask.java
labkey-jeckels: There is not. I considered opening one, but I'd be opening an issue to myself and am the one keeping an eye on the labkey.org behavior. So I opted to be lazy instead.

PR #6548, opened 2025-04-09
Remove support for ancillary studies

labkey-adam:
#### Rationale
Support for ancillary studies was deprecated in 25.3; now it's time to remove it. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=51882

The upgrade script in this PR migrates all existing ancillary studies to published studies. Specimen data is no longer linked from the source study, but otherwise these studies seem to work the same as the ancillary studies.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6392


api/src/org/labkey/api/data/ContainerFilter.java
```diff
@@ -1030,19 +1028,6 @@ public Collection<GUID> generateIds(Container currentContainer, Class<? extends
                 result.add(currentContainer.getEntityId());
             }
 
-            Study study = null;
-            StudyService svc = StudyService.get();
-            if (svc != null)
-                study = svc.getStudy(currentContainer);
-
-            if (study != null && study.isAncillaryStudy())
```
cnathe, line 1038: I don't think this is the case, but would this need to change to check for `study.isSnapshotStudy()` now instead of removing this block of code?

specimen/src/org/labkey/specimen/importer/AbstractSpecimenTask.java
```diff
@@ -168,15 +168,6 @@ public static void doImport(@Nullable Path inputFile, PipelineJob job, SimpleStu
         {
             // do any temp file cleanup
             importHelper.afterImport(ctx);
-
-            // Since changing specimens in this study will impact specimens in ancillary studies dependent on this study,
-            // we need to force a participant/visit refresh in those study containers (if any):
-            for (Study dependentStudy : StudyService.get().getAncillaryStudies(ctx.getContainer()))
```
cnathe, line 174: similar question as above. I don't think this is the case but would we need to change this to some sort of getPublishedStudies() check?

study/src/org/labkey/study/model/StudyManager.java
```diff
@@ -2675,9 +2672,6 @@ public void clearCaches(Container c, boolean unmaterializeDatasets)
         _datasetHelper.clearCache();
         _cohortHelper.clearCache(c);
         _participantCache.remove(c);
-
-        for (StudyImpl substudy : StudyManager.getInstance().getAncillaryStudies(c))
```
cnathe, line 2679: would this and line 1831 above need to iterate over getPublishedStudies() now?

cnathe (approved):

api/src/org/labkey/api/data/ContainerFilter.java
labkey-adam: Previous check was clearly for ancillary studies only, which probably makes sense because they linked back to the source study, where published studies are completely divorced from the source study (other than in the refresh process). This container filter is used in `StudyDesignBaseTable` only... so I'm not too worried about it.

specimen/src/org/labkey/specimen/importer/AbstractSpecimenTask.java
labkey-adam: I think this is important only in the ancillary study case, where specimens are linked, not copied.

study/src/org/labkey/study/model/StudyManager.java
labkey-adam: I don't think so, given that published studies are completely divorced from their source (other than refresh)

PR #6550, opened 2025-04-10
Stop creating and managing the Site Administrators and Developers groups

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=52166

### DIFF
--- a/core/src/org/labkey/core/login/LoginController.java
+++ b/core/src/org/labkey/core/login/LoginController.java
@@ -1911,10 +1911,6 @@ public class LoginController extends SpringActionController
             {
                 errors.reject(ERROR_MSG, "Unable to create user '" + e.getEmail() + "': " + e.getMessage());
             }
-            catch (Exception e)
-            {
-                errors.reject(ERROR_MSG, "Unable to create user: " + e.getMessage());
-            }
 
             return success;
         }
--- a/core/src/org/labkey/core/login/LoginController.java
+++ b/core/src/org/labkey/core/login/LoginController.java
@@ -64,10 +64,10 @@ import org.labkey.api.security.AuthenticationProvider.SSOAuthenticationProvider;
 import org.labkey.api.security.CSRF;
 import org.labkey.api.security.DbLoginService;
 import org.labkey.api.security.EntropyPasswordValidator;
-import org.labkey.api.security.Group;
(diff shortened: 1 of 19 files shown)


labkey-jeckels:
@labkey-adam another couple of test failures to check out

labkey-adam:
> @labkey-adam another couple of test failures to check out

Yep, I'm expecting quite a few failures since the tests have heavy assumptions on the "Site Administrators" group. Fix the BVTs first, then will move to the Daily suite.

core/src/org/labkey/core/login/LoginController.java
```diff
@@ -1899,13 +1903,17 @@ public boolean handlePost(SetPasswordForm form, BindException errors) throws Exc
...
+            catch (InvalidEmailException e)
+            {
+                errors.rejectValue("email", ERROR_MSG, "'" + StringUtils.trimToEmpty(form.getEmail()) + "' is not a valid email address. Please enter an email address in this form: user@domain.tld");
+            }
             catch (UserManagementException e)
             {
                 errors.reject(ERROR_MSG, "Unable to create user '" + e.getEmail() + "': " + e.getMessage());
             }
-            catch (InvalidEmailException e)
+            catch (Exception e)
```
labkey-jeckels: If we're catching `Exception` are we expecting NPEs or other runtime exceptions? Would we want to log the full stack trace?

labkey-jeckels (approved):
Code changes look good. Can review again if there are interesting updates after getting the tests passing, but approving for now.

core/src/org/labkey/core/login/LoginController.java
labkey-adam: We're not expecting anything in particular. I had a coding mistake (forgot to save the policy after assigning the role) and it resulted in a blank page with no exception in the log. I should probably just send it ExceptionUtil so we make noise about it.

labkey-adam: FYI: I removed `catch (Exception)` here since the enclosing block wasn't actually throwing, even in my "forgot to save" case. The blank page was produced on redirect, a ModuleStatus unauthorized response. We might want to investigate that (the page title says 403, but there's no message on the page), but it's a totally different thing.

PR #6552, opened 2025-04-10
Fb dataset update rewrite

labkey-klum:
#### Rationale
Dataset `updateRows` had previously been implemented as a single row delete followed by an insert row. This PR attempts to rewrite this code to look like more typical `QueryUpdateService.updateRow` implementations.

This work was focused on a few main areas and tasks:
- Refactor the helpers in `DatasetDataIteratorBuilder` so they could be used outside of a `DataIterator` context. These helpers were responsible for : LSID, subject ID, sequence number, and participant sequence number generation.
- Add unit tests and selenium tests to validate the integrity of insert and update operations around these special fields
- Swap in the new helpers into `DatasetDataIteratorBuilder`
- Rewrite `DatasetUpdateService.updateRow` to use the new helpers as well as `Table.update` to update the row.


labkey-klum:
@mbellew I'm going to need your opinion on the test failure I'm seeing in `DatasetUpdateService.TestCase.updateRowTest`. It was previously working because of the the code in `SimpleTranslator.MissingValueConvertColumn`, but no longer does because updates are no longer handled through the insert code.

It also looks like other data types don't support this on update either. Do you think the old behavior is still required, should there be code in `DefaultQueryUpdateService` to deal with MV columns?

study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
```diff
@@ -432,8 +429,40 @@ else if (_datasetDefinition.getUseTimeKeyField())
...
+        return null == o ? "" : o.toString();
+    }
+
+    public static Double getOutputDouble(DataIterator it, Integer i)
+    {
+        if (i == null)
+            return null;
+
+        Object o = it.get(i);
+        if (null == o)
```
labkey-matthewb: JdbcType.DOUBLE.convert(o)

study/src/org/labkey/study/model/DatasetDataIteratorBuilder.java
```diff
@@ -432,8 +429,40 @@ else if (_datasetDefinition.getUseTimeKeyField())
         return ret;
     }
 
+    public static String getOutputString(DataIterator it, Integer i)
+    {
+        if (i == null)
+            return null;
+
+        Object o = it.get(i);
+        return null == o ? "" : o.toString();
```
labkey-matthewb: Objects.toString(o,"")

labkey-chrisj (approved):
Tests look good to go

study/src/org/labkey/study/query/DatasetUpdateService.java
```diff
@@ -577,30 +569,16 @@ private void resyncStudy(User user, Container container, @Nullable Set<String> p
...
-            defaultQCState = DataStateManager.getInstance().getStateForRowId(container, defaultQcStateId.intValue());
-
-        String managedKey = null;
-        if (_dataset.getKeyType() == Dataset.KeyType.SUBJECT_VISIT_OTHER && _dataset.getKeyManagementType() != Dataset.KeyManagementType.None)
-            managedKey = _dataset.getKeyPropertyName();
-
         try (DbScope.Transaction transaction = StudyService.get().getDatasetSchema().getScope().ensureTransaction())
         {
+            String lsid = keyFromMap(oldRow);
+            checkDuplicateUpdate(lsid);// Make sure we've found the original participant before doing the update
```
XingY: The comment is for the next line, not `checkDuplicateUpdate`, a merge error I believe.

labkey-matthewb (approved):

PR #6553, opened 2025-04-11
Issue 52666: Ensure that specimen columns don't exceed 48 characters

labkey-jeckels:
#### Rationale
Specimen SQL generation has problems with fields > 48 characters in length, related to issue 52666. We don't need to support them, so we can simply limit the length.

#### Changes
- Prevent saving specimen domains with field names longer than 48 characters


labkey-matthewb (commented):
This strategy is pragmatic for specimen.  However, by itself doesn't solve the intended problem.  Ideally we'd check that the generated storageColumnName matches the column name.  In addition there is the problem of cross-checking that the generated storageColumnName satisfies the complicated postgres name constraint.  This problem affects all types and should be incorporated into DomainImpl.generateStorageColumnName().  

see also StringUtilsLabkey.truncateToUtf8ByteLimit()

labkey-adam (approved):

PR #6558, opened 2025-04-11
Improve display element builders

labkey-adam:
#### Rationale
Making our builders a bit easier to use:

- Code usually interacts directly with builders, so make the builders top-level classes for import convenience and make the display elements inner classes.
- Move all builder classes to the same package (`org.labkey.api.util`)
- Introduce `LinkBuilder` static factory methods `simpleLink()` and `labkeyLink()` for convenience
- Introduce `InputBuilder` static factory methods `checkbox()`, `file()`, `hidden()`, `radio()`, and `text()` for convenience
- Introduce `JspBase.simpleLink()`


labkey-adam:
@bbimber see `LinkBuilder.simpleLink()` and other convenience methods introduced here. Better late than never...

bbimber:
> @bbimber see `LinkBuilder.simpleLink()` and other convenience methods introduced here. Better late than never...

Thank you! I think I had my fill of link refactoring for the time being, but am very glad to see it added. I'll also try to break my habit of over-using PageFlowUtil.

devtools/src/org/labkey/devtools/test/JspTestCaseTest.jsp
```diff
@@ -1,12 +1,19 @@
...
+<%@ page import="org.labkey.api.module.SupportedDatabase" %>
+<%@ page import="java.util.Objects" %>
 <%@ page extends="org.labkey.api.jsp.JspTest.DRT" %>
 
 <%!
     @Test
     public void test1()
     {
-        assertTrue(1==1);
+        ModuleLoader.getInstance().getModules().stream()
```
labkey-jeckels: Is this intended for this PR? `System.out.println()`?

labkey-jeckels (approved):
Maybe we'll need a `simpleLink(String, String)` for some of my suggestions in other PRs. Probably worth it to get rid of more `clearClasses()`

devtools/src/org/labkey/devtools/test/JspTestCaseTest.jsp
labkey-adam: Nope, unintentional. That snuck out of my "ignore" change list. Reverted.

PR #6498, opened 2025-03-27
DatabaseIdentifier

labkey-matthewb:
#### Rationale
Generating correct sql is very dependent on everyone writing identifier consistently.  This is often handled by using append(SqlDialect.makeLegalIdentifier(col.getAlias())) or luck append(col.getName()) or append(col.getAlias()).

The idea is to pass around database identifiers such as table names, column names, both generated and from db metadata, using a class DatabaseIdentifier.

#### Related Pull Requests
- 

#### Changes
-

PR #6564, opened 2025-04-15
Long identifier truncation improvements

labkey-adam:
#### Rationale
Not heeding database-specific truncation rules for identifiers has led to several issues related to long column, table, and alias names, particularly those with non-ASCII characters. Here are two example issues that are resolved by this PR:
- [Issue 52714: Problems with >63 byte at UTF-8 non-ASCII characters in field names](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52714)
- [Issue 52210: LKSM: Not showing fields with cyrillic characters in sample type grid](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52210)

In these cases, `DomainImpl.generateStorageColumnName()` would call an `AliasManager.decideAlias()` variant that usually returned the name as the storage column name, without first making it "legal," which lead to silent truncation to 63 UTF-8 bytes on PostgreSQL. First fix attempt was to call a different method to ensure makeLegalName() was invoked when generating storage names, but this led to some undesirable behavior, such as provisioned columns named "Group" and "User" being stored as "group_" and "user_" (they're keywords on PostgreSQL), which then confused index creation (which of course should know the storage names, but they don't currently). IMO, provisioned column storage names should match the column names as much as possible (I'm less concerned about this for aliases). We're already quoting them appropriately. So the fix here is to stop using AliasManager to generate storage column names; instead, a new simple class `StorageNameGenerator` is now responsible for generating legal and unique storage column names. It calls the dialect-specific truncation method and uniquifies names with a suffix counter, but otherwise it leaves the characters as is. This means we'll start seeing special characters in provisioned tables' column names.

#### Changes
- Introduce `StringUtilsLabKey.truncateStartToUtf8ByteLimit()` that truncates from the *right* end of the string. Add tests. Simplify overly complex `truncateToUtf8ByteLimit()`.
- Move the truncation and "make legal" methods to SqlDialect to allow these to be dialect-specific
- Implement correct truncation in PostgreSQL dialect
- Pass in a `@NotNull` `SqlDialect` to more AliasManager calls
- Introduce `FallBackDialect` for the unfortunate cases where a SqlDialect is not provided to AliasManager; it implements conservative truncation rules that work across all supported databases
- Eliminate `useLegacyMaxLength` flag
- Force callers to specify the number of characters they need reserved for suffixes, etc. Previous code made blanket assumptions that were often incorrect, unnecessary, and/or redundant.
- Switch to surrogate-pair-aware truncation so we don't end up with half characters in our names
- Widen `StorageColumnName` and `mvIndicatorStorageColumnName` to ensure even the longest generated names will fit


api/src/org/labkey/api/util/StringUtilsLabKey.java
```diff
@@ -575,28 +576,43 @@ public static String truncateToUtf8ByteLimit(String s, int maxBytes)
...
             s = new String(bytes, 0, maxBytes, StandardCharsets.UTF_8);
         }
         return s;
     }
 
+    // Truncates a string to UTF-8 bytes <= maxBytes starting from the LAST character (truncating the start of the string)
+    public static String truncateStartToUtf8ByteLimit(String s, int maxBytes)
+    {
+        if (maxBytes < 0)
+            throw new IllegalStateException("maxBytes cannot be negative");
```
labkey-jeckels:
```suggestion
            throw new IllegalArgumentException("maxBytes cannot be negative");

```

api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
```diff
@@ -928,6 +944,64 @@ protected Pattern getSQLScriptProcPattern()
...
+        return identifier.getBytes(StandardCharsets.UTF_8).length > getIdentifierMaxByteLength();
+    }
+
+    @Override
+    public String truncateAndJoin(String... parts)
+    {
+        int maxBytes = getIdentifierMaxByteLength();
+        String ret = String.join("$", parts);
+
+        if (ret.getBytes(StandardCharsets.UTF_8).length > maxBytes)
```
labkey-jeckels: Could use `isIdentifierTooLong()` here

api/src/org/labkey/api/query/AliasManager.java
```diff
@@ -17,262 +17,119 @@
...
-                    case '\\' : sb.append("_bs_"); break;
-                    case '&' : sb.append("_amp_"); break;
-                    case '<' : sb.append("_lt_"); break;
-                    case '>' : sb.append("_gt_"); break;
-                    default: sb.append("_"); break;
-                }
-            }
+            // Oracle rule - truncate to 30 characters
+            String legal = super.makeLegalName(key, reserveCount);
+            // PostgreSQL rule - truncate to 63 b
```
labkey-jeckels:
```suggestion
            // PostgreSQL rule - truncate to 63 bytes

```

api/src/org/labkey/api/query/AliasManager.java
```diff
@@ -405,7 +252,29 @@ public boolean isReserved(String word)
...
+
+            String truncated = m.decideAlias("1234567890123456789012345678901234567890");
+            assertEquals(27, truncated.length());
+            assertTrue(truncated.getBytes(StandardCharsets.UTF_8).length < 60);
+            assertEquals("X13599947545678901234567890", truncated);
+            // For now, not an interesting test at the moment since every non-alphanumeric gets replaced with _. But
+            // this will become interesting if we start allowing Unicode characters in identifier names in the future.
+            String unicode = "\uD83D\uDC7EA\uD83D\uDC7E\uD83E\uDD91\uD83C\uDFBB\uD83C\uDFC2\uD83D\uDC7E\uD83E\uDD91\uD83C\uDFBB\uD83C\uDFC2\uD83D\uDC7E\uD83E\uDD91\uD83C\uDFBB\uD83C\uDFC2";
+            truncated = m.decideAlias(unicode);
+            assertEquals(27, truncated.length());
```
labkey-jeckels, line 329: Is this useful? Won't the last check take care of this too?

api/src/org/labkey/api/util/StringUtilsLabKey.java
```diff
@@ -51,9 +51,69 @@ public class StringUtilsLabKey
     /** Instead of relying on the platform default character encoding, use this Charset */
     public static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
 
+    /** Special character strings that can be used by tests in this class and others */
+    public static final List<String> specialCharacterTestStrings = List.of(
+        "",
+        "A",
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789",
+        "A \" ' ` ~ ! @#$%^&*()_-+= { } [ ] \\ | : ; < > , . ? / 你好 \uD83D\uDC7E",
```
labkey-jeckels, line 60: I'm not understanding the split between the escape sequences and the Unicode characters

labkey-jeckels (approved):
Minor suggestions. I haven't test the code.

api/src/org/labkey/api/util/StringUtilsLabKey.java
labkey-adam: Done

api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
labkey-adam: Done

api/src/org/labkey/api/query/AliasManager.java
labkey-adam: Done

api/src/org/labkey/api/util/StringUtilsLabKey.java
labkey-adam: I tend to use the `\u` syntax for surrogate pairs, because it's easier to distinguish them with that syntax. But I'm probably not consistent and, frankly, this is mostly an artifact of pasting special characters from different places. Do you have a strong preference?

api/src/org/labkey/api/query/AliasManager.java
labkey-adam: Yes, there's definitely some redundancy in these checks. This is mostly about future proofing these tests. It seems likely that we'll change the legal chars in aliases at some point, so the exact return values may change, but the overarching rules need to be heeded.

api/src/org/labkey/api/util/StringUtilsLabKey.java
labkey-jeckels: No, no strong preference. Just struck me as an eclectic assortment. Fine to keep as-is.

labkey-adam:
Merged to https://github.com/LabKey/platform/pull/6498

PR #6568, opened 2025-04-15
Improve handling of bad URL parameters and JSON

labkey-jeckels:
#### Rationale
We can improve messaging and reduce error-level logging in the face of bad input. Specifically, the type of script being used for an export, and the JSON for a request to save a custom query view.

#### Changes
- Use `JSONObject.optString()`
- Switch to `ApiUsageException`
- Default to an HTML template for error views

### DIFF
--- a/query/src/org/labkey/query/CustomViewUtil.java
+++ b/query/src/org/labkey/query/CustomViewUtil.java
@@ -20,6 +20,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.json.JSONArray;
 import org.json.JSONObject;
+import org.labkey.api.action.ApiUsageException;
 import org.labkey.api.data.AnalyticsProviderItem;
 import org.labkey.api.data.ColumnInfo;
 import org.labkey.api.data.DisplayColumn;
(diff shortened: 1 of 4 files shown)


query/src/org/labkey/query/CustomViewUtil.java
```diff
@@ -92,25 +96,33 @@ public static void update(CustomView view, JSONObject jsonView, boolean saveFilt
...
-                String fieldKey = (String)filterInfo.get("fieldKey");
-                String op = Objects.toString(filterInfo.get("op"), "");
-                String value = Objects.toString(filterInfo.get("value"), "");
+                String fieldKey = filterInfo.optString("fieldKey", null);
+                if (fieldKey == null)
+                {
+                    throw new ApiUsageException("No fieldKey for filter");
+                }
+                String op = filterInfo.optString("op");
+                String value = filterInfo.optString("value");
```
labkey-adam, line 109: You do want these to default to empty string, right? Looks like it... but just making sure.

labkey-adam (approved):

query/src/org/labkey/query/CustomViewUtil.java
labkey-jeckels: I guess so. That's what the original code was doing.

PR #6569, opened 2025-04-16
Issue 52854: Add www.googletagmanager.com to img-src CSP when Google Analytics is enabled

labkey-jeckels:
#### Rationale
Enabling Google Analytics reporting should automatically handle all of the CSP settings required

#### Changes
- Add URL to `img-src`
- Ensure audit log entry is transacted
- Reset CSP after the save so the right settings are applied

### DIFF
--- a/core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
+++ b/core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
@@ -117,10 +117,12 @@ public class AnalyticsServiceImpl implements AnalyticsService
     public void resetCSP()
     {
         ContentSecurityPolicyFilter.unregisterAllowedSources(Directive.Connection, ANALYTICS_CSP_KEY);
+        ContentSecurityPolicyFilter.unregisterAllowedSources(Directive.Image, ANALYTICS_CSP_KEY);
 
         if (getTrackingStatus().contains(TrackingStatus.ga4FullUrl))
         {
             ContentSecurityPolicyFilter.registerAllowedSources(Directive.Connection, ANALYTICS_CSP_KEY, "https://*.googletagmanager.com", "https://*.google-analytics.com", "https://*.analytics.google.com");
+            ContentSecurityPolicyFilter.registerAllowedSources(Directive.Image, ANALYTICS_CSP_KEY, "https://www.googletagmanager.com");
         }
     }
 
--- a/core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
+++ b/core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
@@ -20,6 +20,8 @@ import org.jetbrains.annotations.NotNull;
 import org.labkey.api.analytics.AnalyticsService;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
+import org.labkey.api.data.CoreSchema;
+import org.labkey.api.data.DbScope;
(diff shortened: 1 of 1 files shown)


core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
```diff
@@ -117,10 +119,12 @@ public String getDescription()
     public void resetCSP()
     {
         ContentSecurityPolicyFilter.unregisterAllowedSources(Directive.Connection, ANALYTICS_CSP_KEY);
+        ContentSecurityPolicyFilter.unregisterAllowedSources(Directive.Image, ANALYTICS_CSP_KEY);
 
         if (getTrackingStatus().contains(TrackingStatus.ga4FullUrl))
         {
             ContentSecurityPolicyFilter.registerAllowedSources(Directive.Connection, ANALYTICS_CSP_KEY, "https://*.googletagmanager.com", "https://*.google-analytics.com", "https://*.analytics.google.com");
+            ContentSecurityPolicyFilter.registerAllowedSources(Directive.Image, ANALYTICS_CSP_KEY, "https://www.googletagmanager.com");
```
labkey-adam, line 127: Should this be `https://*.googletagmanager.com` like the connection URLs?

core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
```diff
@@ -162,15 +166,19 @@ public void store(Set<TrackingStatus> trackingStatus, String measurementId, Stri
...
-            storeStringValue(AnalyticsProperty.trackingScript.toString(), StringUtils.trimToNull(script));
+            try (DbScope.Transaction t = CoreSchema.getInstance().getScope().ensureTransaction())
+            {
+                String statusString = StringUtils.trimToNull(StringUtils.join(trackingStatus.toArray(), SEPARATOR));
+                storeStringValue(AnalyticsProperty.trackingStatus.toString(), statusString);
+                storeStringValue(AnalyticsProperty.measurementId.toString(), StringUtils.trimToNull(measurementId));
+                storeStringValue(AnalyticsProperty.trackingScript.toString(), StringUtils.trimToNull(script));
+
+                save();
+                writeAuditLogEvent(c, user);
```
labkey-adam, line 177: Use POST_COMMIT task for writing audit log?

labkey-adam (approved):

Sigmonia (approved):

core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
labkey-jeckels: I've always done audit logging in the main transaction. This ensures that we don't commit the real work but then fail to separately commit the audit log message to the DB. But there's the downside that file-based may be written even if the transaction rolls back. Looks like I'm not the first to consider this. https://github.com/LabKey/platform/blob/a2a63507e8de7e13ca355ce487eacb49d48ef9c1/audit/src/org/labkey/audit/model/LogManager.java#L181

core/src/org/labkey/core/analytics/AnalyticsServiceImpl.java
labkey-jeckels: I'm going from the official Google guidance:

https://developers.google.com/tag-platform/security/guides/csp

I considered changing the `Connection` URLs too but decided to be more conservative since this is targeting 25.3.

PR #6571, opened 2025-04-16
Issue 52728: LKSM: Update from file with duplicate rows is only updating once.

XingY:
#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6571
- https://github.com/LabKey/limsModules/pull/1352
- https://github.com/LabKey/testAutomation/pull/2421

#### Changes
- Added checkDuplicateUpdate to `updateRow` in various QUS so the same rows cannot be updated in the same updateRows call
- Modified `ExistingRecordDataIterator` to check for previously seen keys so the same rows are not provided during update/merge
- Add junit test coverage for updating/merging the same list rows using updateRows and DIB

### DIFF
--- a/api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
@@ -262,7 +262,9 @@ public abstract class ExistingRecordDataIterator extends WrapperDataIterator
                 for (int p = 0; p < pkColumns.size(); p++)
                 {
                     sqlf.append(",?");
-                    sqlf.add(pkSuppliers.get(p).get());
+                    Object pkVal = pkSuppliers.get(p).get();
+                    sqlf.add(pkVal);
+                    pkKeys.add(pkVal.toString());
                 }
                 sqlf.append(")");
                 String pkKey = StringUtils.join(pkKeys, ", ");
--- a/api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
@@ -173,7 +173,7 @@ public abstract class ExistingRecordDataIterator extends WrapperDataIterator
         if (!_context.getErrors().hasErrors() && ret && !pkColumns.isEmpty())
         {
             prefetchExisting();
-            if (_context.getErrors().hasErrors())
+            if (_context.shouldCancel())
                 return false;
         }
         return ret;
--- a/api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
@@ -65,7 +65,7 @@ public abstract class ExistingRecordDataIterator extends WrapperDataIterator
     final boolean _checkCrossFolderData;
     final boolean _verifyExisting;
 
-    final Set<String> _pkKeysSeen = new HashSet<>();
+    final Set<Object> _pkKeysSeen = new HashSet<>();
 
     final DataIteratorContext _context;
 
@@ -267,10 +267,9 @@ public abstract class ExistingRecordDataIterator extends WrapperDataIterator
                     pkKeys.add(pkVal.toString());
                 }
                 sqlf.append(")");
-                String pkKey = StringUtils.join(pkKeys, ", ");
-                if (_pkKeysSeen.contains(pkKey))
-                    _context.getErrors().addRowError(new ValidationException("Duplicate key provided: " + pkKey));
-                _pkKeysSeen.add(pkKey);
+                if (_pkKeysSeen.contains(pkKeys))
+                    _context.getErrors().addRowError(new ValidationException("Duplicate key provided: " + StringUtils.join(pkKeys, ", ")));
+                _pkKeysSeen.add(pkKeys);
                 rowNumContainers.put(lastPrefetchRowNumber, container);
             }
             while (--rows > 0 && _delegate.next());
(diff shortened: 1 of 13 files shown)


api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -255,12 +258,19 @@ private Pair<SQLFragment, Map<Integer, String>> getSelectExistingSql(int rows) t
...
                 for (int p = 0; p < pkColumns.size(); p++)
                 {
                     sqlf.append(",?");
-                    sqlf.add(pkSuppliers.get(p).get());
+                    Object pkVal = pkSuppliers.get(p).get();
+                    sqlf.add(pkVal);
+                    pkKeys.add(pkVal.toString());
                 }
                 sqlf.append(")");
+                String pkKey = StringUtils.join(pkKeys, ", ");
```
labkey-susanh: Joining on commas. Are we sure this is safe and we won't have overlap from keys called "a, b" and "other" and two keys called "a" and "bother"?

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -370,8 +380,19 @@ protected void prefetchExisting() throws BatchValidationException
...
+                    List<String> pkKeys = new ArrayList<>();
                     for (int p=0 ; p<pkColumns.size() ; p++)
-                        keyMap.put(pkColumns.get(p).getColumnName(), pkSuppliers.get(p).get());
+                    {
+                        Object pkVal = pkSuppliers.get(p).get();
+                        keyMap.put(pkColumns.get(p).getColumnName(), pkVal);
+                        pkKeys.add(pkVal.toString());
+                    }
+
+                    String pkKey = StringUtils.join(pkKeys, ", ");
```
labkey-susanh: Same question here.

specimen/src/org/labkey/specimen/query/SpecimenUpdateService.java
```diff
@@ -325,7 +325,7 @@ protected Map<String, Object> insertRow(User user, Container container, Map<Stri
 
     @Override
     public List<Map<String, Object>> updateRows(User user, Container container, List<Map<String, Object>> rows, List<Map<String, Object>> oldKeys, BatchValidationException errors, @Nullable Map<Enum, Object> configParameters, Map<String, Object> extraScriptContext)
-            throws InvalidKeyException, BatchValidationException, QueryUpdateServiceException, SQLException
+            throws InvalidKeyException, BatchValidationException, QueryUpdateServiceException, SQLException // TODO
```
labkey-susanh: Still TODO?

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
XingY: good point. Updated.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -834,6 +843,43 @@ public List<Map<String, Object>> updateRows(User user, Container container, List
...
+            List<Object> orderedKeyVals = new ArrayList<>();
+            SortedSet<String> sortedKeys = new TreeSet<>(map.keySet());
+            for (String key : sortedKeys)
+                orderedKeyVals.add(map.get(key));
+            keysObj = orderedKeyVals.toArray();
+        }
+        else
+            keysObj = new Object[]{pkVals};
+
+        if (keysObj.length == 1)
```
labkey-susanh, line 867: Would this and other logic be simpler if `updatedRows` always had a collection (List, Set, Array, etc.) instead of sometimes a collection and sometimes not?

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -1288,6 +1334,15 @@ public void addRowError(ValidationException vex)
             // test new row
             assertEquals("THREE", rows.get(3).get("s"));
             assertNull(rows.get(3).get("i"));
+
+            // merge should fail if duplicate keys are provided
```
labkey-susanh, line 1338: Issue reference here?

experiment/src/client/test/integration/DataClassCrud.ispec.ts
```diff
@@ -235,3 +235,107 @@ describe('Data Class - Required Lineage', () => {
     });
 
 });
+
+describe('CRUD actions', () => {
```
labkey-susanh: Seems like it would be better to give a more specific name here, like 'Duplicate IDs'

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -170,12 +177,19 @@ public boolean next() throws BatchValidationException
...
-            if (_context.getErrors().hasErrors())
+            if (_context.shouldCancel())
                 return false;
         }
         return ret;
     }
 
+    protected void checkDuplicateKeys(List<String> pkKeys)
+    {
+        Object pkKeysObj = pkKeys.size() == 1 ? pkKeys.get(0) : pkKeys;
```
labkey-susanh, line 188: Why not just always store the list, even if a singleton list?

XingY: This is for perf/memory concern. Most data type will have a single PK. With large import, the keys to hold in memory could get quite large. Also, the compare operation would be more effective with single object.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
XingY: The idea here since there is a single key most of the time, there is no need to use a more expensive data structure unnecessarily.

labkey-susanh: Ok, that's what I figured. Just wanted to check.

labkey-susanh (approved):
Looks like there's still a lingering TODO in `SpecimenUpdateService`. Is that meant for this PR? I'll approve, but let me know if you want re-review for any change there (not that I know much about that service).

XingY:
> Looks like there's still a lingering TODO in `SpecimenUpdateService`. Is that meant for this PR? I'll approve, but let me know if you want re-review for any change there (not that I know much about that service).

I've removed that stale TODO. thanks for pointing out.

PR #6578, opened 2025-04-17
Mechanism to limit the JSON length as input to APIs

labkey-jeckels:
#### Rationale
Accepting massive JSON inputs can be bad for server health.

#### Changes
- New annotation option for limiting the maximum length of JSON

### DIFF
--- a/api/src/org/labkey/api/reader/StrictBoundedReader.java
+++ b/api/src/org/labkey/api/reader/StrictBoundedReader.java
@@ -1,0 +1,99 @@
+package org.labkey.api.reader;
+
+import java.io.IOException;
+import java.io.Reader;
+
+/**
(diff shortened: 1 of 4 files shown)


api/src/org/labkey/api/reader/StrictBoundedReader.java
```diff
@@ -0,0 +1,99 @@
+package org.labkey.api.reader;
+
+import java.io.IOException;
+import java.io.Reader;
+
+/**
+ * Restricts the Reader to a character limit. Throws an exception when the limit is exceeded.
+ * Similar to Apache's BoundedReader, which silently truncates.
```
labkey-adam, line 8: Too bad about BoundedReader

labkey-adam (approved):

PR #6580, opened 2025-04-17
Issue 52827: File/attachment fields with special characters

XingY:
#### Rationale
- "multipart/form-data" is used when submitting a html form with file, instead of "application/x-www-form-urlencoded". When submitted as form, form parameters with " are converted to %22 since " is a special character for URL. This is resulting in the parameter being ignored since it cannot be matched to an existing table column.
- :: is used as delimiter for file parameters. The logic to extract parts from the delimiter is flawed in that it doesn't account for fields that contains such substring. 

#### Related Pull Requests
- 

#### Changes
- Added MatchType.mutiPartFormData to relax column matching with encoded "
- fix logic to extract field name from file parameters

### DIFF
--- a/api/src/org/labkey/api/data/TableViewForm.java
+++ b/api/src/org/labkey/api/data/TableViewForm.java
@@ -578,7 +578,7 @@ public class TableViewForm extends ViewForm implements DynaBean, HasBindParamete
             {
                 String quoteEncodedFieldName = getMultiPartFormFieldName(column);
                 boolean isFileColFileRemoved = false;
-                if (values.get(column.getName()) == null && File.class.equals(column.getJavaClass()) && getRequest() instanceof MultipartHttpServletRequest)
+                if (values.get(column.getName()) == null && File.class.equals(column.getJavaClass()))
                 {
                     MultipartHttpServletRequest request = (MultipartHttpServletRequest) getRequest();
                     MultipartFile f = request.getFile(quoteEncodedFieldName);
--- a/api/src/org/labkey/api/data/TableViewForm.java
+++ b/api/src/org/labkey/api/data/TableViewForm.java
@@ -34,6 +34,8 @@ import org.labkey.api.action.HasBindParameters;
 import org.labkey.api.action.NullSafeBindException;
 import org.labkey.api.action.SpringActionController;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
+import org.labkey.api.dataiterator.DataIteratorUtil;
+import org.labkey.api.query.QueryUpdateForm;
 import org.labkey.api.query.SchemaKey;
 import org.labkey.api.security.permissions.DeletePermission;
 import org.labkey.api.security.permissions.InsertPermission;
@@ -572,6 +574,14 @@ public class TableViewForm extends ViewForm implements DynaBean, HasBindParamete
                 values.put(column.getName(), getTypedValue(column));
             else if (includeUntyped && contains(column))
                 values.put(column.getName(), get(column));
+            else if (column.getName().contains("\""))
+            {
+                // TODO: only match when request is of "multipart/form-data" content type?
+                String quoteEncodedName = DataIteratorUtil.MatchType.mutiPartFormData.getMatchedName(column.getName());
+                Object value = getTypedValues().get(QueryUpdateForm.PREFIX + quoteEncodedName);
+                if (value != null)
+                    values.put(column.getName(), value);
+            }
 
             // Check if there was a file uploaded for the column's value
             if (values.get(column.getName()) == null && File.class.equals(column.getJavaClass()) && getRequest() instanceof MultipartHttpServletRequest)
--- a/api/src/org/labkey/api/data/TableViewForm.java
+++ b/api/src/org/labkey/api/data/TableViewForm.java
@@ -560,22 +560,6 @@ public class TableViewForm extends ViewForm implements DynaBean, HasBindParamete
         return _values;
     }
 
-    private boolean setFileColumn(ColumnInfo column, String formFieldName, Map<String, Object> values)
-    {
-        if (values.get(column.getName()) == null && File.class.equals(column.getJavaClass()) && getRequest() instanceof MultipartHttpServletRequest)
-        {
-            MultipartHttpServletRequest request = (MultipartHttpServletRequest) getRequest();
-            MultipartFile f = request.getFile(formFieldName);
-            // Only set the parameter value if there was a form element that was posted
-            if (f != null)
-            {
-                values.put(column.getName(), f.getOriginalFilename() == null || f.getOriginalFilename().isEmpty() ? null : f);
-                return true;
-            }
-        }
-        return false;
-    }
-
     /**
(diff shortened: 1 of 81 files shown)


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -4643,26 +4643,34 @@ private void addRowAttachments(Map<String, Object> rowMap, int rowIndex, @Nullab
...
-                            if (!fieldCommandIndex.equals(commandIndex+"") || !fieldRowIndex.equals(rowIndex+""))
+                            int subDelimIndex = fieldKeyExcludeIndex.lastIndexOf(ROW_ATTACHMENT_INDEX_DELIM);
+                            if (subDelimIndex > -1)
+                            {
+                                fieldKey = fieldKeyExcludeIndex.substring(0, subDelimIndex);
+                                String fieldCommandIndex = fieldKeyExcludeIndex.substring(subDelimIndex + ROW_ATTACHMENT_INDEX_DELIM.length());
+                                if (!fieldCommandIndex.equals(commandIndex+""))
+                                    continue;
+                            }
+                            else
```
labkey-matthewb, line 4682: I think the correct fix is to remove the encoded characters here (early) rather than trying to match them later.  Ideally the encoded chars shouldn't "leak" out of code that is pulling values out of the request object.

Then we wouldn't have to handle these characters later.

labkey-matthewb (changes requested):

query/src/org/labkey/query/controllers/QueryController.java
XingY: The challenge here is that a field might actually contain "%22". Both field" and field%22 will come in as field%22 and we don't know when to decode vs not here.

labkey-matthewb: 🤯

labkey-matthewb: This is really an HTTP bug (it's not OK to escape " to %22 without escaping %).  So this would break any column with " in its name in a type with a file column then.  Not just the file column.

Since these columns could never map to a java property.  I'd propose changing BaseColumnInfo.getPropertyName() to replace **"** with **%_QUOTE_%** or something.  That should solve the problem by never using the " in the input name.  That would fix the dataregion case, anyway.  I'm not sure what the editable grid does.

api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
```diff
@@ -141,7 +141,31 @@ public static Map<String,ColumnInfo> createTableMap(TableInfo target, boolean us
...
+        jdbcname,
+        tsvColumn,
+        mutiPartFormData()
+                {
+                    @Override
+                    public String getMatchedName(@Nullable String name)
+                    {
+                        if (name == null)
+                            return null;
+                        // " is encoded as %22 when content-type is "multipart/form-data"
```
labkey-matthewb:
```suggestion
                        // " is encoded as %22 when content-type is "multipart/form-data" (but is not otherwise encoded so decode() does not work)
```

labkey-matthewb (approved):
We can use your approach, but I would add a longer explanation about the problem with the form encoding.  We might also consider removing double-quote from generated property name.  BaseColumnInfo.getPropertyName().

query/src/org/labkey/query/controllers/QueryController.java
XingY: Editable grid in apps currently doesn't support file fields. The paths supported in Apps are editing detail editing and bulk ediitng. Detail editing uses `updateRows` api with formData (file, plus json data for other row data) while bulk update uses `saveRows`. For Apps, only the file field is impacted since other fields are in json. This approach of replace " with %QUOTE% could work for Apps too, but it's also more error prone. Also, it's hard to pick a substitute that absolutely won't collide with real field names, a field might just actually contain `%QUOTE%`.

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -572,6 +574,26 @@ public Map<String,Object> getTypedColumns(boolean includeUntyped)
                 values.put(column.getName(), getTypedValue(column));
             else if (includeUntyped && contains(column))
                 values.put(column.getName(), get(column));
+            else if (column.getName().contains("\"") && getViewContext().getRequest() instanceof MultipartHttpServletRequest)
+            {
+                String quoteEncodedFieldName = getMultiPartFormFieldName(column);
+                boolean isFileColFileRemoved = false;
+                if (values.get(column.getName()) == null && File.class.equals(column.getJavaClass()) && getRequest() instanceof MultipartHttpServletRequest)
```
cnathe: didn't you already check for `getRequest() instanceof MultipartHttpServletRequest` at line 577 above?

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -572,6 +574,26 @@ public Map<String,Object> getTypedColumns(boolean includeUntyped)
...
+                }
+
+                if (isFileColFileRemoved)
+                    values.put(column.getName(), null);
+                else
+                {
+                    Object value = getTypedValues().get(quoteEncodedFieldName);
+                    if (value != null)
+                        values.put(column.getName(), value);
+                }
```
cnathe, line 595: this seems like a lot of duplicate code from what is already there below for handling the non-slash column name case (lines 599-607). it seems like the main part is the field name to use in `request.getFile()`. Could these blocks of code be collapsed and have the conditional part be using either getMultiPartFormFieldName or getFormFieldName? Then after we have the MultipartFile, the rest of the code seems pretty similar.

api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
```diff
@@ -141,7 +141,60 @@ public static Map<String,ColumnInfo> createTableMap(TableInfo target, boolean us
...
+                        }
+                        return false;
+                    }
+                },
+        low;
+
+        public String getMatchedName(@Nullable String name)
+        {
+            return name;
+        }
```
cnathe, line 184: I'm confused where this getMatchedName() is used. Couldn't that usage just use the name property directly?

cnathe: or is this overridden somewhere that I am just not seeing?

XingY: It's overriden in multiPartFormData

api/src/org/labkey/api/data/TableViewForm.java
XingY: This was how it was written in a previous commit, but that broke LKS attachment update. The different part is what value to use for file field: string or File. The only shared code is the following 2 lines so I removed the shared util in the last commit.
`
MultipartHttpServletRequest request = (MultipartHttpServletRequest) getRequest();
MultipartFile f = request.getFile(quoteEncodedFieldName);
`

api/src/org/labkey/api/data/TableViewForm.java
XingY: Good point. Removed.

PR #6587, opened 2025-04-18
Remove schema export

labkey-klum:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52620

Delete obsolete code including the `bulkImport` support in SQL.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/6587
https://github.com/LabKey/premiumModules/pull/250


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -7651,163 +7748,6 @@ public Object execute(NamedSetForm namedSetForm, BindException errors)
...
-        @Override
-        public void addNavTrail(NavTree root)
-        {
-            root.addChild("Generate Schema");
-        }
-    }
-
-
     @RequiresPermission(AdminPermission.class)
     public static class GetSchemasWithDataSourcesAction extends ReadOnlyApiAction<Object>
```
labkey-adam: Should we delete this action as well? I think it's used only by the jsp that you just removed.

labkey-adam (changes requested):
We should eliminate the other side of this, the "bulkImport" option in SQL scripts. That means `BulkImportBlock` and the "bulkImport" parts of the primary dialect's PROC_PATTERN regexes. Should also drop the `core.bulkImport` stored procedure. I can help, if you want.

labkey-adam (approved):

PR #6591, opened 2025-04-21
Issue 52506 and 52507: add rowId to source detail audit record and throw error if importing into a sibling container

labkey-susanh:
#### Rationale
- Issue 52506 - [LKSM: Issues with importing into a different container than your current context](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52506)
- Issue 52507 -[LKSM: Success after updating Sources via file prompts to select them but doesn't select](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52507)

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1349

#### Changes
- Include RowId as one of the detailed audit record fields so the "select them in the grid" functionality, that filters using RowId will work
- If importing or updating data into a sibling container, throw an error

### DIFF
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -2667,6 +2667,11 @@ public class ExpDataIterators
                                 _context.getErrors().addRowError(new ValidationException("Invalid value '" + rowFolderId +"' provided for '" + getColumnInfo(_folderColIndex).getName() + "'."));
                                 return true;
                             }
+                            else if (_container.getEntityId() != targetContainer.getEntityId() && !_container.hasAncestor(targetContainer) && !targetContainer.hasAncestor(_container))
+                            {
+                                _context.getErrors().addRowError(new ValidationException("Import or update from folder " + _container.getName() + " into folder " + targetContainer.getName() + " not allowed."));
+                                return true;
+                            }
                         }
                     }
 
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -2667,6 +2667,11 @@ public boolean next() throws BatchValidationException
                                 _context.getErrors().addRowError(new ValidationException("Invalid value '" + rowFolderId +"' provided for '" + getColumnInfo(_folderColIndex).getName() + "'."));
                                 return true;
                             }
+                            else if (_container.getEntityId() != targetContainer.getEntityId() && !_container.hasAncestor(targetContainer) && !targetContainer.hasAncestor(_container))
```
XingY: Does this check need to respect the 2 experimental flag: "Less restrictive product folder lookups" and "Product folders display folder-specific data"?

XingY (approved):

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: Also, for "update from file" without container column provided, do we need to check that the import file doesn't include data from sibling?

labkey-susanh: I think not for the folder lookups since this isn't a lookup, but probably so for the other two cases. I'm thinking the `_containerMap` initialization needs to change.

PR #6592, opened 2025-04-21
The new "Delete All" button needs a type parameter

labkey-adam:
#### Rationale
Probably shouldn't delete all the "allowed redirect hosts" when admin attempts to delete "allowed file extensions"

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6588


labkey-jeckels (approved):

labkey-sweta (approved):
Verified. Delete all works as expected.

PR #6596, opened 2025-04-22
Avoid reflecting invalid URL parameter - prevent false positive

labkey-jeckels:
#### Rationale
A security scanner is tripping on a reflected, properly encoded URL parameter value. It's not especially helpful for users so we can just avoid including it on the 404.

#### Changes
- Simplify user-facing error message
- Eliminate long-unused HTTP attribute
- Improve error logging in DavController

### DIFF
--- a/filecontent/src/org/labkey/filecontent/FileContentController.java
+++ b/filecontent/src/org/labkey/filecontent/FileContentController.java
@@ -1466,8 +1466,14 @@ public class FileContentController extends SpringActionController
            if (null == renderAs)
                return FileContentController.RenderStyle.PAGE;
 
-           //Will throw illegal argument exception for other values...
-           return FileContentController.RenderStyle.valueOf(renderAs.toUpperCase());
+           try
+           {
+               return FileContentController.RenderStyle.valueOf(renderAs.toUpperCase());
+           }
+           catch (IllegalArgumentException e)
+           {
+               throw new NotFoundException("Invalid renderAs value");
+           }
        }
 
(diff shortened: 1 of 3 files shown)


filecontent/src/org/labkey/filecontent/FileContentController.java
```diff
@@ -1466,8 +1466,14 @@ public RenderStyle getRenderStyle()
...
-           //Will throw illegal argument exception for other values...
-           return FileContentController.RenderStyle.valueOf(renderAs.toUpperCase());
+           try
+           {
+               return FileContentController.RenderStyle.valueOf(renderAs.toUpperCase());
+           }
+           catch (IllegalArgumentException e)
+           {
+               throw new NotFoundException("Invalid renderAs value");
+           }
```
labkey-adam, line 1476: Since PAGE appears to be the default, you could default to that for invalid value as well:
```
EnumUtils.getEnum(FileContentController.RenderStyle.class, renderAs.toUpperCase(), FileContentController.RenderStyle.PAGE);
```

labkey-adam (approved):

filecontent/src/org/labkey/filecontent/FileContentController.java
labkey-jeckels: I considered that but opted to be a little stricter. I don't think we've seen this exception aside from the crawler/scanner so I don't think we need to be overly tolerant.

Good to know about EnumUtils though - that wasn't on my radar.

PR #6598, opened 2025-04-23
Issue 52592: resolve correct unassigned samples folder column

labkey-nicka:
#### Rationale
See https://github.com/LabKey/limsModules/pull/1358 for rationale.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/740
- https://github.com/LabKey/limsModules/pull/1358
- https://github.com/LabKey/platform/pull/6598

#### Changes
- Add `ExpMaterialTable.Column.fieldKey()` utility

### DIFF
--- a/api/src/org/labkey/api/exp/query/ExpMaterialTable.java
+++ b/api/src/org/labkey/api/exp/query/ExpMaterialTable.java
@@ -17,6 +17,7 @@
 package org.labkey.api.exp.query;
 
 import org.labkey.api.data.UpdateableTableInfo;
+import org.labkey.api.query.FieldKey;
 
 public interface ExpMaterialTable extends ExpTable<ExpMaterialTable.Column>, UpdateableTableInfo
 {
@@ -60,7 +61,12 @@ public interface ExpMaterialTable extends ExpTable<ExpMaterialTable.Column>, Upd
         SourceProtocolLSID,
         StoredAmount,
         Units,
-        IsPlated,
+        IsPlated;
+
+        public FieldKey fieldKey()
+        {
+            return FieldKey.fromParts(name());
+        }
     }
 
     default void setSupportTableRules(boolean supportTableRules)


XingY (commented):
Not sure how I feel about this... a bit overkill?

labkey-nicka:
> Not sure how I feel about this... a bit overkill?

How so?

XingY:
> > Not sure how I feel about this... a bit overkill?
> 
> How so?

I understand the purpose is to enforcing the correct column is used. But still, this is a pretty straightforward util that's only used by that Inventory table. Could we instead have this util in InventoryQuerySchema?
`
private FieldKey getMaterialFieldKey(ExpMaterialTable.Column materialCol)
    {
        return FieldKey.fromParts(materialCol.name());
    }
`

labkey-nicka:
> I understand the purpose is to enforcing the correct column is used. But still, this is a pretty straightforward util that's only used by that Inventory table. Could we instead have this util in InventoryQuerySchema? `private FieldKey getMaterialFieldKey(ExpMaterialTable.Column materialCol) { return FieldKey.fromParts(materialCol.name()); }`

I see. What I'm doing here is just extending a notion that is already in place for other column enumerations. Here are a few examples:

- https://github.com/LabKey/platform/blob/2581ee4a7a/api/src/org/labkey/api/inventory/InventoryService.java#L80
- https://github.com/LabKey/platform/blob/2581ee4a7a/api/src/org/labkey/api/exp/query/ExpSampleTypeTable.java#L61
- https://github.com/LabKey/platform/blob/2581ee4a7a/assay/src/org/labkey/assay/plate/query/WellTable.java#L90
- https://github.com/LabKey/platform/blob/2581ee4a7a/assay/src/org/labkey/assay/plate/query/WellTable.java#L90

This utility could already apply to many usages in platform but I didn't update those here since it's not particularly relevant.

XingY (approved):

PR #6600, opened 2025-04-23
Issue 52570: Assay run move to include source and target containers in the audit event message

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52570

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1361

#### Changes
- include source and target containers in the audit event message


XingY (approved):

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -433,7 +439,8 @@ public void auditRunEvent(User user, ExpProtocol protocol, ExpRun run, @Nullable
         if (run != null)
             event.setRunLsid(run.getLSID());
         event.setProtocolRun(ExperimentAuditProvider.getKey3(protocol, run));
-
+        if (message != null)
```
labkey-susanh, line 442: Don't need the null check here, but no harm.

labkey-susanh (approved):

PR #6606, opened 2025-04-24
Address 52916, and fix allowed extensions list update process

Sigmonia:
#### Rationale
[Issue 52916:](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52916) Allowed file extension: Save is behaving incorrectly. 

#### Changes
- Fixed order of operations, between property save and rebuilding the regex
- Added a JUnit test case


labkey-adam (approved):
Thanks for the junit test!

PR #6607, opened 2025-04-25
Conditionalize option 'Fail import for undefined visits' during folder import

labkey-bpatel:
#### Rationale
The 'Fail import for undefined visits' option is displayed by default on the folder import page in Folder Management, regardless of whether the import is for a Study or a non-Study folder. It's a no-op for non-Study, however, annoying to see. This change was introduced as part of the 'Deprecate advanced import options' - see Related PR.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6419

#### Changes
- Conditionalize option 'Fail import for undefined visits' -- logic to display or not to display this option is the same as in startPipelineImport.jsp (~line 100).


labkey-adam (commented):
@labkey-bpatel I'd like to discuss before reviewing and your merging this PR. Since any folder archive could create a study in a currently non-study folder (and also define visits), it's not clear to me the conditions under which we would suppress this checkbox. I'm also not clear the intent reading the code. Maybe best to chat about this.

PR #6609, opened 2025-04-28
Misc 25.5 assay result issue fixes (assay result delete rows issue 51126, assay unresolved sample lookup issue 52745)

cnathe:
#### Rationale
Issue [51126](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51126): AssayResultUpdateService to call syncLineage() on assay result delete rows when sample lookup column exists
- The related PR fixed the issue where assay run / sample lineage information was resynced after rows for the assay result table were updated (which might have changed a sample lookup value). This PR fixes the related case where we need to resync the lineage info for an assay run / sample after a row is delete from an assay result for a run.

Issue [52745](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52745): The values of some sample fields can successfully be added to assay results grid even if user does not have permissions to view the sample
- This PR adds SQL null checks for the ExprColumn cases for sample lookups so that we don't show the "default" value for the expression if there is no matching sample for the rowid (i.e. unresolved sample)

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5903
- https://github.com/LabKey/platform/pull/6609
- https://github.com/LabKey/limsModules/pull/1372

#### Changes
- AssayResultUpdateService to call syncLineage() on assay result delete rows when sample lookup column exists
- ExprColumn null check for various sample lookup fields (IsAliquot, IsPlated, StorageStatus)

### DIFF
--- a/assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
+++ b/assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
@@ -387,7 +387,7 @@ public class AssayResultUpdateService extends DefaultQueryUpdateService
             for (DomainProperty dp : datatableInfo.getDomain().getNonBaseProperties())
             {
                 if (AssaySampleLookupContext.checkSampleLookup(container, user, dp).isLookup())
-                    _assaySampleLookupContext.trackSampleLookupChange(container, user, datatableInfo, datatableInfo.getColumn("SampleId"), run);
+                    _assaySampleLookupContext.trackSampleLookupChange(container, user, datatableInfo, datatableInfo.getColumn(dp.getName()), run);
             }
         }
 
@@ -400,6 +400,7 @@ public class AssayResultUpdateService extends DefaultQueryUpdateService
         var result = super.deleteRows(user, container, keys, configParameters, extraScriptContext);
 
         BatchValidationException errors = new BatchValidationException();
+        errors.setExtraContext(extraScriptContext);
         _assaySampleLookupContext.syncLineage(container, user, errors);
         if (errors.hasErrors())
             throw errors;
--- a/assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
+++ b/assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
@@ -381,6 +381,29 @@ public class AssayResultUpdateService extends DefaultQueryUpdateService
             }
         }
 
+        // Issue 51126: need to track and resync run/sample lineage on delete in the same way we do for update
+        if (datatableInfo.getDomain() != null)
+        {
+            for (DomainProperty dp : datatableInfo.getDomain().getNonBaseProperties())
+            {
+                if (AssaySampleLookupContext.checkSampleLookup(container, user, dp).isLookup())
+                    _assaySampleLookupContext.trackSampleLookupChange(container, user, datatableInfo, datatableInfo.getColumn("SampleId"), run);
+            }
+        }
+
+        return result;
+    }
+
+    @Override
+    public List<Map<String, Object>> deleteRows(User user, Container container, List<Map<String, Object>> keys, @Nullable Map<Enum, Object> configParameters, @Nullable Map<String, Object> extraScriptContext) throws InvalidKeyException, BatchValidationException, QueryUpdateServiceException, SQLException
+    {
+        var result = super.deleteRows(user, container, keys, configParameters, extraScriptContext);
+
+        BatchValidationException errors = new BatchValidationException();
+        _assaySampleLookupContext.syncLineage(container, user, errors);
+        if (errors.hasErrors())
+            throw errors;
+
         return result;
     }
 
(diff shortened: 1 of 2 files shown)


assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
```diff
@@ -381,6 +381,29 @@ protected Map<String, Object> deleteRow(
...
+
+        return result;
+    }
+
+    @Override
+    public List<Map<String, Object>> deleteRows(User user, Container container, List<Map<String, Object>> keys, @Nullable Map<Enum, Object> configParameters, @Nullable Map<String, Object> extraScriptContext) throws InvalidKeyException, BatchValidationException, QueryUpdateServiceException, SQLException
+    {
+        var result = super.deleteRows(user, container, keys, configParameters, extraScriptContext);
+
+        BatchValidationException errors = new BatchValidationException();
```
labkey-nicka, line 399: Additionally, set the extra context here for errors.

```java
errors.setExtraContext(extraScriptContext);
```

assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
```diff
@@ -381,6 +381,29 @@ protected Map<String, Object> deleteRow(
             }
         }
 
+        // Issue 51126: need to track and resync run/sample lineage on delete in the same way we do for update
+        if (datatableInfo.getDomain() != null)
+        {
+            for (DomainProperty dp : datatableInfo.getDomain().getNonBaseProperties())
+            {
+                if (AssaySampleLookupContext.checkSampleLookup(container, user, dp).isLookup())
+                    _assaySampleLookupContext.trackSampleLookupChange(container, user, datatableInfo, datatableInfo.getColumn("SampleId"), run);
```
labkey-nicka: We support ad-hoc sample lookups. Should this be referring to the underlying column for this domain property rather than hard-coded to `SampleId`?

cnathe: shoot, yes. I had hard coded in "SampleId" there for testing locally. thanks. I fix that

assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
cnathe: done

assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
cnathe: done

XingY (changes requested):
I believe some optimization can be done here. deleteRows already calls deleteRow. Could we instead set a flag in deleteRows to signal updateRow to not do the sync yet if it's a batch delete?

cnathe:
> I believe some optimization can be done here. deleteRows already calls deleteRow. Could we instead set a flag in deleteRows to signal updateRow to not do the sync yet if it's a batch delete?

@XingY I'm not sure I'm following the suggestion here. What do you mean by having deleteRows set a flag for updateRow? Do you mean deleteRow? Currently the code will use the AssaySampleLookupContext to keep track of the list of runs affected (this happens during deleteRow) so that after the batch is deleted we can call syncLineage once (this is done in deleteRows).

XingY:
> > I believe some optimization can be done here. deleteRows already calls deleteRow. Could we instead set a flag in deleteRows to signal updateRow to not do the sync yet if it's a batch delete?
> 
> @XingY I'm not sure I'm following the suggestion here. What do you mean by having deleteRows set a flag for updateRow? Do you mean deleteRow? Currently the code will use the AssaySampleLookupContext to keep track of the list of runs affected (this happens during deleteRow) so that after the batch is deleted we can call syncLineage once (this is done in deleteRows).

OK, nvm, I missed the fact that `deleteRow` is a protected method so it's unlikely to get called outside of `deleteRows`. I thought the sync needed to be done for both code path.

XingY (approved):

assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
```diff
@@ -381,6 +381,30 @@ protected Map<String, Object> deleteRow(
             }
         }
 
+        // Issue 51126: need to track and resync run/sample lineage on delete in the same way we do for update
+        if (datatableInfo.getDomain() != null)
+        {
+            for (DomainProperty dp : datatableInfo.getDomain().getNonBaseProperties())
+            {
+                if (AssaySampleLookupContext.checkSampleLookup(container, user, dp).isLookup())
```
labkey-nicka: `trackSampleLookupChange()` already does this work of checking the sample lookup and it retains it. Consider just calling:

```java
for (DomainProperty dp : datatableInfo.getDomain().getNonBaseProperties())
{
    _assaySampleLookupContext.trackSampleLookupChange(container, user, datatableInfo, datatableInfo.getColumn(dp.getName()), run);
}
```

cnathe: I considered that, but it seemed odd that the `_sampleLookups` would keep track of all columns in the table instead of just the ones that are actually sample lookups. But I suppose it is the same number of checks either way so it seems fine to remove the if statement I have her and just defer that to the one in `trackSampleLookupChange`. Done.

labkey-nicka: Right, the idea is that different parts of the code may need to process a column more than once since so the `AssaySampleLookupContext` holds onto that context.

labkey-nicka (approved):

PR #6612, opened 2025-04-29
Issue 52963: CSP template is not always being fully substituted

labkey-jeckels:
#### Rationale
We can't rely on registrations for CSP directives to correctly substitute the CSP template

#### Changes
- Fix caching to avoid reusing StringExpressions that want different behavior
- Substitute with blanks in the CSP

### DIFF
--- a/api/src/org/labkey/api/util/StringExpressionFactory.java
+++ b/api/src/org/labkey/api/util/StringExpressionFactory.java
@@ -1329,17 +1329,22 @@ public class StringExpressionFactory
             Map<Object,Object> m = new HashMap<>();
 
             FieldKeyStringExpression original = new FieldKeyStringExpression("details.view?id=${rowid}&title=${title}");
-            m.put(FieldKey.fromParts("lookup","rowid"), "BUG");
+            m.put(FieldKey.fromParts("lookup","rowid"), 6);
             m.put(FieldKey.fromParts("lookup"), 5);
             m.put(FieldKey.fromParts("lookup","title"), "title one");
+
+            // Approach #1 - prefix explicitly, column by column
             Map<FieldKey,FieldKey> remap = new HashMap<>();
             remap.put(FieldKey.fromParts("rowid"), FieldKey.fromParts("lookup"));
             remap.put(FieldKey.fromParts("title"), FieldKey.fromParts("lookup", "title"));
-
             FieldKeyStringExpression remapped = original.remapFieldKeys(null, remap);
-
             assertNull(original.eval(m));
             assertEquals("details.view?id=5&title=title%20one", remapped.eval(m));
+
+            // Approach #2 - remap everything with a prefix
+            remapped = original.remapFieldKeys(FieldKey.fromParts("lookup"), null);
+            assertNull(original.eval(m));
+            assertEquals("details.view?id=6&title=title%20one", remapped.eval(m));
         }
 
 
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/filters/ContentSecurityPolicyFilter.java
```diff
@@ -209,7 +209,7 @@ private void regeneratePolicyExpression()
                 .eval(ALLOWED_SOURCES_SUBSTITUTION_MAP);
         }
 
-        _policyExpression = StringExpressionFactory.create(allowSubstitutedPolicy, false, NullValueBehavior.KeepSubstitution);
+        _policyExpression = StringExpressionFactory.create(allowSubstitutedPolicy, false, NullValueBehavior.ReplaceNullAndMissingWithBlank);
```
labkey-adam, line 212: Sure enough, looks like this was switched in a recent refactor. I do find it odd that the behavior is specified at create() time as opposed to eval() time, but no excuses...

labkey-adam (approved):

labkey-jeckels:
@labkey-adam It's worth a quick review of the test update and addition I just pushed if you have a moment.

api/src/org/labkey/api/util/StringExpressionFactory.java
```diff
@@ -1341,12 +1328,15 @@ public void testFieldKey()
...
-            FieldKeyStringExpression fkse = new FieldKeyStringExpression("details.view?id=${rowid}&title=${title}");
+            FieldKeyStringExpression e = new FieldKeyStringExpression("details.view?id=${rowid}&title=${title}");
             m.put(FieldKey.fromParts("A","rowid"), "BUG");
             m.put(new FieldKey(null, "lookup"), 5);
             m.put(FieldKey.fromParts("A","title"), "title one");
             Map<FieldKey,FieldKey> remap = new HashMap<>();
             remap.put(new FieldKey(null,"rowid"), new FieldKey(null,"lookup"));
+
+            e.remapFieldKeys(null, remap);
+            assertEquals("whoknows", e.eval(m));
```
labkey-adam: @labkey-jeckels this line is failing for me (actual eval() return is `null`)

labkey-adam:
> @labkey-adam It's worth a quick review of the test update and addition I just pushed if you have a moment.

@labkey-jeckels looks good to me except for the one failing test case (`testFieldKey()`)

api/src/org/labkey/api/util/StringExpressionFactory.java
```diff
@@ -1341,12 +1328,18 @@ public void testFieldKey()
         {
             Map<Object,Object> m = new HashMap<>();
 
-            FieldKeyStringExpression fkse = new FieldKeyStringExpression("details.view?id=${rowid}&title=${title}");
-            m.put(FieldKey.fromParts("A","rowid"), "BUG");
-            m.put(new FieldKey(null, "lookup"), 5);
-            m.put(FieldKey.fromParts("A","title"), "title one");
+            FieldKeyStringExpression original = new FieldKeyStringExpression("details.view?id=${rowid}&title=${title}");
```
labkey-jeckels, line 1331: @labkey-matthewb requesting your review on this test case

labkey-matthewb: The more common usage is:
FieldKeyStringExpression remapped = original.remapFieldKeys(FieldKey.fromParts("lookup"), null);

Might be worth adding that as well, if it is not covered.

labkey-jeckels: @labkey-matthewb Second version added. I think it's working as expected, but the fact that the original stub for the test had a value of `"BUG"` makes me wonder.

PR #6614, opened 2025-04-29
Issue 52504: Improve messaging for lookup resolution that fails

labkey-susanh:
#### Rationale
Issue [52504](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52504): We are adding the option to validate lookups for the Sample field. The current messaging ("Could not convert value 'Blood-246983' (String) for Integer field 'SampleLookup'") when we fail to find a value for this lookup is not something the typical user will be able to interpret easily since, for one thing, they don't know this is an integer lookup.  By using a `ConversionExceptionWithMessage` within `RemapPostConvert` when we have the field name, we can bubble up an error message that is perhaps more useful.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1376
- https://github.com/LabKey/labkey-ui-components/pull/1783

#### Changes
- Update `RemapPostConvert` to take an optional `fieldName` to use in messaging


experiment/src/org/labkey/experiment/controllers/exp/ExperimentController.java
```diff
@@ -4529,6 +4529,8 @@ public void validateForm(QueryForm form, Errors errors)
         protected void initRequest(QueryForm form) throws ServletException
         {
             QueryDefinition query = form.getQueryDef();
+            // Issue 52504: For lookup validation, we need to use the proper lookup container filter on the table
+            query.setContainerFilter(QueryService.get().getContainerFilterForLookups(getContainer(), getUser()));
```
labkey-nicka, line 4533: Nice to see this is easily configured for these import actions.

labkey-nicka (approved):

PR #6618, opened 2025-04-30
Issue 52578: Disallow subfolder named "home"

labkey-susanh:
#### Rationale
Issue [52578](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52578)

#### Changes
- Add (singleton) set of illegal folder names

### DIFF
--- a/api/src/org/labkey/api/data/Container.java
+++ b/api/src/org/labkey/api/data/Container.java
@@ -844,8 +844,15 @@ public class Container implements Serializable, Comparable<Container>, Securable
             return false;
         }
 
+        if (!isProject && ILLEGAL_FOLDER_NAMES.contains(name))
+        {
+            error.append("Folder name \"").append(name).append("\" is not allowed.");
+            return false;
+        }
         return true;
     }
+
+    private static final Set<String> ILLEGAL_FOLDER_NAMES = Sets.newCaseInsensitiveHashSet("home");
 
     // Any names that we register as special servlets won't work as project names
     private static final Set<String> ILLEGAL_PROJECT_NAMES = Sets.newCaseInsensitiveHashSet("cas", "_rstudio", "_webdav", "filecontent");


labkey-jeckels (commented):
While I didn't log details in the issue, my default assumption was that we'd restrict this the LKSM UI, as this doesn't cause an issue in LKS scenarios. But I also understand that users can create a subfolder via LKS UI or APIs and have it appear in LKSM.

Any product managers care to comment? @labkey-hannah @bernarddlee @labkey-adam 

It's true that "/home" is special-cased already as a project. You can't delete it.

labkey-hannah:
I think if we're going to disallow it in the apps, given that we keep trying to make these types of restrictions more robust, we should be disallowing it everywhere

labkey-adam:
> I think if we're going to disallow it in the apps, given that we keep trying to make these types of restrictions more robust, we should be disallowing it everywhere

We certainly can't disallow creating folders in /home... every LabKey deployment already has at least one ("/home/support") and many probably have more than one. If LKSM wants to hide them I guess that's fine.

labkey-jeckels:
> > I think if we're going to disallow it in the apps, given that we keep trying to make these types of restrictions more robust, we should be disallowing it everywhere
> 
> We certainly can't disallow creating folders in /home... every LabKey deployment already has at least one ("/home/support") and many probably have more than one. If LKSM wants to hide them I guess that's fine.

This is different. We're preventing people from creating child containers named "home", not preventing them from creating child containers in "/home"

labkey-jeckels:
Another approach to consider would be to let the parent container decide what's OK. LKSM parent containers could reject "home" but other folder types could allow it.

labkey-susanh:
> Another approach to consider would be to let the parent container decide what's OK. LKSM parent containers could reject "home" but other folder types could allow it.

Ugh for folder type logic. Seems like a bit of bother for this one name that seems unlikely anyone is using anyway.  I'll look into whether I can resolve the weirdness in the app so this unlikely name is not as problematic.

labkey-jeckels:
Another complication here is the fact that the docs refer to the parent container as the "home folder" but the UI shows whatever the container's actual name is. For LKSM-only customers, that's "Home". For LKS users who use LKSM in a different project, the ambiguity would be when the subfolder is the same name as the project.

I know @labkey-keegang has also thought about this in a handful of different contexts.

I'm open to the approach in this PR if folks feel like it'll help in the places we most want to improve clarity.

PR #6620, opened 2025-05-01
Clarify wiki validator's purpose in life

labkey-adam:
#### Rationale
Update its description to focus on inline JavaScript, which is mainly what it detects


labkey-jeckels (approved):
Such noble purpose

PR #6622, opened 2025-05-02
Require a login for GET and POST in IssuesController.EmailPrefsAction

labkey-jeckels:
#### Rationale
Neither GETs nor POSTs are intended to handle `issues-emailPrefs` requests for guests

#### Changes
- Require a login for the whole action


labkey-jeckels:
https://github.com/LabKey/platform/pull/6623

PR #6624, opened 2025-05-02
Issue 53003: Study dataset JS error with invalid query.maxRows param

cnathe:
#### Rationale
Issue [53003](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53003): Study dataset JS error with invalid query.maxRows param

#### Changes
- add check that DataRegions exists before trying to addClass to form element


labkey-klum (approved):
I forgot how much work that controller action does 😢

PR #6628, opened 2025-05-04
Issue 52839: LKSM: Include the ability to use lineage language in aliquot naming patterns

XingY:
#### Rationale
A sample type can have up to two naming patterns, one for the parent sample, and one for aliquot. A single name generation state is for all samples and aliquots in a transaction, so we can handle sample counters correctly. Currently the state is wired up to sample's name generator only, while aliquot naming patten is passed in as an alternative expression, that supports basic evaluation, but not more complicated substitutions, such as lineage. This PR decouples name generation state from the expression so a state can handle both expressions with full support.   

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6628
- https://github.com/LabKey/limsModules/pull/1387

#### Changes
- Refactored NameGeneratorState outside of NameGenerator so state can work with more than one expressions
- Created SampleNameGeneratorState and removed duplicate code that handle aliquot name generation


experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -2054,4 +2005,53 @@ else if (aliquotedFrom instanceof Number)
...
+            if (aliquotedFromObj != null)
+            {
+                if (aliquotedFromObj instanceof String)
+                {
+                    // Issue 45563: We need the AliquotedFrom name to be quoted so we can properly find the parent,
+                    // but we don't want to include the quotes in the name we generate using AliquotedFrom
+                    aliquotedFrom = StringUtilsLabKey.unquoteString((String) aliquotedFromObj).trim();
+                    map.put(ExpMaterial.ALIQUOTED_FROM_INPUT, aliquotedFrom);
+                }
+                else if (aliquotedFromObj instanceof Number)
```
labkey-susanh, line 2034: Does this type checking work well for number names? If we know at this point that these names will be strings, it may be worth adding a comment to help our future selves feel better about the logic here.

labkey-susanh (approved):

PR #6633, opened 2025-05-06
Dropping an index requires callers to provide its name

labkey-adam:
#### Rationale
Stop guessing index names (step 1). This change eliminates the `ChangeType.DropIndices` option, which guessed the index name based on table and column names, leaving `DropIndicesByName` as the only option. It also adds `StorageProvisioner.dropIndices()` to drop indices by name and adjusts `AbstractAuditTypeProvider` to use this method.

Unfortunately, this does not eliminate all index name guessing: Admin-requested indexes (via metadata XML and domain save API) still create diffs using index names; we need to update this code path to use the audit provider app. Also, the SQL Server dialect still guesses names when updating the data type of a column that's part of an index.  But this change is a step in the right direction.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52913

### DIFF
--- a/api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
+++ b/api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
@@ -1134,7 +1134,6 @@ public abstract class PostgreSql91Dialect extends SqlDialect
             case DropColumns -> sql.add(getDropColumnsStatement(change));
             case RenameColumns -> sql.addAll(getRenameColumnsStatement(change));
             case DropIndicesByName -> sql.addAll(getDropIndexByNameStatements(change));
-            case DropIndices -> sql.addAll(getDropIndexStatements(change));
             case AddIndices -> sql.addAll(getCreateIndexStatements(change));
             case ResizeColumns, ChangeColumnTypes -> sql.addAll(getChangeColumnTypeStatement(change));
             case DropConstraints -> sql.addAll(getDropConstraintsStatement(change));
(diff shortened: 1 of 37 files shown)


api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
```diff
@@ -1134,7 +1134,6 @@ public List<String> getChangeStatements(TableChange change)
             case DropColumns -> sql.add(getDropColumnsStatement(change));
             case RenameColumns -> sql.addAll(getRenameColumnsStatement(change));
             case DropIndicesByName -> sql.addAll(getDropIndexByNameStatements(change));
-            case DropIndices -> sql.addAll(getDropIndexStatements(change));
```
labkey-susanh: Is a similar change wanted or needed for SqlServer?

labkey-adam: Indeed... I neglected to tag you on the related PR: https://github.com/LabKey/premiumModules/pull/243

api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
```diff
@@ -1413,51 +1412,20 @@ private void addCreateIndexStatements(List<String> statements, TableChange chang
...
-                                   CREATE %s INDEX %s ON %s (%s);
-                            END IF;
-                            END$$""",
-                    getStringHandler().quoteStringLiteral(nameIndex(change.getTableName(), index.columnNames)),
-                    getStringHandler().quoteStringLiteral(change.getSchemaName()),
-                    index.isUnique ? "UNIQUE" : "",
-                    nameIndex(change.getTableName(), index.columnNames),
-                    makeTableIdentifier(change),
-                    makePropertyIdentifiers(index.columnNames)));
+            statements.add(String.format("CREATE %sINDEX %s ON %s (%s);",
```
labkey-susanh: Missing space before `INDEX`?

labkey-susanh: Ah, never mind. I see.

labkey-susanh (approved):

PR #6636, opened 2025-05-06
Issue 51126: AssayResultUpdateService.updateRow fix to account for TableSelector getMap() returning column alias

cnathe:
#### Rationale
Issue [51126](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51126): LKSM: Remapping assay results to a new sample disallows delete of old sample

This issue regressed recently after the TableSelector getMap() change to the key used for the map. The code was using the keys of the map to call TableInfo.getColumn() to identify which field values changed in the assay result row so that the assay run syncLineage could be called. Since getColumn() looks for columns by name or fieldKey, this regressed for cases where the sample lookup field had special characters, meaning that the column.getName() does not equal column.getAlias().

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6636
- https://github.com/LabKey/limsModules/pull/1390

#### Changes
- AssayResultUpdateService.updateRow to use the updateRow result map key to get the ColumnInfo

### DIFF
--- a/assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
+++ b/assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
@@ -274,9 +274,8 @@ public class AssayResultUpdateService extends DefaultQueryUpdateService
 
             if (col != null)
             {
-                String entryKey = col.getAlias().getId(); // getRow() ends up using TableSelector.getMap() which is keyed by column alias
-                Object oldValue = originalRow.get(entryKey);
-                Object newValue = updatedValues.get(entryKey);
+                Object oldValue = col.getValue(originalRow);
+                Object newValue = col.getValue(updatedValues);
                 boolean hasValueChanged = !Objects.equals(oldValue, newValue);
 
                 if (hasValueChanged)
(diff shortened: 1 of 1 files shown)


assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
```diff
@@ -266,24 +266,25 @@ protected Map<String, Object> updateRow(
...
             // Also check for properties
             TableInfo table = getQueryTable();
             ColumnInfo col = table.getColumn(entry.getKey());
 
             if (col != null)
             {
-                Object oldValue = originalRow.get(entry.getKey());
-                Object newValue = entry.getValue();
+                String entryKey = col.getAlias().getId(); // getRow() ends up using TableSelector.getMap() which is keyed by column alias
+                Object oldValue = originalRow.get(entryKey);
```
labkey-matthewb: I recommend col.getValue(originalRow) and col.getValue(updatedValues) works.

assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
```diff
@@ -266,24 +266,25 @@ protected Map<String, Object> updateRow(
...
+                Object newValue = updatedValues.get(entryKey);
                 boolean hasValueChanged = !Objects.equals(oldValue, newValue);
 
                 if (hasValueChanged)
                     _assaySampleLookupContext.trackSampleLookupChange(container, user, table, col, run);
 
                 TableInfo fkTableInfo = col.getFkTableInfo();
                 // Don't follow the lookup for specimen IDs, since their FK is very special and based on target study, etc
-                if (hasValueChanged && fkTableInfo != null && !AbstractAssayProvider.SPECIMENID_PROPERTY_NAME.equalsIgnoreCase(entry.getKey()))
+                if (hasValueChanged && fkTableInfo != null && !AbstractAssayProvider.SPECIMENID_PROPERTY_NAME.equalsIgnoreCase(entryKey))
```
labkey-matthewb: entryKey is the alias.  Shouldn't this be equalsIgnoreCase(column.getName())?

cnathe: with my change, entryKey is based off of the result map from the updateRows() response so should be the column name. Though I like just using that key to get the columnInfo object and using that everywhere after so I've made this change.

assay/api-src/org/labkey/api/assay/AssayResultUpdateService.java
cnathe: yes! that col.getValue(Map) helper is much better. done

labkey-matthewb (approved):

labkey-nicka (approved):

PR #6638, opened 2025-05-06
Issue 52968: handle context path in FileLinkDisplayColumn.filePathExist

labkey-nicka:
#### Rationale
This addresses [Issue 52968](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52968) by checking for the context path when checking if a template file exists.

#### Changes
- Handle context path in `FileLinkDisplayColumn.filePathExists`

### DIFF
--- a/api/src/org/labkey/api/study/assay/FileLinkDisplayColumn.java
+++ b/api/src/org/labkey/api/study/assay/FileLinkDisplayColumn.java
@@ -38,6 +38,7 @@ import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.SchemaKey;
 import org.labkey.api.security.User;
 import org.labkey.api.security.permissions.ReadPermission;
+import org.labkey.api.settings.AppProps;
 import org.labkey.api.util.ContainerContext;
 import org.labkey.api.util.FileUtil;
 import org.labkey.api.util.NetworkDrive;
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/study/assay/FileLinkDisplayColumn.java
```diff
@@ -263,6 +264,12 @@ public static boolean filePathExist(String path, Container container, User user)
         var resolver = WebdavService.get().getResolver();
         // Resolve path under webdav root
         Path parsed = Path.parse(StringUtils.trim(davPath));
+
+        // Issue 52968: handle context path
+        Path contextPath = AppProps.getInstance().getParsedContextPath();
```
XingY, line 269: Do we need to worry about path is `/labkey` but contextPath is `labkey` (without leading slash?).

XingY (approved):

api/src/org/labkey/api/study/assay/FileLinkDisplayColumn.java
XingY: NVM, context path is either empty or has to start with /
`assert _contextPathStr.isEmpty() || _contextPathStr.startsWith("/");`

labkey-nicka: As found by you, https://github.com/LabKey/platform/blob/689dc28d2/api/src/org/labkey/api/settings/AppPropsImpl.java#L142, we can assume the context path is empty or begins with a "/" (when assertions are on...).

PR #6641, opened 2025-05-07
Issue 53017: Plate creation from template test fixes for metadata columns with special characters

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53017

The PlateTemplateTest.testCreateTemplateFromPlate and PlateTemplateTest.testCreatePlateFromTemplate started failing recently. There are a few places in the PlateManager code where it is attempting to copy well data, including for custom metadata fields. Those cases were expecting the data map from the well data selectors to be keyed by column name so that when it turns around and calls insertRows that map will contain the keys needed. The recent change made it so that the TableSelector getMapCollection is keyed by column alias. This works fine for most fields since the name and alias are the same, but for the custom fields with special characters this resulted in the data for those columns being dropped.

#### Changes
- key well metadata maps by column name, updated TableSelector getMap is keyed by column alias by default

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateManager.java
+++ b/assay/src/org/labkey/assay/plate/PlateManager.java
@@ -3830,28 +3830,35 @@ public class PlateManager implements PlateService, AssayListener, ExperimentList
 
         try (Results results = new TableSelector(metadataTable, filter, null).getResults())
         {
+            Map<FieldKey, ColumnInfo> fieldMap = results.getFieldMap();
+
             while (results.next())
             {
                 var row = results.getFieldKeyRowMap();
-                var lsid = (String) row.get(WellTable.Column.Lsid.fieldKey());
                 var metadata = new CaseInsensitiveHashMap<>();
 
-                for (var key : row.keySet())
-                {
-                    // Issue 53017: usages of getWellData are expecting the WellData metadata map to be keyed by column names
-                    // (see savePlateImpl wellQus.insertRows() which requires rows to be keyed by column names)
-                    String colName = results.getFieldMap().get(key).getName();
-                    if (row.get(key) != null && !ignoredKeys.contains(colName))
-                        metadata.put(colName, row.get(key));
-                }
+                row.forEach((key, value) -> {
+                    if (value != null)
+                    {
+                        // Issue 53017: usages of getWellData are expecting the WellData metadata map to be keyed
+                        // by column names (see savePlateImpl wellQus.insertRows() which requires rows to be keyed
+                        // by column names)
+                        String colName = fieldMap.get(key).getName();
+                        if (!ignoredKeys.contains(colName))
+                            metadata.put(colName, value);
+                    }
+                });
 
                 if (!metadata.isEmpty())
+                {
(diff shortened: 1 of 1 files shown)


assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3825,19 +3828,31 @@ private List<WellData> getWellMetadata(Container container, User user, List<Well
...
+                        metadata.put(colName, row.get(key));
+                }
+
+                if (!metadata.isEmpty())
+                    metadataMap.put(lsid, metadata);
+            }
+        }
+        catch (SQLException e)
+        {
+            throw new RuntimeException(e);
```
labkey-nicka: Make this `RuntimeSQLException`

assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3825,19 +3828,31 @@ private List<WellData> getWellMetadata(Container container, User user, List<Well
         var metadataMap = new HashMap<String, Map<String, Object>>();
         var ignoredKeys = CaseInsensitiveHashSet.of("_row", WellTable.Column.Lsid.name());
 
-        new TableSelector(metadataTable, filter, null).getMapCollection().forEach(row -> {
-            var lsid = (String) row.get(WellTable.Column.Lsid.name());
-            var metadata = new CaseInsensitiveHashMap<>();
-
-            for (var key : row.keySet())
+        try (Results results = new TableSelector(metadataTable, filter, null).getResults())
```
labkey-nicka, line 3831: nit: Efficiency gains:

```java
try (Results results = new TableSelector(metadataTable, filter, null).getResults())
{
    Map<FieldKey, ColumnInfo> fieldMap = results.getFieldMap();

    while (results.next())
    {
        var row = results.getFieldKeyRowMap();
        var metadata = new CaseInsensitiveHashMap<>();

        row.forEach((key, value) -> {
            if (value != null)
            {
                // Issue 53017: usages of getWellData are expecting the WellData metadata map to be keyed
                // by column names (see savePlateImpl wellQus.insertRows() which requires rows to be keyed
                // by column names)
                String colName = fieldMap.get(key).getName();
                if (!ignoredKeys.contains(colName))
                    metadata.put(colName, value);
            }
        });

        if (!metadata.isEmpty())
        {
            var lsid = (String) row.get(WellTable.Column.Lsid.fieldKey());
            metadataMap.put(lsid, metadata);
        }
    }
}
```

labkey-nicka (approved):
As usual with these kinds of updates be sure to run the plate tests that all reside in Biologics. PlateController.ispec.ts may have coverage here as well.

assay/src/org/labkey/assay/plate/PlateManager.java
cnathe: done

assay/src/org/labkey/assay/plate/PlateManager.java
cnathe: done

PR #6642, opened 2025-05-07
Issue 52098, 49422: when looking up by alternate keys do that first so number names will resolve well

labkey-susanh:
#### Rationale
Issue [52098](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52098): Sample statuses with number labels may not resolve to the correct status
Issue [49422](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49422): List lookups with number-like values may not resolve to the correct status

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1416
- https://github.com/LabKey/testAutomation/pull/2455
- https://github.com/LabKey/ehrModules/pull/960

#### Changes
- Add `LookupResolutionType` as a replacement for `allowLookupByAlternateKey`
- Update `SimpleTranslator` to allow for the different orderings for lookup resolution
- Update `DataIteratorContext` with a new `hasBeenCoerced` property (could perhaps be named better?) so we don't try to re-convert lookups that have already been converted.


api/src/org/labkey/api/data/LookupResolutionType.java
```diff
@@ -0,0 +1,34 @@
+package org.labkey.api.data;
+
+public enum LookupResolutionType
+{
+    primaryKey(true, false, true), // known that the use will always supply the pk value
```
labkey-matthewb:
```suggestion
    primaryKey(true, false, true), // known that the user will always supply the pk value
```

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -352,11 +337,19 @@ private Object fetch(Triple<ColumnInfo, ColumnInfo, MultiValuedMap<?,?>> triple,
                 }
 
                 // ArrayListValuedHashMap returns an empty collection if 'k' is not in the map.
-                if (bulkLoaded == null || bulkLoaded.isEmpty())
+                if (bulkLoaded == null || bulkLoaded.isEmpty() )
                 {
-                    TableSelector ts = createSelector(pkCol, altKeyCol, k);
-                    ts.fillMultiValuedMap(map);
-                    vs = map.get(k);
+                    if (altKeyCol.getJdbcType().getJavaClass().isAssignableFrom(k.getClass()))
```
labkey-matthewb: This doesn't seem quite right.  I expect k will almost always be a string, should we attempt a convert?

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -1338,15 +1340,20 @@ private SimpleConvertColumn createConvertColumn(@NotNull ColumnInfo col, int fro
...
+        if (!hasBeenRemapped && fk != null && lookupResolutionType.useAlternateKey() && fk.allowImportByAlternateKey())
         {
             // Issue 48347: if the lookup field has a "Lookup Validator", then treat the missing values as an error
             boolean hasValidator = pd != null && pd.getValidators().stream().anyMatch(v -> PropertyValidatorType.Lookup.getLabel().equalsIgnoreCase(v.getName()));
 
             RemapMissingBehavior missing = remapMissingBehavior;
             if (missing == null)
                 missing = col.isRequired() || hasValidator ? RemapMissingBehavior.Error : RemapMissingBehavior.Null;
-            c = new RemapPostConvertColumn(c, fromIndex, col, missing, true);
+            // For enum tables, we should not be using number names for the values, so we will look up by primary key first (resolving rowIds) then alternate.
```
labkey-matthewb: I can't picture a situation where would would want to use RemapMissingBehavior.Null?

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
```diff
@@ -1262,10 +1261,11 @@ public int addConvertColumn(ColumnInfo col, int fromIndex)
      * @param fromIndex            Source column to create the output column from.
      * @param mvIndex              Missing value column index.
      * @param remapMissingBehavior The behavior desired when remapping fails.  If null, indicate an error if the column is required or null if not required.
+     * @param hasBeenRemapped      Indicates if remapping of lookup columns has happened or not
```
labkey-matthewb: I don't particularly like this name.  I think it would be better to have an "active" name here.  E.g. remapLookups or withLookupRemapping.

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
labkey-susanh: Ah yes, we were previously doing this for the pre-trigger phase, but not anymore, so, as discussed, I agree we should us `OriginalValue` here instead of `Null`.  this may change existing behavior when we are not validating lookups, but for the better.

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
labkey-susanh: Trying to remember why I needed this... I think it may have been when `k` was an integer after having already been converted and we were going through a second time, so perhaps is no longer needed. I'll take it out and retest.

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
labkey-susanh: Updated to `withLookupRemapping`

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
labkey-susanh: Ah, yes. This is needed because of the system fields (like users) that initially come through with rowIds. Without this `if` statement (or some `if` statement), we get a SQL error because it is trying to compare the rowId to the string alternate key (e.g., email).  I'm open to suggestions for other checks here, but this seems less expensive than a try-catch around a call to convert.

api/src/org/labkey/api/dataiterator/SimpleTranslator.java
labkey-matthewb: That makes sense, but doesn't address the case where the AK is not a String.  Is that scenario we support?  The value coming in will often be String regardless of AK type.

labkey-matthewb: Separately, we don't use Java class comparison much in import/query land.  More often just trying to figure out is it Stringy, Datey, or Numbery.  (Is Integer assignable from Long)

labkey-susanh: > That makes sense, but doesn't address the case where the AK is not a String. Is that scenario we support? The value coming in will often be String regardless of AK type.

I think this is meant to be a comment on my use of  the `isAssignalbleFrom` comparison, so not this part of the code, right? 

The logic in [AbstractForeignKey.allowImportByAlternateKey](https://github.com/LabKey/platform/blob/b0947693db9deb6d7d7f09924aa66b43a83344ae/api/src/org/labkey/api/data/AbstractForeignKey.java#L404) indicates that we don't allow this unless the alternate key type is a String.  Given that, we could perhaps update this to check if `k` is a `String`. How hacky is that?

Pathological case if using `convert`: The value given is a number (the PK value), we use convert and get back a string, then we use that string for this fetching and get a number-named thing even though we started with the PK for some other thing.

labkey-matthewb: Sorry about that.  If we only allow String (which is clearly the most useful case), then just checking instanceof String seems fine (with comment).  Data no longer runs through remap twice though, right?

labkey-susanh: Correct. We are no longer running through remap twice, which is why the case I described is pathological. It would have to be a value for a field that could have number names but has integer values (not Strings that look like integers) to start out with, which would be even less common (you can't have an email address that looks like a number, for example). Less common does not mean it won't happen though, so I'm going to error on the side of caution. 

Code has been updated to check for String type with comment amended. Please add further comments in that context.

labkey-matthewb (approved):

PR #6645, opened 2025-05-07
Move some PostgreSQL-specific behaviors out of Redshift base class

labkey-adam:
#### Rationale
Most important move is the query for "max_identifier_length", which breaks on Redshift. But also, the UTF-8-byte truncation is not applicable to Redshift.

#### Related Pull Requests
- https://github.com/LabKey/premiumModules/pull/247


api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
```diff
@@ -87,16 +86,12 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
-// Base dialect for PostgreSQL. PostgreSQL 9.1 is no longer supported, however, we keep the dialects versioned to
-// track changes we've implemented for each version over time.
+// Base dialect for PostgreSQL AND Redshift. IMPORTANT: Make sure everything added here applies to Redshift as well.
 public abstract class PostgreSql91Dialect extends SqlDialect
```
labkey-matthewb: rename?

labkey-matthewb (approved):

api/src/org/labkey/api/data/dialect/PostgreSql91Dialect.java
labkey-adam: Done

PR #6651, opened 2025-05-09
Add dirty handling to allow list admin pages

labkey-adam:
#### Rationale
It's nice to warn users when navigating away may cause data loss. Add dirty handling to the Allowed External Redirect Hosts, Allowed External Resource Hosts, and Allowed File Extensions pages. Clarify file extensions instructions.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53039
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53038

### DIFF
--- a/core/src/org/labkey/core/admin/existingListValues.jsp
+++ b/core/src/org/labkey/core/admin/existingListValues.jsp
@@ -102,8 +102,19 @@
                 <td><br/>
                     <input type="hidden" id="saveAll" name="saveAll">
                     <%=isTroubleshooter ? button("Done").href(urlProvider(AdminUrls.class).getAdminConsoleURL()) : button("Save").primary(true).onClick("return saveAll();")%>
-                    <%=!isTroubleshooter ? button("Delete All").href(urlFor(DeleteAllValuesAction.class).addParameter("type", bean.getTypeEnum().name()))
-                            .onClick("_formExisting.setClean(); return LABKEY.Utils.confirmAndPost('" + h("Are you sure you want to delete all " + bean.getTypeEnum().getTitle() + "s") + "', this.href);") : HtmlString.EMPTY_STRING%>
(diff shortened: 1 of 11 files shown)


core/src/org/labkey/core/admin/existingListValues.jsp
```diff
@@ -99,7 +102,8 @@
                 <td><br/>
                     <input type="hidden" id="saveAll" name="saveAll">
                     <%=isTroubleshooter ? button("Done").href(urlProvider(AdminUrls.class).getAdminConsoleURL()) : button("Save").primary(true).onClick("return saveAll();")%>
-                    <%=!isTroubleshooter ? button("Delete All").href(urlFor(DeleteAllValuesAction.class).addParameter("type", bean.getTypeEnum().name())).usePost("Are you sure you want to delete all " + bean.getTypeEnum().getTitle() + "s?") : HtmlString.EMPTY_STRING%>
+                    <%=!isTroubleshooter ? button("Delete All").href(urlFor(DeleteAllValuesAction.class).addParameter("type", bean.getTypeEnum().name()))
+                            .onClick("_formExisting.setClean(); return LABKEY.Utils.confirmAndPost('" + h("Are you sure you want to delete all " + bean.getTypeEnum().getTitle() + "s") + "', this.href);") : HtmlString.EMPTY_STRING%>
```
labkey-jeckels: Will this clear the dirty state before the user confirms the action

labkey-jeckels (approved):

core/src/org/labkey/core/admin/existingListValues.jsp
labkey-adam: Yes it will! You think that's a problem? :)

labkey-adam: Updated to preserve dirty state if user doesn't confirm

PR #6655, opened 2025-05-12
Add detailed auditing of domain changes & comment ability

XingY:
#### Rationale
There is currently an App only setting that "requires users provide a reason before completing certain actions", which forces user to enter an audit comment before update/delete a data in the App. Certain update and delete actions of data are wired up to accept `auditUserComment` but so far no enforcement is in place to ensure this comment is provided. This PR extends the actions that accept `auditUserComment` to updateDomain, deleteDomain and assay protocol update  actions. There is a future story to enforce "required" comment. 

This PR also records more details about what's updated during a domain create/update. The domain event and domain property event will now have newRecordMap and oldRecordMap columns, that record all domain or property settings during update/create. Pomain property event will now report more changes, including regex, conditional formatting, text choices, etc. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-api-js/pull/194
* https://github.com/LabKey/labkey-ui-components/pull/1792
* https://github.com/LabKey/labkey-ui-premium/pull/751
* https://github.com/LabKey/platform/pull/6655
* https://github.com/LabKey/commonAssays/pull/885
* https://github.com/LabKey/limsModules/pull/1399
* https://github.com/LabKey/testAutomation/pull/2445
* https://github.com/LabKey/wnprc-modules/pull/827
* https://github.com/LabKey/clientModules/pull/24

#### Changes
- wire up `auditUserComment` to `updateDomain`, `deleteDomain` and assay protocol actions and utils
- wire up Domain Audit Event for dataset delete action
- add old and new record map to Domain and Domain Property Events
- fix up domain event comments for properties update for sample type, dataclass, assay, list and dataset
- fix up domain property event comment for column datetime, number format, conditional format, regex validator, text choices update, etc.
- add selenium tests to check for domain and property audit comment updates


api/src/org/labkey/api/study/Dataset.java
```diff
@@ -364,6 +364,8 @@ public String getRecallFromStudyAuditMessage(String label, int recordCount)
 
     void delete(User user);
 
+    void delete(User user, String auditUserComment);
```
labkey-nicka: Annotate `@Nullable` as well as in the implementation.

```java
void delete(User user, @Nullable String auditUserComment);
```

study/src/org/labkey/study/model/StudyManager.java
```diff
@@ -2631,7 +2632,7 @@ public void deleteDataset(StudyImpl study, User user, DatasetDefinition ds, bool
     /** delete a dataset type and data
      *  does not clear typeURI as we're about to delete the dataset
      */
-    private void deleteDatasetType(Study study, User user, DatasetDefinition ds)
+    private void deleteDatasetType(Study study, User user, DatasetDefinition ds, String auditUserComment) throws DomainNotFoundException
```
labkey-nicka: This does not throw a `DomainNotFoundException`. Annotate `@Nullable String auditUserComment`.

study/src/org/labkey/study/model/StudyManager.java
```diff
@@ -2641,7 +2642,17 @@ private void deleteDatasetType(Study study, User user, DatasetDefinition ds)
         if (!ds.canDeleteDefinition(user))
             throw new IllegalStateException("Can't delete dataset: " + ds.getName());
 
-        StorageProvisioner.get().drop(ds.getDomain());
+        Domain domain = ds.getDomain();
+        if (domain == null)
+            return;
+
+        DomainAuditProvider.DomainAuditEvent event = new DomainAuditProvider.DomainAuditEvent(study.getContainer(), String.format("The domain %s was deleted", domain.getName()));
```
labkey-nicka, line 2581: I see that you tried to call `domain.delete(user, auditUserComment)`. Why did you move away from that? Did you run into deletion problems / lock-up? Seems that would be much preferred to wiring up our own audit event here.

study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -4742,7 +4742,7 @@ public boolean handlePost(IdForm form, BindException errors) throws Exception
             try (DbScope.Transaction transaction = scope.ensureTransaction())
             {
                 // performStudyResync==false so we can do this out of the transaction
-                StudyManager.getInstance().deleteDataset(getStudyRedirectIfNull(), getUser(), ds, false);
+                StudyManager.getInstance().deleteDataset(getStudyRedirectIfNull(), getUser(), ds, false, null);
```
labkey-nicka, line 4806: Should we be exposing `auditUserComment` on this endpoint so it can be passed in?

study/src/org/labkey/study/controllers/DatasetController.java
```diff
@@ -225,7 +225,7 @@ public boolean handlePost(DatasetDeleteForm form, BindException errors)
                 {
                     if (!def.canDeleteDefinition(getUser()))
                         continue;
-                    StudyManager.getInstance().deleteDataset(study, getUser(), def, false);
+                    StudyManager.getInstance().deleteDataset(study, getUser(), def, false, null);
```
labkey-nicka, line 228: Should we be exposing `auditUserComment` on this endpoint so it can be passed in?

api/src/org/labkey/api/exp/api/ExperimentJSONConverter.java
```diff
@@ -957,6 +959,21 @@ else if (value instanceof Map map)
         return aliases;
     }
 
+    public static @Nullable String getImportAliasStringVal(@Nullable Map<String, Map<String, Object>> importAliases)
```
labkey-nicka, line 962: All of the usages of this have their variables names suffixed with `Json`. That seems not right given that these are strings.

api/src/org/labkey/api/exp/api/ExperimentService.java
```diff
@@ -279,7 +279,8 @@ ValidationException updateDataClass(
         @NotNull ExpDataClass dataClass,
         @Nullable DataClassDomainKindProperties options,
         GWTDomain<? extends GWTPropertyDescriptor> original,
-        GWTDomain<? extends GWTPropertyDescriptor> update
+        GWTDomain<? extends GWTPropertyDescriptor> update,
+        String auditUserComment
```
labkey-nicka: Annotate `@Nullable`

api/src/org/labkey/api/exp/api/ExperimentService.java
```diff
@@ -1054,7 +1055,7 @@ List<? extends ExpProtocol> getExpProtocolsWithParameterValue(
 
     void ensureContainerDataTypeExclusions(@NotNull DataTypeForExclusion dataType, @Nullable DataTypeForExclusion relatedDataType, @Nullable Collection<Integer> excludedDataTypeRowIds, @NotNull String excludedContainerId, User user);
 
-    void ensureDataTypeContainerExclusions(@NotNull DataTypeForExclusion dataType, @Nullable Collection<String> excludedContainerIds, @NotNull Integer dataTypeId, User user);
+    @NotNull String ensureDataTypeContainerExclusions(@NotNull DataTypeForExclusion dataType, @Nullable Collection<String> excludedContainerIds, @NotNull Integer dataTypeId, User user);
```
labkey-nicka: Odd from an interface stand-point that this now returns a string but `ensureDataTypeContainerExclusionsNonAdmin` does not.

Additionally, please add documentation for this interface describing what the string is being returned.

api/src/org/labkey/api/exp/list/ListDefinition.java
```diff
@@ -292,6 +292,7 @@ public static BodySetting getForValue(int value)
     void save(User user) throws Exception;
     void save(User user, boolean ensureKey) throws Exception;
     void delete(User user) throws DomainNotFoundException;
+    void delete(User user, String auditUserComment) throws DomainNotFoundException;
```
labkey-nicka: Annotate `@Nullable`. I'll stop repeating this feedback, but should be done on all interface updates.

api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -744,9 +744,60 @@ public static ValidationException updateDomainDescriptor(GWTDomain<? extends GWT
...
     }
 
-    /** @return Errors encountered during the save attempt */
     @NotNull
     public static ValidationException updateDomainDescriptor(GWTDomain<? extends GWTPropertyDescriptor> orig, GWTDomain<? extends GWTPropertyDescriptor> update, Container container, User user, boolean updateDomainName, @Nullable String auditComment)
+    {
+        return updateDomainDescriptor(orig, update, container, user, updateDomainName, auditComment, null);
+    }
+
+    public static String getStringPropChangeMsg(String propertyName, String oldProp, String newProp)
```
labkey-nicka: These prop change message utilities do not seem like they pertain to "domains", but rather, to auditing message generation. Is there somewhere else we can put these that more appropriately locates them and allows for easier identification of intended use?

Additionally, this code feels like it already exists in some form somewhere else. If our existing audit "diff" messaging utilities do not suffice, then we should update them to support what we need to convey in these domain scenarios.

api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -1185,21 +1242,28 @@ private static void _copyProperties(DomainProperty to, GWTPropertyDescriptor fro
...
     }
 
-    private static List<Map<String, Object>> updatePropertyValidators(DomainProperty dp, GWTPropertyDescriptor oldPd, GWTPropertyDescriptor newPd)
+    private static List<Map<String, Object>> updatePropertyValidators(DomainProperty dp, @Nullable GWTPropertyDescriptor oldPd, GWTPropertyDescriptor newPd)
     {
         Map<Integer, GWTPropertyValidator> newProps = new HashMap<>();
         List<Map<String, Object>> valueUpdates = new ArrayList<>();
 
+        PropertyDescriptor oldPropertyDescriptor = dp.getPropertyDescriptor().clone();
+        String oldValidatorStr = dp.getPropertyValidatorStringVal(); // record the old value before dp is mutated
```
labkey-nicka: This `oldValidatorStr` is unused.

api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -1301,20 +1372,35 @@ private static void updateTextChoiceValueRows(Domain domain, User user, String p
         }
     }
 
-    private static void _copyValidator(IPropertyValidator pv, GWTPropertyValidator gpv)
+    /**
```
labkey-nicka: nit: This comment block is relatively useless. I'd say just remove it.

experiment/src/org/labkey/experiment/ExpDataFileListener.java
```diff
@@ -76,7 +76,7 @@ public int fileMoved(@NotNull Path src, @NotNull Path dest, @Nullable User user,
             // if the data object moved containers, set that as well
             if (targetContainer != null && !targetContainer.equals(sourceContainer))
                 data.setContainer(targetContainer);
-            data.save(user);
+            data.save(user); // bad!!
```
labkey-nicka: Is there a fix here? Consider adding a FIXME statement?

labkey-nicka (commented):
In general, this approach of being very explicit about each change and generating the audit message snippet wherever a changes is made is fine but I don't think it is ideal. The amount of checking for changes and the ease with which future changes could miss adding these details makes this cumbersome and fragile.

Seems like there is potential for a way to have domains retain a difference so that when auditing occurs it can be computed what the auditing should record.

That said, I'm not going to hold up this work on this account.

study/src/org/labkey/study/model/StudyManager.java
XingY: It was causing a failure on TC https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_Bvt/3508494, consistently. I cannot repro the failure locally (the problematic dataset is created in ClientAPITest), nor understand it. It might has to do with study container vs definition container vs domain container, but not sure why no repro for me.

study/src/org/labkey/study/controllers/StudyController.java
XingY: That would be a future story I think. We will need LKS UI change to collect that comment first. Currently, domain.delete api can be used to delete a dataset with provided user comment.

study/src/org/labkey/study/controllers/DatasetController.java
XingY: Same as previous comment, a future story, perhaps when we work on enforcing user comment at api level.

api/src/org/labkey/api/exp/api/ExperimentJSONConverter.java
XingY: updated

experiment/src/org/labkey/experiment/ExpDataFileListener.java
XingY: Issue comment added

api/src/org/labkey/api/exp/property/DomainUtil.java
XingY: The implementation has been updated to follow standard old/newRecordMap approach. These utils are no longer needed and removed.

XingY:
> In general, this approach of being very explicit about each change and generating the audit message snippet wherever a changes is made is fine but I don't think it is ideal. The amount of checking for changes and the ease with which future changes could miss adding these details makes this cumbersome and fragile.
> 
> Seems like there is potential for a way to have domains retain a difference so that when auditing occurs it can be computed what the auditing should record.
> 
> That said, I'm not going to hold up this work on this account.

The implementation has been updated to be a bit more aligned with this comment.

api/src/org/labkey/api/audit/AbstractAuditTypeProvider.java
```diff
@@ -394,7 +395,7 @@ public static Map<String, String> decodeFromDataMap(String properties)
         }
     }
 
-    public static String encodeForDataMap(Container c, Map<String, ?> properties)
+    public static String encodeForDataMap(@Nullable Container c, Map<String, ?> properties)
```
labkey-nicka: nit: Looks like the `Container` parameter is not actually used by this method and can be factored out.

api/src/org/labkey/api/data/PropertyStorageSpec.java
```diff
@@ -500,6 +501,14 @@ public int hashCode()
         {
             return Objects.hash(Arrays.hashCode(columnNames), isClustered, isClustered);
         }
+
+        public String toStringVal()
+        {
+            if (columnNames == null)
```
labkey-nicka, line 507: nit: Should this check for empty as well? Seems a bit incongruent to return when `null`, but then join and display the unique flag setting when `columnNames` is empty.

api/src/org/labkey/api/assay/AbstractAssayProvider.java
```diff
@@ -1246,7 +1247,7 @@ public boolean hasUsefulDetailsPage()
     private static final String SCRIPT_PATH_DELIMITER = "|";
 
     @Override
-    public ValidationException setValidationAndAnalysisScripts(ExpProtocol protocol, @NotNull List<AnalysisScript> scripts) throws ExperimentException
+    public Pair<ValidationException, Pair<String, String>> setValidationAndAnalysisScripts(ExpProtocol protocol, @NotNull List<AnalysisScript> scripts) throws ExperimentException
```
labkey-nicka, line 1250: nit: A typed `record` would provide more context than type `Pair<String, String>` about what these strings are.

api/src/org/labkey/api/exp/PropertyDescriptor.java
```diff
@@ -533,6 +535,57 @@ public void setDatabaseDefaultValue(@Nullable Object databaseDefaultValue)
     {
         _databaseDefaultValue = databaseDefaultValue;
     }
+
+    public Map<String, Object> getAuditRecordMap(@Nullable String validatorStr, @Nullable String conditionalFormatStr)
+    {
+        Map<String, Object> map = new LinkedHashMap<>();
```
labkey-nicka, line 541: Should this include any of the following:

- getContainer()
- getPropertyId()
- getDatabaseDefaultValue()

api/src/org/labkey/api/exp/property/Domain.java
```diff
@@ -88,8 +88,8 @@ default void delete(@Nullable User user, @Nullable String auditUserComment) thro
         delete(user);
     }
     void save(User user) throws ChangePropertyDescriptorException;
-    void save(User user, boolean auditComment) throws ChangePropertyDescriptorException;
-    void save(User user, @Nullable String allowAddBaseProperty) throws ChangePropertyDescriptorException;
+    void save(User user, @Nullable Map<String, Object> newRecordMap) throws ChangePropertyDescriptorException;
+    void save(User user, @Nullable String auditComment, @Nullable String auditUserComment, @Nullable Map<String, Object> oldRecordMap, @Nullable Map<String, Object> newRecordMap) throws ChangePropertyDescriptorException;
```
labkey-nicka: nit: A comment here about the difference between `auditComment` and `auditUserComment` could be useful.

api/src/org/labkey/api/exp/property/DomainPropertyAuditProvider.java
```diff
@@ -137,6 +138,14 @@ public TableInfo getLookupTableInfo()
                     col.setLabel("Domain Name");
             }
         };
+
+        appendValueMapColumns(table);
+
+        DetailsURL url = DetailsURL.fromString("audit-detailedAuditChanges.view?auditRowId=${rowId}&auditEventType=" + EVENT_NAME);
```
labkey-nicka: Consider consolidating these URL generation logic:

image

experiment/src/org/labkey/experiment/api/ExpProtocolImpl.java
```diff
@@ -475,4 +480,33 @@ public void setStatus(Status status)
     {
         _object.setStatus(status);
     }
+
+    @Override
+    public Map<String, Object> getAuditRecordMap(AssayProvider provider, Container container)
```
labkey-nicka: Container parameter is unused

api/gwtsrc/org/labkey/api/gwt/client/model/GWTIndex.java
```diff
@@ -69,4 +74,22 @@ public void setUnique(boolean unique)
...
+        if (_columnNames == null || _columnNames.isEmpty())
+            return "";
+
+        return StringUtils.join(_columnNames, ", ") + ", unique: " + isUnique();
+    }
+
+    public static List<String> toStringVals(List<GWTIndex> indices, Set<PropertyStorageSpec.Index> excludeBaseIndices)
+    {
+        if (indices == null || indices.isEmpty())
+            return null;
```
labkey-nicka, line 89: nit: Can just return `Collections.emptyList()` then callers do not need to worry about null values (and could be annotated `@NotNull`.

experiment/src/org/labkey/experiment/api/property/DomainPropertyImpl.java
```diff
@@ -1127,6 +1147,57 @@ public String toString()
         return super.toString() + _pd.getPropertyURI();
     }
 
+    public Map<String, Object> getAuditRecordMap(@Nullable String validatorStr, @Nullable String conditionalFormatStr)
```
labkey-nicka, line 1150: Since `DomainPropertyImpl` wraps a `PropertyDescriptor` should this somehow utilize the audit record of the underlying property descriptor and then extend with whatever additional properties are desired?

labkey-nicka (approved):

labkey-nicka:
> > In general, this approach of being very explicit about each change and generating the audit message snippet wherever a changes is made is fine but I don't think it is ideal. The amount of checking for changes and the ease with which future changes could miss adding these details makes this cumbersome and fragile.
> > Seems like there is potential for a way to have domains retain a difference so that when auditing occurs it can be computed what the auditing should record.
> > That said, I'm not going to hold up this work on this account.
> 
> The implementation has been updated to be a bit more aligned with this comment.

Much improved!

PR #6658, opened 2025-05-12
Inspect cached objects for questionable members

labkey-adam:
#### Rationale
Cached objects can end up holding stale Containers, Users, etc. In general, we want cached objects to hold IDs and retrieve these objects from their own caches as needed.

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53058 

#### Related Pull Requests
- https://github.com/LabKey/ehrModules/pull/952


api/src/org/labkey/api/cache/CacheManager.java
```diff
@@ -175,10 +180,41 @@ public static <V> void validate(String debugName, @Nullable V value)
...
         if (null != description)
         {
-            LOG.warn(debugName + " attempted to cache " + description + ", which could be mutated by callers!");
+            LOG.warn("{} attempted to cache {}, which could be mutated by callers!", debugName, description);
+        }
+
+        // Log questionable members, but don't do the work if we're not going to log it
+        if (LOG.isDebugEnabled() && value != null)
+        {
+            for (Field field : getAllInstanceFields(value.getClass()))
```
labkey-matthewb: keep a Set<Class> for classes we've already inspected?

labkey-matthewb (approved):

api/src/org/labkey/api/cache/CacheManager.java
labkey-adam: I like that idea

labkey-adam: Complete revamp, including recording inspected classes

labkey-matthewb:
Quick manual testing looks good.  I did notice the the log entries created in the catch blocks end up looking a bit mysterious.  A longer message might be be helpful.

PR #6659, opened 2025-05-12
ExpSearchable: common implementation for indexable experiment objects

labkey-nicka:
#### Rationale
Coalesce logic and behaviors for search indexing experiment objects.

#### Changes
- Introduce new implementation interface in `ExpSearchable`
- Refactor usages to produce a search document and absolve them of adding task resources
- Avoid NPEs in index resolution when the results of `getContainer()` is null.

### DIFF
--- a/api/src/org/labkey/api/exp/api/ExpMaterial.java
+++ b/api/src/org/labkey/api/exp/api/ExpMaterial.java
@@ -27,7 +27,7 @@ import java.util.Date;
 import java.util.Map;
 
 /** Represents a physical object in the experiment data model - typically a sample or specimen */
-public interface ExpMaterial extends ExpRunItem
+public interface ExpMaterial extends ExpRunItem, ExpSearchable
 {
     String DEFAULT_CPAS_TYPE = "Material";
     String MATERIAL_INPUT_PARENT = "MaterialInputs";
@@ -83,5 +83,4 @@ public interface ExpMaterial extends ExpRunItem
     Date getMaterialExpDate();
 
     ActionURL detailsURL(Container container, boolean checkForOverride);
-
 }
--- a/api/src/org/labkey/api/exp/api/ExpMaterial.java
+++ b/api/src/org/labkey/api/exp/api/ExpMaterial.java
@@ -27,7 +27,7 @@ import java.util.Date;
 import java.util.Map;
 
 /** Represents a physical object in the experiment data model - typically a sample or specimen */
-public interface ExpMaterial extends ExpRunItem, ExpSearchable
+public interface ExpMaterial extends ExpRunItem
 {
     String DEFAULT_CPAS_TYPE = "Material";
     String MATERIAL_INPUT_PARENT = "MaterialInputs";
--- a/experiment/src/org/labkey/experiment/api/ExpDataImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExpDataImpl.java
@@ -723,7 +723,7 @@ public class ExpDataImpl extends AbstractRunItemImpl<Data> implements ExpData
             identifiersHi.addAll(aliases);
         }
 
-        ExpDataClassImpl dc = getDataClass();
+        ExpDataClassImpl dc = getDataClass(User.getSearchUser());
         if (dc != null)
         {
             ActionURL show = new ActionURL(ExperimentController.ShowDataClassAction.class, container).addParameter("rowId", dc.getRowId());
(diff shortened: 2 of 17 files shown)


experiment/src/org/labkey/experiment/api/ExpDataImpl.java
```diff
@@ -679,29 +679,13 @@ public static Map<String, ExpData> fromDocumentIds(Collection<String> resourceId
...
-        if (task == null)
-        {
-            SearchService ss = SearchService.get();
-            if (null == ss)
-                return;
-            task = ss.defaultTask();
-        }
-
-        var expScope = DbSchema.get("exp", DbSchemaType.Module).getScope();
-        var doc = expScope.executeWithRetryReadOnly((tx) -> createDocument(tableInfo));
```
labkey-nicka, line 693: This is the only place that was doing this kind of optimization to retry. @labkey-matthewb do you have thoughts on why this would be particular to `ExpData` and not other types? I can reintroduce if desired.

mbellew: Probably based on actual Mothership deadlock exceptions.

api/src/org/labkey/api/exp/api/ExpMaterial.java
```diff
@@ -27,7 +27,7 @@
 import java.util.Map;
 
 /** Represents a physical object in the experiment data model - typically a sample or specimen */
-public interface ExpMaterial extends ExpRunItem
+public interface ExpMaterial extends ExpRunItem, ExpSearchable
```
labkey-jeckels: Could make `ExpRunItem` extend `ExpSearchable` instead of having the two subinterfaces separately extend it.

experiment/src/org/labkey/experiment/api/ExpDataImpl.java
labkey-jeckels: That's my guess too. I think we should retain it, though it would be fine to also do the same for the other types.

experiment/src/org/labkey/experiment/api/ExpDataImpl.java
```diff
@@ -755,10 +739,10 @@ public WebdavResource createDocument(@Nullable ExpDataClassDataTableImpl tableIn
             body.append(dc.getName());
         }
 
-        if (tableInfo != null)
+        if (tableInfo instanceof ExpDataClassDataTableImpl expDataClassDataTable)
```
labkey-jeckels: Should we throw if it's not this subclass?

experiment/src/org/labkey/experiment/api/ExpMaterialImpl.java
```diff
@@ -458,10 +398,10 @@ public WebdavResource createIndexDocument(@Nullable ExpMaterialTableImpl tableIn
         if (null != getSampleType())
         {
             if (tableInfo == null)
-                tableInfo = (ExpMaterialTableImpl) QueryService.get().getUserSchema(User.getSearchUser(), container, SCHEMA_SAMPLES).getTable(getSampleType().getName());
+                tableInfo = QueryService.get().getUserSchema(User.getSearchUser(), container, SCHEMA_SAMPLES).getTable(getSampleType().getName());
 
-            if (tableInfo != null)
-                getCustomIndexValues(props, tableInfo, identifiersHi, keyworksHi, jsonData);
+            if (tableInfo instanceof ExpMaterialTableImpl expMaterialTable)
```
labkey-jeckels: Should we throw if it's not this subclass?

labkey-jeckels (approved):

experiment/src/org/labkey/experiment/api/ExpDataImpl.java
labkey-nicka: Alright, I've switched it to do this all the time.

api/src/org/labkey/api/exp/api/ExpMaterial.java
labkey-nicka: Done.

experiment/src/org/labkey/experiment/api/ExpDataImpl.java
labkey-nicka: Good idea. Now throw an `IllegalArgumentException`.

experiment/src/org/labkey/experiment/api/ExpMaterialImpl.java
labkey-nicka: Good idea. Now throw an `IllegalArgumentException`.

PR #6668, opened 2025-05-14
remove method ColumnInfo.getJdbcRsName()

labkey-matthewb:
#### Rationale
This method is redundant.

#### Related Pull Requests
- 

#### Changes
-


api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
```diff
@@ -223,13 +223,6 @@ protected static Map<String,Pair<ColumnInfo,MatchType>> _createTableMap(TableInf
             targetAliasesMap.put(tsvQueryColumnName, new Pair<>(col, MatchType.tsvColumn));
         }
 
-        // should this be under the useImportAliases flag???
```
labkey-matthewb, line 226: This is the only change that gives me pause.  I hope we're not relying on this anywhere, but ...

labkey-martyp (approved):
I ran a ton of TC tests on this and did not see any new failures

PR #6670, opened 2025-05-14
Enable a strong enforce CSP by default

labkey-adam:
#### Rationale
Forward CSP reports to labkey.org if optional feature is turned on

#### Related Pull Requests
- https://github.com/LabKey/premiumModules/pull/258
- https://github.com/LabKey/server/pull/1065

### DIFF
--- a/core/src/org/labkey/core/admin/AdminController.java
+++ b/core/src/org/labkey/core/admin/AdminController.java
@@ -11969,12 +11969,15 @@ public class AdminController extends SpringActionController
                         String path = new URLHelper(urlString).deleteParameters().getURIString();
                         if (null == reports.put(path, Boolean.TRUE) || _log.isDebugEnabled())
                         {
-                            // Forwarded reports already have user, ip, user-agent, etc. from the forwarding server
+                            // Don't modify forwarded reports; they already have user, ip, user-agent, etc. from the forwarding server.
                             boolean forwarded = jsonObj.optBoolean("forwarded", false);
                             if (!forwarded)
                             {
                                 jsonObj.put("user", getUser().getEmail());
-                                jsonObj.put("ip", request.getRemoteAddr());
+                                String ipAddress = request.getHeader("X-FORWARDED-FOR");
+                                if (ipAddress == null)
+                                    ipAddress = request.getRemoteAddr();
+                                jsonObj.put("ip", ipAddress);
                                 if (isNotBlank(userAgent))
                                     jsonObj.put("user-agent", userAgent);
                                 String labkeyVersion = request.getParameter("labkeyVersion");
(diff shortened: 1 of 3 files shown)


core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -11955,19 +11966,61 @@ public Object execute(SimpleApiJsonForm form, BindException errors) throws Excep
...
+                                    .header("Content-Type", request.getContentType()) // Use whatever the browser set
+                                    .POST(HttpRequest.BodyPublishers.ofString(jsonObj.toString(2)))
+                                    .build();
+
+                                // Send the request and get the response
+                                HttpResponse<String> response = client.send(remoteRequest, HttpResponse.BodyHandlers.ofString());
+
+                                if (response.statusCode() != 200)
+                                {
+                                    _log.error("ContentSecurityPolicy report forwarding to https://www.labkey.org failed: {}\n{}", response.statusCode(), response.body());
```
labkey-matthewb, line 12015: log limiting?  (especially since it is ERROR level)

labkey-matthewb (approved):

core/src/org/labkey/core/admin/AdminController.java
```diff
@@ -11955,19 +11966,61 @@ public Object execute(SimpleApiJsonForm form, BindException errors) throws Excep
...
-                                jsonObj.put("labkeyVersion", labkeyVersion);
-                            String cspVersion = request.getParameter("cspVersion");
-                            if (null != cspVersion)
-                                jsonObj.put("cspVersion", cspVersion);
+                            // Forwarded reports already have user, ip, user-agent, etc. from the forwarding server
+                            boolean forwarded = jsonObj.optBoolean("forwarded", false);
+                            if (!forwarded)
+                            {
+                                jsonObj.put("user", getUser().getEmail());
+                                jsonObj.put("ip", request.getRemoteAddr());
```
labkey-jeckels: We probably want the X-FORWARDED-FOR header if it's been set, and can call back on request.getRemoteAddr(). See AuthenticationManager, line 1014

core/src/org/labkey/core/admin/AdminController.java
labkey-jeckels: This is already limited by the check to see if we've received a report for the URL recently. Maybe that's enough

labkey-jeckels (approved):

core/src/org/labkey/core/admin/AdminController.java
labkey-adam: Done

core/src/org/labkey/core/admin/AdminController.java
labkey-adam: Fair point. I have a shelved my change that adds a Throttle for forwarding errors. I can revive that if it turns out the current code is too noisy.

PR #6673, opened 2025-05-15
Issue 53027: Cannot insert a row into a sample type if Allowed File Extensions set, no file is provided

labkey-jeckels:
#### Rationale
We are being too picky about file uploads.

#### Changes
- Skip over file form elements where no file was selected
- Stop asserting about file extensions in makeLegalName(), which isn't necessarily creating a file
- Improve error messaging when an illegal file extension is uploaded

### DIFF
--- a/api/src/org/labkey/api/premium/DefaultAVMultipartResolver.java
+++ b/api/src/org/labkey/api/premium/DefaultAVMultipartResolver.java
@@ -3,8 +3,10 @@ package org.labkey.api.premium;
 import jakarta.servlet.ServletException;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.Part;
+import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 import org.labkey.api.util.FileUtil;
+import org.labkey.api.view.UnauthorizedException;
 import org.springframework.web.multipart.MultipartException;
 import org.springframework.web.multipart.MultipartHttpServletRequest;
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/premium/DefaultAVMultipartResolver.java
```diff
@@ -22,10 +24,17 @@ public class DefaultAVMultipartResolver extends StandardServletMultipartResolver
...
-                    FileUtil.checkAllowedFileName(part.getSubmittedFileName(), true);
-                    validate(part);
+                    try
+                    {
+                        FileUtil.checkAllowedFileName(part.getSubmittedFileName(), true);
+                        validate(part);
+                    }
+                    catch (IOException e)
+                    {
+                        throw new UnauthorizedException(e.getMessage());
```
labkey-susanh: `UnauthorizedException` seems not quite right here. Perhaps `InvalidFileException` would be better?

labkey-susanh (commented):
Are there tests that need to be added in conjunction with this?

api/src/org/labkey/api/premium/DefaultAVMultipartResolver.java
labkey-jeckels: We need something that's a RuntimeException and skips mothership logging. I introduced a new subclass of ApiUsageException

labkey-jeckels:
> Are there tests that need to be added in conjunction with this?

I renamed to match the branch name where Dan's adding test coverage and found this bug. New PR: https://github.com/LabKey/platform/pull/6677

PR #6677, opened 2025-05-16
Issue 53027: Cannot insert a row into a sample type if Allowed File Extensions set, no file is provided

labkey-jeckels:
#### Rationale
We are being too picky about file uploads.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2428

#### Changes
- Skip over file form elements where no file was selected
- Stop asserting about file extensions in makeLegalName(), which isn't necessarily creating a file
- Improve error messaging when an illegal file extension is uploaded
- Minor code cleanup


labkey-adam (approved):
Not for this PR, but should we consider a site validator or similar that checks existing attachments for undesired extensions?

labkey-jeckels:
> Not for this PR, but should we consider a site validator or similar that checks existing attachments for undesired extensions?

I added a new metric, `modules.Core.allowListCounts.FileExtension`, so we can monitor the adoption of this feature. There's probably a lot that people would want to fully lock things down.

Note that I had to remove the assert that was checking for allowed extensions when writing files to the server. Otherwise things like folder exports quickly failed unless you added all of the various extensions they'd produce.

PR #6678, opened 2025-05-17
Escalate SchemaTableInfoCache problems to error-level logging

labkey-jeckels:
#### Rationale
We held off on making this error-level logging last year. Based on initial testing, it's safe to turn up the volume.

Issue 49670: Clean up FKs referencing non-existent tables

Additionally, we can add logging to help identify problems with lookups.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5183

#### Changes
- Switch to a log level that will cause our tests to fail if it's triggered
- `DEBUG` level logging in `QueryForeignKey` when a requested lookup isn't available


labkey-jeckels:
Full EHR suites queued:

https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_EhrPostgres/3515058

https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_EhrSqlserver/3515056

labkey-susanh (approved):

assay/src/org/labkey/assay/plate/query/HitTable.java
```diff
@@ -37,6 +39,14 @@ public HitTable(PlateSchema schema, @Nullable ContainerFilter cf, boolean allowI
         _allowInsertUpdate = allowInsertUpdate;
     }
 
+    @Override
+    public void addColumns()
+    {
+        super.addColumns();
+        getMutableColumnOrThrow("RunId").setFk(new QueryForeignKey.Builder(getUserSchema(), getContainerFilter()).schema(ExpSchema.SCHEMA_NAME).table(ExpSchema.TableType.Runs));
```
labkey-nicka, line 47: Now that you've specified these here should/can they be removed from the `Hit.query.xml` defined in limsModules?

https://github.com/LabKey/limsModules/blob/d9ba7eaf18705d31d4cc9184f05c345c2a1c7764/biologics/resources/queries/plate/Hit.query.xml

labkey-nicka: Looks like it's not complete overlap but just wanted to call this out.

labkey-jeckels: Yes. https://github.com/LabKey/limsModules/pull/1421

labkey-adam (approved):

PR #6686, opened 2025-05-21
Move study design features into a new module

labkey-klum:
#### Rationale
Previously, the study design features (and schema) were behind a deprecated features flag. This work will move most (but not all) code into a new module : `StudyDesign` so that it can be added to the distribution of the single client which uses this feature. 

The overall strategy was to move as much code into the new module as reasonably possible, create/use new interfaces to replace direct access to study classes, and for all others just use the active module to expose/hide functionality. Existing tests which exercised study design functionality should be fixed in this or related PRs.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6686
- https://github.com/LabKey/clientModules/pull/23
- https://github.com/LabKey/testAutomation/pull/2459
- https://github.com/LabKey/customModules/pull/227

#### Changes
- New `StudyDesign` module .
- `Web Parts` and web resources moved to the new module.
- A few treatment and assay schedule unit tests moved.
- `StudyDesignController` moved.
- The schema still remains in the study module but is exposed based on the active module.
- Study design related writers and importers couldn't be easily moved, they remain in the study module but will `noop` if the module is not active. Due to the way internal study writers/importers are registered, I couldn't find an easy way to hide them in the import/export options for a container, this is unfortunate but seems harmless.
-


study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -2528,6 +2536,51 @@ public void addNavTrail(NavTree root)
         }
     }
 
+    /**
+     * Called from the vaccine design webpart for the study design module
+     */
+    @RequiresPermission(UpdatePermission.class)
+    public class CreateVisitVaccineDesign extends MutatingApiAction<VisitForm>
```
cnathe: minor: the name here threw me off for a bit. suggestion `CreateVisitForVaccineDesign` or `CreateVisitFromVaccineDesign`

study/src/org/labkey/study/controllers/StudyController.java
```diff
@@ -2528,6 +2536,51 @@ public void addNavTrail(NavTree root)
...
+    /**
+     * Called from the vaccine design webpart for the study design module
+     */
+    @RequiresPermission(UpdatePermission.class)
+    public class CreateVisitVaccineDesign extends MutatingApiAction<VisitForm>
+    {
+        @Override
+        public void validateForm(VisitForm form, Errors errors)
+        {
+            Study study = getStudy(getContainer());
```
cnathe, line 2555: I see that this action was moved from StudyDesignController, any reason to have a check here for `StudyDesignManager.get().isModuleActive()`?

study/src/org/labkey/study/model/StudyManager.java
```diff
@@ -2721,11 +2654,8 @@ public void deleteAllStudyData(Container c, User user)
             //
             // assay schedule
             //
-            Table.delete(SCHEMA.getTableInfoAssaySpecimenVisit(), containerFilter);
-            assert deletedTables.add(SCHEMA.getTableInfoAssaySpecimenVisit());
-            Table.delete(_assaySpecimenHelper.getTableInfo(), containerFilter);
```
cnathe, line 2726: there were two Table.delete calls here before: AssaySpecimenVisit and AssaySpecimen. Is that second one no longer needed or happening somewhere else now?

study/src/org/labkey/study/view/manageStudy.jsp
```diff
@@ -254,31 +252,30 @@
...
                                 .addReturnUrl(getActionURL());
                         %>
                         <td><%= link("Manage Treatments", manageTreatmentsURL) %></td>
                     </tr>
                     <tr>
                         <td class="lk-study-prop-label">Assay Schedule</td>
-                        <td class="lk-study-prop-desc">This study defines <%= getAssaySpecimenConfigs().size() %> assay configurations</td>
+                        <td class="lk-study-prop-desc">This study defines <%= StudyDesignService.get().getAssaySpecimenConfigs(getContainer()).size() %> assay configurations</td>
                         <%
-                            boolean hasRhoModule = getContainer().getActiveModules().contains(ModuleLoader.getInstance().getModule("rho"));
```
cnathe, line 280: Rho module? wow, that's been awhile!

studydesign/module.properties
```diff
@@ -0,0 +1,5 @@
+ModuleClass: org.labkey.studydesign.StudyDesignModule
+License: Apache 2.0
+LicenseURL: http://www.apache.org/licenses/LICENSE-2.0
+SupportedDatabases: mssql, pgsql
```
cnathe: if this is only going to the Atlas distro, can it be pgsql only?

cnathe (approved):

study/src/org/labkey/study/controllers/StudyController.java
labkey-klum: Let's go with `CreateVisitForVaccineDesign`

study/src/org/labkey/study/controllers/StudyController.java
labkey-klum: Good point.

study/src/org/labkey/study/model/StudyManager.java
labkey-klum: Hmmm, good point. I think I only meant to delete the cache clearing line but that table should still be deleted. I'm going to move both deletes into `StudyDesignManager.deleteAllStudyData` so it can get cleaned up with the rest of the study design tables.

studydesign/module.properties
labkey-klum: Yes!

cnathe:
@labkey-klum I finished manual testing. I verified study export and publish with and without the studydesign module deployed to my localhost server. 

The only thing I noticed is that the study schema looks to show the `AssaySpecimen` and `AssaySpecimenVisit` tables always. I believe those are part of the StudyDesign set of tables. It looks like the folder writers/importers for those tables are within the AssayScheduleImporter/AssayScheduleWriter and the tableInfo is in StudyDesignSchema. Should those be moved into the `StudyDesignQuerySchema`?

PR #6687, opened 2025-05-21
Issue 53050: Option to send list of active users to labkey.org

labkey-adam:
#### Rationale
Help manage multi-server contracts by sending a list of active user email addresses along with usage metrics


labkey-jeckels (approved):
Moving the registration of the flag should be simple enough

labkey-jeckels:
I did a quick manual test. All worked as expected.

PR #6690, opened 2025-05-22
Issue 52824: Don't mutate domains from the cache

labkey-susanh:
#### Rationale
Issue [52824](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52824) - We get occasional exceptions reported to Mothership with error messages like "java.lang.RuntimeException: Duplicate property descriptor name found for: c2d33_donor.SerologyInformation".  The stack traces generally indicate they are coming from some background job (search indexer or ETL, for example). This appears to be happening because our methods for updating domains draw objects from the cache and then manipulate those cached objects before saving them. If some other thread extracts the domain from the cache while it is being manipulated (to rearrange the fields, for example) that other thread can get an invalid domain.

This PR adds an `isMutable` property to domains, which is checked when a domain is saved and throws an exception if it is true. (Note that domains already have a different notion of `locked` so that was not used here.) It would probably be best if we threw an exception from the setter methods for domains as well, but since domains are complex objects that contain other objects that could be mutated, that is not an entirely simple thing, and this PR doesn't go that far. 

There is the possibility that this change will be introducing some inefficiencies since we are deliberately bypassing the cache, but hope they won't be noticeable since we should be getting the immutable domain objects much more often than mutable ones.
 

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1424
- https://github.com/LabKey/ehrModules/pull/962
- https://github.com/LabKey/wnprc-modules/pull/828
- https://github.com/LabKey/DiscvrLabKeyModules/pull/353
- https://github.com/LabKey/OConnorLabModules/pull/165
- https://github.com/LabKey/clientModules/pull/25
- https://github.com/LabKey/commonAssays/pull/891
- https://github.com/LabKey/internalModules/pull/81

#### Changes
- add `getDomain(boolean forUpdate)` and `getDomainDescriptor(int id, boolean forUpdate)` and many other similar overloading methods for getting domains
- update `Domain` class with `isMutable` property


api/src/org/labkey/api/exp/OntologyManager.java
```diff
@@ -2441,25 +2441,47 @@ public DomainDescriptor load(@NotNull Integer key, @Nullable Object argument)
...
     public static DomainDescriptor getDomainDescriptor(int id)
     {
-        return DOMAIN_DESC_BY_ID_CACHE.get(id);
+        return getDomainDescriptor(id, false);
+    }
+
+    public static DomainDescriptor getDomainDescriptor(int id, boolean forUpdate)
+    {
+        DomainDescriptor dd = DOMAIN_DESC_BY_ID_CACHE.get(id);
+        if (forUpdate)
```
labkey-susanh, line 2451: This still leaves a little window for racing if someone gets `dd` before the removal happens. Probably better to not get from the cache at all.

mbellew: I thought I commented, but maybe didn't save?  Anyway, can we use .clone() instead of removing from the cache?

mbellew (changes requested):

api/src/org/labkey/api/exp/OntologyManager.java
labkey-susanh: I'm going the route of not pulling from the cache if it's for update.

api/src/org/labkey/api/assay/AssayProvider.java
```diff
@@ -99,10 +99,16 @@ enum ReRunSupport
 
     Domain getBatchDomain(ExpProtocol protocol);
```
labkey-jeckels, line 101: Worth a comment to say the default `forUpdate` is false?

api/src/org/labkey/api/data/SchemaTableInfo.java
```diff
@@ -857,6 +857,13 @@ public Domain getDomain()
         return null;
     }
 
+    @Nullable
+    @Override
+    public Domain getDomain(boolean forUpdate)
+    {
+        return getDomain();
```
labkey-jeckels, line 866: Maybe this should throw `UnsupportedOperationException` when passed true?

api/src/org/labkey/api/exp/property/PropertyService.java
```diff
@@ -55,9 +55,15 @@ static void setInstance(PropertyService impl)
     @Nullable
     Domain getDomain(Container container, String domainURI);
```
labkey-jeckels, line 56: Maybe comments here about the default value too?

experiment/src/org/labkey/experiment/api/ExpDataClassImpl.java
```diff
@@ -230,19 +230,33 @@ public long getDataCount(Container c, @Nullable ContainerFilter cf)
...
-        _domain = d;
+        if (_domain == null || (forUpdate && !_domain.isMutable()))
+        {
+            _domain = PropertyService.get().getDomain(getContainer(), getLSID(), forUpdate);
+            if (_domain == null)
+            {
+                _domain = PropertyService.get().createDomain(getContainer(), getLSID(), getName());
+                try
+                {
+                    _domain.save(null);
```
labkey-jeckels, line 251: Should this return an immutable copy if `!forUpdate`?

api/src/org/labkey/api/query/SimpleUserSchema.java
```diff
@@ -457,14 +457,20 @@ protected Container getDomainContainer()
...
         {
             if (_objectUriCol == null)
                 return null;
 
-            if (_domain == null)
+            if (_domain == null || (forUpdate && !_domain.isMutable()))
             {
                 String domainURI = getDomainURI();
-                _domain = PropertyService.get().getDomain(getDomainContainer(), domainURI);
+                _domain = PropertyService.get().getDomain(getDomainContainer(), domainURI, forUpdate);
```
labkey-jeckels: It seems wrong to stash a mutable copy and return it from `getDomain(false)`

labkey-jeckels (approved):
Minor suggestions.

I wonder if we should do a followup PR and get rid of some or many of the `getDomain()` methods that don't take the `forUpdate` argument. Or perhaps keep it, but make it a `final` method that just passes through to the overloaded version. I've seen lots of cases where subclass overrides get inconsistent for overloaded methods with defaults.

api/src/org/labkey/api/assay/AssayProvider.java
labkey-susanh: Done

api/src/org/labkey/api/data/SchemaTableInfo.java
labkey-susanh: Done.

api/src/org/labkey/api/exp/property/PropertyService.java
labkey-susanh: Done.

api/src/org/labkey/api/query/SimpleUserSchema.java
labkey-susanh: Yeah, it's a little unsavory, but currently also not really harmful. I was thinking to save the extra query, but I'm sure the unsavory part will be more bothersome in the future. Updated.

experiment/src/org/labkey/experiment/api/ExpDataClassImpl.java
labkey-susanh: Done.

PR #6699, opened 2025-05-27
Issue 52900: LKSM: Moving a sample to another folder breaks other references to shared files

XingY:
#### Rationale
Issue [52900](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52900): LKSM: Moving a sample to another folder breaks other references to shared files
Issue [53070](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53070): Moving files could result in duplicate exp.data records
Issue [53178](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53178): SQL generation problem from TableUpdaterFileListener

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1801
- https://github.com/LabKey/platform/pull/6699
- https://github.com/LabKey/limsModules/pull/1431
- https://github.com/LabKey/commonAssays/pull/890

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/files/TableUpdaterFileListener.java
+++ b/api/src/org/labkey/api/files/TableUpdaterFileListener.java
@@ -198,10 +198,10 @@ public class TableUpdaterFileListener implements FileListener
 
     public String getSourceSelect(SqlDialect sqlDialect)
     {
-        String schema  = sqlDialect.makeLegalIdentifier(_table.getSchema().getName());
-        String table = sqlDialect.makeLegalIdentifier(_table.getName());
-        String column = sqlDialect.makeLegalIdentifier(_pathColumn.getName());
-        return sqlDialect.getStringHandler().quoteStringLiteral(schema + "." + table + "." + column);
+        String schema  = sqlDialect.getStringHandler().quoteStringLiteral(_table.getSchema().getName());
+        String table = sqlDialect.getStringHandler().quoteStringLiteral(_table.getName());
+        String column = sqlDialect.getStringHandler().quoteStringLiteral(_pathColumn.getName());
+        return schema + "." + table + "." + column;
     }
 
     @Override
@@ -401,7 +401,7 @@ public class TableUpdaterFileListener implements FileListener
             selectFrag.append("  NULL AS SourceKey,\n");
 
         //selectFrag.append("  ? AS SourceName\n").add(getName());
-        selectFrag.append("  ").append(getSourceSelect(_table.getSchema().getSqlDialect())).append(" AS SourceName\n");
+        selectFrag.append("  ").appendValue(getSourceSelect(_table.getSchema().getSqlDialect())).append(" AS SourceName\n");
 
         selectFrag.append("FROM ").append(_table, TABLE_ALIAS).append("\n");
         selectFrag.append("WHERE ").appendIdentifier(_pathColumn.getSelectIdentifier());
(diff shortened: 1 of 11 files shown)


api/src/org/labkey/api/files/TableUpdaterFileListener.java
```diff
@@ -195,6 +196,14 @@ public String getSourceName()
         return _table.getSchema().getName() + "." + _table.getName() + "." + _pathColumn.getName();
     }
 
+    public String getSourceSelect(SqlDialect sqlDialect)
+    {
+        String schema  = sqlDialect.makeLegalIdentifier(_table.getSchema().getName());
```
XingY: @labkey-matthewb Any suggestion how to avoid using the deprecated `makeLegalIdentifier` here is appreciated.

labkey-nicka: Agreed. It is helpful when deprecating methods to give indication as to what should be used/considered instead.

api/src/org/labkey/api/files/TableUpdaterFileListener.java
```diff
@@ -393,10 +401,15 @@ else if (_table.getColumn("Folder") != null)
             selectFrag.append("  NULL AS SourceKey,\n");
 
         //selectFrag.append("  ? AS SourceName\n").add(getName());
-        selectFrag.append("  ").append(_table.getSchema().getSqlDialect().getStringHandler().quoteStringLiteral(getSourceName())).append(" AS SourceName\n");
```
labkey-matthewb, line 396: I'm not exactly sure what's going on here (selecing a bit of SQL as a string value?), but can we just do .appendValue(getSourceName())?

api/src/org/labkey/api/files/TableUpdaterFileListener.java
labkey-matthewb: This doesn't really have anything to do with SQLFragment and identifiers, if we're just trying to create a string literal to select.  Just create the string you want and use appendValue() instead of append()

labkey-matthewb (changes requested):

api/src/org/labkey/api/files/TableUpdaterFileListener.java
labkey-matthewb: Are using this selected string to generate SQL later (or is this just for logging/display)?  If so it's probably easier to select the schema and table name separately.

api/src/org/labkey/api/files/TableUpdaterFileListener.java
```diff
@@ -195,6 +196,14 @@ public String getSourceName()
         return _table.getSchema().getName() + "." + _table.getName() + "." + _pathColumn.getName();
     }
 
+    public String getSourceSelect(SqlDialect sqlDialect)
+    {
+        String schema  = sqlDialect.makeLegalIdentifier(_table.getSchema().getName());
+        String table = sqlDialect.makeLegalIdentifier(_table.getName());
+        String column = sqlDialect.makeLegalIdentifier(_pathColumn.getName());
+        return sqlDialect.getStringHandler().quoteStringLiteral(schema + "." + table + "." + column);
```
labkey-matthewb: Maybe use 

// for Db schema table info
return table.getSelectName() + "." + LabKeySql.quoteString(_pathColumn.getName())

// for UserSchema
return table.getUserSchema().getSchemaPath().toSQLString() + "." + LabKeySql.quoteString(_table.getName()) + "." + LabkeySql.quoteString(_pathColumn.getName()

and then appendValue() as mentioned above.

XingY: Thanks. This works.

labkey-matthewb (approved):

PR #6702, opened 2025-05-27
Issue 53167: LKS Chart wizard page can give JS error when selecting Line chart type if LABKEY object doesn't have core moduleContext

cnathe:
#### Rationale
Issue [53164](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53164): LKS new assay design page gives JS error if no specialty assay providers available for server
Issue [53167](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53167): LKS Chart wizard page can give JS error when selecting Line chart type if LABKEY object doesn't have core moduleContext

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1802

#### Changes
- chartTypePanel.js to add null check for core moduleContext

### DIFF
--- a/visualization/resources/web/vis/chartWizard/chartTypePanel.js
+++ b/visualization/resources/web/vis/chartWizard/chartTypePanel.js
@@ -901,7 +901,7 @@ Ext4.define('LABKEY.vis.ChartTypeFieldSelectionsPanel', {
             // special case for trendline field, only visible for LIMS+ when premium module is available
             let hidden = false;
             if (field.name === 'trendline') {
-                const hasChartBuilding = LABKEY.getModuleContext('core').productFeatures?.indexOf('ChartBuilding') !== -1;
+                const hasChartBuilding = LABKEY.getModuleContext('core')?.productFeatures?.indexOf('ChartBuilding') !== -1;
                 const hasPremium = LABKEY.getModuleContext('api').moduleNames.indexOf('premium') !== -1;
                 hidden = !hasPremium || !hasChartBuilding;
             }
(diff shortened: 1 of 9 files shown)


visualization/resources/web/vis/chartWizard/chartTypePanel.js
```diff
@@ -901,7 +901,7 @@ Ext4.define('LABKEY.vis.ChartTypeFieldSelectionsPanel', {
             // special case for trendline field, only visible for LIMS+ when premium module is available
             let hidden = false;
             if (field.name === 'trendline') {
-                const hasChartBuilding = LABKEY.getModuleContext('core').productFeatures?.indexOf('ChartBuilding') !== -1;
+                const hasChartBuilding = LABKEY.getModuleContext('core')?.productFeatures?.indexOf('ChartBuilding') !== -1;
```
labkey-susanh, line 904: Fix seems reasonable enough, but when would we not have the `core` module context? Is this just a temporary state?

labkey-susanh (approved):

visualization/resources/web/vis/chartWizard/chartTypePanel.js
cnathe: I was shocked that we would have a scenario that didn't have the core module context, but the issues container on labkey.org doesn't have it here: https://www.labkey.org/home/Developer/issues/project-begin.view. Looking at the folder management page, it is of type Collaboration with a pretty standard set of modules enabled. The only thing I noticed when comparing it to other collab folders is that it has the "Portal" module unchecked.

PR #6708, opened 2025-05-28
Issue 52684: Update Log4j plugin registration

labkey-jeckels:
#### Rationale
Log4J deprecated the approach we were using for plugin registration and started complaining about conflicting buffering configuration. Since we need to refactor how Log4J gets onto the class path, we can use this as an opportunity to share more extraction code with our tests.

#### Related Pull Requests
- https://github.com/LabKey/server/pull/1069

#### Changes
* Invoke `EmbeddedExtractor` code through reflection
* Remove now redundant extraction code
* Minor code cleanup


pipeline/src/org/labkey/pipeline/api/PipelineServiceImpl.java
```diff
@@ -457,78 +451,37 @@ public boolean isEnterprisePipeline()
...
-
     @Override
-    public List<String> getClusterStartupArguments() throws IOException
+    public List<String> getClusterStartupArguments()
     {
         List<String> args = new ArrayList<>();
         args.add(System.getProperty("java.home") + "/bin/java" + (SystemUtils.IS_OS_WINDOWS ? ".exe" : ""));
-        File labkeyBootstrap = new File(new File(new File(System.getProperty("catalina.home")), "lib"), "labkeyBootstrap.jar");
-        File servletApi = null;
+        File labkeyBootstrap = new File(new File(System.getProperty("catalina.home")), "labkeyBootstrap.jar");
```
labkey-susanh, line 459: I'm confused by this change. Seems like `labkeyBootstrap.jar` is still in a `lib` directory, or am I looking in the wrong place?

labkey-jeckels: image

Here's the layout on my dev machine. The code isn't deleting any previously extracted files. Is it an older copy, maybe?

labkey-susanh (approved):
Have more clarity after our chat.

PR #6714, opened 2025-05-29
Issue 52886: Assure detailed audit log captures differences for long field names

labkey-susanh:
#### Rationale
Issue [52886](https://www.labkey.org/home/Developer/issues/issues-update.view?issueId=52886)

#### Changes
- Update `getRow` method for data classes to return a map with stringified field keys as keys instead of database column names
- Add JSP unit test
- Add `FieldKeyRowMap.toStringMap` static utility method

### DIFF
--- a/api/src/org/labkey/api/data/FieldKeyRowMap.java
+++ b/api/src/org/labkey/api/data/FieldKeyRowMap.java
@@ -125,7 +125,7 @@ public class FieldKeyRowMap implements Map<FieldKey, Object>
         return map;
     }
 
-    public static Map<String, Object> toColumnNameMap(Map<FieldKey, Object> rowMap)
+    public static Map<String, Object> toNameMap(Map<FieldKey, Object> rowMap)
     {
         Map<String, Object> map = new CaseInsensitiveHashMap<>();
         rowMap.forEach((key, value) -> {
--- a/api/src/org/labkey/api/data/FieldKeyRowMap.java
+++ b/api/src/org/labkey/api/data/FieldKeyRowMap.java
@@ -16,6 +16,7 @@
 package org.labkey.api.data;
 
 import org.jetbrains.annotations.NotNull;
+import org.labkey.api.collections.CaseInsensitiveHashMap;
 import org.labkey.api.query.FieldKey;
 
 import java.sql.SQLException;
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/data/FieldKeyRowMap.java
```diff
@@ -124,6 +125,19 @@ public void clear()
...
 
+    public static Map<String, Object> toStringMap(Map<FieldKey, Object> rowMap)
+    {
+        Map<String, Object> map = new CaseInsensitiveHashMap<>();
+        rowMap.forEach((key, value) -> {
+            if (key.getParent() != null)
+                throw new IllegalArgumentException("Multi-part field key '" + key + "' cannot be used as key in string map since it may not be unique.");
+            if (map.containsKey(key.toString()))
+                throw new IllegalArgumentException("Duplicate key '" + key + "' found in fieldKey map.");
+            map.put(key.toString(), value);
```
labkey-matthewb: Why not key.getName() since parent == null?

api/src/org/labkey/api/data/FieldKeyRowMap.java
```diff
@@ -124,6 +125,19 @@ public void clear()
         return map;
     }
 
+    public static Map<String, Object> toStringMap(Map<FieldKey, Object> rowMap)
```
labkey-matthewb: I like this.  I might got even more verbose to disambiguate the possible meaning of toString().  Either toEncodedStringMap() or toNameMap().  See comment below.

labkey-matthewb (changes requested):

api/src/org/labkey/api/data/FieldKeyRowMap.java
labkey-susanh: Good point. Done.

api/src/org/labkey/api/data/FieldKeyRowMap.java
labkey-susanh: I've chosen `toNameMap`. `Encoded` is too ambiguous, at least in my brain.

labkey-matthewb (approved):

cnathe:
Fix verified in fb_auditLongNames branch. Looks good.

PR #6717, opened 2025-05-29
Sort existing allowed file extensions

labkey-nicka:
#### Rationale
A user recently sent a screen cap of this list while triaging and I noticed it was particularly difficult to find a given extension in the list since it was not sorted. This sorts the listing to make extensions easier to find.

Screenshot 2025-05-29 at 3 54 50 PM

#### Changes
- Sort administration page listing of allowed file extensions to make them easier to find

### DIFF
--- a/api/src/org/labkey/api/util/SortHelpers.java
+++ b/api/src/org/labkey/api/util/SortHelpers.java
@@ -17,9 +17,6 @@ package org.labkey.api.util;
 
 import java.util.Comparator;
 
-/**
- * Created by eyounske on 5/10/16.
- */
 public class SortHelpers
 {
     // Natural sort ordering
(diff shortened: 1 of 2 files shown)


labkey-adam (approved):
This JSP is used by the Allowed Redirect Hosts page as well; worth a quick verification that that page also works reasonably with these changes.

labkey-nicka:
> A user recently sent a screen cap of this list while triaging and I noticed it was particularly difficult to find a given extension in the list since it was not sorted. This sorts the listing to make extensions easier to find.

Thanks. I checked it out and this sorting seems logical there as well. I'm going to move ahead with this change.

labkey-jeckels:
@labkey-danield heads up given your work to add test coverage on this feature. Might need to adjust the new test, and/or ensure that it's checking for alphabetization.

PR #6718, opened 2025-05-30
More specific rule to identify signed date/time duration

labkey-martyp:
#### Rationale
When checking potential date and timestamp conversions, only attempt signed duration conversion if the value matches a specific pattern. (sign)(number)(duration unit).

Needs to not match + or -, or temperatures like -80C.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6695

#### Changes
- Use regex in isSignedDuration

### DIFF
--- a/api/src/org/labkey/api/util/DateUtil.java
+++ b/api/src/org/labkey/api/util/DateUtil.java
@@ -1148,7 +1148,7 @@ public class DateUtil
     {
         try
         {
-            if (!durationCandidate.matches("^[+-]\\d+[a-zA-Z].*"))
+            if (!durationCandidate.matches("^[+-]\\d+[pymdhtsPMDHTS].*"))
             {
                 return false;
             }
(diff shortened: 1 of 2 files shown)


labkey-jeckels (changes requested):
As discussed briefly in standup today, please add unit tests to make sure we're not parsing unwanted values going forward. You mentioned +, -, and -80C as good examples.

api/src/org/labkey/api/util/DateUtil.java
```diff
@@ -1150,7 +1148,7 @@ public static boolean isSignedDuration(@NotNull String durationCandidate)
     {
         try
         {
-            if (!(durationCandidate.startsWith("+") || durationCandidate.startsWith("-")))
+            if (!durationCandidate.matches("^[+-]\\d+[pymdhtsPMDHTS].*"))
```
labkey-nicka: Should we compile this matcher for performance?

labkey-nicka: Feels like a comment is warranted about the intention and use case for this expression.

labkey-martyp: good call, added it

labkey-martyp: added

labkey-jeckels (approved):

PR #6721, opened 2025-05-30
Address FileChannel.lock runtime exception

XingY:
#### Rationale
File lock [exception](https://www.labkey.org/_mothership/mothership-showStackTraceDetail.view?exceptionStackTraceId=38415)

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5898

#### Changes
- Catch OverlappingFileLockException (RuntimeException) and throw IOException so it can be handled better downstream


labkey-jeckels:
@XingY does this improve the upload experience or error messaging? Here's my local testing:

Set a breakpoint in FileUtil.copyFile(). Configure it to only suspend the thread. This makes it possible to have two uploads conflict with each other.
In two separate browser tabs, drag the same file into the files web part to upload them.
They'll both hit the breakpoint. Let them continue.

I get two error dialogs in the two browsers and the new IOException logged on the server. I see similar behavior with the original code.

labkey-jeckels (commented):
@XingY @labkey-nicka I don't see changes since I last tested, so I'll ask again: Have I missed a functional change here? The old and the new behavior seemed to be the same, despite the different exception class being used.

labkey-nicka:
I just switched review to you from triage to get this out of everyones views. I don't believe there have been any changes.

XingY:
The PR won't improve user experiment and was aimed at reducing exception noise. The exception cannot be reproduced to validate fix. Closing PR per discussion with @labkey-jeckels. Will keep monitoring for repros.

PR #6724, opened 2025-06-03
Issue 52746: SM Import: Newlines in field names on import

XingY:
#### Related Pull Requests
- 

#### Changes
- disallow newline,  carriage returns and tab characters in field names
- jest api tests


labkey-susanh (approved):
Since this is a change in behavior, you might want to make sure Molly is aware in case we want to include something about this in release notes.

PR #6727, opened 2025-06-03
Fix intermittent test failure on stale password strength indicator

labkey-jeckels:
#### Rationale
When users type in passwords, we make a request to score each of them for strength. The responses may come back out of order, which can lead to showing the wrong strength. This is bad for users and automated tests.

```
org.junit.ComparisonFailure: Strength guidance for password expected:<Very [Strong]> but was:<Very [Weak]>
  at org.junit.Assert.assertEquals(Assert.java:117)
  at org.labkey.test.components.core.login.SetPasswordForm.lambda$0(SetPasswordForm.java:117)
  at org.awaitility.core.AssertionCondition.lambda$new$0(AssertionCondition.java:53)
  at org.awaitility.core.ConditionAwaiter$ConditionPoller.call(ConditionAwaiter.java:248)
```
Example: https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_Ehr_BvtEhrSqlserver/3534802?buildTab=overview&logFilter=debug&logView=flowAware

#### Changes
- Cancel any existing requests when we make a new one to check the latest password

### DIFF
--- a/core/webapp/passwordGauge.js
+++ b/core/webapp/passwordGauge.js
@@ -10,6 +10,7 @@ LABKEY.PasswordGauge = new function() {
     let _emailId;               // the element ID for the email text box
     let _emailAddress;          // optional email address (for the AJAX request)
     let _renderBarFunction;
+    let _pendingScoreRequest;   // The most recent request made to score the proposed password
 
     // private methods
     function _drawOutline(canvas, ctx, ratio) {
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
The easiest way to repro this is to use the debugger to simulate a very slow response. Set a breakpoint in the first line of `LoginController.GetPasswordScoreAction.execute()` with the condition: `form.getPassword() != null && form.getPassword().length() < 2`. Also set the breakpoint to only suspend the one thread.

image

With your password strength requirement set to Strong, start filling out the change password form. Type a long password, letting the debugger stop the request to score the password when it was just one character. Then release the thread in the debugger. Without this patch, the form will show "very weak" as the score. With this patch, it shows whatever the final score actually was.

labkey-nicka (approved):

PR #6732, opened 2025-06-04
Delete participants automated test (plus other study related tests)

labkey-klum:
#### Rationale
Adding automation for recent feature or bug fixes:
- [48158: Delete study participant feature](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48158). I've added a new test in `StudyManageParticipantsTest` which was renamed from : `StudyMergeParticipantsTest`.
- [51208: Creation of 'duplicate' participantIDs with/without leading whitespace is possible using csv import](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51208). This was mostly covered previously by the various unit tests, but I modified them slightly to include more whitespace chars.


labkey-danield (approved):

study/src/org/labkey/study/model/ParticipantIdImportHelper.java
```diff
@@ -165,6 +165,8 @@ public String translateParticipantId(@Nullable Object p) throws ValidationExcept
                 participantId = p.toString();
         }
 
+        participantId = participantId != null ? participantId.trim() : null;
```
labkey-tchad: trim to null? Is `participantId = null` functionally distinct from `participantId = ""`?
```suggestion
        participantId = StringUtils.trimToNull(participantId);
```

study/test/src/org/labkey/test/tests/study/StudyManageParticipantsTest.java
```diff
@@ -1,215 +1,344 @@
...
+
+        cmd.getDomainDesign().setFields(List.of(
+                new FieldDefinition("Name", FieldDefinition.ColumnType.String),
+                new FieldDefinition("IntField", FieldDefinition.ColumnType.Integer),
+                new FieldDefinition("DoubleField", FieldDefinition.ColumnType.Double)
+        ));
+        if ("Demographics".equalsIgnoreCase(datasetName))
+            cmd.setOptions(Map.of("demographics", true));
+        cmd.execute(conn, folderName);
+    }
```
labkey-tchad: Consider making a `DomainProps` subclass for datasets (similar to [`ListDefinition`](https://github.com/LabKey/testAutomation/blob/develop/src/org/labkey/test/params/list/ListDefinition.java) or [`SampleTypeDefinition`](https://github.com/LabKey/testAutomation/blob/develop/src/org/labkey/test/params/experiment/SampleTypeDefinition.java))
Might be useful when we start adding new SDMS tests.

labkey-tchad (approved):

study/src/org/labkey/study/model/ParticipantIdImportHelper.java
labkey-klum: yes, done.

study/test/src/org/labkey/test/tests/study/StudyManageParticipantsTest.java
labkey-klum: Good point, I was not aware of those classes. I'll take a look.

PR #6734, opened 2025-06-04
Don't sort on non-existent column

labkey-matthewb:
#### Rationale
Ignore bad column in sort specification

Issue 53025: BadSqlGrammarException from org.labkey.labbook.NotebookManager.getNotebooks() with invalid sort parameter

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/data/Sort.java
+++ b/api/src/org/labkey/api/data/Sort.java
@@ -491,46 +491,35 @@ public class Sort
                 continue;
             ColumnInfo colinfo = columns.get(fieldKey);
             if (colinfo == null)
+                continue;
+            List<ColumnInfo> sortFields = null;
+            List<FieldKey> sortFieldKeys = colinfo.getSortFieldKeys();
+            if (null != sortFieldKeys)
             {
-                // NOTE Sort.getSortText() passes in dialect==null
-                DatabaseIdentifier id;
-                if (null != dialect)
-                    id = dialect.makeDatabaseIdentifier(fieldKey.getName());
-                else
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
@labkey-matthewb I added this to the board. Let me know if you want a quick tutorial on how to do it - it only takes a couple of clicks.

api/src/org/labkey/api/data/Sort.java
```diff
@@ -491,46 +491,35 @@ public SQLFragment getOrderByClause(SqlDialect dialect, Map<FieldKey, ? extends
                 continue;
             ColumnInfo colinfo = columns.get(fieldKey);
             if (colinfo == null)
+                continue;
```
labkey-jeckels, line 494: Maybe worth a `DEBUG` level log for future troubleshooting?

labkey-jeckels (approved):
Tested fix in the FB (hacking the sort parameter sent from the client to getNotebooks.api via the debugger) and it looks good.

api/src/org/labkey/api/data/Sort.java
labkey-matthewb: That sounds a lot like the DEBUG in MapDataIterator we killed.  No one will ever see it.  Happy to throw instead.

labkey-jeckels: Totally agree that it wouldn't help us track down problems, except when you're specifically trying to troubleshoot behavior and want to confirm a theory by upping the log verbosity. Maybe it's low probability that we'd ever care in this case, but I do occasionally flip prod servers to DEBUG level for individual loggers

PR #6737, opened 2025-06-05
Correct bad merge for DataRegion.renderUpdateForm

labkey-klum:
#### Rationale
Many test failures from the merge forward from yesterday.


labkey-matthewb (commented):
we also lost the try/catch.  I can't remember why those were added, but there is a Results object being held by the RenderContext

labkey-matthewb:
> we also lost the try/catch. I can't remember why those were added, but there is a Results object being held by the RenderContext

I'll try to put this back together

labkey-matthewb (approved):
I see, the getMapCollection() version does not need to the try/catch

PR #6739, opened 2025-06-05
Issue 52547: LKS filter dialog treats many filter types as if they are Equals

XingY:
#### Rationale
The faceted filter in LKS grid has a few flaws. It's overly aggressive in trying to determine checked/unchecked values based on the filter's values, without taking into consideration of the filter operator. For example, "> 4" filter would result in value "4" being selected, which is wrong. This PR doesn't change the logic in selecting values as it's a slightly larger effort that we want to avoid for Ext3. Instead, it makes a small change to default to the "Choose Filter" tab for filter types that are not of equal or IN operator types. 

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6739
- https://github.com/LabKey/testAutomation/pull/2477

#### Changes
- only default to Choose Value tab for equal or IN filter types

### DIFF
--- a/api/webapp/clientapi/ext3/FilterDialog.js
+++ b/api/webapp/clientapi/ext3/FilterDialog.js
@@ -387,7 +387,7 @@ LABKEY.FilterDialog = Ext.extend(Ext.Window, {
                 filters: filters,
                 cacheResults: this.cacheFacetResults,
                 listeners: {
-                    invalidfilter : function() {
+                    invalidfacetedfilter : function() {
                         this.carryfilter = false;
                         this.getContainer().setActiveTab(0);
                         this.getContainer().getActiveTab().doLayout();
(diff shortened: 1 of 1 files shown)


api/webapp/clientapi/ext3/FilterDialog.js
```diff
@@ -1394,6 +1409,10 @@ LABKEY.FilterDialog.View.Faceted = Ext.extend(LABKEY.FilterDialog.ViewPanel, {
                 else
                     grid.selectFilter(this.filters[0]);
 
+                // Issue 52547: LKS filter dialog treats many filter types as if they are Equals
+                if (numFilters > 1 || !this.shouldShowFaceted(this.filters[0]))
+                    this.fireEvent('invalidfacetedfilter');
```
labkey-nicka, line 1414: This is the only case that I see where we fire this event and it is not the last action in the function. Should this be returning after this (sorry, not recalling my Ext lifecycle very well at the moment)?

labkey-nicka (approved):

api/webapp/clientapi/ext3/FilterDialog.js
XingY: The event is also fired in selectFilter call above prior to this change. I think it's fine to not return.

PR #6740, opened 2025-06-06
Kill child processes when cancelling a pipeline job

labkey-jeckels:
#### Rationale
Sometimes running R via the command line ends up spawning multiple processes. This may be OS-specific - Chris saw it locally on Windows. Instead of killing just the parent process, we can also kill the descendants.

#### Changes
- Iterate through child processes and shut them down gracefully and then forcibly
- Add some logging


labkey-chrisj (approved):
I like the behavior of this change on my environment, but I'm a bit worried about the build failures in this FB and it's probably best to see this run successfully on at least the BVTs and DRTS. I'll approve this now, figuring you'll make sure it's all good before merging

labkey-jeckels:
> I like the behavior of this change on my environment, but I'm a bit worried about the build failures in this FB and it's probably best to see this run successfully on at least the BVTs and DRTS. I'll approve this now, figuring you'll make sure it's all good before merging

Yes. The branch was stale and needed the latest changes from develop.

PR #6741, opened 2025-06-06
Always set Results

labkey-matthewb:
#### Rationale
MissingValueDisplayColumn requires FieldKey lookup to work so we have to set Results.  Internally it is hanging on to a ColumnInfo with the right FieldKey, but wrong Alias.

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/data/DataRegion.java
+++ b/api/src/org/labkey/api/data/DataRegion.java
@@ -29,7 +29,6 @@ import org.json.JSONObject;
 import org.junit.Assert;
 import org.junit.Test;
 import org.labkey.api.collections.BoundMap;
-import org.labkey.api.collections.CaseInsensitiveHashMap;
 import org.labkey.api.collections.ResultSetRowMapFactory;
 import org.labkey.api.collections.RowMap;
 import org.labkey.api.collections.Sets;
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
Is there a repro for this? I assume certain column names?

labkey-matthewb:
> Is there a repro for this? I assume certain column names?

Actually, no. BiologicsDataClassTest.testAliasMultiValueDisplayColumn

The bug is simply render and update form with a DataClass/Alias field.  This is due to a slight inconsistency in the implementation of MultiValuedDisplayColumn.  However, there are very few places that these show up in a form.

labkey-jeckels:
I can see that this fixes the test, and I now understand the difference. However, in testing this, I noticed that the Organism select list at the bottom of the update doesn't repopulate correctly. It's because it's comparing a String vs an Integer here:

https://github.com/LabKey/platform/blob/ff689d58499f026066dc45e0a4f88b540108a6e2/api/src/org/labkey/api/data/DataColumn.java#L601

I'm guessing it's been broken since Kevin added this code and the comment about stringifying back in 2013. Do we care about that today?

labkey-jeckels (approved):

labkey-klum (approved):
As noted earlier, reshow on MVFK display columns does not work, and probably never worked. The data region and table view form have different column infos because the display column toggles the bound column between the alias field key and the lookup field key.

PR #6748, opened 2025-06-11
Restore override of handleConversionException in CoerceDataIterator (Issue 52098)

labkey-susanh:
#### Rationale
When addressing Issue 52098, the override of `addConversionException` in `CoerceDataIterator` was removed since we were opting to fail earlier for lookup resolution and it seemed appropriate to fail earlier for other conversions as well. Turns out we want to not fail earlier except for lookup resolution where it would be more burdensome for trigger scripts to understand how to resolve in the face of number names.

#### Related Pull Requests
- #6642
- https://github.com/LabKey/testAutomation/pull/2487

#### Changes
- Restore override and add a comment
- Rename method for better clarity

### DIFF
--- a/api/src/org/labkey/api/dataiterator/CoerceDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/CoerceDataIterator.java
@@ -17,6 +17,7 @@ package org.labkey.api.dataiterator;
 
 import org.labkey.api.collections.CaseInsensitiveHashSet;
 import org.labkey.api.data.ColumnInfo;
+import org.labkey.api.data.JdbcType;
 import org.labkey.api.data.MultiValuedForeignKey;
 import org.labkey.api.data.TableInfo;
 import org.labkey.api.exp.PropertyType;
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/dataiterator/CoerceDataIterator.java
```diff
@@ -87,4 +88,11 @@ else if (to.getFk() instanceof MultiValuedForeignKey)
             }
         }
     }
+
+    // Ignore conversion exceptions during coercion and just pass back the original value.
```
labkey-nicka, line 92: Consider referring to Issue 52098 in this comment.

labkey-jeckels (approved):
I like Nick's suggestion. Otherwise, looks good and I validated the fix.

api/src/org/labkey/api/dataiterator/CoerceDataIterator.java
labkey-susanh: Sure, I can do this, but it's not actually related to the fix for Issue 52098, only mistakenly removed in the course of fixing 52098. If I hadn't made that mistake, I wouldn't have annotated this method, so I'm somewhat inclined not to.

labkey-nicka: Makes sense. Just a contextual hook is all. I'm fine either way.

PR #6750, opened 2025-06-12
Ensure ResultSet get closed

labkey-tchad:
#### Rationale
Tests are hitting "CachedResultSet was not closed." errors during project deletion. It is happening inconsistently, and only on suites with the crawler enabled. Presumably, the crawler is hitting invalid update forms and the `renderUpdateForm` result set isn't getting closed because it finds no data.

#### Related Pull Requests
- #6741 

#### Changes
- Ensure ResultSet get closed


labkey-jeckels (approved):
Seems OK to me. @labkey-matthewb any concerns? We could also change the finally block to close a local variable instead of relying on ctx.getResults()

PR #6752, opened 2025-06-13
Avoid assertion during AssayTransformImportUpdateTest

labkey-jeckels:
#### Rationale
This test is hitting an assert due to a race condition when one thread kills DB connections on another's behalf.

Failure:
https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyCPostgres/3545407

Test of FB:
https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyCPostgres?branch=raceConditionKillingConnection&buildTypeTab=overview

#### Changes
- Always log when the connection has changed, but don't assert when another thread is doing the connection closing
- Show cancel button on assay's job grid
- Avoid race condition by saving brand new "simple" metrics asychronously


labkey-chrisj (commented):
I'm not sure I have much to add here, but it's gratifying to see this test running into issues we didn't expect it to

labkey-nicka (approved):

PR #6757, opened 2025-06-13
Issue 52513: Make upgrade-based search index deletion more explicit

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52513


labkey-jeckels (approved):
I did a quick manual test too. Verified we don't reset during bootstrap or search module bumps. Created another script, copying from 24.000-24.001, to verify that an explicit reset is honored.

PR #6758, opened 2025-06-14
Improve deferred upgrade method robustness

labkey-adam:
#### Rationale
Deferred upgrade code may never run if upgrading servers are shut down prematurely. With this PR, the server queues all deferred upgrade methods in a database table instead of an in-memory list that disappears on restart. This should increase the reliability of upgrades. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43658


api/src/org/labkey/api/data/SqlScriptExecutor.java
```diff
@@ -36,44 +35,39 @@
...
      * @param schema Current schema. Null is allowed for testing purposes.
-     * @param upgradeCode Implementation of UpgradeCode that provides methods for executeJavaCode to run
      * @param moduleContext Current ModuleContext
      * @param conn Connection to use, if non-null
-     * @param literalTrue String value of boolean true for the sql dialect
      */
-    public SqlScriptExecutor(String sql, @Nullable Pattern splitPattern, @NotNull Pattern procPattern, @Nullable DbSchema schema, @Nullable UpgradeCode upgradeCode, ModuleContext moduleContext, @Nullable Connection conn, @NotNull String literalTrue)
+    public SqlScriptExecutor(String script, String sql, @Nullable Pattern splitPattern, @NotNull Pattern procPattern, @Nullable DbSchema schema, ModuleContext moduleContext, @Nullable Connection conn)
     {
+        _script = script;
```
labkey-jeckels: Consider `_scriptFileName` or similar here and in other places. I guessed it was the file name, but it's a little ambiguous.

api/src/org/labkey/api/module/DefaultModule.java
```diff
@@ -1302,28 +1309,136 @@ public boolean isAutoUninstall()
...
+     * Run all deferred upgrade methods in this module and mark them as executed
+     */
+    @Override
+    public void runDeferredUpgradeMethods(ModuleContext moduleContext)
+    {
+        SimpleFilter filter = new SimpleFilter();
+        filter.addCondition(FieldKey.fromParts("ModuleName"), moduleContext.getName());
+        filter.addCondition(FieldKey.fromParts("Executed"), null, CompareType.ISBLANK);
+        TableInfo table = CoreSchema.getInstance().getTableInfoUpgradeSteps();
+        new TableSelector(table, filter, new Sort(FieldKey.fromParts("Created"))).stream(UpgradeMethod.class)
```
labkey-jeckels: Is `Created` sufficiently granular to ensure that things run in the right order? I could imagine two rows getting inserted with the same date/time. Safer to sort on `RowId`?

labkey-jeckels (approved):
A couple of small suggestions. I didn't have a chance to test this out.

api/src/org/labkey/api/module/DefaultModule.java
labkey-adam: `Created` is probably fine, but I agree `RowId` is better

PR #6759, opened 2025-06-16
Another try to avoid showing stale password strength

labkey-jeckels:
#### Rationale
We've still seeing some test failures where the product seems to be showing stale password strength feedback.

https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_InternalSuites_OldDependenciesSqlserver/3548355?buildTab=overview

#### Changes
- Check that the password in the input matches the one that initiated the request

### DIFF
--- a/core/webapp/passwordGauge.js
+++ b/core/webapp/passwordGauge.js
@@ -74,7 +74,7 @@ LABKEY.PasswordGauge = new function() {
                 _pendingScoreRequest.abort();
                 _pendingScoreRequest = null;
             }
-            let passwordAtTimeOfRequest = password.value;
+            const passwordAtTimeOfRequest = password.value;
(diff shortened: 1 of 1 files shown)


core/webapp/passwordGauge.js
```diff
@@ -74,14 +74,18 @@ LABKEY.PasswordGauge = new function() {
                 _pendingScoreRequest.abort();
                 _pendingScoreRequest = null;
             }
+            let passwordAtTimeOfRequest = password.value;
```
labkey-nicka: nit: `const`

labkey-nicka (approved):

labkey-danield (approved):

core/webapp/passwordGauge.js
labkey-jeckels: Updated

PR #6742, opened 2025-06-08
Minor auto-refactor code cleanup on a massive scale

labkey-jeckels:
#### Rationale
We can cleanup tens of thousands of warnings across our codebase with IntelliJ's autorefactors. In some cases, this adopts newer, leaner syntax. In others, it simply removes code that's not serving any purpose.

#### Changes
- Auto-refactors
  - Eliminate redundant toString()
  - Eliminate redundant cast
  - Empty JavaDoc annotations
  - Member variable can be final
  - length() and size() -> isEmpty()
  - Add missing @Override
  - Class can be static
  - Remove unused imports
  - toArray() passed array length
  - Remove unnecessary semicolon
  - Explicit type syntax can be replaced by <>
  - Redundant access modifiers
  - Remove redundant initializer
  - "".equals() -> isEmpty()
  - StringBuilder can be replaced by String
  - Fix misordered arguments to assertEquals()
  - StringUtils.defaultString() - Objects.toString()
  - Cast can be replaced with pattern variable
  - Collapse identical catch blocks
  - Type may be primitive
- Manual fixups
  - Delete unused GWT code
  - Improve generics
  - new UnexpectedException() -> UnexpectedException.wrap()

PR #6763, opened 2025-06-17
Luminex test failure fix - transform script null run properties are being written to runProperties.tsv as "null" instead of ""

cnathe:
#### Rationale
LuminexPositivityTest and LuminexAsyncImportTest started failing recently in develop. Turns out the runProperties.tsv file for the transform script was writing null values as "null" instead of "". This was related to the `StringUtils.defaultString() - Objects.toString()` change from this PR https://github.com/LabKey/platform/pull/6742

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6742

#### Changes
- set nullDefault param as "" for Object.toString()

### DIFF
--- a/api/src/org/labkey/api/qc/TsvDataExchangeHandler.java
+++ b/api/src/org/labkey/api/qc/TsvDataExchangeHandler.java
@@ -454,7 +454,7 @@ public class TsvDataExchangeHandler implements DataExchangeHandler
         {
             pw.append(writer.quoteValue(entry.getKey().getName()));
             pw.append('\t');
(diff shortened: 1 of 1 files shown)


cnathe:
@josheckels do you think any of the other cases that changed from `StringUtils.defaultString()` to `Objects.toString()` need to add this nullDefault parameter as `""` to match previous behavior?

labkey-jeckels (approved):
Thanks Cory. I thought I had tracked down all of these single-argument calls to toString() in the refactor to avoid this problem, but it looks like this snuck through. I double-checked for other examples in my big refactor commit and couldn't find any.

labkey-klum (approved):

PR #6772, opened 2025-06-19
Ensure a valid file link path for file field types

labkey-klum:
#### Rationale
[53320: NPE updating a dataset file field](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53320)

This seems related to the recent changes to `DatasetUpdateService.updateRow`.

- Need to ensure that there is a valid path provided for file field types, this is already handled in the `DataIterator` for inserts.
- Added a test to validate this path. We were so close to probably catching it when we wired up support for files in the reshow case.

### DIFF
--- a/study/src/org/labkey/study/query/DatasetUpdateService.java
+++ b/study/src/org/labkey/study/query/DatasetUpdateService.java
@@ -23,6 +23,7 @@ import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.labkey.api.assay.AssayFileWriter;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
 import org.labkey.api.collections.CaseInsensitiveHashSet;
 import org.labkey.api.collections.ResultSetRowMapFactory;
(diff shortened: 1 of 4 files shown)


study/src/org/labkey/study/query/DatasetUpdateService.java
```diff
@@ -563,6 +565,16 @@ private void resyncStudy(User user, Container container, @Nullable Set<String> p
             throw errors;
     }
 
+    @Override
+    protected void convertTypes(User user, Container c, Map<String, Object> row, TableInfo t, @Nullable Path fileLinkDirPath) throws ValidationException
+    {
+        // Issue 53320 : ensure a valid file link path
+        if (fileLinkDirPath == null)
+            fileLinkDirPath = AssayFileWriter.getUploadDirectoryPath(c, "datasetdata").toNioPathForWrite();
```
labkey-matthewb, line 573: It looks like getUploadDirectoryPath() should be annotated @NotNull?  (though I don't think we handle a misconfigured server in this case anyway)

labkey-matthewb (approved):

PR #6775, opened 2025-06-19
Make delete protocol less prone to deadlock

labkey-matthewb:
#### Rationale
I've seen deadlocks where runs are being deleted at the same time as the protocol/assay is being deleted.  One good approach is to grab locks as soon as possible in the transaction.  I think this will help.

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -4532,6 +4532,11 @@ public class ExperimentServiceImpl implements ExperimentService, ObjectReference
 
         try (DbScope.Transaction transaction = ensureTransaction())
         {
+            // To increase the chances of getting through this w/o a deadlock, lets lock tables now that we plan on deleting
+            // There are other tables/rows we could lock, but this is a start
+            // CONSIDER similar behavior for Domain.delete()
+            _lockProvisionedTables(assayService, expProtocols);
+
             for (ExperimentListener listener : _listeners)
             {
                 listener.beforeProtocolsDeleted(c, user, expProtocols);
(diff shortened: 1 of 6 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -5431,10 +5463,7 @@ private void deleteExpExperiment(Container c, User user, @NotNull ExpExperimentI
             // If we're a batch, delete all the runs too
             if (experiment.getDataObject().getBatchProtocolId() != null)
             {
-                for (ExpRunImpl expRun : experiment.getRuns())
-                {
-                    expRun.delete(user);
-                }
+                ExperimentServiceImpl.get().deleteExperimentRuns(c, user, null, experiment.getRuns());
```
labkey-matthewb: Need to check that we've already checked permissions in order to use deleteExperimentRuns()

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -4532,6 +4532,11 @@ public void deleteProtocolByRowIds(Container c, User user, @Nullable String audi
 
         try (DbScope.Transaction transaction = ensureTransaction())
         {
+            // To increase the chances of getting through this w/o a deadlock, lets lock tables now that we plan on deleting
+            // There are other tables/rows we could lock, but this is a start
+            // CONSIDER similar behavior for Domain.delete()
```
labkey-jeckels, line 4537: `Domain` already has `getDatabaseLock()`. Maybe this could be pushed into the `Domain` interface with comments about how it differs from the existing method?

labkey-jeckels (approved):
Code change looks OK. See suggestion for pushing into `Domain` interface.

What testing do you recommend on this?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-matthewb: The Domain.delete() happens last in this code flow.  We could also grab those locks here at the beginning of the delete protocol transaction.

labkey-jeckels: Yeah, not suggesting we somehow fold this into `Domain.delete()`. Something more like `Domain.getDatabaseLockForDelete()` or `Domain.getDatabaseLockForDrop()` which could handle the SQL (or no-op)

experiment/src/org/labkey/experiment/api/property/DomainImpl.java
```diff
@@ -408,6 +409,22 @@ public Lock getDatabaseLock()
         return getLock(_dd);
     }
 
+    @Override
+    public void lockForDelete(DbSchema expSchema)
+    {
+        // NOTE code relies on the lock returned from Domain.getLock() does not require unlock().
+        var lock = getDatabaseLock();
+        assert lock instanceof DbScope.ServerLock;
```
labkey-jeckels, line 417: Maybe we should also assert that we're inside of a transaction?
```suggestion
        assert lock instanceof DbScope.ServerLock;
        assert ExperimentService.get().getSchema().getScope().isTransactionActive();
```

labkey-jeckels (approved):
One idea for a new assert

labkey-jeckels (approved):

PR #6778, opened 2025-06-21
Option to expose read-only specimen data without the module

labkey-adam:
#### Rationale
Clients who are migrating from Specimen to Sample Manager would like a read-only view of their specimen data. This provides an optional feature that displays the specimen tables in the study schema in the absence of the specimen module.

Most of the changes simply consolidate the many redundant specimen table name constants to the ones defined in `SpecimenTablesProvider`.

### DIFF
--- a/specimen/src/org/labkey/specimen/SpecimenManager.java
+++ b/specimen/src/org/labkey/specimen/SpecimenManager.java
@@ -28,6 +28,7 @@ import org.labkey.api.specimen.Vial;
 import org.labkey.api.specimen.location.LocationImpl;
 import org.labkey.api.specimen.location.LocationManager;
 import org.labkey.api.specimen.model.SpecimenComment;
+import org.labkey.api.specimen.model.SpecimenTablesProvider;
 import org.labkey.specimen.model.SpecimenTypeSummary;
 import org.labkey.specimen.model.SpecimenTypeSummaryRow;
 import org.labkey.api.study.Cohort;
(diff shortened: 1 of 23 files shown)


labkey-adam:
@labkey-klum @labkey-danield Some manual testing scenarios worth exercising:
- Remove the Specimen module. Shut down the server, in <LABKEY_ROOT>/build/deploy/modules, delete the specimen.module file and the specimen folder, and restart the server.
- Enable the new optional feature. Verify that specimen tables appear in the study schema and can be browsed.
- Turn on the "subschemas" experimental feature and verify specimen tables still appear & can be browsed.
- Extra credit: create a lookup from a dataset field to SpecimenDetails (or similar). Verify the dataset displays correctly. Turn off the optional feature; it should still work. (This is essentially https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53336)
- Specimen tables should appear if requested on the URL, even if the optional feature is turned off.

labkey-klum (approved):
Looks good, interesting to note that we don't block creation of any of the specimen tables if the module or flag isn't present. Maybe we don't care but it's slightly different behavior from the `studydesign` tables.

labkey-adam:
> Looks good, interesting to note that we don't block creation of any of the specimen tables if the module or flag isn't present. Maybe we don't care but it's slightly different behavior from the `studydesign` tables.

Way back, I had looked at moving or conditionalizing the specimen table handling, but there's a lot of study code and SQL that assumes they're present, so I bailed on that. I think we'll need to tolerate them until the last client stops using specimen.

labkey-danield:
I'm done with my manual testing. It took a little while because I had to "relearn" specimens, studies and the various table relationships.

The issue is marked as needing automation. I'm not sure there is much automation we could write that would require the removal the specimen module, I would need to brainstorm with @labkey-tchad on that.
We could write automation that validates the experimental feature, and then looks at the schema browser.

labkey-adam:
> I'm done with my manual testing. It took a little while because I had to "relearn" specimens, studies and the various table relationships.
> 
> The issue is marked as needing automation. I'm not sure there is much automation we could write that would require the removal the specimen module, I would need to brainstorm with @labkey-tchad on that. We could write automation that validates the experimental feature, and then looks at the schema browser.

Maybe you saw, but I put my thoughts on automation into the issue. The optional feature shows up only if specimen module is not present, so relevant testing would require removing/disabling the specimen module. Note that we can disable a module that's in a distribution via startup property. But if we decide it's too much bother that's probably okay.

PR #6781, opened 2025-06-23
Issue 52777: use field key if it already exists

labkey-susanh:
#### Rationale
Issue [52777](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52777) - Previous PRs related to this issue addressed the export column name but we are still left with logic that can map a field that should be unrecognizable to a field name with dollar signs.  See the issue for more details.

#### Related Pull Requests
- #6663 
- https://github.com/LabKey/limsModules/pull/1483

#### Changes
- Update `_createTableMap` to use a column's fieldKey to discern the displayString

### DIFF
--- a/api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
+++ b/api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
@@ -217,8 +217,8 @@ public class DataIteratorUtil
         {
             // Issue 21015: Dataset snapshot over flow assay dataset doesn't pick up stat column values
             // TSVColumnWriter.ColumnHeaderType.queryColumnName format is a FieldKey display value from the column name. Blech.
-            // Issue 52777: If the column already has a fieldKey, use that one instead otherwise we convert $$Date to $$ate using fromString() and toDisplayString()
-            String tsvQueryColumnName = col.getFieldKey() != null ? col.getFieldKey().toDisplayString() : FieldKey.fromString(col.getName()).toDisplayString();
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/dataiterator/DataIteratorUtil.java
```diff
@@ -217,7 +217,8 @@ protected static Map<String,Pair<ColumnInfo,MatchType>> _createTableMap(TableInf
         {
             // Issue 21015: Dataset snapshot over flow assay dataset doesn't pick up stat column values
             // TSVColumnWriter.ColumnHeaderType.queryColumnName format is a FieldKey display value from the column name. Blech.
-            String tsvQueryColumnName = FieldKey.fromString(col.getName()).toDisplayString();
+            // Issue 52777: If the column already has a fieldKey, use that one instead otherwise we convert $$Date to $$ate using fromString() and toDisplayString()
+            String tsvQueryColumnName = col.getFieldKey() != null ? col.getFieldKey().toDisplayString() : FieldKey.fromString(col.getName()).toDisplayString();
```
labkey-matthewb: I think we can assert null != col.getFieldKey()
Not sure what the original code is trying to accomplish

labkey-jeckels: Could or should we do that in `BaseColumnInfo`'s constructor? It would be nice to make `getName()` and `getFieldKey()` `NotNull`

labkey-matthewb: We always set them both together.  So we should never have one set and the other not.

labkey-matthewb (approved):

PR #6782, opened 2025-06-23
Don't cache objects that hold Containers

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53058

#### Changes
- Stop holding onto `Container` objects in:
  - `AbstractStudyEntity`
  - `AbstractWebdavResolver`
  - `DatasetDefinition`
  - `DataState`
  - `LocationImpl`
  - `ObjectProperty`
  - `Participant`
  - `PlateImpl`
  - `PositionImpl`
  - `PropertySetImpl`
- Prefer holding and passing container IDs as `GUID` not `String`

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateSetImpl.java
+++ b/assay/src/org/labkey/assay/plate/PlateSetImpl.java
@@ -8,6 +8,7 @@ import org.labkey.api.assay.plate.Plate;
 import org.labkey.api.assay.plate.PlateSet;
 import org.labkey.api.assay.plate.PlateSetType;
 import org.labkey.api.data.Container;
+import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.Entity;
 import org.labkey.api.data.SQLFragment;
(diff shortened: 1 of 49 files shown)


labkey-matthewb (approved):

assay/src/org/labkey/assay/plate/PlateSetImpl.java
```diff
@@ -73,23 +73,6 @@ public void setFolder(Container container)
         setContainer(container);
     }
 
-    @Override
```
labkey-nicka, line 76: Just a heads up that these are modeled here and serialized to the client which definitely utilizes them. You'll need to run the Biologics test suite to get an idea for the coverage here.

labkey-adam: Thanks... getContainerId() and getContainerPath() are now identical to the parent (Entity) impl. I'll add back getContainerName() (and annotate it) if you think that's used.

PR #6788, opened 2025-06-24
ExpObjectDataIterator: consolidate iterator logic

labkey-nicka:
#### Rationale
This is the remanent refactor of prior work done to address [Issue 53153](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53153).

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1815
- https://github.com/LabKey/platform/pull/6788
- https://github.com/LabKey/limsModules/pull/1487

#### Changes
- Introduce `ExpDataTypeDataIterator` to encapsulate logic for data iterators that operate on experiment materials and data.
- Fix `FlagDataIterator` logic to not rely on presence of `lsid` column

### DIFF
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -513,7 +513,7 @@ public class ExpDataIterators
         }
     }
 
-    private static class AliasDataIterator extends ExpObjectDataIterator
+    private static class AliasDataIterator extends ExpDataTypeDataIterator
     {
         // For some reason I don't quite understand we don't want to pass through a column called "alias" so we rename it to ALIASCOLUMNALIAS
         final Supplier<Object> _aliasCol;
@@ -740,7 +740,7 @@ public class ExpDataIterators
         }
     }
 
-    private static class FlagDataIterator extends ExpObjectDataIterator
+    private static class FlagDataIterator extends ExpDataTypeDataIterator
     {
         final DataIteratorContext _context;
         final Integer _lsidCol;
@@ -894,7 +894,7 @@ public class ExpDataIterators
         return values == null ? null : values.collect(Collectors.toList());
     }
 
-    static class DerivationDataIteratorBase extends ExpObjectDataIterator
+    static class DerivationDataIteratorBase extends ExpDataTypeDataIterator
     {
         final Integer _lsidCol;
         final Integer _nameCol;
--- a/experiment/src/org/labkey/experiment/ExpDataIterators.java
+++ b/experiment/src/org/labkey/experiment/ExpDataIterators.java
@@ -3186,16 +3186,14 @@ public class ExpDataIterators
         public static final String ALIQUOT_FROM_RESOLVED_ROW_ID = "_" + AliquotResolutionDataIterator.class.getName() + "#RowId";
 
         final Integer _aliquotFromCol;
+        final Integer _aliquotNameCol;
+        final Integer _aliquotRowIdCol;
         final Container _container;
         final Integer _isAliquotCol;
         final Map<Integer, ExpMaterial> _materialCache;
         final RemapCache _remapCache;
         final ExpSampleTypeImpl _sampleType;
         final User _user;
-
-        String aliquotName = null;
-        Integer aliquotRowId = null;
-        boolean isAliquot = false;
 
         public AliquotResolutionDataIterator(DataIterator di, DataIteratorContext context, Container container, User user, ExpSampleTypeImpl sampleType)
         {
@@ -3212,12 +3210,17 @@ public class ExpDataIterators
 
             _materialCache = hasAliquotFromCol ? new HashMap<>() : null;
             _remapCache = hasAliquotFromCol ? new RemapCache() : null;
-            _isAliquotCol = addColumn(new BaseColumnInfo(ALIQUOT_FROM_IS_ALIQUOT, JdbcType.BOOLEAN), (Supplier<Boolean>)() -> isAliquot);
+            _isAliquotCol = addColumn(new BaseColumnInfo(ALIQUOT_FROM_IS_ALIQUOT, JdbcType.BOOLEAN), (Supplier<Boolean>)() -> hasAliquotFromCol);
 
             if (hasAliquotFromCol)
             {
-                addColumn(new BaseColumnInfo(ALIQUOT_FROM_RESOLVED_NAME, JdbcType.VARCHAR), (Supplier<String>)() -> aliquotName);
-                addColumn(new BaseColumnInfo(ALIQUOT_FROM_RESOLVED_ROW_ID, JdbcType.INTEGER), (Supplier<Integer>)() -> aliquotRowId);
+                _aliquotNameCol = addColumn(new BaseColumnInfo(ALIQUOT_FROM_RESOLVED_NAME, JdbcType.VARCHAR), (Supplier<String>)() -> null);
+                _aliquotRowIdCol = addColumn(new BaseColumnInfo(ALIQUOT_FROM_RESOLVED_ROW_ID, JdbcType.INTEGER), (Supplier<Integer>)() -> null);
+            }
+            else
+            {
+                _aliquotNameCol = null;
+                _aliquotRowIdCol = null;
             }
         }
 
@@ -3228,10 +3231,6 @@ public class ExpDataIterators
 
             if (hasNext && _aliquotFromCol != null)
             {
-                aliquotName = null;
-                aliquotRowId = null;
-                isAliquot = false;
-
                 Object aliquotedFromObj = get(_aliquotFromCol);
                 if (aliquotedFromObj != null)
                 {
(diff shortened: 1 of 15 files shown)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -3283,4 +3179,109 @@ public boolean next() throws BatchValidationException
...
+        {
+            super(di, context);
+            selectAll();
+
+            _container = container;
+            _sampleType = sampleType;
+            _user = user;
+
+            var columnNameMap = getColumnNameMap();
+            _aliquotFromCol = columnNameMap.get(ALIQUOTED_FROM_INPUT);
```
labkey-matthewb: equivalent, just pointing out DataIterator.getSupplier() is also useful

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -3283,4 +3179,109 @@ public boolean next() throws BatchValidationException
...
+                    }
+
+                    boolean isAliquot = !StringUtils.isEmpty(aliquotedFrom);
+                    _row[_isAliquotCol] = isAliquot;
+
+                    if (isAliquot)
+                    {
+                        try
+                        {
+                            ExpMaterial aliquotParent = ExperimentService.get().findExpMaterial(_container, _user, aliquotedFrom, _sampleType, _remapCache, _materialCache);
```
labkey-matthewb: We're resolving by String (name) first even when the caller might intend ROWID?  Can we use an alternate name to be explicit.  E.g. AliquotedFromRowId?  That would allow explicit indication of what the caller inteneded.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -782,43 +774,27 @@ public boolean next() throws BatchValidationException
...
 
                 if (lsidValue instanceof String lsid)
                 {
-                    String flag = Objects.toString(flagValue, null);
-
                     try
                     {
-                        if (_isSample)
+                        ExpObject expObject = null;
+                        if (_context.getInsertOption().mergeRows && _nameCol != null)
```
labkey-susanh: I know this isn't your doing, but it's not clear why this logic about the name column is inside the `if` for the `lsid` being a string. The `lsid` comes into play only if the object isn't found by name.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -2021,25 +1989,18 @@ public String nextName(Map<String, Object> map,
...
-                else if (aliquotedFromObj instanceof Number)
-                {
-                    aliquotedFrom = aliquotedFromObj.toString();
-                }
+                // Issue 53153: When a name has been resolved for the aliquot,
+                // then use that instead of the supplied value for AliquotedFrom.
+                Object name = map.get(ALIQUOT_FROM_RESOLVED_NAME);
+                if (name instanceof String aliquotParentName)
+                    map.put(ExpMaterial.ALIQUOTED_FROM_INPUT, aliquotParentName);
+                else if (name instanceof List<?> nameList)
```
labkey-susanh: When would this happen? Presumably there is only one value in this list? 

Ah yes, I see now in the `AliquotResolutionDataIterator`. Perhaps add a comment here that references that method?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -3283,4 +3179,109 @@ public boolean next() throws BatchValidationException
         }
     }
 
+    public static class AliquotResolutionDataIterator extends SimpleTranslator
+    {
+        public static final String ALIQUOT_FROM_IS_ALIQUOT = "_" + AliquotResolutionDataIterator.class.getName() + "#IsAliquot";
```
labkey-susanh: I think this name is confusing me. This is not indicating that the AliquotedFrom value (the parent) is itself an aliquot, only if this row is representing an aliquot, right?  I would probably just call this `IS_ALIQUOT`.

labkey-susanh (approved):

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-nicka: No longer applicable.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-nicka: No longer applicable.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
labkey-nicka: No longer applicable.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-nicka: No longer applicable.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-nicka: Yikes. I agree, this logic was pretty snarled. I've refactored this to not be nested in the `lsid` check. See 7cb5ff1df4.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -548,37 +544,56 @@ public boolean next() throws BatchValidationException
...
+                            ExpObject obj = getExpObjectByName(name);
+                            if (obj != null)
+                            {
+                                String lsid = obj.getLSID();
+                                if (lsid != null && !lsid.isEmpty())
+                                    _lsidAliasMap.put(lsid, _aliasCol.get());
+                            }
+                        }
+                    }
+                    else if (_lsidCol != null)
```
labkey-susanh: I know it's kind of silly, but you could have both the name and the LSID columns, so perhaps this logic should be more like in the FlagDataIterator where we go into the `lsid` block of `obj` is null?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -548,37 +544,56 @@ public boolean next() throws BatchValidationException
...
+            if (hasNext)
             {
-                final ExperimentService svc = ExperimentService.get();
-
-                try (DbScope.Transaction transaction = svc.getTinfoDataClass().getSchema().getScope().ensureTransaction())
+                // For each iteration, collect the lsid and alias col values.
+                if (_aliasCol != null)
                 {
-                    for (Map.Entry<String, Object> entry : _lsidAliasMap.entrySet())
+                    if (_context.getInsertOption().mergeRows && _nameCol != null)
```
labkey-susanh: I'm not sure the check for `mergeRows` here is correct. Couldn't you also have a `_nameCol` for the update case?

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -779,55 +771,38 @@ public boolean next() throws BatchValidationException
...
 
-            if (_lsidCol != null && _flagCol != null)
-            {
-                Object lsidValue = get(_lsidCol);
+            if (_flagCol == null)
+                return true;
 
-                Object flagValue = get(_flagCol);
+            ExpObject expObject = null;
+            if (_nameCol != null && _context.getInsertOption().mergeRows)
```
labkey-susanh: Same question here about merge vs. update.

labkey-susanh (changes requested):

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-nicka: Updated to be more similar to FlagDataIterator. That said, retaining prior behavior of passing LSID into the map without firing off an additional query for the object.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-nicka: Update to check merge or update.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-nicka: Update to check merge or update.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -1615,76 +1608,69 @@ private static boolean isAliquotRollupHeader(String name)
...
-        Set<String> _existingNames = null;
-        List<Supplier<Map<String, Object>>> _extraPropsFns = new ArrayList<>();
+        final ExpSampleTypeImpl _sampleType;
+        final User _user;
 
+        Set<String> _existingNames = null;
         String generatedName = null;
 
-        _GenerateNamesDataIterator(ExpSampleTypeImpl sampleType, Container dataContainer, User user, MapDataIterator source, DataIteratorContext context, int batchSize)
+        _GenerateNamesDataIterator(ExpSampleTypeImpl sampleType, Container container, User user, MapDataIterator source, DataIteratorContext context, int batchSize)
```
labkey-susanh: Should there be a `@NotNull` annotation on `sampleType` here?

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
labkey-nicka: Done.

PR #6789, opened 2025-06-24
Postgres specific implementation of ROUND()

labkey-matthewb:
#### Rationale
Postgres does not support ROUND(double) use explicit cast to NUMERIC(unspecified) as a work around.

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/query/src/org/labkey/query/sql/Method.java
+++ b/query/src/org/labkey/query/sql/Method.java
@@ -857,6 +857,8 @@ public abstract class Method
             }
             else
             {
+                // NOTE: At the moment there appear to be no Dialects that use this code path (but it still works).
+                // Leave for completeness?
                 int n = Integer.MIN_VALUE;
                 try
                 {
--- a/query/src/org/labkey/query/sql/Method.java
+++ b/query/src/org/labkey/query/sql/Method.java
@@ -360,7 +360,7 @@ public abstract class Method
         labkeyMethod.put("repeat", new JdbcMethod("repeat", JdbcType.VARCHAR, 2, 2));
         labkeyMethod.put("round", new Method("round", JdbcType.DOUBLE, 1, 2)
 			{
-				@Override
+                @Override
 				public MethodInfo getMethodInfo()
 				{
 					return new RoundInfo();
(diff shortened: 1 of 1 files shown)


query/src/org/labkey/query/sql/Method.java
```diff
@@ -842,28 +842,39 @@ public SQLFragment getSQL(SqlDialect dialect, SQLFragment[] arguments)
...
+            if (supportsRoundDouble)
             {
-                i = Integer.parseInt(arguments[1].getSQL());
+                return super.getSQL(dialect, arguments);
             }
-            catch (NumberFormatException x)
+            else if (dialect.isPostgreSQL())
             {
-                /* fall through */
+                // Postgres ROUND() requires NUMERIC argument and will error with DOUBLE
```
labkey-adam, line 851: Move/update/consolidate the comment above (lines #829 - 831) into this comment?

query/src/org/labkey/query/sql/Method.java
```diff
@@ -842,28 +842,39 @@ public SQLFragment getSQL(SqlDialect dialect, SQLFragment[] arguments)
...
+                // This CAST works because in postgres NUMERIC(unspecified) => up to 131072 digits before the decimal point; up to 16383 digits after the decimal point
+                // This is not SQL standard behavior
+                SQLFragment numeric = new SQLFragment();
+                numeric.append("CAST((").append(arguments[0]).append(") AS NUMERIC)");
+                return super.getSQL(dialect, new SQLFragment[] {numeric, arguments[1]});
+            }
+            else
+            {
+                int n = Integer.MIN_VALUE;
```
labkey-adam, line 862: I believe PostgreSQL is the only dialect that doesn't support round double, so we should never get to this case. We could remove this block or at least add a comment. We could also default `supportsRoundDouble()` to return true to make it more obvious that PostgreSQL is the only exception.

query/src/org/labkey/query/sql/Method.java
```diff
@@ -360,7 +360,7 @@ public MethodInfo getMethodInfo()
         labkeyMethod.put("repeat", new JdbcMethod("repeat", JdbcType.VARCHAR, 2, 2));
         labkeyMethod.put("round", new Method("round", JdbcType.DOUBLE, 1, 2)
 			{
-				@Override
+                @Override
```
labkey-adam, line 363: Weird spacing

labkey-adam (approved):
Some suggestions in my comments

query/src/org/labkey/query/sql/Method.java
labkey-matthewb: Lots of TAB chars in this file.  Can address in develop.

query/src/org/labkey/query/sql/Method.java
labkey-matthewb: Added comment.

PR #6793, opened 2025-06-25
Prevent unsupported MVFK XML overrides

labkey-klum:
#### Rationale
related issue : https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51695

A user could previously get into a bad state by editing the table metadata and using a MVFK type that wasn't supported. If that occurred, a user wouldn't be able to fix the query because we were throwing during table FK construction. This PR adds validation to the save action to prevent construction of an invalid FK. This also downgrades the previous exception to a warning to allow corrections to existing metadata.

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/2514


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -1401,6 +1402,19 @@ public void validateForm(SourceForm form, Errors errors)
...
 
+        private void validateForeignKey(ColumnType.Fk fk, Errors errors)
+        {
+            if (fk.isSetFkMultiValued())
+            {
+                // issue 51695 : don't let users create unsupported MVFK types
+                String type = fk.getFkMultiValued();
+                if (!"junction".equals(type))
+                {
+                    errors.reject(ERROR_MSG, String.format("Non-junction multi value column type : \"%s\" is not supported.", type));
```
labkey-nicka: Could this provide more context about which column is failing the validation? I'm not sure what context the user is given when this appears.

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -1401,6 +1402,19 @@ public void validateForm(SourceForm form, Errors errors)
             }
         }
 
+        private void validateForeignKey(ColumnType.Fk fk, Errors errors)
+        {
+            if (fk.isSetFkMultiValued())
+            {
+                // issue 51695 : don't let users create unsupported MVFK types
+                String type = fk.getFkMultiValued();
+                if (!"junction".equals(type))
```
labkey-nicka: nit: The `"junction"` and `"value"` values for this could/should be rolled into an enumeration. There are several places these constants are referenced.

labkey-nicka (approved):

query/src/org/labkey/query/controllers/QueryController.java
labkey-klum: good idea

PR #6794, opened 2025-06-25
Log error if we attempt to cache a "bad" value

labkey-adam:
#### Rationale
We don't want cached objects holding onto `Container` or `User` objects: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53058

We've cleared most (maybe all?) cases, so it's time to crank up the logging to flush out any hiding stragglers. This PR also logs errors when we detect a mutable collection or array getting cached; for a long time, we've logged these as warnings.

#### Related Pull Requests
- https://github.com/LabKey/commonAssays/pull/899


labkey-matthewb (approved):

labkey-adam:
Existing automated tests have flagged the classes I changed and these same tests validate the changes I made. I don't see a value in manual or further automated tests here or in the companion PR.

PR #6795, opened 2025-06-25
 Misc assay issues fixes for 25.7 - part 3

cnathe:
#### Rationale
#[763](https://github.com/LabKey/internal-issues/issues/649): App assay creation allows for assay names that only differ by casing
#[795](https://github.com/LabKey/internal-issues/issues/661): LKS Schema Browser "missing schema" error when assay name has special characters
Issue [52075](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52075): labkey default assay run IDs replace all special characters in the assay name with underscores
Issue [53355](https://www.labkey.org/issues/home/Developer/issues/details.view?issueId=53355): check for null schema

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1816
- https://github.com/LabKey/labkey-ui-premium/pull/781
- https://github.com/LabKey/limsModules/pull/1491
- https://github.com/LabKey/platform/pull/6795

#### Changes
- Assay create/save check name collision to be case-insensitive
- Schema browser click on sub-schema name with special characters
- Generated Assay ID for tmp file case to use original assay name instead of encoded name
- getCompileTimeContainer to check for null schema

### DIFF
--- a/api/src/org/labkey/api/assay/AssayFileWriter.java
+++ b/api/src/org/labkey/api/assay/AssayFileWriter.java
@@ -138,33 +138,43 @@ public class AssayFileWriter<ContextType extends AssayRunUploadContext<? extends
         }
     }
 
+    public static String generateFileName(ExpProtocol protocol, boolean shouldEncode)
+    {
+        Date dateCreated = new Date();
+        String dateString = DateUtil.formatDateTime(dateCreated, "yyy-MM-dd-HHmmss");
+
+        String protocolName = protocol.getName();
+        if (shouldEncode)
+        {
+            char[] characters = protocolName.toCharArray();
+
+            for (int i = 0; i < characters.length; i++)
+            {
+                char character = characters[i];
+                boolean isAtoZchar = character >= 'A' && character <= 'z';
+                if (!Character.isDigit(character) && !isAtoZchar)
+                    characters[i] = '_';
+            }
+            protocolName = new String(characters);
+        }
+
+        return protocolName + "-" + dateString;
+    }
+
     /**
      * Create file name based upon the assay protocol's name and the current time.
      * e.g., <code>assayname-2020-04-14-1602345</code>
      */
     public static FileLike createFile(ExpProtocol protocol, FileLike dir, String extension)
     {
-        Date dateCreated = new Date();
-        String dateString = DateUtil.formatDateTime(dateCreated, "yyy-MM-dd-HHmmss-SSS");
+        String fileNamePrefix = generateFileName(protocol, true);
         int id = 0;
-
-        String protocolName = protocol.getName();
-        char[] characters = protocolName.toCharArray();
-
-        for (int i = 0; i < characters.length; i++)
-        {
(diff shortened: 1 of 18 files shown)


api/src/org/labkey/api/assay/AssayFileWriter.java
```diff
@@ -138,33 +138,43 @@ public static FileLike ensureUploadDirectoryPath(FileLike dir) throws Experiment
         }
     }
 
+    public static String generateFileName(ExpProtocol protocol, boolean shouldEncode)
+    {
+        Date dateCreated = new Date();
+        String dateString = DateUtil.formatDateTime(dateCreated, "yyy-MM-dd-HHmmss");
```
labkey-klum: Should we be using the container configured dateTime format string, or do we care?

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
```diff
@@ -426,16 +426,21 @@ public GWTProtocol saveChanges(GWTProtocol assay, boolean replaceIfExisting) thr
                 boolean hasNameChange = false;
                 AssayProvider assayProvider = AssayService.get().getProvider(assay.getProviderName());
                 String oldAssayName = null;
+
+                // Issue 51321: check reserved domain name: First, All
+                if ("First".equalsIgnoreCase(assay.getName()) || "All".equalsIgnoreCase(assay.getName()))
```
labkey-klum: I've seen a number of places where we check for these strings. Would it make sense to introduce enums or constants or consolidate the checks?

labkey-klum (approved):

api/src/org/labkey/api/assay/AssayFileWriter.java
labkey-nicka: This is an interesting decision point. Feels like having these hard-coded to a single format will be more stable, however, I can understand wanting to align it with user date reading. Are there other places we've had to make a similar decision? Perhaps @labkey-jeckels has thoughts.

labkey-jeckels: We do something similar with a hardcoded format (ISO, with substitutions to make it filename friendly) in `FileUtil.makeFileNameWithTimestamp()`. Maybe we should use that here.

I haven't heard any requests to use the configured date/time formats.

cnathe: I like the idea to use `FileUtil.makeFileNameWithTimestamp()` here. It has the added benefit of removing the custom code for converting/encoding characters for the file name.  Done.

assay/src/org/labkey/assay/AssayDomainServiceImpl.java
cnathe: Good idea. Consolidated checks into DomainUtil.validateReservedName

labkey-klum (approved):

labkey-nicka (approved):

PR #6797, opened 2025-06-25
Fix not-equals check for SecurityQuery

labkey-adam:
#### Rationale
Search tests started failing with an assert in SecurityQuery after related PR was merged. Assert was hit because an inequality check started to always fail due to `r.containerId()` changing its return type to GUID.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6782

### DIFF
--- a/search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
+++ b/search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
@@ -783,7 +783,7 @@ public class LuceneSearchServiceImpl extends AbstractSearchService implements Se
             String resourceId = (String)props.get(PROPERTY.securableResourceId.toString());
             String securityContext = r.getContainerId()
                     + "|" + props.get(PROPERTY.categories.toString()) // multiple categories are separated by spaces, but we shouldn't need to distinguish here
-                    + (null != resourceId && !resourceId.equals(r.getContainerId()) ? "|" + resourceId : "");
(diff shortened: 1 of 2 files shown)


search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
```diff
@@ -783,7 +783,7 @@ else if (owner instanceof User)
             String resourceId = (String)props.get(PROPERTY.securableResourceId.toString());
             String securityContext = r.getContainerId()
                     + "|" + props.get(PROPERTY.categories.toString()) // multiple categories are separated by spaces, but we shouldn't need to distinguish here
-                    + (null != resourceId && !resourceId.equals(r.getContainerId()) ? "|" + resourceId : "");
+                    + (null != resourceId && !resourceId.equals(r.getContainerId().toString()) ? "|" + resourceId : "");
```
labkey-jeckels, line 786: `equalsIgnoreCase()`? I remember having some case sensitivity issues with GUID-as-strings in the past.

labkey-jeckels (approved):
This seems safe enough to merge without manual testing or new automated coverage.

search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
labkey-adam: I'm hopeful that that's not an issue here, but I'm switching the SecurityQuery check to case-insensitive. Trust and verify.

PR #6800, opened 2025-06-26
Add various metrics for sample types and data classes

labkey-susanh:
#### Rationale
We want to track a few more metrics to inform future feature development:
- number of non-media sample types that do not have a metric unit defined 
- number of imports/updates/merges for sources and samples that use a pipeline job
- number of imports/updates/merges for sources and samples that fail because of a validation error

#### Changes
- Add metrics in `AbstactQueryImportAction`
- Add metric in `ExperimentModule`


experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -720,6 +720,7 @@ WHERE op.propertyid IN (
                 results.put("sampleNullAmountCount", new SqlSelector(schema, "SELECT COUNT(*) FROM exp.material WHERE storedamount IS NULL").getObject(Long.class));
                 results.put("sampleNegativeAmountCount", new SqlSelector(schema, "SELECT COUNT(*) FROM exp.material WHERE storedamount < 0").getObject(Long.class));
                 results.put("sampleUnitsDifferCount", new SqlSelector(schema, "SELECT COUNT(*) from exp.material m JOIN exp.materialSource s ON m.materialsourceid = s.rowid WHERE m.units != s.metricunit").getObject(Long.class));
+                results.put("noSampleUnitsCount", new SqlSelector(schema, "SELECT COUNT(*) from exp.materialSource WHERE category IS NULL AND metricunit IS NULL").getObject(Long.class));
```
cnathe: metric works for me locally. Minor: The name to me indicates that we are counting sample rows instead of sample types without units.

cnathe (approved):

PR #6802, opened 2025-06-26
Issue 53313: Fields with non alpha/numeric chars not shown in Custom Props

labkey-jeckels:
#### Rationale
We're intending to show all of the admin-defined sample fields on its LKS details page but aren't looking them up in the map correctly.

#### Changes
- Consolidate handling for calculated and non-calculated fields
- Look up values correctly

### DIFF
--- a/api/src/org/labkey/api/data/SimpleDisplayColumn.java
+++ b/api/src/org/labkey/api/data/SimpleDisplayColumn.java
@@ -107,7 +107,7 @@ public class SimpleDisplayColumn extends DisplayColumn
     {
         Object value = getValue(ctx);
         if (value != null)
-            out.write(value.toString());
+            out.write(value);
     }
 
     @Override
--- a/experiment/src/org/labkey/experiment/CustomPropertiesView.java
+++ b/experiment/src/org/labkey/experiment/CustomPropertiesView.java
@@ -27,6 +27,7 @@ import org.labkey.api.exp.PropertyDescriptor;
 import org.labkey.api.exp.property.Domain;
 import org.labkey.api.exp.property.DomainProperty;
 import org.labkey.api.exp.query.SamplesSchema;
+import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.QueryService;
 import org.labkey.api.query.UserSchema;
 import org.labkey.api.security.User;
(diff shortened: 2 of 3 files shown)


experiment/src/org/labkey/experiment/CustomPropertiesView.java
```diff
@@ -147,18 +150,19 @@ public CustomPropertiesView(ExpMaterialImpl m, Container c, User u)
...
+                    propertyUris.add(property.getPropertyURI());
                 }
+
+                SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("lsid"), parentLSID);
+                Map<String,Object> tableProps = new TableSelector(queryTable, filter, null).getMap();
                 // include calculated fields from the domain / query as well
                 List<ColumnInfo> cols = queryTable.getColumns().stream()
                         .filter(ColumnInfo::isShownInDetailsView)
-                        .filter(ColumnInfo::isValueExpressionColumn)
+                        .filter(col -> propertyUris.contains(col.getPropertyURI()) || col.isValueExpressionColumn())
```
cnathe, line 163: Tested locally, and this works to now show the properties with special characters. Note that we are still only showing properties that have a non-null value because of the if statement below (line 170). I could see changing that to always show all properties regardless of if there is a value or not....but that is not the point of this PR change.

cnathe (approved):

cnathe (approved):

experiment/src/org/labkey/experiment/CustomPropertiesView.java
labkey-jeckels: I pushed changes to show these null fields, as well as test coverage in the related PR.

api/src/org/labkey/api/data/SimpleDisplayColumn.java
```diff
@@ -107,7 +107,7 @@ public void renderDetailsCellContents(RenderContext ctx, HtmlWriter out)
     {
         Object value = getValue(ctx);
         if (value != null)
-            out.write(value.toString());
+            out.write(value);
```
labkey-jeckels, line 110: @labkey-adam FYI, here was another spot we were HTML-encoding an `HTMLString`. I happened to stumble on it when working on this issue.

PR #6803, opened 2025-06-26
Parse NUMERIC/DECIMAL with (precision,scale)

labkey-matthewb:
#### Rationale
For control for formatting it is useful to to be able to explicitly cast to NUMERIC(prec,scale).
Also, add SELECT annotation support `@format="#.00"` (needs doc)

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/query/src/org/labkey/query/sql/Method.java
+++ b/query/src/org/labkey/query/sql/Method.java
@@ -685,30 +685,21 @@ public abstract class Method
         @Override
         public SQLFragment getSQL(SqlDialect dialect, SQLFragment[] fragments)
         {
-            JdbcType jdbcType;
-            SQLFragment length = null;
             if (fragments.length >= 2)
             {
-                String sqlEscapeTypeName = getTypeArgument(fragments[1]);
                 try
                 {
-                    jdbcType = ConvertType.valueOf(sqlEscapeTypeName).jdbcType;
-                    String typeName = dialect.getSqlTypeName(jdbcType);
-                    if (null == typeName)
-                        throw new NullPointerException("No sql type name found for '" + jdbcType.name() + "' in " + dialect.getProductName() + " database");
-                    if (fragments.length > 2)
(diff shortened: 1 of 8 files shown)


query/src/org/labkey/query/sql/Method.java
```diff
@@ -657,21 +657,9 @@ public ConvertInfo()
...
-                }
-                catch (IllegalArgumentException x)
-                {
-                    /* */
-                }
-            }
-
+            JdbcType jdbcType = JdbcType.OTHER;
+            for (ColumnInfo argument : arguments)
+                jdbcType = argument.getJdbcType();
```
labkey-adam, line 662: Why enumerate them? Why not just:
```
JdbcType jdbcType = arguments.length > 0 ? arguments[arguments.length - 1].getJdbcType() : JdbcType.OTHER;
```

labkey-adam (changes requested):
One coding suggestion. Also, I didn't see a test for `@format`.

query/src/org/labkey/query/sql/Method.java
labkey-matthewb: I started writing it that way.  It seemed like a lot of characters for "use the last one"

labkey-matthewb:
> One coding suggestion. Also, I didn't see a test for `@format`.

Do you know of any Junit tests that test DataColumn/DisplayColumn?  I guess I can just peak at the ColumnInfo directly.

labkey-adam:
> > One coding suggestion. Also, I didn't see a test for `@format`.
> 
> Do you know of any Junit tests that test DataColumn/DisplayColumn? I guess I can just peak at the ColumnInfo directly.

There's `NonStandardDateAndTimeFormatTest` which tests formats in lists and data classes. You could maybe extend that with a query over the list.

labkey-adam (approved):

PR #6804, opened 2025-06-27
Issue 53306: Some LKS forms don't distinguish between fields properly

labkey-jeckels:
#### Rationale
Some `<input>` fields are using `name` attributes that don't differentiate from similarly named fields.

#### Changes
- Eliminate `ColumnInfo.getPropertyName()`


api/src/org/labkey/api/data/DisplayColumnGroup.java
```diff
@@ -89,7 +89,8 @@ public void writeSameCheckboxCell(RenderContext ctx, HtmlWriter out)
         ).appendTo(out);
     }
 
-    private String getGroupFormFieldName(RenderContext ctx)
+    /** Use propName because DOM ids and function names can't have spaces */
+     private String getGroupFormFieldName(RenderContext ctx)
```
labkey-nicka: nit: Remove indentation

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -480,7 +480,7 @@ else if (_validateRequired && null != _tinfo)
                     Container container = fk.getLookupContainer() != null ? fk.getLookupContainer() : getContainer();
                     try
                     {
-                        Object remappedValue = cache.remap(SchemaKey.fromParts(fk.getLookupSchemaName()), fk.getLookupTableName(), getUser(), container, ContainerFilter.Type.CurrentPlusProjectAndShared, str);
+                        Object remappedValue = cache.remap(fk.getLookupSchemaKey(), fk.getLookupTableName(), getUser(), container, ContainerFilter.Type.CurrentPlusProjectAndShared, str);
```
labkey-nicka, line 483: Similar change can be made in `ExpRunTableImpl.java:1049`.

api/src/org/labkey/api/data/AbstractWrappedColumnInfo.java
```diff
@@ -215,12 +215,6 @@ public boolean isShouldLog()
         return delegate.isShouldLog();
     }
 
-    @Override
```
labkey-matthewb, line 218: Yeah!

issues/src/org/labkey/issue/IssuesController.java
```diff
@@ -721,7 +721,7 @@ public JSONArray getIssues()
         /**
          * Parse out the issues forms from the JSON data
          */
-        public List<IssuesForm> getIssueForms()
+        public List<IssuesForm> getIssueForms(User user, Container container)
```
labkey-nicka: nit: Initializing a shared object here `_issueForms` with non-constant values for user and container could result in an improper result.

study/test/src/org/labkey/test/tests/search/SearchTest.java
```diff
@@ -451,9 +451,9 @@ private void addSearchableIssues()
         goToModule("Issues");
         File file = TestFileUtils.getSampleData("fileTypes/tsv_sample.tsv");
         issuesHelper.addIssue(ISSUE_TITLE, _userHelper.getDisplayNameForEmail(USER1),
-                Maps.of("type", "UFO",
-                        "area", "Area51",
-                        "priority", "1",
+                Maps.of("Type", "UFO",
+                        "Area", "Area51",
+                        "Priority", "1",
```
labkey-nicka, line 456: Why is "Comments" capitalized in mothership but then "comments" here in issues? Does matching case actually matter I guess is what I'm really asking.

labkey-nicka (approved):

issues/src/org/labkey/issue/IssuesController.java
labkey-matthewb: > nit: Initializing a shared object here `_issueForms` with non-constant values for user and container could result in an improper result.

Which objects are shared?

labkey-nicka: The `_issueForms` is shared but created given variable input.

labkey-matthewb: But IssuesApiForm is allocated per request.

labkey-matthewb: Maybe your point is that User and Container don't need to be passed in at all?

labkey-nicka: My point is that there are multiple calls to `getIssueForms(User user, Container container)` that currently all pass in the same container/user. However, if one did not, and it was different, say an admin user or something, it could end up returning an instance that is not initialized with the admin user but then continuing work assuming it is.

labkey-matthewb: Right, which is why we have getViewContext() for Action classes.

labkey-jeckels: I'll make `IssuesApiForm` non-static so it can get the user and container from the `ViewContext`. This more clearly binds it to an individual request.

api/src/org/labkey/api/data/DisplayColumnGroup.java
labkey-jeckels: Fixed

labkey-matthewb (approved):
This looks great.  My only thought is that the remaining usages of propNameFromName seem to be for id attributes?  Maybe we can put that last nail in this coffin by consolidating with PageConfig.makeId()??   Maybe PageConfig.makeIdFromName()?

api/src/org/labkey/api/data/TableViewForm.java
labkey-jeckels: Done

study/test/src/org/labkey/test/tests/search/SearchTest.java
labkey-jeckels: I don't love the mix, but the issues form is custom-coded. It now has a mix of camel and capitalized fields, depending on whether they're lookups or not. Mothership is generated by a more standard update form codepath.

labkey-jeckels:
> This looks great. My only thought is that the remaining usages of propNameFromName seem to be for id attributes? Maybe we can put that last nail in this coffin by consolidating with PageConfig.makeId()?? Maybe PageConfig.makeIdFromName()?

Yes, plus for making script-friendly names in a couple of places in report code. I moved it from ColumnInfo to PageConfig and renamed it. I didn't attempt to make them all unique though. See my most recent commit in this FB.

PR #6805, opened 2025-06-27
Issue 52339: Try to use shorter URIs for domains when possible

labkey-susanh:
#### Rationale
Issue [52339](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52339)

This doesn't address the creation of long domain URIs that might exist out there, but paves the way toward shorter URIs in other scenarios.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2522

#### Changes
- Move `getLsidPrefixDbSeq` to `LsidManager` for better accessibility
- Use `getLsidPrefixDbSeq` for creating domain URIs for lists
- Update `DefaultValueServiceImpl` to try to use these shorter URIs as appropriate as well.

### DIFF
--- a/api/src/org/labkey/api/exp/LsidManager.java
+++ b/api/src/org/labkey/api/exp/LsidManager.java
@@ -23,6 +23,8 @@ import org.labkey.api.assay.AssayProvider;
 import org.labkey.api.assay.AssayUrls;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
+import org.labkey.api.data.DbSequence;
+import org.labkey.api.data.DbSequenceManager;
 import org.labkey.api.data.SimpleFilter;
 import org.labkey.api.data.Sort;
 import org.labkey.api.data.TableSelector;
@@ -70,6 +72,15 @@ public class LsidManager
     public static LsidManager get()
     {
         return INSTANCE;
+    }
+
+    public static DbSequence getLsidPrefixDbSeq(Container container, String lsidPrefix, int batchSize)
+    {
+        Container projectContainer = container; // use DBSeq at project level to avoid duplicate lsid for types in child folder
+        if (!container.isProject() && container.getProject() != null)
+            projectContainer = container.getProject();
+
+        return DbSequenceManager.getPreallocatingSequence(projectContainer, ExperimentService.LSID_COUNTER_DB_SEQUENCE_PREFIX + lsidPrefix, 0, batchSize);
     }
 
     public interface LsidHandler<I extends Identifiable>
--- a/api/src/org/labkey/api/exp/LsidManager.java
+++ b/api/src/org/labkey/api/exp/LsidManager.java
@@ -74,6 +74,11 @@ public class LsidManager
         return INSTANCE;
     }
 
+    public static DbSequence getLsidPrefixDbSeq(String lsidPrefix, int batchSize)
+    {
+        return getLsidPrefixDbSeq(ContainerManager.getRoot(), lsidPrefix, batchSize);
+    }
+
     public static DbSequence getLsidPrefixDbSeq(Container container, String lsidPrefix, int batchSize)
     {
         Container projectContainer = container; // use DBSeq at project level to avoid duplicate lsid for types in child folder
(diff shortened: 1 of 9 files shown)


api/src/org/labkey/api/exp/LsidManager.java
```diff
@@ -72,6 +74,15 @@ public static LsidManager get()
         return INSTANCE;
     }
 
+    public static DbSequence getLsidPrefixDbSeq(Container container, String lsidPrefix, int batchSize)
```
labkey-jeckels, line 82: Presumably this is safe because container RowIds are included in many LSIDs, but since containers can be moved, would it make sense to have a single sequence used server-wide?

labkey-jeckels (commented):
One comment but looks promising.

api/src/org/labkey/api/exp/LsidManager.java
labkey-susanh: I can certainly change the usage for Lists so it uses the root container. If this method is updated now, we'd need to change the usage from the Experiment module so it can skip over LSIDs already in use from the project-scoped LSIDs. 

@XingY Was there a reason not to make this server-wide when you were adding these sequences in `ExperimentServiceImpl`?

XingY: Historically we had always included folder / container as part of lsid. The change I introduced was switching from data type name to use a db sequence when generating lsid, in order to support datatype/data renaming. LSIDs are not updated during a folder move, and shouldn't cause issues/conflict.

labkey-susanh: It isn't a move that would cause the conflict but the restarting of the sequence if we switch it to use a different container where the conflict may occur. (Previous project's sequence value of 1 vs. new project's sequence value of 1.) But, yes, when there is a folder id in the LSID, there won't be a problem here.

list/src/org/labkey/list/model/ListDomainKind.java
```diff
@@ -268,20 +270,18 @@ public Set<PropertyStorageSpec.Index> getPropertyIndices(Domain domain)
         return Collections.emptySet(); // TODO: Allow this to return the Key Column
     }
 
-    public static Lsid generateDomainURI(String name, Container c, KeyType keyType, @Nullable ListDefinition.Category category)
+    public static Lsid generateDomainURI(Container c, KeyType keyType, @Nullable ListDefinition.Category category)
     {
         String type = getType(keyType, category);
```
XingY, line 275: Not related to this PR, but surprised to see category is hard wired to picklist.
`if (category != null)
                    type = PicklistDomainKind.NAMESPACE_PREFIX;`

experiment/src/org/labkey/experiment/defaults/DefaultValueServiceImpl.java
```diff
@@ -59,13 +60,28 @@ public class DefaultValueServiceImpl implements DefaultValueService
     private String getContainerDefaultsLSID(Container container, Domain domain)
     {
         String suffix = "Folder-" + container.getRowId();
-        return (new Lsid(DOMAIN_DEFAULT_VALUE_LSID_PREFIX, suffix, domain.getName())).toString();
+        DomainKind<?> kind = domain.getDomainKind();
```
XingY, line 63: Would this cause backward compatibility issue? If a default value is set prior to this change, would it be lost since the default lsid is different now?

labkey-susanh: I think this will be ok because when the domain is user-defined, the objectId from the domain LSID is the same as the encoded name. I tested it locally, but it would be good to double check by creating a list while on develop that has some default values defined for properties then verifying the default values still show up when using the list on this branch.

XingY (approved):

api/src/org/labkey/api/exp/LsidManager.java
labkey-susanh: My future self may not thank me for this, but I'm opting for leaving the `Experiment` usages as is and adding a method that provides for an easy way to use a root container sequence, which we'll use for lists.

PR #6807, opened 2025-06-27
Issue list default values shouldn't render in the detail views.

labkey-klum:
#### Rationale
related issue: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53343
 
This was caused by a helper function which automatically set the visible state to `true` and readonly to `false`. It was used as a convenience for the update view and unfortunately got propagated to the detail view. I've deleted that variant so users have to be explicit when setting those states.

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/2525

#### Tasks
- [x] Manual Testing
- [x] Needs Automation
- [x] Verify Fix

### DIFF
--- a/issues/src/org/labkey/issue/model/IssuePage.java
+++ b/issues/src/org/labkey/issue/model/IssuePage.java
@@ -400,11 +400,6 @@ public class IssuePage implements DataRegionSelection.DataSelectionKeyForm
         return _tableInfo;
     }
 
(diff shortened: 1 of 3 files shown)


labkey-adam (approved):
Looks good. I'll pull it down and verify.

labkey-adam:
@labkey-klum I did a quick local verify, so I think you can consider this verified and manually tested... unless you or someone else had more in mind

PR #6811, opened 2025-06-30
Add null check for target table before adding metric

labkey-susanh:
#### Rationale
Null check for target table missed in `AbstractQueryimportAction`

#### Related Pull Requests
- #6800 

#### Changes
- Add missing null check

### DIFF
--- a/api/src/org/labkey/api/query/AbstractQueryImportAction.java
+++ b/api/src/org/labkey/api/query/AbstractQueryImportAction.java
@@ -604,7 +604,8 @@ public abstract class AbstractQueryImportAction<FORM> extends FormApiAction<FORM
                             QueryImportPipelineJob job = new QueryImportPipelineJob(getQueryImportProviderName(), info, root, importContextBuilder);
                             PipelineService.get().queueJob(job, getQueryImportJobNotificationProviderName());
 
-                            SimpleMetricsService.get().increment("query", "fileBackgroundImports", getMetricPrefix(_target));
+                            if (_target != null)
+                                SimpleMetricsService.get().increment("query", "fileBackgroundImports", getMetricPrefix(_target));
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/query/AbstractQueryImportAction.java
```diff
@@ -604,7 +604,8 @@ else if (!dataFileDir.exists())
                             QueryImportPipelineJob job = new QueryImportPipelineJob(getQueryImportProviderName(), info, root, importContextBuilder);
                             PipelineService.get().queueJob(job, getQueryImportJobNotificationProviderName());
 
-                            SimpleMetricsService.get().increment("query", "fileBackgroundImports", getMetricPrefix(_target));
+                            if (_target != null)
+                                SimpleMetricsService.get().increment("query", "fileBackgroundImports", getMetricPrefix(_target));
```
cnathe, line 608: it looks like there are 3 usages of `getMetricPrefix` in this file. 2 of them now do null checks. Should that be added to the third? Or should the null check be moved into `getMetricPrefix`? 

Unrelated, I noticed that `getMetricPrefix` is public but only used in this file, should/could it be made private?

cnathe (changes requested):

api/src/org/labkey/api/query/AbstractQueryImportAction.java
labkey-susanh: The third usage is already inside a check for `target != null`.

Sure, I can make the method private. I think it may have needed to be public in some iteration here.

cnathe: sounds good. your call. thanks for checking on those things.

cnathe (approved):

PR #6818, opened 2025-07-02
Well table contextual role support

labkey-klum:
#### Rationale
[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53194)

We patched around the related issue to avoid the NPE, but the fix was more of a stopgap since it avoided the join to the `Well` table if the user didn't have read access. The appropriate fix is to support and push a `ContextualRole` through the `SampleSchema` and into the `PlateSchema` so it can be used in the `Well` table.

### DIFF
--- a/assay/src/org/labkey/assay/plate/query/PlateSchema.java
+++ b/assay/src/org/labkey/assay/plate/query/PlateSchema.java
@@ -16,6 +16,7 @@
 
 package org.labkey.assay.plate.query;
 
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.collections.CaseInsensitiveTreeSet;
 import org.labkey.api.data.Container;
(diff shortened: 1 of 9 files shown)


assay/src/org/labkey/assay/plate/query/PlateSchema.java
```diff
@@ -48,12 +52,21 @@ public class PlateSchema extends SimpleUserSchema
         AmountUnitsTable.NAME
     ));
 
+    Set<Role> contextualRoles = Set.of();
```
labkey-matthewb, line 55: Should/can this be final?

experiment/src/org/labkey/experiment/api/ExpMaterialTableImpl.java
```diff
@@ -780,8 +802,14 @@ public void addQueryFieldKeys(Set<FieldKey> keys)
         if (InventoryService.get() != null && (st == null || !st.isMedia()))
             defaultCols.addAll(InventoryService.get().addInventoryStatusColumns(st == null ? null : st.getMetricUnit(), this, getContainer(), _userSchema.getUser()));
 
-        UserSchema plateUserSchema = QueryService.get().getUserSchema(_userSchema.getUser(), getContainer(), "plate");
         SQLFragment sql;
+        UserSchema plateUserSchema;
+        // Issue 53194 : this would be the case for linked to study samples
+        if (_userSchema instanceof UserSchema.HasContextualRoles samplesSchema && !samplesSchema.getContextualRoles().isEmpty())
```
labkey-matthewb, line 810: Please add a longer explanation here why it is always OK for the plate schema to inherit the contextual roles of "this" schema.

labkey-matthewb (approved):

PR #6820, opened 2025-07-03
Issue 52657: LKSM: We shouldn't allow creating sample names that differ only in case.

XingY:
#### Rationale
There has been inconsistency when it comes allowing case sensitive duplicate sample/data names. With this PR, duplicate names that varies only in cases should no longer be allowed.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1828
- https://github.com/LabKey/platform/pull/6820
- https://github.com/LabKey/limsModules/pull/1510
- https://github.com/LabKey/testAutomation/pull/2541

#### Changes
- new DuplicateNameCheckIteratorBuilder to check for case insensitive duplicate during insert/import/update/merge
- check for duplicate case insensitive names during sample/data in `QUS._update`
- [Deadlock](https://www.labkey.org/_mothership/mothership-showStackTraceDetail.view?exceptionStackTraceId=39487) as result of aliquot rollup calculation and search indexer.  Process rollup samples in ascending order to match that of indexer, to reduce deadlock


api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -212,9 +214,14 @@ public Map<Integer, Map<String, Object>> getExistingRows(User user, Container co
                     if (!container.getId().equals(dataContainer))
                         throw new InvalidKeyException("Data doesn't belong to folder '" + container.getName() + "': " + key.getValue().values());
                 }
+                // sql server will return case-insensitive match
```
labkey-susanh: I'm not sure how this comment relates to the lines below since there's nothing that says case-sensitive or -insensitive to me.

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -3155,6 +3161,104 @@ public static void incrementCounts(Map<String, Integer> currentCounts, Map<Strin
...
+            Object nameObj = get(_nameCol);
+            if (nameObj == null)
+                return hasNext;
+
+            String newName = String.valueOf(nameObj);
+            if (StringUtils.isEmpty(newName))
+                return hasNext;
+
+            Map<String, Object> existingValues = getExistingRecord();
+            if (existingValues != null  && !existingValues.isEmpty() && existingValues.get(NAME_FIELD).equals(newName))
```
labkey-susanh, line 3227: Should this be `equalsIgnoreCase`?

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -9137,6 +9137,35 @@ public Map<String, Map<String, Object>> getDomainMetrics()
         return metrics;
     }
 
+    public boolean isValidNewOrExistingName(String newOrExistingName, @NotNull TableInfo tableInfo, boolean allowExisting)
+    {
+        SQLFragment dataRowSQL = new SQLFragment("SELECT name FROM ")
+                .append(tableInfo)
+                .append(" WHERE LOWER(name) = LOWER(?)")
+                .add(newOrExistingName);
+        if (allowExisting) // // exclude existing name for merge
```
labkey-susanh: Aren't we allowing existing names for merge?

labkey-susanh (approved):

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
XingY: Updated comment to explain why we need to do a java `equals` check after sql =

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: updated comment

experiment/src/org/labkey/experiment/ExpDataIterators.java
XingY: No, if the case is different, we need to check if renaming is allowed for update and merge. For merge, the work put in is just theoretically at the moment. We don't allow support merge operations that updates names.

PR #6821, opened 2025-07-04
Issue 53253: forego indexing data class in a container being deleted

labkey-nicka:
#### Rationale
This addresses [Issue 53253](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53253) by checking that the `dataClass` instance that is about to be indexed has a container.

#### Changes
- Check for data class container prior to indexing
- Annotate `ExpDataClassImpl.getTinfo()` as `@NotNull` and update usages
- Add assertion check for container in `ExpDataClassImpl.getDomain()`


labkey-susanh (approved):
Nice!

PR #6822, opened 2025-07-07
Issue 53400: resolve AliquotedFrom from a case-insensitive map

labkey-nicka:
#### Rationale
This addresses [Issue 53400](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53400) by using a `CaseInsensitiveHashMap` when processing input values in `_GenerateNamesDataIterator`.

#### Changes
- Use a `CaseInsensitiveHashMap` when mapping input values for aliquot name generation
- Add regression tests

#### Tasks
- [x] Manual Testing @XingY 
- [x] Needs Automation

### DIFF
--- a/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
+++ b/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
@@ -1688,7 +1688,7 @@ public class SampleTypeUpdateServiceDI extends DefaultQueryUpdateService
         @Override
         protected void processNextInput()
         {
-            Map<String, Object> map = new HashMap<>(((MapDataIterator)getInput()).getMap());
+            Map<String, Object> map = new CaseInsensitiveHashMap<>(((MapDataIterator)getInput()).getMap());
 
             String aliquotedFrom = null;
             Object aliquotedFromObj = map.get(ExpMaterial.ALIQUOTED_FROM_INPUT);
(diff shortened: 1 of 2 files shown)


experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -1688,7 +1688,7 @@ void onFirst()
         @Override
         protected void processNextInput()
         {
-            Map<String, Object> map = new HashMap<>(((MapDataIterator)getInput()).getMap());
+            Map<String, Object> map = new CaseInsensitiveHashMap<>(((MapDataIterator)getInput()).getMap());
```
XingY, line 1691: Safer to fix this (additionally) at NameGeneratorState.additionalContext
`Map<FieldKey, Object> ctx = new CaseInsensitiveHashMap<>(NameGenerator.toFieldKeyMap(_batchExpressionContext));`

labkey-nicka: I don't think we need to make any further changes in this regard. That also doesn't support `CaseInsensitiveHashMap` since it is keyed by `FieldKey` which already does case-insensitive comparison.

XingY: > I don't think we need to make any further changes in this regard. That also doesn't support `CaseInsensitiveHashMap` since it is keyed by `FieldKey` which already does case-insensitive comparison.

You are right, FieldKey is already case insensitive when comparing.

XingY (approved):

PR #6831, opened 2025-07-09
Issue 53412: Don't return names or container paths if user doesn't have read permission

labkey-susanh:
#### Rationale
Issue [53412](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53412): Guest Users can figure out Sample/Source Names and their Container Paths if they have access to any folder on a LabKey server

#### Changes
- Check for read permission to containers before adding them or the names of objects in them to the response

#### Tasks
- [x] Manual Testing @labkey-alan 
- [x] ~Needs Automation~ - See [Issue 53415](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53415) 
- [x] Verify Fix @labkey-alan

### DIFF
--- a/assay/src/org/labkey/assay/AssayController.java
+++ b/assay/src/org/labkey/assay/AssayController.java
@@ -1706,7 +1706,8 @@ public class AssayController extends SpringActionController
             for (ExpRun expRun : allRuns)
             {
                 Container c = expRun.getContainer();
-                containers.add(c);
+                if (c.hasPermission(getUser(), ReadPermission.class))
+                    containers.add(c);
                 if (permClass != null && !c.hasPermission(getUser(), permClass))
                     notPermittedIds.add(expRun.getRowId());
             }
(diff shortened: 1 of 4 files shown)


labkey-alan (approved):
I am approving this because fundamentally this does resolve the issue. However, I couldn't find anywhere in our code that actually relies on the Name being returned anywhere, it's possible I just missed something. If we really aren't relying on the Name field anywhere then I think we should consider dropping it from the response, as this would reduce the surface area of a leak if a similar bug is introduced later. As far as I can tell our clients do rely on the RowId and ContainerPath being returned.

labkey-susanh:
> I am approving this because fundamentally this does resolve the issue. However, I couldn't find anywhere in our code that actually relies on the Name being returned anywhere, it's possible I just missed something. If we really aren't relying on the Name field anywhere then I think we should consider dropping it from the response, as this would reduce the surface area of a leak if a similar bug is introduced later. As far as I can tell our clients do rely on the RowId and ContainerPath being returned.

Removed the Name field and reran tests without any hiccups that seem related to that change.  

(One could argue that having the Name field protected by the permission check is better assurance that it won't get added back without the permission check later, but that's no way to write code.)

PR #6837, opened 2025-07-10
Fix sql LIKE usage to properly escape special characters

XingY:
#### Rationale
This has been an ongoing intermittent [TC failure](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_WorkflowEndpointsSqlserver/3576321) on sql server that fails to correctly increment withCounter sequences, when the string contains special characters. Wildcard characters need to be escaped for LIKE operators in SQL. 

#### Related Pull Requests
- 

#### Changes
- 

#### Tasks 📍
~~- [ ] Manual Testing ~~
~~Needs Automation (No, already exist)~~
- [x] code review

### DIFF
--- a/api/src/org/labkey/api/data/dialect/SqlDialect.java
+++ b/api/src/org/labkey/api/data/dialect/SqlDialect.java
@@ -2181,7 +2181,7 @@ public abstract class SqlDialect
             assertEquals("SELECT * FROM A WHERE Name " + d.getCaseInsensitiveLikeOperator() + " '%a![b]C%' ESCAPE '!'", d.appendCaseInsensitiveLikeClause(new SQLFragment("SELECT * FROM A WHERE Name"), "a[b]C").toDebugString());
             assertEquals("SELECT * FROM A WHERE Name " + d.getCaseInsensitiveLikeOperator() + " 'a![b]C_' ESCAPE '!'", d.appendCaseInsensitiveLikeClause(new SQLFragment("SELECT * FROM A WHERE Name"), "a[b]C", null, "_").toDebugString());
             assertEquals("SELECT * FROM A WHERE Name " + d.getCaseInsensitiveLikeOperator() + " '_a!_![b]C%' ESCAPE '!'", d.appendCaseInsensitiveLikeClause(new SQLFragment("SELECT * FROM A WHERE Name"), "a_[b]C", "_", "%").toDebugString());
-            assertEquals("SELECT * FROM A WHERE Name " + d.getCaseInsensitiveLikeOperator() + " '_a[[[]_[[b]C!d%' ESCAPE '['", d.appendCaseInsensitiveLikeClause(new SQLFragment("SELECT * FROM A WHERE Name"), "a_[b]C!d", "_", "%", '[').toDebugString());
+            assertEquals("SELECT * FROM A WHERE Name " + d.getCaseInsensitiveLikeOperator() + " '_a[_[[b]C!d%' ESCAPE '['", d.appendCaseInsensitiveLikeClause(new SQLFragment("SELECT * FROM A WHERE Name"), "a_[b]C!d", "_", "%", '[').toDebugString());
         }
     }
 }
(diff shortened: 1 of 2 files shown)


labkey-nicka (approved):

api/src/org/labkey/api/data/dialect/SqlDialect.java
```diff
@@ -535,6 +536,40 @@ public SQLFragment appendInClauseSql(SQLFragment sql, @NotNull Collection<?> par
         return DEFAULT_GENERATOR.appendInClauseSql(sql, params);
     }
 
+    public SQLFragment appendCaseInsensitiveLikeClause(SQLFragment sql, @NotNull String matchStr, @Nullable String wildcardPrefix, @Nullable String wildcardSuffix, char escapeChar)
+    {
+        String prefix = wildcardPrefix != null ? wildcardPrefix : "";
+        String suffix = wildcardSuffix != null ? wildcardSuffix : "";
+        String prefixLike = prefix + CompareType.escapeLikePattern(matchStr, escapeChar) + suffix;
```
labkey-nicka, line 543: Should this be using `concatenate()` as furnished by the dialect?

api/src/org/labkey/api/data/dialect/SqlDialect.java
```diff
@@ -535,6 +536,40 @@ public SQLFragment appendInClauseSql(SQLFragment sql, @NotNull Collection<?> par
...
+        String escapeToken = " ESCAPE '" + escapeChar + "'";
+        sql.append(" ")
+                .append(getCaseInsensitiveLikeOperator())
+                .append(" ")
+                .appendValue(prefixLike)
+                .append(escapeToken);
+        return sql;
+    }
+
+    public SQLFragment appendCaseInsensitiveLikeClause(SQLFragment sql, @NotNull String matchStr, @Nullable String wildcardPrefix, @Nullable String wildcardSuffix)
```
labkey-nicka, line 553: I recommend unit tests for each of these incantations.

api/src/org/labkey/api/data/dialect/SqlDialect.java
```diff
@@ -535,6 +536,40 @@ public SQLFragment appendInClauseSql(SQLFragment sql, @NotNull Collection<?> par
         return DEFAULT_GENERATOR.appendInClauseSql(sql, params);
     }
 
+    public SQLFragment appendCaseInsensitiveLikeClause(SQLFragment sql, @NotNull String matchStr, @Nullable String wildcardPrefix, @Nullable String wildcardSuffix, char escapeChar)
```
labkey-nicka, line 539: nit: It would be nice to get `ConvertType.QClause.toSQLFragment()` converted over to using this.

api/src/org/labkey/api/data/dialect/SqlDialect.java
labkey-matthewb: If matchStr were not constant (e.g.  SQLFragment) then yes, but since these are all constant strings, this makes more sense.

api/src/org/labkey/api/data/dialect/SqlDialect.java
```diff
@@ -535,6 +536,40 @@ public SQLFragment appendInClauseSql(SQLFragment sql, @NotNull Collection<?> par
...
+    {
+        String prefix = wildcardPrefix != null ? wildcardPrefix : "";
+        String suffix = wildcardSuffix != null ? wildcardSuffix : "";
+        String prefixLike = prefix + CompareType.escapeLikePattern(matchStr, escapeChar) + suffix;
+        String escapeToken = " ESCAPE '" + escapeChar + "'";
+        sql.append(" ")
+                .append(getCaseInsensitiveLikeOperator())
+                .append(" ")
+                .appendValue(prefixLike)
+                .append(escapeToken);
```
labkey-matthewb, line 549: I think escapeToken is always !, but it would be better to use quoteStringLiteral()

XingY: For sql server, quoteStringLiteral adds a "N" prefix and that doesn't work for the escape character.

labkey-matthewb: Got it

api/src/org/labkey/api/data/dialect/SqlDialect.java
XingY: junit tests added

PR #6839, opened 2025-07-11
Issue 53446: misconfigured plate assay design, plate validation query optimization

labkey-nicka:
#### Rationale
This addresses [Issue 53446](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53446) by skipping assay designs that do not have the columns required to determine their run count in relation to plates. Additionally, this optimizes the query for validating unique samples in a primary plate set by making use of a CTE to define the set of plate sets to filter on.

#### Changes
- Skip attempting to count runs against plate-based assay designs which are improperly configured
- Query optimization for validating unique samples in a primary plate set

#### Tasks
- [x] Needs Automation (Already in place)

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateManager.java
+++ b/assay/src/org/labkey/assay/plate/PlateManager.java
@@ -3387,17 +3387,18 @@ public class PlateManager implements PlateService, AssayListener, ExperimentList
         TableInfo wellTable = dbSchema.getTableInfoWell();
 
         // Determines the set of primary plate sets that are being touched from the collection of well rowIds
-        SQLFragment primaryPlateSetsFromWellRowIdsSQL = new SQLFragment("SELECT PS.RowId FROM ").append(wellTable, "W")
-                .append(" INNER JOIN ").append(plateTable, "P").append(" ON P.RowId = W.PlateId")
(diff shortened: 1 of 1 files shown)


labkey-chrisj:
Manual Testing notes:
This FB addresses the issue described in Issue 53446-
The only way I could misconfigure an assay in this way is by:

1. disabling plate metadata, save
2. edit plate column from lookup->plate to lookup->sample type, save
3. re-enable plate metadata, save

At this point, the assays list in a plate set renders without error
Going to the assay upload page renders no errors
As we might expect, entering data into the grid here resolves samples in the sampletype the Plates field looks to
Also predictably, attempting to save data referencing a sample and not a plate by rowId causes an error (the gist of which is: there's no matching plate in the plate set specified here).  Theoretically if the sample rowID happened to be within the range of plate rowIDs we would miss this error and the user's data integrity would be toast.
On the one hand, this is very much an edge case that could be reduced to 'we let users configure the product in ways that don't work'.
On the other, protecting our users from making bad choices with built-in fields might require us to think long and deliberately about how far we're willing to limit user choices in the name of stability.

labkey-klum (approved):

PR #6840, opened 2025-07-11
Issue 53431: Data Class and Sample Type data doesn't round-trip via folder export/import for field names with special char

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53431

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6840
- https://github.com/LabKey/limsModules/pull/1520
- https://github.com/LabKey/testAutomation/pull/2556

#### Changes
- use ColumnHeaderType.ImportField instead of FieldKey for writing folder archive tsv data column headers
- PlateManager.exportPlateData to use ColumnHeaderType.ImportField

#### Tasks 
- [x] Manual Testing @labkey-danield 
- [x] Needs Automation
- [x] Verify Fix

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateManager.java
+++ b/assay/src/org/labkey/assay/plate/PlateManager.java
@@ -3768,7 +3768,7 @@ public class PlateManager implements PlateService, AssayListener, ExperimentList
                 try (TSVGridWriter writer = new TSVGridWriter(plateQueryView::getResults, displayColumns, Collections.singletonMap(sampleIdNameFieldKey.toString(), "Sample ID")))
                 {
                     writer.setDelimiterCharacter(delim);
-                    writer.setColumnHeaderType(ColumnHeaderType.FieldKey);
+                    writer.setColumnHeaderType(ColumnHeaderType.ImportField); // Issue 53431
                     writer.write(plateFileBytes.bytes);
                 }
 
--- a/api/src/org/labkey/api/data/ColumnHeaderType.java
+++ b/api/src/org/labkey/api/data/ColumnHeaderType.java
@@ -109,7 +109,7 @@ public enum ColumnHeaderType
         }
     },
 
-    // Use the ColumnInfo's FieldKey with FieldKey escaping. Useful for import/export round-tripping, but can lead to ugly names.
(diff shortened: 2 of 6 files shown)


assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3768,7 +3768,7 @@ public List<PlateFileBytes> exportPlateData(Container c, User user, ContainerFil
                 try (TSVGridWriter writer = new TSVGridWriter(plateQueryView::getResults, displayColumns, Collections.singletonMap(sampleIdNameFieldKey.toString(), "Sample ID")))
                 {
                     writer.setDelimiterCharacter(delim);
-                    writer.setColumnHeaderType(ColumnHeaderType.FieldKey);
```
labkey-nicka, line 3771: My main question is about the usage of these different `ColumnHeaderType`. Reading through the comments that surround them I'm still not sure which one to choose in which situation.

It looks like up to this point all the usages of `ImportField` align with usage in import templates, however, there are other usages of `ColumnHeaderType.FieldKey` that would seem to align with what you're using them for here.

Is `ImportField` just the new and improved `FieldKey` that we should be using more widely (I assume not)? I may be off-base here as you may have already discussed with @labkey-susanh who looks to familiarity with this enumeration.

cnathe: Yeah, I agree. This enum is confusing. The comment for `FieldKey`, which has been there since 2015 saying "Useful for import/export round-tripping" is wrong. We do not support FieldKey on import (see my attempt to add it here https://github.com/LabKey/platform/pull/6270). Maybe it is time to remove that part of the comment.

labkey-susanh: `ImportField` was added specifically because, as Cory said, round-tripping with `FieldKey` doesn't work currently.  Given this, I'm not sure when `FieldKey` is useful as a column header type. Perhaps it worked at some point, but I doubt it.  Since this targets 25.7, we shouldn't remove it altogether here, I think, but might consider that for develop.

labkey-susanh (approved):

PR #6841, opened 2025-07-14
Fix delete of allowed external resource hosts

labkey-adam:
#### Rationale
Deleting hosts that were saved in a previous server session seems to fail. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53457

#### Tasks 📍
- [x] Verify Fix
- [x] No additional automation - repro requires a server restart, so making automation impractical


labkey-jeckels (approved):
And manually tested

labkey-adam:
> And manually tested

Thanks!

PR #6847, opened 2025-07-15
File path related issue bundle

XingY:
#### Rationale
This PR include the following changes:
- Reject out-of-scope (path outside of file root) file paths and path that doesn't exist on insert/update/import or update via file. 
- Make sure display/messaging doesn’t reveal file existence information by using a generic error message for invalid file on import/update, and generic handling of rendering unavailable files without revealing the reason.
- Only show the file name and not file full path. This include disabling filter/sort for file columns and stop including full path in api responses and custom property views, etc. 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1833
- https://github.com/LabKey/platform/pull/6847
- https://github.com/LabKey/limsModules/pull/1524
- https://github.com/LabKey/testAutomation/pull/2582
- https://github.com/LabKey/commonAssays/pull/910

#### Changes
-


api/src/org/labkey/api/data/ExpDataFileConverter.java
```diff
@@ -341,7 +357,16 @@ private File convertToFile(Object value, @NotNull Container container, @NotNull
                 return result;
             }
         }
+
+        String filePath = value.toString();
+        if (filePath.startsWith("file:"))
+        {
+            URI uri = FileUtil.createUri(filePath);
+            if (FileUtil.FILE_SCHEME.equals(uri.getScheme()))
+                return new File(uri);
```
XingY, line 439: @labkey-matthewb Do you have any suggestions how to get around using `new File` and maybe use `FileLike` here? In order for `f.isAbsolute()` to work for path that starts with `file:/`, I had to convert string to uri first before passing to File constructor.

labkey-matthewb: If you are stuck in File land, use FileUtil.appendPath() and FileUtil.appendName() instead of new File(File,String).

To get out of File land, you have to start with getting FileLike for the directories you expect to find the files in.  E.g. PipeRoot.getRootFileLike() instead of PipeRoot.getPath().   Also PipeRoot has useful resolve() methods and isUnderRoot() methods.

api/src/org/labkey/api/data/ExpDataFileConverter.java
```diff
@@ -218,49 +219,101 @@ public Object convert(Class type, Object value)
         User user = (User) QueryService.get().getEnvironment(QueryService.Environment.USER);
         Object fileRootPathObj = QueryService.get().getEnvironment(QueryService.Environment.FILEROOTPATH);
```
labkey-matthewb, line 219: Consider putting a FileLike in the environment instead of  a String?  I don't love that we're using QueryService.get/setEnvironment() here, but that's a problem for another time.

api/src/org/labkey/api/data/ExpDataFileConverter.java
```diff
@@ -218,49 +219,101 @@ public Object convert(Class type, Object value)
...
 
                 PipeRoot root = PipelineService.get().getPipelineRootSetting(container);
                 if (root != null)
                 {
+                    File fileUnderRoot = f;
                     if (!f.isAbsolute())
                     {
                         // Interpret relative paths based on the file root
-                        f = new File(root.getRootPath(), f.getPath());
+                        fileUnderRoot = new File(root.getRootPath(), f.getPath());
```
labkey-matthewb, line 275: FileUtil.appendPath()

api/src/org/labkey/api/data/ExpDataFileConverter.java
labkey-matthewb: I should have rewritten this whole class to be ExpDataFileLikeConverter and killed usages convert(File.class).  Lets leave that cleanup for a separate PR.

api/src/org/labkey/api/data/ExpDataFileConverter.java
labkey-matthewb: 2nd thought.  Can we refactor this so there is a public convert() method that takes (user, fileRootPath, assayFilesPath) and then have _convert() call that method with these values?  Then we could have QUS and DataLoader call that method directly.  Other usages could still use still rely on thread-local storage until/unless we remove this from ConvertUtils.

api/src/org/labkey/api/reader/DataLoader.java
```diff
@@ -468,11 +465,48 @@ else if (renamedColumns.containsKey(name) && _columnInfoMap.containsKey(renamedC
                 ExceptionUtil.decorateException(e, ExceptionUtil.ExceptionInfo.SkipMothershipLogging, "true", true);
                 throw e;
             }
+
+            // use File converter for known file fields even if inferTypes = false. If inferTypes, this is already done.
+            if (!getInferTypes())
```
labkey-matthewb: I think this is where I get lost.  Why do we we rely on this?  I would hope that import would always work with no conversion in DataLoader.  (AFAIK, the only code that "should" rely on this is currently the infer type UI)

api/src/org/labkey/api/data/ExpDataFileConverter.java
XingY: I changed this to a static method that can be called directly. Currently there are 3 usages for importing assay results that could be refactored to use this static method. However, I didn't go ahead with the refactoring due to the much entangled run/result association. If we are removing this from ConvertUtils for sure then that's an effort worth trying. I've added a TODO comment.

api/src/org/labkey/api/data/ExpDataFileConverter.java
XingY: Unfortunately switching to FileUtil.appendPath broke valid relative path such as "../@files/test.txt". But maybe we can just say this is not supported.

api/src/org/labkey/api/data/ExpDataFileConverter.java
XingY: Changes made to no longer register ExpDataFileConverter for File.class

api/src/org/labkey/api/data/AbstractFileDisplayColumn.java
```diff
@@ -95,34 +92,35 @@ protected boolean isImage(String filename)
...
             // equivalent of DisplayColumn.renderURL.
             // Don't want to call renderUrl (DataColumn.renderUrl) to skip unnecessary displayValue check
             StringExpression s = compileExpression(ctx.getViewContext());
-            String url = null == s ? null : s.eval(ctx);
-            String displayName = getFileName(ctx, filename, true);
-            boolean isImage = isImage(filename);
-            FileImageRenderHelper renderHelper = createRenderHelper(ctx, url, filename, displayName, fileIconUrl, popupIconUrl, thumbnail, isImage);
+            String displayName = getFileName(ctx, fileValue);
+            boolean unavailable = displayName.endsWith(UNAVAILABLE_FILE_SUFFIX);
+            String url = null == s || unavailable ? null : s.eval(ctx);
```
labkey-matthewb, line 104: Nice check here.  It would be nice if getFileName() had an explicit structured return (e.g. record(String fileDisplay,Boolean available), but this is workable.  I wonder about the case in FileLinkDisplayColumn where we fall back to `result = f.getName()` when we don't find the file in an expected "file root".

What is the expected behavior in this case?  Should we treat as "unavailable"?  Should we render a URL?

labkey-matthewb (approved):
I'm not sure I can track all the subclasses here, but this looks like a good step forward.  I only have one question about files that exists on disk, but aren't in a known file root.  Would getFileName() benefit from having a structured return value?

And the tests will be super helpful.

api/src/org/labkey/api/data/AbstractFileDisplayColumn.java
XingY: In the case of `result = f.getName()` when we fail to find file in root, `valid` is false, and the "unavailable" suffix will be added. 

I added selenium test for grid and update view for LKS unavailable file both due to invalid path (exists but out side of file root) and absent file (file deleted after upload) in this [commit](https://github.com/LabKey/testAutomation/pull/2582/commits/57e00198311c86daf4a70149902f63e7e23aa63c)

PR #6848, opened 2025-07-15
Issue 53419: Aliquot parent with number like names that starts with leading zeroes aren't resolved during import

XingY:
#### Rationale
Issue [53419](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53419)
When the values are number like, the column will be inferred as numeric types with leading zeros removed. Previous PRs had forced inputs and parent alias to be inferred as String types. This PR force aliquotedFrom to be inferred as String for sample imports.

#### Related Pull Requests
- #6179 

#### Changes
- Force String type AliquotedFrom column inference for sample type import
- Add jest integration test for issue

#### Tasks 📍
- [x] Automated tests
- [x] code review @labkey-susanh 
- [x] Verify Fix


labkey-susanh:
### Testing notes
This is not related to this change:

If I provide the rowId of a sample (e.g., 4147339) instead of the name (Te-43474), the resulting aliquot name uses that rowId instead of the name (e.g., 4147339-1). Very much of an edge case as it is unlikely anyone will actually be providing rowIds in an import file and the solution is pretty simple (use the names instead), but if we do this via an API call, the results may be unexpected.

labkey-susanh (approved):

XingY:
> ### Testing notes
> This is not related to this change:
> 
> If I provide the rowId of a sample (e.g., 4147339) instead of the name (Te-43474), the resulting aliquot name uses that rowId instead of the name (e.g., 4147339-1). Very much of an edge case as it is unlikely anyone will actually be providing rowIds in an import file and the solution is pretty simple (use the names instead), but if we do this via an API call, the results may be unexpected.

As discussed, this behavior is debatable, but it's consistent with other lineage input (MaterialInputs/Type) or lookup. User can use MaterialInputs/Type/Name to get name instead.

PR #6849, opened 2025-07-16
Issue 53472: OutOfMemoryError during giant deletes from lists

labkey-jeckels:
#### Rationale
We can do better at baseline memory use and what's required for big query API operations.

#### Changes
- Use RowMap and reused canonical case maps for key query APIs
- Cache protocols by RowId instead of ExpProtocolImpl objects
- Log errors when we have conflicting data with different casing for keys
- Use empty maps when possible for caching external schemas

#### Tasks 📍
- [x] Manual Testing @labkey-tchad 📍 
  - Add and remove external schemas via the Schema Administration page. Ensure they appear and disappear as expected  in the admin UI and in the schema browser
  - Add and remove assay designs in the current folder, in the project, and in /Shared. Ensure they appear and disappear as expected as child schemas in the schema browser's assay schema.

### DIFF
--- a/api/src/org/labkey/api/assay/AssayService.java
+++ b/api/src/org/labkey/api/assay/AssayService.java
@@ -87,7 +87,8 @@ public interface AssayService
     AssaySchema createSchema(User user, Container container, @Nullable Container targetStudy);
 
     /**
-     * @return all the assay protocols that are in scope in the given container
+     * @return all the assay protocols that are in scope in the given container, which may include those
+     * defined in other containers
      */
     @NotNull List<ExpProtocol> getAssayProtocols(Container container);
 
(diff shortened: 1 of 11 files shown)


labkey-jeckels:
@labkey-matthewb I pushed the updates you suggested. By caching strictly at the container level, I don't need to switch to RowIds, since we won't be caching the same protocol separately for each container where it's in scope.

labkey-matthewb (approved):

labkey-tchad:
Manual test results:
- ✅ External schemas appear and disappear as expected.
- ✅ Assay definitions look good too but I ran into some confusing behavior.
  - It's a little strange that we allow you to select a subfolder for assay creation in the LKS UI, but ignore that selection and create it in the project if the SampleManagement module is enabled in the project.

labkey-tchad (approved):

PR #6852, opened 2025-07-16
Issue 53476: Improve queries dominating the execution time

labkey-jeckels:
#### Rationale
Four queries dominate the DB execution time on some deployments. They can be optimized to use indices.

#### Changes
- Add missing indices
- Store container aliases as lower case so the DB can use the index
- Join on container RowId instead of EntityId


labkey-matthewb (approved):
I'm surprised we haven't hit paths > 255 length yet.

PR #6853, opened 2025-07-16
Issue 53243: add "includeEmptyPermGroups" to security-getGroupPerms.api

labkey-nicka:
#### Rationale
This addresses [Issue 53243](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53243) by adding a new parameter `includeEmptyPermGroups` to the `security-getGroupPerms.api` which excludes groups that have no effective permissions.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2569

#### Changes
- new parameter `includeEmptyPermGroups` to the `security-getGroupPerms.api` and respect setting when computing groups. Defaults to true.
- Move security check out of groups iteration which can significantly improve performance when there are a large number of folders and/or groups.

### DIFF
--- a/core/src/org/labkey/core/security/SecurityApiActions.java
+++ b/core/src/org/labkey/core/security/SecurityApiActions.java
@@ -19,7 +19,6 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import org.apache.commons.lang3.StringUtils;
 import org.json.JSONObject;
 import org.junit.Test;
-import org.labkey.api.action.ApiJsonWriter;
 import org.labkey.api.action.ApiResponse;
 import org.labkey.api.action.ApiSimpleResponse;
 import org.labkey.api.action.ApiVersion;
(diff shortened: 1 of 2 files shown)


core/src/org/labkey/core/security/SecurityApiActions.java
```diff
@@ -164,15 +175,17 @@ protected Map<String, Object> getContainerPerms(Container container, List<Group>
             return containerPerms;
         }
 
-        protected List<Map<String, Object>> getGroupPerms(Container container, List<Group> groups)
+        protected List<Map<String, Object>> getGroupPerms(Container container, List<Group> groups, boolean includeEmptyPermGroups)
```
labkey-susanh: Add a `@NotNull` annotation on `groups`.

labkey-susanh (approved):

core/src/org/labkey/core/security/SecurityApiActions.java
labkey-nicka: Done.

PR #6855, opened 2025-07-17
Issue 53482: Folder import fails for a sample type / data class that has a field name that is long and has special characters

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53482

#### Changes
- Move call to Lsid.fixupPropertyURI after we replace substitutions in propertyURI

#### Tasks 📍
- [x] Manual Testing/Verify Fix (@cnathe opened the issue) @XingY 
- [x] Needs Automation


labkey-jeckels (approved):
I haven't tested this out but this approach looks good to me.

PR #6858, opened 2025-07-18
Issue 52886: SourcesAuditEvent update event newRecordMap diff missing value for an updated field with a long name

labkey-susanh:
#### Rationale
Issue [52886](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52886): For data classes, the `_selectRow` method had been doing a query against the database table instead of the user schema table. For long field names (at least), the db field names and query schema field names do not match. This results in the diff check for auditing to not see a difference for such fields. The "fix" for Issue [52504](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52504) was also somewhat misguided because it placed the lookup container filter too globally. Instead we want to set the proper container filter on the lookup columns themselves.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1530

#### Changes
- Update `AuditHandler.getOldAndNewRecordForMerge` to not do detailed audit logging for multivalued fields that updated values
- When copying attributes for a property descriptor, assure the proper container filter is used for lookup columns
- Don't use the lookup container filter for the import QueryDef, use the container filter appropriate for the application (if applicable)
- Update `ExpDataClassDataTableimpl._selectRow` method to use the user schema table not the database table
- Set the container filter on the import schema as appropriate for the applications (Issue 52504 fix revised)


experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -805,6 +805,29 @@ public boolean next() throws BatchValidationException
...
+     */
+    public static void setContainerFilterForImport(QueryDefinition qDef, Container container, User user)
+    {
+        if (container.isProductFoldersEnabled())
+        {
+            ContainerFilter cf;
+
+            if (container.isProject())
+                cf = new ContainerFilter.AllInProjectPlusShared(container, user);
+            else
```
labkey-nicka, line 825: Seems like we still want to respect the experimental flag even when importing into a subfolder. I propose:

```java
public static void setContainerFilterForImport(QueryDefinition qDef, Container container, User user)
{
    if (container.isProductFoldersEnabled())
    {
        ContainerFilter cf;

        if (container.isProject())
            cf = new ContainerFilter.AllInProjectPlusShared(container, user);
        else
            cf = QueryService.get().getContainerFilterForLookups(container, user);

        if (cf != null)
            qDef.setContainerFilter(cf);
    }
}
```

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -1284,32 +1273,34 @@ protected Map<String, Object> _select(Container container, Object[] keys) throws
...
-            SQLFragment sql = new SQLFragment()
-                    .append("SELECT t.*, d.RowId, d.Name, d.ClassId, d.Container, d.Description, d.CreatedBy, d.Created, d.ModifiedBy, d.Modified")
-                    .append(" FROM ").append(d, "d")
-                    .append(" LEFT OUTER JOIN ").append(t, "t")
-                    .append(" ON d.lsid = t.lsid WHERE ");
-
+            // Issue 52886: Use queryTable here, not raw database table, so the rows are from the user schema with names
+            // as expected to match row inserts and other querySchema data
+            SimpleFilter filter = new SimpleFilter();
             if (null != rowid)
```
labkey-nicka: nit: Rename `rowid` to `rowId`.

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -1284,32 +1273,34 @@ protected Map<String, Object> _select(Container container, Object[] keys) throws
...
-                sql.append("d.rowid=?").add(rowid);
+                filter.addCondition(FieldKey.fromParts("rowId"), rowid);
```
labkey-nicka: Consider using the column enumeration instead:

```java
SimpleFilter filter = new SimpleFilter();
if (null != rowid)
    filter.addCondition(Column.RowId.fieldKey(), rowid);
else if (null != lsid)
    filter.addCondition(Column.LSID.fieldKey(), lsid);
else
    filter.addCondition(Column.ClassId.fieldKey(), classId)
            .addCondition(Column.Name.fieldKey(), name);
if (!allowCrossContainer)
    filter.addCondition(Column.Folder.fieldKey(), container.getEntityId());
```

experiment/src/org/labkey/experiment/ExpDataIterators.java
```diff
@@ -805,6 +805,29 @@ public boolean next() throws BatchValidationException
...
+    /**
+     * Issue 52504 (sort of): Chooses a container filter that is appropriate for import, merge or update actions in the face of product folders.
+     * Note that this is slightly different from our treatment of lookups:
+     *   - when in a project, we allow import or update to all subfolders,
+     *   - when in a folder, we only allow references to data up the folder tree
+     * @param qDef The QueryDefinition in use for the import action
+     * @param container The container that is the target of the import or update
+     * @param user The user doing the action
+     */
+    public static void setContainerFilterForImport(QueryDefinition qDef, Container container, User user)
```
labkey-nicka, line 817: I can understand why this is in `ExpDataIterators` but it seems like something we'd want to move next to the other container filter operations on `QueryService` at some point.

experiment/src/org/labkey/experiment/ExpDataIterators.java
labkey-susanh: We're setting the container filter for lookups on the lookup columns themselves so when we remap values during import we will look in the right places. Setting the lookup filter on query table doesn't seem quite right (and caused problems with the previous iteration on this fix because the `_selectRow` method is not expecting the table to be filtered in this way), but maybe I'm not thinking of the right use cases.

labkey-nicka: I'm curious what kind of problems this caused. What I'm proposing isn't much of a logical shift, other than the fact that when the experimental flag is true, then it will also use the `AllInProjectPlusShared` filter. Otherwise, it will use `CurrentPlusProjectAndShared`.

I'm guessing the problems you were seeing is when you enabled the flag?

labkey-nicka: After discussing with @labkey-susanh, I was incorrect in my assumptions about what exactly this was scoping. This filter is on the table -- not on the lookups. As an example of why we do not want to use `getContainerFilterForLookups()` here, we do not want to be able to import data/samples in a sibling subfolder from a subfolder. Should only be able to import into the subfolder and the project from the subfolder regardless of the experimental flag setting for less restrictive lookups.

labkey-nicka (approved):

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
labkey-susanh: Done (locally).

PR #6859, opened 2025-07-18
TestDataGenerator randomFieldName and randomDomainName usage of ValidateDomainAndFieldNamesAction

cnathe:
#### Rationale
We've been slowly handling various cases on the test code side to account for randomly generated domain and field names that end up being invalid (because of invalid character sequences, length, etc.). This PR changes that approach so that we instead call a server API (ValidateDomainAndFieldNamesAction) to do those checks on the generated name and regenerate the name if it is invalid.

#### Related Pull Requests
- https://github.com/LabKey/testAutomation/pull/2571
- https://github.com/LabKey/platform/pull/6859
- https://github.com/LabKey/limsModules/pull/1531
- https://github.com/LabKey/premiumModules/pull/313

#### Changes
- PropertyController.ValidateDomainAndFieldNameAction and usage in TestDataGenerator.randomFieldName and randomDomainName
- Add DomainKind.validateDomainName to consolidate some of the duplicate checks in various domainKinds
- Update usages of randomFieldName to pass DomainKind


experiment/src/client/test/integration/utils.ts
```diff
@@ -408,7 +408,7 @@ export async function checkDomainName(server: IntegrationTestServer, domainType:
     let dataTypeRowId = 0;
     if (domainType !== 'SampleSet')
         dataTypeRowId = await getDataClassRowIdByName(server, domainName, folderOptions);
-    const requireMsg = `${domainType} name must not be blank.`
+    const requireMsg = domainType == 'SampleSet' ? 'Sample Type name is required.' : `${domainType} name must not be blank.`;
```
labkey-susanh: Constant?

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
```diff
@@ -501,6 +501,75 @@ public Object execute(DomainApiForm form, BindException errors)
...
+
+            boolean checkFieldNames = !domainDesign.getFields().isEmpty();
+            boolean checkDomainName = domainDesign.getName() != null;
+
+            ApiSimpleResponse resp = new ApiSimpleResponse();
+            ValidationException results = new ValidationException();
+            if (checkFieldNames)
+            {
+                results = DomainUtil.validateProperties(null, domainDesign, kind, null, getUser());
+                if (!results.hasErrors())
```
labkey-susanh: Seems like it would be useful to proceed with the further validation even if there are errors from the first check. Or does that cause problems?

query/test/src/org/labkey/test/tests/query/CrossFolderDataClassTest.java
```diff
@@ -54,15 +56,16 @@ private void doSetup()
      * Issue 45664: addresses the problem where DataClass metadata wasn't available in query when querying cross-folder
      */
     @Test
-    public void testIssue454644() throws Exception
+    public void testIssue45664() throws Exception
```
labkey-susanh, line 59: I know it's not actually from this PR, but I don't love test names like this that are not mnemonic.

labkey-susanh (approved):

experiment/src/client/test/integration/utils.ts
cnathe: done

experiment/src/org/labkey/experiment/controllers/property/PropertyController.java
cnathe: yup, good point. done

PR #6862, opened 2025-07-19
Update redirect() and eliminate HttpRedirectView

labkey-adam:
#### Rationale
Eliminate `HttpView.redirect(URLHelper)` method in favor of other variants that take an `ActionURL` or a `URLHelper` plus a boolean indicating whether redirecting to an external URL is allowed. This tightens up security and reduces surprises.

Also eliminate `HttpRedirectView` in favor of throwing `RedirectException`.

#### Tasks 📍
- [ ] Manual Testing / Verify Fix - not needed?
- [ ] Needs Automation - rely on existing automation?


query/src/org/labkey/query/reports/ReportsController.java
```diff
@@ -1027,7 +1022,7 @@ public ModelAndView getView(ReportDesignBean<?> form, BindException errors) thro
                 return new HtmlView(SPAN(cl("labkey-error"), e.getMessage(), ". Unable to create report."));
             }
 
-            if (!isPrint() && !(reportView instanceof HttpRedirectView) && DiscussionService.get() != null)
+            if (!isPrint() && DiscussionService.get() != null)
```
labkey-jeckels, line 1025: This makes me even happier to eliminate HttpRedirectView.

labkey-jeckels (approved):
I didn't manually test this. I haven't looked at the callers to inform if/how to test. Any recommendations?

labkey-adam:
> I didn't manually test this. I haven't looked at the callers to inform if/how to test. Any recommendations?

No recommendations... I'm hoping automated tests would have shown issues. The `redirect()` changes are effectively no-ops: they are local redirects both before and after. And throwing vs. return a redirect view should be equivalent.

PR #6867, opened 2025-07-23
Option to migrate SQL Server tables to PostgreSQL

labkey-adam:
#### Rationale
We want to provide administrators the ability to easily migrate a SQL Server LabKey SDMS deployment to PostgreSQL. Changes in this PR provide an option to create empty PostgreSQL schemas and migrate all module & provisioned table data into them from a SQL Server deployment.

Specification: https://docs.google.com/document/d/1oY3Mhnusa17OsC3NDj3WR5gS5YlynUP5zzc2tBht_D4/edit?tab=t.0#heading=h.4rg4h5vuek0

This is a significant step forward, but it's still a work in progress. What's missing:
- Testing with non-trivial SQL Server databases
- Testing with large amounts of data

There should be no functional changes in the normal case. The new code is invoked only when the server is started with specific command-line arguments.

#### Changes
If a PostgreSQL database is being bootstrapped and a specific argument is added to the command line, the server will:
- Bootstrap all PostgreSQL schemas without inserting any* data into them
- Flag all PostgreSQL tables that still contain rows and sequences set to a value other than one, plus a few other sanity checks
- Connect to the specified SQL Server database as an external data source and enumerate all module tables, sorted to accommodate foreign key constraints
- Copy data from every table into the just-created empty PostgreSQL tables
- Reset the PostgreSQL sequences to ensure that future keys produced by SERIAL columns don't conflict with the copied-in data
- Create all the provisioned tables based on the rows in `exp.DomainDescriptor`
- Repeat the data migration and sequence updates for the provisioned schemas/tables

*Some details:
- The root and shared containers are actually created for bootstrap purposes (many code paths rely on their presence) and then deleted before populating the tables
- The core Modules, SqlScripts, and UpgradeSteps tables are populated by the PostgreSQL bootstrap process and not cleared. The data in these SQL Server tables is not migrated.
- The `migrationDataSource` argument is used to specify the SQL Server data source to migrate. For example: `-DmigrationDataSource=ssDataSource`
- `-DemptySchemas=true` can be used to test the empty schema creation and verification without attempting an actual migration
- While not typically needed, a few modules register schema-specific `MigrationHandler` implementations to customize behavior (e.g., accommodate cyclical FKs or adjust the tables to migrate)

#### Tasks 📍
- [x] Option to bootstrap schemas without any data
- [x] Verify empty schemas and sequences
- [x] Copy data from SQL Server tables
- [x] Adjust sequences to accommodate copied data
- [x] Fix TableSorter handing for self-referencing foreign keys
- [x] Special case inserts for cyclical FKs in wiki (Pages/PageVersions), prot (Organisms/Identifiers), exp (ExperimentRun/ProtocolApplication)
- [x] Expand testing to all modules and ensure empty lookup tables throughout
- [x] Drop problematic, unused `Deleted` columns from `prot` schema
- [x] Update spec
- [x] Create provisioned tables and migrate data
- [x] Code review @labkey-jeckels 
- [x] Manual Testing @labkey-danield 📍
- Needs Automation - Not at this point. Eventually, we'll need some automated testing.


api/src/org/labkey/api/query/TableSorter.java
```diff
@@ -145,14 +152,14 @@ private static List<TableInfo> sort(String schemaName, Map<String, TableInfo> ta
...
     }
 
-    private static void depthFirstWalk(String schemaName, Map<String, TableInfo> tables, TableInfo table, Set<TableInfo> visited, LinkedList<Tuple3<TableInfo, ColumnInfo, TableInfo>> visitingPath, List<TableInfo> sorted)
+    private static void depthFirstWalk(String schemaName, Map<String, TableInfo> tables, TableInfo table, Set<TableInfo> visited, LinkedList<Tuple3<TableInfo, ColumnInfo, TableInfo>> visitingPath, List<TableInfo> sorted, boolean tolerateLoops)
     {
-        if (hasLoop(visitingPath, table))
+        if (!tolerateLoops && hasLoop(visitingPath, table))
         {
             String msg = "Loop detected in schema '" + schemaName + "':\n" + formatPath(visitingPath);
             if (anyHaveContainerColumn(visitingPath))
```
labkey-jeckels: Is this check relevant for the new migration use case?

audit/src/org/labkey/audit/AuditLogImpl.java
```diff
@@ -92,7 +93,8 @@ public static AuditLogImpl get()
 
     private AuditLogImpl()
     {
-        ContextListener.addStartupListener(this);
+        if (ModuleLoader.getInstance().shouldInsertData())
```
labkey-jeckels, line 97: Consider a comment saying what the startup listener is inserting.

core/src/org/labkey/core/login/LoginController.java
```diff
@@ -1550,7 +1550,7 @@ public ModelAndView getView(SsoRedirectForm form, BindException errors)
 
             url = configuration.getUrl(csrf, getViewContext());
 
-            return HttpView.redirect(url.getURIString());
+            return HttpView.redirect(url, true);
```
labkey-jeckels, line 1553: Is this the intended FB?

api/src/org/labkey/api/data/SqlScriptExecutor.java
```diff
@@ -79,10 +83,51 @@ public void execute()
...
+            // Strip all statements (typically inserts) between the empty-schema START and END annotations, inclusive
+            MutableBoolean skipping = new MutableBoolean(false);
+            MutableInt lineCount = new MutableInt(0);
+            sql = _sql.lines()
+                .filter(line -> {
+                    lineCount.increment();
+                    if (line.contains(START_ANNOTATION))
+                    {
+                        if (skipping.booleanValue())
+                            throw new IllegalStateException("Unexpected " + START_ANNOTATION + " at line " + lineCount.intValue());
```
labkey-jeckels, line 108: Is it clear from other adjacent logging which script has the problem?

api/src/org/labkey/api/module/DatabaseMigration.java
```diff
@@ -0,0 +1,381 @@
...
+    private static void clearSchemas()
+    {
+        TableInfo containers = CoreSchema.getInstance().getTableInfoContainers();
+        Table.delete(containers); // Now that we've bootstrapped, delete root and shared containers
+        DbScope targetScope = DbScope.getLabKeyScope();
+        new SqlExecutor(targetScope).execute("ALTER SEQUENCE core.containers_rowid_seq RESTART"); // Reset Containers sequence
+    }
+
+    // Verify that no data rows were inserted and no sequences were incremented
+    private static void verifyEmptySchemas()
```
labkey-jeckels, line 80: FWIW, it seems like it's more work to avoid inserting rows and then verify that no rows were inserted. Seems like the below code could be truncating the tables (ordered based on FKs) and then we wouldn't have to conditionalize the bootstrap code.

api/src/org/labkey/api/module/DatabaseMigration.java
```diff
@@ -0,0 +1,381 @@
...
+            }
+            else
+            {
+                MigrationHandler handler = getHandler(targetSchema);
+                handler.beforeSchema(targetSchema);
+
+                List<TableInfo> sourceTables = getTables(sourceSchema);
+                List<TableInfo> targetTables = getTables(targetSchema);
+
+                if (sourceTables.size() != targetTables.size())
```
labkey-jeckels: Perhaps unlikely to happen in the real world, but there could be differences even if there are the same number of tables. Perhaps that would fail when we try to select or insert.

api/src/org/labkey/api/module/DatabaseMigration.java
```diff
@@ -0,0 +1,381 @@
...
+                        try (Stream<Map<String, Object>> mapStream = sourceSelector.uncachedMapStream(); Connection conn = targetScope.getConnection())
+                        {
+                            Collection<ColumnInfo> sourceColumns = sourceSelector.getSelectedColumns();
+                            // Map the selected source columns to the target columns so we get the right order and casing for INSERT, etc.
+                            Collection<ColumnInfo> targetColumns = sourceColumns.stream()
+                                .map(sourceCol -> targetTable.getColumn(sourceCol.getName()))
+                                .toList();
+                            String q = StringUtils.join(Collections.nCopies(sourceColumns.size(), "?"), ", ");
+                            SQLFragment sql = new SQLFragment("INSERT INTO ")
+                                .append(targetTable.getSelectName())
```
labkey-jeckels: Should this use `appendIdentifier()`?

api/src/org/labkey/api/module/DatabaseMigration.java
```diff
@@ -0,0 +1,381 @@
...
+                    Set<String> targetTableNames = targetTables.stream().map(TableInfo::getName).collect(LabKeyCollectors.toCaseInsensitiveHashSet());
+                    Set<String> sourceTableNamesCopy = new CaseInsensitiveHashSet(sourceTableNames);
+                    sourceTableNames.removeAll(targetTableNames);
+                    targetTableNames.removeAll(sourceTableNamesCopy);
+                    LOG.warn("Table differences: {} and {}", sourceTableNames, targetTableNames);
+                }
+
+                Map<String, Sequence> schemaSequenceMap = sequenceMap.getOrDefault(sourceSchema.getName(), Map.of());
+
+                for (TableInfo targetTable : tableProducer.apply(targetSchema, handler))
```
labkey-jeckels, line 228: Not needed yet, but this is a good candidate for multi-threading on a per-table basis.

core/resources/schemas/dbscripts/postgresql/core-0.000-24.000.sql
```diff
@@ -46,19 +46,17 @@ CREATE TABLE core.Logins
 
 CREATE TABLE core.Principals
 (
-  UserId SERIAL,                    -- user or group
+  UserId INT GENERATED BY DEFAULT AS IDENTITY ( MINVALUE 1000 ),  -- user or group
```
labkey-jeckels: Should we do this in an upgrade script too or instead? Seems like this column will end up in different states for bootstraps vs upgrades.

experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
```diff
@@ -148,18 +148,31 @@ private String _create(DbScope scope, DomainKind<?> kind, Domain domain)
...
             String tableName = dd.getStorageTableName();
-            if (null != tableName)
+
+            if (useProvidedStorageName)
             {
-                transaction.commit();
-                return tableName;
+                if (null == tableName)
+                {
+                    log.warn("Storage table name was null: {} {}", domain.getTypeId(), domain.getTypeURI());
```
labkey-jeckels: Consider throwing an exception here. I know other code in this method returns null but it looks like some callers won't notice.

labkey-jeckels (approved):

api/src/org/labkey/api/query/TableSorter.java
labkey-adam: Yes, this keeps the sorter from throwing when attempting to sort the wiki and prot tables

core/src/org/labkey/core/login/LoginController.java
labkey-adam: Not directly relevant, of course, but yes it was intentional

core/resources/schemas/dbscripts/postgresql/core-0.000-24.000.sql
labkey-adam: Turns out it's a bit annoying to convert an existing SERIAL into an IDENTITY, so I think I'll just restore this one back to SERIAL. I was trying to avoid explicitly setting a value on this sequence (one less sequence flagged in the verification step) and thought it would be nice to make the starting value consistent between PG and SS, but I don't care enough about either of those. I agree we want to be consistent between bootstrap and upgrade, hence back to the old way.

audit/src/org/labkey/audit/AuditLogImpl.java
labkey-adam: Okay

api/src/org/labkey/api/data/SqlScriptExecutor.java
labkey-adam: Yes
```
INFO  SqlScriptManager         2025-07-31T11:40:35,453           Module Upgrade : Starting to run script: tcrdb-15.31-15.32.sql
FATAL ModuleLoader             2025-07-31T11:40:35,453           Module Upgrade : Startup failure, terminating
org.labkey.api.data.SqlScriptRunner$SqlScriptException: tcrdb-15.31-15.32.sql : Unexpected @SkipOnEmptySchemasBegin at line 18
	at org.labkey.api.data.SqlScriptManager.runScript(SqlScriptManager.java:213) ~[api-25.8-SNAPSHOT.jar:?]
	at org.labkey.api.data.SqlScriptRunner.runScripts(SqlScriptRunner.java:83) ~[api-25.8-SNAPSHOT.jar:?]
```

api/src/org/labkey/api/module/DatabaseMigration.java
labkey-adam: Well, it's not more work for the poor database. But sure, enumerating all tables and truncating would be equivalent. Sequences would be initialized in that case, but we're resetting them anyway. The skipping approach gave me the chance to review each case to ensure it seemed appropriate (i.e., that the table would get populated either by the incoming SQL Server data or by the next server start). The verify step & logging is really just for development purposes... to verify assumptions and point out cases that might need to be investigated. At the end of the day, IMO, it's a toss up and the skip insert approach is the one I went with.

api/src/org/labkey/api/module/DatabaseMigration.java
labkey-adam: Probably...

api/src/org/labkey/api/module/DatabaseMigration.java
labkey-adam: Yes, though we'd have to be careful about foreign keys...

api/src/org/labkey/api/module/DatabaseMigration.java
labkey-adam: Updated the check to look for any differences

experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
labkey-adam: Okay

PR #6856, opened 2025-07-17
Issue 52657: sample name case sensitivity query performance

XingY:
#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6856
- https://github.com/LabKey/premiumModules/pull/312
- https://github.com/LabKey/platform/pull/6868
- https://github.com/LabKey/platform/pull/6820

#### Changes
-

PR #6868, opened 2025-07-23
Revert support for case insensitive sample/data names

XingY:
#### Rationale
There was a set of PRs that got merged to disallow insert/update of case insensitive sample/data names. However, this has resulted in performance degradation of sample imports. Furthermore, this is more of a half way solution that treats name as case insensitive at insert/rename time, but requires exact match at update time. For those reasons, we have decided to revert the previous PRs for now. An alternative database level collation approach will be explored at a future time.

Minor: there was a client side change to show exact case as user provided instead of lower-case when generating duplicate error on the editable grid. This improvement is not included in the revert. 

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6868
- https://github.com/LabKey/limsModules/pull/1542
- https://github.com/LabKey/testAutomation/pull/2576

#### Original Pull Requests that added support for case insensitive names
- https://github.com/LabKey/labkey-ui-components/pull/1828
- https://github.com/LabKey/platform/pull/6820
- https://github.com/LabKey/limsModules/pull/1510
- https://github.com/LabKey/testAutomation/pull/2541

#### Pull Requests attempts to improve import performance (never merged)
- https://github.com/LabKey/platform/pull/6856
- https://github.com/LabKey/premiumModules/pull/312

#### Changes
-

### DIFF
--- a/experiment/src/client/test/integration/DataClassCrud.ispec.ts
+++ b/experiment/src/client/test/integration/DataClassCrud.ispec.ts
@@ -233,7 +233,7 @@ describe('Import with update / merge', () => {
    it ("Issue 52922: Blank sample id in the file are getting ignored in update from file", async () => {
        const BLANK_KEY_UPDATE_ERROR_NO_EXPRESSION = 'Missing value for required property: Name';
        const BLANK_KEY_UPDATE_ERROR_WITH_EXPRESSION = 'Name value not provided on row ';
-       const BOGUS_KEY_UPDATE_ERROR = 'Data not found: ';
+       const BOGUS_KEY_UPDATE_ERROR = 'Data not found for ';
        const DUPLICATE_KEY_ERROR = 'duplicate key value';
 
        const dataType = "NoExpressionNameRequired52922";
--- a/experiment/src/client/test/integration/DataClassCrud.ispec.ts
+++ b/experiment/src/client/test/integration/DataClassCrud.ispec.ts
@@ -233,7 +233,7 @@ describe('Import with update / merge', () => {
    it ("Issue 52922: Blank sample id in the file are getting ignored in update from file", async () => {
        const BLANK_KEY_UPDATE_ERROR_NO_EXPRESSION = 'Missing value for required property: Name';
        const BLANK_KEY_UPDATE_ERROR_WITH_EXPRESSION = 'Name value not provided on row ';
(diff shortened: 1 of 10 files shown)


experiment/src/client/test/integration/DataClassCrud.ispec.ts
```diff
@@ -357,11 +357,11 @@ describe('Duplicate IDs', () => {
             }]
         }, { ...topFolderOptions, ...editorUserOptions }).expect((result) => {
             errorResp = JSON.parse(result.text);
-            expect(errorResp.exception.indexOf('already exists') > -1).toBeTruthy();
+            expect(errorResp.exception.indexOf('duplicate key') > -1).toBeTruthy();
```
labkey-susanh, line 360: Looks like a change to `BOGUS_KEY_UPDATE_ERROR` (above) did not get reverted.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1719,7 +1718,7 @@ private Map<Integer, Pair<Integer, String>> getSampleAliquotCounts(Collection<In
                 .append(")-1 AS CreatedAliquotCount FROM exp.material AS m WHERE m.rowid\s");
         dialect.appendInClauseSql(sql, sampleIds);
 
-        Map<Integer, Pair<Integer, String>> sampleAliquotCounts = new TreeMap<>(); // Order sample by rowId to reduce probability of deadlock with search indexer
+        Map<Integer, Pair<Integer, String>> sampleAliquotCounts = new HashMap<>();
```
labkey-susanh: This change seems like it was rather unrelated to the issue fix. I'm ok with reverting this too, but just wanted to check that it is intentional

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
XingY: Good point. Changes added back.

experiment/src/client/test/integration/DataClassCrud.ispec.ts
XingY: Yep, fixed. Thanks for catching it.

PR #6869, opened 2025-07-23
Add audit logging for file fields and renaming of files

labkey-susanh:
When files are uploaded to store on the file system, we need to avoid name collisions and thus must sometimes rename files. We do this by appending counter suffixes to the file name base. For better accountability, we want to now add audit records for these file uploads and renames. This includes:
- Sample file fields
- Assay file fields
- Assay result files 
- Dataset file fields

We add the original file name as well as a transactionId to the `FileSystemAuditEvent` in order to record the given data and to provide a way to link together the file events with the events from the appropriate domain. 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1836
- https://github.com/LabKey/labkey-ui-premium/pull/795
- https://github.com/LabKey/limsModules/pull/1543

#### Changes
- Move `findUniqueFileName` and relatives from `AssayFileWriter` to `FileUtil`
- Update `FileSytemAuditEvent` to include transactionId and the originally provided name
- Add audit logs for file field uploads, assay result file uploads, moving of assay files
- Add transactionId to more audit logs


api/src/org/labkey/api/assay/AbstractTempDirDataCollector.java
```diff
@@ -229,10 +212,15 @@ public Map<String, FileLike> uploadComplete(ContextType context, @Nullable ExpRu
...
+                    FileSystemAuditProvider.FileSystemAuditEvent event = new FileSystemAuditProvider.FileSystemAuditEvent(context.getContainer(),  "Primary file provided for assay import");
+                    event.setProvidedFileName(tempDirFile.getName());
+                    event.setFile(assayDirFile.getName());
+                    event.setDirectory(assayDirFile.getParent().toURI().getPath());
+                    AuditLogService.get().addEvent(context.getUser(), event);
                 }
                 else
                 {
-                    LOG.warn("Unable to resolve import/upload file location for file: " + tempDirFile);
+                    LOG.warn("Unable to resolve import/upload file location for file: {}", tempDirFile);
```
cnathe, line 223: was this addition of `{}` intentional? was it supposed to be some replacement syntax?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -2139,6 +2132,11 @@ private Map<Integer, List<FileFieldRenameData>> updateSampleFilePaths(ExpSampleT
 
             // TODO, support batch fireFileMoveEvent to avoid excessive FileLinkFileListener.hardTableFileLinkColumns calls
             fileService.fireFileMoveEvent(sourceFile, ref.targetFile, user, targetContainer);
+            FileSystemAuditProvider.FileSystemAuditEvent event = new FileSystemAuditProvider.FileSystemAuditEvent(ref.sourceContainer, "File moved from " + ref.sourceContainer.getPath() + " to " + targetContainer.getPath() + ".");
+            event.setProvidedFileName(sourceFile.getName());
+            event.setFile(ref.targetFile.getName());
+            event.setDirectory(ref.targetFile.getParent());
+            AuditLogService.get().addEvent(user, event);
```
cnathe: There are quite a few usages added in this PR that set these same set or file/dir properties of the event. Thoughts on adding a new constructor for FileSystemAuditEvent or some sort of utility so that a usage wouldn't inadvertently miss one of these lines (i.e. pass in the source file and the target file and it coudl set these 3 things (in most cases).

api/src/org/labkey/api/assay/AssayFileWriter.java
```diff
@@ -305,8 +252,16 @@ public Map<String, FileLike> savePostedFiles(ContextType context, Set<String> pa
...
                             FileLike dir = getFileTargetDir(context);
-                            FileLike file = findUniqueFileName(fileName, dir);
+                            FileLike file = FileUtil.findUniqueFileName(fileName, dir);
                             multipartFile.transferTo(toFileForWrite(file));
+                            if (!dir.toURI().getPath().contains(TEMP_DIR_NAME))
+                            {
+                                FileSystemAuditProvider.FileSystemAuditEvent event = new FileSystemAuditProvider.FileSystemAuditEvent(context.getContainer(), allowMultiple ? "File field provided for assay import" : "Primary file provided for assay import");
+                                event.setProvidedFileName(fileName);
+                                event.setFile(file.getName());
+                                event.setDirectory(dir.getParent().toURI().getPath());
```
cnathe: should this just be `dir` instead of `dir.getParent()` (see comment in test plan related to this)?

api/src/org/labkey/api/audit/SampleTimelineAuditEvent.java
```diff
@@ -92,11 +92,16 @@ public static SampleTimelineEventType getTypeFromName(@Nullable String name)
...
+    public SampleTimelineAuditEvent()
+    {
+        super();
+        setTransactionId(TransactionAuditProvider.getCurrentTransactionAuditId());
+    }
 
     public SampleTimelineAuditEvent(Container container, String comment)
     {
         super(EVENT_TYPE, container, comment);
+        setTransactionId(TransactionAuditProvider.getCurrentTransactionAuditId());
```
cnathe, line 104: I see a bunch of these `setTransactionId()` calls in this PR. Is there any reason not to move this up to the base class? Maybe we don't want this for all audit event types, but just checking.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -1004,28 +1005,41 @@ public static Object saveFile(User user, Container container, String name, Objec
...
-                file = AssayFileWriter.findUniqueFileName(saf.getFilename(), dir);
-                file = checkFileUnderRoot(container, file);
+                file = FileUtil.findUniqueFileName(saf.getFilename(), dir);
+                checkFileUnderRoot(container, file);
                 saf.saveTo(file);
+                event.setComment(String.format(auditMessageFormat, saf.getFilename(), name, container.getPath()));
+                event.setProvidedFileName(saf.getFilename());
+                event.setFile(file.getName());
+                event.setFieldName(name);
+                event.setDirectory(file.getParent().toURI().getPath());
```
cnathe: minor: looks like these three lines are duplicated in the if and the elseif blocks. Could they be moved outside / below. or is there an else case here that wouldn't want this behavior.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -10062,7 +10062,8 @@ public boolean moveFileLinkFile(File sourceFile, File targetFile, Container sour
         }
 
         event.setDirectory(sourceFile.getParentFile().getAbsolutePath());
```
cnathe, line 10065: should `setDirectory` use `targetFile` as well instead of `sourceFile`?

cnathe (changes requested):

api/src/org/labkey/api/assay/AbstractTempDirDataCollector.java
labkey-nicka: Loggers support this syntax for replacement.

labkey-susanh: Yes, intentional.

api/src/org/labkey/api/assay/AssayFileWriter.java
labkey-susanh: Yes. Fixed.

api/src/org/labkey/api/audit/SampleTimelineAuditEvent.java
labkey-susanh: We haven't implemented for all the audit event types, so I think we should leave it this way for now since I don't want to give the false impression that the value is meant to be populated when we haven't wired it up. It would probably be a good idea to do this more globally, but I'm going to leave that for a different PR.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: Yes. Thanks for catching that.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
labkey-susanh: Sure. Done.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: I'm not super excited by having a constructor with many more arguments, especially since sometimes we're dealing with `File`s and sometimes `FileLike` objects and sometimes we don't get the directory directly from the provided File. Happy to discuss, though if you feel strongly about it.

cnathe: Nope. I don't feel strongly. I didn't see that some were File and some FileLike...so I agree that makes it less compelling.

cnathe (approved):

PR #6871, opened 2025-07-23
don't throw if we can't resolve the participant

labkey-klum:
#### Rationale
Found via this crawler [exception](https://teamcity.labkey.org/viewLog.html?buildId=3592883&buildTypeId=LabkeyTrunk_DailyPostgres_2&fromSakuraUI=true#testNameId1662430247021143670).

If we are unable to resolve the participant, don't throw, just render an error.


labkey-tchad (approved):
I though `NotFoundException` would automatically generate an error page that plays nicely with the crawler?

labkey-klum:
> I though `NotFoundException` would automatically generate an error page that plays nicely with the crawler?

It looks like we will ignore 404s if they come from the initial set of links. Does that align with your expectations?

PR #6872, opened 2025-07-23
Track search work by container; purge on delete

labkey-jeckels:
#### Rationale
Instead of waiting for the indexer to drain its queue before deleting a container, we can purge that work from the queue.

#### Changes
- Track search work based on the container for the data
- Purge work for the container being deleted

#### Tasks 📍
- [ ] Manual Testing
- [ ] Needs Automation
- [ ] Verify Fix

### DIFF
--- a/api/src/org/labkey/api/search/SearchService.java
+++ b/api/src/org/labkey/api/search/SearchService.java
@@ -65,6 +65,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 
 public interface SearchService extends SearchMXBean
 {
@@ -96,9 +97,12 @@ public interface SearchService extends SearchMXBean
     // marker value for documents with indexing errors
     Date failDate = new Timestamp(DateUtil.parseISODateTime("1899-12-30"));
 
-    static @Nullable SearchService get()
+    static final SearchService NO_OP = new NoopSearchService();
+
+    static @NotNull SearchService get()
     {
-        return ServiceRegistry.get().getService(SearchService.class);
+        SearchService service = ServiceRegistry.get().getService(SearchService.class);
+        return service == null ? NO_OP : service;
     }
 
     static void setInstance(SearchService impl)
@@ -111,7 +115,10 @@ public interface SearchService extends SearchMXBean
      */
     void reindexContainerFiles(Container c);
 
-    void purgeForContainer(Container container);
+    /**
+     * Removes all non-delete work from the queue associated with the target container
+     */
+    void purgeForContainer(@NotNull Container container);
 
     /**
      * Puts work in the indexer queue at the specified priority and waits up to the timeout for it to complete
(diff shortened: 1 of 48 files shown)


search/src/org/labkey/search/model/AbstractSearchService.java
```diff
@@ -616,11 +627,24 @@ public boolean _eq(URLHelper a, URLHelper b)
...
+    public static @NotNull SearchService get()
+    {
+        return Objects.requireNonNull(ServiceRegistry.get().getService(SearchService.class));
+    }
+
     @Override
-    public boolean drainQueue(PRIORITY priority, long timeout, TimeUnit unit) throws InterruptedException
+    public void purgeForContainer(Container container)
+    {
+        _runQueue.removeIf(i -> container.getEntityId().equals(i._containerId));
```
labkey-matthewb: Shouldn't the task owning this item be notified?  E.g. with TaskListener.indexError() or new method TaskListener.cancel().  Will the task ever know it is done?

api/src/org/labkey/api/search/SearchService.java
```diff
@@ -114,25 +111,30 @@ static void setInstance(SearchService impl)
...
-        group,      // one container
-        item,       // one page/attachment
-        delete
+        /** Only used by the no-op task to detect when there is no other work in the queue */
+        idle,
+
+        /** Lower priority indexing work. Intended for work placed in the queue as part of a call to DocumentProvider.enumerateDocuments() */
+        crawl,
+        /** Intended for content that needed indexing or reindexing because it was modified or created */
+        modified,
```
labkey-matthewb, line 142: I'm not convinced just having Runnables < Item is sufficient?  Runnables can create Runnables, and you really want those to be run at higher pri.   I suppose that there is lower risk of this exploding the queue (compared to the Item explosion), but it still does sort of lead to doing a "breadth-first" expansion of the tasks instead of "depth-first".

labkey-matthewb: Maybe that could be fixed by having "child" runnables be added to the front of the queue, effectively taking the place of the "parent" (expand in place).

labkey-jeckels: I can revert the change if it's too simplistic (I intentionally did that work as a separate commit). My overall takeaway from this work is that we were wildly inconsistent with prioritization and that other people (including myself) didn't understand how the priorities were supposed to be used.

The one place that I found where Runnables beget Runnables is ExperimentServiceImpl.indexMaterials() and its data counterpart. There, a `bulk` Runnable creates `item` Resources and another `bulk` Runnable. I think that's consistent with the flattened set of priorities that sort Resources first. Have I gotten myself confused?

search/src/org/labkey/search/model/AbstractSearchService.java
labkey-jeckels: Good point. Added in my most recent commit.

labkey-jeckels:
Replaced by https://github.com/LabKey/platform/pull/6897

labkey-jeckels:
Replaced by https://github.com/LabKey/platform/pull/6897

PR #6873, opened 2025-07-24
Improve container caching

labkey-jeckels:
#### Rationale
We end up caching multiple copies of containers, adding to memory footprints, as well as commingling keys (EntityId vs Path)

#### Changes
- Eliminate `CACHE_ALL_CHILDREN` in favor of looking up children by `EntityId`
- Cache separately by `Path` and `EntityId`, and use their data types as keys

#### Tasks 📍
- [x] Ideas for manual testing: Check the cache's page
- [x] Manual Testing @labkey-danield 
- [ ] Sanity check on nightly or like.

### DIFF
--- a/api/src/org/labkey/api/data/ContainerManager.java
+++ b/api/src/org/labkey/api/data/ContainerManager.java
@@ -1125,7 +1125,7 @@ public class ContainerManager
         //if the input string is not a GUID, just return null,
         //so that we don't get a SQLException when the database
         //tries to convert it to a unique identifier.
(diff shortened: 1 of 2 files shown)


labkey-adam (approved):

labkey-danield:
I ran some automation, including some of the SampleManager perf testing and the caches view looked ok.
Here is the cache view before the testing:
Before Chaches
Before Transaction Caches

And here it is after the testing:
After Caches
After Transaction Caches

PR #6875, opened 2025-07-24
Adding metrics for checking case-insensitive sample/source names

XingY:
#### Rationale
Adding metrics for case-insensitive sample/source names

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6868

#### Changes
- 

#### Tasks 📍
 - modules.Experiment.duplicateCaseInsensitiveSampleNameCount
 - modules.Experiment.duplicateCaseInsensitiveDataNameCount
 - check sql server and postgres
- [ ] Verify Fix


labkey-susanh (approved):
Verified counts on Postgres. Should be made clear that this does not report the number of samples or data objects involved but the number of duplicate names. This will help us to know who would be affected by a switch to full case-insensitivity but we may later want to provide a report of the names to help users do proper renaming.

For SQL Server,  I have verified that the computation of the metric does not produce an error. Since this database is already case-insensitive by default, my counts are 0.

PR #6878, opened 2025-07-28
give the mouse group checkboxes a moment to become interactive

labkey-chrisj:
#### Rationale
Recently, [StudyPHIExportTest.testSteps](https://teamcity.labkey.org/viewLog.html?buildId=3597644&buildTypeId=LabKey_257Release_Community_DailySqlServer&fromSakuraUI=true#testNameId-2177001183088745253) failed because the mouse-group selection checkboxes weren't ready for interaction when the ext4Helper tried to un-check them during the test. I was able to repro locally, adding a brief wait to give those checkboxes time seems to have helped.

#### Related Pull Requests
n/a

#### Changes
slow down the test a bit


labkey-tchad (approved):
If you want to try a more broad fix, a good first step would be to have `org.labkey.test.components.ext4.Checkbox` override `isEnabled`. Ext has a non-standard way of indicating that elements are disabled (It adds something like `x4-disabled` to the class attribute instead of using the `disabled` attribute).
Then Ext4Helper could wait for it to be enabled.

PR #6886, opened 2025-07-29
Drop abandoned audit table

labkey-adam:
#### Rationale
It's not used


labkey-susanh (approved):
Thanks for the clean up!

PR #6891, opened 2025-07-30
Better error message when basic auth is rejected due to 2FA

labkey-adam:
#### Rationale
Previously, basic auth was rejected with a generic message when 2FA is enabled, which could lead to confusion. Now we'll throw a more useful message.

### DIFF
--- a/api/src/org/labkey/api/security/AuthenticationManager.java
+++ b/api/src/org/labkey/api/security/AuthenticationManager.java
@@ -1237,16 +1237,25 @@ public class AuthenticationManager
 
     // Throws UnauthorizedException if credentials are incorrect, password is expired, password is not complex enough,
     // user doesn't exist, user is inactive, or secondary auth is enabled and an API key hasn't been used.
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
Can we add a test for this? Something in `TOTPAuthenticationTest`? Seems like a `Connection` using `BasisAuthCredentialsProvider` that then makes a request would likely be sufficient.

labkey-adam:
> Can we add a test for this? Something in `TOTPAuthenticationTest`? Seems like a `Connection` using `BasisAuthCredentialsProvider` that then makes a request would likely be sufficient.

Test was added in the `premiumModules` FB: https://github.com/LabKey/premiumModules/pull/323

labkey-jeckels (approved):

labkey-jeckels:
I didn't manually test but the new automated test coverage makes me confident in the change.

labkey-adam:
> I didn't manually test but the new automated test coverage makes me confident in the change.

I tested this via our PostgreSQL protocol implementation, since that's where I first noticed the misleading message

PR #6894, opened 2025-07-31
Issue 53498: Sample Manager: Attachment field is lost after export/import round trip

XingY:
#### Rationale
This PR skip File converter for Attachment columns for import and allow string values to be rejected/validated during AttachmentDataIterator processing. The only time AttachmentDataIterator would accept a string attachment currently is during list archive import, where the attachment parents are set. There will be a follow up PR that addresses QUS of attachment. 

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6894
- https://github.com/LabKey/limsModules/pull/1563
- https://github.com/LabKey/testAutomation/pull/2597

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/reader/DataLoader.java
+++ b/api/src/org/labkey/api/reader/DataLoader.java
@@ -43,6 +43,7 @@ import org.labkey.api.dataiterator.MapDataIterator;
 import org.labkey.api.dataiterator.ScrollableDataIterator;
 import org.labkey.api.exp.MvColumn;
 import org.labkey.api.exp.MvFieldWrapper;
+import org.labkey.api.exp.PropertyType;
 import org.labkey.api.iterator.CloseableFilteredIterator;
 import org.labkey.api.iterator.CloseableIterator;
 import org.labkey.api.query.BatchValidationException;
@@ -327,7 +328,7 @@ public abstract class DataLoader implements Iterable<Map<String, Object>>, Loade
             for (int f = 0; f < nCols; f++)
             {
                 List<Class> classesToTest = new ArrayList<>(Arrays.asList(CONVERT_CLASSES));
-                Class knownColumnClass = null;
+                Class knownColumnClass;
 
                 int classIndex = -1;
                 //NOTE: this means we have a header row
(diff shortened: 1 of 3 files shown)


labkey-chrisj:
Manual testing notes

- Any attempt to update-from-file to an attachment field gets this helpful error: 
image

- This is true of update from excel, csv, tsv
- I tried the same scenario in Biologics and got the same results there
- merge is the same way

api/src/org/labkey/api/reader/DataLoader.java
```diff
@@ -337,27 +338,21 @@ private void inferColumnInfo(@NotNull Map<String, String> renamedColumns) throws
...
+                        ColumnInfo knownColumn = getKnownColumn(name, renamedColumns);
+
+                        if (knownColumn != null)
                         {
-                            knownColumnClass = _columnInfoMap.get(renamedColumns.get(name)).getJavaClass();
+                            knownColumnClass = knownColumn.getJavaClass();
                             classesToTest.add(0, knownColumnClass);
+
+                            // Issue 49830: if we know the column class is File based on the columnInfoMap, use it instead of trying to infer the class based on the data
+                            if (setFileColDescriptor(colDescs[f], knownColumn))
```
labkey-susanh, line 350: Seems like this could come before the setting of `knownColumnClass`.

api/src/org/labkey/api/reader/DataLoader.java
```diff
@@ -468,11 +464,45 @@ else if (renamedColumns.containsKey(name) && _columnInfoMap.containsKey(renamedC
                 ExceptionUtil.decorateException(e, ExceptionUtil.ExceptionInfo.SkipMothershipLogging, "true", true);
                 throw e;
             }
+
+            // use File converter for known file fields even if inferTypes = false. If inferTypes, this is already done.
+            if (!getInferTypes())
```
labkey-susanh, line 469: Seems like you could leave off the if statement here for a cleaner read since setting again won't hurt, but this is slightly more efficient.

labkey-susanh (approved):

PR #6897, opened 2025-08-01
Track search work by container and purge on delete; group related work

labkey-jeckels:
#### Rationale
Instead of waiting for the indexer to drain its queue before deleting a container, we can purge that work from the queue. We can also be more consistent on the priority for work that's queued, cleanly separating crawler-based indexing from indexing in response to data modifications.

#### Changes
- Track search work based on the container for the data
- Purge work for the container being deleted
- Simplify priority levels
- Ensure related indexing work is completed as part of the same batch of work

#### Tasks 📍
  - On a sizable database, Admin Console->Full Text Search->Delete Index. It should start reindexing in the background. Then insert or update some indexable items (wikis, lists, etc) and verify that the change is quickly reflected in the search results (should be only a second or two, but definitely less than 30 seconds)
  - Do a big sample import, and immediately after it finishes, delete the container. You shouldn't see deadlocks or other errors related to background indexing.
- [x] Needs Automation - improved test reliability is the goal


search/src/org/labkey/search/model/AbstractSearchService.java
```diff
@@ -1007,11 +1083,35 @@ public void shutdownStarted()
 
                 if (null != i)
                 {
+                    final Item item = i;
                     while (!_shuttingDown && _itemQueue.size() > 1000)
                     {
                         try {Thread.sleep(100);}catch(InterruptedException ignored){}
                     }
-                    i._run.run();
+                    item._run.accept(new TaskIndexingQueue()
```
labkey-matthewb, line 1092: Would it be cleaner (maybe not) to have _IndexTask and ItemRunnable "privately" implement TaskIndexingQueue to avoid the anonymous inner classes?  They seem one-to-one.

search/src/org/labkey/search/model/AbstractSearchService.java
```diff
@@ -616,11 +671,32 @@ public boolean _eq(URLHelper a, URLHelper b)
...
+    {
+        Predicate<Item> itemChecker = (i) -> container.getEntityId().equals(i._containerId) && i._pri != PRIORITY.delete;
+        _runQueue.removeIf(itemChecker);
+        _itemQueue.removeIf(itemChecker);
+
+        // Let the task know too, treating them as failures (since they were not successfully indexed)
+        for (_IndexTask task : _tasks)
+        {
+            List<Item> toRemove = task._subtasks.keySet().stream().filter(itemChecker).toList();
+            toRemove.forEach(i -> task.completeItem(i, false));
```
labkey-matthewb: move call to completeItem() into the Predicate?  The asynchrony here is hard to reason out.

labkey-matthewb (approved):

search/src/org/labkey/search/model/AbstractSearchService.java
labkey-jeckels: I'd love to collapse them, but currently Tasks can have Items from many different containers and different priorities so that anonymous inner class is what ends up stashing that context

search/src/org/labkey/search/model/AbstractSearchService.java
labkey-jeckels: That seemed to work. See the commit I pushed.

announcements/src/org/labkey/announcements/model/AnnouncementManager.java
```diff
@@ -851,26 +851,25 @@ public static void purgeContainer(Container c)
...
+    public static void indexMessages(SearchService.TaskIndexingQueue queue, Date modifiedSince)
     {
-        indexMessages(task, c.getId(), modifiedSince, null);
+        indexMessages(queue, modifiedSince, null);
     }
 
 
     // TODO: Fix inconsistency -- cid is @NotNull and we check c != null, yet some code below allows for c == null
-    public static void indexMessages(final SearchService.IndexTask task, final @NotNull String containerId, Date modifiedSince, @Nullable String threadId)
+    public static void indexMessages(SearchService.TaskIndexingQueue queue, Date modifiedSince, @Nullable String threadId)
```
labkey-nicka, line 860: nit: Remove stale TODO?

labkey-jeckels: Removed.

labkey-tchad (approved):
manual testing looks good

PR #6898, opened 2025-08-01
Update error messages to include folder or sample type when appropriate

labkey-susanh:
#### Rationale
For better error messaging when we split files up for cross-type or cross-folder import, it is helpful to know if the split has happened. This is done via properties on the `DataIteratorContext`. I opted for new Boolean properties instead of changing the existing `isCrossType` and `isCrossFolder` properties to enum properties but could be convinced otherwise.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1567

#### Changes
- Add new properties to `DataIteratorContext` and set them appropriately in the MultiPart data iterator


XingY (commented):
I feel this is a bit overkill. User still can't easily track down the problematic row from the shuffled row index. Also note that crossFolder update happens as long as there are child folders in the project, even if the data file contains sample/data from a single container, without providing container/folder column. So the error message will be different for merge vs update.

PR #6901, opened 2025-08-04
Fix SAML test IDP redirect 

labkey-adam:
#### Rationale
Slightly stricter `isLocalUri()` check allows redirect to SAML test IDP (same host, but different servlet) as an "external" URI.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/6867

### DIFF
--- a/api/src/org/labkey/api/util/URLHelper.java
+++ b/api/src/org/labkey/api/util/URLHelper.java
@@ -182,9 +182,10 @@ public class URLHelper implements Cloneable, Serializable, JSONString
     public boolean isLocalUri(ViewContext currentContext)
     {
         return getHost() == null
-                || currentContext == null
-                || currentContext.getActionURL() == null
-                || getHost().equalsIgnoreCase(currentContext.getActionURL().getHost());
+            || currentContext == null
+            || currentContext.getActionURL() == null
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/util/URLHelper.java
```diff
@@ -182,9 +182,10 @@ public static StringBuilder getBaseServer(String scheme, String host, int port)
     public boolean isLocalUri(ViewContext currentContext)
     {
         return getHost() == null
-                || currentContext == null
-                || currentContext.getActionURL() == null
-                || getHost().equalsIgnoreCase(currentContext.getActionURL().getHost());
+            || currentContext == null
+            || currentContext.getActionURL() == null
+                // Scheme check is here for test SAML redirect, which is to the same host ("localhost") but a different scheme
+            || getHost().equalsIgnoreCase(currentContext.getActionURL().getHost()) && getScheme().equalsIgnoreCase(currentContext.getActionURL().getScheme());
```
labkey-jeckels: I know operator precedence makes this work but I typically prefer explicit parens.

labkey-jeckels (approved):

api/src/org/labkey/api/util/URLHelper.java
labkey-adam: Addressed

PR #6908, opened 2025-08-06
Reduce logging for SQL exceptions from queries with calculated columns

cnathe:
#### Rationale
We have a few open issues related to the logging for calculated column errors / exceptions:
Issue [52026](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52026): Calculated column validation incorrectly flags expressions containing CAST
Issue [51862](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51862): Invalid calculated columns on DataClass breaks indexing and viewing in LKS grid

This PR looks to reduce that logging by adding a CALCULATED_COLUMN_SQL_TAG to the SQL query for the validation API check and the inclusion of the calc columns in the table. We then look for that tag in the handleSqlException() method.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1840
- https://github.com/LabKey/platform/pull/6908 
- https://github.com/LabKey/limsModules/pull/1583 
- https://github.com/LabKey/premiumModules/pull/330

#### Changes
- Add CALCULATED_COLUMN_SQL_TAG to CalculatedExpressionColumn sql
- handleSqlException to reduce logging for SQL exceptions from queries containing CALCULATED_COLUMN_SQL_TAG

#### Tasks 📍
- [x] Needs Automation


api/src/org/labkey/api/util/ExceptionUtil.java
```diff
@@ -96,6 +96,7 @@
 public class ExceptionUtil
 {
     public static final String REQUEST_EXCEPTION_ATTRIBUTE = ExceptionUtil.class.getName() + "$exception";
+    public static final String CALCULATED_COLUMN_SQL_TAG = "/* CALCULATED--EXPRESSION-COLUMN-QUERY */";
```
labkey-matthewb: It's probably fine, but I'd avoid using sql comment chars "--" inside block comments /* */  (I've seen the JDBC drivers mess up comment handling in lots of ways).

cnathe: oops, good catch. I didn't mean to put two dashes there, just one.

query/src/org/labkey/query/sql/CalculatedExpressionColumn.java
```diff
@@ -345,6 +347,7 @@ public SQLFragment getValueSql(String tableAliasName)
         sql.append("(");
         bound.appendSql(sql, new _Query(tableAliasName));
         sql.append(")");
+        sql.append(CALCULATED_COLUMN_SQL_TAG);
```
labkey-matthewb: append() doesn't yet disallow comments, but it probably should.  Maybe change this to

sql.append(SQLFragment.unsafe(CALCULATED_COLUMN_SQL_TAG))

cnathe: that makes sense. done

labkey-matthewb (approved):

labkey-jeckels (approved):
Code looks good, as does the lack of logging in my local testing.

PR #6910, opened 2025-08-07
Allow "Date" as alias for "Created" column in audit event tables

labkey-adam:
#### Rationale
Fix a bunch of cranky tests and queries. Spend time on more important things.

Probably obvious: this alias will persist after the legacy "audit" table and `_legacyNameMap` are removed (25.12+), though the code here will get simplified

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6903

Testing? Automated tests have discovered all the problems and automated tests shall verify the fix, IMO


labkey-jeckels (approved):
Looks good. Could also initialize `_legacyNameMap` with `"Date"` -> `"Created"`.

labkey-adam:
> Looks good. Could also initialize `_legacyNameMap` with `"Date"` -> `"Created"`.

That map is going away soon

PR #6912, opened 2025-08-08
ExistingRecordDataIterator out of sync

labkey-martyp:
#### Rationale
Getting an error in ExistingRecordDataIterator, like duplicate error, will hit this assert in dev mode before the error has a chance to propagate up, throwing a pretty obscure error (Seen in ETLs). This PR adds a fail fast for ExistingRecordDataIterator to bail out of next() if there are errors. 

#### Changes
- Check for errors in next()

### DIFF
--- a/api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
@@ -177,7 +177,7 @@ public abstract class ExistingRecordDataIterator extends WrapperDataIterator
         if (useMark)
             _unwrapped.mark();  // unwrapped _delegate
         boolean ret = super.next();
-        if (ret && !pkColumns.isEmpty())
+        if (!_context.getErrors().hasErrors() && ret && !pkColumns.isEmpty())
         {
             prefetchExisting();
             if (_context.shouldCancel())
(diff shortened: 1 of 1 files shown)


XingY (commented):
What happens in prod setting, would the assert be skipped and the application get into a bad state? Could we convert this assert to an exception instead?

labkey-martyp:
> What happens in prod setting, would the assert be skipped and the application get into a bad state? Could we convert this assert to an exception instead?

Yeah the assert should be skipped and, at least in ETLs, the row error throws pretty much immediately and displays the duplicate error in the ETL UI. So I'm not sure the state of the DI matters at that point. If we want to keep it in sync even when there are errors, I think we want to remove the error check [here](https://github.com/LabKey/platform/blame/9c9d9b2a414823ae33c26b92807ab496b9203e1f/api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java#L177). But this is long standing behavior, the duplicate error just made it more visible. I'm not sure all the ramifications of changing that check.

api/src/org/labkey/api/dataiterator/ExistingRecordDataIterator.java
```diff
@@ -170,11 +170,14 @@ public Object getConstantValue(int i)
...
     {
+        if (_context.getErrors().hasErrors())
+            return false;
+
         // NOTE: we have to call mark() before we call next() if we want the 'next' row to be cached
         if (useMark)
             _unwrapped.mark();  // unwrapped _delegate
         boolean ret = super.next();
-        if (!_context.getErrors().hasErrors() && ret && !pkColumns.isEmpty())
+        if (ret && !pkColumns.isEmpty())
```
XingY: Probably good idea to keep !_context.getErrors().hasErrors() here as well (in addition to the check earlier) in case new error occurred when calling super.next()

labkey-martyp: good point added back

XingY (approved):

PR #6915, opened 2025-08-11
Fix folder import with domain containing special characters

XingY:
#### Rationale
When a domain name ends with a dot (.) character, FiltUtil.makeLegalName would convert that . to _ when trying to match its import file during folder import. The conversion of dot to _ only happens when the dot is the last character in the filename. The logic is flawed in that it doesn't take into the ".tsv" suffix, which would mean the dot is in fact not the last character in the filename.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1842
- https://github.com/LabKey/labkey-ui-premium/pull/804
- https://github.com/LabKey/platform/pull/6915
- https://github.com/LabKey/limsModules/pull/1594
- https://github.com/LabKey/testAutomation/pull/2617

#### Changes
- Pass full file name to `FileUtil.makeLegalName` to get the correct file name.

### DIFF
--- a/experiment/src/org/labkey/experiment/samples/AbstractExpFolderImporter.java
+++ b/experiment/src/org/labkey/experiment/samples/AbstractExpFolderImporter.java
@@ -272,14 +272,11 @@ public abstract class AbstractExpFolderImporter implements FolderImporter
             for (ExpObject expObject : expObjects)
             {
                 String tableName = expObject.getName();
-                // tsv file name might have been generated from a sanitized table name
-                String fileName = FileUtil.makeLegalName(tableName);
-                if (dataFileMap.containsKey(fileName) || dataFileMap.containsKey(tableName))
+                // tsv file name will have been generated from a sanitized table name
+                String fileName = FileUtil.makeLegalName(tableName + ".tsv");
+                if (dataFileMap.containsKey(fileName))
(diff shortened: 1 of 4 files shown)


experiment/src/org/labkey/experiment/samples/AbstractExpFolderImporter.java
```diff
@@ -272,11 +272,14 @@ protected void importTsvData(FolderImportContext ctx, XarContext xarContext, Str
             for (ExpObject expObject : expObjects)
             {
                 String tableName = expObject.getName();
-                // tsv file name will have been generated from a sanitized table name
+                // tsv file name might have been generated from a sanitized table name
```
labkey-susanh: Perhaps add a note that the uncertainly comes from different treatment in the folder export/import case.

labkey-susanh (approved):

labkey-susanh:
Please update PR details with rationale and list of changes.

experiment/src/org/labkey/experiment/samples/AbstractExpFolderImporter.java
XingY: I just pushed a change with a different implementation.

PR #6918, opened 2025-08-12
Quantity support for volume and mass units and conversion between them

cnathe:
#### Rationale
This pull request introduces support for handling columns with quantity units by recognizing a special unit suffix in column names, updating formatting and conversion logic, and enhancing how units are displayed in the UI. The changes are focused on enabling a feature flag for quantity column suffixes, updating value formatting to handle units, and ensuring proper data conversion and display.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6918
- https://github.com/LabKey/MacCossLabModules/pull/565

#### Changes
- Added logic to detect and process column names with a `__<unit>` suffix when the `QUANTITY_COLUMN_SUFFIX_TESTING` feature is enabled
- Enhanced value formatting and conversion functions to support quantities, including new methods for formatting values with units and converting values to the appropriate quantity type
- Updated rendering logic to append unit labels to column headers and display values
- Refactored and centralized format and conversion logic for columns to use new utility methods


api/src/org/labkey/api/data/BaseColumnInfo.java
```diff
@@ -1353,12 +1355,21 @@ public void setSortFieldKeysFromXml(String xml)
...
     public static String labelFromName(String name)
     {
-        if (name == null)
-            return null;
-
-        if (name.isEmpty())
+        if (StringUtils.isBlank(name))
             return name;
 
+        // NOTE: This is just for testing (let the DataRegion/DataColumn do this)
```
XingY, line 1361: Is this NOTE still valid? It doesn't seem that DataRegion/DataColumn has the change yet.

api/src/org/labkey/api/data/BaseColumnInfo.java
```diff
@@ -1353,12 +1355,21 @@ public void setSortFieldKeysFromXml(String xml)
...
+        if (OptionalFeatureService.get().isFeatureEnabled(AppProps.QUANTITY_COLUMN_SUFFIX_TESTING))
+        {
+            var index = name.indexOf("__");
```
XingY, line 1364: Would our fuzz util generate a field name that trips on this? Perhaps we should update DomainUtil to validate a custom field name doesn't end with "__validUnit"?

api/src/org/labkey/api/data/ColumnRenderPropertiesImpl.java
```diff
@@ -785,11 +786,23 @@ public final Class<?> getJavaClass()
...
+        {
+            ret = pt.getJavaType();
+            isNumeric = pt.getJdbcType().isNumeric();
+        }
+        else
+        {
+            ret = getJdbcType().getJavaClass(isNullable);
+            isNumeric = getJdbcType().isNumeric();
+        }
+        Unit unit = getDisplayUnit();
```
XingY: Can skip `getDisplayUnit` if !isNumeric

api/src/org/labkey/api/data/DataColumn.java
```diff
@@ -940,7 +948,9 @@ public String getSortHandler(RenderContext ctx, Sort.SortDirection sort)
         // TODO: Treat null and empty the same instead?
         if (_caption == null)
             return null;
-        String title = _caption.eval(ctx);
+        var title = _caption.eval(ctx);
+        if (null != _displayColumn && null != _displayColumn.getDisplayUnit())
+            title += " (" + _displayColumn.getDisplayUnit() + ")";
```
XingY, line 953: If _displayColumn.getDisplayUnit() is blank, for example (unit, count, no_unit), this would result in " ()"

api/src/org/labkey/api/data/DisplayColumn.java
```diff
@@ -506,31 +516,65 @@ public String getFormattedText(RenderContext ctx)
...
     @NotNull
-    protected final String formatValue(RenderContext ctx, Object value, StringExpression expr, Format format)
+    protected final String formatValue(RenderContext ctx, final Object value, StringExpression expr, @Nullable Format format, @Nullable Unit displayUnit)
     {
         if (null == value)
             return "";
 
         if (null != expr && expr.canRender(ctx.getFieldMap().keySet()))
         {
+            // TODO handle Quantity values here?
```
XingY, line 524: Still TODO?

api/src/org/labkey/api/data/DisplayColumn.java
```diff
@@ -506,31 +516,65 @@ public String getFormattedText(RenderContext ctx)
...
             return expr.eval(ctx);
         }
+
+        @NotNull String formattedString;
+        if (null != displayUnit && value instanceof Number)
+        {
+            Quantity q = (value instanceof Quantity) ?
+                    (Quantity)value :
+                    displayUnit.getKindOfQuantity().toQuantity((Number)value);
+            /* DISPLAY WITH UNIT SUFFIX
```
XingY, line 534: Is this block commented out on purpose?

api/src/org/labkey/api/data/DataColumn.java
```diff
@@ -623,12 +625,18 @@ protected String getSelectInputDisplayValue(NamedObject entry)
...
+    protected String getStringValue(Object value, Unit unit, boolean disabledInput)
     {
         String strVal = "";
         //UNDONE: Should use output format here.
         if (null != value)
         {
+            if (unit != null && value instanceof Number num)
+            {
+                Quantity quantity = (value instanceof Quantity q) ? q : unit.getKindOfQuantity().toQuantity(num);
+                value = quantity.value(unit);
```
XingY, line 637: Here the value already went though conversion through `quantity.value()`, but then later if there is not `_format` or if format fails, it go through `ConvertUtils.convert` again. Is that expected?

api/src/org/labkey/api/data/DisplayColumn.java
```diff
@@ -382,13 +388,17 @@ public void setTsvFormatString(String formatString)
         _tsvFormat = createFormat(formatString, tsvFormatSymbols);
     }
 
-
     private Format createFormat(String formatString, @Nullable DecimalFormatSymbols dfs)
+    {
+        if (null == formatString)
```
XingY: Looks like this if check is not needed, the telescoped call would check null and return null anyways. Also, add @Nullable to formatString param for consistency.

api/src/org/labkey/api/ontology/Unit.java
```diff
@@ -0,0 +1,293 @@
...
+
+public enum Unit
+{
+    unit(KindOfQuantity.Count, null, 1.0, "",
+            Quantity.class,
+            "unit", "units"),
+    count(KindOfQuantity.Count, unit, 1.0, "",
+            Quantity.class,
+            "count", "count"),
+    no_unit(KindOfQuantity.Count, unit, 1.0, "",
```
XingY: I can understand count having unit as base, but why does no_unit also has unit as base? (That is if I understand why no_unit is needed at all...) Shouldn't no_unit mean unit is not applicable?

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -719,7 +719,7 @@ protected Map<String, Object> coerceTypes(Map<String, Object> row)
                     if (PropertyType.FILE_LINK.equals(col.getPropertyType()) && value instanceof String strVal)
```
XingY, line 719: I see this `if FileLine else` block repeated a few times. Now that we have the `col.getConvertFn()` mechanism to handle conversion, could the FileLink check be handled as part of it?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -1924,7 +1924,8 @@ protected class SampleUnitsConvertColumn extends SimpleTranslator.SimpleConvertC
         {
             public SampleUnitsConvertColumn(String fieldName, int indexFrom, @Nullable JdbcType to)
             {
-                super(fieldName, indexFrom, to, true);
+                // TODO reconcile unit handling
```
XingY, line 1927: Still TODO?

api/src/org/labkey/api/data/BaseColumnInfo.java
cnathe: Correct. This is just the temp workaround for testing this in this feature branch. The next round of changes, which Susan has started on, will remove this and apply the Quantity behavior to the exp.Material amount column.

api/src/org/labkey/api/data/BaseColumnInfo.java
cnathe: We shouldn't trip on this since it is behind an experimental flag. Even if the fuzz test does generate a __ field name, as long as the flag is off, that column shouldn't get this behavior applied.

api/src/org/labkey/api/data/ColumnRenderPropertiesImpl.java
cnathe: done

api/src/org/labkey/api/data/DataColumn.java
cnathe: good catch, you can currently get that if you use the "count" or "unit" type. I've added an isEmpty() check on the unit.toString()

api/src/org/labkey/api/data/DisplayColumn.java
cnathe: @labkey-matthewb ???

api/src/org/labkey/api/data/DisplayColumn.java
cnathe: I assume this was playing around with having the unit shown in the column header / title vs having the unit in the column values (i.e. "10g"). @labkey-matthewb should this be removed for now?

api/src/org/labkey/api/data/DataColumn.java
cnathe: @labkey-matthewb

api/src/org/labkey/api/data/DisplayColumn.java
cnathe: done

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
cnathe: @labkey-matthewb

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
cnathe: Yes, but not in this PR. That will be part of the next story that Susan has started on.

api/src/org/labkey/api/data/DisplayColumn.java
labkey-matthewb: When there is a format we need to convert to defaultDisplayUnit() before calling .format()

api/src/org/labkey/api/data/DisplayColumn.java
labkey-matthewb: I believe this is the current spec.  We can remove this unless we revisit this decision (or want to handle differently in some scenarios).

api/src/org/labkey/api/ontology/Unit.java
labkey-matthewb: I think I had in mind some sort of degenerate case where we had a number but haven't attached it to its Unit/KindOfQUantity yet.  Let's comment or remove this until we have a use case.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
labkey-matthewb: Maybe??? I like the idea, if the FILE_LINK case is handled consistently in the various code paths.  I don't know if that is true.

api/src/org/labkey/api/data/BaseColumnInfo.java
XingY: But perhaps we should start disallowing field names to have __unit suffix in preparation for rolling out of the feature?

cnathe: oh sorry for the confusion, I meant to say that this code will go away in the next set of changes for this epic. We will be removing the "__unit" suffix code here. This was just for testing out the Quantity behavior before the UI parts are wired up.

api/src/org/labkey/api/ontology/Unit.java
cnathe: Sounds good. I'll remove the `no_unit` from the enum.

api/src/org/labkey/api/data/DataColumn.java
XingY: Would be good to have this one addressed before merging.

cnathe: I'm confused. The first conversion in `quantity.value()` is the amounts/units conversion and the second conversion via `ConvertUtils.convert()` is just a convert to string. So I believe they are serving two different purposes, unless I am misunderstanding your comment.

XingY: Ah got it. The ConvertUtils.convert(value) just converts to string.

XingY (approved):

PR #6921, opened 2025-08-12
Don't serialize the generated attachment URL on export.

labkey-klum:
#### Rationale
[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53505)

The list attachment field download URL contains token replacement of the field name that can cause list import issues. There isn't really a need to serialize this to the list archive unless a user has explicitly set it on the `PropertyDescriptor`. The code change needs to unset the value since the base class `writeColumn` doesn't make this distinction. 

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/2624


cnathe (approved):
Code review complete and issue verified in feature branch

PR #6911, opened 2025-08-07
Support  BIGINT Primary Keys

labkey-matthewb:
#### Rationale
Implement support for tables with BIGINT primary keys.  E.g. RowId/ObjectId BIGSERIAL
Because of the large number of shared classes, utilities, services, etc, a lot of code was migrated to used Long.  Code should now assume that an generic integer object key (as opposed to a known table PK), could be a Long.

This PR is aimed at a) Supporting ObjectID BIGSERIAL b) making it possible (but not automatic) to migrate tables that want to use RowId BIGSERIAL.

#### Related Pull Requests
- 

#### Changes
New safer collection classes
  class Int/LongHashMap
  class Int/LongHashSet
explicit Class for key values in Map Objects returned by BaseSelector
  Selector.getValueMap(Class key)
safe "cast" using asInteger or asLong, as opposed to (Integer) and (Long)

PR #6932, opened 2025-08-14
Plate Hit: filter out preexisting hits match on Long

labkey-nicka:
#### Rationale
Plate hit selection removes any preexisting hits from the set to be modified prior to checking the requested hits. This is now comparing an `Integer` to a `Long` which is not desired.

#### Related Pull Requests
- #6911

#### Changes
- Use `Long` instead of `Integer` to match other changes

### DIFF
--- a/assay/src/org/labkey/assay/plate/PlateManager.java
+++ b/assay/src/org/labkey/assay/plate/PlateManager.java
@@ -3176,7 +3176,7 @@ public class PlateManager implements PlateService, AssayListener, ExperimentList
                     SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("ResultId"), rowIds, CompareType.IN);
                     filter.addCondition(FieldKey.fromParts("ProtocolId"), protocol.getRowId());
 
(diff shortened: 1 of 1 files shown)


assay/src/org/labkey/assay/plate/PlateManager.java
```diff
@@ -3176,7 +3176,7 @@ public void markHits(
                     SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("ResultId"), rowIds, CompareType.IN);
                     filter.addCondition(FieldKey.fromParts("ProtocolId"), protocol.getRowId());
 
-                    Set<Integer> preexistingHits = new HashSet<>(new TableSelector(hitTable, Collections.singleton("ResultId"), filter, null).getArrayList(Integer.class));
+                    Set<Long> preexistingHits = new HashSet<>(new TableSelector(hitTable, Collections.singleton("ResultId"), filter, null).getArrayList(Long.class));
```
labkey-matthewb: consider switching to "new LongHashSet(...)"

labkey-matthewb (approved):

assay/src/org/labkey/assay/plate/PlateManager.java
labkey-nicka: Done

PR #6933, opened 2025-08-15
IntegerUtils class

labkey-matthewb:
#### Rationale
Helper for dealing with Numeric objects

#### Related Pull Requests
- 

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/util/IntegerUtils.java
+++ b/api/src/org/labkey/api/util/IntegerUtils.java
@@ -35,7 +35,7 @@ public class IntegerUtils
         return o.getClass() == Long.class ? (Long)o : Long.valueOf(((Number)o).longValue());
     }
 
-    /** returns Numeric object as a Long.  This method will throw if not Object is not Integral or is out of range. */
+    /** returns Numeric object as an Integer.  This method will throw if not Object is not Integral or is out of range. */
     public static Integer asInteger(Object o) throws ClassCastException
     {
         if (null == o)
--- a/api/src/org/labkey/api/util/IntegerUtils.java
+++ b/api/src/org/labkey/api/util/IntegerUtils.java
@@ -1,0 +1,182 @@
+package org.labkey.api.util;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import java.util.Objects;
+
+/** Class to help with dealing with Object that may represent an integer number (char,short,int,long)
+ * Unfortunately, Number does not help with detecting integer/non-integer types, so this class only handles Object
+ * instances that have corresponding to the primitive types.  {@code boolean}, {@code byte}, {@code char},
+ * {@code short}, {@code int}, {@code long}, {@code float}, and {@code double})
+ * <br>
+ * Because "Integer" is kind of ambiguous, I will use "Integral" to mean any integer type.  For now, I am not including
+ * Boolean or Character as "Integral", change this if it seems useful.
+ */
+
+public class IntegerUtils
+{
+    /** Return true if Object is Byte, Short, Integer, or Long */
+    public static boolean isIntegral(Object o)
+    {
+        if (null == o)
+            return false;
+        final Class<?> c = o.getClass();
+        return c == Byte.class || c == Short.class || c==Integer.class || c==Long.class;
+    }
+
+    /** returns Numeric object as a Long.  This method will throw if not Object is not Integral. */
+    public static Long asLong(Object o)
+    {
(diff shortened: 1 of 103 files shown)


api/src/org/labkey/api/util/IntegerUtils.java
```diff
@@ -0,0 +1,203 @@
...
+
+    /** returns Numeric object as a Long.  This method will throw if not Object is not Integral. */
+    public static Long asLong(Object o)
+    {
+        if (!isIntegral(o))
+            return (Long)o;  // throw ClassCastException
+        return o.getClass() == Long.class ? (Long)o : Long.valueOf(((Number)o).longValue());
+    }
+
+    /** returns Numeric object as a Long.  This method will throw if not Object is not Integral or is out of range. */
```
labkey-adam:
```suggestion
    /** returns Numeric object as an Integer.  This method will throw if not Object is not Integral or is out of range. */
```

api/src/org/labkey/api/util/IntegerUtils.java
```diff
@@ -0,0 +1,203 @@
...
+     *      return i;
+     *  else
+     *      throw ..
+     *  <br>
+     *  if (asIntegerElseNull() instanceof Integer i))
+     *      return i
+     *  else
+     *      throw .. }
+     */
+    public static Object asIntegerElseNull(Object o)
```
labkey-adam, line 68: Why not return `Integer`?

api/src/org/labkey/api/util/IntegerUtils.java
```diff
@@ -0,0 +1,203 @@
...
+        if (c == Integer.class)
+            return o;
+        if (c == Byte.class || c == Short.class)
+            return Integer.valueOf(((Number)o).intValue());
+        if (c == Long.class)
+        {
+            long l = ((Number) o).longValue();
+            return Integer.MIN_VALUE <= l && l <= Integer.MAX_VALUE ? Integer.valueOf((int) l) : null;
+        }
+        return null;
```
labkey-adam, line 82: Seems like this as `asInteger()` should be able to share more implementation. But maybe not...

studydesign/src/org/labkey/studydesign/model/AssaySpecimenConfigImpl.java
```diff
@@ -251,26 +253,26 @@ public static AssaySpecimenConfigImpl fromJSON(@NotNull JSONObject o, Container
...
         if (o.has("SampleUnits") && !StringUtils.isEmpty(o.getString("SampleUnits")))
             assay.setSampleUnits(o.getString("SampleUnits"));
-        if (o.has("PrimaryTypeId") && (o.get("PrimaryTypeId") instanceof Integer || o.get("PrimaryTypeId") instanceof Long))
-            assay.setPrimaryTypeId(o.getInt("PrimaryTypeId"));
-        if (o.has("DerivativeTypeId") && (o.get("DerivativeTypeId") instanceof Integer || o.get("DerivativeTypeId") instanceof Long))
-            assay.setDerivativeTypeId(o.getInt("DerivativeTypeId"));
+        if (o.has("PrimaryTypeId") && asIntegerElseNull(o.get("PrimaryTypeId")) instanceof Integer primaryTypeId)
+            assay.setPrimaryTypeId(primaryTypeId);
+        if (o.has("DerivativeTypeId") && asIntegerElseNull(o.get("DerivativeTypeId")) instanceof Integer derivativeTypeId)
+            assay.setDerivativeTypeId(derivativeTypeId);
```
labkey-adam, line 275: So glad we're still maintaining this code... but definitely better patterns here!

labkey-adam (approved):
See comment fix and a couple questions. We should also deprecate `ExperimentService.asLong()/asInteger()`.

api/src/org/labkey/api/util/IntegerUtils.java
labkey-matthewb: The compiler won't let you do `if (asIntegerElseNull(x) instanceof i) {...}` if this returns Integer.  So that would mean doing something like `Integer i; if (null != (i = asIntegerElseNull(x)) {...}`

This is definitely  a judgement call, but this seemed to be the 95% use case.

labkey-adam: Naw, this is fine. And of course you already have a comment about why Object... I just didn't read it.

api/src/org/labkey/api/util/IntegerUtils.java
labkey-matthewb: I might look at these again to make them most efficient in the expected case of  Integer/Long being the most common type.  Maybe they'll end more similar.

PR #6934, opened 2025-08-15
Force exp.Object.ObjectId to use big integers in dev mode

labkey-adam:
#### Rationale
We want development and test deployments to always work with big (> `Integer.MAX_VALUE`) ObjectIds to verify full and continued support for large values

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6911

#### Tasks
- [ ] ~Manual Testing~
- [ ] ~Needs Automation~
- [ ] ~Verify Fix~


labkey-matthewb (approved):
Kinda trusting the reordering code on the SQL changes.

PR #6935, opened 2025-08-15
Abort the startup process faster when we determine we're shutting down

labkey-jeckels:
#### Rationale
We do a lot of logging about having our startup interrupted when Spring Boot decides there's a fatal problem.

#### Changes
- Bail out once we see a `ShuttingDownException`

#### Tasks 📍


labkey-susanh (approved):
Though I can't say that behavior is significantly better with this change, it is, at least, not worse. Still lots of logging before and after the message about the port being in use. I think there is less logging of "NoClassDefFoundError"s after that message, though, which is good.

PR #6939, opened 2025-08-18
Issue 53586: Domain field with long name can result in "PropertyURI cannot exceed 300 characters, but was 302 characters long."

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53586

Several places in the code generated PropertyURIs for a domain field by appending the encoded field name to the domain type URI prefix. This can create long PropertyURI values that exceed our 300 char limit. This PR updates the DomainUtil.createUniquePropertyURI() helper to replace the "encode field name" with a Lsid DB seq as we've moved towards for other URI related cases recently.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6805
- https://github.com/LabKey/platform/pull/6939
- https://github.com/LabKey/testAutomation/pull/2646
- https://github.com/LabKey/clientModules/pull/56
- https://github.com/LabKey/commonAssays/pull/920

#### Changes
- Update DomainUtil.createUniquePropertyURI() to use LsidManager.getLsidPrefixDbSeq()
- Use DomainUtil.createUniquePropertyURI() in several places instead of appending encoded field name


api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -1107,8 +1108,12 @@ public static DomainProperty addProperty(Domain domain, GWTPropertyDescriptor pd
         return p;
     }
 
-    private static String createUniquePropertyURI(String base, Set<String> propertyUrisInUse)
+    public static String createUniquePropertyURI(String typeURI, Set<String> propertyUrisInUse)
```
labkey-nicka: Seems like we want all the callers of this to use a case-insensitive set. I'd recommend either switching them all to use `CaseInsensitiveHashSet`. 

image

Or even better just hide this part of the implementation from the publicly exported variation like this:

```java
public static String createUniquePropertyURI(String typeURI)
{
    return createUniquePropertyURI(typeURI, new CaseInsensitiveHashSet());
}

private static String createUniquePropertyURI(String typeURI, Set<String> propertyUrisInUse)
{
    // ...
}
```

cnathe: Good suggestion. Done

labkey-klum (approved):
I pulled this down and ran with it locally. I'd be fine relying on the automated tests over a manual test pass.

api/src/org/labkey/api/exp/property/DomainUtil.java
labkey-susanh: Does this include changing the callers of `addProperty` to also use a CaseInsensitiveHashSet?

cnathe: I just did that now as well. Thanks.

experiment/src/org/labkey/experiment/XarExporter.java
```diff
@@ -793,7 +793,7 @@ public void addDataClass(ExpDataClass dataClass) throws ExperimentException
         queueDomain(dataClass.getDomain());
     }
 
-    // Return the "name" portion of the propertyURI after the hash
+    // Return the "name" portion of the propertyURI after the hash (Issue 30718)
     private String getPropertyName(DomainProperty dp)
```
labkey-jeckels, line 798: This behavior seems strange now. It's not really returning the property name for (most) new-style PropertyURIs

cnathe: @labkey-jeckels I agree but I don't think this code can actually be hit for any newly created properties. It is executed for Sample Types that have the parentCol, idCol1, idCol2, idCol3 behavior which is deprecated. I couldn't find a way in the UI to create one of these. I took the XAR file from that issue 30718 and imported it locally and this code did the right thing for those PropertyURIs in the XAR. Thoughts on other ways to test this? Since you can't create these types of Sample Types anymore, I think the code should continue working for any existing property descriptors in those deprecated sample types.

labkey-jeckels: Got it. I hadn't looked at the usages. Do you think we should mark this or related methods as deprecated?

cnathe: Yeah, I'll take a look

PR #6947, opened 2025-08-20
Give linked schemas access to sample type, etc. defs in project/shared

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53421

### DIFF
--- a/api/src/org/labkey/api/exp/api/ExperimentService.java
+++ b/api/src/org/labkey/api/exp/api/ExperimentService.java
@@ -289,6 +289,7 @@ public interface ExperimentService extends ExperimentRunTypeSource
      * Get all DataClass definitions in the container.  If <code>includeOtherContainers</code> is true,
      * a user must be provided to check for read permission of the containers in scope.
      */
+    // TODO: Remove user parameter (not used)
     List<? extends ExpDataClass> getDataClasses(@NotNull Container container, @Nullable User user, boolean includeOtherContainers);
 
     /**
@@ -300,6 +301,7 @@ public interface ExperimentService extends ExperimentRunTypeSource
      * Get a DataClass by name within scope -- current, project, and shared.
      * Requires a user to check for container read permission.
      */
+    // TODO: Remove user parameter (not used)
     ExpDataClass getDataClass(@NotNull Container scope, @NotNull User user, @NotNull String dataClassName);
 
     /**
@@ -311,6 +313,7 @@ public interface ExperimentService extends ExperimentRunTypeSource
      * Get a DataClass by rowId within scope -- current, project, and shared.
      * Requires a user to check for container read permission.
      */
+    // TODO: Remove user parameter (not used)
     ExpDataClass getDataClass(@NotNull Container scope, @NotNull User user, int rowId);
 
     /**
@@ -328,6 +331,7 @@ public interface ExperimentService extends ExperimentRunTypeSource
     /**
      * Get a DataClass with name at a specific time.
      */
+    // TODO: Remove user parameter (not used)
     @Nullable ExpDataClass getEffectiveDataClass(
         @NotNull Container definitionContainer,
         @NotNull User user,
@@ -339,6 +343,7 @@ public interface ExperimentService extends ExperimentRunTypeSource
     /**
      * Get a ExpProtocol with name at a specific time.
      */
+    // TODO: Delete?
     ExpProtocol getEffectiveProtocol(Container container, User user, String schemaName, Date effectiveDate, ContainerFilter dataTypeCF);
 
     /**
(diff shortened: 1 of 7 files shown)


labkey-jeckels:
Assay handles this differently. It doesn't care if the user has access to the container where the design is stored. It shows all of the designs that are considered in scope. That is, every user with access to a container will see the same list of assay designs, and their data (if any) that's been loaded in the container.

Sample types don't let you see designs defined in a container you don't have permission to see, even when they have data rows a container where you have permission.

I propose we make sample types behave like assays instead of special-casing linked schemas.

labkey-adam:
> Assay handles this differently. It doesn't care if the user has access to the container where the design is stored. It shows all of the designs that are considered in scope. That is, every user with access to a container will see the same list of assay designs, and their data (if any) that's been loaded in the container.
> 
> Sample types don't let you see designs defined in a container you don't have permission to see, even when they have data rows a container where you have permission.
> 
> I propose we make sample types behave like assays instead of special-casing linked schemas.

That's fine. Should we do the same for data classes and experiments? All three share the same permission-checking logic right now.

labkey-jeckels:
> > Assay handles this differently. It doesn't care if the user has access to the container where the design is stored. It shows all of the designs that are considered in scope. That is, every user with access to a container will see the same list of assay designs, and their data (if any) that's been loaded in the container.
> > Sample types don't let you see designs defined in a container you don't have permission to see, even when they have data rows a container where you have permission.
> > I propose we make sample types behave like assays instead of special-casing linked schemas.
> 
> That's fine. Should we do the same for data classes and experiments? All three share the same permission-checking logic right now.

Yes, I think so. @labkey-hannah do you agree? Let me know if the current or proposed behavior isn't clear from the comments.

labkey-susanh:
> > > Assay handles this differently. It doesn't care if the user has access to the container where the design is stored. It shows all of the designs that are considered in scope. That is, every user with access to a container will see the same list of assay designs, and their data (if any) that's been loaded in the container.
> > > Sample types don't let you see designs defined in a container you don't have permission to see, even when they have data rows a container where you have permission.
> > > I propose we make sample types behave like assays instead of special-casing linked schemas.
> > 
> > 
> > That's fine. Should we do the same for data classes and experiments? All three share the same permission-checking logic right now.
> 
> Yes, I think so. @labkey-hannah do you agree? Let me know if the current or proposed behavior isn't clear from the comments.

If we did this, it sounds like it could address a current permissions sticking point for the apps where some clients want to give access to data in a folder without having to give read access to the project (where all the types are defined).  We'd have to do a similar thing for freezers, though.

labkey-hannah:
> > > > Assay handles this differently. It doesn't care if the user has access to the container where the design is stored. It shows all of the designs that are considered in scope. That is, every user with access to a container will see the same list of assay designs, and their data (if any) that's been loaded in the container.
> > > > Sample types don't let you see designs defined in a container you don't have permission to see, even when they have data rows a container where you have permission.
> > > > I propose we make sample types behave like assays instead of special-casing linked schemas.
> > > 
> > > 
> > > That's fine. Should we do the same for data classes and experiments? All three share the same permission-checking logic right now.
> > 
> > 
> > Yes, I think so. @labkey-hannah do you agree? Let me know if the current or proposed behavior isn't clear from the comments.
> 
> If we did this, it sounds like it could address a current permissions sticking point for the apps where some clients want to give access to data in a folder without having to give read access to the project (where all the types are defined). We'd have to do a similar thing for freezers, though.

Oooh, well definitely on board for that. Are there other ramifications to consider? I'm guessing not too many if assay already works this way...

labkey-adam:
> > > > > Assay handles this differently. It doesn't care if the user has access to the container where the design is stored. It shows all of the designs that are considered in scope. That is, every user with access to a container will see the same list of assay designs, and their data (if any) that's been loaded in the container.
> > > > > Sample types don't let you see designs defined in a container you don't have permission to see, even when they have data rows a container where you have permission.
> > > > > I propose we make sample types behave like assays instead of special-casing linked schemas.
> > > > 
> > > > 
> > > > That's fine. Should we do the same for data classes and experiments? All three share the same permission-checking logic right now.
> > > 
> > > 
> > > Yes, I think so. @labkey-hannah do you agree? Let me know if the current or proposed behavior isn't clear from the comments.
> > 
> > 
> > If we did this, it sounds like it could address a current permissions sticking point for the apps where some clients want to give access to data in a folder without having to give read access to the project (where all the types are defined). We'd have to do a similar thing for freezers, though.
> 
> Oooh, well definitely on board for that. Are there other ramifications to consider? I'm guessing not too many if assay already works this way...

There's really one ramification: today, you see the project's sample type / data class / experiment definitions in subfolders *only* if you have read permissions in the project. (The tables may have no rows in that folder, but you can see the table name, column names/labels, etc.) With this change, you would see all project definitions in all subfolders where you have read permissions, even if you have no permissions in the project.

I expect we'll see some test failures with this change...

labkey-adam:
@labkey-jeckels new approach is ready for review. As you can see, shared and project are now added unconditionally. Assuming this passes muster, I'll remove the `User` parameter from this method and callers.

labkey-jeckels (approved):

PR #6949, opened 2025-08-20
Revise file/attachment exception handling

labkey-nicka:
#### Rationale
This addresses several test failures due to changes in #6941 which resulted in different, albeit desired, conversion exception handling of file/attachment fields.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6949
- https://github.com/LabKey/limsModules/pull/1620
- https://github.com/LabKey/testAutomation/pull/2643

#### Changes
- Revise `AttachmentDataIterator` to give more informative error in non-string value case.
- Assimilate `BatchValidationException` handling when creating a run 😿.

### DIFF
--- a/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
+++ b/api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
@@ -943,7 +943,7 @@ public abstract class AbstractAssayTsvDataHandler extends AbstractExperimentData
                             }
                             catch (ConvertHelper.FileConversionException e)
                             {
-                                throw new ApiUsageException(e);
+                                errors.add(new PropertyValidationError(e.getMessage(), pd.getName()));
                             }
                         }
                     }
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
```diff
@@ -943,7 +943,7 @@ else if (o instanceof MvFieldWrapper mvWrapper)
                             }
                             catch (ConvertHelper.FileConversionException e)
                             {
-                                throw new ApiUsageException(e);
+                                errors.add(new PropertyValidationError(e.getMessage(), pd.getName()));
```
XingY, line 946: Would this result in inconsistent error msg for FileLink? Can we just do errors.add(e.gerMessage())?

XingY (approved):

api/src/org/labkey/api/assay/AbstractAssayTsvDataHandler.java
labkey-nicka: I looked into the exact place and its trivial to handle in a similar manner BUT it causes the exception chaining that we're trying to avoid with those `ApiUsageException` opt-outs. I'm going to investigate a better way so we could do this.

In the meantime, I plan on merging these changes as-is assuming test runs look good.

PR #6950, opened 2025-08-20
Encode URL query and fragment parts in AppPipelineJobNotificationProvider 

labkey-tchad:
#### Rationale
Pipeline notification links for assay imports don't work if the assay protocol name includes a slash. The same goes for file imports from files including an ampersand.

#### Related Pull Requests
- 

#### Changes
- Encode URL query and fragment parts in AppPipelineJobNotificationProvider

### DIFF
--- a/api/src/org/labkey/api/pipeline/AppPipelineJobNotificationProvider.java
+++ b/api/src/org/labkey/api/pipeline/AppPipelineJobNotificationProvider.java
@@ -14,6 +14,7 @@ import org.labkey.api.security.SecurityManager;
 
 import org.labkey.api.security.User;
 import org.labkey.api.security.permissions.ReadPermission;
+import org.labkey.api.util.PageFlowUtil;
 import org.labkey.api.util.URLHelper;
 import org.labkey.api.view.ActionURL;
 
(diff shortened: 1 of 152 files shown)


api/src/org/labkey/api/pipeline/AppPipelineJobNotificationProvider.java
```diff
@@ -276,7 +277,7 @@ else if (job instanceof AssayUploadPipelineJob)
                 String provider = (String) info.get("provider");
                 String assayName = (String) info.get("assayName");
                 long runId = asLong(info.get("runId"));
-                urlFragment = urlFragment + "/" + provider + "/" + assayName + "/runs/" + runId;
+                urlFragment = urlFragment + "/" + provider + "/" + PageFlowUtil.encode(assayName) + "/runs/" + runId;
```
labkey-jeckels: We should encode the provider name too, though that's a lot less likely to include special characters

labkey-jeckels (approved):

labkey-tchad:
Fix went to 25.7: https://github.com/LabKey/platform/pull/6953

PR #6955, opened 2025-08-22
Issue 53672: List export / import roundtrip issue with long key name

labkey-jeckels:
#### Rationale
We should round-trip list archives without errors, even with an egregiously long key field name.

#### Changes
- Use the storage column name when generating SQL

#### Tasks 📍
- [x] Manual Testing @cnathe 
- [x] Needs Automation
- [x] Verify Fix


labkey-adam (approved):

labkey-adam:
Note the "Specimen Postgres" test failure

cnathe (approved):
Manual testing complete in fb as well

PR #6958, opened 2025-08-25
Add trace logging to client abort exception

labkey-martyp:
#### Rationale
Add logging to help troubleshoot client abort exceptions.

#### Changes
- Add trace logging if client abort exception

### DIFF
--- a/api/src/org/labkey/api/util/ExceptionUtil.java
+++ b/api/src/org/labkey/api/util/ExceptionUtil.java
@@ -632,6 +632,20 @@ public class ExceptionUtil
         }
     }
 
+    private static void logClientAbortException(Throwable ex)
+    {
+        String className = ex.getClass().getName();
+        String message = ex.getMessage();
+        Throwable cause = ex.getCause();
+
+        if (message != null) {
+            LOG.trace("Client abort exception: {} - {}", className, message);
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/util/ExceptionUtil.java
```diff
@@ -632,6 +632,20 @@ protected void renderView(Object model, HttpServletRequest request, HttpServletR
...
+        String message = ex.getMessage();
+        Throwable cause = ex.getCause();
+
+        if (message != null) {
+            LOG.trace("Client abort exception: {} - {}", className, message);
+        } else if (cause != null) {
+            LOG.trace("Client abort exception: {} (cause: {})", className, cause.getClass().getName());
+        } else {
+            LOG.trace("Client abort exception: {}", className);
+        }
```
labkey-jeckels: Is there a reason to not just do the standard exception logging?

```suggestion
        LOG.trace("Client abort exception", ex);
```

labkey-martyp: Ok I wasn't sure if we had to be more defensive since we're handling many type of exceptions here. Updated.

labkey-jeckels (approved):

PR #6960, opened 2025-08-25
Drop viability schema on SQL Server

labkey-adam:
#### Rationale
Several developers have run into INT --> BIGINT upgrade problems on SQL Server because their database included an old `viability` schema with an FK to `exp.Object.ObjectId`. Drop that schema if it exists.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6911

No testing needed, IMO.


labkey-jeckels (approved):
I created an empty schema locally and copy/pasted the statement to verify it had the right syntax

PR #6962, opened 2025-08-25
Issue 53784: Field name with quote causes error for linked schema

labkey-jeckels:
#### Rationale
We should support all column names in linked schemas. Even those with quotes in them.

#### Changes
- Properly escape the quotes
- Minor code cleanup

#### Tasks 📍
- [x] Manual Testing / Verify Fix @labkey-danield 
- [x] Needs Automation

### DIFF
--- a/api/src/org/labkey/api/query/QueryKey.java
+++ b/api/src/org/labkey/api/query/QueryKey.java
@@ -200,8 +200,8 @@ import java.util.Objects;
 
     static protected boolean contains(String[] strs, String s)
     {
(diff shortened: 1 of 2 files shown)


labkey-jeckels:
@labkey-danield think we should add a quote to one of the fields in your new test, or somewhere else? Fuzzing will hit it periodically, but should be simple to make sure we have consistent coverage.

https://github.com/LabKey/testAutomation/pull/2645

labkey-adam (approved):

labkey-danield (approved):

PR #6963, opened 2025-08-25
Support dual-series plots for Trailing Mean and Trailing CV

labkey-jeckels:
#### Rationale
The two "Trailing" plot types need tweaks to support dual-series plotting and percent-of-mean and standard-deviation y-axis types.

#### Changes
- Look for the right property names and pass them through


labkey-jeckels:
@cnathe this is in support of a Panorama PR to better support dual-series metric plotting. Some of the plot types didn't support dual-series at all before this change. I'd like to get the platform PR merged sooner so that I don't have to keep maintaining it.
https://github.com/LabKey/targetedms/pull/1111

cnathe (approved):

PR #6967, opened 2025-08-26
Issue 52913: Fix determination of existing index names when dropping

labkey-adam:
#### Rationale
Reconcile and consolidate multiple index handling code paths into a single method that creates/drops indices on provisioned domains. And, especially, stop trying to guess an index's name when dropping it.

In the related PR, `AbstractAuditTypeProvider.updateIndices()` (which was already inspecting index meta data to determine indices to add/drop based on their characteristics) was taught to pass existing index names on to `StorageProvisioner` for dropping. This PR pushes the audit implementation into StorageProvisioner and switches callers of the redundant methods to use it.

The PR includes a bonus side quest that fixes domain URIs and domain URI resolution in a bunch of audit domains to ensure that we can use `domain.getDomainKind()` with confidence.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6633

#### Tasks 📍
- [x] Move AbstractAuditTypeProvider.updateIndices() code into StorageProvisioner
- [x] Fix Domains getting tagged with the wrong DomainKind 
- [x] Junit tests that flag audit DomainKinds with overlapping prefixes
- [x] Switch all callers to use new `StorageProvisioner.ensureTableIndices()` method
- [x] Remove old code
- [x] Manual Testing / Verify Fix @labkey-chrisj 
  - Ideally on a long-running database that's seen multiple upgrades, *Before pulling down the feature branch and upgrading*:
    - Review the "Admin Console - Audit Log" events for LDAP sync events, Study Security Escalations, Logged query events, Query export events, and EHR Security Escalations (if you have EHR built locally...)
    - Review in the Query Schema Browser queries in the "auditLog" schema corresponding to the above - PremiumLdapAuditEvent, StudySecurityEscalationEvent, LoggedQuery, QueryExportAuditEvent, and EHRSecurityEscalationEvent - Use Folder Filter -> All Folders to show all rows like the above do automatically
  - Sync FBs, build and start the server to run the exp upgrade
    - Re-review all "Admin Console - Audit Log" events and "auditLog" queries from above and verify the same row counts appear after upgrade. And no errors, of course.
  - On some lists and datasets, for some columns, check the Advanced checkbox that requires unique values. (Of course the values need to be unique or the table needs to be empty.) Check and save. Uncheck and save. Check and save. This will create and drop indices.
- [x] Needs Automation: No


api/src/org/labkey/api/audit/query/AbstractAuditDomainKind.java
```diff
@@ -418,4 +428,59 @@ public boolean isUserCreatedType()
...
         return false;
     }
+
+    private static final Set<String> PREFIXES = new HashSet<>();
+
+    // Called at audit provider initialization time to forbid domain kinds with shared and overlapping namespace prefixes
+    public void validate()
+    {
+        String prefix = getNamespacePrefix();
+        if (prefix.length() < 12)
```
labkey-susanh, line 440: 12 seems magical here. Comment? Constant?

experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
```diff
@@ -853,147 +855,11 @@ public void createStorageTable(Domain domain, DomainKind<?> kind, DbScope scope)
...
-                transaction.commit();
-            }
-        }
-    }
-
-    private static void ensureIndexToBeAddedHasNoPrimaryKeys(SchemaTableInfo schemaTableInfo, Map.Entry<String, PropertyStorageSpec.Index> requiredIndexEntry)
+    // The old addMissingRequiredIndices() method called this to ensure no index used any column in the PK. Why so
+    // strict? An index that *starts with* the same column(s) as another index or PK is likely redundant, but that's
+    // not unique to PKs and could be valid when a column in the middle of a composite PK is indexed. TODO: Decide
+    // if this should be removed or incorporated into ensureTableIndices().
```
labkey-susanh: I can't think why we would need this.

api/src/org/labkey/api/data/dialect/BasePostgreSqlDialect.java
```diff
@@ -1171,6 +1165,10 @@ private List<String> getRenameColumnsStatement(TableChange change)
             }
         }
 
+        // TODO: This loop should not guess the name of the old indices; instead, it should look them up.
+        // TableChange.setIndexedColumns() could set _indexRenames providing the name, and then this code uses that info.
+        // Or maybe schemaTableInfo.getAllIndices() and then use Index.isSameIndex() to find names.
+        // Or maybe we don't bother renaming the indices since (on PostgreSQL) we always look them up when dropping them.
```
labkey-susanh: Not renaming seems sort of the most straightforward and assures that we wouldn't be guessing, but if we have to/want to rename for SQLServer and it would cause special casing for the two dialects, your second option seems reasonable.

api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -990,6 +989,19 @@ public static ValidationException updateDomainDescriptor(GWTDomain<? extends GWT
...
     }
 
+    public static Map<String, Pair<TableInfo.IndexType, List<ColumnInfo>>> getExistingIndices(Domain domain)
+    {
+        SchemaTableInfo schemaTableInfo = StorageProvisioner.get().getSchemaTableInfo(domain);
+        if (schemaTableInfo != null)
+        {
+            DomainKind<?> dk = domain.getDomainKind();
+            if (null == dk)
+                throw new IllegalStateException("Domain kind is null!");
```
labkey-susanh: The domain kind isn't actually used here. Is this the right place for the null check?

labkey-susanh (commented):
The PR rationale says we stop guessing index names but there is still a TODO related to that. Still something you plan to do in this PR?

api/src/org/labkey/api/exp/property/DomainUtil.java
labkey-adam: It is not... and the method is no longer used. It was used only for temporarily testing the old way vs. the new way. Thanks... I'll remove the method.

experiment/src/org/labkey/experiment/api/property/StorageProvisionerImpl.java
labkey-adam: Agreed, I'll remove

api/src/org/labkey/api/data/dialect/BasePostgreSqlDialect.java
labkey-adam: Agreed. Since I wrote that, I've leaned more in the pro-rename direction... there are other scenarios (add index, rename column -> rename index, add column with old name, add index) where we might conflict. We prevent duplicate indexes (based on the column list) but I don't think we check for index name in use. Agree that the second approach is reasonable.

api/src/org/labkey/api/audit/query/AbstractAuditDomainKind.java
labkey-adam: It's completely arbitrary, mostly meant to prevent "Audit-" from being used, but also trivial prefixes that are likely to overlap. I'll add a comment.

labkey-adam:
> The PR rationale says we stop guessing index names but there is still a TODO related to that. Still something you plan to do in this PR?

I'm keen to get rid of this one last code path that guesses index names (on PostgreSQL... I'm not going to fight SQL Server at this point), but this PR has already taken longer than I wanted and I'd like it to bake a while before making more changes. And I'm anxious to get those SQL scripts and schema version bump merged. I've opened Issue 53838 to address this as a follow-on (and updated the comment in the method).

labkey-susanh (approved):

PR #6969, opened 2025-08-27
Issue 53608 and 53680: Trouble with namings

labkey-susanh:
#### Rationale
Issue [53608](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53608) - The names for assays and workflow templates should not interfere with each other
Issue [53680](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53608) - Extra decoding of LSID objectId is not needed and causes problems.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1632

#### Changes
- Check that both protocols are of the same type and both are assay protocols
- Remove extra call for decoding the domain ObjectId part

#### Tasks 📍
- [x] Needs Automation 

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -315,8 +315,6 @@ import static org.labkey.api.util.IntegerUtils.asLong;
 import static org.labkey.api.exp.api.NameExpressionOptionService.NAME_EXPRESSION_REQUIRED_MSG;
 import static org.labkey.api.exp.api.NameExpressionOptionService.NAME_EXPRESSION_REQUIRED_MSG_WITH_SUBFOLDERS;
 import static org.labkey.api.exp.api.ProvenanceService.PROVENANCE_PROTOCOL_LSID;
-import static org.labkey.api.exp.query.ExpSchema.SCHEMA_EXP_DATA;
-import static org.labkey.api.exp.query.SamplesSchema.SCHEMA_SAMPLES;
 import static org.labkey.api.util.IntegerUtils.asLongElseNull;
 import static org.labkey.experiment.api.SampleTypeServiceImpl.SampleChangeType.rollup;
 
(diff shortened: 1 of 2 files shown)


labkey-jeckels:
Looks good for the [53680](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53608) fix.

This a good example of why I'm wary of making changes to disallow certain names without also fixing up existing names. I know it's hard to know what to do with the preexisting data, but they'll just keep lurking and causing problems.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -6469,9 +6469,9 @@ public Protocol saveProtocol(
             boolean newProtocol = protocol.getRowId() == 0;
             if (newProtocol)
             {
-                // if protocol exist, throw error
+                // if protocol exists, throw error
                 ExpProtocol existing = getExpProtocol(protocol.getContainer(), protocol.getName());
-                if (existing != null && protocol.getLSIDNamespacePrefix().equals("GeneralAssayProtocol"))
+                if (existing != null && protocol.getLSIDNamespacePrefix().equals("GeneralAssayProtocol") && existing.getLSIDNamespacePrefix().equals("GeneralAssayProtocol"))
```
labkey-jeckels: Consider extracting this as a constant somewhere. Maybe AssayService?

labkey-jeckels (approved):

experiment/src/org/labkey/experiment/defaults/DefaultValueServiceImpl.java
```diff
@@ -64,7 +64,7 @@ private String getContainerDefaultsLSID(Container container, Domain domain)
         if (kind != null && kind.isUserCreatedType())
         {
             Lsid domainLsid = new Lsid(domain.getTypeURI());
-            return (new Lsid(DOMAIN_DEFAULT_VALUE_LSID_PREFIX, suffix, Lsid.decodePart(domainLsid.getObjectId()))).toString();
+            return (new Lsid(DOMAIN_DEFAULT_VALUE_LSID_PREFIX, suffix, domainLsid.getObjectId())).toString();
```
XingY, line 67: What about line 81 below, does Lsid.decodePart need to be removed there as well?
`return (new Lsid(USER_DEFAULT_VALUE_DOMAIN_PARENT, suffix, Lsid.decodePart(domainLsid.getObjectId()))).toString();
`

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
XingY: Is similar check needed for renaming assay, in AssayDomainServiceImpl.saveChanges?

experiment/src/org/labkey/experiment/defaults/DefaultValueServiceImpl.java
labkey-susanh: Good call. Thanks.

experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
labkey-susanh: That method already has checks for existing names, using a method from `AssayService`, so I'll just use that here too since it also does the check in a case-insensitive way.

labkey-susanh: No longer used.

PR #6974, opened 2025-08-29
Propagate all existing issue fields for validation

labkey-jeckels:
#### Rationale
I'm working on a script to bulk resolve issues. It's using the `issues-issues` API. That fails when there's an admin-defined field that's marked as required and the API request doesn't provide a value. It should use existing values for validation purposes when there's no value provided.

#### Changes
- Propagate all existing values as defaults, not just the ones that are mandatory to include in the domain
- Handle nulls in the incoming data and properly unset values
- Minor code cleanup

#### Tasks 📍
- [x] Manual Testing @labkey-klum 
- [x] Needs Automation


labkey-klum (approved):

labkey-jeckels:
https://github.com/LabKey/platform/pull/7001

PR #6975, opened 2025-08-29
Entropy score only the first 100 characters of password

labkey-adam:
#### Rationale
We think a recent labkey.org 504 could have been caused by entropy scoring an excessively long password. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53814

We don't have a repro, but scoring only the first 100 characters seems like a reasonable precaution to limit DOS attacks.

### DIFF
--- a/api/src/org/labkey/api/security/EntropyPasswordValidator.java
+++ b/api/src/org/labkey/api/security/EntropyPasswordValidator.java
@@ -54,7 +54,9 @@ public abstract class EntropyPasswordValidator implements PasswordValidator
 
     public static double score(String password, User user)
     {
-        return calculateEntropy(filter(password.toCharArray(), user));
+        // Score only the first 100 characters, Issue 53814
+        String shorter = password.substring(0, 100);
+        return calculateEntropy(filter(shorter.toCharArray(), user));
     }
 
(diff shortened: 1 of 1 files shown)


api/src/org/labkey/api/security/EntropyPasswordValidator.java
```diff
@@ -54,7 +54,9 @@ public boolean shouldShowPasswordGuidance()
 
     public static double score(String password, User user)
     {
-        return calculateEntropy(filter(password.toCharArray(), user));
+        // Score only the first 100 characters, Issue 53814
+        String shorter = password.substring(0, 100);
```
labkey-jeckels: This throws StringOutOfBoundsException when the length < 100 chars.


```suggestion
        String shorter = password.substring(0, Math.min(password.length(), 100));
```

labkey-jeckels (approved):
Need to support "short" passwords too

labkey-jeckels:
I can't reproduce the perf problem locally either.

api/src/org/labkey/api/security/EntropyPasswordValidator.java
labkey-adam: You don't want to encourage long (> 100 char) passwords? Well, okay.

PR #6977, opened 2025-08-30
Adjust TikaTestCase expectations for recent Tika versions

labkey-adam:
#### Rationale
Text extracted from a couple of our test files shows small whitespace differences as of a recent Tika version (not sure which one); adjust test expectations accordingly. Also, run the test with `strict=true` going forward, so it fails when these expectations aren't met, which was the intent. The `false` setting is helpful when adjusting a large number of discrepancies, but not what we want on an on-going basis.

### DIFF
--- a/search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
+++ b/search/src/org/labkey/search/model/LuceneSearchServiceImpl.java
@@ -1881,7 +1881,7 @@ public class LuceneSearchServiceImpl extends AbstractSearchService implements Se
         // expectations are invalid (e.g., expecting body length > 0 but specifying no substrings to search for).
         public void testTikaParsing() throws IOException, TikaException, SAXException
         {
-            boolean strict = false;
+            boolean strict = true;
 
             File sampledata = JunitUtil.getSampleData(null, "fileTypes");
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
I haven't tried running this with the Tika update, but take your word that it will pass.

labkey-adam:
> I haven't tried running this with the Tika update, but take your word that it will pass.

FYI: I noticed these discrepancies while testing 3.2.2, but turns out they aren't new. They show up on develop (3.2.1), but don't fail there because `strict=false`. So, this can be merged to develop prior to the merge forward.

labkey-jeckels:
Sure en

> > I haven't tried running this with the Tika update, but take your word that it will pass.
> 
> FYI: I noticed these discrepancies while testing 3.2.2, but turns out they aren't new. They show up on develop (3.2.1), but don't fail there because `strict=false`. So, this can be merged to develop prior to the merge forward.

Sure enough. https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_Bvt/3646866?buildTab=tests&name=Tika&expandedTest=build%3A%28id%3A3646865%29%2Cid%3A2000000046

PR #6979, opened 2025-08-31
Fix duplicate gpat assay name check

XingY:
#### Rationale
The recent change to assay design duplicate name check expanded the check to non-gpat assay designs.  
[BiologicsAPITest, BiologicsEvalTest, BiologicsRemoteEvalTest](https://teamcity.labkey.org/buildConfiguration/LabKey_HostingAndEvaluationInfrastructure_AutomatedTest_AutomatedTestForTrunk_TrialBiologicsIntegrationTest/3643542?buildTab=tests) failure during test setup with the complaint that "Assay design with name 'SEC' already exists." This test is failing at a step to create a workflow task protocol. The task has a name "SEC", which is the same as an existing gpat assay design. This should be allowed. 
[FlowFolderReimportTest.testReimportNestedWorkspaceFolder](https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyCPostgres/3643490?buildTab=tests&status=failed&name=FlowFolderReimportTest.testReimportNestedWorkspaceFolder): this test is failing when trying to creating a Flow assay design in child folder, when the project folder already has a Flow. This should be allowed.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1632

#### Changes
- Only check for duplicate if current and new protocol are of GPAT type

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
+++ b/experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
@@ -296,6 +296,7 @@ import static java.util.Collections.emptySet;
 import static java.util.Collections.singleton;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
+import static org.labkey.api.assay.AbstractTsvAssayProvider.GPAT_PROTOCOL_LSID_PREFIX;
 import static org.labkey.api.data.CompareType.IN;
 import static org.labkey.api.data.DbScope.CommitTaskOption.POSTCOMMIT;
 import static org.labkey.api.data.DbScope.CommitTaskOption.POSTROLLBACK;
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/api/ExperimentServiceImpl.java
```diff
@@ -6469,8 +6470,12 @@ public Protocol saveProtocol(
             if (newProtocol)
             {
                 // if protocol exists, throw error
-                if (AssayService.get().getAssayProtocolByName(protocol.getContainer(), protocol.getName()) != null)
-                    throw new RuntimeSQLException(new SQLException("Assay design with name '" + protocol.getName() + "' already exists."));
+                if (GPAT_PROTOCOL_LSID_PREFIX.equals(protocol.getLSIDNamespacePrefix()) && AssayService.get() != null)
```
labkey-nicka, line 6477: nit: There are 3 or 4 other unchecked calls to `AssayService.get().` in this service implementation. Consider revising those.

Alternatively, (in the future) we could consider a no-op service implementation for assay since I'm guessing this pattern is in wide use and unchecked.

XingY: Added more null check.

labkey-susanh (approved):

PR #6994, opened 2025-09-02
Update Amounts and Units for better sorting and filtering behavior

labkey-susanh:
#### Rationale
Our original implementation for the StoredAmount and Units columns added display columns over the user-provided data so the application grid displays used consistent units, but left the values as provided by users. Since we accept values in different, compatible units, this results in sorting and filtering on these columns not behaving as one would expect (e.g, samples with amounts of 2 mL and 2 L will be sorted next to each other, though displayed as 2 mL and 2000 mL in the grid).

This work update our handling of `StoredAmount` and `Units` so they always store values in the base units for sample types that have display units. This provides us with a valid basis for sorting and filtering. 

We also are updating the processing of data to always require that the user provides both StoredAmount and Units when either of these values is provided.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6994 
- https://github.com/LabKey/limsModules/pull/1653
- https://github.com/LabKey/testAutomation/pull/2678
- https://github.com/LabKey/labkey-ui-components/pull/1850 
- https://github.com/LabKey/labkey-ui-premium/pull/813 

#### Changes
- Upgrade script to convert existing amounts and units to base units, and drop any units that don't have values
- Remove `Measurement` class in favor of `Quantity` and `Units` classes
- Update detailed audit log to handle provided values that are converted before being stored
- Update aliquot rollup logic to store values in base units
- Remove recompute of rollup values when display unit is changed
- Add `SampleTypeAmountDisplayColumn` and `SampleTyepUnitDisplayColumn` to display amount and units using chosen display units


experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -153,4 +186,166 @@ private static void ensureBigObjectIds(Selector lastValueSelector, Consumer<Long
...
+        if (materialUnit == null && !StringUtils.isEmpty(unitsStr))
+        {
+            // invalid unit stored with sample. Leave as is.
+            LOG.info("Found invalid {} '{}' for sample '{}'. No conversion done.", isAliquot ? "aliquot unit" : "unit", unitsStr, sampleMap.get(Name.name()));
+            sampleCounts.put("invalidUnits", sampleCounts.getOrDefault("invalidUnits", 0) + 1);
+        }
+        else if (materialUnit != null && !materialUnit.isCompatible(baseUnit))
+        {
+            LOG.info("{} '{}' for sample '{}' is not compatible with the base unit '{}'. No conversion done.", isAliquot ? "Aliquot unit" : "Unit", materialUnit.name(), sampleMap.get(Name.name()), baseUnit);
+            sampleCounts.put("invalidUnits", sampleCounts.getOrDefault("invalidUnits", 0) + 1);
```
labkey-susanh, line 245: N.B. the capturing and logging of these counts will either get removed or have log level changed to DEBUG.

api/src/org/labkey/api/dataiterator/DetailedAuditLogDataIterator.java
```diff
@@ -53,12 +55,14 @@ public enum AuditConfigs {
     final QueryService.AuditAction _auditAction;
     final AuditHandler _auditHandler;
     final boolean _useTransactionAuditCache;
+    Function<Map<String, Object>, Map<String, Object>> _extractProvidedValues;
 
     // for batching
     final ArrayList<Map<String,Object>> _updatedRows = new ArrayList<>();
+    final ArrayList<Map<String, Object>> _providedValues = new ArrayList<>(); // TODO
```
XingY: Still TODO?

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -153,4 +186,166 @@ private static void ensureBigObjectIds(Selector lastValueSelector, Consumer<Long
...
+
+    // Converts amounts for all sample types defined in the given container.
+    // Picks up samples from all containers for each sample type.
+    private static void convertAmountsToBaseUnits(Container container, User user, TransactionAuditProvider.TransactionAuditEvent transactionEvent)
+    {
+        DbScope scope = ExperimentService.get().getSchema().getScope();
+        TableInfo tInfo = ExperimentService.get().getTinfoMaterial();
+        for (ExpSampleType sampleType :SampleTypeService.get().getSampleTypes(container, user, false))
+        {
+            try (DbScope.Transaction transaction = scope.ensureTransaction())
```
XingY: Why one transaction per container? If some container succeeded upgrade but some failed, won't that leave DB in a mixed state?

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -153,4 +186,166 @@ private static void ensureBigObjectIds(Selector lastValueSelector, Consumer<Long
...
+
+                Unit currentDisplayUnit = Unit.fromName(sampleType.getMetricUnit());
+                boolean hasDisplayUnit = currentDisplayUnit != null;
+
+                SQLFragment sql = new SQLFragment("SELECT m.RowId, m.Name, m.StoredAmount, m.Units, m.AliquotVolume, m.AliquotUnit, m.AvailableAliquotVolume FROM ")
+                        .append(tInfo, "m")
+                        .append(" WHERE cpastype = ?").add(sampleType.getLSID());
+                SqlSelector selector = new SqlSelector(scope, sql);
+                selector.mapStream().forEach(sampleMap -> {
+                    Map<String, Object> oldDataMap = new HashMap<>();
```
XingY, line 321: Could we use a prepared statement to do batch update?

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
```diff
@@ -719,6 +721,11 @@ protected Map<String, Object> coerceTypes(Map<String, Object> row)
                 {
                     if (PropertyType.FILE_LINK.equals(col.getPropertyType()))
                         value = ExpDataFileConverter.convert(value);
+                    else if (col.getKindOfQuantity() != null)
```
XingY, line 724: I thought with this story getKindOfQuantity() / getKindOfQuantity is now stale and show be removed? It's not clear to me how sample amount / unit is wired up to KindOfQuantity / Unit, seems it's just a one off hard wire?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -557,6 +619,91 @@ public List<Map<String, Object>> updateRows(User user, Container container, List
...
+                    !(value instanceof MultipartFile) &&
+                    !(value instanceof String[]) &&
+                    !(col.getFk() instanceof MultiValuedForeignKey))
+            {
+                try
+                {
+                    if (PropertyType.FILE_LINK.equals(col.getPropertyType()))
+                        value = ExpDataFileConverter.convert(value);
+                    else if (col.getKindOfQuantity() != null)
+                    {
```
XingY, line 687: I would think this can already handle amount/unit columns in the super class?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -557,6 +619,91 @@ public List<Map<String, Object>> updateRows(User user, Container container, List
...
+    @Override
+    protected Map<String, Object> coerceTypes(Map<String, Object> row, Map<String, Object> providedValues, boolean isUpdate)
+    {
+        Map<String, Object> result = new CaseInsensitiveHashMap<>(row.size());
+        Map<String, ColumnInfo> columnMap = ImportAliasable.Helper.createImportMap(_queryTable.getColumns(), true);
+        Object unitsVal = null;
+        ColumnInfo unitsCol = null;
+        Object amountVal = null;
+        ColumnInfo amountCol = null;
+        if (row.containsKey(ExpMaterialTable.Column.Units.name()))
```
XingY, line 636: Not a fan of this overrride of deprecated method. Is there a way to have this in the super method?

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -1922,43 +2077,109 @@ private void _addConvertColumn(ColumnInfo col, int fromIndex, int derivationData
 
         protected class SampleUnitsConvertColumn extends SimpleTranslator.SimpleConvertColumn
```
XingY, line 2090: SampleUnitsConvertColumn and SampleAmountConvertColumn seems generic enough to be used for all volume/amount columns. Can they be moved to SimpleTranslator?

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -153,4 +186,166 @@ private static void ensureBigObjectIds(Selector lastValueSelector, Consumer<Long
...
+        }
+    }
+
+    // Converts amounts for all sample types defined in the given container.
+    // Picks up samples from all containers for each sample type.
+    private static void convertAmountsToBaseUnits(Container container, User user, TransactionAuditProvider.TransactionAuditEvent transactionEvent)
+    {
+        DbScope scope = ExperimentService.get().getSchema().getScope();
+        TableInfo tInfo = ExperimentService.get().getTinfoMaterial();
+        for (ExpSampleType sampleType :SampleTypeService.get().getSampleTypes(container, user, false))
```
XingY: Do we want to exclude specimen from this upgrade?

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -153,4 +186,166 @@ private static void ensureBigObjectIds(Selector lastValueSelector, Consumer<Long
...
+                    Map<String, Object> oldDataMap = new HashMap<>();
+                    Map<String, Object> newDataMap = new HashMap<>();
+                    if (!StringUtils.isEmpty((String) sampleMap.get(Units.name())) && sampleMap.get(StoredAmount.name()) == null)
+                    {
+                        // remove the unit if we had a unit but no amount
+                        oldDataMap.put(Units.name(), sampleMap.get(Units.name()));
+                        newDataMap.put(Units.name(), null);
+                        sampleCounts.put("unitsWithoutAmounts", sampleCounts.getOrDefault("unitsWithoutAmounts", 0) + 1);
+                    }
+                    if (!StringUtils.isEmpty((String) sampleMap.get(AliquotUnit.name())) && sampleMap.get(AliquotVolume.name()) == null && sampleMap.get(AvailableAliquotVolume.name()) == null)
```
XingY, line 338: I think we just need to check for AliquotVolume. If AvailableAliquotVolume is not null, then AliquotVolume is for sure not null.

XingY: But should we check for null, or >0 ? all samples (not aliquot) has not null AliquotVolume. It's initialized to 0.

XingY (commented):
My question is, how is this worked tied to the previous work done to support ColumnRenderProperties.getKindOfQuantity

labkey-susanh:
> My question is, how is this worked tied to the previous work done to support ColumnRenderProperties.getKindOfQuantity

As mentioned in standup, it's not really related to this other than using the `Quantity` class instead of `Measurement`. We are choosing to defer the single-column for a Quantity type field work until sometime later.

api/src/org/labkey/api/dataiterator/DetailedAuditLogDataIterator.java
labkey-susanh: Nope. Thanks.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
labkey-susanh: Moved. This was largely due to an initial plan to have another pathway to use this code but that turned out to not be necessary.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
labkey-susanh: Yes. As discussed, I was wondering about the complications involved with different statements being needed depending on what was to be updated (I just opted for adding some no-op updates when fields are not changing) and the audit events that are necessary to be paired with each update statement, but there is already a method for prepared statements for audit events, so this route seems easier than originally thought.  Thanks.

api/src/org/labkey/api/query/AbstractQueryUpdateService.java
labkey-susanh: Sample Amount is not wired up to this at the moment, so really this code path won't get used yet.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
labkey-susanh: This is just copying the logic from the overridden method.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
labkey-susanh: Not a fan either, but since this fix up is very specific to these two separate columns that are currently tied together, moving anything into the other method would be more messy.  Consider this transitional code that will go away either if we get rid of the need for the `coerceTypes` method used for one-at-a-time updates or we get to do more work so the `StoredAmount` field becomes a `Quantity`-type field and we then know how to deal with the amount and units together there.

experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
labkey-susanh: Only sample types will have these columns. If we get to adding Quantity fields in other domains, we likely won't have a need for these convert columns (or they will be of a different nature).

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
labkey-susanh: I don't think it's necessary to exclude them. Since they don't have display units, very little will happen to the data.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
labkey-susanh: I would prefer to be more cautious with the checks here to be more sure of a consistent outcome. Even if the amount is 0, we may need to adjust the units value.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
XingY: But it would make the upgrade faster. The query below is not cheap, and even though it's iterated as stream, it's still a lot of processing that could have been skipped.
`new SQLFragment("SELECT m.RowId, m.Name, m.StoredAmount, m.Units, m.AliquotVolume, m.AliquotUnit, m.AvailableAliquotVolume FROM ")
                        .append(tInfo, "m")
                        .append(" WHERE cpastype = ?").add(sampleType.getLSID());`

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -153,4 +194,215 @@ private static void ensureBigObjectIds(Selector lastValueSelector, Consumer<Long
...
+                        }
+                        // for rows with an amount and a unit when there is no display unit, no conversion is done
+                    }
+
+
+                    if (!newDataMap.isEmpty())
+                    {
+                        updateStmt.addBatch();
+                        batchCount.getAndIncrement();
+                        SampleTimelineAuditEvent event = new SampleTimelineAuditEvent(container, "Storage amount unit conversion to base unit during upgrade script.");
```
XingY: Shouldn't the audit event be created at sample's container, rather than sample type's container?

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -153,4 +194,215 @@ private static void ensureBigObjectIds(Selector lastValueSelector, Consumer<Long
...
+            Parameter aliquotUnits = new Parameter(AliquotUnit.name(), JdbcType.VARCHAR);
+            Parameter aliquotAmount = new Parameter(AliquotVolume.name(), JdbcType.DOUBLE);
+            Parameter availableAliquotAmount = new Parameter(AvailableAliquotVolume.name(), JdbcType.DOUBLE);
+            ParameterMapStatement updateStmt = new ParameterMapStatement(scope, c,
+                    new SQLFragment("UPDATE ")
+                            .append(tInfo)
+                            .append(" SET Units = ?, StoredAmount = ?, AliquotUnit = ?, AliquotVolume = ?, AvailableAliquotVolume = ? WHERE RowId = ?")
+                            .addAll(units, amount, aliquotUnits, aliquotAmount, availableAliquotAmount, rowId), null);
+
+            for (ExpSampleType sampleType : SampleTypeService.get().getSampleTypes(container, user, false))
```
XingY, line 302: Could we further optimize so we only process the sample type if either exist storedamount is not null OR aliquotvolume is not null?

XingY (commented):
upgrade script changes looks good, with just some further minor optimization comment.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
labkey-susanh: Yes. Thanks for that.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
labkey-susanh: Instead of doing any specialized treatment for specimen samples, I've updated the select statement to be more selective, excluding rows that will not need to be updated, which will probably eliminate further processing for specimen samples.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
labkey-susanh: I don't think we can do anything at the sample type level, but I've updated the select statement for the samples in the sample type to exclude rows that won't need to be updated because they have neither amounts nor units.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -153,4 +188,222 @@ private static void ensureBigObjectIds(Selector lastValueSelector, Consumer<Long
...
+
+                Unit currentDisplayUnit = Unit.fromName(sampleType.getMetricUnit());
+                boolean hasDisplayUnit = currentDisplayUnit != null;
+
+                AtomicInteger batchCount = new AtomicInteger();
+                List<AuditTypeEvent> auditEvents = new ArrayList<>();
+                SQLFragment sql = new SQLFragment("SELECT m.RowId, m.Name, m.StoredAmount, m.Units, m.AliquotVolume, m.AliquotUnit, m.AvailableAliquotVolume, m.container FROM ")
+                        .append(tInfo, "m")
+                        .append(" WHERE cpastype = ?").add(sampleType.getLSID())
+                        .append(" AND (m.StoredAmount IS NOT NULL OR m.Units IS NOT NULL OR m.AliquotVolume IS NOT NULL OR m.AliquotUnit IS NOT NULL OR m.AvailableAliquotVolume IS NOT NULL)");
```
XingY, line 316: OR is generally slow. Could we at least get rid of AvailableAliquotVolume check? If AvailableAliquotVolume is not null, AliquotVolume is for sure not null, unless there is an App bug.

labkey-susanh: We can certainly. Do you have any sense of how much difference the extra condition might make in terms of performance?  Yes, there are other problems if the sample row is in the state where AvailableAliquotVolume is the only thing that is not null in this set of fields, but if we don't do the update this might be a case of compounding an app bug. Again, I'm inclined to be conservative, particularly since this is upgrade code not run-all-the-time code.

XingY: I remember in the past having to switch OR to UNION for performance reasons... And was advised to generally prefer UNION rather than OR. I don't know how to quantify the difference though.

labkey-susanh: @labkey-matthewb Any comment here?

labkey-matthewb: OR is especially bad when it is used across different tables e.g. t1.col IS NOT NULL OR t2.col IS NOT NULL.  In this case where we are probably just scanning the entire table anyway (and just once for upgrade), I think it probably is not a concern.  (expression evaluation is fast, cross-product join is very, very bad)

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -153,4 +188,222 @@ private static void ensureBigObjectIds(Selector lastValueSelector, Consumer<Long
...
+                            event.setLineageUpdate(false);
+                            event.setOldRecordMap(AbstractAuditTypeProvider.encodeForDataMap(oldDataMap));
+                            event.setNewRecordMap(AbstractAuditTypeProvider.encodeForDataMap(newDataMap));
+                            auditEvents.add(event);
+                        }
+                    }
+                    if (batchCount.get() > 1000)
+                    {
+                        updateStmt.executeBatch();
+                        AuditLogService.get().addEvents(user, auditEvents, false);
```
XingY, line 385: So those events aren't really processed in a batch way, because AuditLogImpl._eventTypeQueue flattens the list to single a time. I think _eventTypeQueue should be updated to Queue<Pair<User, List<AuditTypeEvent>>>

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -153,4 +188,202 @@ private static void ensureBigObjectIds(Selector lastValueSelector, Consumer<Long
...
+                            .addAll(units, amount, aliquotUnits, aliquotAmount, availableAliquotAmount, rowId), null);
+
+            for (ExpSampleType sampleType : SampleTypeService.get().getSampleTypes(container, user, false))
+            {
+                LOG.info("** Starting upgrade for sample type {} in folder {}", sampleType.getName(), container.getPath());
+                Map<String, Integer> sampleCounts = new HashMap<>();
+                Map<String, Integer> aliquotCounts = new HashMap<>();
+
+                Unit currentDisplayUnit = Unit.fromName(sampleType.getMetricUnit());
+                boolean hasDisplayUnit = currentDisplayUnit != null;
```
XingY, line 292: Looks like we can just skip the rest if hasDisplayUnit is false

experiment/src/org/labkey/experiment/ExperimentWarningProvider.java
```diff
@@ -0,0 +1,58 @@
...
+import org.labkey.api.security.User;
+import org.labkey.api.util.HtmlString;
+import org.labkey.api.view.ViewContext;
+import org.labkey.api.view.template.WarningProvider;
+import org.labkey.api.view.template.Warnings;
+
+public class ExperimentWarningProvider implements WarningProvider
+{
+    @Override
+    public void addDynamicWarnings(@NotNull Warnings warnings, @Nullable ViewContext context, boolean showAllWarnings)
```
labkey-jeckels, line 27: I believe we'll call this method on every LKS page load for admins, so we should cache the message to avoid requerying (likely very large) audit table when there is a mismatch.

experiment/src/org/labkey/experiment/ExperimentWarningProvider.java
```diff
@@ -0,0 +1,58 @@
...
+    {
+        PropertyManager.WritablePropertyMap props = PropertyManager.getWritableProperties(ExperimentModule.AMOUNT_AND_UNIT_UPGRADE_PROP, false);
+        if (props == null || props.isEmpty())
+            return;
+        String expectedCount = props.get("AuditRecordCount");
+        String transactionIdStr = props.get("AuditTransactionId");
```
labkey-jeckels: Could use constants for these property names.

api/src/org/labkey/api/ontology/Unit.java
```diff
@@ -176,25 +185,56 @@ static Function<Double,Double> convertFn(Unit from, Unit to)
...
     {
         return Quantity.convert(value, this);
     }
 
+    public static Unit getValidatedUnit(@Nullable Object rawUnits, @Nullable Unit defaultUnits)
+    {
+        if (rawUnits == null)
+            return null;
+        if (rawUnits instanceof Unit u)
+            if (defaultUnits == null)
```
XingY, line 207: Can we add the outer braces to the block of if/else if/else ? It's confusing to read and error prone.

experiment/src/org/labkey/experiment/ExperimentWarningProvider.java
XingY: One idea is to store an additional ExperimentModule.ACTUAL_AUDIT_COUNT_PROP in ExperimentModule.AMOUNT_AND_UNIT_UPGRADE_PROP property map.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
XingY: Is it time to change those log level to DEBUG?

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
```diff
@@ -153,4 +192,218 @@ private static void ensureBigObjectIds(Selector lastValueSelector, Consumer<Long
...
+            ContainerManager.getAllChildren(ContainerManager.getRoot()).forEach(c ->
+                    auditCount.addAndGet(convertAmountsToBaseUnits(c, admin))
+            );
+            transaction.commit();
+            LOG.info("{} Total audit events expected", auditCount);
+            if (auditCount.get() > 0)
+            {
+                PropertyManager.WritablePropertyMap props = PropertyManager.getWritableProperties(AMOUNT_AND_UNIT_UPGRADE_PROP, true);
+                props.put(AUDIT_COUNT_PROP, auditCount.toString());
+                props.put(TRANSACTION_ID_PROP, transactionId == null ? null : transactionId.toString());
```
XingY: If transactionId is null, do we still want to set property map?

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
```diff
@@ -1542,37 +1538,31 @@ private Set<Long> getRootSampleIdsFromParentNames(String sampleTypeLsid, Set<Str
...
-        // if sample unit is empty, use 1st aliquot unit
-        if (StringUtils.isEmpty(totalDisplayUnitStr))
-        {
-            String aliquotUnit = volumeUnits.get(0).unit;
-            if (!StringUtils.isEmpty(aliquotUnit))
-                totalDisplayUnitStr = aliquotUnit;
-        }
-
-        if (!StringUtils.isEmpty(totalDisplayUnitStr))
+        Unit totalUnit = null;
```
XingY, line 1546: I feel we are not taking advantage of the fact now all aliquots are now stored in the same base "units", if sample type has a unit set. Rollup should now be much simpler (and faster) for sample types with unit, potentially could be done with a simple sql script.

api/src/org/labkey/api/ontology/Unit.java
labkey-susanh: Sure. Done.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
labkey-susanh: It can't really be null unless my read of the logic is incorrect. This null check just to appease the IntelliJ warnings.

experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
labkey-susanh: Yes. Done.

experiment/src/org/labkey/experiment/ExperimentWarningProvider.java
labkey-susanh: Storing the count as a class member seems to work.

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
labkey-susanh: Yeah, I see what you're saying, but I'm going to not make this change now. We can make this optimization in a separate, smaller PR so we can concentrate on assuring no regressions with the rollup vs. the unit conversion code.

labkey-susanh: Can you open an issue for this so we don't forget?

experiment/src/org/labkey/experiment/ExperimentWarningProvider.java
labkey-susanh: Done

experiment/src/org/labkey/experiment/api/SampleTypeServiceImpl.java
XingY: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53932

XingY (approved):

PR #6951, opened 2025-08-21
Allow SSO, password change, etc. in locked projects

labkey-adam:
#### Rationale
Login from a locked project was failing to display SSO logos. The change password workflow was also broken. And TOTP. And password strength. And user profile updates. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53750

Many `LoginController` actions need to ignore the locked project check.

#### Related Pull Requests
- https://github.com/LabKey/premiumModules/pull/344

#### Tasks 📍
- [x] Allow key login controller actions on locked project
- [x] Fix user profile update form in project with no permissions @labkey-klum 
- [x] Manual Testing / Verify Fix @labkey-tchad 
  - All operations that occur as a part of login (or really prior to reaching the requested page) are worth testing in a locked project as non-admin. That includes:
    - SSO logos/redirect/validate
    - Change password workflow (e.g., set password strength to "good", create "good" password, then set strength to "strong" and attempt to sign in)
    - Forgot password
    - Password strength indicator
    - User profile (set some user fields as required)
    - Terms of use (as part of login and project terms that must be accepted even for users who are logged in)
    - MFA (TOTP, Duo)
- [x] Rewrite locked project permission checking
   - Check a specific permission (`CanAccessLockedProjectsPermission`) that's assigned to admins and can be set as a contextual role
   - Pass contextual roles into locked project permission checking
   - Rewrite permission checking so it calls into a mainstream `SecurityManager.getPermissions()` method instead of attempting to duplicate it (poorly)
- [x] Fix user profile update by setting the above permission as a contextual role on the button bar
- [x] Re-test previously failing manual test cases @labkey-tchad 
- [x] Needs Automation ? - No

PR #7006, opened 2025-09-03
Fix locked project permission

labkey-adam:
#### Rationale
`StudySecurityTest` started failing when I registered my new `CanAccessLockedProjectsPermission` as a role. This role/permission is used only in code so we can designate it as never "applicable"; that appeases the dataset security UI.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6951

### DIFF
--- a/api/src/org/labkey/api/security/permissions/CanAccessLockedProjectsPermission.java
+++ b/api/src/org/labkey/api/security/permissions/CanAccessLockedProjectsPermission.java
@@ -12,6 +12,7 @@ public class CanAccessLockedProjectsPermission extends AbstractPermission
         super("Can Access Locked Projects", "Allows admins to access locked projects");
     }
 
+    // This prevents the dataset security UI (datasets.jsp) from attempting to examine this role
     @Override
     public boolean isApplicable(SecurityPolicy policy, SecurableResource resource)
     {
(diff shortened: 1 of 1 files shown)


labkey-adam:
StudySecurityTest passed: https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailySuites_DailyEPostgres/3651519

api/src/org/labkey/api/security/permissions/CanAccessLockedProjectsPermission.java
```diff
@@ -8,4 +11,10 @@ public CanAccessLockedProjectsPermission()
     {
         super("Can Access Locked Projects", "Allows admins to access locked projects");
     }
+
+    @Override
+    public boolean isApplicable(SecurityPolicy policy, SecurableResource resource)
+    {
+        return false;
```
labkey-jeckels, line 19: Consider a comment that this prevents it from showing up in the dataset security UI or other places we don't want to see it.

labkey-jeckels (approved):

api/src/org/labkey/api/security/permissions/CanAccessLockedProjectsPermission.java
labkey-adam: Done

PR #6989, opened 2025-09-02
Report duplicate and overlapping database indices

labkey-adam:
#### Rationale
New devtools action inspects all indexes in tables in module schemas and flags those with duplicate or overlapping column sets. In most (all?) cases these duplicates are unnecessary, causing bloat and perhaps inefficiency.

Move standard redirect methods and callers away from passing URL strings.

#### Tasks 📍
- [x] Manual Testing @labkey-danield 
  - With the devtools module present, visit tools-begin.view (e.g., http://localhost:8080/home/tools-begin.view)
  - Click the `overlappingIndices` link
  - Observe and lament how many overlapping indices we have

PR #7016, opened 2025-09-05
Drop overlapping indices

labkey-adam:
#### Rationale
Revise the tool that detects redundant database indices to produce SQL scripts with drop commands.

Update our table/column metadata handling to read a table's PK name and use that to determine the index corresponding to the primary key. Previously, the "get index" code path guessed the PK based on the column list, but that incorrectly tagged redundant indices as PKs.

Drop redundant indices from most platform schemas.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/6989

- [x] Manual test @labkey-jeckels


labkey-jeckels (approved):
I gave this some testing locally. I didn't confirm that every index it wanted to delete was correct, but all of the ones I spot checked were correct.

PR #7017, opened 2025-09-05
Issue 53825: LKSM/LKB: Sample Timeline entries for lineage updates with domains containing & or . may be incorrect

XingY:
#### Rationale
When building detailed audit, it's assumed existingRow contains all relevant field, which is not true for the scenario when a new type is added as parent. Newly established parent type fields only exist in the new incoming row, but absent from the original data row. A addDetailedModifiedFields util is added for that scenario to account for new parents. This PR fixes 2 issues related to that util:
1. this is only done for samples, and not dataclass
2. the comparison of field doesn't take encoding difference into consieration

This is revealed by TC failure:
[SMSampleTimelineLineageTest.testAddAndDeleteWithMultipleParents](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SampleManagerGPostgres/3653532?buildTab=tests&status=failed&name=SMSampleTimelineLineageTest.testAddAndDeleteWithMultipleParents)
Lineage update again shows parents of [NA]
The updated fields in the event detail are not as expected.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7005

#### Changes
- check for encoded key when determine if need to add to detailed change list
- move util to AuditHandler to support dataclass audit detail.

### DIFF
--- a/api/src/org/labkey/api/audit/AuditHandler.java
+++ b/api/src/org/labkey/api/audit/AuditHandler.java
@@ -98,8 +98,11 @@ public interface AuditHandler
             String encodedInputColumn = ExperimentService.getEncodedLineageKey(lcName);
             if (encodedInputColumn != null)
             {
-                isExpInput = true;
-                nameFromAlias = encodedInputColumn;
+                if (row.containsKey(encodedInputColumn))
+                {
+                    isExpInput = true;
+                    nameFromAlias = encodedInputColumn;
+                }
             }
             else
                 nameFromAlias = lcName;
--- a/api/src/org/labkey/api/audit/AuditHandler.java
+++ b/api/src/org/labkey/api/audit/AuditHandler.java
@@ -95,17 +95,11 @@ public interface AuditHandler
             String lcName = nameFromAlias.toLowerCase();
             // Preserve casing of inputs so we can show the names properly
             boolean isExpInput = false;
-            if (lcName.startsWith(ExpData.DATA_INPUTS_PREFIX.toLowerCase()) || lcName.startsWith(ExpMaterial.MATERIAL_INPUTS_PREFIX.toLowerCase()))
+            String encodedInputColumn = ExperimentService.getEncodedLineageKey(lcName);
+            if (encodedInputColumn != null)
             {
-                String[] parts = nameFromAlias.split("/", 2);
-                String prefix = parts[0];
-                String dataType = parts[1];
-                // MaterialInputs/datypeTypeEncoded
-                if (row.containsKey(prefix + "/" + QueryKey.encodePart(dataType)))
-                {
-                    isExpInput = true;
-                    nameFromAlias = prefix + "/" + QueryKey.encodePart(dataType);
-                }
+                isExpInput = true;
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/audit/AuditHandler.java
```diff
@@ -177,6 +178,32 @@ else if (isExtraAuditField)
                 modifiedRow.put(nameFromAlias, entry.getValue());
             }
         }
+
+        // we want to include the fields that indicate parent lineage has changed.
+        // Note that we don't need to check for output fields because lineage can be modified only by changing inputs not outputs
+        Set<String> existingEncodedInputColumns = new CaseInsensitiveHashSet();
+        for (String fieldName : existingRow.keySet())
+        {
+            if (fieldName.toLowerCase().startsWith(ExpData.DATA_INPUTS_PREFIX.toLowerCase()) || fieldName.toLowerCase().startsWith(ExpMaterial.MATERIAL_INPUTS_PREFIX.toLowerCase()))
```
labkey-susanh: Could use `MATERIAL_INPUTS_PREFIX_LC` here instead. Do we not have a `isLineageInput` helper method? Perhaps we should.

api/src/org/labkey/api/audit/AuditHandler.java
```diff
@@ -177,6 +178,32 @@ else if (isExtraAuditField)
                 modifiedRow.put(nameFromAlias, entry.getValue());
             }
         }
+
+        // we want to include the fields that indicate parent lineage has changed.
+        // Note that we don't need to check for output fields because lineage can be modified only by changing inputs not outputs
+        Set<String> existingEncodedInputColumns = new CaseInsensitiveHashSet();
```
labkey-susanh, line 181: I don't love that we have more exp-specific stuff in here, but there is precedence, I guess.

api/src/org/labkey/api/audit/AuditHandler.java
```diff
@@ -177,6 +178,32 @@ else if (isExtraAuditField)
...
+        {
+            if (fieldName.toLowerCase().startsWith(ExpData.DATA_INPUTS_PREFIX.toLowerCase()) || fieldName.toLowerCase().startsWith(ExpMaterial.MATERIAL_INPUTS_PREFIX.toLowerCase()))
+            {
+                // Issue 53825: LKSM/LKB: Sample Timeline entries for lineage updates with domains containing & or . may be incorrect
+                String[] parts = fieldName.split("/", 2);
+                if (parts.length == 2)
+                {
+                    String prefix = parts[0];
+                    String dataType = parts[1];
+                    existingEncodedInputColumns.add(prefix + "/" + QueryKey.encodePart(dataType));
```
labkey-susanh: We're doing this splitting to encode logic in a couple of places now, so a small utility method (getEncodedLineageKey?) seems like it would be helpful for readability at least.

labkey-susanh (approved):
Some suggestions for code reuse/readability, but no need for re-review.

Verified fix with a bit of manual testing.

PR #6990, opened 2025-09-02
Deprecate generating controller-first URLs

labkey-adam:
#### Rationale
Path-first URLs have been the default for many, many years. It's time to remove the option that generates URLs in the legacy controller-first format. https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53830

The server normally accepts URLs in the legacy format. However, this PR also adds an experimental feature to reject all controller-first URLs, which should make it easier to find non-standard URLs used in our products and tests. 

#### Tasks 📍
- [x] Manual Testing @labkey-tchad 
- [x] Needs Automation - No

PR #7022, opened 2025-09-08
Modernize some controller-first URLs to support basic functionality

labkey-tchad:
#### Rationale
Running with the `rejectControllerFirstUrls` revealed that we use controller-first URLs to fetch certain CSS and JS files and to log client exceptions. This breaks some pretty basic functionality (including the ability to disable experimental features).

#### Related Pull Requests
- #6990 

#### Changes
- Update some controller-first URLs to support basic functionality

### DIFF
--- a/core/src/org/labkey/core/webdav/davListing.jsp
+++ b/core/src/org/labkey/core/webdav/davListing.jsp
@@ -23,6 +23,8 @@
 <%@ page import="org.labkey.api.view.template.ClientDependencies" %>
 <%@ page import="org.labkey.api.webdav.WebdavResource" %>
 <%@ page import="org.labkey.core.webdav.DavController" %>
+<%@ page import="java.lang.Override" %>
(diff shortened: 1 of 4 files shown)


labkey-adam (approved):

core/src/org/labkey/core/webdav/davListing.jsp
```diff
@@ -48,14 +47,14 @@
         var loginAction = new Ext4.Action({
             text : 'Login',
             handler : function () {
-                window.location = LABKEY.contextPath + '/login/home/login.view?returnUrl=' + returnUrl;
+                window.location = LABKEY.contextPath + '/home/login-login.view?returnUrl=' + returnUrl;
```
labkey-nicka: nit: Ideally, these would both use `LABKEY.ActionURL.buildURL()` rather than stitching the path together themselves.

PR #7028, opened 2025-09-10
Fix domain change NPE

labkey-adam:
#### Rationale
SharedStudyTest was throwing an NPE

StudyDatasetIndexTest  was failing to create custom indices


labkey-klum (approved):
Also validated this against Dataspace.

labkey-adam:
> Also validated this against Dataspace.

Thanks, @labkey-klum ... Study PostgreSQL and CDS suites both look good (one unrelated failure in each, I think)

PR #7035, opened 2025-09-12
Don't need to check for SearchService; it falls back to a no-op now

labkey-jeckels:
#### Rationale
Recent work on the search queue introduced a no-op implementation for the service. Thus, we don't need to check if it's null anymore.

#### Changes
- Just use the available `SearchService` and let it do nothing if it chooses

#### Tasks 📍
- Manual Testing - N/A
- Needs Automation - N/A

### DIFF
--- a/announcements/src/org/labkey/announcements/AnnouncementModule.java
+++ b/announcements/src/org/labkey/announcements/AnnouncementModule.java
@@ -177,12 +177,8 @@ public class AnnouncementModule extends DefaultModule implements SearchService.D
         MessageConfigService.get().registerConfigType(new AnnouncementEmailConfig());
         
         SearchService ss = SearchService.get();
-
-        if (null != ss)
-        {
-            ss.addSearchCategory(AnnouncementManager.searchCategory);
-            ss.addDocumentProvider(this);
-        }
+        ss.addSearchCategory(AnnouncementManager.searchCategory);
+        ss.addDocumentProvider(this);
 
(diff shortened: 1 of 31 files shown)


labkey-adam (changes requested):
Many files here have whitespace changes on every line; I doubt that's what you want. Same thing happened to me a week ago... I don't know why. I created a new FB with the same patch and it produced the expected diff. 🤷

labkey-jeckels:
> Many files here have whitespace changes on every line; I doubt that's what you want. Same thing happened to me a week ago... I don't know why. I created a new FB with the same patch and it produced the expected diff. 🤷

Probably a line ending thing, possibly OSX vs Windows. Did you push your FB to GitHub? I'm not seeing it. I can kill this PR and use your branch instead.

labkey-adam:
> > Many files here have whitespace changes on every line; I doubt that's what you want. Same thing happened to me a week ago... I don't know why. I created a new FB with the same patch and it produced the expected diff. 🤷
> 
> Probably a line ending thing, possibly OSX vs Windows. Did you push your FB to GitHub? I'm not seeing it. I can kill this PR and use your branch instead.

Sorry for confusion... I created a new FB for MY change last week, not this one. I don't think Windows was involved in your PR... you can see the line endings changes via the GitHub Web UI. I'm wondering if IntelliJ is changing line endings in certain cases.

labkey-jeckels:
Fresh PR where GitHub didn't mangle the line endings. Not sure why it behaved differently here. https://github.com/LabKey/platform/pull/7038

PR #7040, opened 2025-09-15
Export full search index as text

labkey-chrisj:
#### Rationale
This change appends a brief check to the end of SearchTest, to export the full-text index and verify some specific items that should be indexed are present. Ideally it also ensures no exceptions in the log arise as a result of exporting the index, and it ensures the full-text log has at least 2k rows

Related issue https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51159

Related PR:
testAutomation helper to get file line count https://github.com/LabKey/testAutomation/pull/2689


labkey-adam (approved):

study/test/src/org/labkey/test/tests/search/SearchTest.java
```diff
@@ -226,14 +229,15 @@ private void addSearchableLists()
...
-    protected void doVerifySteps()
+    protected void doVerifySteps() throws IOException, CommandException
     {
         _searchHelper.verifySearchResults("/" + getProjectName() + "/" + getFolderName());
         testAdvancedSearchScope();
         testAdvancedSearchCategoryFilters();
         renameFolderAndReSearch();
         moveFolderAlterListsAndReSearch();
         deleteFolderAndVerifyNoResults();
+        exportSearchIndexAndVerifyResults();
```
labkey-tchad, line 239: This should check the search index before deleting the folder.

labkey-chrisj: Thanks, will do.

study/test/src/org/labkey/test/tests/search/SearchTest.java
```diff
@@ -226,13 +229,14 @@ private void addSearchableLists()
 
     @Override
     @LogMethod
-    protected void doVerifySteps()
+    protected void doVerifySteps() throws IOException, CommandException
```
labkey-danield: I don't think this throws a CommandException. exportSearchIndexAndVerifyResults throws only an IOException.

study/test/src/org/labkey/test/tests/search/SearchTest.java
```diff
@@ -176,7 +179,7 @@ private Set<String> getSearchResultsProjects(String q) throws IOException, Comma
     }
 
     @Test
-    public void testSearch()
+    public void testSearch() throws IOException, CommandException
```
labkey-danield: Similarly you shouldn't need a CommandException.

labkey-danield (commented):
Wouldn't hurt to run a couple of the search test on TeamCity to verify this works.

study/test/src/org/labkey/test/tests/search/SearchTest.java
```diff
@@ -457,6 +461,27 @@ private void deleteFolderAndVerifyNoResults()
...
+        var indexFile = clickAndWaitForDownload(exportTxtBtnLoc);
+
+        TextSearcher tsvSearcher = new TextSearcher(indexFile);
+        assertTextPresent(tsvSearcher, "Black Bear", "Owlbear","pdf_sample.pdf", "docx_sample.docx",
+                "InlineFile.html", "verifyAssay", "Roquefort", "Brie", "Study 001", "Folder Banana", "Sample",
+                "Urinalysis");
+        int fileRowCount = getFileRowCount(indexFile);
+        checker().wrapAssertion(() -> Assertions.assertThat(fileRowCount)
+                .as("expect minimum 2000 rows")
+                .isGreaterThan(2000));
```
labkey-danield: Where did this 2000 rows come from? 
What if this is the first test, or only test, to run?
If you got this number from a local run I would suggest that you run against a bootstrap database and compare.

study/test/src/org/labkey/test/tests/search/SearchTest.java
labkey-chrisj: OK, I'll clear those out

study/test/src/org/labkey/test/tests/search/SearchTest.java
labkey-chrisj: It was a wild guess, something to see what I'd get in a real-life run on TC
It turns out that my test runs have come back with results as low as ~359ish, I'll try a local bootstrap run as well

PR #7043, opened 2025-09-16
Prevent showing audit count warning if module startup is still in progress

labkey-susanh:
#### Rationale
If the "startup in progress" page refreshes while the system is in the process of adding audit logs for the upgrade script for amounts and units, it will stash the audit log count too early and show a warning before all logs have been created. This PR doesn't generate the warning until module startup is complete and thus all audit logs have been written.

#### Related Pull Requests
- #6994

#### Changes
- Add check for module startup complete

#### Tasks 📍
- [x] Manual Testing @labkey-jeckels

### DIFF
--- a/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
+++ b/experiment/src/org/labkey/experiment/ExperimentUpgradeCode.java
@@ -193,7 +193,8 @@ public class ExperimentUpgradeCode implements UpgradeCode
         }
     }
 
-    // called from exp-25.007-25.008.sql
+    // called from exp-25.009-25.010.sql
+    // When this is removed, the ExperimentWarningProvider class can also be removed.
     @SuppressWarnings("unused")
     public static void upgradeAmountsAndUnits(ModuleContext context)
     {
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
We should add a reminder to delete the warning code when we remove Java upgrade code. Maybe a comment on `ExperimentUpgradeCode.updateAmountsAndUnits()`? @labkey-adam might have ideas on places we'd see it

labkey-adam:
> We should add a reminder to delete the warning code when we remove Java upgrade code. Maybe a comment on `ExperimentUpgradeCode.updateAmountsAndUnits()`? @labkey-adam might have ideas on places we'd see it

Yes, comment on the upgrade method would be best.

labkey-susanh:
> > We should add a reminder to delete the warning code when we remove Java upgrade code. Maybe a comment on `ExperimentUpgradeCode.updateAmountsAndUnits()`? @labkey-adam might have ideas on places we'd see it
> 
> Yes, comment on the upgrade method would be best.

Done.

PR #7044, opened 2025-09-16
Remove datagenerator code which adds units only 75% of the time

labkey-klum:
#### Rationale
Remove the temporary code prior to requiring units with amounts.


labkey-susanh (approved):
Thanks, Karl!

PR #7048, opened 2025-09-17
Stop passing in a null User

labkey-klum:
#### Rationale
[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53929)

Addresses an issue where we were failing to propagate the user object when updating the QC state for a dataset with a lookup field. This PR also contains the regression test for the issue.

#### Tasks 📍
- [x] Needs Automation
- [x] Verify Fix

### DIFF
--- a/study/src/org/labkey/study/model/StudyManager.java
+++ b/study/src/org/labkey/study/model/StudyManager.java
@@ -1797,7 +1797,7 @@ public class StudyManager
         return defaultQCState;
     }
 
-    private Map<String, VisitImpl> getVisitsForDataRows(DatasetDefinition def, Collection<String> dataLsids)
+    private Map<String, VisitImpl> getVisitsForDataRows(User user, DatasetDefinition def, Collection<String> dataLsids)
     {
         final Map<String, VisitImpl> visits = new HashMap<>();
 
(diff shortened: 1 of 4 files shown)


study/src/org/labkey/study/model/StudyManager.java
```diff
@@ -1797,7 +1797,7 @@ public DataState getDefaultQCState(StudyImpl study)
         return defaultQCState;
     }
 
-    private Map<String, VisitImpl> getVisitsForDataRows(DatasetDefinition def, Collection<String> dataLsids)
+    private Map<String, VisitImpl> getVisitsForDataRows(User user, DatasetDefinition def, Collection<String> dataLsids)
```
labkey-jeckels: Should this be NotNull too?

study/src/org/labkey/study/model/StudyManager.java
```diff
@@ -1833,12 +1833,12 @@ private Map<String, VisitImpl> getVisitsForDataRows(DatasetDefinition def, Colle
         return visits;
     }
 
-    public List<VisitImpl> getVisitsForDataset(Container container, int datasetId)
+    public List<VisitImpl> getVisitsForDataset(User user, Container container, int datasetId)
```
labkey-jeckels: Should this be NotNull too?

labkey-jeckels (approved):
Manually tested too

PR #7049, opened 2025-09-18
Fix bug in Ext4 store layer when table has more than one PK

bbimber:
I found a bug that will occur in the Ext4 store layer in fringe cases when the table has more than one PK. When this happens:

- ApiQueryResponse will write the property 'id' into the JSON only when there is a single PK: https://github.com/LabKey/platform/blob/f1e6c1e3715f4b3350c1559b9e55c2d340138ee8/api/src/org/labkey/api/action/ApiQueryResponse.java#L411
- If there's a single PK, 'id' is written, and the ExjJS Reader will store that: https://github.com/LabKey/platform/blob/f1e6c1e3715f4b3350c1559b9e55c2d340138ee8/api/webapp/clientapi/ext4/data/Reader.js#L32. If there are multiple PKs, this doesnt happen, and the model doesnt store the name of the PKs
- When the ExtJS Proxy submits data back to SaveRows, it looks at reader.getIdProperty() and uses that to populate "oldKeys' in the JSON. Since we never set the idProperty, this fails to include any of the PKs: https://github.com/LabKey/platform/blob/f1e6c1e3715f4b3350c1559b9e55c2d340138ee8/api/webapp/clientapi/ext4/data/Proxy.js#L218
- The result is that oldKeys lacks the PKs and the update fails.

This PR inspects the field objects, and if any are tagged as isKeyField, it will include their values in oldKeys.

### DIFF
--- a/api/webapp/clientapi/ext4/data/Proxy.js
+++ b/api/webapp/clientapi/ext4/data/Proxy.js
@@ -224,7 +224,7 @@ Ext4.define('LABKEY.ext4.data.AjaxProxy', {
 
                 // NOTE: if the 'id' property of the ApiQueryResponse is null, which will occur if there is more than 1 PK (see ApiQueryResponse.getMetaData()),
                 // then we need to ensure the original values of keyFields are still included in oldKeys. The check against record.modified should ensure we send the original value
-                this.reader.metaData.fields.filter(f => f.isKeyField).map(f => f.name).forEach(keyFieldName => {
+                this.reader?.metaData?.fields.filter(f => f.isKeyField).map(f => f.name).forEach(keyFieldName => {
                     oldKeys[keyFieldName] = record.modified?.hasOwnProperty(keyFieldName) ? record.modified[keyFieldName] : record.get(keyFieldName);
                 })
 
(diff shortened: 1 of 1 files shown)


bbimber:
Hi @labkey-martyp, thanks for your help on the other PR. Is there any chance you'd be able to at least open a PR on the LabKey side of the fork so test can get run on this? Because it's ~3 line of contiguous code, it might be just as easy to copy/paste my code via github's UI, before line 225, here: https://github.com/LabKey/platform/blob/9c00799ac4a726dca2637965167ea5f9e647cf64/api/webapp/clientapi/ext4/data/Proxy.js#L225

labkey-martyp:
@bbimber Yes, I do have a PR open with this change and it's running TC tests now. I haven't looked through the code yet, do we need to be defensive here with `this.reader.metaData.fields`. Seems like they'd be there for every query but not sure of every use case and if there is checking for that beforehand. Maybe TC tests will inform that.

bbimber:
> @bbimber Yes, I do have a PR open with this change and it's running TC tests now. I haven't looked through the code yet, do we need to be defensive here with `this.reader.metaData.fields`. Seems like they'd be there for every query but not sure of every use case and if there is checking for that beforehand. Maybe TC tests will inform that.

yeah, that's a good point. I dont see how this would ever be called without the reader loading (so that should always exist), but it's much better to be cautious. Changes added to this branch, though that doesnt help the LK side.

bbimber:
Closing in favor of: https://github.com/LabKey/platform/pull/7061

PR #7054, opened 2025-09-18
Consolidate more XML parsing, FileLike usages 

XingY:
#### Rationale
Addressing a few warnings.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7054
- https://github.com/LabKey/commonAssays/pull/929
- https://github.com/LabKey/premiumModules/pull/359

#### Changes
- set more feature flag for DOCUMENT_BUILDER_FACTORY 
- update more usages to use DOCUMENT_BUILDER_FACTORY
- add return value for FileUtil.deleteDir

### DIFF
--- a/pipeline/src/org/labkey/pipeline/api/ParamParserImpl.java
+++ b/pipeline/src/org/labkey/pipeline/api/ParamParserImpl.java
@@ -19,6 +19,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.labkey.api.pipeline.ParamParser;
 import org.labkey.api.util.StringUtilsLabKey;
+import org.labkey.api.util.XmlBeansUtil;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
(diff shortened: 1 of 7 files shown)


pipeline/src/org/labkey/pipeline/api/ParamParserImpl.java
```diff
@@ -129,9 +129,7 @@ public void parse(InputStream inputStream)
     {
         try
         {
-            DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
-            dbf.setValidating(false);
```
labkey-jeckels, line 133: I don't think we're setting this on the centralized factory. Probably OK to not have this scenario disable it, but calling it out in case you hadn't already considered it too.

labkey-jeckels (approved):

pipeline/src/org/labkey/pipeline/api/ParamParserImpl.java
XingY: I've checked and false seems the default setting.

PR #7057, opened 2025-09-20
Issue 53494: Deadlock between sample type design update and indexer

labkey-jeckels:
#### Rationale
We can improve our synchronization to help reduce deadlock potential when loading caches inside of transactions.

#### Changes
- Switch to use DB/Transaction aware caching for sample types
- When checking the shared cache within a transaction, avoid using the same sync object since we won't be loading anyway

### DIFF
--- a/api/src/org/labkey/api/cache/BlockingCache.java
+++ b/api/src/org/labkey/api/cache/BlockingCache.java
@@ -96,9 +96,12 @@ public class BlockingCache<K, V> implements Cache<K, V>
     {
         Wrapper<V> w;
 
+        if (null == loader)
+            loader = _loader;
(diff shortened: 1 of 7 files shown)


labkey-jeckels:
@labkey-adam if you happen to be online, I'd appreciate a quick review of this PR. I think you're most familiar with our caching. If not, I will enlist someone else soon.

labkey-adam:
> @labkey-adam if you happen to be online, I'd appreciate a quick review of this PR. I think you're most familiar with our caching. If not, I will enlist someone else soon.

Sure, I'll take a look

labkey-adam (approved):

PR #7058, opened 2025-09-21
Add a couple foreign keys

labkey-adam:
#### Rationale
We should have foreign keys on `comm.UserList` (to Announcements) and `core.Members` (to Principals)


labkey-jeckels (approved):
Did testing of the upgrade

PR #7059, opened 2025-09-21
Remote API stream to file

labkey-martyp:
#### Rationale
The current remote streaming API will keep a connection open while processing rows in the data iterator. This can cause connection timeout issues with large tables if there is a lot of processing in triggers or other data iterator steps. The refactor in this PR will stream the remote server response to file before any processing in the dataiterator, then stream from file into the dataiterator. This will be the default method for streaming from remote API.

This PR is based on the related PR from @bbimber. Thanks Ben.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7021
- https://github.com/LabKey/premiumModules/pull/363

#### Changes
- Save remote server response to temp file.
- Pass stream from file to dataiterator.
- Cleanup file when done.

### DIFF
--- a/query/api-src/org/labkey/remoteapi/SelectRowsStreamHack.java
+++ b/query/api-src/org/labkey/remoteapi/SelectRowsStreamHack.java
@@ -122,8 +122,7 @@ public class SelectRowsStreamHack
                         public void close() throws IOException
                         {
                             // close the InputStream and http connection
-                            if (is != null)
-                                IOUtils.closeQuietly(is);
+                            IOUtils.closeQuietly(is);
                             response.close();
 
                             // close the JSONDataLoader
@@ -212,20 +211,16 @@ public class SelectRowsStreamHack
 
                 // Now try this using an InputStream that will fail:
                 preexistingTempFiles = getMatchingTempFileNames();
-                boolean exceptionThrown = false;
 
                 SelfDestructiveInputStream sdic = new SelfDestructiveInputStream(100, 10);
                 try
                 {
                     dib = SelectRowsStreamHack.go(cn, ContainerManager.getHomeContainer().getPath(), cmd, targetContainer, sdic);
-                    List<Map<String, Object>> results = dib.getDataIterator(dic).stream().toList();
+                    @SuppressWarnings("unused") List<Map<String, Object>> results = dib.getDataIterator(dic).stream().toList();
+                    fail("SelfDestructiveInputStream did not throw an exception!");
                 }
-                catch (Exception e)
-                {
-                    exceptionThrown = true;
-                }
+                catch (Exception ignored) {}
 
-                assertTrue("SelfDestructiveInputStream did not throw an exception!", exceptionThrown);
                 assertTrue("SelfDestructiveInputStream was not closed!", sdic.isCloseCalled());
 
                 actualTempFiles = getMatchingTempFileNames();
@@ -236,7 +231,9 @@ public class SelectRowsStreamHack
 
         private Set<String> getMatchingTempFileNames()
         {
-            return Arrays.stream(Objects.requireNonNull(new File(System.getProperty("java.io.tmpdir")).list((dir, name) -> name.startsWith(TEMP_FILE_PREFIX) & name.endsWith(TEMP_FILE_SUFFIX)))).collect(Collectors.toSet());
+            return Arrays.stream(Objects.requireNonNull(new File(System.getProperty("java.io.tmpdir")).
+                    list((dir, name) -> name.startsWith(TEMP_FILE_PREFIX) &&
+                            name.endsWith(TEMP_FILE_SUFFIX)))).collect(Collectors.toSet());
         }
 
         private static class SelfDestructiveInputStream extends InputStream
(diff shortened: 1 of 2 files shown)


bbimber (approved):

bbimber:
Hi @labkey-martyp, just to be sure I get the intent of changes, relative to the original PR, you:

1) made this always cache to a file. this seems file.
2) Removed the additional BufferedInputStream (BufferedInputStream(response.getInputStream())), which was possibly redundant
3) Remove tempFile.deleteOnExit(). I agree that's probably not useful.

Unless I'm missing something, those all seem like good changes.

labkey-martyp:
@bbimber 
1. yes I made this always the behavior instead of opting in.
2. yes the BufferedInputStream was redundant for IOUtils.copy
3. I added the deleteOnExit just as a backup to delete the temp files if something happens during this process

bbimber:
Thanks @labkey-martyp - is there anything addition you need/want on this? It appears tests passed.

labkey-martyp:
@bbimber yes all looks good here. I'm just giving @josheckels a chance to review.

query/api-src/org/labkey/remoteapi/SelectRowsStreamHack.java
```diff
@@ -60,7 +73,36 @@ public DataIterator getDataIterator(DataIteratorContext context)
...
+                        _log.debug("Downloading SelectRows JSON to file: " + tempFile);
+                        IOUtils.copy(ris, os);
+                        _log.debug("Finished saving SelectRows results");
+
+                    }
+
+                    is = new BufferedInputStream(new GZIPInputStream(new FileInputStream(tempFile)))
+                    {
+                        @Override
+                        public void close() throws IOException
```
labkey-jeckels: Have we confirmed that this `close()` gets called in success and failure conditions? It's more important to close this stream than the `InputStream` we get from the `HttpServletRequest`.

labkey-jeckels (changes requested):
Here's an idea for test coverage:

For one of the existing tests that uses a remote source, prior to running the ETL, we could set the log level to `DEBUG` and ensure that we see it writing to the file and then deleting afterwards (might want to refactor logging a bit to explicitly log when the file is successfully deleted instead of our intent to delete.

bbimber:
> Here's an idea for test coverage:
> 
> For one of the existing tests that uses a remote source, prior to running the ETL, we could set the log level to `DEBUG` and ensure that we see it writing to the file and then deleting afterwards (might want to refactor logging a bit to explicitly log when the file is successfully deleted instead of our intent to delete.

HI @labkey-jeckels, your point about failure cases is a very good one, although I don't think the logging thing is a great solution. I have locally verified that under normal conditions the file is deleted. Rather than logging, the test code could inspect the temp directory before and after to look for matching files (this is a good idea). 

This doesn't really address whether this code recovers from all strange failures. I will try a few things locally to see how that works.

bbimber:
@labkey-jeckels or @labkey-martyp, my inability to view dataintegration sorta hampers me here. I see the testautomation/ETLTest module, but no java code using it. Does that all exist in dataintegration module? I am happy to do the work of writing those tests, but I'm not currently sure I can. 

In any event, it seems to me that it would be very simple for those tests to: 1) check for files in the tempdir that match the pattern "^SelectRows-", 2) do the test, and 3) verify no new files matching this pattern remain. I'm happy to do this work, but I dont think the relevant tests exist in modules I can read. 

As far as failure cases: no good ideas on how to automate that testing.

bbimber:
@labkey-jeckels, I got an idea for testing overnight. I will open a PR this morning with that.

query/api-src/org/labkey/remoteapi/SelectRowsStreamHack.java
bbimber: Please see my last commit with test cases, described here: https://github.com/LabKey/platform/pull/7065

bbimber:
Hi @labkey-martyp and @labkey-jeckels: the BVTs passed with the new test cases, described in more detail here: https://github.com/LabKey/platform/pull/7065. Is this enough test coverage? Any other requests on this?

bbimber:
hey @labkey-jeckels: just checking back on this. is that level of automated testing sufficient? any thoughts on how to simulate other failure cases? other requests? if it helps to entice a code review, there's some new string constants in this PR...

labkey-jeckels:
> hey @labkey-jeckels: just checking back on this. is that level of automated testing sufficient? any thoughts on how to simulate other failure cases? other requests? if it helps to entice a code review, there's some new string constants in this PR...

I'll take a look tomorrow and coordinate with Marty. I've been swamped and have been out for some personal stuff as well.

bbimber:
Thanks @labkey-jeckels and @labkey-martyp. this isnt an emergency, i'm just trying to tie up some loose ends. 

Again, if there are specific things you're concerned about or would like to me look into, I'm happy to take the lead on investigation/legwork.

labkey-jeckels:
Looks good to me. I'll merge after a final test run.

bbimber:
I write too much in R and too little in Java. It's really throwing off my && and &. Thanks @labkey-jeckels

labkey-jeckels (approved):

PR #7060, opened 2025-09-22
Add `DomainUtil.getNamesAndLabels` utility method for use in reserving variants of column names 

labkey-susanh:
#### Rationale
Because we generally allow users to provide either the name of a field or its label during import, we should consistently be reserving both the name and the label for each domain so there will not be ambiguity when trying to resolve a column during import. This PR gets us closer to consistency by adding a utility method that will create the set of reservable strings given a field's name and then uses that in various domains, also making more of the reserved field sets static constants.

#### Related Pull Requests
- https://github.com/LabKey/clientModules/pull/61
- https://github.com/LabKey/commonAssays/pull/930
- https://github.com/LabKey/limsModules/pull/1682

#### Changes
- Add `DomainUtil.getNamesAndLabels`
- Add `DomainUtil.getNameAndLabels`
- Update various domains to use these new methods
- Mark `getReservedPropertyNames` as `@NotNull`

#### Tasks 📍
- [ ] ~Manual Testing~
- [ ] Code Review @XingY 
- [ ] ~Needs Automation~

### DIFF
--- a/api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
+++ b/api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
@@ -44,8 +44,8 @@ import org.labkey.api.exp.Lsid;
 import org.labkey.api.exp.TemplateInfo;
 import org.labkey.api.exp.property.AbstractDomainKind;
 import org.labkey.api.exp.property.Domain;
-import org.labkey.api.exp.property.DomainKind;
 import org.labkey.api.exp.property.DomainProperty;
+import org.labkey.api.exp.property.DomainUtil;
 import org.labkey.api.exp.query.ExpMaterialTable;
 import org.labkey.api.exp.query.ExpSampleTypeTable;
 import org.labkey.api.exp.query.SamplesSchema;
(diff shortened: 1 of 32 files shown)


api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
```diff
@@ -110,38 +109,31 @@ public class SampleTypeDomainKind extends AbstractDomainKind<SampleTypeDomainKin
...
-        RESERVED_NAMES.add("SampleID"); // alias for Name
-        RESERVED_NAMES.add("Sample ID");
-        RESERVED_NAMES.add("Status");
-        RESERVED_NAMES.add("Amount"); // alias for storedAmount
-        RESERVED_NAMES.add("RunId"); // Issue 50461
-        RESERVED_NAMES.add("Run Id");
+        RESERVED_NAMES = new CaseInsensitiveHashSet(DomainUtil.getNamesAndLabels(
+                BASE_PROPERTIES.stream().map(PropertyStorageSpec::getName).collect(Collectors.toSet()))
+        );
+        RESERVED_NAMES.addAll(DomainUtil.getNamesAndLabels(Arrays.stream(ExpSampleTypeTable.Column.values()).map(Enum::name).toList()));
```
XingY: Instead of each domain pass in NamesAndLabels as reservedNames, could we instead handle the getNamesAndLabels check in DomainUtil in a centralized place?

api/src/org/labkey/api/query/SimpleTableDomainKind.java
```diff
@@ -242,21 +242,16 @@ public boolean canEditDefinition(User user, Domain domain)
...
-            return Sets.newHashSet(Iterables.transform(table.getBuiltInColumns(),
-                new Function<ColumnInfo, String>() {
-                    @Override
-                    public String apply(ColumnInfo col)
-                    {
-                        return col.getName();
-                    }
-                }
-            ));
+            Set<String> set = new CaseInsensitiveHashSet();
```
XingY, line 250: If this is done in DomainUtil then this change is not needed.

XingY (approved):
Could the getNamesAndLabels step be done in the centralized util instead?

api/src/org/labkey/api/exp/api/SampleTypeDomainKind.java
labkey-susanh: I think `getNamesAndLabels` does what you're asking for if used differently. I've adjusted usages so we aren't constructing as many sets from sets.  

Or were you thinking that `DomainKind.getReservedPropertyNames` would call `getNamesAndLabels`? I think we want to keep the sets returned from `getReservedPropertyNames` static where we can so we aren't always constructing new sets.

assay/api-src/org/labkey/api/assay/AssayRunDomainKind.java
```diff
@@ -15,17 +15,32 @@
...
 import java.util.Set;
+import java.util.stream.Collectors;
 
 public class AssayRunDomainKind extends AssayDomainKind
 {
+    private static final Set<String> RESERVED_NAMES;
+    static
+    {
+        RESERVED_NAMES = new CaseInsensitiveHashSet(getAssayReservedPropertyNames());
+        RESERVED_NAMES.addAll(DomainUtil.getNameAndLabels("AsasyId"));
```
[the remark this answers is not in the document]
labkey-susanh: Fixed.

PR #7061, opened 2025-09-22
Fix bug in Ext4 store layer when table has more than one PK

bbimber:
Originally opened under: https://github.com/LabKey/platform/pull/7049

See below:

I found a bug that will occur in the Ext4 store layer in fringe cases when the table has more than one PK. When this happens:

- ApiQueryResponse will write the property 'id' into the JSON only when there is a single PK: https://github.com/LabKey/platform/blob/f1e6c1e3715f4b3350c1559b9e55c2d340138ee8/api/src/org/labkey/api/action/ApiQueryResponse.java#L411
- If there's a single PK, 'id' is written, and the ExjJS Reader will store that: https://github.com/LabKey/platform/blob/f1e6c1e3715f4b3350c1559b9e55c2d340138ee8/api/webapp/clientapi/ext4/data/Reader.js#L32. If there are multiple PKs, this doesnt happen, and the model doesnt store the name of the PKs
- When the ExtJS Proxy submits data back to SaveRows, it looks at reader.getIdProperty() and uses that to populate "oldKeys' in the JSON. Since we never set the idProperty, this fails to include any of the PKs: https://github.com/LabKey/platform/blob/f1e6c1e3715f4b3350c1559b9e55c2d340138ee8/api/webapp/clientapi/ext4/data/Proxy.js#L218
- The result is that oldKeys lacks the PKs and the update fails.

This PR inspects the field objects, and if any are tagged as isKeyField, it will include their values in oldKeys.

### DIFF
--- a/api/webapp/clientapi/ext4/data/Proxy.js
+++ b/api/webapp/clientapi/ext4/data/Proxy.js
@@ -224,7 +224,7 @@ Ext4.define('LABKEY.ext4.data.AjaxProxy', {
 
                 // NOTE: if the 'id' property of the ApiQueryResponse is null, which will occur if there is more than 1 PK (see ApiQueryResponse.getMetaData()),
                 // then we need to ensure the original values of keyFields are still included in oldKeys. The check against record.modified should ensure we send the original value
-                this.reader.metaData.fields.filter(f => f.isKeyField).map(f => f.name).forEach(keyFieldName => {
+                this.reader?.metaData?.fields?.filter(f => f.isKeyField).map(f => f.name).forEach(keyFieldName => {
                     oldKeys[keyFieldName] = record.modified?.hasOwnProperty(keyFieldName) ? record.modified[keyFieldName] : record.get(keyFieldName);
                 })
 
--- a/api/webapp/clientapi/ext4/data/Proxy.js
+++ b/api/webapp/clientapi/ext4/data/Proxy.js
@@ -222,6 +222,12 @@ Ext4.define('LABKEY.ext4.data.AjaxProxy', {
                 oldKeys[this.reader.getIdProperty()] = id;
                 oldKeys['_internalId'] = record.internalId;  //NOTE: also include internalId for records that do not have a server-assigned PK yet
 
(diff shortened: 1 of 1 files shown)


bbimber:
hi @labkey-martyp, my team city user cannot see the BVT results. are there failures relevant to this PR?

labkey-martyp:
@bbimber I ran a pretty broad set of tests here and got some random failures. I don't think they're related but I'm running those particular tests again just to ensure there are no issues. Thanks for opening the PR.

bbimber:
Thanks @labkey-martyp: seems like the last tests were clean. Anything else needed to merge this?

labkey-martyp:
@bbimber I have some other more extensive tests running. The tests on the PR are the minimum that should pass. TeamCity has a lot more. This is old code I just want to make sure we're not breaking anything expecting the old behavior.

bbimber:
@labkey-martyp: how did the other tests look?

While I am pretty sure this PR is perfectly fine as-is (and hopefully the tests confirm that), I tried to think about any fringe cases that could break this. I dont have any; however, in the spirit of extreme caution it would be possible to narrow the scope of this change. I'm trying to address the condition where a table has multiple PKs, which behaves badly in the server API / client Reader codepath listed above. In https://github.com/LabKey/platform/pull/7070, I propose a change that would make the inspection of keyFields only occur when idProperty == _internalId. This should mean that every single store that's functional today should be untouched, and this additional check will only happen for the currently broken multiple PK cases. 

That PR does make the change to use a constant for _internalId (inspired by Josh's computer science fundamentals). This does create the requirement that Reader and Proxy always be loaded together, since the constant is defined in Reader.js. The only code I see loading either of them is ext4.lib.xml, and this already loads both. So as long as this requirement seems OK, using a constant is a marginal improvement.

labkey-martyp (approved):
@bbimber Ok tests completed last night. I don't see any new failures. I think this PR is ok, I don't see any reason to move forward with the other one unless you have any concerns.

Not sure if you have access to merge, if not I can merge this in.

bbimber:
thanks @labkey-martyp. I appears I can merge.

PR #7065, opened 2025-09-23
Ensure temp files are deleted and add testing

bbimber:
@labkey-martyp or @labkey-jeckels: this PR is targeting the existing Labkey org feature branch. I cant commit changes to that. Would you be willing to merge this? 

This addresses some of @labkey-jeckels requests. It:
- Adds more code to ensure the temp file is always deleted
- Adds integration testing over that behavior when there is no error
- It also adds a test case using SelfDestructiveInputStream, which is a contrived InputStream that will throw an error after a period of time. Making SelectRowsStreamHack work with this involved creating an overloaded private go() method. I'd generally prefer not to add something like this for testing, but didnt see a clear alternative, and at least the test-specific code is all private. Happy to discuss other approaches.


bbimber:
correction, apparently I can commit to FBs on the LK org, just not create FBs.

PR #7066, opened 2025-09-23
Issue 53036: LKSM: Aliquot registration event in timeline polish

XingY:
#### Rationale
Aliquot's inherited parent field values shouldn't be in audit log. When querying for existing aliquot rows, those fields should be removed from the map that's passed to AuditHandler to build change details. This PR handles removing of parent fields for field whose label or alias might be different from field name.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1856
- https://github.com/LabKey/labkey-ui-premium/pull/821
- https://github.com/LabKey/platform/pull/7066
- https://github.com/LabKey/limsModules/pull/1684

#### Changes
- check for label and alias when excluding parent fields from aliquot existing data

### DIFF
--- a/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
+++ b/experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
@@ -826,7 +826,18 @@ public class SampleTypeUpdateServiceDI extends DefaultQueryUpdateService
                 .map(ImportAliasable::getName)
                 .collect(Collectors.toSet());
 
-        return new CaseInsensitiveHashSet(fields);
+        Set<String> metaFieldNames = new CaseInsensitiveHashSet(fields);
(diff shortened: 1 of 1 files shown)


experiment/src/org/labkey/experiment/api/SampleTypeUpdateServiceDI.java
```diff
@@ -826,7 +826,18 @@ public Set<String> getSampleMetaFields()
                 .map(ImportAliasable::getName)
                 .collect(Collectors.toSet());
 
-        return new CaseInsensitiveHashSet(fields);
+        Set<String> metaFieldNames = new CaseInsensitiveHashSet(fields);
```
cnathe, line 830: can we add a comment here as to why this is needed and the related issue number?

cnathe (approved):

PR #7067, opened 2025-09-23
Track and report index information via a list of records

labkey-adam:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53887

`TableInfo.getUniqueIndices()` and `TableInfo.getAllIndices()` have been returning `Map<String, Pair<IndexType, List<ColumnInfo>>>`, which is both awkward to work with and hard to extend. Returning `List<IndexDefinition>` increases flexibility and readability. 

Reading and stashing each index's filter condition helps improve the detection of overlapping indices. SQL Server JDBC driver does not report filter conditions in the `getIndexInfo()` `ResultSet`; work around that by stashing the results of an explicit query at data source `prepare()` time.

Drop five redundant indices in the `exp` schema.

#### Tasks 📍
- [x] Manual Testing @labkey-jeckels 

### DIFF
--- a/api/src/org/labkey/api/data/EnumTableInfo.java
+++ b/api/src/org/labkey/api/data/EnumTableInfo.java
@@ -21,10 +21,9 @@ import org.labkey.api.query.ExprColumn;
 import org.labkey.api.query.SchemaKey;
 import org.labkey.api.query.UserSchema;
 
+import java.util.ArrayList;
 import java.util.EnumSet;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 
 /**
  * Exposes a Java enum as a virtual query table. Useful for creating lookups when it's really a hard-coded list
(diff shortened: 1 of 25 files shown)


api/src/org/labkey/api/data/EnumTableInfo.java
```diff
@@ -181,21 +177,24 @@ public boolean isPublic()
...
-    public Map<String, Pair<IndexType, List<ColumnInfo>>> getUniqueIndices()
+    public Map<String, IndexDefinition> getUniqueIndices()
     {
-        Map<String, Pair<IndexType, List<ColumnInfo>>> indices = new HashMap<>();
-        indices.put("uq_value", Pair.of(IndexType.Unique, Arrays.asList(getColumn("Value"))));
-        indices.put("uq_oridinal", Pair.of(IndexType.Unique, Arrays.asList(getColumn("Ordinal"))));
+        Map<String, IndexDefinition> indices = new HashMap<>();
+        String name = "uq_value";
+        indices.put(name, new IndexDefinition(name, IndexType.Unique, List.of(getColumn("Value")), null));
+        name = "uq_ordinal";
```
labkey-jeckels: This is fine, but seems unnecessary to pull out to a separate variable

api/src/org/labkey/api/data/SchemaColumnMetaData.java
```diff
@@ -348,57 +345,45 @@ private void loadIndices(SchemaTableInfo ti) throws SQLException
                             return dbmd.getIndexInfo(l.getCatalogName(), l.getSchemaName(), l.getTableName(), uniqueIndicesOnly, true);
                         }));
 
+                String primaryKeyName = ti.getPrimaryKeyName();
                 Set<String> ignoreIndex = new HashSet<>();
-                Map<String, Pair<TableInfo.IndexType, List<ColumnInfo>>> indexMap = new HashMap<>();
+                Map<String, IndexDefinition> indexMap = new CaseInsensitiveHashMap<>();
```
labkey-jeckels, line 350: This switched to a CaseInsensitiveHashMap but other places didn't. Should they?

experiment/resources/schemas/dbscripts/sqlserver/exp-0.000-24.000.sql
```diff
@@ -14,15 +14,20 @@
...
-/* exp-0.00-11.20.sql */
-
 /*
  *  Creates experiment annotation tables in the exp schema base on FuGE-OM types
  */
 
 CREATE SCHEMA exp;
 GO
 
+-- Provisioned schema used by DataClassDomainKind
```
labkey-jeckels, line 24: I didn't review this file. Let me know if I should.

labkey-jeckels (approved):
I ran this locally and confirmed that the index report showed fewer overlapping indices. I verified that `idx_material_AK` showed up on both DBs as having a filterCondition via the debugger

api/src/org/labkey/api/data/EnumTableInfo.java
labkey-adam: It is (was) used twice in the next line. But, I took the plunge of getting rid of the map altogether, since nobody looks up into it. Those methods just return `List<IndexDefinition>` now.

api/src/org/labkey/api/data/SchemaColumnMetaData.java
labkey-adam: Method no longer populates maps

experiment/resources/schemas/dbscripts/sqlserver/exp-0.000-24.000.sql
labkey-adam: Not needed, IMO

PR #7072, opened 2025-09-24
Regular expression sanitization

XingY:
#### Rationale
clean up warnings.

#### Related Pull Requests
- 

#### Changes
- rewrite regex


labkey-jeckels (commented):
I'm not very good with regex so I generated some descriptions to add as comments. I also added some unit tests which seem to indicate that the quote regex matches differently from the original. Can you take a look?

XingY:
> I'm not very good with regex so I generated some descriptions to add as comments. I also added some unit tests which seem to indicate that the quote regex matches differently from the original. Can you take a look?

Thanks for the unit tests. I updated the regex so to account for the different behavior. The misunderstanding here is (\.) is only suppossed to match newline characters or other special characters, not any backslash+character.

labkey-jeckels:
> > I'm not very good with regex so I generated some descriptions to add as comments. I also added some unit tests which seem to indicate that the quote regex matches differently from the original. Can you take a look?
> 
> Thanks for the unit tests. I updated the regex so to account for the different behavior. The misunderstanding here is (.) is only suppossed to match newline characters or other special characters, not any backslash+character.

GitHub is worried about perf for the updated version. I'm not sure this will be a real-world concern, but I'm not sure if we have to exactly preserve the old behavior either. Maybe my auto-generated unit tests aren't the best place to focus - I'm not seeing any unit test coverage for `JavaCodeFilter` or `XmlCodeFilter`

XingY:
> > > I'm not very good with regex so I generated some descriptions to add as comments. I also added some unit tests which seem to indicate that the quote regex matches differently from the original. Can you take a look?
> > 
> > 
> > Thanks for the unit tests. I updated the regex so to account for the different behavior. The misunderstanding here is (.) is only suppossed to match newline characters or other special characters, not any backslash+character.
> 
> GitHub is worried about perf for the updated version. I'm not sure this will be a real-world concern, but I'm not sure if we have to exactly preserve the old behavior either. Maybe my auto-generated unit tests aren't the best place to focus - I'm not seeing any unit test coverage for `JavaCodeFilter` or `XmlCodeFilter`

I think we probably do want to preserve the old behavior. The regex was meant to allow only special characters such as \n, \t, not any backslash followed by any character. 

As for GitHub complaining about inefficient expression with the rewrite, I'm not sure this is a legit complain since it's only looking at a sub part of the expression in isolation. I checked with a few other tools and they all agree with potential inefficiency with the old expression, but deny such inefficiency with the rewrite.

labkey-jeckels (approved):
OK, thanks for the review of the quote regex.

I pushed a small update with cleanup on that class. If it looks good to you, feel free to merge. Sounds like we'll need to manually suppress the GitHub finding on the regex.

PR #7074, opened 2025-09-25
Issue 53979: Poor and inconsistent error messages for gargantuan amounts

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53979

Decimal fields for sample type (and likely other data types as well) support non-finite values like "Infinity" and "-Infinity" but we currently throw an error when building up the JSON response object. This PR avoids that by calling toString() on the non-finite values when creating the JSONObject response.

Also, we were silently converting double values larger than Double.MAX_VALUE to "Infinity" in the convert helpers. This PR also updates that to now throw a "could not convert" error in this case (for Decimal field types).

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1691

#### Changes
- avoid "JSON does not allow non-finite numbers." error in save rows response when non-finite number is used for a decimal field
- don't allow non-finite values for sample amount
- throw conversion error if double value provided that is > Double.MAX_VALUE


labkey-susanh (approved):

api/src/org/labkey/api/util/JsonUtil.java
```diff
@@ -230,6 +230,16 @@ public static JSONObject toJsonPreserveNulls(Map<String, Object> map)
         return json;
     }
 
+    // Issue 53979: "JSON does not allow non-finite numbers." The new JSONObject throws an exception if you try to
+    // put a Double.NaN or Double.POSITIVE_INFINITY/Double.NEGATIVE_INFINITY value into a JSONObject. This method
+    // preserves those values by converting them to strings.
+    public static Map<String, Object> toMapPreserveNonFinite(Map<String, Object> map)
+    {
+        Map<String, Object> result = new HashMap<>();
+        map.forEach((k, v) -> result.put(k, v instanceof Double d && !Double.isFinite(d) ? d.toString() : v));
```
labkey-jeckels: Seems like we should do the same with `Float`

api/src/org/labkey/api/data/ConvertHelper.java
```diff
@@ -979,8 +983,14 @@ else if (s.equalsIgnoreCase("Inf"))
                     return Double.POSITIVE_INFINITY;
                 else if (s.equalsIgnoreCase("-Inf"))
                     return Double.NEGATIVE_INFINITY;
+                else if (s.equalsIgnoreCase("Infinity"))
```
labkey-jeckels, line 986: Candidates for simple unit testing?

labkey-jeckels (approved):
I didn't test this but code changes look OK.

api/src/org/labkey/api/util/JsonUtil.java
cnathe: Done

api/src/org/labkey/api/data/ConvertHelper.java
cnathe: done

api/src/org/labkey/api/data/ConvertHelper.java
```diff
@@ -967,7 +967,11 @@ else if (value instanceof Number)
             String s = value.toString();
             try
             {
-                return Double.valueOf(s);
+                Double d = Double.valueOf(s);
```
labkey-susanh, line 970: Looks like we don't have a counterpart for this for Float. Do we need such a thing?

labkey-susanh (approved):

PR #7075, opened 2025-09-25
Path related warnings

XingY:
#### Related Pull Requests
- https://github.com/LabKey/commonAssays/pull/934

#### Changes
- remove support for non-webdav path based query import
- refactor File constructor/resolve usages

### DIFF
--- a/api/src/org/labkey/api/attachments/FileAttachmentFile.java
+++ b/api/src/org/labkey/api/attachments/FileAttachmentFile.java
@@ -47,12 +47,12 @@ public class FileAttachmentFile implements AttachmentFile
 
     public FileAttachmentFile(FileLike file)
     {
-        this(file.toNioPathForRead().toFile(), null);
+        this(file == null ? null : file.toNioPathForRead().toFile(), null);
     }
 
     public FileAttachmentFile(FileLike file, @Nullable String originalName)
     {
-        this(file.toNioPathForRead().toFile(), originalName);
+        this(file == null ? null : file.toNioPathForRead().toFile(), originalName);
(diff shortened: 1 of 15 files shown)


api/src/org/labkey/api/attachments/FileAttachmentFile.java
```diff
@@ -47,12 +47,12 @@ public FileAttachmentFile(File file)
 
     public FileAttachmentFile(FileLike file)
     {
-        this(file.toNioPathForRead().toFile(), null);
+        this(file == null ? null : file.toNioPathForRead().toFile(), null);
```
labkey-jeckels: We should probably prohibit null for the file. Lots of methods will NPE if `_file` is null.

api/src/org/labkey/api/query/AbstractQueryImportAction.java
```diff
@@ -478,12 +478,23 @@ else if (StringUtils.isNotBlank(path))
                     PipeRoot root = PipelineService.get().findPipelineRoot(getContainer());
                     if (root != null)
                     {
-                        // Attempt absolute path first, then relative path from pipeline root
-                        File f = new File(path);
-                        if (!root.isUnderRoot(f))
-                            f = root.resolvePath(path);
+                        Path relativePath = Path.parse(path);
```
labkey-jeckels: I looked at the history and came to the same conclusion - this has been broken since 24.11. Let's get rid of both the relative and absolute variants here, but retain the webdav part above.

labkey-jeckels (approved):

PR #7076, opened 2025-09-26
Nullable insert rows

labkey-martyp:
#### Rationale
Some implementations return null so annotate.

#### Related Pull Requests
- TBD

#### Changes
- @Nullable

### DIFF
--- a/api/src/org/labkey/api/query/QueryUpdateService.java
+++ b/api/src/org/labkey/api/query/QueryUpdateService.java
@@ -180,7 +180,7 @@ public interface QueryUpdateService extends HasPermission
      * @throws QueryUpdateServiceException Thrown for implementation-specific exceptions.
      * @throws DuplicateKeyException Thrown if primary key values were supplied in the map
      */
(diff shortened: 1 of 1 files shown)


labkey-jeckels (approved):
I think (and hope) that we should only return null if there are errors that are conveyed through the `errors` argument. Does that match your experience? If so, please update the JavaDoc to match.

I separately wonder if we should throw the `errors` object instead, which seems to be what `updateRows()` does, though obviously that's a larger change

labkey-bpatel:
Merging in for Marty

labkey-martyp:
> I think (and hope) that we should only return null if there are errors that are conveyed through the `errors` argument. Does that match your experience? If so, please update the JavaDoc to match.
> 
> I separately wonder if we should throw the `errors` object instead, which seems to be what `updateRows()` does, though obviously that's a larger change

@labkey-jeckels Yes, the only case I could find (and where this issue originated) was returning null when there are errors. I did think about refactoring to throw errors instead of just passing back null but looked like a big refactor and this was part of a high pri ticket. I can file an issue if you think we should follow up on that.

PR #7079, opened 2025-09-26
Issue 53722: Unable to export multiple sample types with long overlapping names

labkey-susanh:
#### Rationale
Issue [53722](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53722) - Excel limits the length of sheet names in workbooks to 31 characters. Sample type names can be up to 200 characters (I believe). If two sample types share the same 31-character prefix, only one of them will be included in a exported workbook. 

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1694

#### Changes
- Add logic for trimming to unique names when creating sheets for export

#### Tasks 📍
- [x] Needs Automation
- [x] Verify Fix


query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -2109,6 +2109,38 @@ public Object execute(ExportQueriesForm form, BindException errors) throws Excep
...
+            form.getQueryForms().forEach(qf -> {
+                String sheetName = qf.getSheetName();
+                QueryView qv = qf.getQueryView();
+                // use the given sheet name if provided, otherwise try the query definition name
+                String name = StringUtils.isNotBlank(sheetName) ? sheetName : qv.getQueryDef().getName();
+                // if there is no sheet name or queryDefinition name, use a data region name if provided. Otherwise, use "Data"
+                name = StringUtils.isNotBlank(name) ? name : StringUtils.isNotBlank(qv.getDataRegionName()) ? qv.getDataRegionName() : "Data";
+                // clean it to remove undesirable characters and make it of an acceptable length
+                name = ExcelWriter.cleanSheetName(name);
+                nameFormMap.computeIfAbsent(name, k -> new ArrayList<>()).add(qf);
```
XingY, line 2123: Does case sensitivity matter?

query/src/org/labkey/query/controllers/QueryController.java
```diff
@@ -2109,6 +2109,38 @@ public Object execute(ExportQueriesForm form, BindException errors) throws Excep
             ResponseHelper.setContentDisposition(response, ResponseHelper.ContentDispositionType.attachment);
             ViewContext viewContext = getViewContext();
 
+            Map<String, List<QueryForm>> nameFormMap = new HashMap<>();
+            Map<QueryForm, String> sheetNames = new HashMap<>();
```
XingY, line 2113: Does QueryForm as key work well? Could we do form index to name map instead?

XingY (approved):

query/src/org/labkey/query/controllers/QueryController.java
labkey-susanh: Good call. Yes, it does.

query/src/org/labkey/query/controllers/QueryController.java
labkey-susanh: It seems to work with QueryForm as the key. Perhaps not super efficient, but a little more straightforward (and this isn't an operation that is done all that often). I can change to Integer objects if you have a strong opinion here.

PR #7086, opened 2025-09-30
Check file path name

XingY:
#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7083

#### Changes
-

### DIFF
--- a/api/src/org/labkey/api/util/FileType.java
+++ b/api/src/org/labkey/api/util/FileType.java
@@ -374,7 +374,7 @@ public class FileType implements Serializable
 
     public Path getPath(Path parentDir, String basename)
     {
-        return parentDir.resolve(getName(parentDir, basename));
+        return FileUtil.appendName(parentDir, getName(parentDir, basename));
     }
 
     /**
--- a/api/src/org/labkey/api/util/FileType.java
+++ b/api/src/org/labkey/api/util/FileType.java
@@ -349,7 +349,7 @@ public class FileType implements Serializable
             for (String suffix : _suffixes)
             {
                 String name = tryName(parentDir, basename + suffix);
-                Path f = parentDir.resolve(name);
+                Path f = FileUtil.appendName(parentDir, name);
                 if (NetworkDrive.exists(f))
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/util/FileType.java
```diff
@@ -243,7 +243,16 @@ public FileType(List<String> suffixes, String defaultSuffix, boolean dir, gzSupp
...
         if (_supportGZ.booleanValue())  // TPP treats xml.gz as a native format
-        {   // in the case of existing files, non-gz copy wins if present
+        {
+            try
+            {
+                FileUtil.legalPathPartThrow(name);
+            }
+            catch (Exception e)
+            {
+                return name; // don't try to be clever if the name is not legal
```
labkey-susanh: If the name is not legal do we want to return it here or throw?

XingY: The change is meant to skip the NetworkDrive.exists(f) check if the name is illegal. As for throwing, that would be a bigger change for when we enforce check more broadly.

labkey-susanh: At least one of the callers for `tryName` also does a check for `NetworkDrive.exists` with the name that is returned here. Is that not a problem if the name has dots and slashes in it? Or am I misunderstanding the usages?

XingY: Good point. I've updated the caller (FileType line 352) to use a different method to resolve file.  Looking at the callers it seems Path.of(name) would actually never be called since parentDir is never null. So it seems pretty safe to throw. I've updated to throw instead.

labkey-susanh: Might also need to update `FileType.getPath`, which also uses `parentdir.resolve(getName(...))`, which calls `tryName`.  There are also a couple of other usages of `FileType.getName` that might need validating or updating.

XingY: done

labkey-susanh (approved):

PR #7091, opened 2025-10-02
Issue 53168: An aliquot can be updated to be its own parent’s parent

XingY:
#### Rationale
We don't allow a sample/data to be its own parent, but we allow its descendent to be added as ancestors, resulting in circular lineage. This PR add check to see if a new parent is a existing child. This PR doesn't check for >1 generation descendants against parents. 

#### Related Pull Requests
- 

#### Changes
- disallow adding direct children as sample/data's direct parent
- add junit tests for circular lineage for aliquots, samples and sources

#### Tasks 📍
  - [x] check adding sample's derivative as its parent
  - [x] check adding sample's aliquot as its parent
  - [x] check adding sample's child source as its data parent
  - [x] adding source's source child as it's parent
  - [x] check update from App ui 
  - [x] check update/merge from file
- [x] Needs Automation (junit tests added)
- [x] Verify Fix

### DIFF
--- a/experiment/src/org/labkey/experiment/api/ExpSampleTypeTestCase.jsp
+++ b/experiment/src/org/labkey/experiment/api/ExpSampleTypeTestCase.jsp
@@ -696,7 +696,7 @@ public void testUpdateSomeParents() throws Exception
     }
     catch (Exception e)
     {
-        assertThat(e.getMessage(), containsString("'C1' is aliquoted from sample 'P2-1'. Circular relationships are not allowed."));
+        assertThat(e.getMessage(), containsString("'C1' is derived from sample 'P2-1'. Circular relationships are not allowed."));
     }
 
     errors = new BatchValidationException();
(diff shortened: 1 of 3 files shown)


experiment/src/org/labkey/experiment/api/ExpSampleTypeTestCase.jsp
```diff
@@ -682,10 +682,25 @@ public void testUpdateSomeParents() throws Exception
...
+    rows.clear();
+    rows.add(CaseInsensitiveHashMap.of("rowId", parent2Type.getSample(c, "P2-1").getRowId(), "MaterialInputs/ChildSamples", "C1"));
+    try
+    {
+        svcParent2.updateRows(user, c, rows, null, errors, null, null);
+        fail("Expected to throw exception");
+    }
+    catch (Exception e)
+    {
+        assertThat(e.getMessage(), containsString("'C1' is aliquoted from sample 'P2-1'. Circular relationships are not allowed."));
```
labkey-susanh: The message seems wrong here. C1 is derived not aliquoted, isn't it?

labkey-susanh (changes requested):

experiment/src/org/labkey/experiment/api/ExpSampleTypeTestCase.jsp
XingY: Good catch. Fixed.

labkey-susanh (approved):

PR #7097, opened 2025-10-02
Lists: add support for moving rows between folders

labkey-nicka:
#### Rationale
This adds the capability to move rows within a list between folders. This is invoked via the `query-moveRows.api` endpoint.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7097
- https://github.com/LabKey/limsModules/pull/1700
- https://github.com/LabKey/premiumModules/pull/369
- https://github.com/LabKey/testAutomation/pull/2725

#### Changes
**Lists**
- Introduce support for `moveRows` via list query update service
- Add `getDomainOrThrow()` methods to `ListDefinition`
- Update `ListAuditProvider` to support moving list events between folders

**Platform**
- Move `ContainerManager.updateContainer()` to `Table.updateContainer()`
- Update implementation of `updateContainer()` to specify a `Filter`. Much more robust way of handling column/name resolution and filter construction.
- Consolidate some usages to use `Table.updateContainer()` rather than wire up the SQL themselves
- Expand check for moving rows from direct children of parent project to any child of parent project.
- ~~Add assertion for "RowPk" value when a row or existing row is supplied when creating a query audit event~~
- Resolve "RowPk" by `fieldKey.getName()` when creating a query audit event
- Return number of rows modified via `AttachmentService.moveAttachments()`


api/src/org/labkey/api/data/ContainerManager.java
```diff
@@ -2696,29 +2696,9 @@ private static boolean isValidTargetContainer(Container current, Container targe
 
         boolean moveFromProjectToChild = current.isProject() && target.getParent().equals(current);
         boolean moveFromChildToProject = !current.isProject() && current.getParent().isProject() && current.getParent().equals(target);
-        boolean moveFromChildToSibling = !current.isProject() && current.getParent().isProject() && current.getParent().equals(target.getParent());
+        boolean moveFromChildToChildWithinProject = !current.isProject() && !target.isProject() && current.getProject() != null && current.getProject().equals(target.getProject());
```
labkey-susanh: Perhaps `moveToDescendantWithinProject` (could also be a bit misleading but reads better to me at the moment).

Also these checks seem a little incongruent. We allow moving from a child to any other descendant of the project but from the parent only to a direct child. For the apps, these are the same (currently), but if we need to support moving to any level of the hierarchy it seems the checks should be consistent.

api/src/org/labkey/api/data/Table.java
```diff
@@ -1128,6 +1128,65 @@ else if (pkVals instanceof Map)
...
+        filter.addInClause(idColumn.getFieldKey(), ids);
+
+        return updateContainer(table, targetContainer, filter, user, withModified);
+    }
+
+    public static int updateContainer(TableInfo table, Container targetContainer, @NotNull SimpleFilter filter, @Nullable User user, boolean withModified)
+    {
+        assert assertInDb(table);
+        SQLFragment dataUpdate = new SQLFragment("UPDATE ").append(table)
+                .append(" SET container = ").appendValue(targetContainer);
```
labkey-matthewb: constant?  Also see class SimpleTranslator.SpecialColumn (maybe consolidate?)

list/src/org/labkey/list/model/ListQueryUpdateService.java
```diff
@@ -446,6 +465,239 @@ private boolean validateProperty(DomainProperty prop, Object value, Map<String,
         return true;
     }
 
+    private record ListRecord(Object key, String entityId) { }
```
labkey-matthewb, line 472: Same as ListItem w/o a property map?

list/src/org/labkey/list/model/ListQueryUpdateService.java
```diff
@@ -446,6 +465,239 @@ private boolean validateProperty(DomainProperty prop, Object value, Map<String,
...
+
+            for (GUID containerId : containerRows.keySet())
+            {
+                Container sourceContainer = ContainerManager.getForId(containerId);
+                if (sourceContainer == null)
+                    throw new InvalidKeyException("Container '" + containerId + "' does not exist.");
+
+                if (!sourceContainer.hasPermission(user, MoveEntitiesPermission.class))
+                    throw new UnauthorizedException("You do not have permission to move list records out of '" + sourceContainer.getName() + "'.");
+
```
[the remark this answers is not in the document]
labkey-nicka: Yes there is. It is done in `ContainerManager.getMoveTargetContainer()`.

list/src/org/labkey/list/model/ListQueryUpdateService.java
```diff
@@ -446,6 +465,239 @@ private boolean validateProperty(DomainProperty prop, Object value, Map<String,
...
+        String keyName = _list.getKeyName();
+        String sourcePath = sourceContainer.getPath();
+        String targetPath = targetContainer.getPath();
+
+        for (ListRecord record : records)
+        {
+            ListAuditProvider.ListAuditEvent event = new ListAuditProvider.ListAuditEvent(targetContainer, "An existing list record was moved", _list);
+            event.setListItemEntityId(record.entityId);
+            event.setOldRecordMap(AbstractAuditTypeProvider.encodeForDataMap(CaseInsensitiveHashMap.of("Folder", sourcePath, keyName, record.key.toString())));
+            event.setNewRecordMap(AbstractAuditTypeProvider.encodeForDataMap(CaseInsensitiveHashMap.of("Folder", targetPath, keyName, record.key.toString())));
```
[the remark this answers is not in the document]
labkey-nicka: This is not true.

api/src/org/labkey/api/data/Table.java
labkey-nicka: Done.

api/src/org/labkey/api/data/ContainerManager.java
labkey-nicka: Updated to accept any move within a project hierarchy including the project itself.

list/src/org/labkey/list/model/ListQueryUpdateService.java
labkey-nicka: I had considered it, however, I'm trying to keep memory foot print at small as possible here so just utilizing this record.

labkey-susanh (approved):

PR #7098, opened 2025-10-02
Return Set<Class<?>> from Module "get test" methods

labkey-adam:
#### Rationale
Eliminate IntelliJ warnings caused by `Module` methods `getUnitTests()` and `getIntegrationTests()` returning raw `Set<Class>`; they now return `Set<Class<?>>`.

Also, eliminate use of the deprecated `AbstractAuditTypeProvider()` constructor and overrides of `getDomainKind()` in its subclasses; we now always call the constructor that passes in a `DomainKind`.

### DIFF
--- a/announcements/src/org/labkey/announcements/AnnouncementModule.java
+++ b/announcements/src/org/labkey/announcements/AnnouncementModule.java
@@ -200,8 +200,7 @@ public class AnnouncementModule extends DefaultModule implements SearchService.D
     }
 
     @Override
-    @NotNull
-    public Set<Class> getIntegrationTests()
+    public @NotNull Set<Class<?>> getIntegrationTests()
     {
         return Set.of(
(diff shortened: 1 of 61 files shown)


labkey-jeckels (approved):
Seems like we don't need manual testing. We can verify that the BVT run has the same number of tests before and after this change, meaning the unit/integration tests are being found and executed.

labkey-adam:
> Seems like we don't need manual testing. We can verify that the BVT run has the same number of tests before and after this change, meaning the unit/integration tests are being found and executed.

BVT runs on this branch and on develop yesterday both report 553 tests passed:

- https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_Bvt/3686811?buildTab=tests
- https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_Bvt/3686536?buildTab=tests

PR #7099, opened 2025-10-03
Issue 53063: Support "Aliquoted From" with space

XingY:
#### Rationale
"AliquotedFrom" is used as aliquot parent input column, similar to MaterialInputs/Type is used to indicate lineage parent. It's requested that user can alternatively use "Aliquoted From" with a space as the import column header. This PR updates various steps that checks for AliquotedFrom to also allow "Aliquoted From". The exception is with naming expression, where only ${AiquotedFrom} syntax is supported. 

This PR also addresses Issue 53963: Cross-sample-type import gives incorrect row number in message for name expression problems. 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-premium/pull/827
- https://github.com/LabKey/platform/pull/7099
- https://github.com/LabKey/limsModules/pull/1703

#### Tasks 📍
- [x] Manual Testing @labkey-danield  📍 
- [x] Needs Automation @XingY 
~Verify Fix~

#### Manual Test plan
- [x] verify import/merge using "Aliquoted From" as column header
   - [x] Verify warning about unknown field is no longer present
   - [x] verify aliquot roll up
- [x] test parent only, aliquot only and both type fields
   - [x] verify audit log
   - [x] verify actual value stored in database (parent value won't populate aliquot fields)
- [x] cross type/folder import
   - [x] cross type import where all "aliquoted from" value are numeric like
- [x] verify aliquot naming pattern
   - [x] we don't support "aliquoted from" in naming pattern, however, if "aliquoted from" is provided in data file, the naming pattern "aliquotedfrom" should still map to it.
- [x] update from file shows warning about "aliquoted from" column
- [x] if both "aliquotedfrom" and "aliquoted from" is present, import should fail
- [x] API test: insert using "aliquoted from"
- [x] [Issue 53963](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53963)


api/src/org/labkey/api/exp/api/ExpSampleType.java
```diff
@@ -242,6 +243,14 @@ String createSampleName(@NotNull Map<String, Object> rowMap,
 
     @NotNull Map<String, String> getImportAliases() throws IOException;
 
+    @NotNull
+    default Map<String, String> getImportAliasesIncludingAliquot() throws IOException
```
labkey-susanh, line 248: Add an issue number reference here? Or with one of these updates (not sure which is the most obvious), just so it's easier to find what's changed for this issue.

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
```diff
@@ -1370,6 +1370,32 @@ else if (name != null)
...
         }
 
+
+        @Override
+        protected Map<String, Object> updateRow(User user, Container container, Map<String, Object> row, @NotNull Map<String, Object> oldRow, boolean allowOwner, boolean retainCreation)
+                throws InvalidKeyException, ValidationException, QueryUpdateServiceException, SQLException
+        {
+            Map<String, Object> result = super.updateRow(user, container, row, oldRow, allowOwner, retainCreation);
+
+            // add MaterialInput/DataInputs field from parent alias
```
labkey-susanh, line 1380: I'm unclear why this wasn't necessary before. Can you enlighten me?

labkey-susanh (approved):

experiment/src/org/labkey/experiment/api/ExpDataClassDataTableImpl.java
XingY: There was a bug that row by row update using alias gets silently ignored. It's reproducible only from API, since our UI will never send the request using alias, and file import will never hit the row by row code. Due to its narrow repro, I didn't create a separate issue for it. In a way, one could argue ignoring import alias for query api  update is per design, since this feature was called importAliases.

PR #7106, opened 2025-10-07
Issue 52878: Attachment thumbnails from ancestor/lineage columns not rendered grids

XingY:
#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1863
- https://github.com/LabKey/labkey-ui-premium/pull/830
- https://github.com/LabKey/platform/pull/7106
- https://github.com/LabKey/limsModules/pull/1711

#### Changes
- show file icon for ancestor/lineage file columns
- retain properties based on the bounded column for multi-value columns


experiment/src/org/labkey/experiment/api/AncestorLookupDisplayColumn.java
```diff
@@ -120,6 +121,20 @@ public String renderURL(RenderContext ctx)
...
 
+    @Override
+    public void renderGridCellContents(RenderContext ctx, HtmlWriter out)
+    {
+        Integer lookupKey = getLookupId(ctx);
+
+        if (_dc != null && lookupKey != null && lookupKey > 0)
+        {
+            _dc.renderGridCellContents(ctx, out);
+            return;
```
cnathe, line 132: I don't know where this comment belongs (likely not related to this code changes), but when I tried to setup the case locally where I have a sample type grid and customize the view to pull in the "Input/Data/sourcetype/attachment column" case, the result was an empty grid rendering in the app:

Image

When I did this same setup on nightly, it rendered as described in the issue with all expected grid rows.

cnathe (changes requested):

experiment/src/org/labkey/experiment/api/AncestorLookupDisplayColumn.java
cnathe: I tried this again on develop and got the same behavior, so definitely unrelated to changes in this PR.
Turns out the issue was that the source type had an invalid calculated field. So when I tried to add it to any of the sample type grids, it had the same result. Removing that bad calculated field fixed things.

cnathe (approved):

PR #7107, opened 2025-10-07
Issue 53997: Establish a maximum size for query selections

labkey-nicka:
#### Rationale
This addresses [Issue 53997](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53997) by establishing a maximum query selection size of (M=100,000). Some things to note:

**API Changes**

1. The `query-selectAll.api` will now select up to the first M rows. If the underlying query contains more rows than the maximum, then the first M will be selected. If the underlying query contains less than M rows, then all rows will be selected.
2. The `query-setSelection.api` (named `QueryController.SetCheckAction`) will reject any selection that will result in a selection size over the maximum. When rejected the pre-existing selection will remain unchanged.

**UI Changes**

1. In the event that an application grid or a LKS data region is displaying a grid with more than M rows, then the "Select All" button will instead display "Select First M Rows".
2. If a user manually selects more than the maximum amount, then an error will be displayed inline in either the application grid or LKS data region.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1865
- https://github.com/LabKey/platform/pull/7107
- https://github.com/LabKey/limsModules/pull/1713
- https://github.com/LabKey/testAutomation/pull/2738

#### Changes
- Refactor selections to consistently make use of `LinkedHashSet` making comparison/iteration operations much faster. Formerly, we were converting between different structures which resulted in a lot more time composing/iterating (e.g., clearing a selection where the selection size is very large).
- Update client-side data region to render `Select First M Rows` button when applicable.
- Display error message in data region if selection is greater than M.
- Declare constant maximum selection size on server. Pass to client via module context property.


api/src/org/labkey/api/data/DataRegionSelection.java
```diff
@@ -268,14 +268,54 @@ public static int setSelected(ViewContext context, String key, Collection<String
...
+        if (checked && selection.size() > MAX_QUERY_SELECTION_SIZE)
+            throw new BadRequestException(selectionTooLargeMessage(selection.size()));
+
         Set<String> selectedValues = getSet(context, key, true, useSnapshot);
         if (checked)
+        {
+            // Verify that adding these selections will not result in a set that is too large
+            if (selectedValues.size() + selection.size() > MAX_QUERY_SELECTION_SIZE)
+            {
+                // Do not modify the actual selected values
```
labkey-jeckels:
```suggestion
                // Do not modify the actual selected values yet
```

api/src/org/labkey/api/data/DataRegionSelection.java
```diff
@@ -268,14 +268,54 @@ public static int setSelected(ViewContext context, String key, Collection<String
...
     public static int setSelected(ViewContext context, String key, Collection<String> selection, boolean checked, boolean useSnapshot)
     {
+        if (checked && selection.size() > MAX_QUERY_SELECTION_SIZE)
+            throw new BadRequestException(selectionTooLargeMessage(selection.size()));
+
         Set<String> selectedValues = getSet(context, key, true, useSnapshot);
         if (checked)
+        {
+            // Verify that adding these selections will not result in a set that is too large
+            if (selectedValues.size() + selection.size() > MAX_QUERY_SELECTION_SIZE)
```
labkey-jeckels: I think the whole body of this `if()` needs to be synchronized on `selectedValues`. See the note in `getSet()`. This looks like it'll be safe from deadlock potential.

labkey-jeckels (commented):
Maybe I'm misreading the code or expected behavior, but it seems like this won't be true if there are any existing rows selected that aren't in the first M rows:

`The query-selectAll.api will now select up to the first M rows`

I think `createSelectionSet()` will return 100,000 entries, which will then be rejected when merged with the existing selection in `setSelected()`.

Maybe that's fine, but it's a more muddled of a behavior to describe.

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -3544,8 +3550,12 @@ if (!LABKEY.DataRegions) {
 
     var _showSelectMessage = function(region, msg) {
         if (region.showRecordSelectors) {
-            if (region.totalRows && region.totalRows != region.selectedCount) {
-                msg += "&nbsp;<span class='labkey-button select-all'>Select All Rows</span>";
+            if (region.totalRows && region.totalRows !== region.selectedCount && region.selectedCount < MAX_SELECTION_SIZE) {
+                let text = 'Select All Rows';
+                if (region.totalRows > MAX_SELECTION_SIZE) {
+                    text = `Select First ${MAX_SELECTION_SIZE.toLocaleString()} Rows`;
```
cnathe: not sure where this code lives or comment applies, but if I have a current selection of N rows and then I click the "Select First 100,000 Rows" button, I guess I would have expected it to remove my current N row selection and replace it with the first 100k row selection. Instead, as shown in the screenshot on the other PR review, I get the error state because it tried to go beyond the max selection. Would that be possible to handle that case as a "replace" or does not match the proposed solution?

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -4038,8 +4048,8 @@ if (!LABKEY.DataRegions) {
         // If not all rows are visible and some rows are selected, show selection message
         if (region.totalRows && 0 !== region.selectedCount && !region.complete) {
             var msg = (region.selectedCount === region.totalRows) ?
-                        'All <span class="labkey-strong">' + region.totalRows + '</span> rows selected.' :
-                        'Selected <span class="labkey-strong">' + region.selectedCount + '</span> of ' + region.totalRows + ' rows.';
+                        'All <span class="labkey-strong">' + region.totalRows.toLocaleString() + '</span> rows selected.' :
+                        'Selected <span class="labkey-strong">' + region.selectedCount.toLocaleString() + '</span> of ' + region.totalRows.toLocaleString() + ' rows.';
```
cnathe, line 4107: not sure where this comment belongs, but in LKS DataRegion, the header checkbox menu has "Select All" as an option, even when it really will attempt to select only the first 100k. Should this say "Select First 100,000" in this case?

Image

labkey-nicka: Unfortunately, we don't know the result set size when rendering this menu as it is configured by the server. Options:

1. Leave it as "Select All".
2. Switch it to always be "Select First 100,000".
3. Dynamically switch out the rendered content with additional client-side processing.
4. Remove it.

I can investigate option 3, otherwise, I recommend we just leave it alone.

api/src/org/labkey/api/data/DataRegionSelection.java
labkey-nicka: Done.

api/webapp/clientapi/dom/DataRegion.js
labkey-nicka: I've opted to go with what you suggested and have it replace the selection with the first M selection. Take a look at data region interactions and see what you think. I've tried to work out the edge cases.

api/webapp/clientapi/dom/DataRegion.js
labkey-nicka: I went with option 3 but it has tendrils into nav tree rendering. Let me know what you think.

api/webapp/clientapi/dom/DataRegion.js
cnathe: Looks good

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -851,6 +852,10 @@ if (!LABKEY.DataRegions) {
         if (!_selDocClick) {
             _selDocClick = $(document).on('click', _onDocumentClick);
         }
+
+        if (_isShowSelectAll(this)) {
```
cnathe, line 857: maybe worth adding the issue number as a comment here and explain why this if statement is here

api/webapp/clientapi/dom/DataRegion.js
```diff
@@ -1035,16 +1040,26 @@ if (!LABKEY.DataRegions) {
...
-            else if (this.showRows === "unselected") {
+            else if (this.showRows === 'unselected') {
                 _removeParameters(this, [SHOW_ROWS_PREFIX]);
             }
-            else {
+            else if (this.totalRows <= MAX_SELECTION_SIZE) {
                 _toggleAllRows(this, true);
             }
+            else {
+                const lastRowIdx = this.offset + this.rowCount;
```
cnathe, line 1057: I was confused when I first read through this code, maybe worth a comment

cnathe (approved):

api/webapp/clientapi/dom/DataRegion.js
labkey-nicka: Done.

api/webapp/clientapi/dom/DataRegion.js
labkey-nicka: Done.

api/src/org/labkey/api/data/DataRegionSelection.java
```diff
@@ -268,14 +268,73 @@ public static int setSelected(ViewContext context, String key, Collection<String
...
+                            distinctAdds++;
+                    }
+
+                    int prospective = current + distinctAdds;
+                    if (prospective > MAX_QUERY_SELECTION_SIZE)
+                        throw new BadRequestException(selectionTooLargeMessage(prospective));
+                }
+            }
+
             selectedValues.addAll(selection);
```
labkey-jeckels: Might as well put this inside the `synchronized` block and make sure nobody else sneaks in

api/src/org/labkey/api/data/DataRegionSelection.java
```diff
@@ -339,23 +398,21 @@ public static void clearAll(ViewContext context)
...
         {
-            //noinspection SynchronizationOnLocalVariableOrMethodParameter
             synchronized (selection)
             {
                 items.forEach(selection::remove);
             }
         }
-        return Collections.unmodifiableList(items);
+
+        return items;
```
labkey-jeckels: We are now returning a mutable set, presumably for the convenience of `setSelected()`. There appear to be minimal callers of this method so perhaps not important, but it means that others may in the future mess with synchronized control of the collection. It's also different from the version that takes a selectionKey. Maybe this should be just for internal purposes?

labkey-jeckels (approved):

api/src/org/labkey/api/data/DataRegionSelection.java
labkey-nicka: Done.

api/src/org/labkey/api/data/DataRegionSelection.java
labkey-nicka: Remade into an unmodifiable set. I'd removed this while refactoring from a list and did not reintroduce the unmodifiable.

api/src/org/labkey/api/data/DataRegionSelection.java
```diff
@@ -207,11 +208,9 @@ public static String getSelectionKeyFromRequest(ViewContext context)
             values = context.getRequest().getParameterValues(DataRegion.SELECT_CHECKBOX_NAME);
         if (null != values && values.length == 1 && values[0].contains("\t"))
             values = StringUtils.split(values[0],'\t');
-        List<String> parameterSelected = values == null ? new ArrayList<>() : Arrays.asList(values);
-        Set<String> result = new LinkedHashSet<>(parameterSelected);
+        Set<String> result = values == null ? new LinkedHashSet<>() : new LinkedHashSet<>(Arrays.asList(values));
```
XingY, line 211: LinkedHashSet uses significantly more memory than ArrayList.

labkey-nicka: Understood but the `ArrayList` here was just interstitial. Mainly looking to go from a `String[]` to a `LinkedHashSet` as quickly as possible.

labkey-nicka: In general, the change here to stop converting to lists is significantly more performant since we end up doing set logic on these and conversions between the same types is much faster. Saw a significant reduction in processing time switching everything to using the same set structure (e.g., "Clear all" selection).

PR #7108, opened 2025-10-07
QueryForm(schema,table)

labkey-matthewb:
#### Rationale
We don't need to pass schema/table if they are determined by the target action.
Issue [54072](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=54072)

#### Related Pull Requests
- 

#### Changes
-


labkey-jeckels (approved):
I like making the actions more self-sufficient in terms of the initialization. Also consider breaking out the actual `new QueryForm()` call from the other initialization that's duplicated across overrides. Note that I didn't do any manual testing.

labkey-jeckels:
@labkey-matthewb please add this to one of the team boards so we give it a quick visit during standup

PR #7110, opened 2025-10-09
Handle container filtering on more tables

labkey-adam:
#### Rationale
- Address container filtering of `assaywell` and `expsampleset` provisioned schemas plus `exp.ObjectLegacyNames` and `exp.Alias`
- Introduce and use MigrationTableHandler to customize behavior of specific tables. Use it so Labbook can customize the core.Documents select query.
- Use DataIterator to copy data for performance improvement of 10x+
- Implement domain data filters on provisioned columns
- Support domain data filters on built-in Flag columns
- Introduce support for CaseInsensitiveHashSet collector and stable-order detection


experiment/src/org/labkey/experiment/ExperimentModule.java
```diff
@@ -874,34 +880,106 @@ SELECT COUNT(DISTINCT DD.DomainURI) FROM
...
+                        new InClause(FieldKey.fromParts("MaterialId", "Container"), containers),
+                        new InClause(FieldKey.fromParts("TargetApplicationId", "RunId", "Container"), containers)
+                    );
+                    case "Edge" -> new AndClause(
+                        new InClause(FieldKey.fromParts("FromObjectId", "Container"), containers),
+                        new InClause(FieldKey.fromParts("ToObjectId", "Container"), containers)
+                    );
+                    case "Alias" -> new SQLClause(
+                        new SQLFragment("RowId IN (SELECT Alias FROM exp.MaterialAliasMap WHERE Container IN ")
+                            .appendCsvList(containers, sourceTable.getSqlDialect())
```
labkey-matthewb: why not SqlDialect.appendInClauseSql()?

api/src/org/labkey/api/data/DatabaseMigrationService.java
```diff
@@ -162,9 +183,89 @@ public FilterClause getContainerClause(TableInfo sourceTable, FieldKey container
...
+                );
+            }
+        }
+
+        // Special domain data filter method for provisioned tables that have a built-in Flag field
+        protected void addDomainDataFlagFilter(OrClause orClause, DomainFilter filter, TableInfo sourceTable, FieldKey fKey, String guid, Set<String> selectColumnNames)
+        {
+            if (filter.column().equalsIgnoreCase("Flag"))
+            {
+                SQLFragment flagWhere = new SQLFragment("lsid IN (SELECT ObjectURI FROM exp.Object WHERE ObjectId IN (SELECT ObjectId FROM exp.ObjectProperty WHERE StringValue = ? AND PropertyId = (SELECT PropertyId FROM exp.PropertyDescriptor WHERE PropertyURI = 'urn:exp.labkey.org/#Comment')))", filter.condition().getParamVals());
```
labkey-matthewb: (SELECT PropertyId FROM exp.PropertyDescriptor WHERE PropertyURI = 'urn:exp.labkey.org/#Comment') is a constant.

experiment/src/org/labkey/experiment/ExperimentModule.java
labkey-matthewb: Presumably to reduce parameter markers?  Hopefully, appendInClauseSql() would have good default behavior.

api/src/org/labkey/api/data/DatabaseMigrationService.java
labkey-adam: Yep... I figured this should be performant, but I can factor out a separate call and use the constant

experiment/src/org/labkey/experiment/ExperimentModule.java
labkey-adam: Nope, just didn't remember this existed. I'll adjust.

api/src/org/labkey/api/data/DatabaseMigrationService.java
labkey-adam: We now query once, stash, and reuse the constant

experiment/src/org/labkey/experiment/ExperimentModule.java
labkey-adam: In fact, we have `SQLFragment.appendInClause()`, which calls the dialect and is easier to use with chaining. I switched and removed my unnecessary new method.

api/src/org/labkey/api/data/DatabaseMigrationService.java
labkey-matthewb: It _shouldn't_ matter... until you tip the PG optimizer over the edge.  I've just gotten in the habit of avoiding JOINS whenever possible.

labkey-matthewb (approved):

PR #7104, opened 2025-10-06
Issue 48864: Use CURRENT_TIMESTAMP instead of system time

labkey-nicka:
#### Rationale
This addresses [Issue 48864](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48864) and was spurred on by intermittent API test failures caused by discrepancies in server v. database time.

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1710

#### Changes
- Move marker interface `NowTimestamp` from `StatementUtils` to `SQLFragment`.
- Introduce `SQLFragment.appendValue(NowTimestamp)`
- Revise `Table.insert()` and `Table.update()` statement generation to use `NowTimestamp`.
- Remove explicit created/createdBy value setting for new audit events. Let underlying statement generation handle that.

PR #7114, opened 2025-10-10
Convert NowTimestamp to Timestamp

labkey-nicka:
#### Rationale
Serialization of `NowTimestamp` is not resolved in the same way as `java.sql.Timestamp` even though it extends `java.sql.Timestamp`. Rather than looking for `NowTimestamp` everywhere and hooking up custom converters I've elected to just convert back all `NowTimestamp` to `Timestamp` immediately when setting the property values in `Table`.

#### Related Pull Requests
- #7104

#### Changes
- Check for `NowTimestamp` instances and convert to `Timestamp` before setting property.


labkey-jeckels (approved):
Test is passing in this FB. I didn't manually test since it seems like the test is sufficient.

PR #7116, opened 2025-10-10
Permission validation for related issues

labkey-klum:
#### Rationale
[tracking issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53820)

If a user doesn't have permission to see issues in another folder, they can be added (or removed) as related issues from an issue they do have access to. We had checks to prevent adding related issues, but not removing them.

#### Related Pull Requests
https://github.com/LabKey/testAutomation/pull/2739


labkey-danield (approved):
This looks ok to me, but I think @labkey-matthewb is who you wanted to review this.

labkey-adam (approved):

PR #7117, opened 2025-10-13
Charting tool enhancements - Errors bars for Bar and Line chart

cnathe:
#### Rationale
See related ui-components PR for rationale. This PR 

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1869
- https://github.com/LabKey/limsModules/pull/1726
- https://github.com/LabKey/platform/pull/7117 
- https://github.com/LabKey/testAutomation/pull/2740

#### Changes
- Addition of error bar support for bar and line charts with vertical error bars
- New UI controls for aggregate method selection and error bar type configuration
- Enhanced statistical functions for calculating standard deviation and standard error of the mean


core/webapp/vis/src/utils.js
```diff
@@ -258,27 +256,58 @@ LABKEY.vis.getAggregateData = function(data, dimensionName, subDimensionName, me
...
+            }
         }
         else
         {
             throw 'Aggregate ' + aggregate + ' is not yet supported.';
         }
 
+        if (keepNames === undefined || keepNames === null) {
+            keepNames = false;
+        }
```
labkey-alan: This isn't necessary. The following line just looks at `if (keepNames)`, and null/undefined are falsey, we can also make `keepNames` have a default argument of false.

core/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -453,33 +453,82 @@ LABKEY.vis.internal.Axis = function() {
                 }
             }
         }
-        if (tickHover || tickClick || tickMouseOver || tickMouseOut) {
+
+        var hasTickAction = tickHover || tickClick || tickMouseOver || tickMouseOut;
```
labkey-alan: I know that a lot of the code in here was written prior to let/const existing, but we should be able to use const here. Previously we couldn't because our minification library didn't support it, but I believe we've since updated what we use for minification and let/const are now valid.

I'd like to see all new declarations use let or const.

core/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -453,33 +453,82 @@ LABKEY.vis.internal.Axis = function() {
...
+                                });
+                    }
+                } else {
+                    function wrapAxisTickLabel(text) {
+                        var width;
+                        text.each(function(v) {
+                            if (!width)
+                                width = scale(v) - grid.leftEdge;
+
+                            var text = d3.select(this),
```
labkey-alan: I understand that we use this type of formatting elsewhere in the file, but it makes this block of code a lot harder to read, and I don't think we should stick to this standard just because it was used in code we wrote a decade ago. I'd prefer to see us define variables in a way that requires less indentation, so one declaration per line. e.g.

```
let text = d3.select(this);
const words = text.text().split(/[\s]+/).reverse();
let word;
let line = [];
// and so on
```

core/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -453,33 +453,82 @@ LABKEY.vis.internal.Axis = function() {
...
+                                    words = text.text().split(/[\s]+/).reverse(),
+                                    word,
+                                    line = [],
+                                    lineNumber = 0,
+                                    lineHeight = 1.1, // ems
+                                    x = this.getAttribute("x"),
+                                    y = this.getAttribute("y"),
+                                    dy = 0,
```
labkey-alan: dy is always `0`, and lineHeight above is always `1.1`. Should we maybe use constants defined somewhere above?

core/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -453,33 +453,82 @@ LABKEY.vis.internal.Axis = function() {
...
+                                    tspan = text.text(null)
+                                            .append("tspan")
+                                            .attr("x", x)
+                                            .attr("y", y)
+                                            .attr("dy", dy + "em");
+
+                            while (word = words.pop()) {
```
labkey-alan:
```suggestion
                            for (const word of words) {
```

labkey-alan: `for...of` is available in all browsers we support.

labkey-alan: Also, I think this whole for block could be moved to a helper function somewhere else with the signature:

`function wrapTickText(words: string[], width: number, text): void` where `text` is whatever the expected type from `.call` is on line 529.

core/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -453,33 +453,82 @@ LABKEY.vis.internal.Axis = function() {
...
+
+                        addHighlightRects(textAnchors);
+                        textAnchors.selectAll('rect.highlight')
+                                .attr('transform', function (v)
+                                {
+                                    return 'rotate(' + tickOverlapRotation + ',' + textXFn(v) + ',' + textYFn(v) + ')';
+                                });
+                    }
+                } else {
+                    function wrapAxisTickLabel(text) {
```
labkey-alan: There's nothing in this block of code that tells me what it does, except the name of this function. I think we should add some comments in this block of code to explain what we're doing.

core/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -453,33 +453,82 @@ LABKEY.vis.internal.Axis = function() {
...
+                                {
+                                    return 'rotate(' + tickOverlapRotation + ',' + textXFn(v) + ',' + textYFn(v) + ')';
+                                });
+
+                        addHighlightRects(textAnchors);
+                        textAnchors.selectAll('rect.highlight')
+                                .attr('transform', function (v)
+                                {
+                                    return 'rotate(' + tickOverlapRotation + ',' + textXFn(v) + ',' + textYFn(v) + ')';
+                                });
```
labkey-alan:
```suggestion
                        const rotate = (v) => `rotate(${tickOverlapRotation}, ${textXFn(v)}, ${textYFn(v)})`;
                        addTickAreaRects(textAnchors);
                        textAnchors.selectAll("rect." + (tickRectCls ? tickRectCls : "tick-rect"))
                                .attr('transform', rotate);

                        addHighlightRects(textAnchors);
                        textAnchors.selectAll('rect.highlight')
                                .attr('transform', rotate);
```

labkey-alan: Let's use this opportunity to get rid of the duplicate code, which will also improve the formatting. Note: template literals (strings with backticks) are natively supported in all browsers we support.

core/webapp/vis/src/internal/D3Renderer.js
labkey-alan, line 523:
```suggestion

        if (orientation === 'bottom' && hasOverlap) {
            if (hasTickAction || tickOverlapRotation !== undefined || textEls[0].length > 10) {
                if (!tickOverlapRotation) tickOverlapRotation = 35;
                const rotate = (v) => `rotate(${tickOverlapRotation}, ${textXFn(v)}, ${textYFn(v)})`;

                textEls.attr('transform', rotate).attr('text-anchor', 'start');

                if (hasTickAction)
                {
                    addTickAreaRects(textAnchors);
                    textAnchors.selectAll("rect." + (tickRectCls ? tickRectCls : "tick-rect"))
                            .attr('transform', rotate);

                    addHighlightRects(textAnchors);
                    textAnchors.selectAll('rect.highlight').attr('transform', rotate);
                }
            } else {
                function wrapAxisTickLabel(text) {
                    var width;
                    text.each(function(v) {
                        if (!width) width = scale(v) - grid.leftEdge;

                        const textEl = d3.select(this);
                        const words = textEl.text().split(/[\s]+/).reverse();
                        let line = [];
                        let lineNumber = 0;
                        const lineHeight = 1.1; // ems
                        const x = this.getAttribute("x");
                        const y = this.getAttribute("y");
                        const dy = 0;
                        let tspan = textEl.text(null)
                                .append("tspan")
                                .attr("x", x)
                                .attr("y", y)
                                .attr("dy", dy + "em");

                        for (const word of words) {
                            line.push(word);
                            tspan.text(line.join(" "));
                            if (tspan.node().getComputedTextLength() > width) {
                                line.pop();
                                tspan.text(line.join(" "));
                                line = [word];
                                tspan = text.append("tspan")
                                        .attr("x", x)
                                        .attr("y", y)
                                        .attr("dy", ++lineNumber * lineHeight + dy + "em")
                                        .text(word);
                            }
                        }
                    });
                }

                textEls.attr('transform', '').call(wrapAxisTickLabel);
                textAnchors.selectAll('rect.highlight').attr('transform', '');
            }
        } else if (orientation === 'bottom') {
            textEls.attr('transform', '');
            textAnchors.selectAll('rect.highlight').attr('transform', '');
        }
```

labkey-alan: This is a suggestion based on all of my feedback above, in addition to the feedback above I also changed the if branch logic to start with `if (orientation === 'bottom' && hasOverlap) {` followed by an `if (ortientation === 'bottom')`, while this does duplicate the "bottom" check, it also allows us to reduce a level of nesting, making this code easier to read. I would still like to see some more comments in this section of the code though.

core/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -2165,38 +2220,40 @@ LABKEY.vis.internal.D3Renderer = function(plot) {
         }
     };
 
-    var renderErrorBar = function(layer, plot, geom, data) {
-        var colorAcc, sizeAcc, topFn, bottomFn, verticalFn, selection, newBars;
+    var renderErrorBar = function(layer, plot, geom, data, xAcc) {
+        var colorAcc, topFn, bottomFn, verticalFn, selection, newBars;
+        var errorLineWidth = geom.errorWidth != null ? geom.errorWidth : geom.width;
+        var xAcc_ = xAcc || function(row) {return geom.getX(row);};
```
labkey-alan: Remove the `var` line that defines a bunch of variables, and use const to declare `errorLineWidth` and `xAcc_`. Update the variables below to all be `const =` e.g. `const topFn =`, `const bottomFn =`, etc.
```suggestion
        const errorLineWidth = geom.errorWidth != null ? geom.errorWidth : geom.width;
        const xAcc_ = xAcc ?? row => geom.getX(row);
```

core/webapp/vis/src/internal/D3Renderer.js
```diff
@@ -2165,38 +2220,40 @@ LABKEY.vis.internal.D3Renderer = function(plot) {
...
 
         colorAcc = geom.colorAes && geom.colorScale ? function(row) {return geom.colorScale.scale(geom.colorAes.getValue(row) + geom.layerName);} : geom.color;
-        sizeAcc = geom.sizeAes && geom.sizeScale ? function(row) {return geom.sizeScale.scale(geom.sizeAes.getValue(row));} : geom.size;
```
labkey-alan, line 2172: Are we sure it's safe to remove this? Isn't it possible that there are users that have charts configured with `geom.sizeAes` that are now going to have their error bars fixed at 1 width?

core/webapp/vis/src/geom.js
```diff
@@ -340,6 +341,7 @@ LABKEY.vis.Geom.ErrorBar = function(config){
     this.dashed = ('dashed' in config && config.dashed != null && config.dashed != undefined) ? config.dashed : false;
     this.width = ('width' in config && config.width != null && config.width != undefined) ? config.width : 6;
     this.topOnly = ('topOnly' in config && config.topOnly != null && config.topOnly != undefined) ? config.topOnly : false;
+    this.errorShowVertical = ('showVertical' in config && config.showVertical != null && config.showVertical != undefined) ? config.showVertical : false;
```
labkey-alan: I think this could be:

`this.errorShowVertical = config.showVertical ?? : false;`

core/webapp/vis/src/plot.js
```diff
@@ -168,7 +168,7 @@
  * @param {Integer} [config.tickLength] (Optional) The length, in pixels for the x and y axis tick marks. Defaults to 8.
  * @param {Integer} [config.tickWidth] (Optional) The x and y axis tick line width. Defaults to 1.
  * @param {Integer} [config.tickOverlapRotation] (Optional) The degree of rotation for overlapping x axis tick labels.
- *      Defaults to 15 degrees.
+ *      Defaults to 35 degrees.
```
labkey-alan, line 171: Lol, good thing you updated this, because the default was actually 25 degrees.

core/webapp/vis/src/statistics.js
```diff
@@ -199,20 +199,50 @@ LABKEY.vis.Stat.MEAN = LABKEY.vis.Stat.getMean;
 /**
  * Returns the standard deviation.
  * @param values An array of numbers.
+ * @param sample If true, use (n-1) for the denominator of the final step. Defaults to false.
  * @returns {Number}
  */
-LABKEY.vis.Stat.getStdDev = function(values)
+LABKEY.vis.Stat.getStdDev = function(values, sample)
 {
+    if (sample === undefined || sample === null) sample = false;
```
labkey-alan: This conversion to false shouldn't be necessary for two reasons.

1) null and undefined are falsey, and the checks below are compatible with falsey values.
2) we can use default arguments, they are supported all major browsers that we support.
```suggestion
LABKEY.vis.Stat.getStdDev = function(values, sample = false) {
```

core/webapp/vis/src/statistics.js
```diff
@@ -199,20 +199,50 @@ LABKEY.vis.Stat.MEAN = LABKEY.vis.Stat.getMean;
...
+LABKEY.vis.Stat.SD = LABKEY.vis.Stat.getStdDev;
+
+/**
+ * Returns the SEM (standard error of the mean).
+ * @param values An array of numbers.
+ * @param sample If true, use (n-1) for the denominator of the final step. Defaults to false.
+ * @returns {Number}
+ */
+LABKEY.vis.Stat.getStdErr = function(values, sample) {
+    if (sample === undefined || sample === null) sample = false;
```
labkey-alan:
```suggestion
LABKEY.vis.Stat.getStdErr = function(values, sample = false) {
```

core/webapp/vis/src/utils.js
```diff
@@ -258,27 +256,58 @@ LABKEY.vis.getAggregateData = function(data, dimensionName, subDimensionName, me
...
+        if (keepNames) {
+            // if the value was/is a number, convert it back so that the axis domain min/max calculate correctly
+            var dimValue = row['label'];
+            row[dimensionName] = { value: !isNaN(Number(dimValue)) ? Number(dimValue) : dimValue };
+
+            row[measureName] = { value: row['value'] };
+            row[measureName].aggType = aggregate;
+            if (row.hasOwnProperty('subLabel')) {
```
labkey-alan:
```suggestion
            var dimValue = row['label'];
            row[dimensionName] = { value: !isNaN(Number(dimValue)) ? Number(dimValue) : dimValue };
            row[measureName] = { value: row['value'] };
            row[measureName].aggType = aggregate;

            if (row.hasOwnProperty('subLabel')) {
```

core/webapp/vis/src/utils.js
```diff
@@ -258,27 +256,58 @@ LABKEY.vis.getAggregateData = function(data, dimensionName, subDimensionName, me
...
         {
             throw 'Aggregate ' + aggregate + ' is not yet supported.';
         }
 
+        if (keepNames === undefined || keepNames === null) {
+            keepNames = false;
+        }
+        if (keepNames) {
+            // if the value was/is a number, convert it back so that the axis domain min/max calculate correctly
+            var dimValue = row['label'];
```
labkey-alan: Is there a reason we use square bracket accessors e.g. `row['label']` in this block of code instead of something like `row.label` or `row.errorType`?

visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -130,7 +130,9 @@ LABKEY.vis.GenericChartHelper = new function(){
                     ? properties[0].aggregate : properties.aggregate;
 
             if (LABKEY.Utils.isDefined(aggregateProps)) {
-                var aggLabel = LABKEY.Utils.isObject(aggregateProps) ? aggregateProps.name : LABKEY.Utils.capitalize(aggregateProps.toLowerCase());
+                var aggLabel = LABKEY.Utils.isObject(aggregateProps)
+                        ? (aggregateProps.name != null ? aggregateProps.name : aggregateProps.label)
```
labkey-alan: I think you can use `??` here and avoid a nested ternary.
```suggestion
                        ? (aggregateProps.name ?? aggregateProps.label)
```

visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -1060,7 +1067,7 @@ LABKEY.vis.GenericChartHelper = new function(){
             }
 
             if (!scales.y.domain) {
-                var values = $.map(data, function(d) {return d.value;}),
+                var values = $.map(data, function(d) {return d.value + (d.error != null ? d.error : 0);}),
```
labkey-alan:
```suggestion
                var values = $.map(data, d => d.value + (d.error ?? 0)),
```

visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -1367,10 +1382,8 @@ LABKEY.vis.GenericChartHelper = new function(){
                 }
             });
 
-            if (_willRotateXAxisTickText(scales, plotConfig, maxLen, data)) {
-                // min bottom margin: 50, max bottom margin: 150
```
labkey-alan, line 1371: Is it intentional to no longer have a maximum bottom margin?

visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -1828,6 +1841,45 @@ LABKEY.vis.GenericChartHelper = new function(){
         LABKEY.Query.MeasureStore.selectRows(queryConfig);
     };
 
+    var generateDataForChartType = function(chartConfig, chartType, geom, data) {
+        var dimName = null;
+        var subDimName = null;
+        var measureName = null;
+        var aggType = chartType === 'bar_chart' || chartType === 'pie_chart' ? 'COUNT' : null;
+        var aggErrorType = null;
```
labkey-alan:
```suggestion
        let dimName = null;
        let subDimName = null;
        let measureName = null;
        let aggType = chartType === 'bar_chart' || chartType === 'pie_chart' ? 'COUNT' : null;
        let aggErrorType = null;
```

visualization/resources/web/vis/genericChart/genericChartHelper.js
```diff
@@ -1828,6 +1841,45 @@ LABKEY.vis.GenericChartHelper = new function(){
...
+            else if (measureName != null && (chartType === 'bar_chart' || chartType === 'pie_chart')) {
+                // default to SUM for bar and pie charts
+                aggType = 'SUM';
+            }
+        }
+
+        if (aggType) {
+            data = LABKEY.vis.getAggregateData(data, dimName, subDimName, measureName, aggType, '[Blank]', false, aggErrorType, chartType === 'line_plot');
+            if (aggErrorType) {
+                geom.errorAes = { getValue: function(d){ return d.error } };
```
labkey-alan:
```suggestion
                geom.errorAes = { getValue: d => d.error };
```

labkey-alan (changes requested):

core/webapp/vis/src/utils.js
cnathe: done

core/webapp/vis/src/internal/D3Renderer.js
cnathe: done

core/webapp/vis/src/internal/D3Renderer.js
cnathe: done

core/webapp/vis/src/internal/D3Renderer.js
cnathe: done

core/webapp/vis/src/internal/D3Renderer.js
cnathe: done

core/webapp/vis/src/geom.js
cnathe: done

core/webapp/vis/src/statistics.js
cnathe: done

core/webapp/vis/src/statistics.js
cnathe: done

core/webapp/vis/src/utils.js
cnathe: I'm not sure I'm seeing the suggested change here. Do you want me to remove the empty line at 297?

core/webapp/vis/src/utils.js
cnathe: no reason other than to keep things consistent within the function. I've replaced the square bracket usages.

visualization/resources/web/vis/genericChart/genericChartHelper.js
cnathe: done

visualization/resources/web/vis/genericChart/genericChartHelper.js
cnathe: done

visualization/resources/web/vis/genericChart/genericChartHelper.js
cnathe: done

visualization/resources/web/vis/genericChart/genericChartHelper.js
cnathe: done

core/webapp/vis/src/internal/D3Renderer.js
cnathe: I suppose I could but that seems unnecessary. The variables/constants are only used in this function.

core/webapp/vis/src/internal/D3Renderer.js
cnathe: I changed to the for loop instead of the while, but splitting out the `function wrapTickText` seems unnecessary as it is only a few lines and only used in this one place.

core/webapp/vis/src/internal/D3Renderer.js
cnathe: done

core/webapp/vis/src/internal/D3Renderer.js
cnathe: done

core/webapp/vis/src/internal/D3Renderer.js
cnathe: yes, I guess it is possible that someone was setting this size for the error bar....but why? That seems odd that they'd actually want something other than 1 for this size. I can add it back if you feel strongly.

visualization/resources/web/vis/genericChart/genericChartHelper.js
cnathe: yes it was intentional, but after review it doesn't hurt to leave in that max. Done

core/webapp/vis/src/internal/D3Renderer.js
labkey-alan: I don't think I feel strongly, just wanted to make sure we considered it. I think it's safe to commit this with your change, and if someone does need this functionality we can re-add it.

core/webapp/vis/src/utils.js
labkey-alan: yes, it's very nitpicky, but I think we don't need the empty line between the value assignments, but we should have an empty line before the if statement. I think it's easier to read that way.

labkey-alan (approved):

core/webapp/vis/src/utils.js
cnathe: done

PR #7101, opened 2025-10-03
Show Plate Samples on Sample Assays Page

labkey-alan:
#### Rationale
This PR updates our server side code to mark samples on Plate Wells as MaterialInputs when importing assay data for plate enabled assays. This allows us to show plate assay data for samples on the Sample details page in our apps.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1860
- https://github.com/LabKey/labkey-ui-premium/pull/828
- https://github.com/LabKey/platform/pull/7101
- https://github.com/LabKey/limsModules/pull/1705

#### Changes
- AbstractAssayTsvDataHandler: Mark plate well samples as assay run materials inputs during checkData
- AssaySampleLookupContext: Clean up the material inputs during assay data row delete
- AssayPlateMetadataService
    - Add getWellLocationToSampleIdMap
    - Add isWellLookup
- Bump ui-components

PR #7124, opened 2025-10-15
Improve plate assay import performance

labkey-alan:
#### Rationale
This PR slightly optimizes the way we query for well samples, and preloads all exp materials for a plate into the `materialCache` during import, which improves performance.

Benchmarks from my machine show that on develop import time is approximately 58% slower than 25.10. With the changes in this PR import time is approximately 24% slower.

Assay import time is expected to be slower by some amount, because when marking well samples as Material Inputs we are inserting at least 3x the amount of rows into the database, and we are also making additional queries to the Well table during import (in order to determine what samples to mark as inputs).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7101

#### Changes
- AssayPlateMetadataServiceImpl.getWellLocationToSampleIdMap: select only the columns we need when querying the well table
- AbstractAssayTsvDataHandler.checkData: preload all exp materials for a plate into the `materialCache`

### DIFF
--- a/assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
+++ b/assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
@@ -1518,14 +1518,22 @@ public class AssayPlateMetadataServiceImpl implements AssayPlateMetadataService
         return new PlateSchema(querySchema, contextualRoles);
     }
 
-    @Override
-    public Map<String, Long> getWellLocationToSampleIdMap(Container container, User user, Long plateId)
-    {
-        Map<String, Long> wellLocationToSampleIdMap = new HashMap<>();
-        List<WellData> wellData = PlateManager.get().getWellData(container, user, plateId, true, false);
+    record WellSampleData(Long sampleId, Integer row, Integer col) {}
 
-        for (WellData data : wellData)
-            wellLocationToSampleIdMap.put(data.getPosition(), data.getSampleId());
+    @Override
+    public Map<String, Long> getWellLocationToSampleIdMap(Long plateId)
+    {
+        // Note: this method intentionally does not use PlateManager.get().getWellData, by selecting only the columns
+        // we need there is a small but measurable performance boost when importing plate assay data
+        SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("plateid"), plateId);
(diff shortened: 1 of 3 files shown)


assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -1518,14 +1518,22 @@ public String format(FieldKey fieldKey)
...
 
+    record WellSampleData(Long sampleId, Integer row, Integer col) {}
+
     @Override
-    public Map<String, Long> getWellLocationToSampleIdMap(Container container, User user, Long plateId)
+    public Map<String, Long> getWellLocationToSampleIdMap(Long plateId)
     {
+        // Note: this method intentionally does not use PlateManager.get().getWellData, by selecting only the columns
+        // we need there is a small but measurable performance boost when importing plate assay data
+        SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("plateid"), plateId);
```
labkey-nicka: nit: There are constants for these columns.

```java
SimpleFilter filter = new SimpleFilter(WellTable.Column.PlateId.fieldKey(), plateId);
Set<String> columns = Set.of(WellTable.Column.SampleID.name(), WellTable.Column.Row.name(), WellTable.Column.Col.name());
List<WellSampleData> wells = new TableSelector(AssayDbSchema.getInstance().getTableInfoWell(), columns, filter, null).getArrayList(WellSampleData.class);
```

assay/src/org/labkey/assay/plate/AssayPlateMetadataServiceImpl.java
```diff
@@ -1518,14 +1518,22 @@ public String format(FieldKey fieldKey)
         return new PlateSchema(querySchema, contextualRoles);
     }
 
+    record WellSampleData(Long sampleId, Integer row, Integer col) {}
+
     @Override
-    public Map<String, Long> getWellLocationToSampleIdMap(Container container, User user, Long plateId)
+    public Map<String, Long> getWellLocationToSampleIdMap(Long plateId)
     {
+        // Note: this method intentionally does not use PlateManager.get().getWellData, by selecting only the columns
```
labkey-nicka, line 1526: My bad here with the misleading advice. The well table makes some pretty complex queries in itself so it's not as direct as I had recalled.

labkey-nicka (approved):

PR #7130, opened 2025-10-16
Expose more PropertyDescriptor data in exp.Fields

labkey-jeckels:
#### Rationale
In order to find fields of type Flag, we need to expose the ConceptURI.

#### Changes
- Add ConceptURI to `exp.Fields` and `exp.PhiFields`
- Expose all of the other columns as hidden columns

#### Tasks 📍
- [x] Manual Testing @labkey-adam 
  - Tested both Fields and PhiFields, verified ConceptURI looked correct, added all hidden columns, applied container filters (e.g., All Folders), filtered/sorted on various columns
- [x] Needs Automation - Not Needed


labkey-jeckels:
Are any of these columns sensitive enough to avoid exposing, or only to certain users? I didn't see anything concerning when I scanned the list.

experiment/src/org/labkey/experiment/api/BaseFieldsTable.java
```diff
@@ -32,6 +34,18 @@ public BaseFieldsTable(String tableName, @NotNull ExpSchema userSchema, @Nullabl
         addWrapColumn(_rootTable.getColumn("Description"));
         addWrapColumn(_rootTable.getColumn("RangeURI"));
         addWrapColumn(_rootTable.getColumn("StorageColumnName"));
+        addWrapColumn(_rootTable.getColumn("ConceptURI"));
+
+        // Add the other columns in case they're useful for future troubleshooting, but keep them out
+        // of the way by default
+        for (ColumnInfo column : _rootTable.getColumns())
```
labkey-adam, line 41: Looks like you'll need to adjust this loop to exclude "PHI" or adjust PhiFieldsTable to not add the column (but adjust it to be shown by default).

labkey-adam (approved):
Need to fix PhiFieldsTable to not blow up, but no need for follow-on review

PR #7132, opened 2025-10-17
ObjectFactory fails to map reserved field names

XingY:
#### Rationale
A recent change to how file fields are displayed in Sample Timeline revealed a flaw in our audit log query utils. Specifically, AuditLogService.getAuditEvents fails to populate the event beans if any of the event field name happens to be in the sql dialect's reservedWordSet. For example, "File" is a reserved word for SQL Server, but not for postgresDB. When a column is a reserved word, it's alias to "col_", "col_1"... in the select sql. The ResultSet uses "col_" as the column name, and has no knowledge of its original name, which results in bean property mapping failure. 

This PR adds a workaround to allow mapping ignoring trailing "_". It also logs warning about reserved word being added to domain. 

Something to consider for future work: the set of reserved word is large, and contains words like "case", "primary" that seems harmless to use as domain fields, so it's not practical to disallow them in user domains. 

[SMSampleTimelineBasicTest](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SmFeatures_BvtSampleManagerSqlserver/3695275?buildTab=tests&status=failed&suite=org.labkey.test.Runner%3A+&package=org.labkey.test.tests.samplemanagement.timeline&class=SMSampleTimelineBasicTest).[testFileFields](https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_SampleManager_SmFeatures_BvtSampleManagerSqlserver/3695275?buildTab=tests&status=failed&expandedTest=build%3A%28id%3A3695275%29%2Cid%3A2000000007) (MSSQL)
NPE when loading sample timeline with file field changes

#### Related Pull Requests
- https://github.com/LabKey/limsModules/pull/1643

#### Changes
- attempt to resolve bean field mapping by stripping trailing "_"
- add warning about use of reserved word for domain

### DIFF
--- a/api/src/org/labkey/api/exp/property/DomainUtil.java
+++ b/api/src/org/labkey/api/exp/property/DomainUtil.java
@@ -35,6 +35,7 @@ import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerFilter;
 import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.ContainerService;
+import org.labkey.api.data.CoreSchema;
 import org.labkey.api.data.NameGenerator;
 import org.labkey.api.data.PHI;
 import org.labkey.api.data.PropertyStorageSpec;
@@ -1484,6 +1485,9 @@ public class DomainUtil
                     continue;
                 }
             }
+
+            if (CoreSchema.getInstance().getSqlDialect().isReserved(name))
+                LOG.warn("Field name '" + name + "' is a reserved word in the current SQL dialect.");
 
             if (namePropertyIdMap.containsKey(name))
             {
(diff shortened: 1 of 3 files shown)


api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -1485,6 +1486,9 @@ public static ValidationException validateProperties(@Nullable Domain domain, @N
                 }
             }
 
+            if (CoreSchema.getInstance().getSqlDialect().isReserved(name))
```
labkey-matthewb: If this is supported, then I would not include these warnings (or lower to DEBUG).  Customers will see these and ask us how to resolve the problem.  If this is not fully supported...

labkey-matthewb (approved):

api/src/org/labkey/api/exp/property/DomainUtil.java
XingY: Good point. log removed for user defined domains.

PR #7115, opened 2025-10-10
Improve file path validation

labkey-jeckels:
#### Rationale
We have APIs that let us do stronger validation on input file paths. We can use FileLike in more assay interfaces, and use the simpler FileUtil methods as a drop-in replacement for the java.io.File constructor

#### Changes
- Switch from File to FileLike
- Use FileUtil.appendName() and appendPath() to ensure no path traversal

#### Tasks 📍
- [x] Manual Testing @labkey-danield 📍 
- [x] Needs Automation - Not needed

PR #7135, opened 2025-10-20
Don't always fail test for missing sample data

labkey-tchad:
#### Rationale
Server-side JUnit tests should be skipped if they require sampledata from a module but that module's source isn't present.

#### Related Pull Requests
- #7115 

#### Changes
- Don't fail tests when module source isn't present


labkey-tchad:
Verifying on TeamCity: https://teamcity.labkey.org/buildConfiguration/bt3/3706256

labkey-jeckels (approved):

PR #7147, opened 2025-10-24
Extract copySourceTableToTargetTable() and add to interface

labkey-adam:
#### Rationale
Refactor to allow handlers to copy data using the optimized code path. `DataClassMigrationSchemaHandler` uses new entry point to populate `biologics.SequenceIdentity` table after all data classes have been populated.

Fix mistaken StatementDataIterator `_batchSize` default calculation

### DIFF
--- a/api/src/org/labkey/api/dataiterator/StatementDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/StatementDataIterator.java
@@ -214,7 +214,7 @@ public class StatementDataIterator extends AbstractDataIterator
         _currentBinding = _bindings[0];
 
         if (_batchSize < 1 && null == _rowIdIndex && null == _objectIdIndex && null == _objectUriIndex)
-            _batchSize = Math.max(10, 10000/Math.max(2, _bindings[0].length));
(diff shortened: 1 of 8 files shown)


api/src/org/labkey/api/dataiterator/StatementDataIterator.java
```diff
@@ -214,7 +214,7 @@ void init()
         _currentBinding = _bindings[0];
 
         if (_batchSize < 1 && null == _rowIdIndex && null == _objectIdIndex && null == _objectUriIndex)
-            _batchSize = Math.max(10, 10000/Math.max(2,_bindings.length));
+            _batchSize = Math.max(10, 10000/Math.max(2, _bindings[0].length));
```
labkey-nicka: Use `_currentBinding`?

labkey-nicka (approved):

api/src/org/labkey/api/dataiterator/StatementDataIterator.java
labkey-adam: Yes, good call!

PR #7150, opened 2025-10-27
Issue 53831: check the true assay name length since we append onto the name when creating the assay domains (ex. "<assay name> Batch Fields")

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53831

This PR doesn't fix or update anything on the server side to actually allow for assay names of the full 200 characters. This instead adds a check for what is currently the "true assay name max length" at assay creation time.

Note that this definitely feels like a stop gap. If we really want to allow assay names that are 200 characters and not this arbitrary length, we should put that on the backlog and figure out an approach that works for assays but probably applies to all data types so they all have the same name max length.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7150
- https://github.com/LabKey/testAutomation/pull/2764

#### Changes
- add assay name specific length check in AssayDomainServiceImpl

#### Tasks 📍
- [x] Needs Automation
- [x] Verify Fix

### DIFF
--- a/api/src/org/labkey/api/exp/property/DomainUtil.java
+++ b/api/src/org/labkey/api/exp/property/DomainUtil.java
@@ -1018,12 +1018,6 @@ public class DomainUtil
                 return prefix + invalidPattenError;
         }
 
-        // Issue 53831: add a specific check for assay name length since we append onto the name when creating the assay domains (ex. "<assay nam> Batch Fields")
-        // which makes that actual max less than the DB size of 200
-        int assayNameLengthMax = 150;
-        if ("Assay Design".equalsIgnoreCase(kindName) && domainName.length() > assayNameLengthMax)
-            return "Value is too long for assay design name, a maximum length of " + assayNameLengthMax + " is allowed. The supplied value, '" + StringUtils.abbreviateMiddle(domainName, "...", 50) + "', was " + domainName.length() + " characters long.";
-
         return null;
     }
 
--- a/api/src/org/labkey/api/exp/property/DomainUtil.java
+++ b/api/src/org/labkey/api/exp/property/DomainUtil.java
@@ -1018,6 +1018,11 @@ public class DomainUtil
                 return prefix + invalidPattenError;
         }
 
+        // Issue 53831: check the true assay name length since we append onto the name when creating the assay domains (ex. "<assay nam> Batch Fields")
+        int actualAssayNameLengthMax = 186;
+        if ("Assay Design".equalsIgnoreCase(kindName) && domainName.length() > actualAssayNameLengthMax)
+            return "Value is too long for assay design name, a maximum length of " + actualAssayNameLengthMax + " is allowed. The supplied value, '" + StringUtils.abbreviateMiddle(domainName, "...", 50) + "', was " + domainName.length() + " characters long.";
+
         return null;
     }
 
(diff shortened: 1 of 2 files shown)


api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -1018,6 +1018,11 @@ public static ValidationException updateDomainDescriptor(GWTDomain<? extends GWT
                 return prefix + invalidPattenError;
         }
 
+        // Issue 53831: check the true assay name length since we append onto the name when creating the assay domains (ex. "<assay nam> Batch Fields")
+        int actualAssayNameLengthMax = 186;
```
labkey-jeckels: Is this the best place for this check? From what I can see, `"Assay Design"` is only passed as `kindName` in the call from `AssayDomainServiceImpl`. Maybe that should do the check there instead, which would also save us from having the kind name hard-coded in multiple places?

Also, perhaps we should just cap this at 150 or similar instead of trying to eke out every possible character. The prior sentence is 110 characters, for example, and would be miserably long as a design name.

labkey-jeckels (approved):

api/src/org/labkey/api/exp/property/DomainUtil.java
cnathe: To your first point: I originally had it within AssayDomainServiceImpl but moved it to this location in DomainUtil so that it would be hit while called in the TestDataGenerator.isNameValidRemote() code (which calls property-validateDomainAndFieldNames.api). @labkey-tchad I know you were moving away from that api call to the local test check. Do you know when / if this isNameValidRemote() is still being called?

cnathe: To the second item: I totally agree and almost went that route. I can definitely change this to something like 150 chars as the max. Some "round" number would be much more explainable to the users. Anyone have strong feelings on what that number should be? If not, I'll go with 150.

labkey-tchad: Currently, `isNameValidRemote` only gets called if you run the test with `webtest.remote.domain.validation=true`. Nothing on TeamCity runs that way.

api/src/org/labkey/api/exp/property/DomainUtil.java
```diff
@@ -1018,6 +1018,12 @@ public static ValidationException updateDomainDescriptor(GWTDomain<? extends GWT
                 return prefix + invalidPattenError;
         }
 
+        // Issue 53831: add a specific check for assay name length since we append onto the name when creating the assay domains (ex. "<assay nam> Batch Fields")
```
labkey-tchad: typo
```suggestion
        // Issue 53831: add a specific check for assay name length since we append onto the name when creating the assay domains (ex. "<assay name> Batch Fields")
```

labkey-tchad (approved):

api/src/org/labkey/api/exp/property/DomainUtil.java
cnathe: in that case I will move this code back to AssayDomainServiceImpl

api/src/org/labkey/api/exp/property/DomainUtil.java
cnathe: done

PR #7152, opened 2025-10-27
Issue 53567: Improve "Alias" column performance

labkey-nicka:
#### Rationale
This addresses [Issue 53567](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53567) by refactoring the "Alias" column for data classes and sample types away from the traditional multi-value foreign key (MVFK) to an inline join on the values.

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1875
- https://github.com/LabKey/platform/pull/7152
- https://github.com/LabKey/limsModules/pull/1759

#### Changes
- Alias column is now an array of string values.
- Alias is no longer a lookup column.
- Introduce `ColumnInfo.isMultiValued()` to allow columns to declare they are multi-value. Formerly, everything needed to be an instance of `MultiValueForeignKey`.
- Refactor group concat sql generation for postgres to use `string_agg()` and `::text`. Reduces per item processing.
- Consolidate duplicated logic for coercing type values in query update services.
- Migrate regression test for [Issue 43241](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43241).

### DIFF
--- a/api/src/org/labkey/api/dataiterator/CoerceDataIterator.java
+++ b/api/src/org/labkey/api/dataiterator/CoerceDataIterator.java
@@ -65,7 +65,7 @@ public class CoerceDataIterator extends SimpleTranslator
                 seen.add(to.getName());
                 if (to.getPropertyType() == PropertyType.ATTACHMENT || to.getPropertyType() == PropertyType.FILE_LINK)
                     addColumn(to, i);
-                else if (to.getFk() instanceof MultiValuedForeignKey)
+                else if (to.isMultiValued() || to.getFk() instanceof MultiValuedForeignKey)
                     addColumn(to.getName(), i); // pass-through multi-value columns -- converting will stringify a collection
                 else
                     addConvertColumn(to.getName(), i, to.getJdbcType(), to.getFk(), RemapMissingBehavior.Error, true);
--- a/experiment/src/org/labkey/experiment/api/AliasDisplayColumnFactory.java
+++ b/experiment/src/org/labkey/experiment/api/AliasDisplayColumnFactory.java
@@ -6,6 +6,7 @@ import org.labkey.api.data.DisplayColumn;
 import org.labkey.api.data.DisplayColumnFactory;
 import org.labkey.api.data.MultiValuedDisplayColumn;
 import org.labkey.api.data.RenderContext;
+import org.labkey.api.util.PageFlowUtil;
 
 import java.util.List;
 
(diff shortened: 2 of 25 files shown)


experiment/src/org/labkey/experiment/api/AliasDisplayColumnFactory.java
```diff
@@ -6,43 +6,26 @@
...
-                            String name = new TableSelector(ExperimentService.get().getTinfoAlias(), Collections.singleton("Name")).getObject(item, String.class);
-
-                            sb.append(delim);
-                            sb.append(name);
-                            delim = ", ";
-                        }
-                    }
-                }
-                return sb.toString();
+                    return String.join(", ", (List) value);
```
labkey-matthewb: I see this isn't new, but I really don't like it.  This should be a robustly round-trippable format.  In https://docs.google.com/document/d/1oXkzxDD7hIQy9bSp3KQ9pjG69kdcoCoRPRHT4y759fo/edit?tab=t.0#heading=h.xa25n05t4ysx I proposed using the same format Google Sheets uses for multi-choice fields.  The closest thing we have is PageFlowUtil.joinValuesToStrin()/splitStringToValues().  But it is not Sheets compatible.  We should write a new method maybe PageFlowUtil.joinValuesForExport()/splitValuesForImport().

labkey-nicka: `AliasDisplayColumnFactory` now makes use of `PageFlowUtil.joinValuesToStringForExport()`.

api/src/org/labkey/api/dataiterator/CoerceDataIterator.java
```diff
@@ -65,7 +65,7 @@ void init(TableInfo target, boolean useImportAliases, boolean outputAllColumns)
                 seen.add(to.getName());
                 if (to.getPropertyType() == PropertyType.ATTACHMENT || to.getPropertyType() == PropertyType.FILE_LINK)
                     addColumn(to, i);
-                else if (to.getFk() instanceof MultiValuedForeignKey)
+                else if (to.isMultiValued() || to.getFk() instanceof MultiValuedForeignKey)
                     addColumn(to.getName(), i); // pass-through multi-value columns -- converting will stringify a collection
```
labkey-matthewb, line 69: Hmm, yes ConvertUtils.convert() is bad this way...  Maybe, we need to stop using it directly?

labkey-matthewb (approved):

PR #7155, opened 2025-10-27
Adjust fetch size. Purge exp.* rows.

labkey-adam:
#### Rationale
Provide the ability to override the default ResultSet fetch size. This is important when each row could be very large.

Delete from exp.Data, exp.Material, exp.Object, and associated tables after populating sample types and data classes.

### DIFF
--- a/experiment/src/org/labkey/experiment/SampleTypeMigrationSchemaHandler.java
+++ b/experiment/src/org/labkey/experiment/SampleTypeMigrationSchemaHandler.java
@@ -98,13 +98,8 @@ class SampleTypeMigrationSchemaHandler extends DefaultMigrationSchemaHandler
 
         if (getJoinColumnName(sourceTable).equals("LSID"))
         {
-            // SELECT RowId FROM exp.Material m WHERE lsid IN (SELECT lsid FROM expsampleset.c3258d9570_dinosaurs WHERE ...)
-            selector = new SqlSelector(getSchema(), new SQLFragment("SELECT RowId FROM exp.Material m WHERE lsid IN (SELECT lsid FROM ")
-                .appendIdentifier(sourceTable.getSelectName())
-                .append(" WHERE ")
-                .append(notCopiedFilter.getSQLFragment(dialect))
-                .append(")")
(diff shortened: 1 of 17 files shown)


experiment/src/org/labkey/experiment/SampleTypeMigrationSchemaHandler.java
```diff
@@ -88,9 +93,68 @@ private String getJoinColumnName(TableInfo sourceTable)
...
+                .append(notCopiedFilter.getSQLFragment(dialect))
+                .append(")")
+            );
+        }
+        else
+        {
+            selector = new TableSelector(sourceTable, Collections.singleton("RowId"), notCopiedFilter, null);
+        }
+
+        Collection<Integer> notCopiedRows = selector.getCollection(Integer.class);
```
labkey-matthewb, line 109: Are these lists big?  Could consider SELECT INTO temp table.

labkey-matthewb (approved):

experiment/src/org/labkey/experiment/SampleTypeMigrationSchemaHandler.java
labkey-matthewb: Actually you could mayve accumulate these not copied values in afterTable() and do the DELETEs once per exp.table.  (e.g. instead of exp.tables X samples.tables)

labkey-adam: Yeah, that's the pattern I use for SequenceIdentity, though accumulating the unique sequences in memory in that case. I'll keep an eye on the sizes, but I think they've been in the 10s of thousands of integers... so not concerning.

labkey-adam: Also note that these IN clauses (once they get > 1,000 items) are automatically backed by temp tables with an index on the values. I get that accumulating them all into a temp table and doing one delete per table would have benefits, but current code works pretty well without any direct temp table interactions.

PR #7159, opened 2025-10-28
Remove extends DynaBean 

labkey-matthewb:
#### Rationale
Simplify TableViewForm with an eye toward making array binding less convoluted

#### Related Pull Requests
- https://github.com/LabKey/premiumModules/pull/398

#### Changes
-


api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -67,13 +69,14 @@
  * Supports insert, update, delete functionality with a minimum of fuss
  * <p/>
  */
-public class TableViewForm extends ViewForm implements DynaBean, HasBindParameters
+public class TableViewForm extends ViewForm implements HasBindParameters
 {
     private static final Logger _log = LogHelper.getLogger(TableViewForm.class, "Table operation warnings");
 
-    protected Map<String, String> _stringValues = new CaseInsensitiveHashMap<>();
+    // This is called "stringValues" as this is expected to come from a form POST (but it was never just string valure)
```
labkey-nicka: Typo: "valure" and "abd"

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -318,16 +296,22 @@ public void setPkVals(String s)
     {
         //Issue 42042: Lists with text primary key don't handle commas in key value when viewing row details
         if (getPkNamesList().size() == 1)
-            set(getPkNamesList().get(0), s);
+        {
+            setValueToBind(getPkNamesList().get(0), s);
+        }
         else
+        {
+            // TODO replace with new splitStringToValues
```
labkey-nicka: Should this be a `TODO MultiChoice`?

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -642,108 +640,64 @@ public void setTypedValues(Map<String, Object> values, boolean merge)
...
+    public Map<String, Object> getValuesToBind()
     {
         return _stringValues;
     }
 
-    public boolean contains(ColumnInfo col)
-    {
-        return _stringValues.containsKey(getFormFieldName(col));
-    }
+//    public boolean contains(ColumnInfo col)
```
labkey-nicka: Remove commented out code.

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -642,108 +640,64 @@ public void setTypedValues(Map<String, Object> values, boolean merge)
...
-    public DynaClass getDynaClass()
+    public String getAsString(ColumnInfo col)
     {
-        return _dynaClass;
+        return getAsString(getFormFieldName(col));
     }
 
-    @Override
-    public void remove(String arg0, String arg1)
+    public void setValueToBind(String propName, Object value)
```
labkey-nicka, line 688: Should this function be trimming? If not, then who performs the trimming?

api/src/org/labkey/api/util/HtmlString.java
```diff
@@ -30,6 +30,7 @@ public final class HtmlString implements SafeToRender, DOM.Renderable, Comparabl
 {
     // Helpful constants for convenience (and efficiency)
     public static final HtmlString EMPTY_STRING = HtmlString.of("");
+    public static final HtmlString SP = HtmlString.of(" ");
```
labkey-nicka: nit: Doesn't appear to be used.

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -642,108 +640,64 @@ public void setTypedValues(Map<String, Object> values, boolean merge)
...
+    public String getAsString(String propName)
     {
-        String v;
-        if (arg1 == null)
-            v = null;
-        else if (arg1 instanceof Object[])
+        Object value = _stringValues.get(propName);
+        if (value == null || value instanceof String)
+            return (String)value;
+        if (value instanceof String[])
```
labkey-nicka: nit: Mark as `@Nullable` and make use of the pattern variable:

```java
public @Nullable String getAsString(String propName)
{
    Object value = _stringValues.get(propName);
    if (value == null)
        return null;
    if (value instanceof String s)
        return s;
    if (value instanceof String[] arr && arr.length == 0)
        return null;
    return ConvertUtils.convert(value);
}
```

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -642,108 +640,64 @@ public void setTypedValues(Map<String, Object> values, boolean merge)
...
-        assert null != _dynaClass;
-
-        _stringValues = strings;
+        _stringValues.clear();
+        _stringValues.putAll(strings);
         _values = null;
     }
 
-    public Map<String, String> getStrings()
+    public Map<String, Object> getValuesToBind()
```
labkey-nicka, line 660: Should this be returning an immutable map?

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -642,108 +640,64 @@ public void setTypedValues(Map<String, Object> values, boolean merge)
...
-    @Override
-    public void set(String arg0, String arg1, Object arg2)
-    {
-        throw new UnsupportedOperationException("No mapped properties in a table");
+        if (null == value || value instanceof String || value instanceof String[])
+            _stringValues.put(propName, value);
+        else if (value instanceof Collection<?> col && col.stream().allMatch(e -> null==e || e instanceof String))
+            _stringValues.put(propName, col.toArray(new String[0]));
+        else
+            _stringValues.put(propName, ConvertUtils.convert(value));
```
labkey-nicka, line 695: Since we are converting going in (`setValueToBind`) then why do we need to convert again going out (`getAsString`)?

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -537,7 +534,9 @@ public boolean hasTypedValue(ColumnInfo column)
     public void setTypedValue(String propName, Object val)
     {
         getTypedValues().put(propName, val);
-        _stringValues.put(propName, ConvertUtils.convert(val));
+        // setValueToBind() but w/o invalidate.
```
labkey-nicka: Should this be a `TODO MultiChoice`?

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -861,11 +839,10 @@ public void setViewContext(@NotNull ViewContext context)
 
         for (PropertyValue pv : params.getPropertyValues())
         {
-            Object value = pv.getValue();
-            if (value instanceof String || value instanceof String[])
-                set(pv.getName(), value);
+            _stringValues.put(pv.getName(), pv.getValue());
```
labkey-nicka: Shouldn't this call `setValueToBind()` instead?

api/src/org/labkey/api/data/TableViewForm.java
```diff
@@ -642,108 +640,64 @@ public void setTypedValues(Map<String, Object> values, boolean merge)
...
     }
 
-    public void setStrings(Map<String, String> strings)
+    public void setValuesToBind(Map<String, Object> strings)
     {
-        assert null != _dynaClass;
-
-        _stringValues = strings;
+        _stringValues.clear();
+        _stringValues.putAll(strings);
```
labkey-nicka: Should this be calling `setValueToBind()` for each prop/value instead?

api/src/org/labkey/api/data/BeanViewForm.java
```diff
@@ -118,4 +122,68 @@ else if (o instanceof Map)
...
+    @Override
+    public Object get(String arg0, String arg1)
+    {
+        throw new UnsupportedOperationException("No mapped properties in a table");
+    }
+
+    @Override
+    public Object get(String arg0, int arg1)
+    {
+        throw new UnsupportedOperationException("No indexed properties in a table");
```
labkey-nicka, line 169: nit: Why do some of these say `No indexed properties...` and some say `No mapped properties...`?

api/src/org/labkey/api/data/BeanViewForm.java
```diff
@@ -83,19 +87,19 @@ public void setBean(K bean)
...
+        Map<String, Object> strings = super.getValuesToBind();
         if (null == strings || strings.isEmpty() && (null != _values && !_values.isEmpty()))
         {
             strings = new HashMap<>();
             for (Map.Entry<String, Object> entry : _values.entrySet())
             {
                 strings.put(entry.getKey(), ConvertUtils.convert(entry.getValue()));
             }
-            _stringValues = strings;
+            setValuesToBind(strings);
```
labkey-nicka, line 102: If the change was made to `setValuesToBind()` to internally call `setValueToBind` for each prop/value, then this could be simplified to:

```java
@Override
public Map<String, Object> getValuesToBind()
{
    //If we don't have strings and do have typed values then
    //make the strings match the typed values
    Map<String, Object> strings = super.getValuesToBind();
    if (null == strings || strings.isEmpty() && (null != _values && !_values.isEmpty()))
    {
        setValuesToBind(_values);
        strings = super.getValuesToBind();
    }

    return strings;
}
```

api/src/org/labkey/api/data/BeanViewForm.java
labkey-matthewb: one is used to bind array-like members and one is used to bind map-like members.  E.g. for Spring url binding myfield[0]=foo  or myfield.subfield=bar

api/src/org/labkey/api/data/BeanViewForm.java
labkey-matthewb: This was (mostly) just auto-refactored.  Using setValuesToBind() here is wrong.  It has the side-effect of clearing _values which we don't want to do.  I have grown skeptical of the need to convert everything when moving values->strings.

api/src/org/labkey/api/util/HtmlString.java
labkey-matthewb: stray (used in MultiChoice rendering)

api/src/org/labkey/api/data/TableViewForm.java
labkey-matthewb: Not really??? This is existing potentially bad behavior.  I guess we don't have PKs with commas in them?  I can expand on the comment though.

api/src/org/labkey/api/data/TableViewForm.java
labkey-matthewb: No, that was just meant as a comment why it's intentionally not using setValueToBind().  I can use more words.

api/src/org/labkey/api/data/TableViewForm.java
labkey-matthewb: Probably better.  I think convert handles String[] better than Collection, so there's some value in using setValueToBind() (I don't think this happens in the current usages). 
(Note that It's not clear that there is any value in converting when copying values->strings because that value is clearly already the right type, and we don't expect it to be reconverted anyway)

api/src/org/labkey/api/data/TableViewForm.java
labkey-matthewb: Currently I'm not converting when moving objects from _values->_strings.

api/src/org/labkey/api/data/TableViewForm.java
labkey-matthewb: pv.getValue() is always String or String[] array.  But that's better.

api/src/org/labkey/api/data/TableViewForm.java
labkey-matthewb: Good catch.  Too much "cleanup".  I think getValueToBind() is the best place e.g. just before convert() in populateValues().
The previous code split the work between set() which did trim() and populateValues() which checked isEmpty().

labkey-matthewb: No test???

labkey-nicka (approved):

PR #7165, opened 2025-10-30
Add InventoryFilter. Clear orphaned issues.

labkey-adam:
#### Rationale
Refactor domain filtering to allow code sharing between tables with automatic Flag columns and extensible tables (like inventory Location, Box, and Item).

Clean up orphaned issues rows.

### DIFF
--- a/api/src/org/labkey/api/data/DatabaseMigrationService.java
+++ b/api/src/org/labkey/api/data/DatabaseMigrationService.java
@@ -9,10 +9,12 @@ import org.labkey.api.data.SimpleFilter.AndClause;
 import org.labkey.api.data.SimpleFilter.FilterClause;
 import org.labkey.api.data.SimpleFilter.InClause;
 import org.labkey.api.data.SimpleFilter.OrClause;
+import org.labkey.api.data.SimpleFilter.SQLClause;
 import org.labkey.api.query.FieldKey;
 import org.labkey.api.query.SchemaKey;
 import org.labkey.api.query.TableSorter;
 import org.labkey.api.services.ServiceRegistry;
+import org.labkey.api.util.ConfigurationException;
 import org.labkey.api.util.GUID;
 import org.labkey.api.util.logging.LogHelper;
 import org.labkey.vfs.FileLike;
@@ -30,7 +32,7 @@ public interface DatabaseMigrationService
 {
     Logger LOG = LogHelper.getLogger(DatabaseMigrationService.class, "Information about database migration");
 
-    record DomainFilter(Set<GUID> containers, String column, FilterClause condition) {}
+    record DataFilter(Set<GUID> containers, String column, FilterClause condition) {}
 
     static @NotNull DatabaseMigrationService get()
     {
@@ -84,11 +86,13 @@ public interface DatabaseMigrationService
 
         List<TableInfo> getTablesToCopy();
 
+        FilterClause getTableFilter(TableInfo sourceTable, FieldKey containerFieldKey, Set<GUID> containers);
+
         FilterClause getContainerClause(TableInfo sourceTable, FieldKey containerFieldKey, Set<GUID> containers);
 
         @Nullable FieldKey getContainerFieldKey(TableInfo sourceTable);
 
-        void addDomainDataFilter(OrClause orClause, DomainFilter filter, TableInfo sourceTable, FieldKey fKey, Set<String> selectColumnNames);
+        void addDomainDataFilter(OrClause orClause, DataFilter filter, TableInfo sourceTable, FieldKey fKey, Set<String> selectColumnNames);
 
         // Do any necessary clean up after the target table has been populated. notCopiedFilter selects all rows in the
         // source table that were NOT copied to the target table. (For example, they were filtered out due to container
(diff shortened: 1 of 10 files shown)


api/src/org/labkey/api/data/DatabaseMigrationService.java
```diff
@@ -209,37 +255,44 @@ protected void addDomainDataStandardFilter(OrClause orClause, DomainFilter filte
...
+
+            return columnExists;
         }
 
-        // Special domain data filter method for provisioned tables that have a built-in Flag field
-        protected void addDomainDataFlagFilter(OrClause orClause, DomainFilter filter, TableInfo sourceTable, FieldKey fKey, Set<String> selectColumnNames)
+        // Add a filter to select all rows where the object property with <propertyId> equals the filter value
+        protected void addObjectPropertyFilter(OrClause orClause, DataFilter filter, TableInfo sourceTable, FieldKey fKey, int propertyId)
+        {
+            SQLFragment flagWhere = new SQLFragment("lsid IN (SELECT ObjectURI FROM exp.Object WHERE ObjectId IN (SELECT ObjectId FROM exp.ObjectProperty WHERE StringValue = ? AND PropertyId = ?))", filter.condition().getParamVals()[0], propertyId);
```
labkey-matthewb, line 265: Since most exp tables have an object it, it might be worth having a parameter that indicates whether to use "lsid" or "objectid"

labkey-matthewb: This runs the risk of the database actually evaluating ((SELECT ObjectURI FROM exp.Object WHERE ObjectId IN (SELECT ObjectId FROM exp.ObjectProperty WHERE StringValue = ? AND PropertyId = ?)) in a somewhat expensive way.  It's index are optimized for objectid=? and propertyid=? joins.
Iff there is a performance issue this might be worth trying 

lsid::  WHERE ? = (SELECT StringValue FROM (O WHERE objecturi=lsid) JOIN OP ON objectid  WHERE  PropertyId=?)
or
objectid:: WHERE ? = (SELECT StringValue FROM OP WHERE  OP.ObjectId=<outer>.objectid PropertyId=?)

api/src/org/labkey/api/data/DatabaseMigrationService.java
```diff
@@ -301,4 +338,26 @@ interface MigrationFilter
         // Implementations should validate guid nullity
         void saveFilter(@Nullable GUID guid, String value);
     }
+
+    // Helper method that parses a data filter then adds it and its container to the provided collections, coalescing
+    // cases where multiple containers specify the same filter
+    static void addDataFilter(String filterName, List<DataFilter> dataFilters, Set<GUID> filteredContainers, GUID guid, String filter)
```
labkey-matthewb, line 344: Too bad filter isn't a FilterClause (but I don't know how we get to this point)

labkey-matthewb (approved):
I'm curious about what metadata/conventions we could add to our schemas to simplify this work.

api/src/org/labkey/api/data/DatabaseMigrationService.java
labkey-adam: > Since most exp tables have an object it, it might be worth having a parameter that indicates whether to use "lsid" or "objectid"

This is called to filter provisioned tables, none of which have ObjectId (that I've seen). Sample Type tables do have RowId (a material RowId) and that handler generates a custom join that takes advantage. I don't think we can generalize this method to handle all the ways we join into exp.ObjectProperty... in my current FB, I've improved the comments to make it clear this is for cases where LSID is the only option.

labkey-adam: > This runs the risk of the database actually evaluating ((SELECT ObjectURI FROM exp.Object WHERE ObjectId IN (SELECT ObjectId FROM exp.ObjectProperty WHERE StringValue = ? AND PropertyId = ?)) in a somewhat expensive way. It's index are optimized for objectid=? and propertyid=? joins. Iff there is a performance issue this might be worth trying
> 
> lsid:: WHERE ? = (SELECT StringValue FROM (O WHERE objecturi=lsid) JOIN OP ON objectid WHERE PropertyId=?) or objectid:: WHERE ? = (SELECT StringValue FROM OP WHERE OP.ObjectId=.objectid PropertyId=?)

@labkey-matthewb Here's how I've rewritten the full data class provisioned table SELECT statement (with hard-coded values instead of placeholders). Any further suggestions? (Note that because we're constructing and passing around `FilterClause`s, the first join has to be a sub-select.):
```
SELECT * FROM expdataclass.c1563d4300_protsequence dc WHERE lsid IN (SELECT ObjectURI FROM exp.Object o INNER JOIN exp.ObjectProperty op ON o.ObjectId = op.ObjectId WHERE StringValue = 'J.POD2' AND PropertyId = 70)
```

labkey-matthewb: I see.  That makes sense.  Simpler is usually better (or not worse) so the rewrite “looks” better to me (does that count as vibe coding)

labkey-adam: Or maybe you're saying this is better:
```
SELECT * FROM expdataclass.c1563d4300_protsequence dc WHERE 'J.POD2' = (
	SELECT StringValue FROM exp.ObjectProperty op INNER JOIN exp.Object o ON op.ObjectId = o.ObjectId AND ObjectURI = dc.LSID WHERE PropertyId = 70
)
```

labkey-matthewb: That’s how I would have written it by hand as a first guess (that kinda mirrors what we do for PropertyColumn),  but if you have to stuff it into a filter clause, I can see why your rewrite makes sense.   I would just keep these ideas in your back pocket if the performance is a problem.

PR #7168, opened 2025-10-31
PropertyType.MULTI_CHOICE

labkey-matthewb:
#### Rationale
First story of MultiChoice epic.  
Support PropertyType.MULTI_CHOICE and JdbcType.ARRAY
DataRegion Form  for support \<input\> and \<select multiple\>

#### Related Pull Requests
- 

#### Changes
-


api/src/org/labkey/api/exp/PropertyType.java
```diff
@@ -202,6 +205,53 @@ public Object getPreviewValue(@Nullable String prefix)
...
+        @Override
+        public SimpleTypeNames.Enum getXmlBeanType()
+        {
+            return SimpleTypeNames.STRING;
+        }
+
+        @Override
+        protected void init(PropertyRow row, Object value)
+        {
+            row.stringValue = (String)value;
```
XingY: Should this throw new UnsupportedOperationException() instead?

experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
```diff
@@ -323,7 +323,10 @@ public void registerValidatorKind(ValidatorKind validatorKind)
         if (_validatorTypes.containsKey(validatorKind.getTypeURI()))
             throw new IllegalArgumentException("Validator type : " + validatorKind.getTypeURI() + " is already registered");
 
-        _validatorTypes.put(validatorKind.getTypeURI(), validatorKind);
+        var uri = validatorKind.getTypeURI();
+        _validatorTypes.put(uri, validatorKind);
+        _validatorTypes.put(uri.substring(uri.lastIndexOf(':')+1), validatorKind);
```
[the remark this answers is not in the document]
XingY: Seems a valid comment

api/src/org/labkey/api/exp/PropertyType.java
```diff
@@ -202,6 +205,53 @@ public Object getPreviewValue(@Nullable String prefix)
...
+        @Override
+        protected Object getValue(ObjectProperty property)
+        {
+            return property.getStringValue();
+        }
+
+        @Override
+        public Object getPreviewValue(@Nullable String prefix)
+        {
+            return prefix + "Value";
```
XingY: Maybe something like "item1, item2"?

api/src/org/labkey/api/data/JdbcType.java
```diff
@@ -299,6 +300,8 @@ protected Collection<Integer> getSqlTypes()
         }
     },
 
+    ARRAY(Types.ARRAY, Array.class),
```
XingY, line 303: It's still unclear to me how are we going to support comma character in choice. TabLoader would have stripped the enclosing quotes from "a,b" and then ARRAY needs to decide if this is a single choice a,b or [a, b]. But sure that can be dealt with in a later story.

api/src/org/labkey/api/data/ColumnRenderProperties.java
```diff
@@ -168,8 +169,9 @@ else if (java.sql.Date.class.isAssignableFrom(javaClass))
             return "Date";
         else if (Date.class.isAssignableFrom(javaClass))
             return "Date and Time";
-        else
-            return "Other";
+        else if (List.class.isAssignableFrom(javaClass) || javaClass.isArray())
```
XingY, line 172: There have been inconsistent use of List or Collection when checking the data type.

api/src/org/labkey/api/data/MultiChoice.java
```diff
@@ -0,0 +1,510 @@
...
+import java.util.stream.StreamSupport;
+
+import static org.labkey.api.util.DOM.Attribute.style;
+import static org.labkey.api.util.DOM.DIV;
+import static org.labkey.api.util.DOM.SPAN;
+import static org.labkey.api.util.DOM.at;
+
+public  class MultiChoice
+{
+    public static final String ARRAY_MARKER = "[]";
```
XingY, line 49: Maybe we should start disallowing domain fields to end with [].

api/src/org/labkey/api/data/MultiChoice.java
```diff
@@ -0,0 +1,510 @@
...
+        @Override
+        public <T> T convert(Class<T> aClass, Object o)
+        {
+            if (null == o)
+                return (T) List.of();
+            if (o instanceof String s)
+                return (T) Array.from(s);
+            if (o instanceof org.json.JSONArray json)
+                return (T) Array.from(json);
+            if (o.getClass().isArray())
```
XingY, line 519: Do we need to handle Collection?

XingY (approved):

experiment/src/org/labkey/experiment/api/property/PropertyServiceImpl.java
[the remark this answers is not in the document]
labkey-matthewb: Don't need this change

api/src/org/labkey/api/data/ColumnRenderProperties.java
labkey-matthewb: List is probably better.  I'll look for those.

api/src/org/labkey/api/data/JdbcType.java
labkey-matthewb: See PageFlowUtil$TestCase.testGoogleSheetMultiValue(), the string value will be
  a, b, "A,B"
This will then get escaped according to the rules of the particular file file format.  This will look much better in a TAB separated file thans COMMA separated file.

XingY: a, b, "A,B" will parse to array.
but what about just "A,B"? TabLoader would have stripped the enclosing quotes so we still have to decide if it's A,B or [A, B]

api/src/org/labkey/api/exp/PropertyType.java
labkey-matthewb: yes.

api/src/org/labkey/api/exp/PropertyType.java
labkey-matthewb: I forgot to go back to look at all these "side" methods on this enum.  I need to see how these are used.

api/src/org/labkey/api/data/JdbcType.java
labkey-matthewb: RowId, MultiChoice
5, "a, b, ""A,B"""

labkey-matthewb: It's kinda like writing a regular expression in a .java file.  Regex has it's own escape rules and Java strings have their own escape rules.  And you have to nest them.

api/src/org/labkey/api/data/MultiChoice.java
labkey-matthewb: Maybe.  I convinced myself this was OK, because this looks like Spring's form binding syntax (e.g. Spring will interpret mycolumn[0]=1).  I could change it to be a prefix instead of a suffix like @.

api/src/org/labkey/api/data/MultiChoice.java
labkey-matthewb: I added List

PR #7169, opened 2025-10-31
Rework sequence querying & updating

labkey-adam:
#### Rationale
The previous sequences metadata query was flawed: it joined on sequence name (which is not globally unique) without qualifying by schema name. That produced confusing row duplication in cases where multiple serial columns existed on a single table (e.g., etltest.Source). Switch dialect method to return a collection of sequences since a given table could have more than one.

Add copying of assay results.

Correct Sample Type's delete to pass a list of ObjectIds (it was passing sample IDs!). Clear deleted samples from assay wells.

Fix our handling of data classes that lack a DataId column and sample types that lack a RowId.

Attempt using a small batch size for core.Documents to speed up copy.

ExperimentDeleteService allows assay results handler to delete from exp tables.

Refactor the filtering method hierarchy and get rid of `DUMMY_FIELD_KEY`. Previously, getTableSelector() would call getContainerFieldKey() and pass the result through getTableFilterClause() and getContainerClause(). This was awkward and forced the use of DUMMY_FIELD_KEY to ensure getContainerClause() overrides were called. A much more sensible approach: getTableSelector() now calls getTableFilterClause() which calls getContainerClause() which calls getContainerFieldKey(). Handlers now override at whatever level is appropriate and the dumb marker object is no longer needed.

Delete not copied MaterialIds from inventory.Item.


assay/src/org/labkey/assay/AssayResultMigrationSchemaHandler.java
```diff
@@ -0,0 +1,80 @@
...
+class AssayResultMigrationSchemaHandler extends DefaultMigrationSchemaHandler
+{
+    private static final Logger LOG = LogHelper.getLogger(AssayResultMigrationSchemaHandler.class, "Assay result migration status");
+
+    public AssayResultMigrationSchemaHandler()
+    {
+        super(DbSchema.get(AbstractTsvAssayProvider.ASSAY_SCHEMA_NAME, DbSchemaType.Provisioned));
+    }
+
+    // Provisioned assay result tables occasionally have no DataId column; hopefully they have an LSID column.
```
labkey-jeckels, line 32: If they have an LSID, I think it would be for the assay result row and not an `exp.data` row. Do we know anything about these domains without a `DataId`? Are they very old?

labkey-jeckels (approved):
See concern about `DataId` vs `LSID`

assay/src/org/labkey/assay/AssayResultMigrationSchemaHandler.java
labkey-adam: @labkey-jeckels it's a single assayresult provisioned table named `c5d21781_assayplatereplicatestats` (all other provisioned tables in the schema end with `_data_fields`). The table is empty, making it impossible to verify via data. It doesn't look old... the domain ("AssayPlateReplicateStats", unsurprisingly) and properties (18 of them) were all created 7/29/2025.

I implemented this guess as a precaution, and you likely have a better guess than I. I could just log a warning about this case and skip it. Let's discuss.

labkey-adam: Okay, I see that this is associated with `PlateReplicateStatsDomainKind`. I'll take a look at that class to understand how it's hooked up to everything else.

